PR #149, opened 2018-05-09
[WIP] add log verbosity

### LINKED ISSUE #147: support verbosity control in log info messages
Suggest using `glog.V(n).Info` and pass the verbose flag from the controller to the importer pod.

jeffvance:
fixes #147 
WIP: we need to be fairly consistent in our choice of verbose level so I'm sure there's room for improvement here.
Overall design: controller is run with `-v N` where N is the verbosity level and defaults to 3. Controller creates the importer pod with the same N. So, for now the importer pod, spawned by the controller cannot be at a different verbosity than the controller, though we could decide to support a new pvc annotation for this...
DO NOT MERGE!


cmd/controller/controller.go
```diff
@@ -31,6 +32,7 @@ func init() {
 	// flags
 	flag.StringVar(&configPath, "kubeconfig", os.Getenv("KUBECONFIG"), "(Optional) Overrides $KUBECONFIG")
 	flag.StringVar(&masterURL, "server", "", "(Optional) URL address of a remote api server.  Do not set for local clusters.")
+	flag.IntVar(&verbose, "v", 3, "(Optional) Verbosity level. Higher number provides more detailed log messages.")
```
jeffvance: not needed. glog's _init_ func does this

cmd/controller/controller.go
```diff
@@ -42,10 +44,12 @@ func init() {
 	if pp := os.Getenv(common.IMPORTER_PULL_POLICY); len(pp) != 0 {
 		pullPolicy = pp
 	}
-	glog.Infof("init: complete: CDI controller will create the %q version of the importer\n", importerImage)
+	glog.V(5).Infof("init: complete: CDI controller will create importer using image %q with verbosity %d\n", importerImage, verbose)
```
jeffvance: define constants for the levels. Define only 3 levels named: VUser=1, VAdmin=2, VDebug=3

jeffvance:
@copejon Note: many of the changes are due to `gofmt` and that I dropped the need for the _common_ pgk name for the constants.

pkg/controller/controller.go
```diff
@@ -98,7 +100,7 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
 	if !cache.WaitForCacheSync(stopCh, c.podInformer.HasSynced) {
 		return fmt.Errorf("controller.Run: Timeout waiting for pod cache sync")
 	}
-	glog.Infoln("Controller cache has synced")
+	glog.V(Vdebug).Infoln("Controller cache has synced")
```
copejon: Could probably be `Vadmin` level

pkg/controller/controller.go
```diff
@@ -145,13 +147,13 @@ func (c *Controller) ProcessNextPodItem() bool {
...
 func (c *Controller) processPodItem(pod *v1.Pod) error {
-	glog.Infof("processPodItem: processing pod named %q\n", pod.Name)
+	glog.V(Vdebug).Infof("processPodItem: processing pod named %q\n", pod.Name)
 
 	// First get the pod's CDI-relative pvc name
 	var pvcKey string
 	for _, vol := range pod.Spec.Volumes {
 		if vol.Name == DataVolName {
-			glog.Infof("processPodItem: Pod has volume matching CDI claim")
+			glog.V(Vadmin).Infof("processPodItem: Pod has volume matching CDI claim")
```
copejon: Since we're exposing it for `Vadmin`, could use the pod name in the log line

pkg/controller/util.go
```diff
@@ -65,12 +65,12 @@ func (c *Controller) podFromKey(key interface{}) (*v1.Pod, error) {
...
 	// check if we have proper AnnEndPoint annotation
 	if !metav1.HasAnnotation(pvc.ObjectMeta, AnnEndpoint) {
-		glog.Infof("checkPVC: annotation %q not found, skipping pvc\n", AnnEndpoint)
+		glog.V(Vdebug).Infof("checkPVC: annotation %q not found, skipping pvc\n", AnnEndpoint)
 		return false, nil
 	}
 	//check if the pvc is being processed
 	if metav1.HasAnnotation(pvc.ObjectMeta, AnnImportPod) {
-		glog.Infof("checkPVC: pvc annotation %q exists indicating it is being or has been processed, skipping pvc\n", AnnImportPod)
+		glog.V(Vadmin).Infof("checkPVC: pvc annotation %q exists indicating it is being or has been processed, skipping pvc\n", AnnImportPod)
```
copejon: `Vadmin` level logs probably shouldn't mention the function name

pkg/controller/util.go
```diff
@@ -80,15 +80,14 @@ func (c *Controller) checkPVC(pvc *v1.PersistentVolumeClaim, get bool) (bool, er
...
+	glog.V(Vdebug).Infof("checkPVC: getting latest version of pvc for in-process annotation")
 	latest, err := c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name, metav1.GetOptions{})
 	if err != nil {
-		glog.Infof("checkPVC: pvc Get error: %v\n", err)
 		return false, err
 	}
 	// check if we are processing this pvc now that we have the lastest copy
 	if metav1.HasAnnotation(latest.ObjectMeta, AnnImportPod) {
-		glog.Infof("checkPVC: pvc Get annotation %q exists indicating it is being or has been processed, skipping pvc\n", AnnImportPod)
+		glog.V(Vadmin).Infof("checkPVC: pvc Get annotation %q exists indicating it is being or has been processed, skipping pvc\n", AnnImportPod)
```
copejon: ditto

pkg/controller/util.go
```diff
@@ -121,13 +120,13 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
 		} else {
 			msg += "secret name is missing from annotation %q in pvc \"%s/%s\""
 		}
-		glog.Infof(msg+"\n", AnnSecret, ns, pvc.Name)
+		glog.V(Vadmin).Infof(msg+"\n", AnnSecret, ns, pvc.Name)
 		return "", nil // importer pod will not contain secret credentials
 	}
-	glog.Infof("getEndpointSecret: retrieving Secret \"%s/%s\"\n", ns, name)
+	glog.V(Vdebug).Infof("getEndpointSecret: retrieving Secret \"%s/%s\"\n", ns, name)
```
copejon: Admin's may like to see that a supplied secret was being processed

pkg/controller/util.go
```diff
@@ -121,13 +120,13 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
 	_, err := c.clientset.CoreV1().Secrets(ns).Get(name, metav1.GetOptions{})
 	if apierrs.IsNotFound(err) {
-		glog.Infof("getEndpointSecret: secret %q defined in pvc \"%s/%s\" is missing. Importer pod will run once this secret is created\n", name, ns, pvc.Name)
+		glog.V(Vuser).Infof("getEndpointSecret: secret %q defined in pvc \"%s/%s\" is missing. Importer pod will run once this secret is created\n", name, ns, pvc.Name)
```
copejon: drop func name for user level

pkg/controller/util.go
```diff
@@ -190,7 +189,7 @@ func (c *Controller) setPVCAnnotation(pvc *v1.PersistentVolumeClaim, key, val st
 
 // checks if particular label exists in pvc
 func (c *Controller) checkIfLabelExists(pvc *v1.PersistentVolumeClaim, lbl string, val string) bool {
-	glog.Info("checkIfLabelExists")
+	glog.V(Vdebug).Info("checkIfLabelExists")
```
copejon: i don't think this adds value to debugging logs

pkg/image/decompress.go
```diff
@@ -26,12 +27,12 @@ func XzReader(r io.ReadCloser) (io.ReadCloser, error) {
...
 func TarReader(r io.ReadCloser) (io.ReadCloser, error) {
-	glog.Infoln("TarReader: tar format")
+	glog.V(Vdebug).Infoln("TarReader: tar format")
 	tr := tar.NewReader(r)
 	hdr, err := tr.Next() // advance cursor to 1st (and only) file in tarball
 	if err != nil {
 		return r, fmt.Errorf("TarReader: reading tarfile header: %v\n", err)
 	}
-	glog.Infof("TarReader: extracting %q\n", hdr.Name)
+	glog.V(Vadmin).Infof("TarReader: extracting %q\n", hdr.Name)
```
copejon: We could make the logs in this file more useful by adding something like this to each compression wrapper function and removing the simple `glog.V(Vdebug).Infoln("TarReader: tar format")` style logs.

pkg/importer/dataStream.go
```diff
@@ -69,14 +69,14 @@ func (d *dataStream) dataStreamSelector() (io.ReadCloser, error) {
...
+	glog.V(Vdebug).Infoln("Using S3 client to get data")
 	bucket := d.url.Host
 	object := strings.Trim(d.url.Path, "/")
-	mc, err := minio.NewV4(common.IMPORTER_S3_HOST, d.accessKeyId, d.secretKey, false)
+	mc, err := minio.NewV4(IMPORTER_S3_HOST, d.accessKeyId, d.secretKey, false)
 	if err != nil {
 		return nil, fmt.Errorf("getDataWithS3Client: error building minio client for %q\n", d.url.Host)
 	}
-	glog.Infof("Attempting to get object %q via S3 client\n", d.url.String())
+	glog.V(Vadmin).Infof("Attempting to get object %q via S3 client\n", d.url.String())
```
copejon: The level is inconsistent w/ the http().  I definitely think an admin should be told which client is being used, so bumping a lvl in http() would be prefereable

pkg/controller/controller.go
```diff
@@ -85,7 +87,7 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
 		c.pvcQueue.ShutDown()
 		c.podQueue.ShutDown()
 	}()
-	glog.Infoln("Starting CDI controller loop")
+	glog.V(Vadmin).Infoln("Starting cdi controller Run loop")
```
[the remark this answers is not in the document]
jeffvance: 5 occurrences of "cdi" vs. 3 of "CDI". Switching to all lowercase.

pkg/controller/controller.go
jeffvance: OK, but pushing back a little :)  Why is this something an admin cares about vs. someone who needs to figure out why the controller isn't working (debugging)?

pkg/controller/controller.go
jeffvance: Agree.

pkg/controller/util.go
jeffvance: ya, probably right. Will change all Vadmin and Vuser to not mention func name.

pkg/controller/util.go
jeffvance: yes, but we don't know if the Get will succeed. I'll keep this as a debug msg and add a user level msg.

pkg/image/decompress.go
jeffvance: Did same for gz but xz does not have exported fields for name etc.

pkg/controller/controller.go
copejon: I can't really provide a specific reason, it just feels more Admin-ish :P I'm fine with leaving as it though.

pkg/image/decompress.go
```diff
@@ -8,30 +8,34 @@ import (
...
+		return nil, fmt.Errorf("GzReader: error creating gz Reader: %v\n", err)
+	}
+	glog.V(Vadmin).Infof("gzip: extracting %q\n", gz.Name)
+	return gz, nil
 }
 
 func XzReader(r io.ReadCloser) (io.ReadCloser, error) {
-	glog.Infoln("XzReader: xz format")
-	rdr, err := xz.NewReader(r, 0) //note: default dict size may be too small
+	glog.V(Vdebug).Infoln("XzReader: xz format")
```
copejon: This line isn't symmetric with GzReader or TarReader

pkg/importer/dataStream.go
```diff
@@ -180,20 +180,20 @@ func (d *dataStream) Copy(outPath string) error {
...
 func copyImage(r io.Reader, out string, qemu bool) error {
 	out = filepath.Clean(out)
-	glog.Infof("copyImage: copying image file to %q", out)
+	glog.V(Vadmin).Infof("copying image file to %q", out)
 	dest := out
 	if qemu {
 		// copy to tmp; qemu conversion will write to passed-in destination
 		dest = randTmpName(out)
-		glog.Infof(fmt.Sprintf("copyImage: temp file for qcow2 conversion: %q", dest))
+		glog.V(Vdebug).Infof(fmt.Sprintf("copyImage: temp file for qcow2 conversion: %q", dest))
```
copejon: `Sprintf` inside `Infof` is redundant

pkg/image/decompress.go
jeffvance: True. I don't see in the xz reader (or hdr) where to get the compressed file name(s). Any pointers?

copejon: oh, yes, fair point.  I don't think our xz lib has that capability

copejon:
LGTM!

PR #146, opened 2018-05-07
WIP Improve error returns

copejon:
Right now we're manually building stack traces by doing something like

```
if err != nil {
    return fmt.Errorf("myFuncName: %v", err)
}
```
to generate output where each caller wraps the error with the caller's name to create a pseudo stack trace:
```
E0508 22:24:27.064832       7 importer.go:50] main: copy error: Copy: dataStreamSelector: invalid url scheme: 
```
This is tedious and prone to typos and staleness during refactors.

In this PR, I'm introducing a custom errors package.  The functions are designed to be a 1:1 replacement for go built-in error methods (`fmt.Errorf()`, `errors.New()`) and provide a much more readable output.

Sample Output:
```
E0508 22:22:59.562419       7 importer.go:51] ERROR: invalid url scheme: ""
        Stack Trace:
        => main.main(L49)
        => github.com/kubevirt/containerized-data-importer/pkg/importer.(*dataStream).Copy(L129)
        => github.com/kubevirt/containerized-data-importer/pkg/importer.(*dataStream).dataStreamSelector(L69)
```

Here you see the call stack from `main()` down to the erroneous call and the line number the error was generated at.

PR #150, opened 2018-05-10
Provide context to errors

copejon:
Replaces PR #146 by using Dave Cheney's excellent errors package [(see his blog)](https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package).

Currently errors do not return call stacks. Instead, each return from a call wraps the returned error by doing something like

```
if err != nil {
    return fmt.Errorf("currentFunc: %v", err)
}
```
This is tedious and prone to a number of problems.  By using `github.com/pkg/errors`, we can intelligently generated error messages with readable call stacks.


cmd/controller/controller.go
```diff
@@ -48,11 +49,11 @@ func init() {
 func main() {
 	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, configPath)
 	if err != nil {
-		glog.Fatalf("Error getting kube config: %v\n", err)
+		glog.Fatalf("%+v", errors.Wrap(err, "Unable to get kube config"))
```
jeffvance: It looks like `[glog.Fatal](https://github.com/golang/glog/blob/master/glog.go#L1144)` already provides a trace

cmd/controller/controller.go
```diff
@@ -48,11 +48,11 @@ func init() {
 func main() {
 	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, configPath)
 	if err != nil {
-		glog.Fatalf("Error getting kube config: %v\n", err)
+		glog.Fatalf("Unable to get kube config: %v", err)
```
jeffvance: is `\n` not needed even though using _xxxx**f**()_?

pkg/controller/controller.go
```diff
@@ -87,16 +88,16 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
 	}()
 	glog.Infoln("Starting CDI controller loop")
 	if threadiness < 1 {
-		return fmt.Errorf("controller.Run: expected >0 threads, got %d", threadiness)
+		return errors.Errorf("expected >0 threads, got %d", threadiness)
```
jeffvance: What's the reason to use `errors.Errorf` here and `errors.New` below?

pkg/controller/controller.go
```diff
@@ -157,25 +158,26 @@ func (c *Controller) processPodItem(pod *v1.Pod) error {
...
 	glog.Infof("processPodItem: Getting PVC object for key %q", pvcKey)
 	pvc, err := c.pvcFromKey(pvcKey)
 	if err != nil {
-		return fmt.Errorf("processPodItem: error getting pvc from key: %v", err)
+		return errors.Wrapf(err, "error getting pvc object with key %q", pvcKey)
 	}
 	err = c.setPVCAnnotation(pvc, AnnPodPhase, string(pod.Status.Phase))
 	if err != nil {
-		return fmt.Errorf("processPodItem: error setting PVC annotation: %v", err)
+		return errors.Wrapf(err, "error setting annotation %q:%q in pvc %q", AnnPodPhase, pod.Status.Phase, pvc.Name)
```
jeffvance: here we call `errors.Wrapf` but on L241, after also calling setPVCAnnotation, we call `errors.WithMessage`. Why the difference?

pkg/importer/dataStream.go
```diff
@@ -95,14 +96,17 @@ func (d *dataStream) http() (io.ReadCloser, error) {
 	if len(d.accessKeyId) > 0 && len(d.secretKey) > 0 {
 		req.SetBasicAuth(d.accessKeyId, d.secretKey)
 	}
+	if err != nil {
```
jeffvance: shouldn't this err block be move to L95.1?

PR #158, opened 2018-05-16
Functional Testing:  PVC gets pod phase in annotation

copejon:


copejon:
For logging and ginkgo output, cd to pkg/controller and run

`go test -tags=unit_test -ginkgo.v -args -v 3 -alsologtostderr`

pkg/controller/controller_test.go
```diff
@@ -32,29 +31,26 @@ var _ = Describe("Controller", func() {
...
-
-	setUpInformer := func(pvc *v1.PersistentVolumeClaim, op operation) {
 
-		stop = make(chan struct{})
-		fakeClient = fake.NewSimpleClientset(pvc)
-		pvcSource := k8stesting.NewFakePVCControllerSource()
-		podSource := k8stesting.NewFakeControllerSource()
+	setUpInformers := func(pvc *v1.PersistentVolumeClaim, pod *v1.Pod, ns string, op operation, stop <-chan struct{}) {
+		// Populate the informer caches with test objects
+		if pvc != nil {
```
jeffvance: If pvc or pod are nil should we end the tests (or panic)? I know we don't want to reference nil ptrs but is this check merited?

pkg/controller/controller_test.go
```diff
@@ -63,83 +59,139 @@ var _ = Describe("Controller", func() {
...
-				Expect(err).ToNot(BeNil(), fmt.Sprintf("importer pod %s... should not exist\n", exptPod))
-			} else {
-				Expect(err).To(BeNil(), fmt.Sprintf("importer pod: %v\n", err))
-				Expect(pod).ToNot(BeNil(), fmt.Sprintf("importer pod %q missing", exptPod))
-				Expect(pod.GenerateName).To(HavePrefix(exptPod))
+
+		tests := []testT{
+			{
+				descr:       "pvc, endpoint, blank ns: controller creates importer pod",
+				ns:          "", // blank used originally in these unit tests
```
jeffvance: I _think_ @screeley44  and I agreed we don't need or want to test blank namespaces, so these cases can be deleted (and any special case code to construct the key).

pkg/controller/controller_test.go
```diff
@@ -63,83 +59,139 @@ var _ = Describe("Controller", func() {
...
+		var (
+			pvc *v1.PersistentVolumeClaim
+			//pod     *v1.Pod
+			pvcName string
+			ns      string
+			ann     map[string]string
+		)
+
+		BeforeEach(func() {
+			pvcName = "test-pvc"
```
jeffvance: maybe this can be a constant?

pkg/controller/controller_test.go
copejon: that's fine, but i'd prefer to do that in a separate PR

pkg/controller/controller_test.go
copejon: down the road yes, all the code is scratch at the moment just to debug the Update() problem

pkg/controller/controller_test.go
```diff
@@ -32,114 +31,201 @@ var _ = Describe("Controller", func() {
...
+				hasVolume:     true,
+				shouldSucceed: false,
+			},
+			{
+				desc:          "Should do nothing if there is no volume is attached to the pod",
+				podAnn:        map[string]string{AnnCreatedBy: ""},
+				hasVolume:     false,
+				shouldSucceed: false,
+			},
+		}
```
jeffvance: how about adding an "in-progress" test where expect it to do nothing?

pkg/controller/controller_test.go
```diff
@@ -32,114 +31,201 @@ var _ = Describe("Controller", func() {
...
+				shouldSucceed: true,
+			},
+			{
+				desc:          fmt.Sprintf("Should do nothing when the pod is missing annotation \"%s: %s\"", AnnCreatedBy, ""),
+				podAnn:        map[string]string{},
+				hasVolume:     true,
+				shouldSucceed: false,
+			},
+			{
+				desc:          "Should do nothing if there is no volume is attached to the pod",
```
screeley44: remove *is* - ```should do nothing if there is no volume attached to the pod```

PR #161, opened 2018-05-17
change Patch to Update

### LINKED ISSUE #160: to support testing use Update rather than Patch to add annotations and labels
fakeClient.Patch() is not implemented. No error is returned when Patch is called and the returned object is messed up. Our tests have been getting back false positives.
Update() has a likely rare window where the object's version is updated during the patch window. For now we will not worry about this.

jeffvance:
fixes #160 
Improves (hopefully) some log messages with better consistency.
Some err messages also updated.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -121,36 +121,38 @@ func (c *Controller) runPVCWorkers() {
 	}
 }
 
-// ProcessNextPodItem gets the next pod key from the queue and verifies that it was created
-// by the cdi controller. If not, the key is discarded. Otherwise, the pod object is passed
-// to processPodItem.
+// ProcessNextPodItem gets the next pod key from the queue and verifies that it was created by the
+// controller. If not the key is discarded; otherwise, the pod object is passed to processPodItem.
+// Note: pods are already filtered by label "app=containerized-data-importer".
 func (c *Controller) ProcessNextPodItem() bool {
 	key, shutdown := c.podQueue.Get()
 	if shutdown {
 		return false
 	}
 	defer c.podQueue.Done(key)
+
 	pod, err := c.podFromKey(key)
 	if err != nil {
-		c.forgetKey(key, fmt.Sprintf("Unable to get pod object: %v", err))
+		c.forgetKey(key, fmt.Sprintf("ProcessNextPodItem: unable to get pod from key %v: %v", key, err))
 		return true
 	}
 	if !metav1.HasAnnotation(pod.ObjectMeta, AnnCreatedBy) {
-		c.forgetKey(key, "Pod does not have annotation "+AnnCreatedBy)
+		c.forgetKey(key, fmt.Sprintf("ProcessNextPodItem: pod %q does not have annotation %q", key, AnnCreatedBy))
 		return true
 	}
-	if err := c.processPodItem(pod); err == nil {
-		c.forgetKey(key, fmt.Sprintf("Processing Pod %q completed", pod.Name))
+	glog.V(Vdebug).Infof("ProcessNextPodItem: next pod to process: %s\n", key)
+	err = c.processPodItem(pod)
+	if err != nil { // processPodItem errors may not have been logged so log here
+		glog.Errorf("error processing pod %q: %v", key, err)
+		return true
 	}
-	return true
+	return c.forgetKey(key, fmt.Sprintf("ProcessNextPodItem: processing pod %q completed", key))
 }
 
 // processPodItem verifies that the passed in pod is genuine and, if so, annotates the Phase
 // of the pod in the PVC to indicate the status of the import process.
 func (c *Controller) processPodItem(pod *v1.Pod) error {
-	glog.V(Vdebug).Infof("processPodItem: processing pod named %q\n", pod.Name)
-
-	// First get the pod's CDI-relative pvc name
+	// verify that this pod has the expected pvc name
 	var pvcKey string
 	for _, vol := range pod.Spec.Volumes {
 		if vol.Name == DataVolName {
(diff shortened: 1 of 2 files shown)


pkg/controller/util.go
```diff
@@ -138,55 +137,39 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
+			pvcClone, err = c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name, metav1.GetOptions{})
+			if err == nil {
+				return false, nil // try again
+			}
+			verb = "getting"
+		}
+		glog.V(Vadmin).Infof("error %s pvc \"%s/%s\": %v\n", verb, pvc.Namespace, pvc.Name, err)
+		return false, nil // try again
+	})
+	return pvc, err
```
copejon: err needs to be `errors.Wrap()`'ed

pkg/controller/util.go
```diff
@@ -138,55 +137,39 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
+	// loop a few times in case the cloned pvc is stale
+	err = wait.PollImmediate(time.Second*1, time.Second*5, func() (bool, error) {
+		pvc, err = c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Update(pvcClone)
+		if err == nil {
+			return true, nil // successful update
+		}
+		verb := "updating"
+		if apierrs.IsConflict(err) { // pvc is likely stale
+			glog.V(Vdebug).Infof("pvc \"%s/%s\" is stale, re-getting pvc\n", pvc.Namespace, pvc.Name)
+			pvcClone, err = c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name, metav1.GetOptions{})
```
jeffvance: cannot re-Get the pvc because the changed applied by the caller will be lost.

pkg/controller/util.go
```diff
@@ -138,55 +137,47 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
+		if apierrs.IsConflict(err) { // pvc is likely stale
+			glog.V(Vdebug).Infof("pvc %q is stale, re-trying\n", nsName)
+			pvcCopy, err = c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name, metav1.GetOptions{})
+			if err == nil {
+				pvcCopy.ObjectMeta = metadata // remember orig metadata
+				return false, nil             // retry update
+			}
+		}
+		glog.V(Vdebug).Infof("%q update/get error: %v\n", nsName, err)
+		return false, nil // retry
```
jeffvance: bug here: if the Get fails we retry Update with a zero-value pvc.

pkg/controller/util.go
```diff
@@ -138,55 +137,47 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
-// Note: Patch() is used instead of Update() to handle version related field changes.
-func (c *Controller) setPVCAnnotation(pvc *v1.PersistentVolumeClaim, key, val string) error {
-	glog.V(Vdebug).Infof("Adding annotation \"%s: %s\" to pvc \"%s/%s\"\n", key, val, pvc.Namespace, pvc.Name)
+	// loop a few times in case the pvc is stale
+	err = wait.PollImmediate(time.Second*1, time.Second*4, func() (bool, error) {
+		updtPvc, err = c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Update(pvcCopy)
+		if err == nil {
+			return true, nil // successful update
+		}
+		if apierrs.IsConflict(err) { // pvc is likely stale
```
jeffvance: better to not retry at all, not even have a loop here. Let the re-queue logic of the pvc work queue handle the retry.

pkg/controller/util.go
```diff
@@ -138,55 +137,47 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
-	}
-	return pvcClone, data, nil
-}
-
-// Sets an annotation `key: val` in the given PVC.
-// Note: Patch() is used instead of Update() to handle version related field changes.
-func (c *Controller) setPVCAnnotation(pvc *v1.PersistentVolumeClaim, key, val string) error {
-	glog.V(Vdebug).Infof("Adding annotation \"%s: %s\" to pvc \"%s/%s\"\n", key, val, pvc.Namespace, pvc.Name)
+	// loop a few times in case the pvc is stale
+	err = wait.PollImmediate(time.Second*1, time.Second*4, func() (bool, error) {
```
jeffvance: increase 4s to maybe 20s?

pkg/controller/controller.go
```diff
@@ -121,36 +121,38 @@ func (c *Controller) runPVCWorkers() {
...
 	key, shutdown := c.podQueue.Get()
 	if shutdown {
 		return false
 	}
 	defer c.podQueue.Done(key)
+
 	pod, err := c.podFromKey(key)
 	if err != nil {
-		c.forgetKey(key, fmt.Sprintf("Unable to get pod object: %v", err))
+		c.forgetKey(key, fmt.Sprintf("ProcessNextPodItem: unable to get pod from key %q: %v", key, err))
```
copejon: One error case is that `key` is not parsable as a string, so `%q` will print out garbage.

pkg/controller/util.go
```diff
@@ -138,55 +137,67 @@ func (c *Controller) getSecretName(pvc *v1.PersistentVolumeClaim) (string, error
...
-	if err != nil {
-		return errors.Wrap(err, "error creating pvc patch")
-	}
-	_, err = c.clientset.CoreV1().PersistentVolumeClaims(pvc.Namespace).Patch(pvc.Name, types.StrategicMergePatchType, patch)
-	if err != nil {
-		return errors.Wrapf(err, "error patching pvc %q", pvc.Name)
-	}
-	return nil
-}
+// Update and return a copy of the passed-in pvc. Only one of the annotation or label maps is requred though
```
copejon: typo in "required"

PR #163, opened 2018-05-18
refactor making util funcs more generic

### LINKED ISSUE #162: refactor: move controller funcs to exported funcs in util for more generic use with tests
rather than using controller for the client, we should pass the client to the func in an exported manner

screeley44:
fixes #162

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -190,7 +190,7 @@ func (c *Controller) ProcessNextPvcItem() bool {
 	}
 	defer c.pvcQueue.Done(key)
 
-	pvc, err := c.pvcFromKey(key)
+	pvc, err := pvcFromKey(c.pvcInformer, key)
 	if err != nil || pvc == nil {
 		return c.forgetKey(key, fmt.Sprintf("ProcessNextPvcItem: error converting key %q to pvc: %v", key, err))
 	}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -130,7 +130,8 @@ func (c *Controller) ProcessNextPodItem() bool {
 		return false
 	}
 	defer c.podQueue.Done(key)
-	pod, err := c.podFromKey(key)
+	//pod, err := c.podFromKey(key)
+	pod, err := podFromKey2(c.podInformer, key)
 	if err != nil {
 		c.forgetKey(key, fmt.Sprintf("Unable to get pod object: %v", err))
 		return true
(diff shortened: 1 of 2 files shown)


pkg/controller/util.go
```diff
@@ -36,6 +38,25 @@ func (c *Controller) pvcFromKey(key interface{}) (*v1.PersistentVolumeClaim, err
...
+		return nil, errors.New("key object not of type string")
+	}
+	obj, ok, err := informer.GetIndexer().GetByKey(keyString)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error getting key %q from cache", keyString)
+	}
+	if !ok {
+		return nil, errors.Errorf("key %q not found in cache", keyString)
+	}
+	pvc, ok := obj.(*v1.PersistentVolumeClaim)
```
copejon: this func and podFromKey2 are the same down to this point in each.  could probably reduce some duplicate code

pkg/controller/controller.go
```diff
@@ -130,7 +130,8 @@ func (c *Controller) ProcessNextPodItem() bool {
 		return false
 	}
 	defer c.podQueue.Done(key)
-	pod, err := c.podFromKey(key)
+	//pod, err := c.podFromKey(key)
+	pod, err := podFromKey(c.podInformer, key)
```
jeffvance: do we plan on any unit tests calling `podFromKey` or `pvcFromKey`? If so then they need to be exported. But my guess is no in which case they can remain methods.

pkg/controller/controller.go
```diff
@@ -165,11 +166,13 @@ func (c *Controller) processPodItem(pod *v1.Pod) error {
 		return errors.Errorf("Pod does not contain volume %q", DataVolName)
 	}
 	glog.V(Vdebug).Infof("processPodItem: Getting PVC object for key %q", pvcKey)
-	pvc, err := c.pvcFromKey(pvcKey)
+	//pvc, err := c.pvcFromKey(pvcKey)
+	pvc, err := pvcFromKey(c.pvcInformer, pvcKey)
```
jeffvance: see above comment

pkg/controller/controller.go
```diff
@@ -187,13 +190,14 @@ func (c *Controller) ProcessNextPvcItem() bool {
 	}
 	defer c.pvcQueue.Done(key)
 
-	pvc, err := c.pvcFromKey(key)
+	pvc, err := pvcFromKey(c.pvcInformer, key)
```
jeffvance: ditto

pkg/controller/controller.go
```diff
@@ -187,13 +190,14 @@ func (c *Controller) ProcessNextPvcItem() bool {
...
 	if err != nil || pvc == nil {
 		return c.forgetKey(key, fmt.Sprintf("ProcessNextPvcItem: error converting key %q to pvc: %v", key, err))
 	}
 
 	// filter pvc and decide if the importer pod should be created
-	createPod, _ := c.checkPVC(pvc, false)
+	//createPod, _ := c.checkPVC(pvc, false)
+	createPod, _ := checkPVC(c.clientset, pvc, false)
```
jeffvance: I'm not sure of the reason to convert a method to a func and not export it?  I don't know if we expect a unit test to test `checkPVC` but I guess there's a way to test private funcs by appending _Test_ to the func name...

screeley44:
Closing this in favor of #168

pkg/controller/controller.go
```diff
@@ -130,7 +130,8 @@ func (c *Controller) ProcessNextPodItem() bool {
 		return false
 	}
 	defer c.podQueue.Done(key)
-	pod, err := c.podFromKey(key)
+	//pod, err := c.podFromKey(key)
```
jeffvance: delete comment

pkg/controller/controller.go
jeffvance: keep this a method since it take an informer

pkg/controller/controller.go
jeffvance: ditto on staying a method

PR #168, opened 2018-05-21
refactor of util funcs

### LINKED ISSUE #162: refactor: move controller funcs to exported funcs in util for more generic use with tests

screeley44:
fixes #162

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -174,8 +174,8 @@ func (c *Controller) processPodItem(pod *v1.Pod) error {
 	}
 	// see if pvc's importer pod phase anno needs to be added/updated
 	phase := string(pod.Status.Phase)
-	if !c.checkIfAnnoExists(pvc, AnnPodPhase, phase) {
-		pvc, err = c.setPVCAnnotation(pvc, AnnPodPhase, phase)
+	if !checkIfAnnoExists(pvc, AnnPodPhase, phase) {
+		pvc, err = setPVCAnnotation(c.clientset, pvc, AnnPodPhase, phase)
 		if err != nil {
 			return errors.WithMessage(err, fmt.Sprintf("could not set annotation \"%s: %s\" on pvc %q", AnnPodPhase, phase, pvc.Name))
 			glog.V(Vdebug).Infof("processPodItem: pod phase %q annotated in pvc %q", pod.Status.Phase, pvcKey)
@@ -199,7 +199,7 @@ func (c *Controller) ProcessNextPvcItem() bool {
 		return c.forgetKey(key, fmt.Sprintf("ProcessNextPvcItem: error converting key %v to pvc: %v", key, err))
 	}
 	// filter pvc and decide if the importer pod should be created
-	if continue_processing, _, _ := c.checkPVC(pvc, false); !continue_processing {
+	if continue_processing, _, _ := checkPVC(c.clientset, pvc, false); !continue_processing {
 		return c.forgetKey(key, fmt.Sprintf("ProcessNextPvcItem: skipping pvc %q\n", key))
 	}
 	glog.V(Vdebug).Infof("ProcessNextPvcItem: next pvc to process: %s\n", key)
@@ -219,7 +219,7 @@ func (c *Controller) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	if err != nil {
 		return err
 	}
-	secretName, err := c.getSecretName(pvc)
+	secretName, err := getSecretName(c.clientset, pvc)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -238,14 +238,14 @@ func (c *Controller) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	}
 
 	// all checks passed, let's create the importer pod!
-	pod, err := c.createImporterPod(ep, secretName, pvc)
+	pod, err := CreateImporterPod(c.clientset, c.importerImage, c.verbose, c.pullPolicy, ep, secretName, pvc)
```
jeffvance: This now needs several controller fields passed. Does it _need_ to be a func?

pkg/controller/util.go
```diff
@@ -11,23 +11,19 @@ import (
...
 const DataVolName = "cdi-data-vol"
 
 // return a pvc pointer based on the passed-in work queue key.
 func (c *Controller) pvcFromKey(key interface{}) (*v1.PersistentVolumeClaim, error) {
-	keyString, ok := key.(string)
-	if !ok {
-		return nil, errors.New("key object not of type string")
-	}
-	obj, ok, err := c.pvcInformer.GetIndexer().GetByKey(keyString)
+	obj, err := objFromKey(c.pvcInformer, key)
```
jeffvance: can the type of obj desired be passed to `objFromKey`? That would reduce some of the similar err handling.

pkg/controller/util.go
```diff
@@ -212,20 +217,20 @@ func (c *Controller) checkIfLabelExists(pvc *v1.PersistentVolumeClaim, lbl strin
...
 		return nil, errors.Wrap(err, "importer pod API create errored")
 	}
-	glog.V(Vuser).Infof("importer pod \"%s/%s\" (image: %q) created\n", pod.Namespace, pod.Name, c.importerImage)
+	glog.V(Vuser).Infof("importer pod \"%s/%s\" (image: %q) created\n", pod.Namespace, pod.Name, image)
 	return pod, nil
 }
 
 // return the importer pod spec based on the passed-in endpoint, secret and pvc.
-func (c *Controller) makeImporterPodSpec(ep, secret string, pvc *v1.PersistentVolumeClaim) *v1.Pod {
+func makeImporterPodSpec(image, verbose, pullPolicy, ep, secret string, pvc *v1.PersistentVolumeClaim) *v1.Pod {
```
jeffvance: ditto on several controller fields passed in. And this func is not public...

pkg/controller/controller.go
screeley44: I think so, IMHO this should be a generic function that can be called from anywhere, I don't think we need the clientset though, thought I fixed that. Also I thought we discussed this one on the original PR and decided it was fine to pass in params? @copejon any thoughts on this one?

copejon: I'm in favor of it being a func.  While it's accepting a lot of controller params, those are still pretty generic values

pkg/controller/util.go
screeley44: which error handling do you mean? we are passing in an informer, how would the type of object desired be useful in this case?

pkg/controller/controller.go
screeley44: I guess I could see c.CreateImporterPod keeping as a method but MakeImporterPodSpec should probably be exported func?  thoughts?

pkg/controller/util.go
copejon: the informer is a core component of the controller, so `objFromKey` may be more suited as a method

screeley44: ack

pkg/controller/controller.go
jeffvance: For sure above @screeley44 .  I also agree w/ @copejon that these new parms are generic, so I can go either way.

pkg/controller/controller.go
```diff
@@ -174,7 +174,7 @@ func (c *Controller) processPodItem(pod *v1.Pod) error {
 	}
 	// see if pvc's importer pod phase anno needs to be added/updated
 	phase := string(pod.Status.Phase)
-	if !c.checkIfAnnoExists(pvc, AnnPodPhase, phase) {
+	if !checkIfAnnoExists(pvc, AnnPodPhase, phase) {
```
jeffvance: Why isn't `setPVCAnnotation` also converted to a func?

pkg/controller/util.go
```diff
@@ -11,23 +11,19 @@ import (
...
-	obj, ok, err := c.pvcInformer.GetIndexer().GetByKey(keyString)
+	obj, err := c.objFromKey(c.pvcInformer, key)
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error getting key %q from cache", keyString)
-	}
-	if !ok {
-		return nil, errors.Errorf("key %q not found in cache", keyString)
+		return nil, errors.Wrap(err, "could not get pvc object from key")
 	}
+
```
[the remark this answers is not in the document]
screeley44: hmmm...not sure I understand this one, go is a statically typed language right? So the type is determined at compile time, even if I used type switching, how would this work?  Maybe I'm missing something that is obvious, but either way, might need your help to explore this.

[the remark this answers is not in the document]
jeffvance: I was thinking maybe Go's [reflect](https://golang.org/pkg/reflect/) package could be used.

jeffvance:
/lgtm
squash and we'll merge

PR #172, opened 2018-05-23
remove release stanza

copejon:
Because we do not version our project, the versioned binaries are not indicative of the changes in CDI, making them essentially meaningless.  As the project progresses, a random amount of changes will just be lumped into the next binary.


jeffvance:
/lgtm
may want to quick check from @screeley44 but looks fine to me and ok to merge

PR #174, opened 2018-05-24
renaming and moving functional tests and setting up unit tests template framework

screeley44:


pkg/importer/dataStream_test.go
```diff
@@ -0,0 +1,349 @@
...
+import (
+	"io"
+	"net/url"
+	"reflect"
+	"testing"
+
+	. "github.com/kubevirt/containerized-data-importer/pkg/common"
+)
+
+func Test_dataStream_Read(t *testing.T) {
```
jeffvance: Why `Test_` here vs. `TestXyz` elsewhere?

screeley44: I think so

test/functional/controller/controller_suite_test.go
```diff
@@ -1,6 +1,6 @@
 // +build unit_test
```
jeffvance: is this still applicable after the test has moved to _functional/_?

test/functional/controller/controller_test.go
```diff
@@ -0,0 +1,291 @@
+// +build unit_test
```
jeffvance: ditto

test/framework/fileConversion.go
```diff
@@ -19,29 +19,26 @@ var formatTable = map[string]func(string) (string, error){
 
 func FormatTestData(srcFile string, targetFormats ...string) (string, error) {
 
-	var (
-		err     error
-		outFile = srcFile
-	)
+	outFile := srcFile
```
jeffvance: In my diff I added a comment to this func which you may wish to keep:
```
 // create file based on targetFormat extensions and return created file's name.
// note: intermediate files are removed.

```

test/functional/controller/controller_suite_test.go
copejon: Probably not,  they were useful early on but all build tags should probably be dropped until we come up with a standard for them.

PR #182, opened 2018-05-31
plumbing for introspection

jeffvance:
Importer refactoring to support the need to get the size of the source image but not actually copy it. Note: the `ImageSize` function is stubbed.
I have not yet created a new pkg (eg. "importer/size") but we should consider this.


pkg/importer/dataStream.go
```diff
@@ -121,75 +131,109 @@ func (d *dataStream) local() (io.ReadCloser, error) {
...
+
+// Return the endpoint size based on passed-in reader.
+func (d *dataStream) size() (int64, error) {
+	rdr := d.readers[len(d.readers)-1]
+	return Size(rdr, d.qemu)
 }
 
 // Copy the file using its Reader (r) to the passed-in destination (`out`).
-func copyImage(r io.Reader, out string, qemu bool) error {
+func Copy(r io.Reader, out string, qemu bool) error {
```
jeffvance: should this be exported?

PR #185, opened 2018-06-05
new size pkg

jeffvance:
Created new **size** package (in _pkg/lib/size/_) for consumption by kubevirt CRD.


pkg/importer/size/size.go
```diff
@@ -0,0 +1,26 @@
+package size
```
copejon: If we don't intend this package to be used by the importer, then maybe it should be located under `pkg/library/size` or something similar.

jeffvance: I thought of that too. But the size pkg seems like a subset of `importer`, similar to golang's `path` and `path/filepath` -- I was thinking.

copejon: Within CDI, I'd agree.  This package would be used by other controllers to predict the size of PVCs they generate though.  In this case, it becomes a controller level package.

Also, I feel that putting it  in the importer or the controller package may imply that it's a necessary component of them, when it isn't.

jeffvance: That's true (_I feel that putting it in the importer or the controller package may imply that it's a necessary component of them, when it isn't._). I agree _pkg/lib/*_ is more meaningful for this type of package. I'll do that and update Makefile.

Makefile
```diff
@@ -35,14 +35,15 @@ IMPT_IMG_NAME=cdi-$(IMPORTER)
 GIT_USER=$(shell git config --get user.email | sed 's/@.*//')
 TAG=$(GIT_USER)-latest
 
-.PHONY: controller importer controller-bin importer-bin controller-image importer-image push-controller push-controller-release push-importer-release push-importer clean test
-all: clean test controller importer
-pre-release: all
+.PHONY: controller importer controller-bin importer-bin controller-image importer-image push-controller push-controller-release push-importer-release push-importer lib clean test
+all: clean test controller importer lib
+#pre-release: all #why??
```
copejon: I _think_ that maybe the travis script relies on this, which still doesn't make it right.  Feel free to remove it, but check .travis.yaml if you do.

copejon:
lgtm!

PR #188, opened 2018-06-07
more work for ImageSize() plumbing

jeffvance:
More refactoring so that we can support the `ImageSize` function for use by the kubevirt CRD.
This pr implements the new `size.ImageSize()` function, which currently only supports _qcow2_ files.


pkg/importer/dataStream.go
```diff
@@ -173,10 +173,19 @@ func CopyImage(dest, endpoint, accessKey, secKey string) error {
 	return ds.copy(dest)
 }
 
-// Parse the endpoint extensions and return a slice of read-closers. The readers are in order starting with
-// the lowest level reader (eg. http) with the last entry being the highest level reader (eg, gzip).
-// Note: the .qcow2 ext is ignored however qemu files are detected based on their magic number.
-// Note: readers are not closed here.
+// Parse the endpoint extensions and return a slice of read-closers. The readers order starts with the
+// lowest level reader (eg. http). The next reader is the final reader in the chain (eg. gz). The last
```
copejon: >  The next reader is the final reader in the chain (eg. gz).

This is true when there is 1 level of compression. For more than one, this is not true.  It might be more descriptive to say that the readers are appended in the unpacking order of operation.

pkg/importer/dataStream.go
```diff
@@ -137,17 +169,25 @@ func CopyImage(dest, endpoint, accessKey, secKey string) error {
...
 	return ds.copy(dest)
 }
 
-// Parse the endpoint extensions and return a slice of read-closers. The readers are in order starting with
-// the lowest level reader (eg. http) with the last entry being the highest level reader (eg, gzip).
-// Note: the .qcow2 ext is ignored however qemu files are detected based on their magic number.
-// Note: readers are not closed here.
-func (d *dataStream) constructReaders() ([]io.ReadCloser, error) {
+// Parse the endpoint extensions and return a slice of read-closers. The readers order starts with the
+// lowest level reader (eg. http). The next reader is the final reader in the chain (eg. gz). The last
```
copejon: > The next reader is the final reader in the chain (eg. gz).

This is true when there is 1 level of compression. For more than 1, this is not true. It might be more descriptive to say that the readers are indexed in the unpacking order of operation.

pkg/importer/dataStream.go
```diff
@@ -182,31 +222,71 @@ func (d *dataStream) constructReaders() ([]io.ReadCloser, error) {
...
+	if len(readers) <= 1 {
+		// no readers to close (multi-reader does not need to be closed)
+		return nil
+	}
+	var err error
+	// move first reader to end of slice
+	r := readers[0]
+	readers = readers[1:]
+	readers = append(readers, r)
+	for _, r = range readers {
```
copejon: This closes a .tar.gz reader stack as (gz ->  tar -> http).  This is not the order the pipeline is structured (http -> gz -> tar).  Is the intent to close the readers in reverse order?

pkg/lib/size/size.go
```diff
@@ -1,26 +1,105 @@
...
+	const (
+		sizeOffset = 24 // bytes
+		fieldSize  = 8  // bytes
+		bufSize    = sizeOffset + fieldSize
+	)
+
+	buf := make([]byte, bufSize)
+	var bufLen int
+
+	// the first multi-reader Read returns only 4 bytes (magic number) so a loop is needed
```
copejon: Is this true?  The multi reader should pipe through the entire byte stream, right?

pkg/lib/size/size.go
```diff
@@ -1,26 +1,105 @@
...
+	for {
+		len, err := r.Read(buf)
```
copejon: if len != len(buf), then something unexpected happened

copejon: is it an error to use `len` as a var name?  seems like that might cause problems.

pkg/lib/size/size.go
```diff
@@ -1,26 +1,105 @@
...
+	}
+	if bufLen < bufSize {
+		return 0, errors.Errorf("unexpected qemu buffer length: %d", bufLen)
+	}
+	delta := bufLen - bufSize // amount not read into buffer (== amount read by the multi reader)
+	offset := sizeOffset - delta
+	glog.V(Vdebug).Infof("qemuSize: reading at offset %d for %d bytes\n", offset, fieldSize)
+
+	// skip offset bytes and keep the "window" into buf `fieldSize` long
+	buf = buf[offset : bufSize-delta]
```
copejon: why not `buf[offset : offset+fieldSize]`?

pkg/lib/size/size.go
```diff
@@ -1,26 +1,105 @@
...
+	s := hex.EncodeToString(buf)
+	size, err := strconv.ParseInt(s, 16, 32)
```
copejon: The function returns ( and the way the size is represented in the field) is 8 bytes.  32 should be 64 here.

pkg/importer/dataStream.go
jeffvance: +1

pkg/lib/size/size.go
```diff
@@ -1,26 +1,105 @@
...
 
 // Return the size of the endpoint corresponding to the passed-in reader.
-func Size(rdr io.ReadCloser, qemu bool) (int64, error) {
-	// TODO: figure out the size!
+func Size(readers []Reader, qemu bool) (int64, error) {
+	r := readers[len(readers)-1] // top-level reader
+	if qemu {
+		return qemuSize(r.Rdr)
+	}
+	switch r.RdrType {
```
jeffvance: The top reader is always a multi-reader so this switch doesn't work

pkg/lib/size/size.go
jeffvance: agree. Will change var name.

pkg/lib/size/size.go
```diff
@@ -56,48 +54,31 @@ func httpSize(r io.Reader) (int64, error) {
...
-		if bufLen >= bufSize {
-			break
-		}
+	l, err := io.ReadFull(r, buf)
+	if err != nil {
+		return 0, errors.Wrapf(err, "qemu ReadFull error")
 	}
-	if bufLen < bufSize {
-		return 0, errors.Errorf("unexpected qemu buffer length: %d", bufLen)
+	if l < bufSize {
```
copejon: ReadFull returns an error in this case, so the code in this block is unreachable.

PR #190, opened 2018-06-08
Containerized Build Env

copejon:
Standardize build environments by switching to a Docker container to compile project binaries.


jeffvance:
This assumes _/go/src/github.com/kubevirt/containerized-data-importer_ as the repo dir which is the best practice. Still, is there any value in parameterizing the work dir?

copejon:
Not that I can see.  In the golang image, `/go` is GOPATH, and so for any imports to work, it must be the way it's coded now.

PR #200, opened 2018-06-13
Allow non-master branch as long as there is no diff with UPSTREAM master

copejon:


hack/version/lib.sh
```diff
@@ -42,13 +42,23 @@ function verifyOnMaster(){
...
-    upstream="$upstream/master"
-    if [ -n "$(git rev-list --left-right "$upstream"...master)" ]; then
-        printf "Detected commit difference between %s and local master.  Please merge/rebase %s and retry\n" "$upstream" "$upstream"
+    if [ -z "$curBranch" ]; then
+        printf "No current branch was found, exiting\n"
+        exit 1
+    fi
+
+    if ! git fetch "$upstream" master; then
+        exit 1
```
jeffvance: no error msg here?

copejon: I've tested it and a `git fetch` error prints it's own relevant message (invalid repo, no permissions), etc.

PR #204, opened 2018-06-13
fix issue 199 - add containerized and non-containerized ability to Makefile

### LINKED ISSUE #199: Containerized only build breaks downstream
Containerizing the build process removed the ability to do a non-containerized build. Can the ability to do a non-containerized build be put back into the makefile so that the option is available?

screeley44:
fixes #199

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,14 @@ F_TEST_BIN=$(BIN)/$(F_TEST)
 # Source dirs
 CMD_DIR=$(REPO_ROOT)/cmd
 PKG_DIR=$(REPO_ROOT)/pkg
+CONTROLLER_CMD=$(CMD_DIR)/$(CONTROLLER)
+IMPORTER_CMD=$(CMD_DIR)/$(IMPORTER)
 LIB_PKG_DIR=$(PKG_DIR)/lib
 LIB_SIZE_DIR=$(LIB_PKG_DIR)/size
 F_TEST_DIR=$(REPO_ROOT)/test/functional/importer
 F_IMG_DIR=$(REPO_ROOT)/test/images/tinyCore.iso
+BUILD_CMD=GOOS=$(GOOS) GOARCH=$(ARCH) CGO_ENABLED=$(CGO_ENABLED) go build -a -ldflags $(LDFLAGS)
+DOCKER_BUILD_CMD=docker run -it --rm -v $(REPO_ROOT):$(WORK_DIR):Z -w $(WORK_DIR) -e GOOS=$(GOOS) -e GOARCH=$(ARCH) -e CGO_ENABLED=$(CGO_ENABLED) $(BUILD_IMAGE) go build
 
 # Build Dirs
 BUILD_DIR=$(REPO_ROOT)/build
@@ -32,6 +36,10 @@ CTRL_IMG_NAME=cdi-$(CONTROLLER)
 IMPT_IMG_NAME=cdi-$(IMPORTER)
 GIT_USER=$(shell git config --get user.email | sed 's/@.*//')
 TAG=$(GIT_USER)-latest
+
+# Preflight Check Defaults
+USE_DOCKER=1
+DOCKER_OUT=$(shell docker ps)
 
 .PHONY: controller importer controller-bin importer-bin controller-image importer-image push-controller push-controller-release push-importer-release push-importer lib clean test
 all: clean test controller importer lib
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,7 +58,15 @@ LDFLAGS='-extldflags "-static"'
 controller-bin:
 	@echo '********'
 	@echo 'Compiling controller binary'
-	docker run -it --rm -v $(REPO_ROOT):$(WORK_DIR):Z -w $(WORK_DIR) -e GOOS=$(GOOS) -e GOARCH=$(ARCH) -e CGO_ENABLED=$(CGO_ENABLED) $(BUILD_IMAGE) go build -a -ldflags $(LDFLAGS) -o $(WORK_DIR)/bin/$(CONTROLLER_BIN) $(WORK_DIR)/cmd/controller/controller.go
+	$(eval DOCKER_CMD_COUNT=$(shell docker ps | wc -c))
```
jeffvance: I think it is sufficient to just check if `docker ps` works. There do not have to be running containers in order for us to decide to use a containerized make.

jeffvance:
Where is the non-containerized build for `importer-bin` and the containerized build for `test-bin`?

Makefile
```diff
@@ -33,6 +34,10 @@ IMPT_IMG_NAME=cdi-$(IMPORTER)
 GIT_USER=$(shell git config --get user.email | sed 's/@.*//')
 TAG=$(GIT_USER)-latest
 
+# Preflight Checks
+USE_DOCKER=1
+DOCKER_CMD_COUNT=0
```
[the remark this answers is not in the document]
screeley44: how do you plan on evaluating `docker ps` ? if docker isn't running the wc returns 0 if it is it's > 0, and I need a way to pass that back into my evaluation. If I use exit code, I'm still going to have to create a variable to store that information and access again won't I?

Makefile
```diff
@@ -53,7 +58,15 @@ LDFLAGS='-extldflags "-static"'
 controller-bin:
 	@echo '********'
 	@echo 'Compiling controller binary'
-	docker run -it --rm -v $(REPO_ROOT):$(WORK_DIR):Z -w $(WORK_DIR) -e GOOS=$(GOOS) -e GOARCH=$(ARCH) -e CGO_ENABLED=$(CGO_ENABLED) $(BUILD_IMAGE) go build -a -ldflags $(LDFLAGS) -o $(WORK_DIR)/bin/$(CONTROLLER_BIN) $(WORK_DIR)/cmd/controller/controller.go
+	$(eval DOCKER_CMD_COUNT=$(shell docker ps | wc -c))
+	export DOCKER_CMD_COUNT
```
copejon: L61 should have set this var in the env already

Makefile
```diff
@@ -53,7 +58,15 @@ LDFLAGS='-extldflags "-static"'
...
+	@if [ $(DOCKER_CMD_COUNT) -eq 0 ] || [ $(USE_DOCKER) -eq 0 ]; then \
```
copejon: it _may_ be cleaner to use make if/else blocks rather than in shell.

Makefile
screeley44: not that I can access from within that little context of the makefile, the only way I could get DOCKER_CMD_COUNT was to export, it's strange, but basically if I don't export and then try to access that var, I get empty

Makefile
screeley44: this does not require containers to be present, if docker is not running it evaluates correctly, if docker is not installed it evaluates correctly, and it does not need containers to be installed.

Makefile
screeley44: I've tried, and the `ifeq else endif` blocks didn't work, they errored and/or they never evaluated the vars correctly. that is why I moved to the shell, but probably the bigger issue I didn't like is there is no complex conditional support (if AND/OR) so we would have to nest our ifs

Makefile
```diff
@@ -33,6 +35,12 @@ IMPT_IMG_NAME=cdi-$(IMPORTER)
 GIT_USER=$(shell git config --get user.email | sed 's/@.*//')
 TAG=$(GIT_USER)-latest
 
+# Preflight Check Defaults
+USE_DOCKER=1
+DOCKER_CMD_COUNT=0
+$(eval DOCKER_CMD_COUNT=$(shell docker ps | wc -c))
+export DOCKER_CMD_COUNT
```
copejon: mixing make and shell syntax

Makefile
```diff
@@ -33,6 +39,11 @@ IMPT_IMG_NAME=cdi-$(IMPORTER)
 GIT_USER=$(shell git config --get user.email | sed 's/@.*//')
 TAG=$(GIT_USER)-latest
 
+# Preflight Check Defaults
+USE_DOCKER=1
+$(eval DOCKER_OUT=$(shell docker ps))
```
copejon: This mixes make syntax (everything left justified) and bash syntax (everything under a target indented by 1).

To set the variable globally in the Makefile, you can do
DOCKER_OUT=$(shell docker ps)
DOCKER_OUT != docker ps

Makefile
```diff
@@ -33,6 +39,11 @@ IMPT_IMG_NAME=cdi-$(IMPORTER)
...
+export DOCKER_OUT
```
copejon: don't need to export if set as defined above

copejon: also, not sure this would work, since it's bash syntax

screeley44: doesn't like the DOCKER_OUT != docker ps

PR #209, opened 2018-06-14
Reinstate travis deployment, but only on tag pushes

copejon:
Travis will only deploy versioned releases when a new git tag is pushed.


jeffvance:
lgtm

PR #215, opened 2018-06-18
make set-version: do global sed replacements of versions

copejon:
Add 'g' to sed command for global substitutions


jeffvance:
lgtm

PR #222, opened 2018-06-25
Discard docker ps output and capture exit code

copejon:
`docker ps` prints some value to stdin/stderr regardless of whether it passes, making it inadequate for testing if docker is running.  This PR captures the exit code instead to `DOCKER_OUT`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ TAG=$(GIT_USER)-latest
 
 # Preflight Check Defaults
 USE_DOCKER=1
-DOCKER_OUT=$(shell docker ps)
+RUNNING_DOCKER=$(shell docker ps &>/dev/null; echo $$?)
 
 .PHONY: controller importer controller-bin importer-bin controller-image importer-image push-controller push-controller-release push-importer-release push-importer lib clean test
 all: clean test controller importer lib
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,7 +39,7 @@ TAG=$(GIT_USER)-latest
 
 # Preflight Check Defaults
 USE_DOCKER=1
-DOCKER_OUT=$(shell docker ps)
+RUNNING_DOCKER=$(shell docker ps &>/dev/null; echo $$?)
```
jeffvance: might be nice if you could invert this value. But likely tricky in `make`... Thinking of something like:
```
RUNNING_DOCKER=1 - $(shell docker ps &>/dev/null; echo $$?)  # probably wrong syntax

```
Then below both vars can be tested for 1...

Makefile
```diff
@@ -89,6 +89,10 @@ func-test-bin:
 	GOOS=$(GOOS) GOARCH=$(ARCH) CGO_ENABLED=$(CGO_ENABLED) go test -a -c -ldflags $(LDFLAGS) -o $(F_TEST_BIN) $(F_TEST_DIR)/*.go
 
 # build the controller image
+ifeq ($(RUNNING_DOCKER), 1)
```
jeffvance: good catch here

jeffvance:
lgtm

Makefile
copejon: That seems like it would be confusing.  If we doing the command in line, it would be 

`$ if [ $(docker ps &>/dev/null; echo $?) ]; then echo is-true; fi`
If running docker, the exit code would be 0 and it would print `is-true`.

PR #224, opened 2018-06-26
Build and deploy resources in CI, release images on deploy

copejon:
Images should be built and tested in the `script:` stanza of travis.  If the objects deploy as expected and the build is for a release tag, release the images.


copejon:
For whatever reason, Travis is attempting to test a commit that is not part of this PR.  I can't get it to correct itself.  Closing and reopening.

PR #225, opened 2018-06-26
Ci release images

copejon:
Images should be built and tested in the script: stanza of travis. If the objects deploy as expected and the build is for a release tag, release the images.


jeffvance:
lgtm

PR #227, opened 2018-06-29
compute original size for .iso files

### LINKED ISSUE #226: The original size of .gz and .xz files is not available
pr #220 provides the size for _iso.tar_, _iso.tar.gz_, _iso.tar.xz_, _.qcow2_, _qcow2.tar_, _qcow2.gz_,  _qcow2.tar.gz_, _qcow2.xz_ and _qcow2.tar.xz_ files based on header metadata. These file formats are supported because the inner most "wrapper" is either tar or qcow2 and both of these formats contain size metadata. However, _.iso_, _iso.gz_, and _iso.xz_ files are excluded because there is no way to get the original size without copying the entire file. (Note: we may be able to get the size of _https://foo.iso_ via the http response header -- more investigation is needed into http/s and s3.) 

It seems risky to temporarily copy the endpoint in order to get its size. For example, cluster resources (cpu, mem, n/w bandwith, node disk, etc) may become exhausted, or the temp copy may fail due to quota restrictions. However, at this point I do not know how to reliably calculate the size of .gz or .xz files without decompressing them.

Gzip info: [rfc 1952](https://tools.ietf.org/html/rfc1952#page-5) and from _man_:
>  BUGS
    According to RFC 1952, the recorded file size is stored in a 32-bit
    integer, therefore, it can not represent files larger than 4GB. This
    limitation also applies to -l option of gzip utility.

Note: the size of a gzip file is contained in the last 4 bytes of the file content so the entire file needs to be read.

From .xz [man](https://www.systutorials.com/docs/linux/man/1-xz/):
> No size information is stored in block headers, thus files created in single-threaded mode won't be identical to files created in multi-threaded mode.

jeffvance:
Fixes #226 

pr #220 set the framework for this pr. If, after resolving all readers needed to read the endpoint, the file's size is still 0 then `isoSize` is called to calculate the byte size of the original iso file.

Note: this pr understates the size of the _tinyCore.iso_ file by appox 710k bytes (~347 sectors). But, note that even the `ioutil` utility shows the size for the same file to be 8869 * 2048 = 17,862,166; whereas, `ls -l ` shows 18,874,368. So `ioutil` also understates the size by 1,012,202 bytes (~494 sectors) [1].  So we're a bit closer ;-]. Perhaps there's a fudge factor that needs to be applied (to really small ISOs) but I haven't come across that yet...

See UPDATE below where the computed size of a larger .iso file is exact!

@copejon @screeley44 @davidvossel 
[1] :
```
$ ls tinyCore.iso -l
-rw-r--r--. 1 root root 18874368 Jun 27 15:03 tinyCore.iso

$ isoinfo -debug -d -i tinyCore.iso
CD-ROM is in ISO 9660 format
System id: LINUX
Volume id: TinyCore
Volume set id: 
Publisher id: 
Data preparer id: 
Application id: MKISOFS ISO9660/HFS/UDF FILESYSTEM BUILDER & CDRECORD CD/DVD/BluRay CREATOR (C) 1993 E.YOUNGDALE (C) 1997 J.PEARSON/J.SCHILLING
Copyright File id: 
Abstract File id: 
Bibliographic File id: 
Volume set size is: 1
Volume set sequence number is: 1
Logical block size is: 2048    <=========== sector size
Volume size is: 8869           <=========== num of sectors (multiply)
Root directory extent:  29 size: 2048
Path table size is:     66
L Path table start:     21
L Path opt table start: 0
M Path table start:     23
M Path opt table start: 0
...
File structure version: 1
...
```


screeley44:
@jeffvance - maybe the differences could be from padding from the writing software and other little nuances? Not really sure, but I would think that you are roughly 3% off actual size on disk, maybe it's ok to assume whatever you get from your routine you add an additional 5-10% as return value? If I'm following this correctly, this will always be an estimation, even `isoinfo` is an approximate apparently, as long as we are not returning a value that is smaller than actual size I don't see the harm.

jeffvance:
UPDATE: using a much larger _.iso_ file (windows) this pr calculates the correct source size:
@jcope @screeley @erinboyd @alitke I rebased pr 227 ("iso size") and ran tests on the much larger windows _.iso_ file:
```
...
endpoint "file:///en_windows_server_2018_x64_dvd_11636692.iso"'s computed byte size: 6006587392    <==== see ls below
I0706 17:33:12.234782   14450 dataStream.go:474] copying image file to "/data/disk.img"
I0706 17:33:12.235087   14450 util.go:49] begin import...
I0706 17:33:45.845734   14450 importer.go:42] import complete

$ ls /data/ -l
-rwxr-xr-x. 1 root root 6006587392 Jul  6 17:33 disk.img  <======= same!
```

PR #246, opened 2018-07-09
Refactor makefile into scripts

copejon:
The makefile is becoming unmaintainable and unreadable as recipes have been added for very granular tasks.  This PR adds some generality via bash scripts to allow for precise targeting of packages without having to write a recipe for each one.

Non-docker builds must now be explicitly set in make e.g. `make DOCKER=1 build`, otherwise, the containerized build env will be used.


copejon:
Functional tests currently fail because the path they access for test images is invalid.

jeffvance:
Here we go again :)
> `make DOCKER=1 build`
To me, setting DOCKER=1 means setting it to true. I know bash exit codes are zero for success/true but other bash constructs use 1 to mean "set", eg `((DOCKER)) && do-it`

Makefile
```diff
@@ -1,261 +1,33 @@
...
-
-.PHONY: deploy-controller
-deploy-controller: $(REPO_ROOT)/manifests/controller/cdi-controller-deployment.yaml
-	sed -E -e 's#kubevirt/cdi-controller.*#cdi-controller#g' -e 's#imagePullPolicy:.*#imagePullPolicy: Never#g' $(REPO_ROOT)/manifests/controller/cdi-controller-deployment.yaml | kubectl apply -f -
-
-.PHONY: patch-controller
-patch-controller:
-	kubectl patch deployment cdi-deployment --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/env", "value": [{"name": "IMPORTER_IMAGE", "value": "cdi-importer"}]}]'
+.PHONY: vet
+vet:
```
jeffvance: what is `vet`?

bin/.gitignore
```diff
@@ -1 +1 @@
-*
\ No newline at end of file
+./*
```
jeffvance: do you mean for .gitignore to be part of this pr?

build/importer/Dockerfile
```diff
@@ -1,16 +1,19 @@
...
-RUN apk update && apk add qemu-img \
-  xz \
-  gzip
-RUN apk add ca-certificates && rm -rf /var/cache/apk/*
+RUN dnf install -y qemu-img xz gzip
+RUN mkdir -p /gimme && curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=/gimme bash >> /etc/profile.d/gimme.sh
+ENV GOPATH="/go" GOBIN="/usr/bin"
+RUN \
+    mkdir -p /go && \
+    source /etc/profile.d/gimme.sh
```
jeffvance: should `-v=` be an arg to ENTRYPOINT below?

hack/build/build-go.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+
+set -e
+
+source hack/build/common.sh
```
jeffvance: so we assume user is always at repo root...

hack/build/build-go.sh
```diff
@@ -0,0 +1,86 @@
...
+fi
+
+# If no binaries are given, assume all
+if [ $# -gt 0 ]; then
+    targets=$@
+else
+    targets=$BINARIES
+fi
+
+if [ "$go_opt" == "test" ]; then
```
jeffvance: might be cleaner/easier-to-maintain if these `if`s become a `case`?

hack/build/build-go.sh
```diff
@@ -0,0 +1,86 @@
...
+
+            GOOS=${BUILD_OS} GOARCH=${BUILD_ARCH} go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${LINUX_NAME} -ldflags '-extldflags "static"'
+
+            ln -sf ${LINUX_NAME} ${CDI_DIR}/bin/${LINUX_NAME}
+        )
+    done
+fi
+
+# If no pgk paths are given, test or execute go $target for pkgs
+if [ $# -eq 0 ]; then
```
jeffvance: seem like we're testing either # args or ${go_opt} several times. Is there a clean way to consolidate/DRY these checks?

hack/build/docker/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM fedora:27
```
jeffvance: why fedora 27 here and f28 elsewhere?

jeffvance:
very nice! Will definitely improve `make`

build/importer/Dockerfile
copejon: Right now, there is nothing written to build the controller or importer images, so that's TBD.  I'm  going to simplify their images into something a little more flexible.

hack/build/build-go.sh
copejon: The assumption here is that it's being executed via the Makefile, so the user has to be in the root dir.

hack/build/build-go.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+
+set -e
+
+source hack/build/common.sh
+source hack/build/config.sh
+
+if [ -z $1 ]; then
```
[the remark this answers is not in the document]
copejon: right, thanks!

hack/build/build-go.sh
```diff
@@ -0,0 +1,86 @@
...
+# If no binaries are given, assume all
+if [ $# -gt 0 ]; then
+    targets=$@
+else
+    targets=$BINARIES
+fi
+
+if [ "$go_opt" == "test" ]; then
+    if [ $# -eq 0 ]; then
+    (
```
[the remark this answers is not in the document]
copejon: `./...` means recurse downward from here to the go tool

hack/build/docker/Dockerfile
copejon: Ah, I think I saw that kubevirt was using 27.  I'll unify it throughout.  The controller and importer images are yet to get overhauled.

hack/build/build-go.sh
```diff
@@ -0,0 +1,86 @@
...
+        cd ${CDI_DIR}/pkg
+        go test -v ./...
+    )
+    (
+        cd ${CDI_DIR}/test
+        go test -v ./...
+    )
+    else
+        for tgt in ${targets}; do
```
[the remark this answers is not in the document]
copejon: as it is, $targets only specifies the binary outputs by default.  I can refactor it to assign pkg, test, and cmd to targets, then iterate over that

hack/build/build-go.sh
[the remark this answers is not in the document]
jeffvance: I mean the () parenthesis which I believe create a sub-shell... Is this a make requirement?

hack/build/build-go.sh
```diff
@@ -19,7 +19,7 @@ else
     targets=$BINARIES
 fi
 
-# If no pgks paths are given, test or execute go $target for pkgs
+# If no pgk paths are given, test or execute go $target for pkgs
```
jeffvance: type: "pkg"

hack/build/build-go.sh
```diff
@@ -1,86 +1,49 @@
...
-            go $go_opt ./...
-        )
-        (
-            cd ${CDI_DIR}/test
-            go $go_opt ./...
-        )
+    if [ -z "${targets}" ]; then
+        targets="${BINARIES}"
     fi
+	for tgt in ${targets}; do
```
jeffvance: +1

Makefile
```diff
@@ -1,4 +1,4 @@
-DOCKER=0
+DOCKER=1
```
jeffvance: +1

hack/build/version.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+
+
+function kubevirt::version::get_version_vars() {
```
jeffvance: why _kubevirt_ prefix instead of, say, _cdi::_ ?

copejon: this is ... inspired... by version.sh in kubevirt/kubevirt :)  Just began testing it before altering it for CDI's use.  Honestly, it seems like overkills for determining whether it's a dirty branch or not.

old.Makefile
```diff
@@ -0,0 +1,218 @@
+REPO_ROOT=$(abspath $(dir $(lastword $(MAKEFILE_LIST))))
```
aglitke: Any real reason to keep this file around?  It exists in the git history.

hack/README.md
```diff
@@ -1,60 +1,60 @@
...
-`glide install -v`
+`$ make all` executes the full workflow.  For granular control of the workflow, several Make targets are defined:
 
-`glide install` scans imports and resolves missing and unused dependencies. `-v` removes nested vendor and Godeps/_workspace directories.
+- `all`: cleans up previous build, then compiles all CDI packages and builds containers
+- `clean`: cleans up previous build artifacts
+- `build`: compile all CDI binary artifacts
+    - `build-controller`: compile cdi-controller binary
+    - `build-importer`: compile cdi-importer binary
+- `test`: execute all tests
```
jeffvance: still true?

hack/README.md
```diff
@@ -1,60 +1,60 @@
...
+    - `test-unit`: execute all tests under `./pkg`
+    - `test-functional`: execute all tests under `./test`
+- `docker`: compile all binaries and build all containerized
+    - `docker-controller`: compile cdi-controller and build cdi-controller image
+    - `docker-importer`: compile cdi-importer and build cdi-importer image
+    - `docker-cloner`: build the cdi-cloner image ( cloner is driven by a shell script, not a binary )
+- `vet`: lint all CDI packages
```
jeffvance: still true?

hack/README.md
```diff
@@ -1,60 +1,60 @@
...
+    Docker builds may be disabled by setting DOCKER=0; e.g.
+    $ make all DOCKER=0
 
-`glide install -v`
+`$ make all` executes the full workflow.  For granular control of the workflow, several Make targets are defined:
 
-`glide install` scans imports and resolves missing and unused dependencies. `-v` removes nested vendor and Godeps/_workspace directories.
+- `all`: cleans up previous build, then compiles all CDI packages and builds containers
+- `clean`: cleans up previous build artifacts
+- `build`: compile all CDI binary artifacts
```
jeffvance: is `vet` still part of this recipe?

hack/README.md
```diff
@@ -1,60 +1,60 @@
...
+    - `build-controller`: compile cdi-controller binary
+    - `build-importer`: compile cdi-importer binary
+- `test`: execute all tests
+    - `test-unit`: execute all tests under `./pkg`
+    - `test-functional`: execute all tests under `./test`
+- `docker`: compile all binaries and build all containerized
+    - `docker-controller`: compile cdi-controller and build cdi-controller image
+    - `docker-importer`: compile cdi-importer and build cdi-importer image
+    - `docker-cloner`: build the cdi-cloner image ( cloner is driven by a shell script, not a binary )
```
jeffvance: super nit: leading blank between "(" and "cloner"

hack/README.md
```diff
@@ -1,60 +1,60 @@
...
+- `vet`: lint all CDI packages
+- `format`: Execute `shfmt`, `goimports`, and `go vet` on all CDI packages.  Writes back to the source files.
```
jeffvance: +1!

Makefile
```diff
@@ -1,261 +1,66 @@
...
 
-# Binary Path
-BIN=$(REPO_ROOT)/bin
-CONTROLLER_BIN=import-controller
-IMPORTER_BIN=importer
-F_TEST_BIN=$(BIN)/$(F_TEST)
-U_TEST_BIN=$(BIN)/$(U_TEST)
-U_TEST_BIN_CONTROLLER=$(BIN)/$(U_TEST_CONTROLLER)
-U_TEST_BIN_IMAGE=$(BIN)/$(U_TEST_IMAGE)
+all: clean docker
```
jeffvance: do you want _docker_ to be the recipe name here? I liked _image_ from the orig Makefile, or maybe something with _container_ in the name, ie not docker specific? What if later we want cri-o containers built?

Makefile
```diff
@@ -1,261 +1,66 @@
-REPO_ROOT=$(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+DOCKER=1
```
jeffvance: see _docker_comment below.
Maybe the var should be named `CONTAINER`?

hack/build/build-docker.sh
```diff
@@ -0,0 +1,39 @@
...
+            if [ "${tgt}" == "${CLONER}" ]; then
+                docker build -t ${DOCKER_REPO}/${tgt}:${DOCKER_TAG} .
+            else
+                rm -rf ${tmp_dir}
+                mkdir -p ${tmp_dir}
+                cp Dockerfile ${tmp_dir}/
+                cp "${CMD_OUT_DIR}/${tgt}/${tgt}" ${tmp_dir}/
+                cd ${tmp_dir}/
+                docker build -t ${DOCKER_REPO}/${tgt}:${DOCKER_TAG} .
+                rm -rf ${tmp_dir}
```
jeffvance: just a note that cwd still points to this just-deleted dir.

hack/build/build-docker.sh
```diff
@@ -0,0 +1,39 @@
...
+            # Handle cdi-cloner special case - it has no build artifact, just the script.sh
+            if [ "${tgt}" == "${CLONER}" ]; then
+                docker build -t ${DOCKER_REPO}/${tgt}:${DOCKER_TAG} .
+            else
+                rm -rf ${tmp_dir}
+                mkdir -p ${tmp_dir}
+                cp Dockerfile ${tmp_dir}/
+                cp "${CMD_OUT_DIR}/${tgt}/${tgt}" ${tmp_dir}/
+                cd ${tmp_dir}/
+                docker build -t ${DOCKER_REPO}/${tgt}:${DOCKER_TAG} .
```
jeffvance: symmetry nit: why hard-coded `build` verb here but you reference`${docker_opt}` in the docker command for _push_ below?

Makefile
```diff
@@ -1,261 +1,66 @@
...
+	./hack/build/build-docker.sh build ${WHAT}
+
+docker-controller: WHAT = cdi-controller
+docker-controller: docker
+docker-importer: WHAT = cdi-importer
+docker-importer: docker
+docker-cloner: WHAT = cdi-cloner
+docker-cloner: docker
+
+publish: docker
```
jeffvance: +1!

Makefile
```diff
@@ -1,261 +1,66 @@
...
+	./hack/build/build-docker.sh push ${WHAT}
+
+vet:
```
jeffvance: +1!

hack/build/build-go.sh
```diff
@@ -0,0 +1,67 @@
...
+        # To be fixed by https://github.com/kubernetes/kubernetes/pull/60584
+        targets=$(sed "s,kubevirt.io/containerized-data-importer,${CDI_DIR},g" <(go list ./... | grep -v -E "vendor|pkg/client" | sort -u ))
+    fi
+    for tgt in ${targets}; do
+        (
+            cd "${tgt}"
+            go vet -v ./...
+        )
+    done
+else # Pass go commands directly on to packages expect vendor
```
jeffvance: typo _except_ (I think?)

old.Makefile
copejon: Just using it as a reference for build targets, but no, it'll go away before we merge

hack/version/shift-tag.sh
```diff
@@ -1,27 +0,0 @@
-#!/usr/bin/env bash
```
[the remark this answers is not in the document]
copejon: yes, it's just bad practice and will not be used

Makefile
```diff
@@ -1,261 +1,91 @@
...
-unit-test-bin:
-	@echo '********'
-	@echo 'Compiling unit test binary'
-	-rm -f $(U_TEST_BIN)
-	-rm -f $(U_TEST_BIN_CONTROLLER)
-	-rm -f $(U_TEST_BIN_IMAGE)
-	GOOS=$(GOOS) GOARCH=$(ARCH) CGO_ENABLED=$(CGO_ENABLED) go test -c -v -tags=unit_test ./pkg/controller -a -ldflags $(LDFLAGS) -o $(U_TEST_BIN_CONTROLLER) $(U_TEST_DIR_ALL)/*_test.go
-	GOOS=$(GOOS) GOARCH=$(ARCH) CGO_ENABLED=$(CGO_ENABLED) go test -c -v -tags=unit_test ./pkg/image -a -ldflags $(LDFLAGS) -o $(U_TEST_BIN_IMAGE) $(U_TEST_DIR_ALL)/*_test.go
+clean:
+	./hack/build/in-docker "./hack/build/build-go.sh clean && rm -rf bin/* _out/*"
```
aglitke: We always clean in docker even if DOCKER=0?  Will that work?

copejon: The reason is that when built with a container, the files will have `root:root` ownership, since the container runs as root internally.  This means that if make clean is executed in the host env, not in a container, it won't have privilege to delete the files.  (ownership translation is apparently something people have been asking docker for for years)

That said, I'm not against adding the non-dockerized version.  It will succeed the build is only ever performed with DOCKER=0.

jeffvance:
lgtm

PR #255, opened 2018-07-12
Fix improperly exported field in test struct

copejon:
The datastream implementations in datastream_test.go misdefine the datastream's `qemu` field as `Qemu`, causing the tests to fail.   This PR renames the field to `qemu` to fix conflicts with the datastream struct definition.


jeffvance:
lgtm

PR #256, opened 2018-07-13
[wip] Size() unit test

jeffvance:
First pass at `lib.Size()` unit tests.
TODO:
  - containerized tests with target iso files
  - discuss acceptable range of sizes still considered "success"
  - add more test cases
  - add functional tests

@copejon @screeley44


jeffvance:
2nd commit allows for functional test exceptions to known Size errors on certain image files w/o failing the test.
Issue remaining:
. there are 2 _suite_ files in _test/functional/importer/_ and a result of this is that the dataStream tests are run twice, and (for some reason) the importer suite is not run at all.
. Size() should return the correct size for _.iso.gz_ and _.iso.xz_ image files but these 2 tests fail.

pkg/lib/size/size_test.go
```diff
@@ -0,0 +1,55 @@
...
+	tests := []struct {
+		name  string
+		args  args
+		low   int64 // lowest acceptable size due to variations in iso hdrs
+		exact int64
+		high  int64 // highest acceptable size due to variations in iso hdrs
+	}{
+		{
+			name:  "small tinyCore.iso file size",
+			args:  args{"file:///root/go/src/kubevirt.io/containerized-data-importer/tinyCore.iso", "", ""},
```
copejon: these paths should be set dynamically using the `os` package and made relative to the test file rather than absolute.

pkg/lib/size/size_test.go
```diff
@@ -0,0 +1,55 @@
...
+			low:   17860000,
+			exact: 18874368,
+			high:  18900000,
+		},
+		{
+			name:  "large windows .iso.xz file size",
+			args:  args{"file:///root/go/src/kubevirt.io/containerized-data-importer/en_windows_server_2016_updated_feb_2018_x64_dvd_11636692.iso.xz", "", ""},
```
copejon: where does this file come from?  Also, if it's being downloaded at runtime, why into the repo root?

pkg/lib/size/size_test.go
copejon: e.g.
```
	srcIso, err := filepath.Abs("../../../test/images/tinycore.iso")
	if err != nil {
		t.Fatalf("Failed to get source image path: %s\n", err)
	}
```

pkg/importer/dataStream.go
```diff
@@ -411,6 +411,7 @@ func (d *dataStream) isoSize() (int64, error) {
 	// ensure we have an iso file by checking the type and id value
 	vdtyp, err := strconv.Atoi(hex.EncodeToString(buf[typeOff : typeOff+typeLen]))
 	if err != nil {
+		glog.Errorf("isoSize: Atoi error on endpoint %q: %v", d.Url.Path, err)
```
copejon: `err` should be wrapped w/ the errors package for stack trace consistency.

pkg/lib/size/size_test.go
jeffvance: agree. I forgot to delete some debugging test case.

pkg/lib/size/size_test.go
jeffvance: same debugging related mistake

copejon:
would you mind removing https://github.com/jeffvance/containerized-data-importer/blob/10e4728769193db408bd814302928707c1a2b4a3/test/framework/fileConversion.go#L80
and 
 https://github.com/jeffvance/containerized-data-importer/blob/10e4728769193db408bd814302928707c1a2b4a3/test/framework/fileConversion.go#L82

test/functional/importer/importer_suite_test.go
```diff
@@ -1,4 +1,4 @@
-package importer_test
+package importer
```
copejon: Do we want 2 packages in this directory (`importer_test` && `importer`)?

PR #257, opened 2018-07-13
[WIP] Functional Test to use relative base image path

copejon:
Currently, functional tests require a specialized container environment where the test is compiled into a binary and copied to `/` along with an expected base iso.  This puts too much responsibility on `make` and disallows devs from running the tests in a host env.  Additionally, it requires a make recipe for every single functional test individually, which is not extensible.

This PR refactors the datastream functional test to access the included ISO directly and to write test images to a known tmp directory.  This makes the test solely responsible for setting up it's own dependencies and decouples it from the build workflow entirely.

### DIFF
--- a/test/functional/importer/datastream_test.go
+++ b/test/functional/importer/datastream_test.go
@@ -19,6 +19,7 @@ import (
 	imagesize "kubevirt.io/containerized-data-importer/pkg/lib/size"
 	f "kubevirt.io/containerized-data-importer/test/framework"
 	"time"
+	"kubevirt.io/containerized-data-importer/pkg/common"
 )
 
 var _ = Describe("Streaming Data Conversion", func() {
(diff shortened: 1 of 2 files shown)


test/functional/importer/datastream_test.go
```diff
@@ -19,111 +18,134 @@ import (
...
+			}
+
+			err = os.Mkdir(tempTestDir, 0666)
+			if err != nil {
+				Fail(fmt.Sprintf("Could not create tmp file: %v\n ", err))
+			}
+			By("Mkdir " + tempTestDir)
+		})
+
+		//AfterEach(func() {
```
copejon: commented out so i can poke into the .tmp-* dirs

copejon:
note to set: must cp source file to tmp dir, then gzip/xz.  cannot `gzip -c src > tgt` since there's not tty present.

The tgt path is corrupted by the url package when `file://myPath/to/the/target` is processed.  it treats the path[0] element as the host, cropping it from the path e.g.
```
host: "myPath"
path: "/to/the/target"
```

copejon:
This refactor will be merged into #246 as they changes here are dependent on that PR and will break master is merged alone.

PR #262, opened 2018-07-16
Fix local transport path to file

copejon:
The file path for local transport is divided between the url.Host and url.Path fields, resulting in a truncated path to a file.

e.g. `file:///root/go/src/github.com/` becomes
```
Host: /root/
Path: go/src/github.com
```

This PR fixes the bug to return a full path to the file.


copejon:
nvm, not a bug

PR #276, opened 2018-07-26
Release announcement generator

copejon:
Mimics kubevirt's realease-announcement generator script and make target
Refactors makefile to remove ugly duplication

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,6 @@
 #See the License for the specific language governing permissions and
 #limitations under the License.
 
-DOCKER=1
-
 .PHONY: build build-controller build-importer \
 		docker docker-controller docker-cloner docker-importer \
 		test test-functional test-unit \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -12,32 +12,31 @@
...
-
-all: docker
+		goveralls \
+		release-description
 
-clean:
+DOCKER=1
 ifeq (${DOCKER}, 1)
-	./hack/build/in-docker "./hack/build/build-go.sh clean; rm -rf bin/* _out/* manifests/generated/* .coverprofile"
+DO=./hack/build/in-docker
```
jeffvance: Does make allow for indentation here (and below after else)?

Makefile
```diff
@@ -12,32 +12,31 @@
...
 else
-	./hack/build/build-go.sh clean; rm -rf bin/* _out/* manifests/generated/* .coverprofile
+DO=eval
```
jeffvance: do you **need** `eval` for this case? IOW, would DO="#" work?

copejon: I tried that initially, but without `eval` it's equivalent to `$ "./do/my/script.sh"` which isn't legal.

doc/releases.md
```diff
@@ -14,30 +14,33 @@ CDI adheres to the [semantic version definitions](https://semver.org/) format of
 
 ### Releasing a New Version
 
- The version number is tracked in several files in CDI as well as through a git tag.  To reduce the chance of human error, a help script is used to change the version in all known locations.
+Release branches are used to isolate a stable version of CDI.  Git tags are used within these release branches to track incrementation of Minor and Patch versions.  When a Major incrementation occurs, a new stable branch should be created corresponding to the release.
```
jeffvance: is "incrementation" a word? :)

Makefile
copejon: `DO` must be declared as a make variable, so it must be left justified.  Indenting it would cause to to be declared as a shell var in a sub shell I think.  This means it wouldn't be accessible to the recipes since they are all executed within their own sub shells as well.

doc/releases.md
```diff
@@ -14,30 +14,33 @@ CDI adheres to the [semantic version definitions](https://semver.org/) format of
...
 
-     DO NOT EDIT ANY VERSION STRINGS IN CDI!!
+- Release branches should adhere to the `release-v#.#.#` pattern.
 
-1. Set the new release version
+- Tags should adhere to the `v#.#.#(-alpha.#)*` pattern.
```
jeffvance: why parenthesis here? If you mean optional then I think square brackets are more typical, eg:
`v#.#.#[-alpha.#*`
Also, confused by the `*`. Is that a regex to mean _repeated 0 or more times_?

doc/releases.md
copejon: Wellllll .... 185K hits on google, none of which is the oxford dictionary :P

hack/build/release-description.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
+
```
jeffvance: no `set -exu -o pipefail` ?

hack/build/release-description.sh
```diff
@@ -0,0 +1,89 @@
...
+log() { echo "$@" >&2; }
+title() { underline "=" "$@"; }
+section() { underline "-" "$@"; }
+
+#
+# All sorts of content
+#
+release_notes() {
+    log "Fetching release notes"
+    cat manual-release-notes || echo "FIXME manual notes needed"
```
jeffvance: CWD is always the right dir here? No repo-root stuff needed?

copejon: the script doesn't access other files, so it doesn't need to know cwd

hack/build/release-description.sh
copejon: added

doc/releases.md
```diff
@@ -14,41 +19,46 @@ CDI adheres to the [semantic version definitions](https://semver.org/) format of
...
 ### Releasing a New Version
 
- The version number is tracked in several files in CDI as well as through a git tag.  To reduce the chance of human error, a help script is used to change the version in all known locations.
+Release branches are used to isolate a stable version of CDI.  Git tags are used within these release branches to track incrementing of Minor and Patch versions.  When a Major version is incremented, a new stable branch should be created corresponding to the release.
 
-     DO NOT EDIT ANY VERSION STRINGS IN CDI!!
+- Release branches should adhere to the `release-v#.#.#` pattern.
 
-1. Set the new release version
+- Tags should adhere to the `v#.#.#(-alpha.#)` pattern.
```
jeffvance: Better without the `*` but the parenthesis still confuse me. What do they mean in this context?

copejon: As in `The dog(s) is (are) playing in the yard.`

Issue #279, opened 2018-07-29
Remove cloning label from target pods to fix pod affinity


zvikorn:
Having the cloning label in the target pod, make the pod affinity fails.
The target pod looks for a pod with a specific label (specified in the pod affinity) that matches the source pod label.
In my case the target pod included this label as well, so we can see that the target pod found matching pod, but it is the WRONG pod. It's itself!!
The target was running without finding the source pod first.
If we remove this label from the target pod, it will find the source pod and then will be scheduled on the same node.
If it does not find the source pod (because the scheduler tried to schedule it before the source pod), it will be in 'Pending' state until the source pod is scheduled, and then will be running on the same node.

zvikorn:
Fixed. PR - https://github.com/kubevirt/containerized-data-importer/pull/280

jeffvance:
fixed #280

PR #280, opened 2018-07-29
Remove cloning label from target pods to fix pod affinity. 

zvikorn:
Issue:
https://github.com/kubevirt/containerized-data-importer/issues/279
If we remove the cloning label from the target pod then the target pod will be scheduled on the same node as the source pod.

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -446,8 +446,7 @@ func MakeCloneTargetPodSpec(image, verbose, pullPolicy string, pvc *v1.Persisten
 				AnnCloningCreatedBy: "yes",
 			},
 			Labels: map[string]string{
-				CDI_LABEL_KEY:     CDI_LABEL_VALUE,                               //filtered by the podInformer
-				CLONING_LABEL_KEY: CLONING_LABEL_VALUE + "-" + generatedLabelStr, //used by PodAffinity
+				CDI_LABEL_KEY:     CDI_LABEL_VALUE,    //filtered by the podInformer
 			},
 		},
 		Spec: v1.PodSpec{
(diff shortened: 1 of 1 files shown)


pkg/controller/util.go
```diff
@@ -491,6 +490,10 @@ func MakeCloneTargetPodSpec(image, verbose, pullPolicy string, pvc *v1.Persisten
 					Name:            CLONER_TARGET_PODNAME,
 					Image:           image,
 					ImagePullPolicy: v1.PullPolicy(pullPolicy),
+					SecurityContext: &v1.SecurityContext{
+						Privileged: &[]bool{true}[0],
```
jeffvance: Why does the target pod need to run privileged? Is there a more granular capability this pod can be granted?

BTW, good job deducing the problem!

zvikorn: Thanks!
It runs privileged as the source pod does. It was added to the source when I add permission issue cloning on minishift. I used to get this error:
"error processing pvc "target-ns/target-pvc": source pod API create errored: pods "source-pod-" is forbidden: unable to validate against any security context constraint: [spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 0: must be in the ranges: [1000080000, 1000089999]]"

aglitke:
Please add the description of the problem (from the issue) to the commit message.  That way we can see the explanation along with the code right in the git history.

zvikorn:
You right, I'm sorry.

pkg/controller/util.go
jeffvance: Ack. However, running a container privileged is the "sledge hammer" fix. If possible it is preferable to add just the CAPS needed by the container to read from the socket. Maybe that does require a fully privileged container but I'm asking here in case not.

zvikorn: @aglitke ?

aglitke: Actually, I am not sure how this is related to removing a label on the pod?

zvikorn:
It’s not. It’s something I forgot to do before. I can remove it for this PR and submit a different PR for this.

pkg/controller/util.go
zvikorn: It’s not. It’s something I forgot to do before. I can remove it for this PR and submit a different PR for this.

PR #282, opened 2018-07-30
Add kubevirt cluster environment with sync

awels:
- Added ability to call make cluster-up to start a kubevirt cluster.
- Added ability to call make cluster-down to stop a kubevirt cluster.
- Added ability to call make cluster-sync to build CDI and sync it with
  the running cluster. This allows for testing during development.
- Added cli.sh and kubectl.sh just like kubevirt, in the CDI project to
  manipulate a running cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@
 
 .PHONY: build build-controller build-importer \
 		docker docker-controller docker-cloner docker-importer \
+		cluster-sync cluster-sync-controller cluster-sync-cloner cluster-sync-importer \
 		test test-functional test-unit \
 		publish \
 		vet \
@@ -89,3 +90,19 @@ goveralls:
 
 release-description:
 	./hack/build/release-description.sh ${RELREF} ${PREREF}
+
+cluster-up:
+	./cluster/up.sh
+
+cluster-down:
+	./cluster/down.sh
+
+cluster-sync: build ${WHAT}
+	./cluster/sync.sh ${WHAT}
+
+cluster-sync-controller: WHAT = cmd/cdi-controller
+cluster-sync-controller: cluster-sync
+cluster-sync-importer: WHAT = cmd/cdi-importer
+cluster-sync-importer: cluster-sync
+cluster-sync-cloner: WHAT = cmd/cdi-cloner
+cluster-sync-cloner: cluster-sync
--- a/cluster/sync.sh
+++ b/cluster/sync.sh
@@ -1,0 +1,30 @@
+#!/bin/bash -e
+
+cdi=$1
+cdi="${cdi##*/}"
+
+echo $cdi
+
+source ./hack/build/config.sh
+source ./cluster/gocli.sh
+
+registry_port=$($gocli ports registry | tr -d '\r')
+registry=localhost:$registry_port
+
+DOCKER_REPO=${registry} make docker push
+DOCKER_REPO="registry:5000" make manifests
+
+# Make sure that all nodes use the newest images
+container=""
+container_alias=""
+images="${@:-${DOCKER_IMAGES}}"
+for arg in $images; do
+    name=$(basename $arg)
+    container="${container} registry:5000/${name}:latest"
+done
+for i in $(seq 1 ${KUBEVIRT_NUM_NODES}); do
+    echo "node$(printf "%02d" ${i})" "echo \"${container}\" | xargs \-\-max-args=1 sudo docker pull"
+    ./cluster/cli.sh ssh "node$(printf "%02d" ${i})" "echo \"${container}\" | xargs \-\-max-args=1 sudo docker pull"
+done
+
+./cluster/kubectl.sh apply -f ./manifests/generated/cdi-controller.yaml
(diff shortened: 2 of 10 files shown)


jeffvance:
@awels are cli.sh and kubectl.sh copies of the same in kubevirt? If yes what is the plan to keep them in-sync with kubevirt?

awels:
No, they are simplified versions of the ones in kubevirt as they don't need to support all the intricate interactions the kubevirt one does.

cluster/up.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
+
+source ./cluster/gocli.sh
+
+$gocli run --random-ports --nodes 2 --background kubevirtci/k8s-1.10.3
```
awels: Need to just adjust this to take different providers and node counts based on the same KUBEVIRT_PROVIDER and KUBEVIRT_NODES variables, its hard coded here.

jeffvance:
needs rebase

cluster/sync.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo $cdi
+
+source ./hack/build/config.sh
+source ./cluster/gocli.sh
+
+registry_port=$($gocli ports registry | tr -d '\r')
+registry=localhost:$registry_port
+
+DOCKER_REPO=${registry} make build
```
copejon: this should be `make docker` to generate fresh images

awels: done

cluster/up.sh
```diff
@@ -0,0 +1,57 @@
...
+source ./cluster/gocli.sh
+
+num_nodes=${KUBEVIRT_NUM_NODES:-1}
+re='^-?[0-9]+$'
+if ! [[ $num_nodes =~ $re ]] || [[ $num_nodes -lt 1 ]] ; then
+    num_nodes=1
+fi
+
+image="k8s-1.10.4"
+case "${KUBEVIRT_PROVIDER}" in
```
copejon: It looks like `KUBEVIRT_PROVIDER` is meant to be set at the cli - if so, then it would be nice to simplify the cases to `kubernetes` and `openshift` and set `k8s-1.10.4` or `os-3.10.0` accordingly.

e.g.
```
...
 "kubernetes")
    image = "${KUBE_IMAGE}"
...
```

cluster/up.sh
```diff
@@ -0,0 +1,57 @@
...
+"k8s-1.10.4")
+    image="k8s-1.10.4"
+    ;;
+"os-3.10.0")
+    image="os-3.10.0"
```
copejon: I would prefer to store `k8s-1.10-4` and `os-3.10.0` in config.sh and then source in the script.  The intent of that file is to give devs an at-a-glance idea of the build / test / deploy configuration, so making the kubevirtci images visible there would be useful.

copejon:
Can you also add the new make targets and variables to [hack/README.md](https://github.com/kubevirt/containerized-data-importer/blob/master/hack/README.md#make-targets)?

cluster/sync.sh
```diff
@@ -0,0 +1,31 @@
...
+echo $cdi
+
+source ./hack/build/config.sh
+source ./cluster/gocli.sh
+
+registry_port=$($gocli ports registry | tr -d '\r')
+registry=localhost:$registry_port
+
+DOCKER_REPO=${registry} make docker
+DOCKER_REPO=${registry} make push
```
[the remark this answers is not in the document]
awels: No particular reason that I am aware of, so I can combine these no problem.

cluster/sync.sh
```diff
@@ -0,0 +1,31 @@
...
+DOCKER_REPO="registry:5000" make manifests
```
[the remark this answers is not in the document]
awels: Yes, we are generating the manifests here, and we need to point to the 'internal' registry, which is exposed by a dnsmasq as 'registry:5000' inside the cluster, but exposed as localhost:$registry_port outside the cluster. This is the mechanism that is used to get the created images into the cluster. However the exposed port changes based on the run, and the gocli command looks up the port for 'localhost', but it is static inside the cluster as 'registry:5000'

cluster/up.sh
awels: Sure we can do that.

Makefile
```diff
@@ -89,3 +90,19 @@ goveralls:
...
+cluster-up:
+	./cluster/up.sh
+
+cluster-down:
+	./cluster/down.sh
+
+cluster-sync: build ${WHAT}
+	./cluster/sync.sh ${WHAT}
+
+cluster-sync-controller: WHAT = cmd/cdi-controller
```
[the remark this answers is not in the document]
awels: That is a mistake, it should be cmd/ for all of them, as you are right the sync.sh strips out the cmd/ part.

cluster/sync.sh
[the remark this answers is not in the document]
copejon: Got it, thanks for explaining :)

cluster/up.sh
copejon: Given that, I'm fine with the values as is then.

awels:
@copejon Added new targets to the README.

PR #293, opened 2018-07-31
Use the correct API group for import pod owner reference

### LINKED ISSUE #292: Wrong API Group specified for blockOwnerDeletion
When attempting to import into a PVC the following error appears in the logs...

E0731 17:53:43.133558       1 import-controller.go:206] error processing pvc "default/golden-pvc": importer pod API create errored: pods "importer-golden-pvc-t5zvn" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion apps/v1 Kind PersistentVolumeClaim: allowed by openshift authorizer, no matches for apps/, Kind=PersistentVolumeClaim

aglitke:
The api group should be v1 not appsv1.  Fixes #292.

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -2,9 +2,11 @@ package controller
 
 import (
 	"fmt"
+	"strings"
+	"time"
+
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
-	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/util.go
```diff
@@ -479,7 +479,7 @@ func MakeCloneTargetPodSpec(image, verbose, pullPolicy string, pvc *v1.Persisten
 									},
 								},
 							},
-							Namespaces: []string{podAffinityNamespace},
+							Namespaces:  []string{podAffinityNamespace},
```
jeffvance: super nit: did `gofmt` add the trailing space?

jeffvance:
lgtm

pkg/controller/util.go
aglitke: yes

mhenriks:
appsv1 is referenced in util_test.go as well so should probably be changed there as well

jeffvance:
@aglitke 
> appsv1 is referenced in util_test.go as well so should probably be changed there as well
 Do you want this to be a separate pr or added to your pr here?

PR #308, opened 2018-08-01
bridging change 303 and 291 differences

screeley44:
seems we had multiple PR's making changes to the rbac, this is to blend the differences together


jeffvance:
lgtm

PR #311, opened 2018-08-02
add cluster considerations for cloning to readme

### LINKED ISSUE #306: update readme regarding cluster requirements for k8s < 1.10 and pod scheduling
Add to README cluster requirements so that the cloner pods reside on the same node and run in the correct order.

The reason for this issue is in consideration of a user, running the cloner on a cluster not configured as:
```
PersistentLocalVolumes=true
VolumeScheduling=true
MountPropagation=true
volumeBindingMode = WaitForFirstConsumer
```
and the target and source pods end up being scheduled on different nodes. In this case the target pod will wait for 15s before giving up. Log info does help identify the problem.

Additionally, 15s seems like too short of a wait on a busy cluster, so my sense is to increase the wait time.

See issue #299 for additional details.

jeffvance:
and increase num of retries in _cloner-startup.sh_ script to account for slower nodes.
Note: all feature gates mentioned in this pr default to true in k8s 1.10 and later. I expect the addition to the readme to be short-lived but still worthwhile as long as cdi is used on k8s 1.9 and earlier clusters.
Fixes #306 
Feature gate statuses:  https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,26 @@ Make copies of the [example manifests](./manifests/example) for editing. The nec
 
 ### Cloning VM Images
 
-Cloning is achieved by creating a new PVC with the 'k8s.io/CloneRequest' annotation indicating the name of the PVC the image is copied from. Once the controller detects the PVC, it starts two pods which are responsible for the cloning of the image from one PVC to another using a unix socket that is created on the host itself. When the cloning is completed, the PVC which the image was copied to, is assigned with the 'k8s.io/CloneOf' annotation to indicate cloning completion. The copied VM image can be used by a new pod only after the cloning process is completed.
+Cloning is achieved by creating a new PVC with the 'k8s.io/CloneRequest' annotation indicating the name of the PVC the image is copied from.
+Once the controller detects the PVC, it starts two pods which are responsible for the cloning of the image from one PVC to another using a unix socket that is created on the host itself.
+When the cloning is completed, the PVC which the image was copied to, is assigned with the 'k8s.io/CloneOf' annotation to indicate cloning completion.
+The copied VM image can be used by a new pod only after the cloning process is completed.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -152,7 +152,28 @@ Make copies of the [example manifests](./manifests/example) for editing. The nec
...
+Cloning is achieved by creating a new PVC with the 'k8s.io/CloneRequest' annotation indicating the name of the PVC the image is copied from.
+Once the controller detects the PVC, it starts two pods which are responsible for the cloning of the image from one PVC to another using a unix socket that is created on the host itself.
+When the cloning is completed, the PVC which the image was copied to, is assigned with the 'k8s.io/CloneOf' annotation to indicate cloning completion.
+The copied VM image can be used by a new pod only after the cloning process is completed.
+
+Cloning requires two pods, the source pod and target pod, which must execute on the same node.
+Pod adffinity is used to enforce this requirement; however, the cluster also needs to be configured to delay volume binding until pod scheduling has completed.
+
+In Kubernetes 1.10 and older export KUBE_FEATURE_GATES before bringing up the cluster:
```
mhenriks: Not required in 1.11?

jeffvance: That's my understanding. @aglitke ?

screeley44: I thought the following feature gates are defaulted to true starting in 1.10 (promoted to beta)...so only needed in 1.9 and below...I could be mistaken?

jeffvance: Correct @screeley44 . I've updated the readme. Thanks!

jeffvance:
Note: I refactored a lot of _clone_startup.sh_ to improve readability and to fix some test cases where the max retries test was not triggered.

copejon:
LGTM

PR #312, opened 2018-08-02
Add disclaimer and a log warning about local (:file//) endpoints

### LINKED ISSUE #286: Document & log file:/// endpoints as debug use only
Use of the `file:///` endpoint location is causing confusion with users who attempt to access a file on the host filesystem.    The local file scheme is designed with debugging in mine ( e.g. when running the importer binary on the host machine outside of a container).   This needs to be documented and log output needs to be produced to better inform users.

It should be noted in the [README.md](https://github.com/kubevirt/containerized-data-importer/blob/master/README.md#start-importing-images) and log output should be written from the [`local()`](https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/importer/dataStream.go#L184) method.

The log output should be a Warning vLog level 1.

jeffvance:
Add a warning log msg when a _file://_ (local) endpoint is being imported and update the readme to indicate that local imports are not supported.
Fixes #286 

Here is a sample warning:
```
W0802 16:00:29.104350       5 dataStream.go:186] the file:/// protocol is intended *only* for debugging and testing. Files outside of the container cannot be accessed.
```


mhenriks:
LGTM

README.md
```diff
@@ -57,6 +57,8 @@ Supported file formats are:
 - A running Kubernetes cluster with roles and role bindings implementing security necesary for the CDI controller to watch PVCs and pods across all namespaces.
 - A storage class and provisioner.
 - An HTTP or S3 file server hosting VM images
+> Note: CDI is also able to import a local file endpoint but this is not supported except for testing and debugging. The `file::///path-to-host-file` scheme is used and the endpoint's host path must be mounted in the importer container. This bind-mount is not done via the CDI controller nor importer pod.
```
copejon: 1 too many `:` in `file::///...` :)

pkg/importer/dataStream.go
```diff
@@ -183,6 +183,7 @@ func (d dataStream) http() (io.ReadCloser, error) {
 
 func (d dataStream) local() (io.ReadCloser, error) {
 	fn := d.Url.Path
+	glog.Warningf("attempting to import local file %q which is not officially supported", fn)
```
copejon: Can this be a little more verbose?  Something like
> Warning - the file:/// protocol is intended for debug and testing only as it cannot access filesystems external to the container.

Or something along those lines?

README.md
copejon: It might better to be more explicit with something like
> ... file:/// endpoints can only access the container's filesystem and requires that files on the node be mounted via hostPath to the Pod ...

PR #313, opened 2018-08-02
Refactor test/framkework's FormatTestData for reusability

copejon:
Right now, the `gzip` and `xz` commands will destroy the `tinyCore.iso` during test generation.  This is a problem during noop tests that expect a `tinyCore.iso` as the commands are destructive.  Instead, copy `test/images/tinyCore.iso` to a tmp dir, gzip/xzip it there (destructively) then copy the resulting file to the test directory.  This preserves noop tests expecting the `.iso` and makes the `FormatTestData()` function more predictable and thus usable by other packages.

edit: 
Additionally, if the compression commands are non-destructive (via the `-k` flag), then gzip and xz tests will leave by duplicated tinyCore.iso files laying around without cleaning them up.


jeffvance:
@copejon my suggestion for a minor change in `copyIfNotPresent`:
```
// If no copy is performed, the original src string is returned.  <== originally false for err or if tgt exists
func copyIfNotPresent(src, tgtDir string) (string, error) {
	tgt :=  filepath.Join(tgtDir, filepath.Base(src))
	// Only copy the source image if it does not exist in the temp directory
	err := os.Stat(tgt)
        if os.IsNotExist(err) {
		if err = doCmd("cp", src, tgtDir); err == nil {
			return tgt, nil
		}
                return src, err
	}
	return src, nil
}
```

jeffvance:
thinking off the top of my head.... would something like this save us from doing up to 2 copies for a gz|xz file?
```
func gzCmd(src, tgtDir string) (string, error) {
        tgt :=  filepath.Join(tgtDir, filepath.Base(src)) + image.ExtGz
	if err := doCmd("gzip", "-c", src, ">", tgt); err != nil {
		return "", err
	}
	return tgt, nil
```

test/framework/fileConversion.go
```diff
@@ -49,29 +49,27 @@ func tarCmd(src, tgtDir string) (string, error) {
 }
 
 func gzCmd(src, tgtDir string) (string, error) {
```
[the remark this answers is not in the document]
copejon: There's a gzip/xz flag to keep the file, however that doesn't help us when only running a gzip test since it will leave a 2nd .iso laying around.

Without the flag though, a .iso that another test might be expecting will be destroyed.

copejon:
I tried that initially, but the command is exec'd with no tty, so streaming stdout causes an error.

pkg/importer/dataStream.go
```diff
@@ -344,7 +344,7 @@ func (d dataStream) qcow2NopReader(h *image.Header) (io.Reader, int64, error) {
 //  order to get its original size. For now 0 is returned.
 //TODO: support gz size.
 func (d dataStream) xzReader() (io.Reader, int64, error) {
-	xz, err := xz.NewReader(d.topReader(), 0) //note: default dict size may be too small
+	xz, err := xz.NewReader(d.topReader()) //note: default dict size may be too small
```
jeffvance: if a dictionary is not applicable to new xz pkg then pls remove this comment.

test/framework/fileConversion.go
```diff
@@ -37,44 +40,86 @@ func FormatTestData(srcFile, tgtDir string, targetFormats ...string) (string, er
...
+func toTar(src, tgtDir string) (string, error) {
+	tgtFile, tgtPath, err := createTargetFile(src, tgtDir, image.ExtTar)
+	defer tgtFile.Close()
 
-	if err := doCmdAndVerifyFile(tgt, "tar", args...); err != nil {
-		return "", err
+	w := tar.NewWriter(tgtFile)
+	defer w.Close()
+
+	srcFile, err := os.Open(src)
```
jeffvance: something to think about (maybe) is if you want to wrap a buffer reader around this for (maybe) better performance? Depends on what buffer is done with a file reader.

test/framework/fileConversion.go
```diff
@@ -37,44 +40,86 @@ func FormatTestData(srcFile, tgtDir string, targetFormats ...string) (string, er
...
+	if err != nil {
+		return "", errors.Wrap(err, "Error stating file")
+	}
+
+	hdr, err := tar.FileInfoHeader(srcFileInfo, "")
+	if err != nil {
+		return "", errors.Wrap(err, "Error generating tar file header")
+	}
+
+	err = w.WriteHeader(hdr)
```
jeffvance: In other examples I've seen of writing a tar file there's a loop for each file being archived.  Maybe add a comment stating assumption of single file (though I guess that's pretty obvious now that I think about it)...

test/framework/fileConversion.go
```diff
@@ -37,44 +40,86 @@ func FormatTestData(srcFile, tgtDir string, targetFormats ...string) (string, er
...
+	if err != nil {
+		return "", errors.Wrap(err, "Error writing header")
+	}
+	defer w.Close()
```
jeffvance: why a 2nd defer of `w,Close()` here?

test/framework/fileConversion.go
```diff
@@ -37,44 +40,86 @@ func FormatTestData(srcFile, tgtDir string, targetFormats ...string) (string, er
...
+	_, err = io.Copy(w, srcFile)
+	if err != nil {
+		return "", errors.Wrap(err, "Error writing to file")
+	}
+	return tgtPath, nil
 }
 
-func gzCmd(src, tgtDir string) (string, error) {
-	src, err := copyIfNotPresent(src, tgtDir)
+func toGz(src, tgtDir string) (string, error) {
```
jeffvance: do you see a way to leverage the similar code between toTar, toGz and toXz?

test/framework/fileConversion.go
```diff
@@ -37,44 +40,86 @@ func FormatTestData(srcFile, tgtDir string, targetFormats ...string) (string, er
...
+
+	_, err = io.Copy(w, srcFile)
+	if err != nil {
+		return "", errors.Wrap(err, "Error writing to file")
+	}
+	return tgtPath, nil
 }
 
-func qcow2Cmd(srcfile, tgtDir string) (string, error) {
+func toQcow2(srcfile, tgtDir string) (string, error) {
```
jeffvance: `Replace` below doesn't work if _.iso_ is part of base name (which is very unlikely but could occur).

test/framework/fileConversion.go
```diff
@@ -85,7 +130,7 @@ func qcow2Cmd(srcfile, tgtDir string) (string, error) {
 	return tgt, nil
 }
 
-func noopCmd(src, tgtDir string) (string, error) {
+func toNoop(src, tgtDir string) (string, error) {
```
jeffvance: what do you think about this entire func begin:
```
func toNoop(src, tgtDir string) (string, error) {
     return copyIfNotPresent(src, tgtDir)
}
```

test/framework/fileConversion.go
```diff
@@ -112,17 +157,29 @@ func doCmd(osCmd string, osArgs ...string) error {
...
-	if _, err := os.Stat(filepath.Join(tgtDir, base)); err != nil {
-		if err := doCmd("cp", "-f", src, tgtDir); err != nil {
-			return "", err
-		}
-		src = filepath.Join(tgtDir, base)
+	ret := filepath.Join(tgtDir, filepath.Base(src))
+	_, err := os.Stat(ret)
+	if err != nil && !os.IsNotExist(err) {
+		return "", errors.Wrap(err, "Unexpected error stating file")
+	}
```
jeffvance: at this point err == nil or == IsNotExist. If err == nil we don't want to do the cp.

test/framework/fileConversion.go
copejon: Yeah, the package is designed to handle single files only.  I can add a package level comment to describe the overall design / behavior though.

test/framework/fileConversion.go
copejon: oops! Fixed

test/framework/fileConversion.go
copejon: Will look to see what i can generalize

test/framework/fileConversion.go
jeffvance: The import copy is not wrapped with a buffer reader, though I thought about it. For now this is fine as is.

test/framework/fileConversion.go
copejon: Yup! Thanks!

test/framework/fileConversion.go
```diff
@@ -120,7 +122,9 @@ func toXz(src, tgtDir string) (string, error) {
 }
 
 func toQcow2(srcfile, tgtDir string) (string, error) {
-	tgt := strings.Replace(filepath.Base(srcfile), ".iso", image.ExtQcow2, 1)
+	base := filepath.Base(srcfile)
+	i := strings.LastIndex(base, ".iso")
+	tgt := strings.Replace(base[i:], ".iso", image.ExtQcow2, 1)
```
jeffvance: +1

test/framework/fileConversion.go
```diff
@@ -166,8 +166,10 @@ func copyIfNotPresent(src, tgtDir string) (string, error) {
 	if err != nil && !os.IsNotExist(err) {
 		return "", errors.Wrap(err, "Unexpected error stating file")
 	}
-	if err = doCmd("cp", src, tgtDir); err != nil {
-		return "", err
+	if os.IsNotExist(err) {
+		if err = doCmd("cp", src, ret); err != nil {
```
jeffvance: nit: could be combined with `&&` but no big deal...

test/framework/fileConversion.go
```diff
@@ -22,8 +22,11 @@ var formatTable = map[string]func(string, string) (string, error){
 	"":             toNoop,
 }
 
-// create file based on targetFormat extensions and return created file's name.
-// note: intermediate files are removed.
+// FormatTestData accepts the path of a single file (srcFile) and attempts to generate an output
```
jeffvance: +1

test/framework/fileConversion.go
jeffvance: Other than the code not quite matching the comments, this is just personal preference. I'd do:
```
// If no copy is performed, the original src string is returned.   // <==== true?
func copyIfNotPresent(src, tgtDir string) (string, error) {
	ret := filepath.Join(tgtDir, filepath.Base(src))
	_, err := os.Stat(ret)
	if err == nil {
                return src, nil
        }
	if !os.IsNotExist(err) {
                return src, errors.Wrap(err, "Unexpected error stating file")
        }
        if err = doCmd("cp", src, ret); err != nil {
		return src, err
	}
	return ret, nil
}
```

test/framework/fileConversion.go
copejon: There's not a real clean way to generalize the functions without hurting readability.  The pattern that each is written in is identical by design, but the calls they make are specialized and dissimilar in their returns.

test/framework/fileConversion.go
```diff
@@ -112,17 +155,31 @@ func doCmd(osCmd string, osArgs ...string) error {
...
-// copyIfNotPresent checks for the src file in the tgtDir.  If it is not there, it attempts to copy if from src to tgtdir.
+// copyIfNotPresent checks for the src file in the tgtDir.  If it is not there, it attempts to copy it from src to tgtdir.
 // If a copy is performed, the path to the copy is returned.
 // If no copy is performed, the original src string is returned.
 func copyIfNotPresent(src, tgtDir string) (string, error) {
-	base := filepath.Base(src)
-	// Only copy the source image if it does not exist in the temp directory
-	if _, err := os.Stat(filepath.Join(tgtDir, base)); err != nil {
-		if err := doCmd("cp", "-f", src, tgtDir); err != nil {
+	ret := filepath.Join(tgtDir, filepath.Base(src))
```
jeffvance: This func does not exactly implement it's comment (last comment). Me, personally, I'd refactor as:
```
// If no copy is performed, the original src string is returned.  <========= true??
func copyIfNotPresent(src, tgtDir string) (string, error) {
	ret := filepath.Join(tgtDir, filepath.Base(src))
	err := os.Stat(ret)
	if  err == nil { // file exists, don't copy
                return src, nil  // do we return src per comment or should ret be returned??
        }
	if !os.IsNotExist(err) {
		return src, errors.Wrap(err, "Unexpected error stating file")
	}
	if err = doCmd("cp", src, ret); err != nil {
			return src, err
	}
	return ret, nil
}
```
but it's really personal preference...

copejon: If the source and the target are the same, then ret == src, no copy is done, and the original path is returned.

jeffvance: true, but one of the err return cases explicitly returns `"", err` which contradicts the last comment. I guess that's my main point....

jeffvance:
lgtm

PR #317, opened 2018-08-06
Streaming qcow2->raw conversion and resource limits on qemu-img subprocess

### LINKED ISSUE #254: enhance importer to use qemu-img pseudo streaming to save /tmp step
For qcow2 files, rather than copying to _/tmp/rand-name_ and then running `qcow2-img` to convert it to raw, we can use `qcow2-img` to copy the file and convert it, saving the intermediate step and saving disk/cpu on the host.  Of course _.tar.gz_, _.xz_ conversion needs to still be done first so this may only help for the simple case of importing _.qcow2_ --> raw.

Some timings from a recent email:
```
$ time wget http://local.server/Fedora-AtomicHost-28-20180625.1.x86_64.qcow2 
...
Length: 638043136 (608M) [application/octet-stream]Saving to: ‘Fedora-AtomicHost-28-20180625.1.x86_64.qcow2’

Fedora-AtomicHost-28-20180625.1.x86_64.qcow2    100%[=====================================================================================================>] 608.49M   107MB/s    in 5.6s    

2018-07-09 21:38:39 (108 MB/s) - ‘Fedora-AtomicHost-28-20180625.1.x86_64.qcow2’ saved [638043136/638043136]


real	0m5.941s
user	0m0.183s
sys	0m1.185s

$ time qemu-img convert -p -f qcow2 -O raw http://local.server/Fedora-AtomicHost-28-20180625.1.x86_64.qcow2 converted.raw
    (100.00/100%)

real	0m14.217s
user	0m5.235s
sys	0m2.343s

$ time qemu-img convert -p -f qcow2 -O raw Fedora-AtomicHost-28-20180625.1.x86_64.qcow2 converted.raw
    (100.00/100%)

real	0m11.909s
user	0m4.728s
sys	0m1.595s

So converting on the fly is even little faster then downloading to temporary file and converting.
```

mhenriks:
Two goals here:

Support streaming qcow2 to raw conversion.  Don't want to have to create temporary file.  Currently only for http or https source images.  But maybe we want to avoid the extra copy for non gzipped/tarred as well?

Put resource limits on qemu-img.  Basically, it is possible to craft a qcow2 file in such a way that it'll DOS your system.  Initial checkin uses prlimit() to set the same cpu/memory limits used in Nova.  Subsequent update will include call to `qemu-img info` to get details on source image to detect bad images before attempting conversion.
See this thread for more details about streaming conversion and DOS prevention:  https://groups.google.com/forum/#!msg/kubevirt-dev/4h7R8fPq0eU/fyL2IaiSBAAJ

- [x] streaming qcow2 to raw conversion
- [x] cpu/memory limits for qemu-img
- [x] pre-validate qcow2 with `qemu-img info`
- [x] tests

Fixes #254


jeffvance:
should we also use fedors28 in _hack/build/docker/builder/Dockerfile_?

pkg/image/qemu.go
```diff
@@ -1,16 +1,80 @@
...
+	"github.com/golang/glog"
+)
+
+const (
+	networkTimeoutSecs = 3600    //max is 10000
+	maxMemory          = 1 << 30 //value from OpenStack Nova
+	maxCPUSecs         = 30      //value from OpenStack Nova
 )
 
+// ConvertQcow2ToRaw converts a local qcou2 image to raw format
```
jeffvance: nit: typo

pkg/image/qemu.go
```diff
@@ -1,16 +1,80 @@
...
-	cmd := exec.Command("qemu-img", "convert", "-f", "qcow2", "-O", "raw", src, dest)
-	err := cmd.Run()
+	err := execWithLimits("qemu-img", "convert", "-f", "qcow2", "-O", "raw", src, dest)
+	if err != nil {
+		return errors.Wrap(err, "could not convert local qcow2 image to raw")
+	}
+	return nil
+}
+
+// ConvertQcow2ToRawStream converts an http accessible qcow2 image to raf format without locally caching the qcow2 image
```
jeffvance: nit: typo

pkg/importer/dataStream.go
```diff
@@ -467,8 +483,18 @@ func closeReaders(readers []Reader) (rtnerr error) {
 	return rtnerr
 }
 
+func (d dataStream) isHTTPQcow2() bool {
+	// assuming len(d.Readers) == 3 is [http, mr, mr]
+	// should prob get rid of the redundant MultiReader
```
jeffvance: agree. Where is the extra mr coming from?

pkg/importer/dataStream.go
```diff
@@ -467,8 +483,18 @@ func closeReaders(readers []Reader) (rtnerr error) {
...
+	return (d.Url.Scheme == "http" || d.Url.Scheme == "https") && d.qemu && len(d.Readers) == 3
```
jeffvance: `== 3`? vs. `>=3` (or 2 if mr fix). It looks like we only invoke streaming if the endpoint has not been compressed and/or archived, correct? I thought that http(s) supported some form of inline decompression and, if true, is this something we can use?

pkg/importer/dataStream.go
```diff
@@ -467,8 +483,18 @@ func closeReaders(readers []Reader) (rtnerr error) {
...
+}
+
 // Copy endpoint to dest based on passed-in reader.
 func (d dataStream) copy(dest string) error {
+	if d.isHTTPQcow2() {
+		glog.V(Vuser).Infoln("Doing streaming qcow2 to raw conversion")
```
jeffvance: perhaps logging level here should be Vadmin? Just thinking that the end user does really care if we use qemu streaming, but the admin might due to SLA or perf guarantees.

pkg/importer/dataStream.go
mhenriks: the loop in constructReaders() always adds a MultiReader to the stack, when matchHeader() returns a match.  This is a problem in the qcow2 case because unlike the other possible matches, no qcow2 reader is added to the stack.

I think there is a pretty straightforward fix for this.

pkg/importer/dataStream.go
mhenriks: Yeah, qemu-img can't work directly on compressed/tarred files.

I'm not aware of any http features that can help us here.    An http client can request that the server compress data for transmission via the `Accept-Encoding` header.  But I'm not aware of any protocol support for decompressing a request for a gzipped file.

mhenriks: Also, `qemu-img convert` doesn't support reading from stdin so we can't pipe data into it from a qzip reader.  If the file is compressed/tarred, I think we have to copy it locally,

pkg/importer/dataStream.go
mhenriks: Makes sense

pkg/importer/dataStream.go
jeffvance: Ok, my bad and thanks for finding this. If you want to fix this in a separate pr let me know asap, else I'll fix it. Thanks.

mhenriks: @jeffvance I spent a bit of time looking into this earlier.  Turns out a fix is not as simple as I thought.  I think a separate PR would be appropriate.

pkg/image/qemu.go
```diff
@@ -33,48 +33,95 @@ const (
...
+	_, err = execWithLimits("qemu-img", "convert", "-f", "qcow2", "-O", "raw", src, dest)
 	if err != nil {
 		return errors.Wrap(err, "could not convert local qcow2 image to raw")
 	}
+
 	return nil
 }
 
-// ConvertQcow2ToRawStream converts an http accessible qcow2 image to raf format without locally caching the qcow2 image
+// ConvertQcow2ToRawStream converts an http accessible qcow2 image to raw format without locally caching the qcow2 image
```
jeffvance: what do you think of having a single `ConvertQcow2ToRaw` func with an "optional" json arg?

pkg/image/qemu.go
```diff
@@ -33,48 +33,95 @@ const (
...
+	return nil
+}
+
+func validate(image, format string) error {
+	type imageInfo struct {
+		Format      string `json:"format"`
+		BackingFile string `json:"backing-filename"`
+	}
+
+	output, err := execWithLimits("qemu-img", "info", "--output=json", image)
```
jeffvance: I do not know how `qemu-img` does reads on _qcow2_ files. But, does this func cause the entire (or a lot of?) the _qcow2_ file to be read?

pkg/image/qemu.go
```diff
@@ -33,48 +33,95 @@ const (
...
 	if err != nil {
 		return errors.Wrap(err, "could not stream/convert qcow2 image to raw")
 	}
+
+	return nil
+}
+
+func validate(image, format string) error {
+	type imageInfo struct {
+		Format      string `json:"format"`
```
jeffvance: do these fields need to be exported? Maybe for testing purposes?

pkg/image/qemu.go
```diff
@@ -33,48 +33,95 @@ const (
...
+
+	var info imageInfo
+	err = json.Unmarshal(output, &info)
+	if err != nil {
+		glog.Errorf("Invalid JSON:\n%s\n", string(output))
+		return errors.Wrapf(err, "Invalid json for image %s", image)
+	}
+
+	if info.Format != format {
+		return errors.Wrapf(err, "Invalid format %s for image %s", info.Format, image)
```
jeffvance: but `err` is nil here so no need to wrap.

pkg/image/qemu.go
```diff
@@ -33,48 +33,95 @@ const (
...
+	}
+
+	if len(info.BackingFile) > 0 {
+		return errors.Wrapf(err, "Image %s is invalid because it has backing file %s", image, info.BackingFile)
```
jeffvance: ditto re nil err

pkg/image/qemu.go
mhenriks: Pretty sure it doesn't read the whole file.  Would be surprised if it reads more than the header.

pkg/image/qemu.go
mhenriks: Get this warning if I make private:  `struct field format has json tag but is not exported` so will keep exported.

jeffvance: hmmm, thanks for checking.

pkg/image/prlimit_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+const killedByTestError = "Had to kill process"
+
+func TestLimits(t *testing.T) {
```
mhenriks: Wondering if these tests are really necessary and if so should they be functional tests?

pkg/image/qemu.go
```diff
@@ -1,16 +1,127 @@
...
+	_, err = execWithLimits("qemu-img", "convert", "-f", "qcow2", "-O", "raw", src, dest)
+	if err != nil {
+		return errors.Wrap(err, "could not convert local qcow2 image to raw")
+	}
+
 	return nil
 }
+
+// ConvertQcow2ToRawStream converts an http accessible qcow2 image to raw format without locally caching the qcow2 image
+func ConvertQcow2ToRawStream(url *url.URL, dest string) error {
```
mhenriks: Should this be a functional test?

mhenriks:
ci test please

pkg/image/qemu.go
aglitke: I'd say so.  Will depend on @copejon 's server container in the environment.

pkg/image/prlimit_test.go
aglitke: I am not sure we've ever found a real image in the wild that can cause problems.  Maybe we can run using a fake qemu binary that spinloops or something?

pkg/image/qemu.go
aglitke: right, it's just reading the header.

test/scripts/memory-consumer.sh
```diff
@@ -0,0 +1,23 @@
...
+#distributed under the License is distributed on an "AS IS" BASIS,
+#WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#See the License for the specific language governing permissions and
+#limitations under the License.
+
+set -eo pipefail
+
+arr=()
+for ((i=1; i<=$1; i++)); do
+    random="$(head -c 4096 /dev/urandom | base64)"
```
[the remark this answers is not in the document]
mhenriks: bash didn't like certain characters (definitely nulls and maybe other stuff) so just base64 encoded everything.

pkg/image/qemu.go
mhenriks: @aglitke this test actually starts an embedded http server that serves up image files.  but yeah may be redundant in functional test environment

.gitignore
```diff
@@ -7,3 +7,5 @@ release-announcement
 .project
 .vscode
 cluster/.*
+vendor/**
```
copejon: A few other PRs have added this line and it's caused dep issues for devs.  Is there a reason for excluding new vendored dependencies?

mhenriks: Good catch!  This was a merge/rebase issue.  I only intended to add .history.

pkg/importer/dataStream.go
```diff
@@ -109,12 +127,12 @@ func NewDataStream(endpt, accKey, secKey string) (*dataStream, error) {
 
 // Read from top-most reader. Note: ReadFull is needed since there may be intermediate,
 // smaller multi-readers in the reader stack, and we need to be able to fill buf.
-func (d dataStream) Read(buf []byte) (int, error) {
+func (d *dataStream) Read(buf []byte) (int, error) {
```
jeffvance: why pass by reference here? We are not updating the receiver? But since the byte slice has been added to the dataStream stuct it is now using more memory.

pkg/importer/dataStream.go
jeffvance: I just added pr #339. Thanks for mentioning this as it turned out to be a bug in `constructReaders` and the unit test too.

pkg/importer/dataStream.go
mhenriks: Basically, I've always followed the pattern that if any methods have pointer receivers, they all should.  From golang faq:  https://golang.org/doc/faq#methods_on_values_or_pointers

But maybe something special is going on here that I don't get.  Like I don't understand how it would be using 512 bytes extra

If you think I should change this back or any others I changed, I will.  Just let me know.

jeffvance: @mhenriks From the link above:
> First, and most important, does the method need to modify the receiver?

This is what caught my attention in the first place (plus I just fixed my bug in `matchHeader` where that method should have used a pointer to dataStream since it reads into the dataStream buf).  Continuing from the link above:
> Second is the consideration of efficiency. If the receiver is large, a big struct for instance, it will be much cheaper to use a pointer receiver.

This is why I agree that it makes sense to pass the dataStream receiver as a pointer. And, based on this reason it also makes sense the all dataStream method be by reference.

jeffvance:
why is pr_limit_test.go in_pkg/util/prlimit_? The only func in _pkg/util/_ is `RandAlphaNum` which is not related to pr-limits. Maybe it should be moved to _pkg/image_?

pkg/importer/dataStream.go
```diff
@@ -507,9 +507,7 @@ func closeReaders(readers []Reader) (rtnerr error) {
 }
 
 func (d *dataStream) isHTTPQcow2() bool {
-	// assuming len(d.Readers) == 3 is [http, mr, mr]
-	// should prob get rid of the redundant MultiReader
-	return (d.Url.Scheme == "http" || d.Url.Scheme == "https") && d.qemu && len(d.Readers) == 3
+	return (d.Url.Scheme == "http" || d.Url.Scheme == "https") && d.qemu && len(d.Readers) == 2
```
jeffvance: +1

mhenriks:
@jeffvance I moved the process limit stuff from image to `util` because it's generic and could be used for any process, not just qemu-img.  Since it could in theory be shared with other packages, `util` seemed like a better place to put it.  Maybe a `system` package would be better?

jeffvance:
Yes, maybe a new _pkg/system/_ dir, or maybe in _pkg/common/_?

pkg/image/qemu.go
```diff
@@ -1,16 +1,118 @@
...
+
+// QEMUOperations defines the interface for executing qemu subprocesses
+type QEMUOperations interface {
+	ConvertQcow2ToRaw(string, string) error
+	ConvertQcow2ToRawStream(*url.URL, string) error
+	Validate(string, string) error
+}
+
+type qemuOperations struct{}
+
```
copejon: Something to consider - interface implementation can be checked at compile time with

`var _ qemuOperations = QEMUOperations` 

to ensure that changes in the interface definition are reflected by `qemuOperations`

mhenriks:
@jeffvance sounds good.  Thanks for all the help!

pkg/image/qemu_test.go
```diff
@@ -228,9 +228,7 @@ func replaceExecFunction(replacement execFunctionType, f func()) {
 	orig := qemuExecFunction
 	if replacement != nil {
 		qemuExecFunction = replacement
+		defer func() { qemuExecFunction = orig }()
```
jeffvance: cleaner!

aglitke:
@mhenriks Can you rebase this?  Do the existing functional tests cover this feature satisfactorily?

aglitke:
@mhenriks Can we remove the WIP label?

mhenriks:
@aglitke I honestly have no idea how to remove the WIP label!  Tried to do it awhile ago.  Maybe because I still have restricted permissions to this repo?

PR #319, opened 2018-08-06
Bug Fix - Match all tar magic numbers

### LINKED ISSUE #318: Format identification only targets ustar version "00"
The byte slice used for format identification is overly specific for [tar files](https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/image/filefmt.go#L38).  As is, the slice matches `ustar.00` and not `ustar .`.  This causes the datastream to noop `ustar .` when it should unarchive.  The result is the unrecognized tarball passed through the pipeline and written to disk as `disk.img`.

copejon:
Fixes #318 

Truncate the magic number slice for tar to match all tarball identifies.

### DIFF
--- a/pkg/image/filefmt.go
+++ b/pkg/image/filefmt.go
@@ -35,7 +35,7 @@ var knownHeaders = Headers{
 	},
 	"tar": Header{
 		Format:      "tar",
-		magicNumber: []byte{0x75, 0x73, 0x74, 0x61, 0x72, 0x20},
+		magicNumber: []byte{0x75, 0x73, 0x74, 0x61, 0x72},
 		mgOffset:    0x101,
 		SizeOff:     124,
 		SizeLen:     8,


jeffvance:
lgtm

aglitke:
Any reason why we didn't match both exact possibilities rather than making the check less sensitive to variation?

jeffvance:
> Any reason why we didn't match both exact possibilities rather than making the check less sensitive to variation?

Fair question. and that was considered. The implementation of how we process format types [L21](https://github.com/copejon/containerized-data-importer/blob/master/pkg/image/filefmt.go#L21) and [L61](https://github.com/copejon/containerized-data-importer/blob/master/pkg/image/filefmt.go#L61) doesn't work for multiple versions of the same format type. We could make ustar and ustar 0.0 separate formats with the same offset but different magic numbers and it would work. If you fell strongly about it please create a new issue.

PR #330, opened 2018-08-10
[WIP] Start adding cloning tests

aglitke:
This is a WIP PR to help add some cloning tests.  I've created some
utility functions to help create a PVC and a Pod which will run a bash
script while attached to a PVC (for checking the PVC).  Right now we're
not actually testing cloning but just getting some of the machinations
nailed down.

Test is not passing yet due to bad shell when checking PVC.

Requires a cluster with a default storage class, ie.
   export KUBEVIRT_PROVIDER=os-3.10.0


tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
+package tests_test
```
jeffvance: I'd like to see 3 new dirs under _tests/_: importer/, cloner/, and dataVolume/. I think it may get messy have all of our e2e tests under a single dir.

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+const (
+	TestNS = "default"
+)
+
+var _ = Describe("Cloning", func() {
+	var (
+		err error
+	)
+
+	flag.Parse()
```
jeffvance: This relies on _utils.go_'s `init` func. Maybe this should be documented/commented somewhere? I've been thinking that it might be better to move `init` to a flags helper func and have it set a pkg global flags struct so that all tests have easier access to all flag values.

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	BeforeEach(func() {
+		tests.CreatePVC(TestNS, "src", "1Mi")
```
jeffvance: Combining this with pr #328. we should use random namespaces that are created in a `BeforeSuite` and cleaned up in an `AfterSuite` -- see related comments in that pr.

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	k8sv1 "k8s.io/api/core/v1"
+	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"kubevirt.io/containerized-data-importer/tests"
+)
+
+const (
+	TestNS = "default"
```
jeffvance: Combining this with pr #328. we should use random namespaces that are created in a `BeforeSuite` and cleaned up in an `AfterSuite` -- see related comments in that

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+		Expect(err).To(BeNil())
+	})
+
+	AfterEach(func() {
+		tests.DeletePVC(TestNS, "src")
+	})
+
+	Specify("New source PVC should be empty", func() {
+		check := "[ $(ls -1 /pvc | wc -l) == 0 ]"
+		client, err := tests.GetKubeClient()
```
jeffvance: also commented in pr #328: I'd like to see a `framework` struct that contains the standard values need by an e2e test. Eg,  ns, kube client, ocp client, cdi client, and other fields as they become obvious...

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+		tests.PanicOnError(err)
```
jeffvance: why aren't we using _Gomega_'s expected results pkg? The typical pattern is something like:
```
err := do-something()
Expect(err).ToBe(Nil), "extra informative err msg")
...
```

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+				Requests: k8sv1.ResourceList{
+					"storage": quantity,
+				},
+			},
+		},
+	}
+}
+
+func RunPodWithPVC(namespace string, podName string, pvcName string, cmd string) {
+	client, err := GetKubeClient()
```
jeffvance: almost every func has to do this. k8s e2e tests use the previously mentioned _framework_ struct to hold the client fields, and often the client is passed to the target func to avoid getting the client each call.

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+			Resources: k8sv1.ResourceRequirements{
+				Requests: k8sv1.ResourceList{
+					"storage": quantity,
+				},
+			},
+		},
+	}
+}
+
+func RunPodWithPVC(namespace string, podName string, pvcName string, cmd string) {
```
jeffvance: to me, the func name is a bit misleading. I expected the pod to be run (inject some cmd via `kubectl exec` or similar). Instead the pod is only created. Ignore if this is just the skeleton func and more will be added. <--- this for all my comments.

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+	client, err := GetKubeClient()
+	PanicOnError(err)
+	_, err = client.CoreV1().Pods(namespace).Create(newCmdPod(podName, pvcName, cmd))
+	PanicOnError(err)
```
jeffvance: what if the err is pod already exists? or, api retry?

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+}
+
+func DeletePod(namespace string, name string) {
+	client, err := GetKubeClient()
+	PanicOnError(err)
+	err = client.CoreV1().Pods(namespace).Delete(name, nil)
```
jeffvance: ditto re. err: could be pod not-exists, could be api retry

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+				},
+			},
+		},
+	}
+}
+
+func RunPodWithPVC(namespace string, podName string, pvcName string, cmd string) {
+	client, err := GetKubeClient()
+	PanicOnError(err)
+	_, err = client.CoreV1().Pods(namespace).Create(newCmdPod(podName, pvcName, cmd))
```
jeffvance: this is a clean call to`Create`. But we may want to add (later is ok) retry logic

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+
+func newCmdPod(name string, pvcName string, cmd string) *k8sv1.Pod {
+	return &k8sv1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+		Spec: k8sv1.PodSpec{
+			RestartPolicy: k8sv1.RestartPolicyNever,
+			Containers: []k8sv1.Container{
+				{
+					Name:    "runner",
+					Image:   "fedora:28",
```
jeffvance: most cdi images are f27. Cloner is f28. We should standardize.

tests/utils.go
```diff
@@ -100,3 +103,80 @@ func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, er
...
+					Command: []string{"/bin/bash", "-c", cmd},
+					VolumeMounts: []k8sv1.VolumeMount{
+						{
+							Name:      "pvc",
+							MountPath: "/pvc",
```
jeffvance: should _/pvc_ be a const somewhere? _/pvc_ is also hard-coded in L34.

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+		defer client.CoreV1().Pods(TestNS).Delete("src", nil)
+		tests.RunPodWithPVC(TestNS, "src", "src", check)
+
+		By("Checking the pod state")
+		Eventually(func() k8sv1.PodPhase {
+			pod, err := client.CoreV1().Pods(TestNS).Get("src", k8smetav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
```
jeffvance: again, there can be api failures in Getting the pod. Do we want to be resilient to this? If not e2e failures will be less meaningful due to the chance that the cluster was busy or just bad timing.

tests/clone_test.go
```diff
@@ -0,0 +1,53 @@
...
+			return pod.Status.Phase
+		}, 30, 1).Should(Equal(k8sv1.PodSucceeded))
+	})
+
+})
+
+func runWithPVC(pvcName string, cmd string) (string, int) {
+	return "", 1
```
jeffvance: skipped...

awels:
@aglitke So the test is failing because someone (and I won't name names) added the specific sha256 to which image to use, which happens to be an image that doesn't have the local provisioner enabled yet. update build/hack/config.sh and change the KUBERNETES_IMAGE to:
KUBERNETES_IMAGE="k8s-1.10.4@sha256:486064eddea289b17e150e6600fefc89dab9164d5cba07153c02888a35fed4f1"
And the provisioner will be available and now the test works.

tests/utils.go
awels: We should probably take some inspiration from https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/util.go on how to handle checking for pod states.

jeffvance:
@aglitke Adam, can you please split this pr into two: 1) _utils.go_ for test framework changes, 2) actual clone tests. We want to prioritize test framework prs over actual test prs for now. This will reduce rebasing on the actual test prs.

awels:
@jeffvance odds are fairly good this patch will get abandoned (unless @aglitke gets to it before I do) I am currently working on the Framework PR, and its coming along nicely. Also half this PR is now in #340

tests/clone_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var _ = Describe("Cloning", func() {
+	var (
+		pvc *k8sv1.PersistentVolumeClaim
+	)
+
+	flag.Parse()
+	client, err := tests.GetKubeClient()
+	tests.PanicOnError(err)
```
copejon: Given this uses the ginkgo framework, it's better to use `Expect(err).NotTo(HaveOccurred)` which  can take a 2nd string arg for a custom message.  The 2nd benefit being that it will report in a ginkgo defined format.

awels:
Once #354 is merged, this can be closed I think, everything in here will be in there already.

aglitke:
Covered by other PRs.  Closing.

Issue #331, opened 2018-08-10
should there be a default provider defined in cluster/up.sh?


jeffvance:
and if so should it be "k8s-1.10.4"?

aglitke:
Yes.  I was going to propose something like this as well.  It should probably be the latest reasonable k8s release so the above looks good to me.

nellyc:
Is there a PR for this one already?

jeffvance:
@nellyc No pr yet.

jeffvance:
@awels mentions in pr #332 that there is a provider default ( k8s-1.10.4)

PR #332, opened 2018-08-13
add e2e functional test steps to hack/README.md

### LINKED ISSUE #327: Update and add new make functest instructions to hack/README.md
after PR #314 - need to also add updates to the hack/README.md with basic description and instructions on how to run this

screeley44:
fixes #327 

Add simple instructions on how to execute the `make functest` properly

### DIFF
--- a/hack/README.md
+++ b/hack/README.md
@@ -71,6 +71,13 @@ These may be passed to a target as `$ make VARIABLE=value target`
 - `PULL_POLICY`: (default: IfNotPresent) Set global CDI pull policy
 
 #### Execute Functional Tests
+Environment Variables and Supported Values
+
+| Env Variable       | Default       | Additional Values  |
+|--------------------|---------------|--------------------|
+|KUBEVIRT_PROVIDER   | k8s-1.10.4    | os-3.10.0          |
+|NUM_NODES           | 1             | 2-5                |
+
 To Run Tests
 ```
  make cluster-up
--- a/hack/README.md
+++ b/hack/README.md
@@ -4,6 +4,7 @@
     * [Lint, Test, Build](#lint-test-build)
         * [Make Targets](#make-targets)
         * [Make Variables](#make-variables)
+        * [Execute Functional Tests](#execute-functional-tests)
         * [Submit PRs](#submit-prs)
         * [Releases](#releases)
         * [Vendoring Dependencies](#vendoring-dependencies)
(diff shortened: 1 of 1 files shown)


awels:
lgtm

hack/README.md
```diff
@@ -69,6 +70,21 @@ These may be passed to a target as `$ make VARIABLE=value target`
 - `VERBOSITY`: (default: 1) Set global log level verbosity
 - `PULL_POLICY`: (default: IfNotPresent) Set global CDI pull policy
 
+#### Execute Functional Tests
+To Run Tests
+```
```
aglitke: We need to document the KUBEVIRT_PROVIDER variable and add a link to the available options.

Maybe https://github.com/kubevirt/kubevirtci/blob/master/README.md ?

awels: Well we specifically opted to support just k8s-1.10.4 and os-3.10.0 right now, to keep the logic that determines which image to grab simple, with the provision that if k8s-1.11.x or os-3.11.x comes out that we will switch to those. So those two are the only actual options available right now.

hack/README.md
```diff
@@ -69,6 +70,21 @@ These may be passed to a target as `$ make VARIABLE=value target`
...
+ make cluster-up
```
jeffvance: My version of this command was `KUBEVIRT_PROVIDER=k8s-1.10.4 KUBEVIRT_NUM_NODES=3 make cluster-up`. Are these variables not needed? If so what's their default?

hack/README.md
```diff
@@ -69,6 +70,21 @@ These may be passed to a target as `$ make VARIABLE=value target`
...
+ make cluster-sync
```
jeffvance: I'd like to see a short description of each of these targets.

hack/README.md
awels: Technically they are not needed, as their defaults are k8s-1.10.4 and the number of nodes = 1. The environment variables allow you to override the defaults with something else like increase the number of nodes or use the open shift provider.

jeffvance: @awels I didn't realize there was a default for `KUBEVIRT_PROVIDER`. Good to know and I'll close issue #331

hack/README.md
```diff
@@ -69,6 +70,28 @@ These may be passed to a target as `$ make VARIABLE=value target`
 - `VERBOSITY`: (default: 1) Set global log level verbosity
 - `PULL_POLICY`: (default: IfNotPresent) Set global CDI pull policy
 
+#### Execute Functional Tests
+Environment Variables and Supported Values
```
jeffvance: +1 nice!

jeffvance:
lgtm

PR #334, opened 2018-08-14
Add disclaimer and a log warning about local (:file//) endpoints (#312)

copejon:
Cherry Pick of PR #312 (9f6455)

Adds documentation explaining that `file:///` endpoint  is strictly for debugging in non-containerized run times.


aglitke:
Can you reference the original PR in this PR and the original commit in this commit?  It will help us to know that this is merely a cherrypick from master.

aglitke:
lgtm

PR #337, opened 2018-08-14
Add CDI client getter

awels:
- Added code to get an instance of a CDI client, copied from PR #328

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -15,6 +15,8 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+
+	clientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 )
 
 var KubectlPath = ""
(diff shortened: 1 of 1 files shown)


tests/utils.go
```diff
@@ -76,6 +78,17 @@ func PanicOnError(err error) {
 	}
 }
 
+// Gets an instance of a kubernetes client that includes all the CDI extensions.
+func GetCDIClient() *clientset.Clientset {
+
+	cfg, err := clientcmd.BuildConfigFromFlags(master, kubeconfig)
+	PanicOnError(err)
```
jeffvance: panic vs Fail?

tests/utils.go
```diff
@@ -76,6 +78,17 @@ func PanicOnError(err error) {
...
+	cdiClient, err := clientset.NewForConfig(cfg)
+	PanicOnError(err)
```
jeffvance: ditto

jeffvance:
In func GetKubeClientFromRESTConfig:
```
 func GetKubeClientFromRESTConfig(config *rest.Config) (*kubernetes.Clientset, error) {
 ...
 	coreClient, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		return nil, err
 	}
 	return coreClient, nil
 }
```
why not just: `return  kubernetes.NewForConfig(config)`  ?

tests/utils.go
awels: So the way I read panic vs Fail, is that panic is 100% fatal, while Fail is not good, other tests might still work.

awels: Not quite sure what GetCDIclientOrDir() means, the OrDir part is confusing me. Is it a typo and should be OrDie?

jeffvance: Sorry. Fat finger on my part. Yes I meant `GetCDIclientOrDie`

jeffvance:
@awels what do you think of renaming _tests/util.go_ to _tests/framework/util.go|framework.go_?

awels:
@jeffvance That is an interesting question, and my answer will be it depends. I want to avoid trying to prematurely organize files into directories where we end up with a bunch of directories with one or two files in it. On the other hand we don't want to end up with a ton of files in the tests directory, where it is hard to find what you are looking for.

At this point we don't have many files yet, so I think I would prefer to keep everything in tests for now.

awels:
So anything stopping this one from getting merged?

jeffvance:
lgtm

PR #339, opened 2018-08-15
fix dataStream reader and test bugs

### LINKED ISSUE #324: make sure the datastream reader stack does not contain an extra multi-reader
In a recent pr it was discovered that there is an extra multi-reader in the importer's dataStream's reader slice. If this is true we should fix it.

jeffvance:
fixes #324 
One of the dataStream unit tests called `constructReaders` twice per dataStream. Instead, that test now tests that the number of readers in the _Readers_ slice matches the expected size.
Also, `constructReaders` now exits its `for` loop once a _qcow2_ file header is detected. Previously, the loop did an extra (unnecessary) iteration resulting in an additional multi-reader being appended to the _Readers_ slice.

@copejon @screeley44 @mhenriks

### DIFF
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -56,7 +56,6 @@ const (
 	RdrFile
 	RdrGz
 	RdrMulti
-	RdrQcow2
 	RdrTar
 	RdrXz
 )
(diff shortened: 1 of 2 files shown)


screeley44:
lgtm

PR #343, opened 2018-08-16
Remove duplicate default timeout

awels:
Removed default timeout declaration that was duplicated in two files.

make functest will not compile without this.


tests/pvc_utils.go
```diff
@@ -14,7 +14,6 @@ import (
 )
 
 const (
-	defaultTimeout = 30*time.Second
 )
```
copejon: Can you remove the empty `const` block as well

jeffvance:
lgtm

PR #348, opened 2018-08-16
Colocate high level unit tests with their respective packages

copejon:
This PR moves all high level unit tests and their utilities from test/functional/... into their respective packages.  

e.g. `test/functional/importer/*` -> `pkg/importer/`

Minor code changes are required to correct imports, de-dupe some commons variables, and fix the relative path to the test image.

```
test/functional/controller/controller_suite_test.go -> pkg/controller/import_controller_ginkgo_suite_test.go
test/functional/controller/controller_test.go -> pkg/controller/import_controller_ginkgo_test.go
test/functional/importer/datastream/datastream_test.go -> pkg/importer/datastream_ginkgo_test.go
test/functional/importer/datastream/datastream_suite_test.go -> pkg/importer/datastream_suite_test.go
test/functional/importer/importer_suite_test.go -> pkg/impoter/importer_suite_test.go
test/functional/importer/importer_test.go -> pkg/impoter/importer_test.go
test/framework/fileConversion.go -> tests/utils/fileConversion.go
test/images/cirros-qcow2.img -> tests/images/cirros-qcow2.img
test/images/tinyCore.iso -> tests/images/tinyCore.iso
```

Since these former functional tests are now considered unit tests, the `make test-functional` target is removed from travis ci.  The reason is that it now executes the new set of functional tests which are reliant on a running cluster.

TODO

- [x]   Adjust build flow to handle mixed (standard and ginkgo) tests colocated in the same dir


copejon:
cc @jeffvance @screeley44 @awels @davidvossel @aglitke

awels:
@copejon So I think I am missing something. I thought you were simply going to move stuff from /test into their respective directories where the unit tests should go. But I am seeing references to tests (which are our e2e tests) now, from the unit tests. I don't think we want to do that.

pkg/importer/dataStream_test.go
```diff
@@ -14,10 +14,10 @@ import (
 	"testing"
 
 	"kubevirt.io/containerized-data-importer/pkg/image"
-	"kubevirt.io/containerized-data-importer/test/framework"
+	"kubevirt.io/containerized-data-importer/tests/framework"
```
awels: Should this be like this, or was it just a global search and replace that went wrong?

pkg/importer/datastream_ginkgo_test.go
```diff
@@ -11,25 +11,21 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"kubevirt.io/containerized-data-importer/pkg/image"
-	"kubevirt.io/containerized-data-importer/pkg/importer"
 	imagesize "kubevirt.io/containerized-data-importer/pkg/lib/size"
-	f "kubevirt.io/containerized-data-importer/test/framework"
+	f "kubevirt.io/containerized-data-importer/tests/framework"
```
awels: Same, don't think this is right. There are several more of these.

pkg/importer/dataStream_test.go
copejon: framework/ was moved under tests/ from test/.  Given that we don't have a unified package for the refactored unit tests, this seemed like a reasonable place for it.  I'm open to suggestions though.

copejon: I've updated the PR description to list the packages / files that were moved.

jeffvance: @copejon My pr 349 creates _tests/framewor/_ for common e2e funcs, so we need to coordinate

pkg/impoter/importer_suite_test.go
```diff
@@ -1,4 +1,4 @@
-package importer
+package impoter
```
jeffvance: typo?

pkg/impoter/importer_test.go
```diff
@@ -1,4 +1,4 @@
-package importer
+package impoter
```
jeffvance: typo?

pkg/importer/dataStream_test.go
copejon: @jeffvance This should be handled pretty well by git in merging.

pkg/impoter/importer_suite_test.go
copejon: yes! Thanks, will fix

pkg/impoter/importer_test.go
copejon: ditto

pkg/importer/dataStream_test.go
jeffvance: I know you know: but this file contains the new `framework` pkg. Is is still relevant to the renaming and restructuring of unit tests?

copejon: It shouldn't be.  Git will resolve the diffs between these files.  It won't have an effect on unit tests that importer framework.

pkg/importer/datastream_ginkgo_test.go
copejon: The `test` package is being removed and the `framework` package is being moved to `tests`.  This is the reason for the change.

pkg/importer/dataStream_test.go
jeffvance: This is probably ok. My concern is that the Framework struct and methods today are motivated and evolving based on e2e usage. That _may_ conflict with unit test usage down the road.

copejon: @jeffvance The changes here do not affect the framework struct at all.  That said, maybe this 
belongs in a separate package.  `tests/utils/`?

pkg/importer/dataStream_test.go
```diff
@@ -285,12 +285,12 @@ func TestCopyImage(t *testing.T) {
...
 
 func createTestData() map[string]string {
-	imageDir, _ := filepath.Abs(testImagesDir)
+	imageDir, _ := filepath.Abs(TestImagesDir)
 
 	// xz data
-	xzfile, _ := framework.FormatTestData(filepath.Join(imageDir, "tinyCore.iso"), os.TempDir(), image.ExtXz)
-	gzfile, _ := framework.FormatTestData(filepath.Join(imageDir, "tinyCore.iso"), os.TempDir(), image.ExtGz)
-	xztarfile, _ := framework.FormatTestData(filepath.Join(imageDir, "tinyCore.iso"), os.TempDir(), []string{image.ExtTar, image.ExtGz}...)
+	xzfile, _ := utils.FormatTestData(filepath.Join(imageDir, "tinyCore.iso"), os.TempDir(), image.ExtXz)
```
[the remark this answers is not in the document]
copejon: It has thus far in CI.  This refactor doesn't change anything wrt to test file generation.

pkg/importer/dataStream_test.go
```diff
@@ -300,7 +300,7 @@ func createTestData() map[string]string {
 }
 
 func Test_dataStream_constructReaders(t *testing.T) {
-	imageDir, _ := filepath.Abs(testImagesDir)
+	imageDir, _ := filepath.Abs(TestImagesDir)
```
jeffvance: you'll want to look at the changes to this func.  Since the filename has changed I'm not sure how it will get merged. Bottom line is I fixed a bug where `constructReaders` was called twice. I changed the test to verify number of readers instead of err.

jeffvance:
A couple of comments but lgtm

pkg/importer/dataStream_test.go
copejon: This PR does not change code other than import paths (here it removes the package prefix since it's now in the same package).  There should be no issues with changes other PRs have already made to this test.

jeffvance:
lgtm

PR #352, opened 2018-08-17
Pass ENVIRONMENT variables as test runner arguments.

awels:
- Added the ability to set environment variables to override arguments
  for the e2e test runner.

Signed-off-by: Alexander Wels <awels@redhat.com>

PR #353, opened 2018-08-17
Unify redundant test scripts under test target and aliases

copejon:
The Makefile contains 2 separate targets for functional testing.  This PR looks to combine these under the `test-functional` target by abstracting the driver script and adding custom args to the `test-functional` make target.

The idea here is to have a single script to drive testing.  The script is only picky about the package path meeting the go test ./path/to/pkg/... format, everything else is considered a test arguement.

Make targets `test-unit` and `test-functional` delineate the tests and will pass the relevant test args to the script according to what the suites expect.

- [ ] Update hack/README.md


copejon:
@awels can you TAL also please

hack/build/build-test.sh
```diff
@@ -17,19 +17,37 @@
...
     oc=${kubectl}
 fi
 
-echo ${kubectl}
+echo "kubectl: $kubectl"
 
-kubeconfig=${CDI_DIR}/cluster/.kubeconfig
-# ${TESTS_OUT_DIR}/tests.test -kubeconfig=${kubeconfig} -oc-path=${oc} -kubectl-path=${kubectl} -test.timeout 90m ${FUNC_TEST_ARGS}
-go test -v ./tests/... -args -ginkgo.v -kubeconfig=${kubeconfig} -oc-path=${oc} -kubectl-path=${kubectl} -test.timeout 90m ${FUNC_TEST_ARGS}
+go test -v -test.timeout 90m ${target} -args ${test_args}
```
awels: We will need to converge this with #352 so all those environment variables are passed properly.

copejon:
@awels 
> We will need to converge this with #352 so all those environment variables are passed properly.
Agreed.  Can you PTAL at the latest commit here.

One variable - oc-path - is a test flag but isn't used anywhere.  Even if it is, it looks like it would have been set to `cluster/.kubectl` anyway.  Is this an artifact or copy/paste?  And why have 2 variables with distinct names reference the same executable?

awels:
@copejon So I know it looks a little funky, here is why. The oc-path is to create the path to oc to run oc commands. With kubevirtci when we use the open shift provider, it copies the oc command into the same file name .kubectl as when we would be using the k8s provider (that explains why we are setting it the same as kubectl).

The reason we have it as a separate variable is so we can create open shift specific tests that require oc instead of kubectl, and use OC instead of kubectl.

I agree right now its not needed as we don't have any of them, but that is the reason we have so that its available when we do. If you feel we should not have it until we actually need it, we can remove it without issues.

copejon:
@awels 
I see your point that tracking the `oc` path discretely makes sense over having a generic cluster client variable for both k8s and oc. This is moot though if the `oc` and `kubectl` executables are mutually exclusive files.

> With kubevirtci when we use the open shift provider, it copies the oc command into the same file name .kubectl as when we would be using the k8s provider (that explains why we are setting it the same as kubectl).

Since we can't change the way kubevirtCI writes the `cluster/.kubectl` file, I think it would be better to make the distinction at the test suite Describe level and let the cluster client variable be generic.  The distinction between the cluster providers can be made with tags we define and target  e.g. `Describe("[Openshift] Basic Tests", func(){ ... })` and `-ginkgo.focus="[Openshift].*"`  (_In some future PR_)

awels:
Okay, I have no issues with removing the oc-path flag, it really is only important for when we want to run oc specific oc commands, we can use kubectl for everything else.

copejon:
@awels Sounds good, I'll look at that in another PR.

awels:
Added a comment for the [OC] tag.

hack/README.md
```diff
@@ -41,7 +41,7 @@ The standard workflow is performed inside a helper container to normalize the bu
     - No `build-cloner` target exists as the code is written in bash
 - `test`: execute all tests
     - `test-unit`: execute all tests under `./pkg`
-    - `test-functional`: execute all tests under `./test`
+    - `test-functional`: execute all tests under `./tests`
```
jeffvance: what do you think about adding a comment here that functional tests are also known as e2e tests?

hack/build/config.sh
```diff
@@ -30,12 +30,37 @@ VERBOSITY=${VERBOSITY:-1}
...
 function allPkgs {
     ret=$(sed "s,kubevirt.io/containerized-data-importer,${CDI_DIR},g" <(go list ./... | grep -v "pkg/client" | sort -u ))
     echo "$ret"
 }
 
-KUBERNETES_IMAGE="k8s-1.10.4@sha256:ee6846957b58e1f56b240d9ba6410f082e4787a4c4f1e0d60f6b907b76146b3e"
-OPENSHIFT_IMAGE="os-3.10.0@sha256:cdc9f998e19915b28b5c5be1ccc4c6fa2c8336435f38a37855f75b206977cbc2"
-
-KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-k8s-1.10.4}
+function parseTestOpts {
```
jeffvance: add comment that this function sets `test_args` and `pkgs`

Makefile
```diff
@@ -117,7 +120,6 @@ cluster-sync-importer: cluster-sync
 cluster-sync-cloner: WHAT = cmd/cdi-cloner
 cluster-sync-cloner: cluster-sync
 
-functest:
-	./hack/build/functests.sh
-
-
+functest-image-host: WHAT=tools/cdi-func-test-file-host-init
```
awels: I don't think these are needed anymore. build-cdi-func-test-file-host.sh is called with build, and then both init and http are part of the build target as well.

hack/build/config.sh
copejon: will do

hack/README.md
copejon: That might be more confusing that illuminating I think.  Also, kubevirt defines e2e as something else i think.

jeffvance: ok, didn't realize kubevirt has a def for e2e. Ok, never-mind.

hack/build/run-functional-tests.sh
```diff
@@ -16,11 +16,24 @@
...
+# functional testing
+KUBECTL=${KUBECTL:-${CDI_DIR}/cluster/.kubectl}
+KUBECONFIG=${KUBECONFIG:-${CDI_DIR}/cluster/.kubeconfig}
+KUBE_MASTER_URL=${KUBE_MASTER_URL:-""}
+CDI_NAMESPACE=${CDI_NAMESPACE:-kubesystem}
+
 parseTestOpts "${@}"
 
-test_args="${test_args} -ginkgo.v -oc-path=${KUBECTL} -kubectl-path=${KUBECTL} -kubeconfig=${KUBECONFIG}"
+arg_master="${KUBE_MASTER_URL:+-master=$KUBE_MASTER_URL}"
```
jeffvance: bash _magic_ !

hack/build/run-functional-tests.sh
```diff
@@ -16,11 +16,24 @@
...
+script_dir="$(readlink -f $(dirname $0))"
 source hack/build/config.sh
 source hack/build/common.sh
 
+# functional testing
+KUBECTL=${KUBECTL:-${CDI_DIR}/cluster/.kubectl}
+KUBECONFIG=${KUBECONFIG:-${CDI_DIR}/cluster/.kubeconfig}
+KUBE_MASTER_URL=${KUBE_MASTER_URL:-""}
+CDI_NAMESPACE=${CDI_NAMESPACE:-kubesystem}
+
```
jeffvance: In pr 352 the env var is named KUBECTL_PATH and the readme shows that. Is this changing?
Also, what's the current status of an OC_PATH env var?
@copejon @awels

copejon: For now i'm leaving OC_PATH in as it's still a flag in the tests.  It's inert, so there's not a need to remove it right now.  I'd prefer to do that in another PR.  I'll fix up the readme to reflect these changes.

awels:
lgtm

jeffvance:
lgtm

PR #354, opened 2018-08-19
Add regular PVC creation e2e test.

awels:
- Added e2e test to see if regular PVC doesn't get populated.
- Moved PVC utility functions under framework.
- Added exec utility functions to execute commands against running
  Pods.


jeffvance:
why so many files?

tests/framework/exec_util.go
```diff
@@ -0,0 +1,143 @@
...
+	"k8s.io/client-go/tools/remotecommand"
+
+	. "github.com/onsi/gomega"
+)
+
+// ExecOptions passed to ExecWithOptions
+type ExecOptions struct {
+	Command []string
+
+	Namespace     string
```
jeffvance: name + ns are fine. Did you consider passing *v1.Pod?

tests/framework/exec_util.go
```diff
@@ -0,0 +1,143 @@
...
+	PodName       string
+	ContainerName string
+
+	Stdin         io.Reader
+	CaptureStdout bool
+	CaptureStderr bool
+	// If false, whitespace in std{err,out} will be removed.
+	PreserveWhitespace bool
```
jeffvance: note that false is the default (zero) value for a bool.

tests/framework/exec_util.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// ExecWithOptions executes a command in the specified container,
+// returning stdout, stderr and error. `options` allowed for
+// additional parameters to be passed.
+func (f *Framework) ExecWithOptions(options ExecOptions) (string, string, error) {
+	config, err := LoadConfig()
```
jeffvance: should the rest client be part of the core framework struct, along with k8s and cdi clients?

tests/framework/exec_util.go
```diff
@@ -0,0 +1,143 @@
...
+		Namespace(options.Namespace).
+		SubResource("exec").
+		Param("container", options.ContainerName)
+	req.VersionedParams(&v1.PodExecOptions{
+		Container: options.ContainerName,
+		Command:   options.Command,
+		Stdin:     options.Stdin != nil,
+		Stdout:    options.CaptureStdout,
+		Stderr:    options.CaptureStderr,
+		TTY:       tty,
```
jeffvance: false is the default for bools

tests/framework/exec_util.go
```diff
@@ -0,0 +1,143 @@
...
+	// If false, whitespace in std{err,out} will be removed.
+	PreserveWhitespace bool
+}
+
+// ExecWithOptions executes a command in the specified container,
+// returning stdout, stderr and error. `options` allowed for
+// additional parameters to be passed.
+func (f *Framework) ExecWithOptions(options ExecOptions) (string, string, error) {
+	config, err := LoadConfig()
+	Expect(err).NotTo(HaveOccurred(), "failed to load restclient config")
```
jeffvance: you have to decide if framework funcs should call `Expect`. If yes then typically the func name should end with `...OrDie` so it's more obvious that the func may not return. This is especially true of the func calls `Expect` or `Fali` or `panic`  **and** also has an error return value.

tests/framework/exec_util.go
jeffvance: Do we want to load the config every time this func is called?

tests/framework/exec_util.go
```diff
@@ -0,0 +1,143 @@
...
+func (f *Framework) ExecCommandInContainer(podName, containerName string, cmd ...string) string {
+	stdout, _, err := f.ExecCommandInContainerWithFullOutput(podName, containerName, cmd...)
+	Expect(err).NotTo(HaveOccurred(),
+		"failed to execute command in pod %v, container %v: %v",
+		podName, containerName, err)
+	return stdout
+}
+
+func (f *Framework) ExecShellInContainer(podName, containerName string, cmd string) string {
+	return f.ExecCommandInContainer(podName, containerName, "/bin/sh", "-c", cmd)
```
jeffvance: A more common pattern (but may not fit here) is for the exported func name to call an non-exported func passing it some fields from the exported receiver. So in this case `ExecCommandInContainer` would be private, unless it is also useful on its own as an exported func

tests/framework/framework.go
```diff
@@ -26,6 +26,7 @@ import (
 const (
 	NsCreateTime = 30 * time.Second
 	NsDeleteTime = 5 * time.Minute
+	DefaultTimeout = 30 * time.Second
```
jeffvance: The 2 above consts are specific to a resource (ns). I see `DefaultTimeout` is being used for pvcs so I think it's better to chose a name indicating that.

tests/framework/framework.go
```diff
@@ -83,27 +84,27 @@ func NewFramework(prefix string) *Framework {
...
+		f.OcPath = *ocPath
+	}
+	if cdiInstallNs != nil {
+		f.CdiInstallNs = *cdiInstallNs
+	}
+	if kubeConfig != nil {
+		f.KubeConfig = *kubeConfig
+	}
+	if master != nil {
+		f.Master = *master
```
jeffvance: thanks for formatting. I ran `gofmt -w -s` a few times but must have missed this...

tests/framework/framework.go
```diff
@@ -213,6 +214,8 @@ func DeleteNS(c *kubernetes.Clientset, ns string) error {
 
 // Gets an instance of a kubernetes client that includes all the CDI extensions.
 func (f *Framework) GetCdiClient() (*cdiClientset.Clientset, error) {
+	fmt.Println(f.Master)
```
jeffvance: is this debugging? If not then you want to use the GinkgoWriter is seen in other framework uses of `fmt.Fprintf`

tests/framework/framework.go
```diff
@@ -229,13 +232,17 @@ func (f *Framework) GetKubeClient() (*kubernetes.Clientset, error) {
...
-	config, err := clientcmd.BuildConfigFromFlags(master, kubeconfig)
+	config, err := LoadConfig()
 	if err != nil {
 		return nil, err
 	}
 	return GetKubeClientFromRESTConfig(config)
 }
 
+func LoadConfig() (*rest.Config, error) {
+	return clientcmd.BuildConfigFromFlags(*master, *kubeConfig)
```
jeffvance: don't refer to globals when these ptrs are passed to `GetKubeClientFromFlags`. Just need to pass them to `LoadConfig` too.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,188 @@
...
+	err := wait.PollImmediate(poll, DefaultTimeout, func() (bool, error) {
+		var err error
+		pvc, err = f.K8sClient.CoreV1().PersistentVolumeClaims(namespace).Create(def)
+		if err != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		Fail("Unable to create PVC: " + def.GetName() + ", error: " + err.Error())
```
jeffvance: so func name should end (by convention) with `OrDie`. Or, better, imo, to return the error to the caller since this is a framework func and let the caller call Expect, etc.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Creates a PVC in the passed in namespace from the passed in PersistentVolumeClaim definition.
+// An example of creating a PVC without annotations looks like this:
+// CreatePVCFromDefinition(client, namespace, NewPVCDefinition(name, size, nil))
+func (f *Framework) CreatePVCFromDefinition(namespace string, def *k8sv1.PersistentVolumeClaim) *k8sv1.PersistentVolumeClaim {
+	var pvc *k8sv1.PersistentVolumeClaim
+	err := wait.PollImmediate(poll, DefaultTimeout, func() (bool, error) {
+		var err error
+		pvc, err = f.K8sClient.CoreV1().PersistentVolumeClaims(namespace).Create(def)
+		if err != nil {
```
jeffvance: what if err == apierrs.AlreadyExists. See framework's `CreateNS` func.

awels:
Yes the large number of vendor files is due to the SPDY stuff. Also I copied exec_util.go almost whole sale from k8s e2e and made some small changes to have it fit in our version of Framework, but all the code is in tact. I suppose I can alter it a little more based on the comments (but we should start making reports against k8s at that point ;).

tests/framework/framework.go
awels: It was debugging, I forgot to remove them.

tests/framework/framework.go
awels: Actually that was a bug, we create the Framework struct in each high level test file, but we only parse the flags once, so we would only copy the values into the Framework struct the first time. So notice I changed the if to only be around parse.

tests/framework/exec_util.go
awels: Well the rest client is part of the k8s and cdi client. See the DiscoveryClient struct which is part of ClientSet. In this particular case we are interested in the config of the client, not the client itself. I can see that being useful as part of Framework struct.

awels: If we add it to the Framework, we won't have to worry about it.

tests/framework/exec_util.go
awels: This would not be relevant either if we add the config to the Framework struct.

tests/framework/exec_util.go
awels: Tell that to the Google engineer who wrote this :), but we can always improve it here.

tests/framework/exec_util.go
awels: No since I copied and pasted this from the k8s e2e framework. Seems like that part of the code is somewhat old and could use some improvements.

tests/framework/exec_util.go
awels: The way I read this, all these are useful on their own in particular cases and thus they are all exported.

tests/framework/framework.go
awels: Fixed

tests/framework/pvc_utils.go
awels: Fixed

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,188 @@
...
+// Creates a PVC definition using the passed in name and requested size.
+// You can use the following annotation keys to request an import or clone. The values are defined in the controller package
+// AnnEndpoint
+// AnnSecret
+// AnnCloneRequest
+func NewPVCDefinition(pvcName string, size string, annotations map[string]string) *k8sv1.PersistentVolumeClaim {
+	return &k8sv1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        pvcName,
+			Annotations: annotations,
```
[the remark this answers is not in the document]
awels: Good point labels seem to be useful for a lot of cases, so we should accept them.

tests/framework/exec_util.go
jeffvance: @awels Yes, that is one of the challenges of copying code -- you're never sure if what you're grabbing is actually the preferred way of doing something. K8s e2e is a good example of both old and current implementations of e2e testing, and I do not always know which is which either. What you have above is fine. I was just thinking that having a pointer to a pod might be better because 1) it's just one field, 2) you likely have already created this pod, and 3) maybe future enhancement to these functions will be need other pod fields in the exec options.

tests/framework/exec_util.go
jeffvance: Feel free to move it to the Framework struct if you see general usefulness.

tests/framework/exec_util.go
jeffvance: Agree. but is the `TTY` const used anywhere else? See line [70](https://github.com/kubevirt/containerized-data-importer/pull/354/files#diff-dbd89751a11d5d9f645bd240b9f4f686R70) where `false` is used instead of the TTY const.

tests/framework/exec_util.go
jeffvance: Can you explain more? I don't understand your reply.

tests/framework/exec_util.go
jeffvance: @awels maybe you have already done this, but if not can you push a new commit that has config in the Framework?

jeffvance:
@awels can you justify your vendoring in all the code for spdy? client-go has a spdy pkg so it's unclear to me why we need all of the related docker files.

awels:
@jeffvance That is what glide adds after I do a glide cc and then glide up -v, it adds a ton of files. I am assuming glide knows what it is doing.

tests/import_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+const (
+	testSuiteName   = "Importer Test Suite"
+	namespacePrefix = "importer"
+)
+
+var _ = Describe(testSuiteName, func() {
+	f := framework.NewFramework(namespacePrefix)
+
+	Describe("Creating standard PVC should not trigger importer", func() {
```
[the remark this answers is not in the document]
copejon: Sounds good - let's start that convo in the issues and ping QE folks.  It'd be good to get their input on this.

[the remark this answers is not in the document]
jeffvance: See issue [359](https://github.com/kubevirt/containerized-data-importer/issues/359)

tests/framework/framework.go
jeffvance: per mention from @copejon none of the _if flag == nil {_ checks are necessary and the code is cleaner w/o them.

tests/framework/framework.go
```diff
@@ -24,14 +24,15 @@ import (
 )
 
 const (
-	NsCreateTime = 30 * time.Second
-	NsDeleteTime = 5 * time.Minute
+	NsCreateTime   = 30 * time.Second
+	NsDeleteTime   = 5 * time.Minute
+	DefaultTimeout = 30 * time.Second
```
jeffvance: again, pls use a more descriptive const name.

tests/framework/framework.go
```diff
@@ -47,6 +48,9 @@ type Framework struct {
 	K8sClient *kubernetes.Clientset
 	// cdi client
 	CdiClient *cdiClientset.Clientset
+	// REST client config.
+	RestConfig *rest.Config
+
```
jeffvance: nit: for consistency with the rest of this struct delete this empty line.

tests/framework/framework.go
```diff
@@ -76,34 +80,42 @@ func init() {
 // NewFramework makes a new framework and sets up the global BeforeEach/AfterEach's.
 // Test run-time flags are parsed and added to the Framework struct.
 func NewFramework(prefix string) *Framework {
+	var err error
```
jeffvance: defining err here is very C-like. The go approach is, when possible, define your local vars directly in an assignment. When that won't work (say due to scoping rules) then define the var right before it's needed. In this case you do not need a definition for `err`

tests/framework/framework.go
```diff
@@ -76,34 +80,42 @@ func init() {
...
+	if kubeConfig != nil {
+		f.KubeConfig = *kubeConfig
+	}
+	if master != nil {
+		f.Master = *master
+	}
+
+	f.RestConfig, err = f.LoadConfig()
+	if err != nil {
+		// Can't use Expect here due to fail handler not being registered at this point
```
jeffvance: when is the Fail handler registered? I thought it was in the zz_test_suite.go code, which is run prior to `NewFramework`

tests/framework/framework.go
jeffvance: I think it's probably better to add an err return to this func. Then we don't have to think about calling Fail or Expect -- we just return the error to the caller.

tests/framework/framework.go
```diff
@@ -76,34 +80,42 @@ func init() {
...
+		fmt.Fprintf(GinkgoWriter, "ERROR, unable to load RestConfig")
```
jeffvance: This 
Probably should include the cdi _errors_ import so that err can be wrapped and a stack trace is automated when the err is formatted. In this case, I'd write something like:
```

```

tests/framework/framework.go
```diff
@@ -225,15 +237,15 @@ func (f *Framework) GetCdiClient() (*cdiClientset.Clientset, error) {
 }
 
 func (f *Framework) GetKubeClient() (*kubernetes.Clientset, error) {
-	return GetKubeClientFromFlags(f.Master, f.KubeConfig)
+	return f.GetKubeClientFromFlags(f.Master, f.KubeConfig)
```
jeffvance: if you're changing this to a method then you do not need to pass the 2 args here since they are defined in the receiver.

tests/framework/framework.go
```diff
@@ -225,15 +237,15 @@ func (f *Framework) GetCdiClient() (*cdiClientset.Clientset, error) {
...
 }
 
-func GetKubeClientFromFlags(master, kubeconfig string) (*kubernetes.Clientset, error) {
-	config, err := clientcmd.BuildConfigFromFlags(master, kubeconfig)
-	if err != nil {
-		return nil, err
-	}
-	return GetKubeClientFromRESTConfig(config)
+func (f *Framework) GetKubeClientFromFlags(master, kubeconfig string) (*kubernetes.Clientset, error) {
```
jeffvance: ditto re the 2 parms need needed since it's now a method.

tests/import_test.go
```diff
@@ -0,0 +1,30 @@
...
+		It("PVC should remain empty after being created", func() {
+			pvc, err := f.CreatePVCFromDefinition(f.Namespace.GetName(), framework.NewPVCDefinition("no-import", "1G", nil, nil))
```
jeffvance: why use `GetName` vs `Namespace.Name`?

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+		pvc, err = f.K8sClient.CoreV1().PersistentVolumeClaims(namespace).Create(def)
+		if err == nil || apierrs.IsAlreadyExists(err) {
+			return true, nil
+		}
+		return false, err
+	})
+	if err != nil {
+		return nil, err
+	}
+	return pvc, nil
```
jeffvance: so we know here there are no api errors but the pvc may not yet be created.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// Delete the passed in PVC
+func (f *Framework) DeletePVC(pvc *k8sv1.PersistentVolumeClaim) error {
+	err := wait.PollImmediate(poll, DefaultTimeout, func() (bool, error) {
```
jeffvance: this can be simplified as:
```
     return wait.PollImmediate(poll, DefaultTimeout, func() (bool, error) {
               ...
     })
```

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+				},
+			},
+		},
+	}
+}
+
+// Wait for the PVC to be in a particular phase (Pending, Bound, or Lost)
+func (f *Framework) WaitForPersistentVolumeClaimPhase(phase k8sv1.PersistentVolumeClaimPhase, pvcName string) error {
+	wait.PollImmediate(poll, DefaultTimeout, func() (bool, error) {
+		pvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(f.Namespace.GetName()).Get(pvcName, metav1.GetOptions{})
```
jeffvance: why assume here that the ns is the one defined in the Framework. What if it's the cloner's ns2? I'd like to see consistent use of ns. Now it's sometimes it's passed, other times used from the Framework.

jeffvance: ditto re using GetName()

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+		if err != nil {
+			return false, err
+		} else {
```
jeffvance: more go-like to not use `else` here. Instead (several changes combined below):
```
	err := wait.PollImmediate(poll, DefaultTimeout, func() (bool, error) { <==== always capture err
		pvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Get(pvcName, metav1.GetOptions{})
		if err != nil || pvc.Status.Phase != phase {
			return false, err
		}
	        return true, nil
	})
        if err != nil {
	       return fmt.Errorf("PersistentVolumeClaim %s not in phase %s within %v", pvcName, phase, DefaultTimeout)
        }
       return nil
```

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+func newExecutorPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1.Pod {
+	return &k8sv1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: podName},
+		Spec: k8sv1.PodSpec{
+			RestartPolicy: k8sv1.RestartPolicyNever,
+			Containers: []k8sv1.Container{
+				{
+					Name:    "runner",
+					Image:   "fedora:28",
+					Command: []string{"/bin/sh", "-c", "sleep 5; echo I am an executor pod;"},
```
jeffvance: why?

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+					VolumeMounts: []k8sv1.VolumeMount{
+						{
+							Name:      pvc.GetName(),
```
jeffvance: unless GetName solves some problem I am unaware of I don't see what's wrong w/ `pvc.Name`

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+
+func newExecutorPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1.Pod {
+	return &k8sv1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: podName},
+		Spec: k8sv1.PodSpec{
+			RestartPolicy: k8sv1.RestartPolicyNever,
+			Containers: []k8sv1.Container{
+				{
+					Name:    "runner",
+					Image:   "fedora:28",
```
jeffvance: hard-coding f28 will be problematic down the road...

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+		},
+	}
+}
+
+func (f *Framework) waitTimeoutForPodReady(podName string, timeout time.Duration) error {
+	return wait.PollImmediate(2*time.Second, timeout, f.podRunning(podName, f.Namespace.GetName()))
+}
+
+func (f *Framework) podRunning(podName, namespace string) wait.ConditionFunc {
```
jeffvance: why is ns passed here?

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+	return func() (bool, error) {
+		pod, err := f.K8sClient.CoreV1().Pods(namespace).Get(podName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		switch pod.Status.Phase {
+		case k8sv1.PodRunning:
+			return true, nil
+		case k8sv1.PodFailed, k8sv1.PodSucceeded:
```
jeffvance: why is this `case` needed since it doesn't add anything different from dropping through?

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+			return false, nil
+		}
+		return false, nil
+	}
+}
+
+// Verify a particular PVC is empty, returns true if it is, false otherwise.
```
jeffvance: confusing comment.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+func (f *Framework) VerifyPVCIsEmpty(pvc *k8sv1.PersistentVolumeClaim) bool {
+	executorPod := f.CreateExecutorPodWithPVC("verify-pvc-empty", pvc)
```
jeffvance: It seems more efficient to create one executor pod at the beginning of the test (outer Describe) and then use that one pod for all node-base commands. Is there a problem doing that?

jeffvance:
@awels I think this pr will be a model/template for many future cdi e2e tests. Therefore I am being fairly picky and critical in my review. Pls don't take it personally. I just want this 1st e2e test pr to be as good an example as possible.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+	err := f.waitTimeoutForPodReady(executorPod.GetName(), DefaultTimeout)
+	Expect(err).To(BeNil())
```
jeffvance: My vote is that all framework funcs and methods return err rather than calling `Expect`, `Fail` or panic. I think this is the best practice for a lower-level, common framework pkg.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+	output := f.ExecShellInPod(executorPod.GetName(), "ls -1 /pvc | wc -l")
```
jeffvance: pending your comment, this could be simply `executorPod,Name`

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+	f.DeletePod(executorPod)
+	return strings.Compare("0", output) == 0
```
jeffvance: Is there no way to capture the exit code of the ls command?

tests/import_test.go
```diff
@@ -0,0 +1,30 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			// Wait a while to see if CDI puts anything in the PVC.
+			time.Sleep(framework.DefaultTimeout)
```
jeffvance: need specific timeout names

tests/import_test.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var _ = Describe(testSuiteName, func() {
+	f := framework.NewFramework(namespacePrefix)
+
+	Describe("Creating standard PVC should not trigger importer", func() {
+		It("PVC should remain empty after being created", func() {
+			pvc, err := f.CreatePVCFromDefinition(f.Namespace.GetName(), framework.NewPVCDefinition("no-import", "1G", nil, nil))
+			Expect(err).ToNot(HaveOccurred())
+			// Wait a while to see if CDI puts anything in the PVC.
```
jeffvance: the concept of "a while" is ambiguous. In most (but not all) cases k8s polls for a state to lessen the need for an unconditional wait. Maybe we need to Poll the executor-pod with a cmd to test the existence of _disk.img_?

tests/import_test.go
```diff
@@ -0,0 +1,30 @@
...
+			time.Sleep(framework.DefaultTimeout)
+			Expect(f.VerifyPVCIsEmpty(pvc)).To(BeTrue())
+			// Not deleting PVC as it will be removed with the NS removal.
```
jeffvance: This is true, however it is better practice to explicitly delete all resources created by the test. That way, if the pvc cannot be deleted you'll be able to report a meaningful error compared to the err you see if the pvc delete fails while deleting the ns,

tests/framework/exec_util.go
awels: @jeffvance In general I prefer structs over a bunch of strings as it makes it much clearer what we are talking about. In this particular case I am on the fence. As you say its highly likely someone will have the pod already and passing it in is easy. On the other hand we will internally simply do pod.GetName and pod.GetNamespace. And if someone didn't have the pod, they will have to construct it from the name and namespace, only for us to just grab those two fields.

So in the case that the caller has the pod already, its easy for them to pass the two fields we are interested in. And in the case they don't its also easy for them to pass the fields as they have them already. And they don't have to construct the Pod struct at all.

tests/framework/exec_util.go
awels: Right, the comment is just explaining the different behaviors of the possible values.

tests/framework/framework.go
awels: Sounds good, let me change that.

tests/framework/framework.go
awels: @jeffvance I actually put the Expect in there, and it errored out with a message saying the Fail handler wasn't registered yet. I was shocked at this, so I put a note in there to remind me why I didn't put an Expect in there.

tests/framework/framework.go
awels: So this DefaultTimeout is a generic timeout value in the case you don't have a more specific const defined. I won't want to end up with 50 XyTimeout, one for each type of entity. If we do want specific timeouts for a particular entity because of something in that entity makes it necessary to have different defaults, I would suggest we put it in the const block for that entity utility file, instead of in the framework const block.

This particular one, is generic, and the Ns ones are in here because they are part of the framework file.

tests/framework/framework.go
awels: Done

tests/framework/framework.go
awels: Gotcha, well I can assign with := so removing.

tests/framework/framework.go
awels: panic: You are trying to make an assertion, but Gomega's fail handler is nil.
If you're using Ginkgo then you probably forgot to put your assertion in an It().
Alternatively, you may have forgotten to register a fail handler with RegisterFailHandler() or RegisterTestingT().

That is the error message. So I think the problem is that we are calling the Expect outside of the It block.

tests/framework/framework.go
awels: Sounds good, will change.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+	k8sv1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	defaultPvcMountPath = "/pvc"
+	poll                = 2 * time.Second
```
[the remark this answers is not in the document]
awels: Yeah copied it from k8s code. renamed it to defaultPollInterval which is what it is.

tests/framework/pvc_utils.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// Creates a PVC in the passed in namespace from the passed in PersistentVolumeClaim definition.
+// An example of creating a PVC without annotations looks like this:
+// CreatePVCFromDefinition(client, namespace, NewPVCDefinition(name, size, nil))
+func (f *Framework) CreatePVCFromDefinition(namespace string, def *k8sv1.PersistentVolumeClaim) (*k8sv1.PersistentVolumeClaim, error) {
```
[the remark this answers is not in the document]
awels: Yes or if I need to create a PVC in an existing NS for whatever reason. Because a test can potentially work in multiple namespaces, basically all our utility functions will need a namespace parameter.

tests/import_test.go
awels: Its a Java habit, where its always preferable to use the getter/setter over direct accessing the field, even if you have access to the field. This allows sub classes (in java) to override the setter/getter and get the correct behavior. Obviously it makes no difference in GO. @jeffvance is there a preference in GO to use the field over the getter?

tests/framework/pvc_utils.go
awels: So that is an interesting question, with a complex answer. PVC phase has 3 options. Pending/Bound/Lost. And depending on the type of underlying storage Bound may not happen until a pod attaches to the PVC (in particular local volume storage, as the PV used will be on the host the pod is scheduled on). Or it might be bound immediately to the PV (most other storage). So what exactly will we wait on? We can't check the phase, unless we want to check for either Pending or Bound, but I believe the phase goes to Pending immediately so there is nothing to wait on.

tests/framework/pvc_utils.go
awels: Good point. fixed

tests/framework/pvc_utils.go
awels: You are right, I will make namespace a parameter to all the functions, for the reason that we might not be using the Namespace in the framework for that particular operation.

awels: So I don't have a strong opinion either way, it seems you would like .Name, so I will update all the GetName to be Name

tests/framework/pvc_utils.go
awels: See my answers above, its my Java habit, but I don't have a strong opinion either way.

tests/framework/pvc_utils.go
awels: Correct, we could make an executor pod when we create the framework, and use that throughout, if we can make a long running pod. I haven't been able to figure out how to make a long running pod. If you know, let me know how, and I will do that.

tests/framework/pvc_utils.go
awels: Changed, like I said before no real preference one way or the other, so lets try and be consistent with just .Name

tests/framework/pvc_utils.go
awels: I am not sure I understand what you are getting at, how would the exit code tell me if the PVC is empty or not? Or are you trying to see if something happened with the ls command?

tests/import_test.go
awels: In almost all other cases we can do that, but this is the negative test case (since I don't have http server yet to try a positive case). The test is basically create a PVC without annotation, and wait a while and check if nothing changed. A while like you said is ambiguous, but how else do you check nothing changed?

tests/import_test.go
awels: So here is why I decided to let the NS removal clean up stuff. When the NS cleans up stuff, its all internal in the api server. If we do it manually, we are hammering the api server with different request, so the load on it increases, and we increase the network traffic as well. At the cost of less meaningful messages if something goes wrong during the deleting of the NS.

tests/import_test.go
awels: See my comment on the specific vs generic timeout names. If we want specific timeouts for an entity we can do that, but I think its useful to have a generic timeout.

tests/framework/pvc_utils.go
awels: This is a good question, and I am not sure I have a good answer. Where do we draw the line on what is a utility function, and what is a framework function. This one in particular is a good one to ask that question. Its a higher level function that calls a bunch of lower level ones, so I would lean more towards a utility function. But the other ones are all single purpose functions, but they will now all take namespace as a parameter.

Maybe put them all in a utility package, and put matching functions without the namespace in the framework, and delegate the work to the utility functions with the namspace filled in.

In short, I don't know, maybe we need to talk about that.

awels:
@jeffvance I believe I have addressed all your comments, sorry for the big PR.

tests/import_test.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		Fail("Unable to create framework struct")
+	}
+
+	Describe("Creating standard PVC should not trigger importer", func() {
+		It("PVC should remain empty after being created", func() {
+			pvc, err := f.CreatePVCFromDefinition(utils.NewPVCDefinition("no-import", "1G", nil, nil))
+			Expect(err).ToNot(HaveOccurred())
+			// Wait a while to see if CDI puts anything in the PVC.
+			time.Sleep(waitForCDIImportTimeout)
```
awels: @jeffvance I am sure you are not a fan of the sleep here, but for this particular test I see no other option. We are basically verifying that nothing happens. So we need to pick a particular time on how long to wait for 'nothing' to change.

tests/framework/pvc.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (f *Framework) DeletePVC(pvc *k8sv1.PersistentVolumeClaim) error {
+	return utils.DeletePVC(f.K8sClient, f.Namespace.Name, pvc)
+}
+
+func (f *Framework) WaitForPersistentVolumeClaimPhase(phase k8sv1.PersistentVolumeClaimPhase, pvcName string) error {
+	return utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, f.Namespace.Name, phase, pvcName)
+}
+
+func (f *Framework) CreateExecutorPodWithPVC(podName string, pvc *k8sv1.PersistentVolumeClaim) (*k8sv1.Pod, error) {
```
awels: This function crossed over into Pods and PVCs, so not sure if I should move that into a separate Pod file or keep it here. I kept it here for now, but I am not opposed to moving it into a pod.go with the other related Pod functions.

tests/import_test.go
```diff
@@ -0,0 +1,35 @@
...
+			Expect(framework.VerifyPVCIsEmpty(f, pvc)).To(BeTrue())
+			// Not deleting PVC as it will be removed with the NS removal.
```
awels: I also know you prefer me to delete this PVC explicitly as well. I explained that I prefer to let the NS cascade the removal because its fewer requests to the API server, and the internal delete of the NS is faster than doing everything explicitly.

tests/framework/exec_util.go
jeffvance: @awels I'm ok leaving it the way you originally wrote it. If there's a need later it can easily be changed.

tests/framework/exec_util.go
jeffvance: Yes, I read too quickly last night. Fine as is.

tests/framework/exec_util.go
jeffvance: Are you thinking of defining a Config struct and embedding it in Framework? If so, consider what the xz compression pkg did where he has a public Config struct, a default constructor, and also a constructor _method_ on Config, where Config fields (non-default) can be set to alter the values set by the constructor. No config args passed, no config struct passed -- pretty clean imo. See [xz](https://github.com/ulikunitz/xz/blob/master/lzma/writer.go#L158)

tests/framework/exec_util.go
jeffvance: By Config are you referring to the flag strings? I don't see a Config in Framework.

tests/framework/framework.go
```diff
@@ -75,41 +78,40 @@ func init() {
...
+		fmt.Fprintf(GinkgoWriter, "   %s = %q\n", f.Name, f.Value.String())
+	})
+	fmt.Fprintf(GinkgoWriter, "**\n")
+	f.KubectlPath = *kubectlPath
+	f.OcPath = *ocPath
+	f.CdiInstallNs = *cdiInstallNs
+	f.KubeConfig = *kubeConfig
+	f.Master = *master
+	// It won't allow me to direct assign f.RestConfig, so using intermediate variable.
+	restConfig, err := f.LoadConfig()
```
jeffvance: missing err check

tests/utils/pvc.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	DefaultPvcMountPath = "/pvc"
+	defaultPollInterval = 2 * time.Second
+	PVCCreateTime       = 30 * time.Second
+	PVCDeleteTime       = 30 * time.Second
+	PVCWaitForTime      = 30 * time.Second
```
jeffvance: +1

tests/utils/pvc.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	DefaultPvcMountPath = "/pvc"
+	defaultPollInterval = 2 * time.Second
+	PVCCreateTime       = 30 * time.Second
```
jeffvance: +1

tests/utils/pvc.go
jeffvance: Even more specific -- what do you think of `PVCPhaseTime = ...`

tests/framework/framework.go
jeffvance: Agree. That must be the issue here.

tests/utils/pvc.go
```diff
@@ -0,0 +1,87 @@
...
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	DefaultPvcMountPath = "/pvc"
+	defaultPollInterval = 2 * time.Second
```
jeffvance: I was ambiguous in my comment lat night. I meant to suggest that you include in the poll const _what_ is being polled. Eg `pvcPollInterval` so we know the resource this applies to is a pvc. The _utils_ pkg may eventually have many different resources that are polled and it nice to have an interval specific to each. Though, in fairness, I commonly see `2s` used as the interval in many unrelated polls.

tests/import_test.go
jeffvance: I honestly don't know if there's a go preference between direct access to an exported field vs. a getter method. I did a quick google search and came up with this [SO answer](https://stackoverflow.com/questions/45189126/go-getter-methods-vs-fields-proper-naming):
_Don't use getters and setters if your fields are exported. This just confuses the interface._

tests/framework/pvc_utils.go
jeffvance: Phase changes are not necessarily ordered (a->b->c). Sometimes, depending on randomness in the cluster, the Phase might go from a->c completely skipping b.
Nevermind, thinking more I think this code is fine. Eg. Bound may never be seen if Phase is set to Lost right after being set to Bound. This is fine and the poll with expire. It's also possible that the poll sees Bound but right after that Phase changes to Lost...

tests/framework/pvc_utils.go
jeffvance: I guess one way to make a long-running pod is a `while true` cmd loop with a sleep X in it.

tests/framework/pvc_utils.go
jeffvance: In my view, all functions needed to set up the running cluster so that general e2e tests can be reliably and repeatedly run should be in the framework pkg. This may eventually consist of several files but now there's just one.  So this includes namespaces, service-acct, cluster roles, services, maybe an executor pod, possibly a general purpose, common storage class...
Helper functions used to manage more specialized resource such as importer pod, pvc, pv, ... should live in a _utils_ pkg. The framework pkg never calls Expect or Fail or panic -- always returns err. (Though the Before and AfterEach's in framework call Expect now). The functions in a _utils_ pkg may call Expect with the assumption they are **always** called in a ginkgo Describe, Context or It block. What do you think?

tests/framework/pvc_utils.go
jeffvance: It would only tell you if the file is present which it always will be since it's mounted. So nevermind...

tests/import_test.go
jeffvance: Fair enough for now.

tests/import_test.go
jeffvance: There is a balance between how long it takes to run a full e2e test suite vs meaningful error messages. I lean more toward the later because if a clear err msg helps QE or a developer find and fix a bug sooner I see that time saved as being very valuable. I guess this balance point is a personal choice and will likely fluctuate over the life of cdi.

tests/import_test.go
jeffvance: understood.

tests/framework/pvc.go
jeffvance: If no clear answer then leave it where it is now.

tests/import_test.go
jeffvance: Right. See my comment above and then decide how you want to do it.

jeffvance:
looking good. Nice job Alex!

tests/framework/framework.go
awels: The check is below, but you are right, I should only assign if err == nil

tests/framework/framework.go
```diff
@@ -75,41 +78,40 @@ func init() {
...
+	f.RestConfig = restConfig
+
+	if err != nil {
+		// Can't use Expect here due this being called outside of an It block, and Expect
+		// requires any calls to it to be inside an It block.
```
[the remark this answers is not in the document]
awels: :+1:

tests/framework/pvc.go
awels: :+1:

tests/utils/pvc.go
awels: We already have create and delete, why not phase. I will rename it.

tests/import_test.go
awels: Right its about run time vs ease of debugging... I am going to go with using the cascade from the NS delete for now. I might change my mind in the future.

tests/utils/pvc.go
awels: The same argument applies to the timeout as the poll value. So I will change it.

awels:
@jeffvance made some minor changes based on your comments, should be good to go now.

tests/import_test.go
```diff
@@ -0,0 +1,35 @@
...
+	waitForCDIImportTimeout = 30 * time.Second
+)
+
+var _ = Describe(testSuiteName, func() {
+	f, err := framework.NewFramework(namespacePrefix)
+	if err != nil {
+		Fail("Unable to create framework struct")
+	}
+
+	Describe("Creating standard PVC should not trigger importer", func() {
```
copejon: The title string here is pretty specific, which makes it seem like the existing `It()` is the only one that will ever be here.  If that's the case, it would be fine to extract the `It()` block to this scope and remove the nested `Describe()`.  And then just make the `It()` title a little more descriptive. (e.g. `It("should not perform CDI operations on PVCs without expected annotations", ... )`)

tests/framework/pvc.go
```diff
@@ -0,0 +1,47 @@
...
+	"time"
+
+	. "github.com/onsi/gomega"
+
+	k8sv1 "k8s.io/api/core/v1"
+
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+func (f *Framework) CreatePVCFromDefinition(def *k8sv1.PersistentVolumeClaim) (*k8sv1.PersistentVolumeClaim, error) {
```
copejon: Any desire to have a sister function that takes a `namespace string` parameter for multi namespace tests?

tests/utils/pod.go
```diff
@@ -0,0 +1,93 @@
...
+	return wait.PollImmediate(2*time.Second, PodDeleteTime, func() (bool, error) {
+		err := clientSet.CoreV1().Pods(namespace).Delete(pod.GetName(), &metav1.DeleteOptions{})
+		if err != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
+func newExecutorPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1.Pod {
```
copejon: `cmd` appears to be ignored

jeffvance:
@awels This pr is very close now. Pls address Jon's comments and let's try to get this merged today.

jeffvance:
@awels Here's our current plan for your pr: Jon will grab ENV vars from pr 352 and add them to his 353. This is because 353 changes filenames consolidates make targets. But we also want your ENV vars. So after these ENV vars are added to 353 we'll close (not merge) 352. 
This pr (354) will be merged after you've add (another) chance to address the last couple of comments and after 353 is merge. Still shooting for today. How does that sound?

tests/framework/pvc.go
awels: As you can see this delegates to a utility function, that utility function will take care of multi NS cases.

tests/utils/pod.go
awels: Correct, it is, we want this pod to be here so we can run our utility methods against it that can work with stdout/err/etc. The command simply needs to keep the pod alive long enough for us to connect to it.

awels:
@jeffvance Sounds good, #352 is fairly small so it should be easy to grab what you need and merge it into #353. I will address the comments, well its only one code change I think, the description of the describe in the actual test.

tests/import_test.go
awels: Maybe I am a little confused at what goes in which level. The goal is the Describe is for all the import tests, and the It is just one of the tests, in particular the negative nothing changes test. Is this better?
```go
	Describe("PVC import tests", func() {
		It("Creating standard PVC should not trigger importer, and the PVC should remain empty", func() {
			pvc, err := f.CreatePVCFromDefinition(utils.NewPVCDefinition("no-import", "1G", nil, nil))
			// Wait a while to see if CDI puts anything in the PVC.
			time.Sleep(waitForCDIImportTimeout)
			Expect(framework.VerifyPVCIsEmpty(f, pvc)).To(BeTrue())
			// Not deleting PVC as it will be removed with the NS removal.
		})
	})
```

copejon:
LGTM!

tests/import_test.go
```diff
@@ -0,0 +1,33 @@
...
+	f, err := framework.NewFramework(namespacePrefix)
+	if err != nil {
+		Fail("Unable to create framework struct")
+	}
+
+	It("Should not perform CDI operations on PVC without annotations", func() {
+		pvc, err := f.CreatePVCFromDefinition(utils.NewPVCDefinition("no-import", "1G", nil, nil))
+		Expect(err).ToNot(HaveOccurred())
+		// Wait a while to see if CDI puts anything in the PVC.
+		time.Sleep(waitForCDIImportTimeout)
```
aglitke: I think a better way to do this is to grep the import controller pod logs for messages that indicate the pvc was found by the controller but not picked up.  I think kubevirt has some functional tests like this that you could use as an example.

jeffvance: I agree @aglitke . Let's do this is a new pr.

PR #356, opened 2018-08-20
Sync test http server to cluster in make cluster-sync

awels:
- Added init/http docker images creation during sync
- Sync init/http image to cluster registry.
- Start http server service during sync.


awels:
@jeffvance @copejon @davidvossel please review. This PR syncs the http/s3 server into the cluster during a make sync, and then starts it. This will be a pre-requisite for running the e2e test.

awels:
I am going to have to test if the service which is created in the the kube-system namespace (should we do it somewhere else?) is exposed to other namespaces or not, if not we need to solve that for the tests.

awels:
Looks like I messed up some stuff, working on fixing the docker make target.

Makefile
```diff
@@ -111,5 +111,12 @@ functest:
 	./hack/build/functests.sh
 
 functest-image-host: WHAT=tools/cdi-func-test-file-host-init
-functest-image-host:  manifests build
+functest-image-host:  manifests docker
```
copejon: Why build all project binaries and images from scratch when we build the image server?

copejon: This also creates redundant work for line L115 `./hack/build/build-docker.sh "tools/cdi-func-test-file-host-init tools/cdi-func-test-file-host-http"`

hack/build/config.sh
```diff
@@ -23,7 +23,7 @@ CONTROLLER_MAIN="cmd/${CONTROLLER}"
 IMPORTER_MAIN="cmd/${IMPORTER}"
 CLONER_MAIN="cmd/${CLONER}"
 
-DOCKER_IMAGES="cmd/${CONTROLLER} cmd/${IMPORTER} cmd/${CLONER}"
+DOCKER_IMAGES="cmd/${CONTROLLER} cmd/${IMPORTER} cmd/${CLONER} tools/cdi-func-test-file-host-init tools/cdi-func-test-file-host-http"
```
copejon: Can you extract the `cdi-func-test-file-host-*` names into separate variables

Makefile
```diff
@@ -111,5 +111,12 @@ functest:
...
 	${DO} ./hack/build/build-cdi-func-test-file-host.sh && ./hack/build/build-docker.sh "tools/cdi-func-test-file-host-init tools/cdi-func-test-file-host-http"
+
+cluster-sync-functest-image-init: WHAT = tools/cdi-func-test-file-host-init
+cluster-sync-functest-image-init: functest-image-host
```
copejon: This isn't syncing anything to the cluster from what I can see.  The prerequisite should probably be `cluster-sync`.

Makefile
```diff
@@ -111,5 +111,12 @@ functest:
...
+
+cluster-sync-functest-image-http: WHAT = tools/cdi-func-test-file-host-http
+cluster-sync-functest-image-http: functest-image-host
```
copejon: ditto

hack/build/config.sh
awels: yes

Makefile
awels: Yeah, I was a little optimistic here, I had run some prerequisite manually and didn't do a make clean to make sure everything got build properly. I will push an update soon that should be much better.

cluster/sync.sh
```diff
@@ -27,4 +27,7 @@ for i in $(seq 1 ${KUBEVIRT_NUM_NODES}); do
     ./cluster/cli.sh ssh "node$(printf "%02d" ${i})" "echo \"${container}\" | xargs \-\-max-args=1 sudo docker pull"
 done
 
+# Install CDI
 ./cluster/kubectl.sh apply -f ./manifests/generated/cdi-controller.yaml
+# Start functional test HTTP server.
+./cluster/kubectl.sh apply -f ./manifests/generated/file-host.yaml
```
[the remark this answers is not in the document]
awels: You make a really good point, but it should be fairly easy to remedy, simply do a delete, and wait for the pods to disappear, then do the create. Or any number of other options to recreate the pods. I think the delete and wait is the simplest though.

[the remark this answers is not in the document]
davidvossel: take a look at how kubevirt cleans the cluster. They use the kubevirt.io take to remove all objects of a specific type. We already have the cdi.kubevirt.io label and can do the same thing. 

https://github.com/kubevirt/kubevirt/blob/master/cluster/clean.sh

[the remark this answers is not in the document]
awels: Nice, makes sense now, cluster-clean calls the cleaner, then in cluster-sync we call clean, and then the actual sync.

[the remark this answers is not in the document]
awels: @davidvossel stole the cleanup from kubevirt. We are now actually syncing.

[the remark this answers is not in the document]
davidvossel: @awels yep, looks good!

awels:
@copejon I believe the latest version addresses your concerns, please re-review.

davidvossel:
I tested this locally and it worked for me

jeffvance:
lgtm
@copejon you have a review approval request.

Issue #335, opened 2018-08-14
Intermittent CI failures in master


copejon:
CI is failing sporadically on the [CI master branch buildes](https://travis-ci.org/kubevirt/containerized-data-importer/builds/415627625).

The failures occur with tests involving `xz`:

Message: `could not format test data: Error writing to file: limit reached`

Complete test output:

```
Streaming Data Conversion
/go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:22
  when data is in a supported file format
  /go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:24
    should decompress xz [It]
    /go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:143
    *** skipping endpoint extension ".iso.xz" as exception
    /go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:184
------------------------------
••STEP: [BeforeEach] Creating temporary dir /tmp/.tmp-G2zfqjqX30Zj
STEP: Getting size of source file "/go/src/kubevirt.io/containerized-data-importer/test/images/tinyCore.iso"
INFO: size = 18874368
STEP: Converting source file to format: [.tar .xz]
STEP: [AfterEach] Cleaning up temporary dir /tmp/.tmp-G2zfqjqX30Zj
------------------------------
• Failure [3.099 seconds]
Streaming Data Conversion
/go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:22
  when data is in a supported file format
  /go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:24
    should unpack .tar.xz [It]
    /go/src/kubevirt.io/containerized-data-importer/test/functional/importer/datastream/datastream_test.go:143
    Error formatting test data.
    Expected error:
        <*errors.withStack | 0xc420300180>: {
            error: {
                cause: {
                    error: {
                        cause: {s: "limit reached"},
                        msg: "Error writing to file",
                    },
[...]

PR #360, opened 2018-08-22
issue 335 logging

jeffvance:
This pr adds some logging to test cases where we convert a file to _.tar.xz_. See issue #335 for all the details.

PR #361, opened 2018-08-22
Disable .tar.xz until issue 335 is resolved

copejon:
Unblocks PRs affected by #335 by disabling the the `.tar.xz` test.


jeffvance:
If we disable the test case that causes the problem how will be utilize the new logging added in pr #360?

jeffvance:
lgtm

mhenriks:
I wonder if this issue has a similar root cause when gemu-img was aborting in a mostly consistent way for one of my tests when running in travis

PR #363, opened 2018-08-22
make rbac tests compatible across kube versions update sanity functional test check

screeley44:
it appears that there are some implementation changes that have happened in Kube from version 1.10 for example which is used in kubevirtci `make cluster-up`.  The behavioral change is around checking rbac permissions using the `can-i` directive, in older versions if a resource+verb were not supported, the return would be no error and the result of `no` - in later versions (somewhere between 1.11 and latest 1.13) if a *No* case was tested it would return an error making our test invalid. I think switching test to only check the value and not check for error should handle all cases.


awels:
lgtm, maybe have @jeffvance and @copejon try it on their setup?

jeffvance:
@awels Jon and I cannot run `make cluster-up` in our dev env. Unfortunately, we need to rely on you and Scott for hands-on testing.
Otherwise, this LGTM. Thanks @screeley44 !

awels:
I verified on my end using a 1.11 kubectl against a 1.10 cluster.

PR #364, opened 2018-08-23
Fix default CDI namespace typo

awels:
- The default was set to kubesystem, instead of kube-system.


awels:
@copejon I think there was a typo in #353 this fixes it.

copejon:
@awels lgtm

PR #367, opened 2018-08-23
Run gofmt with simplifications

j-griffith:
Run gofmt with simplification                                                                                                                                                                                                
                                                                                                                                                                                                                             
This is the result of the new `make format` option and the results of                                                                                                                                                        
gofmt being run against the existing source files.


jeffvance:
lgtm

hack/build/format.sh
```diff
@@ -23,3 +23,4 @@ source "${script_dir}"/config.sh
 shfmt -i 4 -w ${CDI_DIR}/hack ${CLONER_MAIN}
 goimports -w -local kubevirt.io ${CDI_DIR}/cmd/ ${CDI_DIR}/pkg/ ${CDI_DIR}/test/
 (cd ${CDI_DIR} && go vet $(go list ./... | grep -v -E "vendor|pkg/client" | sort -u))
+gofmt -l -s -w $(SOURCE_DIRS)
```
copejon: `goimports` wraps `gofmt` and add import sorting.  I think the result here would be that L26 never catches format issues since L24 will have fixed them already.

j-griffith:
Going to close this and try and simplify the commits since it seems I've succesfully confused everyone :)

#370 and #371 supersede this.

PR #366, opened 2018-08-23
Add formatting and linter checks to Makefile

j-griffith:
This commit adds 2 new options to the Makefile, and extends the existing
`format` option:
1. checkformat
    This runs `gofmt -l -s` against all gofiles in the provided
    SOURCE_DIRS list (pkg, tests and tools), writing suggestions to
    stdout.

2. lint
    Runs `golint` against all gofiles in the provided SOURCE_DIRS list

3. Extends the existing format option
    In addition to running `go vet` and `go imports` via the hack/format.sh script, we
    add `gofmt -w -s` against the go files in SOURCE_DIRS trying to
    automatically update and write out simplifications to the files as
    necessary.

PR #368, opened 2018-08-23
Add gofmt checks to travis

j-griffith:
Add formatting check to travis                                                                                                                                                                            
                                                                                                                                                                                                                             
This just adds the `make checkformat` test to .travis.yml file to                                                                                                                                                        
include these as part of the build.  

We'll get all the lint errors cleaned 
up and pushed then add the `make lint`
as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,9 @@
 		format \
 		manifests \
 		goveralls \
-		release-description
+		release-description \
+		checkformat \
+		lint
 
 DOCKER=1
 ifeq (${DOCKER}, 1)
@@ -29,6 +31,7 @@ DO=./hack/build/in-docker.sh
 else
 DO=eval
 endif
+SOURCE_DIRS = pkg tests tools
 
 all: docker
 
@@ -97,6 +100,13 @@ manifests:
 
 goveralls:
 	${DO} "TRAVIS_JOB_ID=${TRAVIS_JOB_ID} TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST} TRAVIS_BRANCH=${TRAVIS_BRANCH} ./hack/build/goveralls.sh"
+
+checkformat:
+	## Just run gofmt without attempting to update files
+	@gofmt -l -s $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
+lint:
+	## Run golint on source files
+	@$(foreach dir,$(SOURCE_DIRS),echo "loop dir: $(dir)\n" && golint "$(dir)/...";)
 
 release-description:
 	./hack/build/release-description.sh ${RELREF} ${PREREF}
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ install:
   - make manifests
 
 before_script:
+- make goveralls
+- make checkformat
 - curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v$K8S_VER/bin/linux/amd64/kubectl
   && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
 - curl -Lo minikube https://storage.googleapis.com/minikube/releases/v$MK_VER/minikube-linux-amd64
@@ -47,7 +49,6 @@ script:
 - "sed \"s#cdi.kubevirt.io/storage.import.endpoint:.*#cdi.kubevirt.io/storage.import.endpoint: \"$SRC#\"\" manifests/example/golden-pvc.yaml | kubectl apply -f -"
 - k_wait_all_running pods
 - kubectl get pods --all-namespaces
-- make goveralls
 
 before_deploy:
 - docker login -u="$DOCKER_USER" -p="$DOCKER_PASS"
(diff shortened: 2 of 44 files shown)


j-griffith:
Cool, the `make lint` is expected to fail here; now to figure out if/how we want to introduce that check.  I'd propose for now updating this commit to comment out that check, and after I finish cleaning all the lint errors then remove the comment.

Makefile
```diff
@@ -98,6 +101,13 @@ manifests:
 goveralls:
 	${DO} "TRAVIS_JOB_ID=${TRAVIS_JOB_ID} TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST} TRAVIS_BRANCH=${TRAVIS_BRANCH} ./hack/build/goveralls.sh"
 
+checkformat:
+	## Just run gofmt without attempting to update files
+	@gofmt -l -s $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
```
copejon: I'd like to see this follow the design of the rest of the Makefile - that is having the make recipe all a script.  There's a few benefits to this
- config values like SOURCE_DIRS can be written in hack/build/config.sh||common.sh
- the scripts can be run by default in the build container which houses all the build deps

Makefile
```diff
@@ -98,6 +101,13 @@ manifests:
...
+lint:
+	## Run golint on source files
+	@$(foreach dir,$(SOURCE_DIRS),echo "loop dir: $(dir)\n" && golint "$(dir)/...";)
```
copejon: ditto the above comment.  Also, if you think that it'd be preferred to run all the checks at once, a single script that combines them all would be fine.

Makefile
j-griffith: Sure, I can change these up to live in the hack scripts, good point about the build container.

Makefile
j-griffith: I'm open to however folks would like to see this.  I was following the goveralls pattern (ie using the makefile directive) but moving the lint and format checks into the format script makes sense if everybody's good with that.

I would leave the actual format option in the Makefile to run "gofmt -s -w" though if that makes sense?  Maybe it needs a more distinctive name like "make runformatter" ?

copejon: Yup, that makes sense to me.

In case you haven't run into it yet, there's a generated file that fails `go vet`.  An issue was raised in kube about it but nothing has come of it yet
```
pkg/client/clientset/versioned/fake/clientset_generated.go:56: literal copies lock value from fakePtr: kubevirt.io/containerized-data-importer/vendor/k8s.io/client-go/testing.Fake
```

j-griffith: Ahh, I haven't but thanks for the heads up.  I did notice that there's something wonky with setting the CDI_DIR when running `make format`, it fails on my local machine because I'm not building in docker (dir is always set to /go/src/kubevirt.io/....).  Will see if I can get that sorted.

copejon: That's odd, it shouldn't be acting like it.  The idea is that CDI_DIR always reflects repo root, regardless of the env / parent dir path.

Makefile
jeffvance: @j-griffith See my strikeout comment in #366. This would be the 1st time that CDI CI alters the content of a pr. We've not wanted to do that in the past. Are there examples where this is done in other repos?

copejon: > 1. Just check that gofmt passes
That's actually in the hack/build/format script like you suggest

 +1 this is a good addition for CI - it's been needed but back burnered for a long time

> 2. Actually run a gofmt -s -w and update all src files automagically
That one seemed like it was ok to call directly from Make, but I can put it in another
build script if it makes more sense for the pattern?

I think this is what `format.sh` accomplishes - `goimports` wraps `gofmt` and adds import sorting.  Given how many scripts we have and continue to add, `shfmt` seems like a good idea as well.  Right now `format.sh` also calls `go vet` but we could extract that into a `verify-format.sh` script with other linting calls.

jeffvance: and `shfmt` has been added to the builder container? It's not included in my rhel dev env

j-griffith: @jeffvance Yeah, that's exactly why I was doing it the way I had it here; CDI/CI systems modifying code is not something I think is a good idea at all (and not exactly sure how you'd do that sanely anyway).  I think there might be some confusion on my intent here.  For gofmt I want two things:
1. A hook to make travis test it and fail the PR if it's not formatted (still up to the owner to fix it or to merge anyway)

2. Provide a simple, easy automated wrapper around gofmt so a comitter could (in their own branch locally) run `gofmt -s -w` without thinking about it and then see all the modified files, review them and commit the changes if they desire.

I went ahead and looked at putting that into the format.sh build script, which "works" for case number 2, but I'm not sure what it does in the CI context.  If we want to skip number 2 altogether I'm perfectly fine with that and just adding the hook in the makefile and travis to check formatting and linting.  Thoughts?

jeffvance: @j-griffith My misunderstanding. I totally agree w/ your above comment.

j-griffith:
Going to close this and try and simplify the commits since it seems I've succesfully confused everyone :)

#370 and #371 supersede this.

Issue #374, opened 2018-08-27
Cloner container creation has hard coded script path


awels:
The clone controller has a hard coded script path for calling the script when the clone source/target container is instantiated.

In release v1.1.1 its hard coded to /tmp/script.sh
In master its hard coded to /tmp/cloner_startup.sh

We should use the ENTRYPOINT for the container to specify the script path instead of hard coding it in the controller.

aglitke:
In addition, the script should be installed in /usr/bin, not /tmp

nellyc:
fixed

PR #375, opened 2018-08-27
cloner image uses ENTRYPOINT instead of hard coded path

awels:
- Updated controller to use ENTRYPOINT (no cmd) instead of /bin/sh
- Updated Dockerfile to set path to /usr/bin/cloner_startup.sh instead of /tmp
- Updated Dockerfile to use ENTRYPOINT.

Fixes issue #374


jeffvance:
lgtm

PR #377, opened 2018-08-27
Fix WaitForPersistentVolumeClaimPhase returning error unconditionally

copejon:
Fix function returning an error regardless of the output from the `wait` call.

### DIFF
--- a/tests/utils/pvc.go
+++ b/tests/utils/pvc.go
@@ -76,12 +76,15 @@ func NewPVCDefinition(pvcName string, size string, annotations, labels map[strin
 
 // Wait for the PVC to be in a particular phase (Pending, Bound, or Lost)
 func WaitForPersistentVolumeClaimPhase(clientSet *kubernetes.Clientset, namespace string, phase k8sv1.PersistentVolumeClaimPhase, pvcName string) error {
-	wait.PollImmediate(pvcPollInterval, PVCPhaseTime, func() (bool, error) {
+	err := wait.PollImmediate(pvcPollInterval, PVCPhaseTime, func() (bool, error) {
 		pvc, err := clientSet.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


tests/utils/pvc.go
```diff
@@ -76,12 +76,15 @@ func NewPVCDefinition(pvcName string, size string, annotations, labels map[strin
...
+	err := wait.PollImmediate(pvcPollInterval, PVCPhaseTime, func() (bool, error) {
 		pvc, err := clientSet.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
 		if err != nil || pvc.Status.Phase != phase {
 			return false, err
 		}
 		return true, nil
 	})
-	return fmt.Errorf("PersistentVolumeClaim %s not in phase %s within %v", pvcName, phase, PVCPhaseTime)
+	if err != nil{
+		return fmt.Errorf("PersistentVolumeClaim %s not in phase %s within %v", pvcName, phase, PVCPhaseTime)
```
jeffvance: what about:
```
func WaitForPersistentVolumeClaimPhase(clientSet *kubernetes.Clientset, namespace string, phase k8sv1.PersistentVolumeClaimPhase, pvcName string) error {
	return wait.PollImmediate(pvcPollInterval, PVCPhaseTime, func() (bool, error) {
		pvc, err := clientSet.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
		if err != nil || pvc.Status.Phase != phase {
			return false, err
		}
		return true, nil
	})
}
```

jeffvance: Although, a downside to my suggestion is that the caller doesn't see the nice err msg stating the pvc was not in the expected phase after X mins...

copejon: I'd considered that as well, but opted for the verbose error message.

jeffvance: ack. Ok as is!

jeffvance:
closed by #378

PR #383, opened 2018-08-29
improve hack/readme for TEST_ARGS

jeffvance:
Re-worded the description on using `TEST_ARGS`.
This may not be how we intended to use `TEST_ARGS` for functional tests, but it explains the way it works currently.


jeffvance:
ci add to whitelist

PR #394, opened 2018-08-30
Fixes the broken make target for functest http container

copejon:
Removes the `build` target for functest-file-host-http.  There is no code to compile and the target dir does not exist, causing a `file not found` error.

Adjusts `docker-functest-image-http` target to remove `build-functest-image-http`, call `build-docker.sh` instead.


awels:
Looks good to me, but waiting on merges until I get the CI one merged.

PR #399, opened 2018-08-31
`NewFrameworkOrDie`

jeffvance:
Replace the common functional test pattern of:
```
f, err := framework.NewFramework("foo", framework.Config{})
if err != nil {
     Fail("...")
}
```
with
```
f := framework.NewFrameworkOrDie("foo")
```
Note that the Config parameter is optional now.

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -96,6 +96,15 @@ func init() {
 	master = flag.String("master", "", "master url:port")
 }
 
+// NewFrameworkOrDie calls NewFramework and handles errors by calling Fail.
+func NewFrameworkOrDie(prefix string, config Config) *Framework {
+	f, err := NewFramework(prefix, config)
+	if err != nil {
+		Fail(fmt.Sprintf("failed to create test framework: %v", err))
+	}
+	return f
+}
+
 // NewFramework makes a new framework and sets up the global BeforeEach/AfterEach's.
 // Test run-time flags are parsed and added to the Framework struct.
 func NewFramework(prefix string, config Config) (*Framework, error) {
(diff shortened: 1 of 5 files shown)


jeffvance:
This pr fails the cloner functional tests but so does the master branch right now.

tests/framework/framework.go
```diff
@@ -96,6 +96,23 @@ func init() {
 	master = flag.String("master", "", "master url:port")
 }
 
+// NewFrameworkOrDie calls NewFramework and handles errors by calling Fail. Config is optional, but
+// if passed there can only be one.
+func NewFrameworkOrDie(prefix string, config ...Config) *Framework {
```
screeley44: Why is config variadic?

tests/framework/framework.go
```diff
@@ -96,6 +96,23 @@ func init() {
...
+	if len(config) > 1 {
+		Fail("only 1 config can be passed to NewFrameworkOrDie")
+	}
+	cfg := Config{}
+	if len(config) == 1 {
```
screeley44: what if config isn't passed in at all?

tests/framework/framework.go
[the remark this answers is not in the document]
jeffvance: I haven't tried that. What would happen if a list of Configs was passed? I wanted to protect against mis-calling `NewFrameworkOrDie`.

jeffvance: So that it becomes an _optional_ parameter.

tests/framework/framework.go
jeffvance: Then len(config) == 0 and cfg == `Config{}` an empty Config struct.

tests/framework/framework.go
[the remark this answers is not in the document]
jeffvance: I could do this:
```
func NewFrameworkOrDie(prefix string, config ...Config) *Framework {
	cfg := Config{}
	if len(config) > 0 {
		cfg = config[0]
	}
	f, err := NewFramework(prefix, cfg)
	if err != nil {
		Fail(fmt.Sprintf("failed to create test framework with config %+v: %v", cfg, err))
	}
	return f
}
```

PR #401, opened 2018-08-31
Client creation in NewFramework

awels:
- Moved client creation into NewFramework instead of BeforeEach
  block.


awels:
ci please test

awels:
ci test please

tests/framework/framework.go
```diff
@@ -129,27 +129,28 @@ func NewFramework(prefix string, config Config) (*Framework, error) {
...
 
+	// clients
+	kcs, err := f.GetKubeClient()
+	if err != nil {
+		err = errors.Wrap(err, "ERROR, unable to create K8SClient")
+	} else {
+		f.K8sClient = kcs
+	}
+
+	cs, err := f.GetCdiClient()
```
jeffvance: If you fail to create the kubeClient above, do you still want to create the cdi client here? I expected to see a pattern like this:
```
	kcs, err := f.GetKubeClient()
 	if err != nil {
 		return nil, errors.Wrap(err, "ERROR, unable to create K8SClient")
 	f.K8sClient = kcs
...
```

tests/framework/framework.go
```diff
@@ -129,27 +129,28 @@ func NewFramework(prefix string, config Config) (*Framework, error) {
...
+	if err != nil {
```
jeffvance: ditto

tests/framework/framework.go
awels: You are right, I will push follow up soon.

PR #402, opened 2018-08-31
Fix find pod by prefix

awels:
- Implemented missing LabelSelector code.
- Fixed bug where err was not nil in inner function.

### DIFF
--- a/tests/utils/pod.go
+++ b/tests/utils/pod.go
@@ -84,34 +84,32 @@ func NewPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1
 // Finds the first pod which has the passed in prefix. Returns error if multiple pods with the same prefix are found.
 func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSelector string) (*k8sv1.Pod, error) {
 	var result k8sv1.Pod
-	var multiplePods bool
+	var foundPod bool
 	err := wait.PollImmediate(2*time.Second, PodCreateTime, func() (bool, error) {
-		multiplePods = false
-		podList, err := clientSet.CoreV1().Pods(namespace).List(metav1.ListOptions{})
+		podList, err := clientSet.CoreV1().Pods(namespace).List(metav1.ListOptions{
+			LabelSelector: labelSelector,
+		})
 		if err == nil {
-			var foundPod bool
 			for _, pod := range podList.Items {
 				if strings.HasPrefix(pod.Name, prefix) {
 					if !foundPod {
 						foundPod = true
 						result = pod
 					} else {
-						fmt.Fprintf(GinkgoWriter, "INFO: First pod name %s\n", result.Name)
-						fmt.Fprintf(GinkgoWriter, "INFO: Second pod name %s\n", pod.Name)
-						multiplePods = true
+						fmt.Fprintf(GinkgoWriter, "INFO: First pod name %s in namespace %s\n", result.Name, result.Namespace)
+						fmt.Fprintf(GinkgoWriter, "INFO: Second pod name %s in namespace %s\n", pod.Name, pod.Namespace)
+						return true, fmt.Errorf("Multiple pods starting with prefix %q in namespace %q", prefix, namespace)
 					}
 				}
 			}
 			if foundPod {
 				return true, nil
 			}
-			return false, fmt.Errorf("Unable to find pod starting with prefix %s", prefix)
-		} else {
-			return false, err
 		}
+		return false, nil
 	})
-	if multiplePods {
-		return nil, fmt.Errorf("Multiple pods starting with prefix %s", prefix)
+	if !foundPod {
+		return nil, fmt.Errorf("Unable to find pod starting with prefix %s", prefix)
 	}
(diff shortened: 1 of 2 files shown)


tests/utils/pod.go
```diff
@@ -107,7 +109,7 @@ func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSe
 			}
 			return false, fmt.Errorf("Unable to find pod starting with prefix %s", prefix)
```
jeffvance: This line `return false, fmt.Errorf("Unable to find pod starting with prefix %s", prefix)`, inside the Poll, is saying both _continue to poll_ (done==false), AND _stop the poll_ (err != nil). The err != nil take precedence so the Poll will stop. Is this what you want?

jeffvance:
Looks good. One comment re Poll and returning err.

tests/utils/pod.go
awels: You are right its probably not what I want, but I do want to indicate that I didn't find anything, but I should do it outside of the poll.

tests/utils/pod.go
```diff
@@ -84,12 +84,13 @@ func NewPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1
 // Finds the first pod which has the passed in prefix. Returns error if multiple pods with the same prefix are found.
 func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSelector string) (*k8sv1.Pod, error) {
 	var result k8sv1.Pod
-	var multiplePods bool
+	var multiplePods, foundPod bool
 	err := wait.PollImmediate(2*time.Second, PodCreateTime, func() (bool, error) {
 		multiplePods = false
-		podList, err := clientSet.CoreV1().Pods(namespace).List(metav1.ListOptions{})
+		podList, err := clientSet.CoreV1().Pods(namespace).List(metav1.ListOptions{
```
jeffvance: Line above: `multiplePods = false`
You don't need to do this. False is its default/"zero" value. Once it becomes true you want to exit the Poll. IOW, it doesn't matter if you have 2 pods w/ same prefix or 100, once there's >1 exit the Poll.

tests/utils/pod.go
```diff
@@ -105,11 +106,12 @@ func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSe
 			if foundPod {
 				return true, nil
 			}
-			return false, fmt.Errorf("Unable to find pod starting with prefix %s", prefix)
-		} else {
-			return false, err
 		}
+		return false, nil
```
jeffvance: L100 above: it might help with debugging to include the ns in the Fprintfs above.

tests/utils/pod.go
jeffvance: Actually, I don't think you need `multiplePods`, see below

tests/utils/pod.go
awels: Yes good plan, I will add that.

tests/utils/pod.go
awels: I am missing something, I need multiplePods to let function outside the poll loop know I found multiples. In each poll loop I loop over all found pods and thus I can find multiples.

tests/utils/pod.go
```diff
@@ -105,11 +106,12 @@ func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSe
...
 	})
+	if !foundPod {
+		return nil, fmt.Errorf("Unable to find pod starting with prefix %s", prefix)
+	}
```
jeffvance: See [PollImmediate](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L244)
Potential refactor, what do you think?
```
func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSelector string) (*k8sv1.Pod, error) {
	var result k8sv1.Pod
	var foundPod bool
		podList, err := clientSet.CoreV1().Pods(namespace).List(metav1.ListOptions{
			LabelSelector: labelSelector,
		})
		if err != nil {
                        return false, nil // continue Poll
                }
			if strings.HasPrefix(pod.Name, prefix) {
				if !foundPod {
					foundPod = true
					result = pod
				} else {
					fmt.Fprintf(GinkgoWriter, "INFO: First pod name %s\n", result.Name)
					fmt.Fprintf(GinkgoWriter, "INFO: Second pod name %s\n", pod.Name)
                                        return true, fmt.Errorf("Multiple pods starting with prefix %q in namespace %q", prefix, namespace)
				}
			}
		}
                // if we get here all pods compared with at most 1 matching pod
		return true, nil
	})
	if !foundPod {
		return nil, fmt.Errorf("Unable to find pod in namespace %q starting with prefix %q", namespace, prefix)
	}
	return &result, err
}
```

tests/utils/pod.go
jeffvance: That happens with the refactor because you're inner func is returning _true_ to stop the Poll and _err_ which is then returned by `PollImmediate`

awels: Yes I got it now, tricky little details.

PR #404, opened 2018-08-31
Fix golint errors and add golint check on CDI Packages

j-griffith:
We've agreed to use gofmt and golint, gofmt has been added; now it's time for golint.  There's a lot of little lint errors with the existing code base so rather than try and dump one gigantic change set, this series of commits will work from the bottom (module layer) up.  We'll start which just focuses on fixing each individual package and then work our way up to some more far reaching changes.

Most of what's here is just adding comments to exported functions and constants, there is some other clean up around how exported/un exported consts were being used and also some other minor things.

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -43,19 +43,23 @@ import (
 	cdischeme "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned/scheme"
 	informers "kubevirt.io/containerized-data-importer/pkg/client/informers/externalversions/datavolumecontroller/v1alpha1"
 	listers "kubevirt.io/containerized-data-importer/pkg/client/listers/datavolumecontroller/v1alpha1"
-	. "kubevirt.io/containerized-data-importer/pkg/common"
 	expectations "kubevirt.io/containerized-data-importer/pkg/expectations"
 )
 
 const controllerAgentName = "datavolume-controller"
 
 const (
-	SuccessSynced         = "Synced"
-	ErrResourceExists     = "ErrResourceExists"
+	// SuccessSynced provides a const to represent a Synced status
+	SuccessSynced = "Synced"
+	// ErrResourceExists provides a const to indicate a resource exists error
+	ErrResourceExists = "ErrResourceExists"
+	// MessageResourceExists provides a const to form a resource exists error message
 	MessageResourceExists = "Resource %q already exists and is not managed by DataVolume"
+	// MessageResourceSynced provides a const to standardize a Resource Synced message
 	MessageResourceSynced = "DataVolume synced successfully"
 )
 
+// DataVolumeController represents the CDI Data Volume Controller
 type DataVolumeController struct {
 	// kubeclientset is a standard kubernetes clientset
 	kubeclientset kubernetes.Interface
@@ -74,6 +78,8 @@ type DataVolumeController struct {
 	pvcExpectations *expectations.UIDTrackingControllerExpectations
 }
 
+// NewDataVolumeController sets up a Data Volume Controller, and return a pointer to
+// the newly created Controller
 func NewDataVolumeController(
 	kubeclientset kubernetes.Interface,
 	cdiClientSet clientset.Interface,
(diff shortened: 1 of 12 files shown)


pkg/lib/size/size.go
```diff
@@ -1,5 +1,7 @@
 package size
 
+// should this pkg just be dropped and made an exported function of the datastream pkg?
```
j-griffith: Seems like we should move this into an exported function of the importer package, rather than having a full package (that imports the importer package anyway) to get the size of the endpoint.  Maybe there are intentions of adding more to this lib package?

pkg/controller/datavolume-controller.go
```diff
@@ -84,9 +90,9 @@ func NewDataVolumeController(
 	// Add datavolume-controller types to the default Kubernetes Scheme so Events can be
 	// logged for datavolume-controller types.
 	cdischeme.AddToScheme(scheme.Scheme)
-	glog.V(Vdebug).Info("Creating event broadcaster")
+	glog.V(3).Info("Creating event broadcaster")
```
copejon: What's the distinction between using `cdi.vDebug` above and hard coded ints in this file?

pkg/controller/import-controller.go
```diff
@@ -124,9 +132,9 @@ func (c *ImportController) handlePodObject(obj interface{}, verb string) {
 			runtime.HandleError(errors.Errorf("error decoding object tombstone, invalid type"))
 			return
 		}
-		glog.V(Vdebug).Infof("Recovered deleted object '%s' from tombstone", object.GetName())
+		glog.V(3).Infof("Recovered deleted object '%s' from tombstone", object.GetName())
```
copejon: Ditto hardcoded log levels.

pkg/controller/import-controller.go
```diff
@@ -16,21 +16,27 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
-	. "kubevirt.io/containerized-data-importer/pkg/common"
+	//. "kubevirt.io/containerized-data-importer/pkg/common"
```
awels: This can be removed?

pkg/lib/size/size.go
copejon: There was some pressure a while ago to provide an exportable package that CNV would import to determine the size of a remote file.  The intent being that front end automation would take be given an endpoint by a user, determine the size of the remote image, and generate a PVC just large enough to hold it.

Placing `size` in a standalone lib made sense then form an organizational standpoint, but I don't think it's being used in that way now.

awels:
Looks pretty good to me, had same question as Jon, at first you do cdi.VAdmin/etc, but then later you switch to 1/2/3 for the debug levels. Do we want to be consistent (I know its a pain)? Also any particular reason you changed dataStream to DataStream? It doesn't bother me, just wondering.

pkg/controller/datavolume-controller.go
j-griffith: It's not a major thing, but I figured we didn't really need a constant for 1, 2 and 3; and could manage to just use glog in a more normal manner so to speak.  The main motivation is I was working on elimintating pkg/common/common.go; there's one hold out left there but everything else sort of just "falls out" once you start poking at it.  This was the only thing really that needed to be in common by itself; seemed like it was more straight forward to just use glog normally like in K8s.

pkg/controller/import-controller.go
j-griffith: Oops, yeah thanks!

awels:
I don't have strong feelings either way, I just noticed we are mixing either Vuser, or 1 depending on the file. I just would like to be consistent one way or the other.

j-griffith:
@awels OH!  Yeah, so I'm doing these pkg by pkg, the goal being that "soon" they're all replaced.  I can spin another patch to just do it all at once to avoid confusion as things go along.

j-griffith:
I pushed up 408 to tackle the log levels all at once.

j-griffith:
Breaking these out into individual PR's to make things a bit more digestible and better organized.  See PRs #414 through #416

PR #407, opened 2018-08-31
Set file host nginx servernames

copejon:
Moves `servername` directive to appropriate config blocks.


jeffvance:
lgtm, will merge after ci

PR #408, opened 2018-08-31
Remove log-level constants

j-griffith:
This change just removes the log level constants that were set in
pkg/common/common.go and falls back to just using the int arguments (1,
2 or 3) to glog.  While the constants are handy to limit the choices for example it also introduces a pkg dependency on common.go.  It seems just as easy to type 1, 2 or 3 and if we want to enforce the use of levels inclusive of a certain range we can do that through lint checks or review criteria.  I'm not advocating we do that, just mentioning that if that's a restriction we want we can certainly make it work.

Now instead of issuing a log call with the constant:
  `glog.V(common.Vuser)`

We just provide the argument as an int:
  `glog.V(1)`

I started doing this as part of the effort to fix linting, but that's
going package by package and it was rightly pointed out that the result
is a period of inconsistency and confusion.  So this patch just jumps
ahead and takes care of the low hanging fruit in one shot.

Related PR #404 #410

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -59,8 +59,9 @@ func init() {
 		pullPolicy = pp
 	}
 
-	// get the verbose level so it can be passed to the importer pod
-	defVerbose := fmt.Sprintf("%d", DEFAULT_VERBOSE) // note flag values are strings
+	// NOTE we used to have a constant here and we're now just passing in the level directly
+	// that should be fine since it was a constant and not a mutable variable
+	defVerbose := fmt.Sprintf("%d", 1) // note flag values are strings
 	verbose = defVerbose
 	// visit actual flags passed in and if passed check -v and set verbose
 	flag.Visit(func(f *flag.Flag) {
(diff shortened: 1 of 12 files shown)


jeffvance:
is there any connection/dep between this pr and 410?

jeffvance:
The pr description does not mention _why_ you want to change log constants to numbers? Is it mostly because that's a common approach?
One thing I've noticed with log verbosity numbers is you're not sure which number to use for some of the logs. Eg is this a 4 or a 5 level? You then search code and see little consistency. Our thoughts on   having only 3 verbose levels is that it would be easier to decide which level to use. Of course, this perspective does not necessarily endorse the use of constants, but constants help keep the number of choices smaller...

j-griffith:
@jeffvance I added a few lines to the commit message.  Yes, this is indirectly related to the golint changes; it's not really a requirement as such but when I started going through the exported constants it turned out that most were ONLY used by the Controller package.  So then I started looking a bit more, and it turns out that the bulk of the constants are limited to a single package, or they're used in a secondary nature by packages that already include the primary package they're used in.  

What basically happened is that controller and image packages were pretty much able to hold most of the constants themeselves and there was no real need to have them exported.  Here's an example of what a const.go file in the controller package started to look like:  https://gist.github.com/j-griffith/a65e80a2bc008258efe4b2399f0cd636

Maybe it would help if next week we set up some time and I can talk through some of this and where it's going?  If folks decide they don't agree or that I'm going the wrong direction (I certainly don't have the background and experience in the project) it's totally fair.

Mostly I'm trying to add some value reworking the code a bit to hopefully in the end make it a bit easier to follow, simpler to maintain and make bringing on new users with go experience a bit easier by just following some of the more standardized golang practices.  That all started with just gofmt and golint.

jeffvance:
>  I'm trying to add some value reworking the code a bit to hopefully in the end make it a bit easier to follow, simpler to maintain and make bringing on new users with go experience a bit easier by just following some of the more standardized golang practices.

This of course is a great goal and one that @copejon and I followed in of our initial design, coding and pr reviews.
In my use of `gofmt`, `golint` and `goimports` I have not gotten errors when constants were used for the verbosity levels. Are you seeing errors here? That said, I think we should schedule time to discuss the history and some of the decisions that shaped CDI. Jon and I will be leaving the project shortly and it may benefit those taking over to know how it got to where it is today.

jeffvance:
Must rebase. Can't use github's merge

PR #410, opened 2018-08-31
Image pkg lint

j-griffith:
Just do the bare minimum to run a clean golint on the image package and add it to our lint-test.


jeffvance:
Not related to this pr, but it seems like after the STDCI pr was merged we now fail the 3 new CI tests. @awels can you ptal?
cc. @aglitke

hack/build/run-lint-checks.sh
```diff
@@ -27,4 +27,27 @@ if [[ ${out} ]]; then
     echo "${out}"
     ec=1
 fi
+# Once we get everything fixed up we'll just have a single lint check
+# in the meantime add pkgs as they're ready
+# controller
+out="$(golint pkg/controller/...)"
```
jeffvance: what about a loop here to invoke `golint` and the various dirs?

pkg/controller/clone-controller.go
```diff
@@ -16,21 +16,26 @@ import (
...
 	"k8s.io/client-go/util/workqueue"
 
-	. "kubevirt.io/containerized-data-importer/pkg/common"
+	cdi "kubevirt.io/containerized-data-importer/pkg/common"
 	expectations "kubevirt.io/containerized-data-importer/pkg/expectations"
 )
 
 const (
-	// pvc annotations
+	//AnnCloneRequest sets our expected annotation for a CloneRequest
```
jeffvance: It might be useful to define all annotations in _common_? That could make it easier to find the anno value, though IDE's pretty much mitigate that advantage...

pkg/controller/clone-controller.go
```diff
@@ -16,21 +16,26 @@ import (
...
 	AnnCloneRequest = "k8s.io/CloneRequest"
-	AnnCloneOf      = "k8s.io/CloneOf"
-	// cloner pods annotations
-	AnnCloningCreatedBy   = "cdi.kubevirt.io/storage.cloningCreatedByController"
-	AnnClonePodPhase      = "cdi.kubevirt.io/storage.clone.pod.phase"
-	CloneUniqueID         = "cdi.kubevirt.io/storage.clone.cloneUniqeId"
+	// AnnCloneOf sets our expected annotation to inidcate where a clone originated from
```
jeffvance: +1 Nice to see the additional of comments

pkg/controller/datavolume-controller.go
```diff
@@ -84,9 +90,9 @@ func NewDataVolumeController(
 	// Add datavolume-controller types to the default Kubernetes Scheme so Events can be
 	// logged for datavolume-controller types.
 	cdischeme.AddToScheme(scheme.Scheme)
-	glog.V(Vdebug).Info("Creating event broadcaster")
+	glog.V(3).Info("Creating event broadcaster")
```
jeffvance: Why not using the _common_ const here?

pkg/controller/datavolume-controller.go
```diff
@@ -84,9 +90,9 @@ func NewDataVolumeController(
...
 	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartLogging(glog.V(Vadmin).Infof)
+	eventBroadcaster.StartLogging(glog.V(2).Infof)
```
jeffvance: ditto.

pkg/controller/datavolume-controller.go
```diff
@@ -102,7 +108,7 @@ func NewDataVolumeController(
 		pvcExpectations:   expectations.NewUIDTrackingControllerExpectations(expectations.NewControllerExpectations()),
 	}
 
-	glog.V(Vadmin).Info("Setting up event handlers")
+	glog.V(2).Info("Setting up event handlers")
```
jeffvance: and again.

pkg/controller/util_test.go
```diff
@@ -70,9 +71,9 @@ func TestCloneController_pvcFromKey(t *testing.T) {
 
 	//create staging pvc and pods
 	pvcWithCloneRequestAnno := createPvc("target-pvc", "target-ns", map[string]string{AnnCloneRequest: "source-ns/golden-pvc"}, nil)
-	pvcUid := string(pvcWithCloneRequestAnno.GetUID())
-	sourcePod := createSourcePod(pvcWithCloneRequestAnno, DataVolName, pvcUid)
-	targetPod := createTargetPod(pvcWithCloneRequestAnno, DataVolName, pvcUid, sourcePod.Namespace)
+	id := string(pvcWithCloneRequestAnno.GetUID())
```
jeffvance: not sure of the motivation to change so many occurrences of _pvcUid_ to _id_ since it is the id of the pvc? But it's a nit comment.

pkg/image/filefmt.go
```diff
@@ -67,10 +66,12 @@ func CopyKnownHdrs() Headers {
 	return m
 }
 
+// Match performs a check to see if the provided byte array matches the bytes in our header data
 func (h Header) Match(b []byte) bool {
 	return bytes.Equal(b[h.mgOffset:h.mgOffset+len(h.magicNumber)], h.magicNumber)
 }
 
+// Size calculates the size of the provided byte array as an int64 and returns it or an error
```
jeffvance: Not really accurate. `Size` extracts from the receiver the byte size of the original file represented by this header's metadata.

jeffvance:
A suggestion: you might get quicker and more focused reviews by splitting up prs like this into a couple of smaller, narrower pr. Eg one pr for public vs private types, fields, funcs etc. A separate pr to change log levels to numbers. A separate for the lint enhancement. And, the title of this pr is not reflective of most of its content...

hack/build/run-lint-checks.sh
j-griffith: Sure, but this is just temporary, eventually we just run against all the src dirs in one shot.

pkg/controller/clone-controller.go
j-griffith: Yeah, agree for sure!  I'm just focused on doing the bare minimum necessary to make golint pass for now.  Then I have some changes that will do things like grouping and maybe rename things a bit.  Still kinda learning some of this code base so I'll look forward to help from you and others when I get to that point.

pkg/controller/datavolume-controller.go
j-griffith: Yeah, surprisingly this seems to be the most controversial thing :)  So personally I'm used to just using glog and providing the arguments explicitly (1-5), k8s, kubevirt etc.  Just thought it would eliminate an import and was straight forward.  I also thought it was trivial, but it seems that maybe that's not the case.

pkg/controller/util_test.go
j-griffith: Because it fails golint :)

pkg/image/filefmt.go
j-griffith: Thanks, like I said; there wasn't a road map or any comments so I pieced these together best I could; I'll see if I can fix that up.

pkg/importer/dataStream.go
```diff
@@ -37,59 +38,59 @@ type DataStreamInterface interface {
...
-var _ DataStreamInterface = &dataStream{}
+var _ DataStreamInterface = &DataStream{}
 
-// implements the ReadCloser interface
-type dataStream struct {
-	Url         *url.URL
-	Readers     []Reader
+// DataStream implements the ReadCloser interface
+type DataStream struct {
+	url         *url.URL
```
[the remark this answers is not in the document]
j-griffith: Well... DataStream was not exported, that's something I added/changed because we return it in multiple exported functions, but we were returning a pointer to a non-exported item (dataStream).  I just modifed that, and the name was again a "non" linting name (Url), could be URL or url I went with url because there was no place I could find in the code that accessed it.

pkg/importer/dataStream.go
```diff
@@ -37,59 +38,59 @@ type DataStreamInterface interface {
...
 }
 
-// Return a dataStream object after validating the endpoint and constructing the reader/closer chain.
+// NewDataStream returns a DataStream object after validating the endpoint and constructing the reader/closer chain.
 // Note: the caller must close the `Readers` in reverse order. See Close().
-func NewDataStream(endpt, accKey, secKey string) (*dataStream, error) {
+func NewDataStream(endpt, accKey, secKey string) (*DataStream, error) {
 	if len(accKey) == 0 || len(secKey) == 0 {
-		glog.V(Vadmin).Infof("%s and/or %s are empty\n", IMPORTER_ACCESS_KEY_ID, IMPORTER_SECRET_KEY)
+		glog.V(2).Infof("%s and/or %s are empty\n", cdi.IMPORTER_ACCESS_KEY_ID, cdi.IMPORTER_SECRET_KEY)
```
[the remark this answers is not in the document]
j-griffith: yeah I know I know :)  I've been trying to do these initial changes package per package, that's caused some confusion with glog so I've since submitted a patch that just updated all of the glog calls in one patch. PR #408

pkg/importer/dataStream.go
```diff
@@ -412,7 +413,7 @@ func (d *dataStream) isoSize() (int64, error) {
 	const (
 		vdTypeOff       = 0
 		typeLen         = 1
-		vdIdOff         = 1
+		vdIDOff         = 1
```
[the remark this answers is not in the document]
j-griffith: It's personal choice until you run golint against it.  Honestly most of these aren't just things I've pulled out of the air, they're actual standardized *things*.  This isn't my hill to die on, I just volunteered to turn on golint; if we don't want it that's ok too but I do think it's valuable and very much worth while even if it is annoying at first.  I'm a very firm believer in gofmt, golint and govet, ESPECIALLY in an Open Source project where you hope to have a community of developers.

j-griffith:
@jeffvance Thanks, I'm trying to make these as small as I can by going with individual packages, granted I blew it here and put two of them in the same PR as seperate commits (my bad).  Not sure if you saw there's also #404 and #404 that follow this same pattern.  I'm avoiding dumping all of them at once, will probably just trickle these out as we go assuming they merge.

Thank you for the thorough review and taking the time, I do appreciate it!

hack/build/run-lint-checks.sh
copejon: @j-griffith have you seen the allPkgs()s function in config.sh?  This will return all go files in the projects (sans `vendor/` and the bugged generated file`).

j-griffith:
Latest update goes back to using the const for glog since that was a big sticking point and probably should've been in it's own change anyway (now in PR #408 ):
Vuser = 1
Vadmin = 2
Vdebug = 3

pkg/controller/util_test.go
jeffvance: Really? I wonder yhy golint cares about a var named `pvcUid` vs `id`? That seems odd to me... I haven't run it in a while so I guess I need to do that today.

pkg/importer/dataStream.go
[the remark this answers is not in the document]
jeffvance: I appreciate your analysis on which types, funcs, fields, etc should be exported vs. private! Jon and I put some thought in that but also some of the export vs private stuff just evolved as testing was added.

pkg/importer/dataStream.go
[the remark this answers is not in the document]
jeffvance: Agree. Take my review comments with a grain of salt because I didn't realize, while reviewing your prs last Fri, that the changes you made were necessary for `golint` to run.

jeffvance:
@j-griffith I may have come off a bit short last Fri when I reviewed this pr and I apologize for that. I reviewed this pr first and then your earlier prs next, so I didn't have their context in mind when I started this review. As I mentioned in another pr of yours, I think it would be beneficial to schedule a mtg with you, myself, @copejon and others who are interested and discuss CDI at a high-level and our choices for our coding practices.

PR #412, opened 2018-09-02
common: fixed key constants capitalization

danielerez:
Fixed capitalization to correalte with other constants in the file.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -38,8 +38,8 @@ const (
 	CLONER_SOCKET_PATH    = "/tmp/clone/socket"
 
 	// key names expected in credential secret
-	KeyAccess = "accessKeyId"
-	KeySecret = "secretKey"
+	KEY_ACCESS = "accessKeyId"
+	KEY_SECRET = "secretKey"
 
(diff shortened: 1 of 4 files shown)


danielerez:
@aglitke @awels @rollandf

rollandf:
I see in a lot of other places that we use mixed caps.
We need to decide for consistency.

copejon:
> We need to decide for consistency.

Agreed.  It'd be best to adhere to go idiom with cameCased const names.

j-griffith:
The Go standard is actually camel case instead of all caps.  I have a patch series that's leading up to those changes.. .something like this  https://gist.github.com/j-griffith/a65e80a2bc008258efe4b2399f0cd636

jeffvance:
> waiting to merge when previous review comments are addressed...
Since kubevirt implements all of golint`'s recommendations I'd say let golint decide. Not being tongue-in-cheek, if golint has an opinion on this then let's follow that recommendation.

jeffvance:
@danielerez Pls work with @j-griffith on this. My opinion as of now is that we should close your pr because of his series of prs that are about to be merged.

PR #415, opened 2018-09-02
Fix lint erros in image pkg

j-griffith:
Fix golint errors in image pkg

### DIFF
--- a/pkg/image/filefmt.go
+++ b/pkg/image/filefmt.go
@@ -7,16 +7,15 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
-
-	. "kubevirt.io/containerized-data-importer/pkg/common"
+	"kubevirt.io/containerized-data-importer/pkg/common"
 )
 
-// Size of buffer used to read file headers.
+// MaxExpectedHdrSize defines the Size of buffer used to read file headers.
 // Note: this is the size of tar's header. If a larger number is used the tar unarchive operation
 //   creates the destination file too large, by the difference between this const and 512.
 const MaxExpectedHdrSize = 512
 
-// key is file format, eg. "gz" or "tar", value is metadata describing the layout for this hdr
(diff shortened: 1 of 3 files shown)


j-griffith:
These errors seem to be intermittent, doing an empty re base to test that theory.

pkg/image/filefmt.go
```diff
@@ -67,10 +67,12 @@ func CopyKnownHdrs() Headers {
 	return m
 }
 
+// Match performs a check to see if the provided byte array matches the bytes in our header data
```
jeffvance: nit: technically byte slice not array...

pkg/image/filefmt.go
```diff
@@ -67,10 +67,12 @@ func CopyKnownHdrs() Headers {
...
 func (h Header) Match(b []byte) bool {
 	return bytes.Equal(b[h.mgOffset:h.mgOffset+len(h.magicNumber)], h.magicNumber)
 }
 
+// Size uses the Header receiver and the provided byte array to extract the size or the original file represented in the headers metadata
```
jeffvance: a bit more accurate would be:
```
// Size uses the Header receiver offset and length fields to extract, from the passed-in file header slice (b), the size of the original file.
```

pkg/image/validate.go
```diff
@@ -1,12 +1,20 @@
 package image
 
 const (
-	ExtImg   = ".img"
-	ExtIso   = ".iso"
-	ExtGz    = ".gz"
+	// ExtImg is a constant for the .img extenstion
+	ExtImg = ".img"
+	// ExtIso is a constant for the .img extenstion
```
jeffvance: copy/paste errors from here to bottom...

pkg/image/filefmt.go
j-griffith: done

jeffvance:
lgtm
Waiting for ci

PR #417, opened 2018-09-03
[WIP] Adding CDI e2e negative tests

zvikorn:
One of the tests I created is 'import invalid credentials' test case which fails right now as the source of the disk image does not require credentials. what should I test this against? should we create another container/pod for this specific test that returns error for invalid credentials?

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -57,6 +58,10 @@ var _ = Describe(testSuiteName, func() {
 			Expect(err).NotTo(HaveOccurred())
 			f.AddNamespaceToDelete(targetNs)
 			doCloneTest(f, targetNs)
+		})
+
+		It("Should not clone anything when CloneOf annotation exists", func() {
+			cloneOfAnnoExistenceTest(f, f.Namespace)
 		})
 	})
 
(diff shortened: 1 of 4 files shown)


tests/cloner_test.go
```diff
@@ -121,3 +125,32 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
 	}
 }
+
+func verifyTargetContent(f *framework.Framework, namespace *v1.Namespace, pvc *v1.PersistentVolumeClaim) bool {
+	By("Verify target PVC content matches source PVC")
+	executorPod, err := utils.CreateExecutorPodWithPVC(f.K8sClient, "verify-pvc-content", namespace.Name, pvc)
+	Expect(err).ToNot(HaveOccurred())
+	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
+	Expect(err).ToNot(HaveOccurred())
+	output := f.ExecShellInPod(executorPod.Name, namespace.Name, "cat "+testFile)
```
jeffvance: Is there value in this func also returning err?

tests/import_test.go
```diff
@@ -6,18 +6,44 @@ import (
...
+	bedEndpoint             = "http://gopats.com/who-is-the goat.iso"
+	endpoint                = "http://distro.ibiblio.org/tinycorelinux/9.x/x86/release/Core-current.iso" 
+	annServiceAccountName   = "kubernetes.io/service-account.name"
+	secretName              = "mysecret"
 	waitForCDIImportTimeout = 30 * time.Second
 )
 
 var _ = Describe(testSuiteName, func() {
-	f := framework.NewFrameworkOrDie(namespacePrefix)
+	f, err := framework.NewFramework(namespacePrefix, framework.Config{})
```
jeffvance: Consider calling the new [`NewFrameworkOrDie`](https://github.com/kubevirt/containerized-data-importer/blob/master/tests/framework/framework.go#L101). You do not need to check err and you do not need to pass in Config.

jeffvance:
looks good to me.
Needs rebase but should be ready to merge.

awels:
@zvikorn did you close this on purpose?

tests/cloner_test.go
```diff
@@ -121,3 +126,32 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
+func cloneOfAnnoExistenceTest(f *framework.Framework, targetNs *v1.Namespace) {
+	// Create targetPvc
+	By("Creating target pvc")
+	targetPvc, err := utils.CreatePVCFromDefinition(f.K8sClient, targetNs.Name, utils.NewPVCDefinition(
+		"target-pvc",
+		"1G",
+		map[string]string{controller.AnnCloneRequest: f.Namespace.Name + "/" + sourcePVCName, controller.AnnCloneOf: "true"},
+		nil))
+	Expect(err).ToNot(HaveOccurred())
+	err = utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, targetNs.Name, v1.ClaimBound, targetPvc.Name)
```
awels: I don't think you can wait for ClaimBound, as with local volume that will not happen unless a pod attaches the PVC to itself. Which won't happen.

zvikorn:
Yes. I’m going to open a new one soon. 
I saw your comment though. Thanks!

tests/cloner_test.go
awels: So just merged a patch that should make it easier to inspect the controller log to see if it is skipping the PVC you are interested in.

PR #419, opened 2018-09-04
Fix golint errors on expectations pkg

j-griffith:
Avoiding any real changes other than just the required items to make
golint pass.  Realizing that this is a copy/past of k8s pkg.

### DIFF
--- a/pkg/expectations/expectations.go
+++ b/pkg/expectations/expectations.go
@@ -36,11 +36,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
-
-	. "kubevirt.io/containerized-data-importer/pkg/common"
+	"kubevirt.io/containerized-data-importer/pkg/common"
 )
 
 const (
+	// ExpectationsTimeout sets a constant timeout value to 5 minutes
 	// If a watch drops a delete event for a pod, it'll take this long
 	// before a dormant controller waiting for those packets is woken up anyway. It is
 	// specifically targeted at the case where some problem prevents an update
@@ -54,6 +54,7 @@ const (
 	ExpectationsTimeout = 5 * time.Minute
 )
 
+//UpdateTaintBackoff provides a variable to hold a wait.Backoff item
 var UpdateTaintBackoff = wait.Backoff{
 	Steps:    5,
 	Duration: 100 * time.Millisecond,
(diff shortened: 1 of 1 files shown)


pkg/expectations/expectations.go
```diff
@@ -36,11 +36,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
-
-	. "kubevirt.io/containerized-data-importer/pkg/common"
```
mhenriks: Love that you are getting rid of these dot imports

pkg/expectations/expectations.go
```diff
@@ -125,18 +128,18 @@ type ControllerExpectations struct {
 
 // GetExpectations returns the ControlleeExpectations of the given controller.
 func (r *ControllerExpectations) GetExpectations(controllerKey string) (*ControlleeExpectations, bool, error) {
+	var err error
 	if exp, exists, err := r.GetByKey(controllerKey); err == nil && exists {
 		return exp.(*ControlleeExpectations), true, nil
-	} else {
-		return nil, false, err
 	}
+	return nil, false, err
```
jeffvance: can just return `nil, false, nil`, right?

j-griffith: @jeffvance Actually no, the (nil, false, err) case there is hit when there actual *is* an error so it's important.  The need to define the variable above is because I took out the explicit else to simplify the function, and in that case the go compiler gets upset because the def of the variable is in an if statement.

Anyway, not a lint problem; I can just put it back the way it was.  Or drop the compound statement and do a more traditional "if err != nil" pattern.

pkg/expectations/expectations.go
```diff
@@ -125,18 +128,18 @@ type ControllerExpectations struct {
 
 // GetExpectations returns the ControlleeExpectations of the given controller.
 func (r *ControllerExpectations) GetExpectations(controllerKey string) (*ControlleeExpectations, bool, error) {
+	var err error
```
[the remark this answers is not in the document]
j-griffith: Because the declaration in the if statement below is "in an if statement" so what happens is without an "else" the compiler sees it as possibly never set/initialized.

It's sort of one problem leading to another; typically returns in if/else blocks is wasteful, as opposed to just falling through to a default return; so fixing that means now there's not a guarantee of err being set (which in this logic the default return is when an error exists so it needs to be set) and as a result it needs to be declared.

You can check some of this out by just running `golint xxxx.go`

pkg/expectations/expectations.go
jeffvance: I should have been clearer above. Here's what I was meaning to say:
```
func (r *ControllerExpectations) GetExpectations(controllerKey string) (*ControlleeExpectations, bool, error) {
	var err error
	if exp, exists, err := r.GetByKey(controllerKey); err == nil && exists {
/// these 3 ^^ vars are scoped in the if block and not the func
		return exp.(*ControlleeExpectations), true, nil
	}
// here we have either 1) an err or 2) no err but !exists (but exists it not scoped here)
	return nil, false, err
// however, err in the return is always false (its "zero" value)
}
```
I think I would refactor as:
```
func (r *ControllerExpectations) GetExpectations(controllerKey string) (*ControlleeExpectations, bool, error) {
	exp, exists, err := r.GetByKey(controllerKey)
        if exp != nil {
		return exp.(*ControlleeExpectations), exists, err
	}
	return nil, exists, err
}
```

j-griffith: Much better, I'll add it, thanks!

PR #422, opened 2018-09-04
Fix golint errors on register package

j-griffith:


pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -31,56 +32,79 @@ type DataVolume struct {
...
 	PVC  *DataVolumeSourcePVC  `json:"pvc,omitempty"`
 }
 
+// DataVolumeSourcePVC provides the parameters to create a Data Volume from an existing PVC
 type DataVolumeSourcePVC struct {
 	Namespace string `json:"namespace,omitempty"`
 	Name      string `json:"name,omitempty"`
 }
 
+// DataVolumeSourceS3 provides the parameters to create a Data Volume from an S3 source
```
jeffvance: DataVolumes also work with http sources, AFAIK.

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -31,56 +32,79 @@ type DataVolume struct {
...
 type DataVolumeSourceS3 struct {
-	URL       string `json:"url,omitempty"`
+	//URL is the url of the S3 source
```
jeffvance: ditto re http sources

pkg/apis/datavolumecontroller/v1alpha1/types.go
j-griffith: ?  This comment is for a DataVolumeS3 struct?

pkg/apis/datavolumecontroller/v1alpha1/types.go
j-griffith: Same as above, this is the DataVolumeSourceS3.URL member.

pkg/apis/datavolumecontroller/v1alpha1/types.go
jeffvance: Nevermind... I didn't notice that.

jeffvance:
lgtm

PR #427, opened 2018-09-05
CDI e2e negative tests

zvikorn:

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -2,30 +2,28 @@ package tests
 
 import (
 	"fmt"
-	"time"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
 	"k8s.io/api/core/v1"
-
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
 	"kubevirt.io/containerized-data-importer/tests/framework"
 	"kubevirt.io/containerized-data-importer/tests/utils"
+	"strings"
+	"time"
 )
 
 const (
-	testSuiteName         = "Cloner Test Suite"
-	namespacePrefix       = "cloner"
-	sourcePodFillerName   = "fill-source"
-	sourcePVCName         = "source-pvc"
-	fillData              = "123456789012345678901234567890123456789012345678901234567890"
-	testFile              = utils.DefaultPvcMountPath + "/source.txt"
-	fillCommand           = "echo \"" + fillData + "\" >> " + testFile
-	assertionPollInterval = 2 * time.Second
-	cloneCompleteTimeout  = 60 * time.Second
-	shortTimeout          = 2 * time.Second
+	testSuiteName                    = "Cloner Test Suite"
+	namespacePrefix                  = "cloner"
+	sourcePodFillerName              = "fill-source"
+	sourcePVCName                    = "source-pvc"
+	fillData                         = "123456789012345678901234567890123456789012345678901234567890"
+	testFile                         = utils.DefaultPvcMountPath + "/source.txt"
+	fillCommand                      = "echo \"" + fillData + "\" >> " + testFile
+	assertionPollInterval            = 2 * time.Second
+	cloneCompleteTimeout             = 60 * time.Second
+	controllerSkipPVCCompleteTimeout = 60 * time.Second
 )
 
 var _ = Describe(testSuiteName, func() {
@@ -57,6 +55,9 @@ var _ = Describe(testSuiteName, func() {
 			Expect(err).NotTo(HaveOccurred())
 			f.AddNamespaceToDelete(targetNs)
 			doCloneTest(f, targetNs)
+		})
+		It("Should not clone anything when CloneOf annotation exists", func() {
+			cloneOfAnnoExistenceTest(f, f.Namespace.Name)
 		})
 	})
 
@@ -126,4 +127,23 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
 	deleted, err = utils.WaitPodDeleted(f.K8sClient, targetPod.Name, targetNs.Name, time.Second*40)
 	Expect(err).ToNot(HaveOccurred())
 	Expect(deleted).To(BeTrue())
+}
+
+func cloneOfAnnoExistenceTest(f *framework.Framework, targetNamespaceName string) {
+	// Create targetPvc
+	By(fmt.Sprintf("Creating target pvc: %s/target-pvc", targetNamespaceName))
+	_, err := utils.CreatePVCFromDefinition(f.K8sClient, targetNamespaceName, utils.NewPVCDefinition(
+		"target-pvc",
+		"1G",
+		map[string]string{controller.AnnCloneRequest: f.Namespace.Name + "/" + sourcePVCName, controller.AnnCloneOf: "true"},
+		nil))
+	Expect(err).ToNot(HaveOccurred())
+	By("Checking no cloning pods were created")
+
+	Eventually(func() bool {
+		log, err := RunKubectlCommand(f, "logs", f.ControllerPod.Name, "-n", f.CdiInstallNs)
+		Expect(err).NotTo(HaveOccurred())
+		return strings.Contains(log, "pvc annotation \""+controller.AnnCloneOf+"\" exists indicating cloning completed, skipping pvc \""+f.Namespace.Name+"/target-pvc\"")
+	}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(BeTrue())
+	Expect(err).ToNot(HaveOccurred())
 }
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -7,6 +7,8 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"k8s.io/api/core/v1"
+
 	"kubevirt.io/containerized-data-importer/pkg/controller"
 	"kubevirt.io/containerized-data-importer/tests"
 	"kubevirt.io/containerized-data-importer/tests/framework"
@@ -18,6 +20,8 @@ const (
 	namespacePrefix                  = "importer"
 	assertionPollInterval            = 2 * time.Second
 	controllerSkipPVCCompleteTimeout = 60 * time.Second
+	invalidEndpoint                  = "http://gopats.com/who-is-the-goat.iso"
+	waitForCDIImportTimeout          = 30 * time.Second
 )
 
 var _ = Describe(testSuiteName, func() {
@@ -35,5 +39,18 @@ var _ = Describe(testSuiteName, func() {
 		// Wait a while to see if CDI puts anything in the PVC.
 		Expect(framework.VerifyPVCIsEmpty(f, pvc)).To(BeTrue())
 		// Not deleting PVC as it will be removed with the NS removal.
+	})
+	It("Import pod status should be Fail on unavailable endpoint", func() {
+		pvc, err := f.CreatePVCFromDefinition(utils.NewPVCDefinition(
+			"no-import",
+			"1G",
+			map[string]string{controller.AnnEndpoint: invalidEndpoint},
+			nil))
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Verify the pod status is Failed on the target PVC")
+		status, phaseAnnotation, err := utils.WaitForPVCAnnotation(f.K8sClient, f.Namespace.Name, pvc, controller.AnnPodPhase)
+		Expect(phaseAnnotation).To(BeTrue())
+		Expect(status).Should(BeEquivalentTo(v1.PodPending))
 	})
 })
--- a/tests/transport_test.go
+++ b/tests/transport_test.go
@@ -27,9 +27,10 @@ var _ = Describe("Transport Tests", func() {
 	)
 
 	var (
-		ns string
-		f  = framework.NewFrameworkOrDie("transport", framework.Config{SkipNamespaceCreation: false})
-		c  = f.K8sClient
+		ns  string
+		f   = framework.NewFrameworkOrDie("transport", framework.Config{SkipNamespaceCreation: false})
+		c   = f.K8sClient
+		sec *v1.Secret
 	)
 
 	BeforeEach(func() {
@@ -39,10 +40,11 @@ var _ = Describe("Transport Tests", func() {
 	})
 
 	// it() is the body of the test and is executed once per Entry() by DescribeTable()
-	it := func(ep, file, accessKey, secretKey string) {
+	// closes over c and ns
+	it := func(ep, file, accessKey, secretKey string, shouldSucceed bool) {
+
 		var (
 			err error // prevent shadowing
-			sec *v1.Secret
 		)
 
 		pvcAnn := map[string]string{
@@ -52,10 +54,16 @@ var _ = Describe("Transport Tests", func() {
 
 		if accessKey != "" || secretKey != "" {
 			By(fmt.Sprintf("Creating secret for endpoint %s", ep))
-			stringData := map[string]string{
-				common.KeyAccess: utils.AccessKeyValue,
-				common.KeySecret: utils.SecretKeyValue,
+			if accessKey == "" {
+				accessKey = utils.AccessKeyValue
 			}
+			if secretKey == "" {
+				secretKey = utils.SecretKeyValue
+			}
+			stringData := make(map[string]string)
+			stringData[common.KeyAccess] = accessKey
+			stringData[common.KeySecret] = secretKey
+
 			sec, err = utils.CreateSecretFromDefinition(c, utils.NewSecretDefinition(nil, stringData, nil, ns, secretPrefix))
 			Expect(err).NotTo(HaveOccurred(), "Error creating test secret")
 			pvcAnn[controller.AnnSecret] = sec.Name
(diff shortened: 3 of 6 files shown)


tests/import_test.go
```diff
@@ -18,6 +20,9 @@ const (
 	namespacePrefix                  = "importer"
 	assertionPollInterval            = 2 * time.Second
 	controllerSkipPVCCompleteTimeout = 60 * time.Second
+	bedEndpoint                      = "http://gopats.com/who-is-the goat.iso"
```
awels: Should this be 'badEndpoint'? Also you have a space in the URL.

zvikorn: oops for both. thanks.

tests/transport_test.go
```diff
@@ -60,12 +60,21 @@ var _ = Describe("Transport Tests", func() {
...
+					common.KEY_ACCESS: accessKey,
+					common.KEY_SECRET: secretKey,
+				}
+			} else {
+				stringData = map[string]string{
+					common.KEY_ACCESS: utils.AccessKeyValue,
+					common.KEY_SECRET: utils.SecretKeyValue,
+				}
 			}
+
```
jeffvance: slightly cleaner, imo, to refactor as:
```
		var stringData map[string]string
		if credentialed {
			By(fmt.Sprintf("Creating secret for endpoint %s", ep))
                        if accessKey == "" {
                                      accessKey =  utils.AccessKeyValue
                        }
                        if secretKey == "" {
                                       secretKey == utils.SecretKeyValu
                        }
		       stringData[common.KEY_ACCESS] = accessKey
                       stringData[common.KEY_SECRET] = secretKey
...
```

tests/utils/secrets.go
```diff
@@ -48,3 +48,13 @@ func CreateSecretFromDefinition(c *kubernetes.Clientset, secret *v1.Secret) (*v1
...
 	return secret, nil
 }
+
+func DeleteSecret(clientSet *kubernetes.Clientset, namespace string, secret v1.Secret) error {
+	return wait.PollImmediate(SecretPollInterval, SecretPollPeriod, func() (bool, error) {
+		err := clientSet.CoreV1().Secrets(namespace).Delete(secret.GetName(), nil)
+		if err == nil || apierrs.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
```
jeffvance: The return of a non-nil err stops `PollImmediate`

tests/cloner_test.go
```diff
@@ -121,3 +125,30 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
+	By("Creating target pvc")
+	_, err := utils.CreatePVCFromDefinition(f.K8sClient, targetNs.Name, utils.NewPVCDefinition(
+		"target-pvc",
+		"1G",
+		map[string]string{controller.AnnCloneRequest: f.Namespace.Name + "/" + sourcePVCName, controller.AnnCloneOf: "true"},
+		nil))
+	Expect(err).ToNot(HaveOccurred())
+	By("Checking no cloning pods were created")
+	_, err = f.FindPodByPrefix(common.CLONER_SOURCE_PODNAME)
+	Expect(err).To(HaveOccurred())
```
copejon: `apierrs.IsNotFound()` might be a more precise check here (from
`apierrs "k8s.io/apimachinery/pkg/api/errors"`)

tests/cloner_test.go
```diff
@@ -121,3 +125,30 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
+	_, err = utils.FindPodByPrefix(f.K8sClient, targetNs.Name, common.CLONER_TARGET_PODNAME, common.CDI_LABEL_SELECTOR)
+	Expect(err).To(HaveOccurred())
```
copejon: same

tests/transport_test.go
```diff
@@ -48,7 +48,7 @@ var _ = Describe("Transport Tests", func() {
 
 	// it() is the body of the test and is executed once per Entry() by DescribeTable()
 	// closes over c and ns
-	it := func(ep string, credentialed bool) {
+	it := func(ep string, credentialed bool, accessKey, secretKey string, shouldSucceed bool) {
```
copejon: `credentialed bool` should be dropped as it's redundant with `accessKey` and `secretKey`.

tests/transport_test.go
```diff
@@ -60,12 +60,21 @@ var _ = Describe("Transport Tests", func() {
 			sec *v1.Secret
 		)
 
+		var stringData map[string]string
 		if credentialed {
```
copejon: Should be
`if accessKey != "" || secretKey != "" {...}`

tests/transport_test.go
jeffvance: I was thinking similar... if accessKey != "" || secretKey != "" { // then we have credentialed==true

tests/transport_test.go
```diff
@@ -60,12 +60,21 @@ var _ = Describe("Transport Tests", func() {
...
-				common.KEY_ACCESS: utils.AccessKeyValue,
-				common.KEY_SECRET: utils.SecretKeyValue,
+			if accessKey != "" && secretKey != "" {
+				stringData = map[string]string{
+					common.KEY_ACCESS: accessKey,
+					common.KEY_SECRET: secretKey,
+				}
+			} else {
+				stringData = map[string]string{
+					common.KEY_ACCESS: utils.AccessKeyValue,
```
copejon: If the credentials aren't supplied, they should be left empty

tests/transport_test.go
```diff
@@ -78,13 +87,19 @@ var _ = Describe("Transport Tests", func() {
...
+			By("Verifying PVC is empty")
+			Expect(framework.VerifyPVCIsEmpty(f, pvc)).ToNot(BeFalse())
+		}
 	}
 
 	DescribeTable("Transport Test Table", it,
-		Entry("should connect to http endpoint without credentials", httpNoAuthEp, false),
-		Entry("should connect to http endpoint with credentials", httpAuthEp, true))
+		Entry("should connect to http endpoint without credentials", httpNoAuthEp, false, "", "", true),
+		Entry("should connect to http endpoint with valid credentials", httpAuthEp, true, "", "", true),
```
copejon: 2nd Entry() should pass utils.AccessKeyValue and utils.SecretKeyValue

tests/transport_test.go
jeffvance: It may that credential is indicated only by `accessKey != ""` and the secret value can be ignored for this test?

tests/cloner_test.go
```diff
@@ -121,3 +125,30 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
 		Expect(err).ToNot(HaveOccurred())
 	}
 }
+
+func verifyTargetContent(f *framework.Framework, namespace *v1.Namespace, pvc *v1.PersistentVolumeClaim) bool {
```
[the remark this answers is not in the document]
zvikorn: Don't know where I imported this function from and why. removed it.
you absolutely right. passing  the name now.

tests/cloner_test.go
```diff
@@ -121,3 +125,30 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
+	executorPod, err := utils.CreateExecutorPodWithPVC(f.K8sClient, "verify-pvc-content", namespace.Name, pvc)
+	Expect(err).ToNot(HaveOccurred())
+	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
+	Expect(err).ToNot(HaveOccurred())
+	output := f.ExecShellInPod(executorPod.Name, namespace.Name, "cat "+testFile)
+	f.DeletePod(executorPod)
+	return strings.Compare(fillData, output) == 0
+}
+
+func cloneOfAnnoExistenceTest(f *framework.Framework, targetNs *v1.Namespace) {
```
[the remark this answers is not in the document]
zvikorn: you absolutely right. passing  the name now.

tests/transport_test.go
zvikorn: Maybe I'm missing something....
there are two scenarios we need to address:
1) no auth
2) with auth

In case of auth, we have two other options:
a) accessKey and secretKey taken from Utils
b) accessKey and secretKey passed as arguments.

The the example that you guys show above, it does not cover all options.
Jeff's example does not cover the case of Auth with keys from utils. In this case the keys we pass are empty but we need to take them from utils. We do not do that outside the if clause.
In other words, Jeff's example implement a case where at least one of the keys contains a value as an argument.

Bottom line, the logic I am trying to convey is that first we want to know if we ask for auth or not.
After we know that, in case we do need auth, we need to check whether the keys where passed or should we take the default (utils).

If we are not on the same track, we can talk over bj (Jon, since Jeff is in the west cost, we can do it just us as you and Jeff come with the same approach here).

tests/transport_test.go
zvikorn: meaning, no secret? just accessKey? Is it po ssible? if so, I can change this

tests/utils/secrets.go
zvikorn: so if we do not want to "stop polling", should we ignore the err and return nil instead?

tests/transport_test.go
zvikorn: done

tests/transport_test.go
zvikorn: changed the entire logic as above

tests/transport_test.go
zvikorn: sure

tests/transport_test.go
zvikorn: initially, I did not understand that you guys want me to pass the default keys in the second Entry call, so I let the code do this logic.  I change the code to reflect this now. Thanks!

tests/cloner_test.go
zvikorn: how do we use it?

tests/transport_test.go
zvikorn: okay

tests/cloner_test.go
jeffvance:
```
import ( ...
        apierrs "k8s.io/apimachinery/pkg/api/errors"
...

	_, err = f.FindPodByPrefix(common.CLONER_SOURCE_PODNAME)
	Expect(apierrs.IsNotFound(err)).ToBe(True())
	_, err = utils.FindPodByPrefix(f.K8sClient, targetNamespaceName, common.CLONER_TARGET_PODNAME, common.CDI_LABEL_SELECTOR)
```

zvikorn:
Still do not understand why we get the timeout error. @awels ?

should connect to http endpoint with invalid credentials [It]    //this should be changed to 'should NOT'
    /go/src/kubevirt.io/containerized-data-importer/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected success, but got an error:
        <*errors.errorString | 0xc4200bf530>: {
            s: "timed out waiting for the condition",
        }
        timed out waiting for the condition

    /go/src/kubevirt.io/containerized-data-importer/tests/**transport_test.go:78**
------------------------------

tests/import_test.go
```diff
@@ -36,4 +40,18 @@ var _ = Describe(testSuiteName, func() {
...
 	})
+	It("Import pod status should be Fail on unavailable endpoint", func() {
+		pvc, err := f.CreatePVCFromDefinition(utils.NewPVCDefinition(
+			"no-import",
+			"1G",
+			map[string]string{controller.AnnEndpoint: invalidEndpoint},
+			nil))
+		Expect(err).ToNot(HaveOccurred())
+		// Wait a while to see if CDI puts anything in the PVC.
+		time.Sleep(waitForCDIImportTimeout)
```
jeffvance: unconditional sleeps generally should be avoided. It's usually better to poll for a condition with a small sleep between each iteration. The advantages of this approach are 1) potential for a much shorter sleep time thus helping CI to run faster, and 2) better chance of sleeping long enough for the desired condition to occur.

tests/import_test.go
```diff
@@ -36,4 +40,18 @@ var _ = Describe(testSuiteName, func() {
...
 		// Not deleting PVC as it will be removed with the NS removal.
 	})
+	It("Import pod status should be Fail on unavailable endpoint", func() {
+		pvc, err := f.CreatePVCFromDefinition(utils.NewPVCDefinition(
+			"no-import",
+			"1G",
+			map[string]string{controller.AnnEndpoint: invalidEndpoint},
+			nil))
+		Expect(err).ToNot(HaveOccurred())
+		// Wait a while to see if CDI puts anything in the PVC.
```
jeffvance: this comments does not represent what the code does

tests/utils.go
```diff
@@ -54,6 +56,7 @@ func PrintControllerLog(f *framework.Framework) {
 	}
 }
 
+//PanicOnError ...
```
jeffvance: Is this here just to pass lint checks? Otherwise not really a useful comment...

tests/utils/secrets.go
jeffvance: Yes. Polling stops 1st if err != nil and 2nd if done is true and 3rd if the overall time expires.

zvikorn: Changed it a bit. Please let me know what do you think...

tests/import_test.go
zvikorn: utils.WaitForPVCAnnotation in line 53 does the polling already.
Should we remove the sleep then? Do you see why it's there if we are polling anyway?

tests/utils/secrets.go
zvikorn: Also added a call to this method in transport_test.go

tests/import_test.go
jeffvance: I know your didn't change the func below but I reviewed it based on your reply above, and I think it should be refactored a bit due to a) idiomatic go, b) bug where nil pvc ptr can be referenced:
```
func WaitForPVCAnnotation(clientSet *kubernetes.Clientset, namespace string, pvc *k8sv1.PersistentVolumeClaim, annoKey string) (annoValue string, found bool, err error) {
	err = wait.PollImmediate(PVCPollInterval, PVCCreateTime, func() (bool, error) {
		pvc, err := FindPVC(clientSet, namespace, pvc.Name)
		if err != nil {
			return false, nil // keep trying
		}
		if annoValue, found = pvc.ObjectMeta.Annotations[annoKey] ; found {
			return true, nil
		}
		return false, err
	})
	return
}
```

jeffvance: and, to answer your question, I don't see a reason why the `sleep` above is needed.

tests/transport_test.go
zvikorn: command := `expSize=$(curl -s http://cdi-file-host.kube-system | grep -E '"name":"tinyCore.iso"' | grep -Eo '"size":[0-9]+' | tr -d '":[:alpha:]'); haveSize=$(wc -c < /pvc/disk.img); (( $expSize == $haveSize )); echo $?`

exitCode := f.ExecShellInPod(pod.Name, ns, command)
// A 0 exitCode should indicate that $expSize == $haveSize
Expect(strconv.Atoi(exitCode)).To(BeZero())

returns error:

Transport Tests
/go/src/kubevirt.io/containerized-data-importer/tests/transport_test.go:20
  Transport Test Table
  /go/src/kubevirt.io/containerized-data-importer/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    should not connect to http endpoint with invalid credentials [It]
    /go/src/kubevirt.io/containerized-data-importer/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected
        <int>: 1
    to be zero-valued

    /go/src/kubevirt.io/containerized-data-importer/tests/transport_test.go:96

tests/import_test.go
zvikorn: ok

tests/utils/secrets.go
```diff
@@ -48,3 +48,14 @@ func CreateSecretFromDefinition(c *kubernetes.Clientset, secret *v1.Secret) (*v1
 	}
 	return secret, nil
 }
+
+func DeleteSecret(clientSet *kubernetes.Clientset, namespace string, secret v1.Secret) error {
+	e := wait.PollImmediate(SecretPollInterval, SecretPollPeriod, func() (bool, error) {
+		err := clientSet.CoreV1().Secrets(namespace).Delete(secret.GetName(), nil)
```
jeffvance: don't worry about `err` being assigned here. This copy of `err` is referred to as a _shadow_ variable and is its own instance of error.

tests/utils/secrets.go
```diff
@@ -48,3 +48,14 @@ func CreateSecretFromDefinition(c *kubernetes.Clientset, secret *v1.Secret) (*v1
...
+		if err == nil || apierrs.IsNotFound(err) {
+			return true, nil
+		}
+		return false, nil //keep polling
+	})
+	return e
+}
```
jeffvance: looks good.

tests/cloner_test.go
```diff
@@ -120,3 +124,19 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
 		Expect(err).ToNot(HaveOccurred())
 	}
 }
+
+func cloneOfAnnoExistenceTest(f *framework.Framework, targetNamespaceName string) {
+	// Create targetPvc
+	By("Creating target pvc")
```
jeffvance: Please append namespace to this text. Standard format, if you don't want two separate inserts is:
   _pvcNamespace/pvcName_

tests/cloner_test.go
```diff
@@ -120,3 +124,19 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
+	_, err := utils.CreatePVCFromDefinition(f.K8sClient, targetNamespaceName, utils.NewPVCDefinition(
+		"target-pvc",
+		"1G",
+		map[string]string{controller.AnnCloneRequest: f.Namespace.Name + "/" + sourcePVCName, controller.AnnCloneOf: "true"},
+		nil))
+	Expect(err).ToNot(HaveOccurred())
+	By("Checking no cloning pods were created")
+	_, err = f.FindPodByPrefix(common.ClonerSourcePodName)
```
jeffvance: Sorry, I know this isn't you code but there's a slight idiomatic change I'd make to _FindPodByPrefix_:
```
func FindPodByPrefix(clientSet *kubernetes.Clientset, namespace, prefix, labelSelector string) (*k8sv1.Pod, error) {
...
		podList, err := clientSet.CoreV1().Pods(namespace).List(metav1.ListOptions{
			LabelSelector: labelSelector,
		})
		if err == nil {
				...
			}
		}
		return found, nil  // <==== here, don't need 2 returns, remove return in `err == nil` block
	})
```

tests/cloner_test.go
```diff
@@ -121,3 +125,30 @@ func doCloneTest(f *framework.Framework, targetNs *v1.Namespace) {
...
+func cloneOfAnnoExistenceTest(f *framework.Framework, targetNs *v1.Namespace) {
+	// Create targetPvc
+	By("Creating target pvc")
+	_, err := utils.CreatePVCFromDefinition(f.K8sClient, targetNs.Name, utils.NewPVCDefinition(
+		"target-pvc",
+		"1G",
+		map[string]string{controller.AnnCloneRequest: f.Namespace.Name + "/" + sourcePVCName, controller.AnnCloneOf: "true"},
+		nil))
+	Expect(err).ToNot(HaveOccurred())
+	By("Checking no cloning pods were created")
```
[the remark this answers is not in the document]
jeffvance: @awels `FindPodByPrefix` polls for the pod in the given namespace with the given prefix to be present, or returns an err if the elapsed time expires. I may be missing something but I think the existing code works.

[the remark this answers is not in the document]
awels: You are right it will work, however we have to wait the entire timeout. My version we can check immediately if it is being skipped.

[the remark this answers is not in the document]
jeffvance: > however we have to wait the entire timeout. My version we can check immediately if it is being skipped.

I may not be fully understanding your comment above. Are you saying to replace the call to `FindPodByPrefix` with an `Eventually` loop? Or are you saying we should add the `Eventually` code before calling `FindPodByPrefix`?

tests/cloner_test.go
zvikorn: sure.

tests/cloner_test.go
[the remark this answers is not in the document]
jeffvance: @awels Thanks and I agree.

zvikorn:
@aglitke @mhenriks @j-griffith @awels This was reviewed almost a month ago. If there are no more comments, could you please approve it?

awels:
@zvikorn I have open comments about the negative tests using the controller log instead of waiting 30 seconds for the pod to not appear.

zvikorn:
Aren't we talking about the 'cloneOfAnnoExistenceTest' method? I changed it e discussed this...
Can you point me to the line you are talking about?

awels:
Oh you did change it, I hadn't seen the conversation resolved (which usually happens when you change code, or if you manually click resolve). In that case looks good to me.

awels:
ci please test

awels:
ci test please

PR #428, opened 2018-09-05
Add golint enforcement to all pkgs but common

j-griffith:
Now that the updates have merged to fix golint errors in all of the
CDI pkgs (except common) turn on lint checks to keep things up to date.

This patch does that by:
1.  Adding a check in hack/build/run-lint-checks.sh
    NOTE:  We temporarily set a list variable in the script to indicate
    the directories that we've fixed and want to enforce.  This will be
    removed when common is "lintified"
2.  Adding golint to our build containers Dockerfile

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -23,7 +23,8 @@ RUN \
     go install ./... ) && \
     ( go get -d github.com/mattn/goveralls && \
     cd $GOPATH/src/github.com/mattn/goveralls && \
-    go install ./... )
+    go install ./... ) && \
+    ( go get -u golang.org/x/lint/golint )
 
 ADD entrypoint.sh /entrypoint.sh
 
--- a/hack/build/run-lint-checks.sh
+++ b/hack/build/run-lint-checks.sh
@@ -20,6 +20,9 @@
 source hack/build/config.sh
 source hack/build/common.sh
 
+# Temporary var for pkgs that are ready to enforce linting
+LINTABLE=(pkg/controller pkg/apis pkg/client pkg/controller pkg/expectations pkg/image pkg/importer pkg/lib pkg/util)
+
 ec=0
 out="$(gofmt -l -s ${SOURCE_DIRS} | grep ".*\.go")"
 if [[ ${out} ]]; then
@@ -27,4 +30,15 @@ if [[ ${out} ]]; then
     echo "${out}"
     ec=1
 fi
+
+for p in "${LINTABLE[@]}"; do
+    echo "Performing lint checks on: ${p}"
+    out="$(golint ${p}/...)"
+    if [[ ${out} ]]; then
+        echo "FAIL: following golint errors found:"
+        echo "${out}"
+        ec=1
+    fi
+done
+
 exit ${ec}


hack/build/docker/builder/Dockerfile
```diff
@@ -23,7 +23,8 @@ RUN \
     go install ./... ) && \
     ( go get -d github.com/mattn/goveralls && \
     cd $GOPATH/src/github.com/mattn/goveralls && \
-    go install ./... )
+    go install ./... ) && \
+    ( go get -u golang.org/x/lint/golint )
```
awels: Do we need a 
cd $GOPATH/src/golang.org/x/lint/golint && \
go install ./...
?

hack/build/run-lint-checks.sh
```diff
@@ -20,11 +20,25 @@
...
     echo "FAIL: Format errors found in the following files:"
     echo "${out}"
     ec=1
 fi
+
+for p in "${LINTABLE}"
+do
+  out="$(golint ${p}/...)"
+    if [[ ${out}  ]]; then
+    echo "FAIL: following golint errors found:"
```
jeffvance: Indent here and below. Does `shfmt` report this?

hack/build/run-lint-checks.sh
```diff
@@ -20,11 +20,25 @@
...
 ec=0
 out="$(gofmt -l -s ${SOURCE_DIRS} | grep ".*\.go")"
 if [[ ${out} ]]; then
     echo "FAIL: Format errors found in the following files:"
     echo "${out}"
     ec=1
 fi
+
+for p in "${LINTABLE}"
+do
```
jeffvance: I think I'd echo the pkg inside the loop

hack/build/run-lint-checks.sh
```diff
@@ -20,11 +20,25 @@
 source hack/build/config.sh
 source hack/build/common.sh
 
+# Temporary var for pkgs that are ready to enforce linting
+LINTABLE=(pkg/controller pkg/apis pkg/client pkg/controller pkg/expectations pkg/image pkg/importer pkg/lib pkg/util)
```
copejon: We should include `tests/...` in this as well.

hack/build/docker/builder/Dockerfile
j-griffith: Yeah, I intentionally did not use the "-d" flag here for that reason.  I wasn't sure if there was a reason for the pattern here of using "got get -d" and then manually installing.  "go get -d" is special in that it only downloads and skips the install.  Things like "go get -u" however will download (including update) and install for you so you don't need the extra steps.

I figured there might be a reason for doing it the way it is here but for the case of golint it just seemed like an unnecessary extra step.

hack/build/run-lint-checks.sh
j-griffith: Ahh... yeah; and more importantly that for call to iterate on the array is wrong (missing the extra [@])

hack/build/run-lint-checks.sh
j-griffith: Great point, and you're right when that's added I noticed the bug in the syntax of the array.  Nice catch!

hack/build/run-lint-checks.sh
j-griffith: Yeah, definitely except they're not passing yet.  There's a ton of work to do in common first then the tests.  Tests are bottom of the list for now IMO, but they're not in this change because they won't pass.

jeffvance: Nit, butr just fyi: `LINTABLE` can be a simple string with space separators. Then the `for` loop below is just `for p in "$LINTABLE ; do`

j-griffith: Oh nice!  Thanks, much easier to remember than the crazy array syntax for bash too as a bonus.

j-griffith:
ci test please

jeffvance:
lgtm

PR #435, opened 2018-09-06
cleaning up dangling volumes from cluster-up

### LINKED ISSUE #434: left over docker volumes from cluster-up after running functional tests and running make cluster-down
after running `make cluster-up` and `make cluster-down` volumes are left dangling in docker - these can add up on my local system, not sure how this impacts travis-ci/jenkins
```
$ docker volume ls
DRIVER              VOLUME NAME
local               082f3490e8f26d6d1ebed6b0edef69354e8bd8351043015171710f0b73a94029
local               7cfb864090d5fb7f388698e80875a52fc84f38bd4167ed2c4ceef6737a168529
```

screeley44:
fixes #434 
not sure this is necessary in ci environment but I noticed locally these volumes add up on my system and eat up all my space, so adding a little clean up routine.

### DIFF
--- a/cluster/down.sh
+++ b/cluster/down.sh
@@ -2,4 +2,13 @@
 
 source ./cluster/gocli.sh
 
+echo "Bringing down cluster and client ..."
 $gocli rm
+
+# clean up unused docker volumes
+danglingVols="$(docker volume ls -qf dangling=true)"
+if [[ $? == 0 && -n "$danglingVols" ]]; then
+    echo "Cleaning up docker and dangling volumes ..."
+    docker volume rm $danglingVols
+fi
+


cluster/down.sh
```diff
@@ -2,4 +2,11 @@
 
 source ./cluster/gocli.sh
 
+echo "Bringing down cluster and client ..."
 $gocli rm
+
+# clean up unused docker volumes
+if [ "$(docker volume ls | wc -l)" -gt 1 ]; then
```
awels: Should you check for only dangling=true so something like:
```bash
if [ "$(docker volume ls -qf dangling=true | wc -l)" -gt 0]; then
```

j-griffith: These are isolated inside the test VM so I'm not sure we can't just treat this as our own env and completely clean it.  That being said, I like the suggested refactor from @jeffvance

aglitke: Could this interfere with other docker workflows happening on the same machine?  Should we try to find and remove only images that this infrastructure created?

j-griffith: Given that this runs in the VM (Cluster) built by cluster-up seems like these are *all* our docker workflows.  I'd suggest either the refactor @jeffvance suggested, or even a slight variant and just drop the dangling=true.  If you reworked slightly with either of those I'd be +2/A on this.

jeffvance:
lgtm
@awels are you ok with the _dangling=true_ filter?

awels:
I am fine with it.

PR #443, opened 2018-09-07
deprecation of local endpoint support

jeffvance:
Completes [338](https://github.com/kubevirt/containerized-data-importer/issues/338).

Removal of all code and tests related to local (file://) endpoints.

TODO: unit tests still are coded for local endpoints and need to be changed to use the new file server.

@aglitke @copejon @nellyc

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,6 @@ In Addition, the Containerized Data Importer gives the option to clone the impor
 1. [Hacking (WIP)](hack/README.md#getting-started-for-developers)
 1. [Security Configurations](#security-configurations)
 
-> DEPRECATION NOTICE:
-Support for local (file://) endpoints will be removed from CDI in the next release. There is no replacement and no work-around. All import endpoints must reference http(s) or s3 endpoints.
-
-
 ## Overview
 
 ### Purpose
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -41,7 +41,4 @@ func main() {
 		os.Exit(1)
 	}
 	glog.V(1).Infoln("import complete")
-
-	// temporary local import deprecation notice
-	glog.Warningf("\nDEPRECATION NOTICE:\n   Support for local (file://) endpoints will be removed from CDI in the next release.\n   There is no replacement and no work-around.\n   All import endpoints must reference http(s) or s3 endpoints\n")
 }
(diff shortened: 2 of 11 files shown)


ovirt-infra:

jeffvance:
This pr depends on #429 afaict

awels:
@jeffvance Doesn't look like it, this simply deprecates the file:/// and updates the unit tests to match (well maybe) and 429 is about the e2e tests.

jeffvance:
agree with @awels  above

jeffvance:
@davidvossel @awels @mhenriks Is there an http mock library that you find useful? I don't want to reinvent the wheel here. Thanks.

awels:
@jeffvance I was looking at splitting up the NewDataStream into two sections, one section that parses and creates the transport stream (http/s3/etc) and one section that creates the other readers. That way in the unit test you can manually create a DataStream, setup file stream as the transport, and then call the part that sets up the other readers.

Looking at the unit test code that is partially there. That way we don't 'support' file streams, but it is not impossible to use them for testing purposes.

awels:
Having worked on this a while, not sure my approach will work. I did find out golang has a net/http/httptest server build in, so we could use that.

awels:
@jeffvance I was trying something like this:
```go
ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    w.WriteHeader(http.StatusOK)
    f, err := os.Open(fileName)
    if err != nil {
        w.WriteHeader(http.StatusNotFound)
        return
    }
    defer f.Close()
    n, err := io.copy(w, f)
    if err != nil {
        w.WriteHeader(http.StatusNotFound)
    }
}))
```
But it doesn't seem to work, I keep ending up with a deadlock.

jeffvance:
@awels interesting. Thanks for trying. I'll keep working on it and I appreciate your time spent on this!

jeffvance:
@awels I suspect the `defer f.Close()` above. Don't we want to keep the file open until after the `io.Copy`?

awels:
@jeffvance Isn't the whole point of the defer that it doesn't get executed until the end of the code block, in this case the end of the function?

jeffvance:
> Isn't the whole point of the defer that it doesn't get executed until the end of the code block

Yes. My thinking was that we didn't want to close the file before we read from it, ie a copy operation. But, most of the unit tests don't actually read the endpoint's content...

awels:
@jeffvance my deadlock issue was that a lot of those tests expect file:// behavior and weren't reading the data since it works fine with a file://, however the http server is blocking waiting for data to be read from the stream, which a lot of the tests aren't doing. After I do a ds.Read(buf) it works a lot better.

jeffvance:
> ...a lot of those tests expect file:// behavior ...

Yes! This pr is a bit more complicated than I originally expected but I'm still making progress to address the many tests they rely on :file// endpoints.

awels:
I am actually also working on this, so if you want, you can stop

jeffvance:
But this pr is to deprecate the code for local endpoints. Why do we want/need an additional pr. This is almost working...

awels:
I wasn't going to make an additional PR, I will be using this one, with some updates.

jeffvance:
my 6th commit is for @awels since I did not know he was creating his own pr based on this one. I haven't pushed my local changes in a couple of days and now that I know this pr is being used by someone else, it's time for a refresh...

jeffvance:
replaced by #460

PR #444, opened 2018-09-07
CDI controller code dedupe

zvikorn:
This PR reduces the duplicated code in the clone, datavolume, and import controllers making maintenance easier.

Fixes #

**Release note**:

```release-note
NONE
```
**

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -2,52 +2,30 @@ package controller
 
 import (
 	"fmt"
-	"time"
-
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
-	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/workqueue"
-
 	"kubevirt.io/containerized-data-importer/pkg/common"
-	expectations "kubevirt.io/containerized-data-importer/pkg/expectations"
 )
 
 const (
 	//AnnCloneRequest sets our expected annotation for a CloneRequest
 	AnnCloneRequest = "k8s.io/CloneRequest"
-	// AnnCloneOf sets our expected annotation to inidcate where a clone originated from
+	//AnnCloneOf is used to indicate that cloning was complete
 	AnnCloneOf = "k8s.io/CloneOf"
-	// AnnCloningCreatedBy provides an annotation to indicate the object was created by the CDI controller
-	AnnCloningCreatedBy = "cdi.kubevirt.io/storage.cloningCreatedByController"
-	// AnnClonePodPhase provides an annotation to indicate the pod phase of the our CDI related pod
-	AnnClonePodPhase = "cdi.kubevirt.io/storage.clone.pod.phase"
-	// CloneUniqueID provides an annotation to uniquely identify the Clone object
+	//CloneUniqueID is used as a special label to be used when we search for the pod
 	CloneUniqueID = "cdi.kubevirt.io/storage.clone.cloneUniqeId"
-	// AnnTargetPodNamespace provides an annotation to indicate the namespace of the Target Pod in the CDI process
+	//AnnTargetPodNamespace is being used as a pod label to find the related target PVC
 	AnnTargetPodNamespace = "cdi.kubevirt.io/storage.clone.targetPod.namespace"
 )
 
 // CloneController represents the CDI Clone Controller
 type CloneController struct {
-	clientset                kubernetes.Interface
-	queue                    workqueue.RateLimitingInterface
-	pvcInformer, podInformer cache.SharedIndexInformer
-	pvcLister                corelisters.PersistentVolumeClaimLister
-	podLister                corelisters.PodLister
-	pvcsSynced               cache.InformerSynced
-	podsSynced               cache.InformerSynced
-	cloneImage               string
-	pullPolicy               string // Options: IfNotPresent, Always, or Never
-	verbose                  string // verbose levels: 1, 2, ...
-	podExpectations          *expectations.UIDTrackingControllerExpectations
+	Controller
 }
 
 // NewCloneController sets up a Clone Controller, and returns a pointer to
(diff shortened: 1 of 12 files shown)


zvikorn:
tarvis ci output:
0910 11:08:26.198029    1564 prlimit.go:103] faker [1  ] failed output is:
E0910 11:08:26.198522    1564 prlimit.go:104] 

What is that?

zvikorn:
Ready for review.

pkg/controller/util.go
```diff
@@ -24,45 +22,13 @@ const DataVolName = "cdi-data-vol"
...
 	if pvc.DeletionTimestamp != nil {
 		return false
 	}
-
-	// check if we have proper AnnEndPoint annotation
-	if !metav1.HasAnnotation(pvc.ObjectMeta, AnnEndpoint) {
-		glog.V(2).Infof("pvc annotation %q not found, skipping pvc \"%s/%s\"\n", AnnEndpoint, pvc.Namespace, pvc.Name)
+	// check if we have proper annotation
+	if !metav1.HasAnnotation(pvc.ObjectMeta, annotation) {
+		glog.V(3).Infof("pvc annotation %q not found, skipping pvc\n", annotation)
```
awels: Can you have this look like:
```go
glog.V(2).Infof("pvc annotation %q not found, skipping pvc \"%s/%s\"\n", AnnEndpoint, pvc.Namespace, pvc.Name)
```
Some of the e2e tests look for that string, this is causing the failure in the CI

pkg/controller/clone-controller.go
```diff
@@ -363,18 +163,71 @@ func (c *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
...
+	if pvc == nil {
+		return nil
+	}
+	//check if AnnoCloneRequest annotation exists
+	if !checkPVC(pvc, AnnCloneRequest) {
+		return nil
+	}
+	//checking for CloneOf annotation indicating that the clone was already taken care of by the provisioner (smart clone).
+	if metav1.HasAnnotation(pvc.ObjectMeta, AnnCloneOf) {
+		glog.V(3).Infof("pvc annotation %q exists indicating cloning completed, skipping pvc\n", AnnCloneOf)
```
awels: This needs the same format as the others, which include the name and namespace of the PVC.

awels:
```go
glog.V(2).Infof("pvc annotation %q not found, skipping pvc \"%s/%s\"\n", AnnCloneRequest, pvc.Namespace, pvc.Name)
```

zvikorn: Are there other places were I need to change this? what pattern should I look for? Is it only when we need to print the PVC details?

awels: Loook for "skipping pvc", the test code looks for "skipping pvc "<namespace>/<pvcname>". This is to make the 'dont do anything tests' faster, if there is a line in the log about skipping the pvc, we know it works.

j-griffith:
retest this please

zvikorn:
Travis fails on the following and I need to understand why:

--- FAIL: TestImportPodCreationExpectation (0.20s)
	controller_test.go:72: Expected
			testing.UpdateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"default", Verb:"update", Resource:schema.GroupVersionResource{Group:"", Version:"v1", Resource:"persistentvolumeclaims"}, Subresource:""}, Object:(*v1.PersistentVolumeClaim)(0xc4204a81a0)}
		got
			testing.CreateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"default"Verb:"create", Resource:schema.GroupVersionResource{Group:"", Version:"v1", Resource:"pods"}, Subresource:""}, Name:"", Object:(*v1.Pod)(0xc420224000)}

awels:
@zvikorn Looks like:
k8sActions := filterActions(f.kubeclient.Actions())
and
f.kubeactions
Don't match for some reason in the import test.

zvikorn:
It used to work though. was something changed?

awels:
Well this PR made a lot of changes to both the controller and the test, so its likely something changed. I haven't had a chance to locally apply this PR to take a closer look.

awels:
Okay so I looked at this a little more, and this test is failing:
```go
// Verifies pod creation does not occur when waiting for expectation.
func TestImportPodCreationExpectation(t *testing.T) {
	f := newImportFixture(t)
	pvc := createPvc("testPvc1", "default", map[string]string{AnnEndpoint: "http://test"}, nil)

	f.pvcLister = append(f.pvcLister, pvc)
	f.kubeobjects = append(f.kubeobjects, pvc)

	expPvc := pvc.DeepCopy()
	expPvc.ObjectMeta.Labels = map[string]string{CDILabelKey: CDILabelValue}
	f.expectUpdatePvcAction(expPvc)

	f.runWithExpectation(getPvcKey(pvc, t))
}
```
It appears that a POD is being created in syncPvc in the import-controller and thus this test is failing. Now I am not familiar enough with how all of this works to figure out why the Pod is being created, but it is.

zvikorn:
I saw that too. 
@davidvossel
Shouln't the pod be created because of the PVC?  
Could you please take a look and see why the test fails?

pkg/controller/import-controller_test.go
```diff
@@ -7,86 +7,45 @@ import (
...
-func (f *importFixture) runController(pvcName string,
+func (f *ImportFixture) runController(pvcName string,
 	startInformers bool,
 	expectError bool,
 	withCreateExpectation bool) {
-	c := f.newController()
+	c := f.newImportController()
 	if startInformers {
 		stopCh := make(chan struct{})
 		defer close(stopCh)
```
awels:
```go
cache.WaitForCacheSync(stopCh, c.pvcInformer.HasSynced)
cache.WaitForCacheSync(stopCh, c.pvcInformer.HasSynced)
```
Seems wrong as well, shouldn't we wait for Pvc AND Pod sync?

zvikorn: Don't know. This is the original code. @davidvossel ?

pkg/controller/import-controller.go
```diff
@@ -333,18 +127,62 @@ func (c *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
...
+	//check if AnnoEndPoint annotation exists
+	if !checkPVC(pvc, AnnEndpoint) {
+		return nil
+	}
+	glog.V(3).Infof("ProcessNextPvcItem: next pvc to process: %s\n", key)
+	return ic.processPvcItem(pvc)
+}
+
+//Run is being called from cdi-controller (cmd)
+func (ic *ImportController) Run(threadiness int, stopCh <-chan struct{}) error {
```
mhenriks: Looks like all the controllers have basically the same Run() method.  Maybe it makes sense to add a helper function in base Controller that takes `ic.runPVCWorkers` as a parameter?  Actually, maybe a `runPVCWorkers` equivalent can be common as well.  `ic.ProcessNextPvcItem()` is the real controller dependent function

pkg/controller/clone-controller.go
```diff
@@ -311,17 +111,17 @@ func (c *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 
 	// expectations prevent us from creating multiple pods. An expectation forces
 	// us to observe a pod's creation in the cache.
-	needsSync := c.podExpectations.SatisfiedExpectations(pvcKey)
+	needsSync := cc.podExpectations.SatisfiedExpectations(pvcKey)
 
 	// make sure not to reprocess a PVC that has already completed successfully,
 	// even if the pod no longer exists
-	phase, exists := pvc.ObjectMeta.Annotations[AnnClonePodPhase]
+	phase, exists := pvc.ObjectMeta.Annotations[AnnPodPhase]
```
mhenriks: Is this correct?

zvikorn: yes. The importer has the same annotation which can be common for both now.

pkg/controller/import-controller.go
zvikorn: :+1: Will move the run method. 
runPVCWorkers cannot get argument so I can't see how I go further...

pkg/controller/import-controller_test.go
davidvossel: yeah, i don't know what's going on with that :/ It doesn't look right though.

davidvossel:
TestImportPodCreationExpectation is failing because the refactor modified the expectation logic.

Previously, the importer controller "sets" the expectation and the clone controller "initialized then raised" the expectation per pod. Now, both controllers are "raising" the expectation. The problem is that the import controller never initializes the expectation for the key in use. This initialization used to automatically take place because we'd always "set" the expectation to 1 rather than raise it to 1. Since there was only ever a single pod for the import controller, we were essentially initializing the expectation to a value of 1 in a single step.

zvikorn:
Changed the import controller back to set the expectation and the test passed. Thanks you!

pkg/controller/controller.go
```diff
@@ -0,0 +1,252 @@
...
+	if !cache.WaitForCacheSync(stopCh, c.pvcInformer.HasSynced) {
+		return errors.New("Timeout waiting for pvc cache sync")
+	}
+	if !cache.WaitForCacheSync(stopCh, c.podInformer.HasSynced) {
+		return errors.New("Timeout waiting for pod cache sync")
+	}
+	glog.V(3).Infoln("Controller cache has synced")
+	for i := 0; i < threadiness; i++ {
+		//Go is not pure object oriented language. The command repetition below is a result of that.
+		switch t := controller.(type) {
```
awels: Am I losing it, or does this switch do nothing?
Wouldn't this be the same?
```go
t := controller.(type)
go wait.Until(t.runPVCWorkers, time.Second, stopCh)
```
I guess if the type isn't one of the two, but is that even possible?

zvikorn: Go does not let me do this... Spent half day to find a solution.
You cannot do what you wrote above as you get an error:
Use of .(type) outside type switch

awels: Oh I see, I missed the fact controller was an interface.

pkg/controller/clone-controller.go
```diff
@@ -363,18 +161,72 @@ func (c *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
...
 
-// forget the passed-in key for this event and optionally log a message.
-func (c *CloneController) forgetKey(key interface{}, msg string) bool {
-	if len(msg) > 0 {
-		glog.V(3).Info(msg)
+func (c *Controller) initializeExpectations(pvcKey string) error {
+	return c.podExpectations.SetExpectations(pvcKey, 0, 0)
+}
+
+func (c *Controller) expectPodCreate(pvcKey string) {
```
j-griffith: so just a though now that we noticed this... rather than have multiple functions with the same name that does different things; maybe we could/should implement these in the general controller still but give them meaningful names.  What I'm getting at is something like adding two functions in controller.go:

> func(*c controller)raisePodExpect()
> func(*c controller)createPodExpect()

That would have eliminated this sort of confusion, IMO if the functoin does different things it shouldn't have the same name :)

zvikorn:
You right. Can’t put them in controller as each does different thing but definitely change the expect name in cloner to raise.

PR #448, opened 2018-09-07
doc: Add storage class to example PVC

### LINKED ISSUE #221: add storageclass to golden-pvc example
I think it will be good to add the storageclass (marked as optional) to the example file
manifests/example/golden-pvc.yaml

it would be extra nice to have at least some of the available values there as well,
i keep forgetting how to mark glusterfs :)

aglitke:
Fixes #221.

### DIFF
--- a/manifests/example/golden-pvc.yaml
+++ b/manifests/example/golden-pvc.yaml
@@ -13,3 +13,5 @@ spec:
   resources:
     requests:
       storage: 10Gi
+  # Optional: Set the storage class or omit to accept the default
+  # storageClassName: local


j-griffith:
ci test please

j-griffith:
Hmm... that doesn't re-kick Travis?

awels:
No just STD-CI, I think the problem is that one of the tests actually use that yaml file, which now has a storage class that doesn't exist.

aglitke:
Interesting.  When I do a make cluster-up I get a storage class called 'local'

awels:
Right that is the 'local volume' storage class. But none of that exists in travis ci.

PR #456, opened 2018-09-14
Have DataVolume clones follow the PVC status.

### LINKED ISSUE #455: DataVolume clone source doesn't set data volume phase properly.

>
/kind bug
> /kind enhancement

**What happened**:
I have a working source PVC with some data in it.
I create a DV with source for instance like this:
```yaml
      source:
        pvc:
          name: template-test-dv-01
          namespace: default
```
It starts to clone the source PVC into my new DV and the cloning successfully completes. However it never updates the status of the DV to be Successful.
The phase to be Successful instead of 'PVCBound', since everything completed successfully.

Create a PVC called test-pvc in the default name space, and populate some data in it.
Then create a DV with this yaml
```yaml
kind: DataVolume
metadata:
  name: "template-test-dv-01"
spec:
    source:
      pvc:
        name: test-pvc
        namespace: default
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: "128Mi"
```
The process will complete but the status phase on the DV will remain PVCBound

The functional tests did not catch this issue because it tests the contents of the target PVC (which are correct) and doesn't test for the phase of the DV.

**Environment**:
latest master
1.10.4
- Install tools:
- Others:

awels:
This PR fixes an issue where the Data Volume Phase was not being updated properly where the source was a PVC instead of HTTP

Fixes #455

**Release note**:

```release-note
NONE
```
**


pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -101,6 +101,12 @@ const (
 	// ImportInProgress represents a data volume with a current phase of ImportInProgress
 	ImportInProgress DataVolumePhase = "ImportInProgress"
 
+	// CloneScheduled represents a data volume with a current phase of ImportScheduled
```
aglitke: CloneScheduled instead of ImportScheduled.

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -101,6 +101,12 @@ const (
...
+	CloneScheduled DataVolumePhase = "CloneScheduled"
+
+	// CloneInProgress represents a data volume with a current phase of ImportInProgress
+	CloneInProgress DataVolumePhase = "CloneInProgress"
```
aglitke: CloneInProgress?

awels:
@j-griffith What additions are we talking about? This fixes a bug where it was looking at the wrong annotations, and thus not switching to the right phases. The phase constants didn't exist because of the bug. Is that what you mean?

pkg/apis/datavolumecontroller/v1alpha1/types.go
awels: Yeah you are right. See how easy it is to copy and paste and forget something. We should find a way to unify these into a single constant for all transfer type operations (import/clone/upload/whatever)

pkg/apis/datavolumecontroller/v1alpha1/types.go
awels: Fixed.

j-griffith:
@awels 

> What additions are we talking about? This fixes a bug where it was looking at the wrong annotations, and thus not switching to the right phases. The phase constants didn't exist because of the bug. Is that what you mean?

Yes, even though it's a bug fix it introduces those new phases.  I'm not sure if matters or if we have any sort of policy on this, but seeing a new constant in the API package it seems like it might be worth noting.

awels:
Yeah we probably need to give some thought on if we want to unify all those phases/annotations/etc into a single set usable by all forms of data transfer, so we don't have to keep duplicating them. But I will update the notes to include those new phases.

j-griffith:
@zvikorn I was thinking about that based on the PodPhase change in your patch.  I think it's something we can/should look at, but I didn't think it was appropriate for this bug fix.  Seems like a great follow up improvement though.

PR #460, opened 2018-09-19
deprecation of local endpoint support

### LINKED ISSUE #338: deprecate import of local (file://) endpoints
This is **deprecation** only. Actual removal of local (file://) endpoint support will be in a forthcoming issue.
See pr #341

@nellyc @aglitke @davidvossel @erinboyd

awels:
- Removed file:// end points from code base.
- Updated unit tests to use httptest server instead of file://
- Refactored data stream unit tests to use ginko/gomega

This PR removes all references to file:// URIs and removes support for those types of end points.

Fixes #338

**Release note**:

```release-note
Support for file:// URIs has officially been removed.
```


jeffvance:
reviewed all files except _pkg/importer/dataStream_test.go_, which I'll review tomorrow.
So far it lgtm.

awels:
I am going to wait with merging until after the upload PR is merged as this one conflicts with it.

pkg/lib/size/size.go
```diff
@@ -1,18 +0,0 @@
...
-
-import (
-	"github.com/pkg/errors"
-
-	"kubevirt.io/containerized-data-importer/pkg/importer"
-)
-
-// Size returns the size in bytes of the provided endpoint. If the endpoint was archived, compressed or
-// converted to qcow2 the original image size is returned.
-func Size(endpoint, accessKey, secKey string) (int64, error) {
```
mhenriks: Not sad to see this stuff go.  But it is referenced in the README.  So should probably update that.

jeffvance: Yes, readme needs updating.
@aglitke @davidvossel I don't recall who wanted the Size() func. It's fine to deprecate it but I think some notification would be a good idea. Eg let QE and kubevirt-ansible know of this api change.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/image"
+	"kubevirt.io/containerized-data-importer/pkg/util"
 	"kubevirt.io/containerized-data-importer/tests/utils"
 )
 
 const TestImagesDir = "../../tests/images"
 
+var imageDir, _ = filepath.Abs(TestImagesDir)
+var cirrosData, _ = readFile(filepath.Join(imageDir, "cirros-qcow2.img"))
```
jeffvance: Is there any concern that the test container may run out of memory due to reading entire images into variables?

awels: The image are 13 and 18M and we are manipulating and reading them tons of times during the test run. So if an extra 31M of memory is a concern, we have bigger issues.

pkg/lib/size/size.go
awels: I will update the readme. I looked hard for anything that was actually referencing this code, and the only place I could find was the test for it. IIRC the original intent was for it to be used in some of the controllers to figure out the size for some operations (maybe report progress, who knows) but its a tiny wrapper around the data stream Size field, so there is not much use to it at this point.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+			By("Reading data from buffer")
+			ds.Read(resultBuffer)
+			Expect(ds.accessKeyID).To(Equal(accessKeyID))
+			Expect(ds.secretKey).To(Equal(secretKey))
+			Expect(ds.qemu).To(Equal(qemu))
+			Expect(reflect.DeepEqual(resultBuffer, want)).To(BeTrue())
+		} else {
+			Expect(err).To(HaveOccurred())
+		}
+	},
```
jeffvance: where are the Readers closed?

awels: Good point, they aren't being closed and should, let me add that.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
-	return NewFakeQEMUOperations(err, err, err)
-}
+	table.DescribeTable("can build selector", func(image, ep string, wantErr bool) {
+		if ep == "" {
+			ep = ts.URL
+		}
+		ds, err := NewDataStream(ep+"/"+image, "", "")
+		if !wantErr {
+			Expect(err).NotTo(HaveOccurred())
+			defer ds.Close()
```
jeffvance: I think you only want to close the reader when err==nil. Maybe something like this:

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
-			wantErr: true,
-		},
-		{
-			name:    "expect new DataStream to fail with invalid or missing endpoint",
-			args:    args{"", "", ""},
-			want:    nil,
-			wantErr: true,
-		},
+func emptyBuffer(ds *DataStream) {
+	buf := make([]byte, 32*1024)
```
jeffvance: If the point of this func is to reach eof on all readers then I think you need a larger buf. Otherwise, I am not sure why you are reading 32k worth of the image? A comment would be helpful.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+		if !wantErr {
+			Expect(err).NotTo(HaveOccurred())
+			defer ds.Close()
+			emptyBuffer(ds)
+			err = ds.dataStreamSelector()
+			Expect(err).NotTo(HaveOccurred())
+		} else {
+			Expect(err).To(HaveOccurred())
+		}
+	},
```
jeffvance: Should there be a check that the correct scheme was returned? I guess for now it's always http but maybe later s3 will be added?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+		table.Entry("success building selector for valid endpoint and file", "tinyCore.iso", "", false),
+		table.Entry("fail trying to build selector for invalid http endpoint", "", "http://www.google.com", true),
+		table.Entry("fail trying to build invalid selector", "tinyCore.iso", "fake://somefakefile", true),
+	)
+
+	table.DescribeTable("can construct readers", func(outfile, filename string, numRdrs int, wantErr bool) {
+		fmt.Println(filename)
+		By(fmt.Sprintf("Creating new fileserver for %s and file %s", filepath.Dir(filename), filepath.Base(filename)))
+		tempTestServer := createTestServer(filepath.Dir(filename))
```
jeffvance: so now there are 2 test servers running? One here and one in the BeforeEach

pkg/importer/dataStream_test.go
awels: Which is what it is doing, if the Expect fails then err != nil and I don't want to close the reader

pkg/importer/dataStream_test.go
awels: See the for loop below here? It runs until the readers are exhausted, I am not actually interested in the content.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+
+		if !wantErr {
+			Expect(err).NotTo(HaveOccurred())
+			defer ds.Close()
+			for _, r := range ds.Readers {
+				fmt.Fprintf(GinkgoWriter, "INFO: Reader type: %d\n", r.rdrType)
+			}
+			Expect(len(ds.Readers)).To(Equal(numRdrs))
+		} else {
+			Expect(err).To(HaveOccurred())
```
jeffvance: If we want an error and there was no err then we should still close the readers, imo

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+		ds, err := NewDataStream(tempTestServer.URL+"/"+filepath.Base(filename), "", "")
+		defer func() {
+			tempTestServer.Close()
+		}()
+
+		if !wantErr {
+			Expect(err).NotTo(HaveOccurred())
+			defer ds.Close()
+			for _, r := range ds.Readers {
+				fmt.Fprintf(GinkgoWriter, "INFO: Reader type: %d\n", r.rdrType)
```
jeffvance: logging the readers is nice. I think this should be done whenever NewDataStream returns nil err. Separately you can check if an err was wanted or not...

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+			Expect(err).To(HaveOccurred())
+		}
+	},
+		table.Entry("success building selector for valid endpoint and file", "tinyCore.iso", "", false),
+		table.Entry("fail trying to build selector for invalid http endpoint", "", "http://www.google.com", true),
+		table.Entry("fail trying to build invalid selector", "tinyCore.iso", "fake://somefakefile", true),
+	)
+
+	table.DescribeTable("can construct readers", func(outfile, filename string, numRdrs int, wantErr bool) {
+		fmt.Println(filename)
```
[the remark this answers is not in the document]
awels: Sorry that was leftover debug code.

pkg/importer/dataStream_test.go
awels: Correct, I can't use the one in the before each because the directories change that we are using for the server based on which iteration of the test we are running.

pkg/importer/dataStream_test.go
awels: We can't since the ds variable will be nil, and thus there are no readers.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,409 @@ import (
...
+			defer ds.Close()
+			for _, r := range ds.Readers {
+				fmt.Fprintf(GinkgoWriter, "INFO: Reader type: %d\n", r.rdrType)
+			}
+			Expect(len(ds.Readers)).To(Equal(numRdrs))
+		} else {
+			Expect(err).To(HaveOccurred())
+		}
+	},
+		table.Entry("successfully construct a xz reader", "tinyCore.iso.xz", testfiles[".xz"], 5, false),              // [http, multi-r, xz, multi-r]
```
jeffvance: +1, nice approach here!

pkg/importer/dataStream_test.go
jeffvance: agree. May become an issue if we test much larger (or many) images...

pkg/importer/dataStream_test.go
awels: If err != nil, then ds will be nil (look at implementation of NewDataStream) and thus we should stop processing stuff (this is why the Expect is before anything else)

pkg/importer/dataStream_test.go
awels: I would argue we shouldn't do that in unit tests, that would be an e2e test.

pkg/importer/dataStream_test.go
awels: Unless we can somehow fake an S3 server, I think its always going to be http, and thus is not very interesting to test for (since we add ts.URL to the ep if its empty)

pkg/importer/dataStream_test.go
awels: Thanks the original tests were all in table form already, I just converted them to the DescribeTable versions of gingko

pkg/importer/dataStream_test.go
jeffvance: I'm also thinking of the case where wantErr == true and NewDataStream returns no err (and thus may have created the readers).
Maybe the best way to ensure reader are close would be:
```
ds, err := NewDataStream()
if ds != nil && len(ds.Readers) > 0 {
      defer ds.Close()
}

```

pkg/importer/dataStream_test.go
jeffvance: It looks like minio has a mock s3 server so this could happen...

pkg/importer/dataStream_test.go
jeffvance: I'm describing a case where we want an err but NewDataStream returns a ds and no err. In that situation readers may exist and won't be closed. See my refactor above where the block is decided by err != nil rather than wantErr...

pkg/importer/dataStream_test.go
jeffvance: Yes I see that. But what is the point in "exhausting" the readers and what does that actually mean. I was thinking you wanted to reach EOF but maybe it is something else. Is the hard-coded 32k buf a magic size (ie related to the biggest buf a reader could have? I assume this func is needed so you can call NewDataStream more than once and have a fresh image to process each time. There is the concept of _resetting_ a file which basically moves the current record ptr to 0. Is this the idea here?

awels: Hmm okay, well I must have been doing something wrong in my earlier experiments because the httptest server would just hang waiting for something to read the stream, so I had to explicitly read the stream to make it work. But I since switched to the file server and it is no longer needed.

But originally it was just to simply read the stream so the server wouldn't lock waiting for something to read the stream.

pkg/importer/dataStream_test.go
awels: Right, I will add that.

pkg/importer/dataStream_test.go
jeffvance: @awels I'm not saying you are doing something wrong here. More that I don't understand it...

pkg/importer/dataStream_test.go
awels: If you do it your way, you are simply adding an extra layer of checks that the Expect is already doing. That is why I reversed it and made it driven by wantErr, this is possible due to the mutually exclusiveness of ds and err. One of them is nil while the other is not, there is no scenario where both or neither are nil.

pkg/importer/dataStream_test.go
awels: We probably should put the closing logic in the AfterEach regardless of success or failure.

pkg/importer/dataStream_test.go
awels: The problem is, in the BeforeEach I don't know the directory which we will be using to serve the images from (could be a temporary directory that created some versions of the file we want).

pkg/importer/dataStream_test.go
awels: The Readers are now always closed in the AfterEach block

pkg/importer/dataStream_test.go
awels: We are still investigating how we can do S3 tests, I don't think the original tests that I copied checked for scheme either.

jeffvance: @awels you are correct.

pkg/importer/dataStream_test.go
jeffvance: I don't see my own refactor any longer... did it get deleted somehow?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
-	if replacement != nil {
-		qemuOperations = replacement
-		defer func() { qemuOperations = orig }()
+	AfterEach(func() {
+		By("[AfterEach] closing test server")
+		ts.Close()
+	})
+
+	table.DescribeTable("Image, with import source should", func(dest, endpt string, qemuOperations image.QEMUOperations, wantErr bool) {
+		By("Configuring end point")
```
jeffvance: usually one word, _endpoint_, right?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+		if !strings.HasPrefix(endpt, "http") {
+			endpt = ts.URL + "/" + endpt
+		}
+		By(fmt.Sprintf("end point is %s", endpt))
+		defer os.RemoveAll(dest)
```
jeffvance: so this is equivalent to `rm -r /tmp/cdi-test-copy` for the 1st and 2nd test cases, and `rm -r ./cirros-qcow2.raw` for the 3rd, 4th, and 5th test cases. Unless the rm -r target is a a dir I think it's clearer to use `os.Remove`.

jeffvance: I guess since the `os.RemoveAll` err is ignored it's ok to call this even when `CopyImage` fails and there is no destination file...

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+		})
+	},
+		table.Entry("successfully copy local image", filepath.Join(os.TempDir(), "cdi-testcopy"), "tinyCore.iso", NewQEMUAllErrors(), false),
+		table.Entry("expect failure trying to copy non-existing local image", filepath.Join(os.TempDir(), "cdi-testcopy"), "tinyCoreBad.iso", NewQEMUAllErrors(), true),
+		table.Entry("successfully copy streaming image", "cirros-qcow2.raw", "cirros-qcow2.img", NewFakeQEMUOperations(errors.New("should not be called"), nil, nil), false),
+		table.Entry("streaming image qemu validation fails", "cirros-qcow2.raw", "cirros-qcow2.img", NewFakeQEMUOperations(nil, nil, errors.New("invalid image")), true),
+		table.Entry("streaming image qemu convert fails", "cirros-qcow2.raw", "cirros-qcow2.img", NewFakeQEMUOperations(nil, errors.New("exit 1"), nil), true),
+	)
+
+	table.DescribeTable("internal copy", func(r io.Reader, out string, qemu bool, qemuOperations image.QEMUOperations, wantErr bool) {
```
jeffvance: Not sure where to put this comment but, imo, I think the `isHTTPQcow2` block in `*DataStream.copy` should  be moved to `func copy()`. The reason being that the lowest level copy func, which is `func copy()` should contain all of the logic to do the copy which includes qcow2 streaming.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
 )
 
 const TestImagesDir = "../../tests/images"
 
+var imageDir, _ = filepath.Abs(TestImagesDir)
+var cirrosData, _ = readFile(filepath.Join(imageDir, "cirros-qcow2.img"))
+var tinyCoreData, _ = readFile(filepath.Join(imageDir, "tinyCore.iso"))
+var testfiles = createTestData()
+var stringRdr = strings.NewReader("test data for reader 1")
+var rdrfile, _ = os.Open(filepath.Join(imageDir, "cirros-qcow2.img"))
```
jeffvance: I think this var name is too generic since it always contains only a qcow2 reader. Maybe `qcow2Rdr` is more descriptive?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+		readers []reader
 	}
-	f()
-}
 
-func createDataStream(ep, accKey, secKey string) *DataStream {
-	dsurl, _ := ParseEndpoint(ep)
+	rdrs1 := ioutil.NopCloser(strings.NewReader("test data for reader 1"))
+	rdrs2 := ioutil.NopCloser(strings.NewReader("test data for reader 2"))
+	rdrA := reader{4, rdrs1}
```
jeffvance: better to not hard-code the RdrType values.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+	const numbyte = 8
 
-func createDataStreamBytes(testfile, ep string, defaultBuf, singlereader bool) (*DataStream, []byte, error) {
-	urlPath := filepath.Join(ep, testfile)
-	if len(ep) == 0 {
-		urlPath = getURLPath(testfile)
-	}
-	testFilePath := getTestFilePath(testfile)
-	ds := createDataStream(urlPath, "", "")
+	It("create expected random name", func() {
```
jeffvance: This test is really only testing that the random name is 8 bytes in length. It will always pass unless the length of random name used in `randTmpName` changes.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
 
-	return ds, testBytes, nil
-}
+var _ = Describe("Streaming Data Conversion", func() {
 
-func getFileSize(testfile string) (int, error) {
-	f, err := os.Open(getTestFilePath(testfile))
-	defer f.Close()
+	tinyCoreIso := "tinyCore.iso"
+	baseTestImage, err := filepath.Abs(filepath.Join(TestImagesDir, tinyCoreIso))
```
jeffvance: Could use `imageDir` instead of `TestImageDir` and not need to call `Abs()`

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
-func getURLPath(testfile string) string {
-	// CWD is set within go test to the directory of the test
-	// being executed, so using relative path
-	imageDir, _ := filepath.Abs(TestImagesDir)
-	return "file://" + imageDir + "/" + testfile
-}
+	var tmpTestDir string
+	BeforeEach(func() {
+		tmpTestDir = testDir(os.TempDir())
+		if err != nil {
```
jeffvance: what err is being checked here?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+			Fail(fmt.Sprintf("[BeforeEach] Failed created test dir: %v\n", err))
+		}
+		By(fmt.Sprintf("[BeforeEach] Creating temporary dir %s", tmpTestDir))
+		syscall.Umask(0000)
```
jeffvance: not sure this is needed. Were you getting perms errors in your testing?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
-		break
+		ds, err := NewDataStream(tempTestServer.URL+"/"+testBase, "", "")
+
+		Expect(err).NotTo(HaveOccurred())
+		defer ds.Close()
+		Expect(ds.Size).To(Equal(sourceSize))
+
+		fmt.Fprintf(GinkgoWriter, "End test on test file %q\n", testSample)
+	},
+		table.Entry("should decompress gzip", baseTestImage, false, []string{image.ExtGz}),
```
jeffvance: instead casting to `[]string` for each row, consider making this last arg a variadic string.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+		sourceSize := finfo.Size()
+		fmt.Fprintf(GinkgoWriter, "INFO: size = %d\n", sourceSize)
+
+		By(fmt.Sprintf("Converting source file to format: %s", expectFormats))
+		// Generate the expected data format from the random bytes
+		testSample, err := utils.FormatTestData(originalFile, tmpTestDir, expectFormats...)
+		Expect(err).NotTo(HaveOccurred(), "Error formatting test data.")
+		fmt.Fprintf(GinkgoWriter, "INFO: converted source file name is %q, in dir %q\n", testSample, tmpTestDir)
+
+		tempTestServer := createTestServer(filepath.Dir(testSample))
```
jeffvance: why is a new http file server needed for each iteration of this test?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+	AfterEach(func() {
+		By(fmt.Sprintf("[AfterEach] Cleaning up temporary dir %s", tmpTestDir))
+		os.RemoveAll(tmpTestDir)
+	})
 
-	go func() {
-		if err := server.ListenAndServe(); err != nil {
-			// cannot panic, because this probably is an intentional close
-			log.Printf("Httpserver: ListenAndServe() error: %s", err)
+	table.DescribeTable("when data is in a supported file format", func(originalFile string, useVirtSize bool, expectFormats []string) {
```
jeffvance: seems like this test case is more than a unit test of a single main func. Can size be tested w/o calling Copy or CopyImage? For all tar files and qcow2 files the answer is yes -- the ds.Size field should work. But for iso files (including .gz and .xz) the answer is no AFAIK.

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
-	for i := 0; i < 10; i++ {
-		resp, err := http.Get(fmt.Sprintf("http://localhost:%d/", port))
-		if err != nil {
-			time.Sleep(500 * time.Millisecond)
-			continue
+		fmt.Fprintf(GinkgoWriter, "INFO: byte size = %d\n", targetSize)
+
+		By(fmt.Sprintf("Calling `size.Size()` on same endpoint %q", tempTestServer.URL))
+		// extract the file extension(s) and check if file should be skipped
+		testBase := filepath.Base(testSample)
```
jeffvance: Why not move this line up to 281.1 to save a call to Base()?

pkg/importer/dataStream_test.go
```diff
@@ -7,416 +7,404 @@ import (
...
+		//table.Entry("should unpack .tar.xz", baseTestImage, false, []string{image.ExtTar, image.ExtXz}),
+		table.Entry("should convert .qcow2", baseTestImage, true, []string{image.ExtQcow2}),
+		table.Entry("should convert and unpack .qcow2.gz", baseTestImage, false, []string{image.ExtQcow2, image.ExtGz}),
+		table.Entry("should convert and unpack .qcow2.xz", baseTestImage, false, []string{image.ExtQcow2, image.ExtXz}),
+		table.Entry("should convert and untar .qcow2.tar", baseTestImage, false, []string{image.ExtQcow2, image.ExtTar}),
+		table.Entry("should convert and untar and unpack .qcow2.tar.gz", baseTestImage, false, []string{image.ExtQcow2, image.ExtTar, image.ExtGz}),
+		table.Entry("should convert and untar and unpack .qcow2.tar.xz", baseTestImage, false, []string{image.ExtQcow2, image.ExtTar, image.ExtXz}),
+		table.Entry("should pass through unformatted data", baseTestImage, false, []string{""}),
+	)
+})
```
jeffvance: nicely done!

pkg/importer/dataStream_test.go
awels: Done

pkg/importer/dataStream_test.go
awels: I never did see a refactor, but I guess your issue is with the wantErr driving the flow of the test instead of 'err'. Logically there is no difference between the two. The only potential issue was that the readers might not be closed in certain flows (which I solved by always calling close in the AfterEach). The reason I prefer the wantErr driving the test is that I can use the Expects to validate the values of err. These are logically the same:

```go
if err != nil {
  if !wantErr {
    Expect(Err).ToNot(HaveOccured)) <-- is redundant at this point, since we know err != nil
  }
  ...
} else {
  if wantErr {
    Expect(Err).To(HaveOccured)) <-- is redundant at this point, since we know err == nil
  }
  ...
}
```
vs
```go
if !wantErr {
  Expect(Err).ToNot(HaveOccured))
  ...
} else {
  Expect(Err).To(HaveOccured))
  ...
}
```
Both are the same logically, and since we know ds == nil is always err != nil and vice versa, having wantErr drive the test is identical to err driving the test.

pkg/importer/dataStream_test.go
awels: Not sure English is not my first language, but sure. I also notice we don't use endpoint in variable names consistently, I see ep, endpt, and endpoint in various places.

pkg/importer/dataStream_test.go
awels: Done

pkg/importer/dataStream_test.go
awels: I copied this directly out of the original file, not sure why it is needed.

pkg/importer/dataStream_test.go
awels: Not sure, no error being returned.

pkg/importer/dataStream_test.go
awels: Yeah, was like that in the original file, updated to use the constants instead.

pkg/importer/dataStream_test.go
awels: I copied this code verbatim from the original file, didn't inspect it too much, but I don't see a reason to use tmp if not needed (might not even have tmp in certain environments). Also switched to os.Remove

pkg/importer/dataStream_test.go
awels: Done

pkg/importer/dataStream_test.go
awels: httptest servers are cheap to start (I am creating a new one for each test in previous tests as well)

pkg/importer/dataStream_test.go
awels: Done

pkg/importer/dataStream_test.go
awels: Done

pkg/importer/dataStream_test.go
awels: This particular test was in the datastream_ginko_test file originally, so I am guessing it used to be what you guys called a functional test, so yes it is more than a single function test.

awels:
ci test please

awels:
ci test please

PR #478, opened 2018-10-03
CDI - Upload Controller Dedupe

zvikorn:

CDI upload controller dedupe, code refactoring based on https://github.com/kubevirt/containerized-data-importer/issues/235
Fixes #

**Release note**:

```release-note
   NONE
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -7,8 +7,6 @@ import (
 )
 
 // Common types and constants used by the importer and controller.
-// TODO: maybe the vm cloner can use these common values
-
 const (
 	// CDILabelKey provides a constant for CDI PVC labels
 	CDILabelKey = "app"
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"kubevirt.io/containerized-data-importer/pkg/common"
+	"reflect"
 )
 
 const (
@@ -22,6 +23,8 @@ const (
 	//AnnTargetPodNamespace is being used as a pod label to find the related target PVC
 	AnnTargetPodNamespace = "cdi.kubevirt.io/storage.clone.targetPod.namespace"
 )
+
+var cloner = "cloner"
 
 // CloneController represents the CDI Clone Controller
 type CloneController struct {
@@ -37,7 +40,7 @@ func NewCloneController(client kubernetes.Interface,
 	pullPolicy string,
 	verbose string) *CloneController {
 	c := &CloneController{
-		Controller: *NewController(client, pvcInformer, podInformer, image, pullPolicy, verbose),
+		Controller: *NewController(client, pvcInformer, podInformer, image, pullPolicy, verbose, cloner),
 	}
 	return c
 }
@@ -226,7 +229,14 @@ func (cc *CloneController) Run(threadiness int, stopCh <-chan struct{}) error {
 	return nil
 }
 
-func (cc *CloneController) runPVCWorkers() {
+//RunWorkers is declared in Controller WorkerRunner interface
+func (cc *CloneController) RunWorkers() {
 	for cc.ProcessNextPvcItem() {
 	}
+}
+
+//WaitForCacheSync is declared in Controller WorkerRunner interface
+func (cc *CloneController) WaitForCacheSync(stopCh <-chan struct{}, c *Controller, v reflect.Value, controller interface{}) error {
+	waitForCacheSync("CloneController", stopCh, c, v, controller, nil)
+	return nil
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"kubevirt.io/containerized-data-importer/pkg/expectations"
+	"reflect"
 	"time"
 )
 
@@ -36,6 +37,7 @@ type Controller struct {
 	pullPolicy               string // Options: IfNotPresent, Always, or Never
 	verbose                  string // verbose levels: 1, 2, ...
 	podExpectations          *expectations.UIDTrackingControllerExpectations
+	ctlType                  string
 }
 
 //NewController is called when we instantiate any CDI controller.
@@ -44,7 +46,8 @@ func NewController(client kubernetes.Interface,
 	podInformer coreinformers.PodInformer,
 	image string,
 	pullPolicy string,
-	verbose string) *Controller {
+	verbose string,
+	ctlType string) *Controller {
 	c := &Controller{
 		clientset:       client,
 		queue:           workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
@@ -57,14 +60,15 @@ func NewController(client kubernetes.Interface,
 		image:           image,
 		pullPolicy:      pullPolicy,
 		verbose:         verbose,
+		ctlType:         ctlType,
 		podExpectations: expectations.NewUIDTrackingControllerExpectations(expectations.NewControllerExpectations()),
 	}
 
 	// Bind the pvc SharedIndexInformer to the pvc queue
 	c.pvcInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: c.enqueuePVC,
+		AddFunc: c.enqueueObject,
 		UpdateFunc: func(old, new interface{}) {
-			c.enqueuePVC(new)
+			c.enqueueObject(new)
 		},
 	})
 
@@ -89,7 +93,14 @@ func NewController(client kubernetes.Interface,
 }
 
 func (c *Controller) handlePodAdd(obj interface{}) {
-	c.handlePodObject(obj, "add")
+	switch c.ctlType {
+	case "uploder":
+		c.handleObject("")
+		break
+	default:
+		c.handlePodObject(obj, "add")
+	}
+
 }
 func (c *Controller) handlePodUpdate(obj interface{}) {
 	c.handlePodObject(obj, "update")
@@ -100,6 +111,10 @@ func (c *Controller) handlePodDelete(obj interface{}) {
 
 func (c *Controller) observePodCreate(pvcKey string) {
 	c.podExpectations.CreationObserved(pvcKey)
+}
+
+func (c *Controller) handleObject(obj interface{}) {
+	c.handlePodObject(obj, "")
 }
 
 func (c *Controller) handlePodObject(obj interface{}, verb string) {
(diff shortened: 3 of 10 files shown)


pkg/common/common.go
```diff
@@ -78,4 +80,27 @@ const (
 
 	// DefaultResyncPeriod sets a 10 minute resync period, used in the controller pkg and the controller cmd executable
 	DefaultResyncPeriod = 10 * time.Minute
+
```
j-griffith: Thanks for adding these, I don't know if we need the full blown mapping or not; for example I had just added a new member to the base controller "ctrlType: string" and variables (not constants even though we LOVE constants) for importer, cloner, uploader.  That way the NewController function would just expect that to be set, and it's super easy to access (c.ctrlType).

The only thing I don't like about your approach is using this common.go file (it can become a dumping ground again rather quickly) and exporting everything.  Not a big deal at all but just something to throw out there if you're interested in considering it.

pkg/common/common.go
```diff
@@ -78,4 +80,27 @@ const (
...
+	Importer ControllerType = 0
+	//Cloner enum for ControllerType
+	Cloner ControllerType = 1
+	//Uploader enum for ControllerType
+	Uploader ControllerType = 2
+	//Unknown enum for ControllerType
+	Unknown ControllerType = 3
 )
+
+func (controller ControllerType) getControllerName() string {
```
j-griffith: If we just modify the Controller to have a member that points to this func in the main controller class it might be even easier?  Or just use a string to begin with, the only downside to that is when I modified all the logging it does get a little noisy.

pkg/controller/controller.go
```diff
@@ -36,6 +38,7 @@ type Controller struct {
 	pullPolicy               string // Options: IfNotPresent, Always, or Never
 	verbose                  string // verbose levels: 1, 2, ...
 	podExpectations          *expectations.UIDTrackingControllerExpectations
+	ownerController          common.ControllerType
```
j-griffith: So maybe just something like:
var (ImportController = "importer",
       CloneController = "cloner",
       UploadController = "upload",
       ...)
Then just add:
    "controllerType: string"
and pass that in when calling NewController.

If we're really concerned about the variable then it can be a constant, that's fine.  The reason this matters though IMO is if we don't need 3 locations and 3 data structures to handle this in multiple files we should try and keep it as simple as we can (ie stuff it all into the controller package, that's the only place it should be appropriate for the most part).

pkg/controller/controller.go
```diff
@@ -88,8 +93,21 @@ func NewController(client kubernetes.Interface,
 	return c
 }
 
+//NewEmptyController is an empty constructor
+func NewEmptyController() *Controller {
```
j-griffith: Do we really need this helper?  I see you're using it in the test but I can't quite figure out why the test can't just do "c := &Controller{}" on it's own?

pkg/controller/controller.go
```diff
@@ -151,45 +173,84 @@ func (c *Controller) handlePodObject(obj interface{}, verb string) {
...
 		runtime.HandleError(err)
 		return
 	}
 	c.queue.AddRateLimited(key)
 }
 
+func waitForCacheSync(controllerName string, stopCh <-chan struct{}, c *Controller, v reflect.Value, controller interface{},
+	uploadController *UploadController) error {
+
+	glog.V(3).Infof("starting %s Run loop", controllerName)
```
j-griffith: Nice!  This is helpful IMO

pkg/controller/controller.go
```diff
@@ -151,45 +173,84 @@ func (c *Controller) handlePodObject(obj interface{}, verb string) {
...
 	if threadiness < 1 {
 		return errors.Errorf("expected >0 threads, got %d", threadiness)
 	}
 
-	if !cache.WaitForCacheSync(stopCh, c.pvcInformer.HasSynced) {
-		return errors.New("Timeout waiting for pvc cache sync")
-	}
-	if !cache.WaitForCacheSync(stopCh, c.podInformer.HasSynced) {
-		return errors.New("Timeout waiting for pod cache sync")
+	v := reflect.ValueOf(controller)
```
j-griffith: That's a cool way to do the type check, I usually run away from "reflect" but it's a good use for it here I think.

pkg/controller/controller.go
```diff
@@ -151,45 +173,84 @@ func (c *Controller) handlePodObject(obj interface{}, verb string) {
...
-			go wait.Until(t.runPVCWorkers, time.Second, stopCh)
+		if v.Type() == reflect.TypeOf(importController) {
+			ownerController := controller.(*ImportController)
+			go wait.Until(ownerController.runPVCWorkers, time.Second, stopCh)
+		} else if v.Type() == reflect.TypeOf(cloneController) {
+			ownerController := controller.(*CloneController)
+			go wait.Until(ownerController.runPVCWorkers, time.Second, stopCh)
+		} else if v.Type() == reflect.TypeOf(uploadController) {
+			ownerController := controller.(*UploadController)
+			go wait.Until(ownerController.runWorker, time.Second, stopCh) //<------ change runWorker to runPVCWorkers
```
j-griffith: If nothing else I like that this makes the code a bit more explicit so you don't need your comment explaining why the seemingly same calls are repeated like you had to do in the old code.  I'm curious though, is there a reason you switched from using a switch to the if/else-if format here?

mhenriks: Rather than doing this type switching in a 'base' class, I think it would be much cleaner for the derived or 'super' controllers to implement a set of interfaces that the base can work with.  For example, the above code would turn into:

```
type WorkerRunner {
    RunWorkers()
}

// super controllers implement RunWorkers() (can prob jest rename existing functions)

ownerController, ok := controller.(WorkerRunner)
// maybe the super controller doesn't want to implement this method for some reason
if ok {
    go wait.Until(ownerController.RunWorkers, time.Second, stopCh)
}

```

zvikorn:
ci test please

pkg/controller/clone-controller.go
```diff
@@ -37,7 +38,7 @@ func NewCloneController(client kubernetes.Interface,
 	pullPolicy string,
 	verbose string) *CloneController {
 	c := &CloneController{
-		Controller: *NewController(client, pvcInformer, podInformer, image, pullPolicy, verbose),
+		Controller: *NewController(client, pvcInformer, podInformer, image, pullPolicy, verbose, "cloner"),
```
j-griffith: We should probably create a variable set for these, in Controller you have "ctlType string" that works, you could just add a const in each of these controllers (or even a variable would be better than free form string) to define each of the controller types strings, and make them something easily referenced throughout the code.

Another possible suggestion would be to define an enum in Controller.go and use that but that's probably overkill here I think.

pkg/controller/clone-controller.go
```diff
@@ -226,7 +227,14 @@ func (cc *CloneController) Run(threadiness int, stopCh <-chan struct{}) error {
 	return nil
 }
 
-func (cc *CloneController) runPVCWorkers() {
+//RunWorkers is declared in Controller WorkerRunner interface
+func (cc *CloneController) RunWorkers() {
```
j-griffith: Just an idea for further consolidation; we could probably move these into Controller.go and operate on the Controller struct itself instead of the individual import, upload and clone controllers.  As it stands now we have an exported Run, ProcessNextPvcItem etc duplicated in each controller, that can be cleaned up pretty easily.  We could put these in the Controller.go pkg and just use a Controller as the receiver.  In the case of unique things that are still left we could just use the ctrlType identifier to call an unexported helper function in the appropriate controller pkg.  One problem with this is we're kinda reveresed on how I think we should do the 'has-a' relationship.  Rather than the import, clone or upload controllers having a base Controller, we should have everything be a Controller, and it include an instance of the specific controller we want.  If we do that then a lot of the logic falls into place with these and as an added bonus from an exported process flow perspective we only have to trace the flow through a single set of Controller->XYZ funcs.  It gives us a nice pattern and a sort  of contract for our controllers.

I'm not proposing you rework this patch to add that sort of thing, but I think if you're interested it would be a good next step in the consolidation/cleanup process.  I'd be happy to share some ideas and collaborate with your or anyone else that is interested.

zvikorn: ProcessNextPvcItem in upload controller does different things a little bit than the other controllers so this is the reason I left it like this. 
Run is being called from cmd/cdi-controller/controller.go and I need to see if we can move it to the (sub class) controller. did not think about it.
What do you think about merging it as it now (still there was lots of work here refactoring the code and I'm not sure you can change ProcessNextPvcItem in the uploader to be like the others, need to see) and tackle it as you said in another cleanup  pr or something..... Will the pvc transfer controller goes into CDI as well? that will be another chance. Let me know.... In the meantime, I will push the other change (ctlType vars).

zvikorn:
ci test please

j-griffith:
Maybe we should all get together and hash out some ideas for a restructuring?

j-griffith:
I think maybe we should talk a bit about our controller design.  There are good and bad things about the way things are structured, and the dedupe work is a good first pass at some dedupe it doesn't seem to help with the clarity or unwieldiness of things.

There are a number of approaches that I think we can (and should) consider here.  
1. There's the simple model that's used in K8s 
This is nice because there's a good template for it that we can leverage, there's a lot of duplicate code if you ask some, but that's also intentional.  The goal being to isolate and make things robust and very explicit as opposed to trying to optimize or reduce LOC.  One advantage to this approach in CDI is that we follow the same patterns that K8's developers are already familiar with and more importantly IMHO it makes the controller code and process flow significantly easier to follow.

2. A nested controller design
This is another approach that provides a compromise between duplicated code and clarity (again my opinion).  The idea here would be you'd have a tree structure that looked something like this:  

`
.
├── cloner
│   └── clone_controller.go
├── controller.go
├── importer
│   └── import_controller.go
└── uploader
    └── upload_controller.go

`
There are a number of things you can do here, like combine the starter process into a single executable and spawn the runners etc.  You'd be able to reuse *some* basic controller code, but each functional controller would have it's own independent code space.  

In BOTH of the approaches, the idea is to compartmentalize the code and make it significantly easier to follow and maintain, I also think that it would make it more robust in the long run.

aglitke:
Thanks for bringing this up @j-griffith.  While I can appreciate both approaches, I have a slight preference for the nested controller design as it should help us to keep all controllers updated in tandem.  I worry about someone fixing the import controller and forgetting to apply the same fix to the clone controller.  While the nested approach won't guarantee consistency, it will reduce the surface area somewhat and encourage us to implement the controllers more consistently with each other.

mhenriks:
@aglitke @j-griffith 
I think the first approach works well for a huge project like Kubernetes.  Joe Programmer can come off the street and contribute more quickly.  Prob not as important for us.

Second model I think could work well for us and is basically what Kubebuilder does.  Controllers live in their own package.  All controllers are run in a single "Manager" process.

j-griffith:
So what do folks think about this particular PR?  Do we want to continue moving forward with this approach and possibly revisit a more large scale change in the future?  Or just hold off on this one until we do a refactor of things?

zvikorn:
The only way to reduce the duplicated code in an object-oriented approach in Go is composition (as far as I know, but I'm not an expert in go).
Interfaces are defined to make classes, implement them (if needed) with a unique code. So class A implements method x in one way and class B implements method x in a different way (logic/functionality/whatever). This is not our case where much of the code in all controllers looks the same.
Since Go is not a pure Object-Oriented language, I do not care closing this PR, and live with what we have now, as it is more readable.
If someone else knows how to reduce the duplicated code (the majority of the duplicated code) in a better way (if we still want to do that), I would be more than happy to know how.

awels:
I think we can close this

PR #485, opened 2018-10-10
Orchestrate upload using a DataVolume

zvikorn:

As a user I want to orchestrate the VM disk upload process using a DataVolume object so that kubevirt will be aware of the state of the process and can prevent my VM from starting before the upload has completed.

Fixes #

**Release note**:

```release-note
NONE
```


zvikorn:
The field 'local' in upload-datavolume can be changed to anything else....
Also, I run hack/update-codegen.sh and no new 'zz_generated.deepcopy.go' file is created although I changed the type.go file to add the 'local' field. Shouldn't it be updated?

zvikorn:
ci test please

zvikorn:
ci test please

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -42,9 +42,10 @@ type DataVolumeSpec struct {
 
 // DataVolumeSource represents the source for our Data Volume, this can be HTTP, S3 or an existing PVC
 type DataVolumeSource struct {
-	HTTP *DataVolumeSourceHTTP `json:"http,omitempty"`
-	S3   *DataVolumeSourceS3   `json:"s3,omitempty"`
-	PVC  *DataVolumeSourcePVC  `json:"pvc,omitempty"`
+	HTTP  *DataVolumeSourceHTTP `json:"http,omitempty"`
+	S3    *DataVolumeSourceS3   `json:"s3,omitempty"`
+	PVC   *DataVolumeSourcePVC  `json:"pvc,omitempty"`
+	LOCAL string                `json:"local,omitempty"`
```
mhenriks: I would change LOCAL to ```UPLOAD *bool `json:"upload,omitempty"```

Also, since I'm here and you were wondering why code zz_generated.deepcopy.go wasn't updated for `LOCAL` it's because it wasn't necessary.  It's not a pointer.  The following line would copy it:

https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/apis/datavolumecontroller/v1alpha1/zz_generated.deepcopy.go#L91

Change type to `*bool` and code generation will get updated

zvikorn: or maybe even just bool

mhenriks: You should use a pointer to be consistent and signify that the value is optional.  To me bool makes more sense than string.

manifests/example/upload-datavolume.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: DataVolume
+metadata:
+  name: upload-datavolume
+spec:
+  source:
+      local: "true"
```
aglitke: As @mhenriks mentioned, 'upload' seems a better fit here.

pkg/controller/datavolume-controller.go
```diff
@@ -334,6 +334,24 @@ func (c *DataVolumeController) updateCloneStatusPhase(pvc *corev1.PersistentVolu
...
+func (c *DataVolumeController) updateUploadStatusPhase(pvc *corev1.PersistentVolumeClaim, dataVolumeCopy *cdiv1.DataVolume) {
+	phase, ok := pvc.Annotations[AnnPodPhase]
+	if ok {
+		switch phase {
+		case string(corev1.PodPending):
+			// TODO: Use a more generic Scheduled, like maybe TransferScheduled.
+			dataVolumeCopy.Status.Phase = cdiv1.UploadScheduled
+		case string(corev1.PodRunning):
+			// TODO: Use a more generic In Progess, like maybe TransferInProgress.
+			dataVolumeCopy.Status.Phase = cdiv1.UploadInProgress
```
aglitke: This is a bit misleading.  UploadInProgress makes me think someone has started sending data.  How about UploadReady?

zvikorn: take into account that when the image is uploaded for real, the pod status is still running, and the status you will have in the datavolume is still UplaodReady...  There is not difference in the pods status (running) when the image is ready to be uploaded and when it's loaded...

mhenriks: I prefer UploadReady as well.

manifests/example/upload-datavolume.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: DataVolume
+metadata:
+  name: upload-datavolume
+spec:
+  source:
+      upload: true
+  pvc:
+    storageClassName: "hostpath"
```
awels: Do we really need to specify a storageClassName in the example? If someone doesn't have this particular storage class it will fail.

manifests/example/upload-token.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: upload.cdi.kubevirt.io/v1alpha1
 kind: UploadTokenRequest
 metadata:
-  name: upload-test
+  name: upload-datavolume
```
awels: This change breaks the upload-pvc example, as that is expecting upload-test name, please create another upload-dv-token.yaml and leave this one as is.

awels:
Besides the example nits, lgtm.

manifests/example/upload-datavolume.yaml
zvikorn: yeah, I thought about too, but datavolume.yaml contains this too, so I followed this. What do you think?

aglitke: I'd say leave it in but comment it out with a note saying it's optional if you don't want to use the default storage class.

awels: datavolume.yaml shouldn't have it either actually, but we can fix that in another PR.

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -42,9 +42,10 @@ type DataVolumeSpec struct {
 
 // DataVolumeSource represents the source for our Data Volume, this can be HTTP, S3 or an existing PVC
 type DataVolumeSource struct {
-	HTTP *DataVolumeSourceHTTP `json:"http,omitempty"`
-	S3   *DataVolumeSourceS3   `json:"s3,omitempty"`
-	PVC  *DataVolumeSourcePVC  `json:"pvc,omitempty"`
+	HTTP   *DataVolumeSourceHTTP `json:"http,omitempty"`
+	S3     *DataVolumeSourceS3   `json:"s3,omitempty"`
+	PVC    *DataVolumeSourcePVC  `json:"pvc,omitempty"`
+	UPLOAD *bool                 `json:"upload,omitempty"`
```
aglitke: I think we should follow the pattern with other sources and introduce a DataVolumeSourceUpload type.  For now this type can be an empty struct.  This will allow us to extend the functionality more easily in the future.

pkg/controller/datavolume-controller_test.go
```diff
@@ -543,3 +559,106 @@ func TestCloneClaimLost(t *testing.T) {
 	f.expectUpdateDataVolumeStatusAction(result)
 	f.run(getKey(dataVolume, t))
 }
+
+// Cloning tests
```
aglitke: This is an upload test

pkg/apis/datavolumecontroller/v1alpha1/types.go
zvikorn: how the upload-datavolume.yaml should look like if we define it as an empty struct?
the code in datavolume-controller.go check if the struct is not null before setting the annotation.
How can you define it in the yaml as not null but not setting any field inside?

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -42,9 +42,10 @@ type DataVolumeSpec struct {
 
 // DataVolumeSource represents the source for our Data Volume, this can be HTTP, S3 or an existing PVC
 type DataVolumeSource struct {
-	HTTP *DataVolumeSourceHTTP `json:"http,omitempty"`
-	S3   *DataVolumeSourceS3   `json:"s3,omitempty"`
-	PVC  *DataVolumeSourcePVC  `json:"pvc,omitempty"`
+	HTTP   *DataVolumeSourceHTTP   `json:"http,omitempty"`
+	S3     *DataVolumeSourceS3     `json:"s3,omitempty"`
+	PVC    *DataVolumeSourcePVC    `json:"pvc,omitempty"`
+	UPLOAD *DataVolumeSourceUpload `json:"upload,omitempty"`
```
mhenriks: Don't you re-run the code generator to accommodate this new type?

zvikorn: yes. forgot about that. Thanks!

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -53,6 +54,11 @@ type DataVolumeSourcePVC struct {
 	Name      string `json:"name,omitempty"`
 }
 
+// DataVolumeSourceUpload provides the parameters to create a Data Volume by uploading the source
+type DataVolumeSourceUpload struct {
+	ShouldUpload bool `json:"shouldUpload,omitempty"`
```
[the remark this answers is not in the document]
zvikorn: Fine with me. However, the controller needs to check existence of this field "target", and the check should be something like if 'dataVolume.Spec.Source.target != "" '.  Is it ok to set it now to 'default' since the controller ignore it anyway. Otherwise I need to check if 'dataVolume.Spec.Source.target == "" ' since it is a string. and we will have to change this in the future versus changing the yaml file

manifests/example/upload-datavolume.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: DataVolume
+metadata:
+  name: upload-datavolume
+spec:
+  source:
+      upload:
+        shouldUpload: true
```
aglitke: Would there ever be a situation where this should be false?  Can't we just have an empty dict such as:

```
source:
    upload: {}
```

pkg/apis/datavolumecontroller/v1alpha1/types.go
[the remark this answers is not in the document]
zvikorn: I will set the yaml to be
target: "local" for now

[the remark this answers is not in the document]
aglitke: I don't understand this conversation.  It seems like a bad idea to pass around something arbitrary.  Shouldn't the presence of `upload: {}` in the source field of the spec be enough to indicate that an upload is requested?  If some time in the future we have one or more annotations to pass wrt to upload we can add them at that time.

zvikorn:
@aglitke @mhenriks please agree on if you want DataVolumeSourceUpload to be empty struct or contains the string field that might be related to the annotation in the future. I will change the code tomorrow my morning after there is an agreement.

mhenriks:
@zvikorn @aglitke `upload: {}` works for me.

zvikorn:
ci test please

pkg/controller/datavolume-controller_test.go
zvikorn: fixed

pkg/apis/datavolumecontroller/v1alpha1/types.go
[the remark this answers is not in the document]
zvikorn: done

manifests/example/upload-datavolume.yaml
zvikorn: done

PR #490, opened 2018-10-13
[WIP] expand imported disk

danielerez:
Expand imported disk to fill maximum space available on the PV:
* Parsed PVC storage size from 'resources.requests.storage'
* Passed the size value as an env variable to the improter pod
  (taking into account the filesystem overhead).
* Added 'Resize' and 'Info' to QEMUOperations interface.
* Invoked expand image on the importer pod (verifying necessity
  according to the 'virtual-size' extracted from image info).

Change-Id: Ie574cda297f658be2013ec885e4c5a43f88d985b

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"k8s.io/client-go/util/cert/triple"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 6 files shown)


danielerez:
@aglitke @awels

pkg/controller/util.go
```diff
@@ -82,6 +84,19 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
 	return name, nil
 }
 
+// returns the PVC size with fs overhead percentage included.
+func getPvcSize(pvc *v1.PersistentVolumeClaim) (string, error) {
```
j-griffith: Will this actually be used anywhere else besides the import controller?  It might be good to keep it local in that file rather than continuing to dump random things in this util.go file.

j-griffith: This needs to be called something different.  I'd read 'getPvcSize' simply as "get me the size of the specified pvc", that's not what this is.  Maybe "getFilesystemAdjustedSize" or something?

danielerez: sure. maybe simply "getAdjustedPvcSize"?

danielerez:
Hi @j-griffith, the suggested solution here is indeed to expand the image by default. I've only added a preliminary validation that the requested size is larger than the image (by fetching virtual-size from 'qemu-img info'). We can simplify by dropping this validation if seems redundant...

awels:
#522 supersedes this.

PR #498, opened 2018-10-23
Unpacking archives

idanshaby:
This pull request adds CDI the support for unpacking archives.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -49,9 +49,12 @@ func main() {
 	contentType, _ := util.ParseEnvVar(common.ImporterContentType, false)
 	imageSize, _ := util.ParseEnvVar(common.ImporterImageSize, false)
 
-	glog.V(1).Infoln("begin import process")
+	dest := common.ImporterWritePath
+	if contentType == controller.ContentTypeArchive {
+		dest = common.ImporterVolumePath
+	}
 	dso := &importer.DataStreamOptions{
-		common.ImporterWritePath,
+		dest,
 		ep,
 		acc,
 		sec,
@@ -73,7 +76,8 @@ func main() {
 			os.Exit(1)
 		}
 	} else {
-		err = importer.CopyImage(dso)
+		glog.V(1).Infoln("begin import process")
+		err = importer.CopyData(dso)
 		if err != nil {
 			glog.Errorf("%+v", err)
 			os.Exit(1)
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -25,12 +25,12 @@ const (
 
 	// host file constants:
 
-	// ImporterWriteDir provides a constant to build ImporterWritePath
-	ImporterWriteDir = "/data"
-	// ImporterWriteFile provides a constant for our importer/datastream_ginkgo_test and to build ImporterWritePath
-	ImporterWriteFile = "disk.img"
+	// ImporterVolumePath provides a constant for the directory where the PV is mounted.
+	ImporterVolumePath = "/data"
+	// DiskImageName provides a constant for our importer/datastream_ginkgo_test and to build ImporterWritePath
+	DiskImageName = "disk.img"
 	// ImporterWritePath provides a constant for the cmd/cdi-importer/importer.go executable
-	ImporterWritePath = ImporterWriteDir + "/" + ImporterWriteFile
+	ImporterWritePath = ImporterVolumePath + "/" + DiskImageName
 
 	// ImporterPodName provides a constant to use as a prefix for Pods created by CDI (controller only)
 	ImporterPodName = "importer"
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -161,6 +161,7 @@ func newDataStream(dso *DataStreamOptions, stream io.ReadCloser) (*DataStream, e
 	// establish readers for endpoint's formats and do initial calc of size of raw endpt
 	err = ds.constructReaders(stream)
 	if err != nil {
+		ds.Close()
 		return nil, errors.Wrapf(err, "unable to construct readers")
 	}
 
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -287,14 +287,20 @@ func (d *DataStream) pollProgress(reader *util.CountingReader, idleTime, pollInt
 	}
 }
 
-// CopyImage copies the source endpoint (vm image) to the provided destination path.
-func CopyImage(dso *DataStreamOptions) error {
+// CopyData copies the source endpoint (vm image) to the provided destination path.
+func CopyData(dso *DataStreamOptions) error {
 	glog.V(1).Infof("copying %q to %q...\n", dso.Endpoint, dso.Dest)
 	ds, err := NewDataStream(dso)
 	if err != nil {
 		return errors.Wrap(err, "unable to create data stream")
 	}
 	defer ds.Close()
+	if dso.ContentType == controller.ContentTypeArchive {
+		if err := UnArchiveTar(ds.topReader(), dso.Dest); err != nil {
+			return errors.Wrap(err, "unable to untar files from endpoint")
+		}
+		return nil
+	}
 	return ds.copy(dso.Dest)
 }
 
@@ -390,6 +396,7 @@ func (d *DataStream) constructReaders(stream io.ReadCloser) error {
 	//   the iso file is tar'd we can get its size via the tar hdr -- see intro comments.
 	knownHdrs := image.CopyKnownHdrs() // need local copy since keys are removed
 	glog.V(3).Infof("constructReaders: checking compression and archive formats: %s\n", d.url.Path)
+	var isTarFile bool
 	for {
 		hdr, err := d.matchHeader(&knownHdrs)
 		if err != nil {
@@ -404,11 +411,16 @@ func (d *DataStream) constructReaders(stream io.ReadCloser) error {
 		if err != nil {
 			return errors.WithMessage(err, "could not create compression/unarchive reader")
 		}
+		isTarFile = isTarFile || hdr.Format == "tar"
 		// exit loop if hdr is qcow2 since that's the equivalent of a raw (iso) file,
 		// mentioned above as the orig source file
 		if hdr.Format == "qcow2" {
 			break
 		}
+	}
+
+	if d.ContentType == controller.ContentTypeArchive && !isTarFile {
+		return errors.Errorf("cannot process a non tar file as an archive")
 	}
 
 	if len(d.Readers) <= 2 {
@@ -515,6 +527,9 @@ func (d *DataStream) xzReader() (io.Reader, int64, error) {
 // Assumes a single file was archived.
 // Note: the size stored in the header is used rather than raw metadata.
 func (d *DataStream) tarReader() (io.Reader, int64, error) {
+	if d.ContentType == controller.ContentTypeArchive {
+		return d.mulFileTarReader()
+	}
 	tr := tar.NewReader(d.topReader())
 	hdr, err := tr.Next() // advance cursor to 1st (and only) file in tarball
 	if err != nil {
@@ -522,6 +537,28 @@ func (d *DataStream) tarReader() (io.Reader, int64, error) {
 	}
 	glog.V(2).Infof("tar: extracting %q\n", hdr.Name)
 	return tr, hdr.Size, nil
+}
+
+// Note - the tar file is processed in dataStream.CopyData
+// directly by calling util.UnArchiveTar.
+func (d *DataStream) mulFileTarReader() (io.Reader, int64, error) {
+	buf, err := ioutil.ReadAll(d.topReader())
+	if err != nil {
+		return nil, 0, err
+	}
+	tr := tar.NewReader(bytes.NewReader(buf))
+	var size int64
+	for {
+		header, err := tr.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, 0, err
+		}
+		size += header.Size
+	}
+	return bytes.NewReader(buf), size, nil
 }
 
 // If the raw endpoint is an ISO file then set the receiver's Size via the iso metadata.
--- a/pkg/importer/dataStream_test.go
+++ b/pkg/importer/dataStream_test.go
@@ -37,11 +37,17 @@ import (
 const TestImagesDir = "../../tests/images"
 
 var imageDir, _ = filepath.Abs(TestImagesDir)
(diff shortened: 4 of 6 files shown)


pkg/importer/util.go
```diff
@@ -56,3 +58,44 @@ func StreamDataToFile(dataReader io.Reader, filePath string) error {
 	}
 	return nil
 }
+
+// UnArchiveTar unarchives a tar file and streams its files
+// using the specified io.Reader to the specified destination.
+func UnArchiveTar(reader io.Reader, destFolder string) error {
```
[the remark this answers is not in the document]
aglitke: Could you stream the data to a named pipe and then invoke the actual tar command to extract?  Then we avoid needing to maintain tar extract code.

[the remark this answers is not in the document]
awels: You could probably just call exec with tar as the command, and write the contents of the reader to the stdin of the exec command with an io.Copy or something.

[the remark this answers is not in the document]
awels: Something like this should do it (note not tested so might not work), and I am not using the destDir at all, but I am sure that is just a flag to tar.
```go
func untar(r io.Reader) error {
	untar := exec.Command("tar" "xvf" "-")
	untar.Stdin = r
	err := untar.Start()
	if err != nil {
		return err
	}
	return untar.Wait()
}
```

pkg/controller/import-controller.go
```diff
@@ -106,7 +114,7 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
 		}
 		// all checks passed, let's create the importer pod!
 		ic.expectPodCreate(pvcKey)
-		pod, err = CreateImporterPod(ic.clientset, ic.image, ic.verbose, ic.pullPolicy, ep, secretName, pvc)
+		pod, err = CreateImporterPod(ic.clientset, ic.image, ic.verbose, ic.pullPolicy, ep, strconv.FormatBool(archive), secretName, pvc)
```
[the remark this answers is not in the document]
j-griffith: Once we get up to this many arguments, maybe create a request struct for CreateImporterPod?  We could likely leverage that struct for all of these CreatexxxxPod calls eventually.  Maybe something like:

`
type createPodRequest struct {
  client: kubeclient
  image: string
  verbosity: string
  pullPolicy: string
  endpoint: string
  podType: podTypes
  .....
}
`

pkg/controller/util.go
```diff
@@ -176,8 +188,8 @@ func CreateImporterPod(client kubernetes.Interface, image, verbose, pullPolicy,
 	return pod, nil
 }
 
-// MakeImporterPodSpec creates and return the importer pod spec based on the passed-in endpoint, secret and pvc.
-func MakeImporterPodSpec(image, verbose, pullPolicy, ep, secret string, pvc *v1.PersistentVolumeClaim) *v1.Pod {
+// MakeImporterPodSpec creates and return the importer pod spec based on the passed-in endpoint, archive, secret and pvc.
+func MakeImporterPodSpec(image, verbose, pullPolicy, ep, archive, secret string, pvc *v1.PersistentVolumeClaim) *v1.Pod {
```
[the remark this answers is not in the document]
j-griffith: Same opinion with respect to number of args here.

pkg/importer/dataStream_test.go
```diff
@@ -315,7 +315,7 @@ var _ = Describe("Streaming Data Conversion", func() {
 				Skip(fmt.Sprintf("*** skipping endpoint extension %q as exception", targetExt))
 			}
 		}
-		ds, err := NewDataStream(tempTestServer.URL+"/"+testBase, "", "")
+		ds, err := NewDataStream(tempTestServer.URL+"/"+testBase, "", "", false)
 
 		Expect(err).NotTo(HaveOccurred())
 		defer ds.Close()
```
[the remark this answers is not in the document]
idanshaby: I will add some unit tests on the next cycle.

pkg/importer/util.go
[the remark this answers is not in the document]
idanshaby: @awels , I saw that you added a similar helper here:
https://github.com/kubevirt/containerized-data-importer/pull/509/files#diff-a6d0d00c7b526cd6f561d7a6dffe46f3R91
Won't it be more useful here in util.go?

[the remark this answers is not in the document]
awels: Yes, feel free to move it into a utility library, at the time it was the only place using it.

pkg/common/common.go
```diff
@@ -21,11 +21,11 @@ const (
 	CDIComponentLabel = "cdi.kubevirt.io"
 
 	// host file constants:
-	importerWriteDir = "/data"
+	ImporterWriteDir = "/data"
 	// ImporterWriteFile provides a constant for our importer/datastream_ginkgo_test and to build ImporterWritePath
 	ImporterWriteFile = "disk.img"
 	//ImporterWritePath provides a constant for for the cmd/cdi-importer/importer.go executable
-	ImporterWritePath = importerWriteDir + "/" + ImporterWriteFile
+	ImporterWritePath = ImporterWriteDir + "/" + ImporterWriteFile
```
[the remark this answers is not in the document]
j-griffith: Yeah, I've always found these a bit annoying; the suggestions by @aglitke would be a nice improvement and add quite a bit of clarity while being succinct.

pkg/common/common.go
```diff
@@ -47,6 +48,13 @@ const (
 	ImporterAccessKeyID = "IMPORTER_ACCESS_KEY_ID"
 	// ImporterSecretKey provides a constant to capture our env variable "IMPORTER_SECRET_KEY"
 	ImporterSecretKey = "IMPORTER_SECRET_KEY"
+	// ImporterContent provides a constant to capture our env variable "IMPORTER_CONTENT"
+	ImporterContent = "IMPORTER_CONTENT"
+
+	// ImporterContentKubvirt (the default content) - copy the endpoint into an image.
+	ImporterContentKubvirt = "kubevirt"
```
j-griffith: Type on L#55,  s/Kubvirt/Kubevirt/

pkg/common/common.go
```diff
@@ -47,6 +48,13 @@ const (
...
+	// ImporterContentArchive - expand a tar file as an archive to the target PVC.
+	ImporterContentArchive = "archive"
```
j-griffith: I'm not sure about this one "ImporterContentArchive = "archive""...  maybe ContentType would be better?

pkg/controller/util.go
```diff
@@ -91,6 +91,23 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
...
+// returns the import content.
+func getImportContent(pvc *v1.PersistentVolumeClaim) (string, error) {
+	contents := []string{common.ImporterContentKubvirt, common.ImporterContentArchive}
+	contentsSet := make(map[string]struct{})
+	for _, v := range contents {
+		contentsSet[v] = struct{}{}
+	}
+	content, found := pvc.Annotations[AnnContent]
+	if !found {
+		return common.ImporterContentKubvirt, nil
```
j-griffith: Kubvirt here as well

pkg/controller/util.go
```diff
@@ -91,6 +91,23 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
 	return name, nil
 }
 
+// returns the import content.
```
j-griffith: // returns the import content.  --- well, not really; it's actually doing a `getOrSetContentType`

pkg/controller/util.go
```diff
@@ -91,6 +91,23 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
...
+func getImportContent(pvc *v1.PersistentVolumeClaim) (string, error) {
```
j-griffith: Sorry, I'm not completely following what's going on in the logic below here?  Also, it seems like we could just declare a variable to represent this map and set it (it's just joining the existing constants we set).  Maybe I'm missing something, but this seems like a lot of churn just to get the string back from the pvc content annotation.  I'm sorry if I'm missing this.

pkg/importer/dataStream.go
```diff
@@ -98,15 +99,15 @@ const httpClientTimeout = time.Minute * 60
...
 // Note: the caller must close the `Readers` in reverse order. See Close().
-func NewDataStream(endpt, accKey, secKey string) (*DataStream, error) {
-	return newDataStream(endpt, accKey, secKey, nil)
+func NewDataStream(endpt, accKey, secKey, content string) (*DataStream, error) {
+	return newDataStream(endpt, accKey, secKey, content, nil)
 }
 
 func newDataStreamFromStream(stream io.ReadCloser) (*DataStream, error) {
-	return newDataStream("stream://data", "", "", stream)
+	return newDataStream("stream://data", "", "", common.ImporterContentKubvirt, stream)
```
j-griffith: Kubevirt

cmd/cdi-importer/importer.go
```diff
@@ -33,9 +35,18 @@ func main() {
...
 	acc, _ := importer.ParseEnvVar(common.ImporterAccessKeyID, false)
 	sec, _ := importer.ParseEnvVar(common.ImporterSecretKey, false)
+	archive, _ := importer.ParseEnvVar(common.ImporterArchive, false)
+	archiveBool, _ := strconv.ParseBool(archive)
+	var dest string
+	if archiveBool {
+		dest = common.ImporterWriteDir
+	} else {
+		dest = common.ImporterWritePath
+	}
```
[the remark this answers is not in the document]
awels: Now that #520 has been merged, you can look at the contentType to determine the path where you want to write the files.

pkg/importer/util.go
```diff
@@ -42,3 +44,23 @@ func StreamDataToFile(dataReader io.Reader, filePath string) error {
 	}
 	return nil
 }
+
+// UnArchiveTar unarchives a tar file and streams its files
+// using the specified io.Reader to the specified destination.
+func UnArchiveTar(reader io.Reader, destDir string) error {
```
awels: Maybe we can re-use the untar from clonertarget.go. Move it into util or something, then importing from repo can use it as well.

pkg/importer/dataStream.go
```diff
@@ -361,7 +375,11 @@ func (d *DataStream) fileFormatSelector(hdr *image.Header) (err error) {
 		r, d.Size, err = d.qcow2NopReader(hdr)
 		d.qemu = true
 	case "tar":
-		r, d.Size, err = d.tarReader()
+		if d.content == common.ImporterContentArchive {
+			r, d.Size, err = d.mulFileTarReader()
```
j-griffith: would be good to abstract the d.mulFileTarReader and the tarReader functions into a single wrapper function that takes d.content and let it figure out which one to call.

pkg/importer/dataStream.go
```diff
@@ -428,6 +446,14 @@ func (d *DataStream) tarReader() (io.Reader, int64, error) {
 	return tr, hdr.Size, nil
 }
 
+func (d *DataStream) mulFileTarReader() (io.Reader, int64, error) {
```
j-griffith: Since we don't have an impl here yet, if you introduce the wrapper as I mentioned above it could just return nil,0,nil for this type without adding/calling a function with no implementation.

pkg/importer/util.go
```diff
@@ -42,3 +44,23 @@ func StreamDataToFile(dataReader io.Reader, filePath string) error {
...
+	untar.Stdin = reader
+	var errBuf bytes.Buffer
+	untar.Stderr = &errBuf
+	err := untar.Start()
+	if err != nil {
+		return err
+	}
+	err = untar.Wait()
+	if err != nil {
+		glog.Errorf("%s\n", string(errBuf.Bytes()))
```
j-griffith: Should probably have some sort of statement before dumping the errBuf here maybe?  Could probably include the err.Error() string for the error log, and a V3 or greater for the actual buffer.. but I might be wrong on that one.  I just think there at least needs to be something telling me what I'm looking at in the logs when you hit this error.

cmd/cdi-importer/importer.go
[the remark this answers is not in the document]
idanshaby: Done

pkg/importer/util.go
idanshaby: It's already in util.
If you think I should change anything else, please say so.

cmd/cdi-importer/importer.go
```diff
@@ -11,6 +11,9 @@ package main
 //    ImporterAccessKeyID  Optional. Access key is the user ID that uniquely identifies your
 //			      account.
 //    ImporterSecretKey     Optional. Secret key is the password to your account.
+//    ImporterContent		Optional. Specifies the content type of the endpoint:
```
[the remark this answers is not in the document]
idanshaby: Irrelevant anymore after rebasing on top of #520.

cmd/cdi-importer/importer.go
```diff
@@ -11,6 +11,9 @@ package main
...
+//					kubevirt (the default content) - the endpoint (image) is copied into the target PVC.
+//					archive - the endpoint (tar file) is processed as an archive and expanded into the target PVC.
```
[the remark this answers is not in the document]
idanshaby: Irrelevant anymore after rebasing on top of #520.

pkg/common/common.go
idanshaby: Irrelevant anymore after rebasing on top of #520.

pkg/controller/util.go
idanshaby: Irrelevant anymore after rebasing on top of #520.

pkg/importer/dataStream.go
idanshaby: Irrelevant anymore after rebasing on top of #520.

pkg/importer/dataStream.go
idanshaby: Good idea, thanks!

pkg/importer/util.go
idanshaby: Done, is that what you suggested?

idanshaby:
ci add to whitelist

danielerez:
@annastopel

pkg/importer/util.go
awels: Oops didn't realize that, well I guess it would nice if you can update the clonertarget to use this one instead of the one it uses. Then at the same time you can steal the unit test I have for that untar method for this one.

awels:
Please rebase, lots of changes happened.

pkg/importer/util.go
idanshaby: I guess that it would be more right to do it on a separate PR.

awels:
There is a lint error, make test-lint will run the linting.

awels:
ci test please

PR #502, opened 2018-10-25
Add structure to hold all CDI importer environment variables #501

### LINKED ISSUE #501: Add structure to hold all CDI importer environment variables

> /kind enhancement

As we add more functionality to the CDI importer, we need to use more environment variables when we create the importer pod. In order to avoid passing more and more parameters that represent these  environment variables to the methods that are responsible for the pod creation, a new struct that define these parameters is created to be passed as argument into the methods . From now on, every time we add environment variable to the importer pod, we just need to add it to the new structure and access it when we create the pod. 

**What happened**:

**Environment**:
- Install tools:
- Others:

zvikorn:

Add structure to hold all CDI importer environment variables

Fixes #501 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -30,6 +30,12 @@ type ImportController struct {
 	Controller
 }
 
+type importPodEnvVar struct {
+	ep, secretName string
+}
+
+var podEnvVar importPodEnvVar
+
 // NewImportController sets up an Import Controller, and returns a pointer to
 // the newly created Import Controller
 func NewImportController(client kubernetes.Interface,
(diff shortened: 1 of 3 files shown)


zvikorn:
@aglitke @awels @mhenriks @idanshaby Please review as Idan's and my next PR's (Unpacking archives, blank image) rely on this.
Thanks!

awels:
lgtm

pkg/controller/import-controller.go
```diff
@@ -93,20 +99,20 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
 		needsSync = false
 	}
 	if pod == nil && needsSync {
-		ep, err := getEndpoint(pvc)
+		podEnvVar.ep, err = getEndpoint(pvc)
```
mhenriks: Wondering about the use of a global variable here.  The controller would have to be single threaded for this to be safe.  Which I believe would be the case if threadiness is > 1 here:  https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/controller/controller.go#L170

awels: You are right, that isn't safe, we should create a new struct here and pass it down instead of using a global.

zvikorn: Thanks! Forgot this can run in a multi-threaded environment. I am going to declare the variable not global. inside the if block.I believe it satisfies this.

zvikorn: how to we change this after the merge? Can we reopen this? or I need a new PR?

awels: Just create a new PR @zvikorn that fixes the issue.

PR #503, opened 2018-10-25
Add ability to create empty disk images #497

### LINKED ISSUE #497: Add ability to create empty disk images

/kind enhancement

**What happened**:
I've been working with a combination of CNV, glusterfs and PXE. I've been creating VMs to PXE boot from a Satellite server to be provisioned. I've been using CDI to import a blank disk image from a URL.

Would it be possible for CDI to have an option to create empty/blank disks rather than import an existing image?

n/a

**Environment**:
n/a. using `kubevirtci/disk-importer`
Client Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2018-09-26T12:30:09Z", GoVersion:"go1.10.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2018-09-26T12:30:09Z", GoVersion:"go1.10.2", Compiler:"gc", Platform:"linux/amd64"}
bare metal PowerEdge servers
- Install tools:
openshift-ansible, cnv installed directly from manifest

zvikorn:

[epic-improve-import-clone-1.4] Support blank disks

Fixes #497

**Release note**:

```release-note
This feature adds the ability to create an empty raw image, triggered by special PVC annotation.
```


pkg/controller/import-controller.go
```diff
@@ -23,13 +23,22 @@ const (
...
 type ImportController struct {
 	Controller
 }
 
+type importPodEnvVar struct {
+	ep, secretName, imageSize string
+	createImage               bool
+}
+
+var podEnvVar importPodEnvVar
```
mhenriks: See my comment in related (closed) PR:  https://github.com/kubevirt/containerized-data-importer/pull/502

Basically, I don't think it's safe to use a global variable here.

zvikorn: Rebased to fix that.

zvikorn:
ci test please

cmd/cdi-importer/importer.go
```diff
@@ -33,12 +33,22 @@ func main() {
...
-	if err != nil {
-		glog.Errorf("%+v", err)
-		os.Exit(1)
+	if ep != "" {
+		err := importer.CopyImage(common.ImporterWritePath, ep, acc, sec)
+		if err != nil {
+			glog.Errorf("%+v", err)
+			os.Exit(1)
+		}
+	}else if createImage == "true"{
```
awels: Needs a space between } and else?

tests/import_test.go
```diff
@@ -53,4 +53,27 @@ var _ = Describe(testSuiteName, func() {
 		Expect(phaseAnnotation).To(BeTrue())
 		Expect(status).Should(BeEquivalentTo(v1.PodPending))
 	})
+	/*It("Should create import pod for creating raw image", func() {
```
awels: This should be commented?

awels:
@zvikorn A few questions/comments. This will need a release note, so please put a line or two explaining what this is in the release note section of the PR.

I think we need to stop for a second and consider the API we are building. We are adding a new source "empty" and an empty size, but in another PR we are adding a new operation resize disk with a size as well. Maybe we should combine the 'size' field for all the operations that can specify a size, like resize,new disk, whatever. Heck we should even be able to specify a resize on a clone, after the clone it will resize the disk.

@aglitke @mhenriks what do you think about above thought?

zvikorn:
If you mean 1) create image size 0
2) resize it by the other PR
It’s something I was thinking about before implementing this, but somehow thought that maybe we want to do it in one ‘Atomic’ step. I don’t mind either way. Let me guys know.

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -46,6 +46,7 @@ type DataVolumeSource struct {
 	S3     *DataVolumeSourceS3     `json:"s3,omitempty"`
 	PVC    *DataVolumeSourcePVC    `json:"pvc,omitempty"`
 	UPLOAD *DataVolumeSourceUpload `json:"upload,omitempty"`
+	CREATE *DataVolumeCreateImage  `json:"create,omitempty"`
```
mhenriks: This was something that slipped by me with upload datavolume integration and I think we should probably fix/not perpetuate here.  `HTTP/PVC/S3` are acronyms, so it makes sense for them to be all caps.  `UPLOAD` and `CREATE` are not acronyms so they should be changed to `Upload` and `Create` respectively.

Although, I'm not crazy about the name `Create`.  I wonder if `Empty` of `Blank` would be better?  Or any other suggestions?

pkg/controller/import-controller.go
```diff
@@ -23,6 +23,8 @@ const (
 	AnnImportPod = "cdi.kubevirt.io/storage.import.importPodName"
 	//LabelImportPvc is a pod label used to find the import pod that was created by the relevant PVC
 	LabelImportPvc = "cdi.kubevirt.io/storage.import.importPvcName"
+	//AnnCreateImage provides a const for our PVC createImage annotation
+	AnnCreateImage = "cdi.kubevirt.io/storage.import.createImage"
```
mhenriks: Again, not sure `create` is the right word.  Because an image always gets created.  What makes this special is that it's blank/empty.  I may be off here, but what do others think?

pkg/controller/import-controller.go
```diff
@@ -99,19 +102,30 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
 	if pod == nil && needsSync {
 		var podEnvVar importPodEnvVar
 
+		//check for endpoint annotation
 		podEnvVar.ep, err = getEndpoint(pvc)
```
mhenriks: Nat a changed in this PR, but it seems like is is OK for a pvc to not have an endpoint.  So maybe that case shouldn't be treated as an error?  

Maybe it would be best to have a function that returns a fully populated `importPodEnvVar` based on a PVC, then the code can make decisions based on that?

mhenriks:
@awels maybe I'm not understanding your suggestion, but the size for the empty disk is coming from the PVC specification.  Which seems like a natural place to me for this case.  

BUT I wonder if, after considering filesystem overhead, the PVC is guaranteed to have free space equal to the requested PVC size?  I'm guessing no.  Maybe qemu-img handles that?  Or maybe we should check the actual free space on the PVC and adjust the size passed to `qemu-img create` accordingly?

cc: @aglitke @j-griffith

mhenriks:
@awels I understand that the `volume` would be at least the requested size.  But that does not necessarily mean that the `filesystem` has that much free space.  Or is that a guarantee in docs/code somewhere?

cmd/cdi-importer/importer.go
zvikorn: fixed

pkg/controller/util.go
```diff
@@ -90,6 +91,22 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
...
+// returns whether the AnnCreateImage annotation exists and the expected image size from the PVC
+func getCreateImage(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (bool, string, error) {
+	_, found := pvc.Annotations[AnnCreateImage]
+	if !found {
+		return false, "", errors.Errorf("annotation %q in pvc \"%s/%s\" is missing\n", AnnCreateImage, pvc.Namespace, pvc.Name)
+	}
+	glog.V(3).Infof("getCreateImage: retrieving requested storage size from pvc \"%s/%s\"\n", pvc.Namespace, pvc.Name)
+	storageSize := pvc.Spec.Resources.Requests[k8sv1.ResourceStorage]
+	storageSizeStr := storageSize.String()
+	//this is to eliminate the 'i' from the binary presentation of the image size ("Gi") as qemu does not recognize this.
```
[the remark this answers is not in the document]
zvikorn: @awels Are you sure ParseQuantity remove the 'i' from the end of the String ("500Mi -> 500M)?
I tried that and it's not..... 
QEMU cannot get the 'i' at the end.

tests/import_test.go
zvikorn: sorry, forgot to uncomment.

pkg/apis/datavolumecontroller/v1alpha1/types.go
zvikorn: okay. Will change it to Blank for now.

pkg/controller/import-controller.go
zvikorn: until we get another decision, I changed to 'blank'

pkg/controller/import-controller.go
zvikorn: I'm not sure I understood.
"So maybe that case shouldn't be treated as an error?"
In my flow, if we do not find endpoint anno,we look for blankImage anno. If both does not exist, it's an error. Does not it make sense?

mhenriks: Sure, but on line 108, an error is logged if PVC does not have an endpoint.  That is not really an error as it is expected in the blank disk use case.

I personally think it may be cleaner to get rid of `getEndpoint()`, `getBlankInage()`, and `getSectetName()` (and any other similar functions that assign members of podEnvVar) and have a single function that takes a PVC and returns a fully initialized `importPodEnvVar`.  The controller can then make decisions (is this an import or a blank disk) based on the info in the structure.

zvikorn:
@awels @mhenriks updated  the PR. Still waiting for an answer from Michael about the PVC missing endpoint annotation and for both of you regarding the PVC space.

pkg/controller/import-controller.go
zvikorn: Oh.. I see now. Sure. I will write one method to do that...

zvikorn:
ci test please

zvikorn:
ci test please

pkg/controller/import-controller.go
```diff
@@ -67,6 +70,30 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
 	return podList[0], nil
 }
 
+func (ic *ImportController) getPodEnvVar(pvc *v1.PersistentVolumeClaim) (importPodEnvVar, error) {
```
mhenriks: Not a big deal, but I think it is generally preferable to return pointers when there may be an error, so `(*importPodEnvVar, error)` with first value being nil in error case.

zvikorn:
@mhenriks 
"I think this looks pretty good. But I think it would b wise to make sure the PVC has enough free space to create the image." 
So I create a blank image with a size taken from the PVC (requested size). I'm confused a bit by your sentence. you mean checking PV actual size?

pkg/controller/import-controller.go
zvikorn: good idea. Thanks.

mhenriks:
@zvikorn My concern is something like this:  a user requests a 10G PVC.  A dynamic storage provisioner allocates a 10G volume, but once a filesystem is allocated on that volume, free space is < 10G.

mhenriks:
@zvikorn I'm thinking the size of the image should be `min(requested_pvc_size, pvc_free_space)` with maybe a small fudge factor subtracted from `pvc_free_space` to accommodate filesystem metadata for disk.img.

zvikorn:
@mhenriks I'm missing something here.
If this is static provisioning of PV's, then the PVC will bound to a PV where it's capacity is at least as the requested amount. So for example if the requested size is 10G, it will bound to a PV with capacity equals or greater than 10G. In that case,  the size of the image can be the requested size less the FS needed.
If this is a dynamic provisioning of PV's, then a PV with the requested size will be created. Isn't it?
Meaning, the PV capacity will be equals to the requested size. In that case, again, the image size can be the requested size less FS needed.
Please correct me if I'm wrong....

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
@aglitke @awels @mhenriks  still need to understand if we need to change something or not according to @mhenriks comment above.
What I know so far is, if a PV found with a size(capacity?) at least as the pvc requested size, then it is bound to the PVC. Otherwise, a dynamic PV is provisioned with the a size equals to the PVC requested size. Is there an option to have PVC bound to a PV with capacity less than the PVC requested size? If not, then I do not understand what 'min(requested_pvc_size, pvc_free_space)' means.

pkg/controller/import-controller.go
```diff
@@ -23,6 +23,8 @@ const (
 	AnnImportPod = "cdi.kubevirt.io/storage.import.importPodName"
 	//LabelImportPvc is a pod label used to find the import pod that was created by the relevant PVC
 	LabelImportPvc = "cdi.kubevirt.io/storage.import.importPvcName"
+	//AnnBlankImage provides a const for our PVC blankImage annotation
+	AnnBlankImage = "cdi.kubevirt.io/storage.import.blankImage"
```
awels: @zvikorn 
I believe we decided on giving everything a 'source', so a blank image becomes

cdi.kubevirt.io/storage.import.source: blank

Then the controller can decided what to do based on the value of the source. I am creating a card to do the same for the existing sources.

zvikorn: oh, I see. I thought that we will change everything together at some point. Ok, I will change it. Thanks!

zvikorn: @awels I take my words back :-) I think that we should have a PR first to change the existing annotation, and based on that I will update my bank image feature. I do not want to change anything which is not related to the feature itself like creating one common annotation.  This will be in the other PR. What do you think?

awels: I agree, I don't think it will be that much work to do an intermediate PR with making the change to include the 'source'.

awels:
@zvikorn So if I understand Michaels comment correctly, his question is if I request a 10G empty image, by requesting a 10G PVC, will you actually get a 10G empty image. And I think the answer is no, you will get a 9.somethingG sized image, because the PVC size will be 10G, and there is some file system overhead. I don't think there would be a way to request an actual 10G image, without the user pre-calculating some extra size into the PVC.

manifests/example/blank-image-pvc.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: "blank-image-pvc"
+  labels:
+    app: containerized-data-importer
+  annotations:
+    cdi.kubevirt.io/storage.import.blankImage: ""
```
awels: If we change the annotation, this will need to change as well

pkg/controller/datavolume-controller.go
```diff
@@ -486,6 +486,9 @@ func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolu
 			}
 
 			_, ok := pvc.Annotations[AnnImportPod]
+			if !ok {
+				_, ok = pvc.Annotations[AnnBlankImage]
```
awels: This will probably need to change in the pre PR so we get the value of the annotation.

pkg/controller/datavolume-controller.go
```diff
@@ -632,6 +635,8 @@ func newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume) (*corev1.PersistentV
 		}
 	} else if dataVolume.Spec.Source.Upload != nil {
 		annotations[AnnUploadRequest] = ""
+	} else if dataVolume.Spec.Source.Blank != nil {
+		annotations[AnnBlankImage] = ""
```
awels: Same here should be something like 

  annotations[AnnSource] = Blank

Where Blank is "Blank"

awels:
In general looks good to me, some minor comments, and I am missing a new unit test for the empty disk. I see extra parameters for the existing unit tests, but no new ones for create.

zvikorn:
@awels Thanks! you are right about the unit test. Will TestCreatesImportPod with the blank annotation be enough? or you think we need the other tests that involve creation of PVC (for expectation, status) too?

awels:
@zvikorn The PVC creation is handled by the e2e test. So just adding some tests when you call the function should be enough.

awels:
Looks good to me

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
Re-based. Ready for another review.

zvikorn:
ci test please

cmd/cdi-importer/importer.go
```diff
@@ -44,6 +46,7 @@ func main() {
 	sec, _ := util.ParseEnvVar(common.ImporterSecretKey, false)
 	source, _ := util.ParseEnvVar(common.ImporterSource, false)
 	contentType, _ := util.ParseEnvVar(common.ImporterContentType, false)
+	blankImageSize, _ := util.ParseEnvVar(common.ImporterBlankImageSize, false)
```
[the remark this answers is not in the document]
zvikorn: of course :-)

pkg/controller/util.go
```diff
@@ -148,6 +149,17 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
...
 }
 
+// returns the expected image size from the PVC
+func getBlankImageSize(pvc *v1.PersistentVolumeClaim) (string, error) {
+	glog.V(3).Infof("getBlankImage: retrieving requested storage size from pvc \"%s/%s\"\n", pvc.Namespace, pvc.Name)
+	storageSize := pvc.Spec.Resources.Requests[k8sv1.ResourceStorage]
+	iStorageSize := storageSize.Value()
+	// Accounting filesystem overhead - up to 4% according to:
+	// https://rwmj.wordpress.com/2009/11/08/filesystem-metadata-overhead/
+	fStorageSize := float64(iStorageSize) * float64(0.96)
```
awels: We don't need to take 96% of the requested size. we will solve on the importer side.

pkg/controller/util.go
```diff
@@ -148,6 +149,17 @@ func getSecretName(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim) (
 	return name, nil
 }
 
+// returns the expected image size from the PVC
+func getBlankImageSize(pvc *v1.PersistentVolumeClaim) (string, error) {
```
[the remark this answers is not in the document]
zvikorn: you will have more places.....

awels:
@zvikorn Can you address the renaming comments? I would like to merge this sooner rather than later.

zvikorn:
Sure. Merge this and I will rename mine

awels:
Also the 96% calculation is wrong, so that needs fixing as well.

awels:
Oh I misunderstood you want me to merge my size patch first and you will rebase. sounds good.

zvikorn:
@aglitke @awels @mhenriks @j-griffith 
There were lots of comments on this patch about adjusting the requested image size before we create it.
Initially, we were talking about reducing 4% for FS overhead. Then as far as I understood, this was not needed anymore as we just taking the minimum of the PV available value and the requested image size. This is not working because, the importer pod will not start running if the requested size is greater than the PV available size (pending state). So you cannot run the code that finds the minimum if the pod is not running. Meaning, the pod will run only if there will be a PV with a size at least as the requested size in the PVC.

Can someone please write (elaborate would be great) here or in trello (what's more convenient), what are the requirements regarding the size image we want to create?

Thanks!

zvikorn:
@awels by the way, I could not expect something in details better than this. I appreciate that!

zvikorn:
@awels 
can util.getAvailableSpace("/data") return smaller value than the available size on the entire PV? If yes, when can this happen?

zvikorn:
@awels updated patch. Thanks!

pkg/controller/util.go
```diff
@@ -863,3 +858,25 @@ func deletePod(req podDeleteRequest) error {
 	}
 	return err
 }
+
+func getPodEnvVar(pvc *v1.PersistentVolumeClaim, ic *ImportController) (*importPodEnvVar, error) {
```
awels: Can you rename this to be createImportEnvVar to better reflect its meaning?

pkg/controller/util.go
```diff
@@ -863,3 +858,25 @@ func deletePod(req podDeleteRequest) error {
...
+	podEnvVar := &importPodEnvVar{}
+	podEnvVar.source = getSource(pvc)
+	podEnvVar.contentType = getContentType(pvc)
+
+	var err error
```
awels: ep/secret is only relevant if source != none, if you check here, then you can keep the original logging code in getEndpoint

pkg/controller/util.go
```diff
@@ -68,19 +66,16 @@ func checkPVC(pvc *v1.PersistentVolumeClaim, annotation string) bool {
...
-func getEndpoint(pvc *v1.PersistentVolumeClaim) (string, error) {
+func getEndpoint(pvc *v1.PersistentVolumeClaim) string {
 	ep, found := pvc.Annotations[AnnEndpoint]
-	if !found || ep == "" {
-		verb := "empty"
-		if !found {
-			verb = "missing"
-		}
-		return ep, errors.Errorf("annotation %q in pvc \"%s/%s\" is %s\n", AnnEndpoint, pvc.Namespace, pvc.Name, verb)
+
```
awels: We should keep the logging, its important to know there is no endpoint when you are expecting it. We just have to not call this function when source is none.

pkg/image/qemu.go
```diff
@@ -192,3 +193,19 @@ func reportProgress(line string) {
...
 }
+
+// CreateBlankImage creates empty raw image
+func CreateBlankImage(dest, size string) error {
+	glog.V(1).Infof("creating raw image with size %s", size)
+	return qemuIterface.CreateBlankImage(dest, size)
+}
+
+// CreateBlankImage creates a raw image with a given size
+func (o *qemuOperations) CreateBlankImage(dest, size string) error {
```
awels: Please rebase and look at the Resize function and follow that pattern. The size is a Quantity there and there is a function that translates the Quantity into a qemu compatible string. Right now this code won't work for a value in the K range.

cmd/cdi-importer/importer.go
```diff
@@ -56,10 +59,27 @@ func main() {
 		contentType,
 		imageSize,
 	}
-	err = importer.CopyImage(dso)
-	if err != nil {
-		glog.Errorf("%+v", err)
-		os.Exit(1)
+
+	if source == controller.SourceNone && contentType == controller.ContentTypeKubevirt {
+		qImageSize := resource.MustParse(imageSize)
```
awels: Doesn't MustParse produce a quantity already, why are we changing it to a string and then back to a quantity again?

pkg/image/qemu.go
zvikorn: done

PR #511, opened 2018-11-01
swagger/apidoc generation

mhenriks:

Generates the openapi spec (swagger.json) and from there apidocs and python client.  Most of this code was stolen from KubeVirt.  Still need new repos created and deploy finished/tested.

Fixes #

**Release note**:

```release-note
1.  OpenAPI spec (swagger.json) generation
2.  API docs generation from swagger.json
```


mhenriks:
ci test please

hack/build/common.sh
```diff
@@ -21,3 +21,5 @@ BUILD_DIR=${CDI_DIR}/hack/build
 MANIFEST_TEMPLATE_DIR=${CDI_DIR}/manifests/templates
 MANIFEST_GENERATED_DIR=${CDI_DIR}/manifests/generated
 SOURCE_DIRS="pkg tests tools"
+APIDOCS_OUT_DIR=${OUT_DIR}/apidocs
+PYTHON_CLIENT_OUT_DIR=${OUT_DIR}/cdi-client-python
```
awels: We still need this variable since we are not doing the python client anymore?

hack/build/docker/builder/Dockerfile
```diff
@@ -24,7 +24,17 @@ RUN \
     ( go get -d github.com/mattn/goveralls && \
     cd $GOPATH/src/github.com/mattn/goveralls && \
     go install ./... ) && \
-    ( go get -u golang.org/x/lint/golint )
+    ( go get -u golang.org/x/lint/golint ) && \
+    ( go get -u github.com/mhenriks/go-swagger-utils/swagger-doc ) && \
```
awels: Sure we want your private repo in here?

hack/build/docker/builder/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM fedora:28
 
-RUN dnf install -y qemu xz gzip git && dnf clean all
+RUN dnf install -y qemu xz gzip git gradle && dnf clean all
```
awels: @aglitke This will need an update for downstream to match I think.

hack/build/common.sh
mhenriks: good catch

hack/build/docker/builder/Dockerfile
mhenriks: It's either mine or @rmohr's!  Should prob use Roman's though as that's what KubeVirt uses.

rmohr: If you can, go with the official openapi-gen from k8s. We had an issue blocking us in the past to switch. Since my code generator also just works for us right now we did not switch yet. Here the PR: https://github.com/kubevirt/kubevirt/pull/1419.

rmohr: Oh, it might only generate swagger v3. Forgot that. Not sure if you need v2 or v3.

PR #518, opened 2018-11-08
Rolling timeout on streaming transfer.

awels:

This PR allows for rolling timeouts on long transfers. As long as there is some activity for 10 minutes the transfer will continue. Right now we have a timeout on imports of one hour, anything that takes longer will abort regardless of progress being made. This allows for long running imports to not abort as long as progress is happening, and for it to abort if the other end is really not sending anything.

**Release note**:

```release-note
Rolling timeouts on long imports, as long as there is activity within the last 10 minutes the import will not timeout
```

### DIFF
--- a/cmd/cdi-cloner/clonertarget.go
+++ b/cmd/cdi-cloner/clonertarget.go
@@ -21,9 +21,8 @@ import (
 )
 
 type prometheusProgressReader struct {
-	reader  io.Reader
-	current int64
-	total   int64
+	util.CountingReader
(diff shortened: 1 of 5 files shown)


j-griffith:
Why do we need it?

Issue?  Remove the form data since you're not using it

Other than that the change seems solid at first glance.

awels:
ci test please

awels:
Updated the message to better explain why this is something we want.

PR #520, opened 2018-11-12
Add source and contentType annotations

awels:

This PR adds new annotations to help determine the source and content type used for an import.

**Release note**:

```release-note
New annotation for source: cdi.kubevirt.io/storage.import.source
New annotation for content-type: cdi.kubevirt.io/storage.contentType
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -42,9 +42,19 @@ func main() {
 	ep, _ := util.ParseEnvVar(common.ImporterEndpoint, false)
 	acc, _ := util.ParseEnvVar(common.ImporterAccessKeyID, false)
 	sec, _ := util.ParseEnvVar(common.ImporterSecretKey, false)
+	source, _ := util.ParseEnvVar(common.ImporterSource, false)
+	contentType, _ := util.ParseEnvVar(common.ImporterContentType, false)
 
 	glog.V(1).Infoln("begin import process")
-	err = importer.CopyImage(common.ImporterWritePath, ep, acc, sec)
+	dso := &importer.DataStreamOptions{
+		common.ImporterWritePath,
+		ep,
+		acc,
+		sec,
+		source,
+		contentType,
+	}
+	err = importer.CopyImage(dso)
 	if err != nil {
 		glog.Errorf("%+v", err)
 		os.Exit(1)
--- a/doc/annotations.md
+++ b/doc/annotations.md
@@ -1,0 +1,94 @@
+# Sources and contentType
+All annotations associated with Containerized Data Importer (CDI) have a prefix of: cdi.kubevirt.io. This denotes that the annotations are meant to be consumed by the CDI controller.
+
+## Source
+Source describes the type of data source CDI will be collecting the data from. Based on the value of source, additional annotations may be required to successfully import the data. The full annotation for source is: cdi.kubevirt.io/storage.import.source. The following values are currently available:
+* http
+* S3
+* none (don't import, but create data based on the contentType annotation)
+
+### http and s3
+The http and s3 sources require an additional annotation to describe the end point CDI needs to connect to. The annotation is cdi.kubevirt.io/storage.import.endpoint. If the end point requires authentication one can add an optional annotation to point to a Kubernetes Secret to get authentication information from. This annotation is: cdi.kubevirt.io/storage.import.secretName. If the source annotation is missing it will default to "http". 
+
+#### contentType
+There is an additional annotation that determines the content type of the http/s3 source, the content type can be one of the following:
+* kubevirt (Virtual Machine image)
+* archive (tar archive)
+If the contentType is missing, it is defaulted to kubevirt.
+
+#### examples
+Creating a PVC that imports data from an http source with kubevirt contentType:
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: "example-pvc"
+  labels:
+    app: containerized-data-importer
+  annotations:
+    cdi.kubevirt.io/storage.import.source: "http" #defaults to http if missing or invalid
+    cdi.kubevirt.io/storage.import.contentType: "kubevirt" #defaults to kubevirt if missing or invalid.
+    cdi.kubevirt.io/storage.import.endpoint: "https://www.source.example/path/of/data" # http or https is supported
+    cdi.kubevirt.io/storage.import.secretName: "" # Optional. The name of the secret containing credentials for the end point
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi # Request a size that is large enough to accept the data from the source, including conversion
+  # Optional: Set the storage class or omit to accept the default
+  # storageClassName: local
+``` 
+
+Creating a PVC that imports data from an http source with archive contentType:
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: "example-pvc"
+  labels:
+    app: containerized-data-importer
+  annotations:
+    cdi.kubevirt.io/storage.import.source: "http" #defaults to http if missing or invalid
+    cdi.kubevirt.io/storage.import.contentType: "archive" #defaults to kubevirt if missing or invalid.
+    cdi.kubevirt.io/storage.import.endpoint: "http://www.source.example/path/of/data.tar" # http or https is supported
+    cdi.kubevirt.io/storage.import.secretName: "" # Optional. The name of the secret containing credentials for the end point
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi # Request a size that is large enough to accept the data from the source, including conversion
+  # Optional: Set the storage class or omit to accept the default
+  # storageClassName: local
+``` 
+
+### None
+The none source indicates there is no source to get data from and instead a default action should be taken based on the contentType.
+
+#### contentType
+There is currently only one contentType that any meaning with a source of None. If the contentType is kubevirt, it will create an empty Virtual Machine image of the size specified in the PersistentVolumeClaim(PVC) request. Specifying a source of none and a contentType of archive will not do anything.
+
+#### example
+Creating a PVC that creates an empty virtual machine disk:
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: "example-pvc"
+  labels:
+    app: containerized-data-importer
+  annotations:
+    cdi.kubevirt.io/storage.import.source: "none"
+    cdi.kubevirt.io/storage.import.contentType: "kubevirt" #defaults to kubevirt if missing or invalid.
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi # Request a size that is large enough to accept the data from the source, including conversion
+  # Optional: Set the storage class or omit to accept the default
+  # storageClassName: local
+``` 
+
+
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -15,14 +15,20 @@ import (
 )
 
 const (
+	// AnnAPIGroup is the APIGroup for CDI
+	AnnAPIGroup = "cdi.kubevirt.io"
+	// AnnSource provide a const for our PVC import source annotation
+	AnnSource = AnnAPIGroup + "/storage.import.source"
 	// AnnEndpoint provides a const for our PVC endpoint annotation
-	AnnEndpoint = "cdi.kubevirt.io/storage.import.endpoint"
+	AnnEndpoint = AnnAPIGroup + "/storage.import.endpoint"
 	// AnnSecret provides a const for our PVC secretName annotation
-	AnnSecret = "cdi.kubevirt.io/storage.import.secretName"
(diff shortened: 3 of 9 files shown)


cmd/cdi-importer/importer.go
```diff
@@ -42,9 +42,19 @@ func main() {
...
-	err = importer.CopyImage(common.ImporterWritePath, ep, acc, sec)
+	dso := &importer.DataStreamOptions{
+		common.ImporterWritePath,
+		ep,
+		acc,
+		sec,
+		source,
+		contentType,
+	}
+	err = importer.CopyImage(dso)
```
zvikorn: So for the blank image feature, I'm using the image package where I will have a method to create the image in the qemu.go file. I do not need to use the DataStream file of course. Anyway... The  only parameters I will need to create the image are the imageSize (which will be another env var) and the common.ImporterWritePath. I do not want to pass the entire dso struct. Can we move the ImporterWritePath out of it?

pkg/controller/import-controller.go
```diff
@@ -187,8 +197,8 @@ func (ic *ImportController) syncPvc(key string) error {
 	if pvc == nil {
 		return nil
 	}
-	//check if AnnoEndPoint annotation exists
-	if !checkPVC(pvc, AnnEndpoint) {
+	//check if AnnEndPoint or AnnSource annotation exists
+	if !checkPVC(pvc, AnnEndpoint) && !checkPVC(pvc, AnnSource) {
```
zvikorn: If the EndPoint anno exists and Source anno does not, we still processing the PVC. I think the source annotation should be 'must' in order to keep processing the PVC. Otherwise it's very confusing. The source annotation is not used only for Enpoint case of course. Look also at my other comments related to this.

awels:
ci test please

pkg/controller/util.go
```diff
@@ -22,12 +22,30 @@ import (
...
+	// ImagePathName provides a const to use for creating volumes in pod specs
+	ImagePathName  = "image-path"
+	socketPathName = "socket-path"
+
+	// SourceHTTP is the source type HTTP, if unspecified or invalid, it defaults to SourceHTTP
+	SourceHTTP = "http"
+	// SourceS3 is the source type S3
+	SourceS3 = "s3"
+	// SourceGlance is the source type of glance
+	SourceGlance = "glance"
```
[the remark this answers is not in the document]
zvikorn: Yes. Otherwise I feel deprived (is this the appropriate word) since "blank" is not there :-)

pkg/controller/util.go
```diff
@@ -62,6 +80,45 @@ func getEndpoint(pvc *v1.PersistentVolumeClaim) (string, error) {
...
+		SourceS3,
+		SourceGlance,
+		SourceNone,
+		SourceRegistry:
+		glog.V(2).Infof("pvc source annotation found for pvc \"%s/%s\", value %s\n", pvc.Namespace, pvc.Name, source)
+	default:
+		glog.V(2).Infof("No valid source annotation found for pvc \"%s/%s\", default to http\n", pvc.Namespace, pvc.Name)
+		source = SourceHTTP
+	}
+	return source
```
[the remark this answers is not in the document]
awels: We made the decision to stay backwards compatible, so we default to http source is none is specified and fall back to the old behavior. Same for contentType if it is not there it defaults to kubevirt.

doc/annotations.md
```diff
@@ -0,0 +1,88 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: "example-pvc"
+  labels:
+    app: containerized-data-importer
+  annotations:
+    cdi.kubevirt.io/storage.import.source: "none"
+    cdi.kubevirt.io/storage.import.contentType: "kubevirt" #defaults to kubevirt if missing or invalid.
+    cdi.kubevirt.io/storage.import.endpoint: "http://www.source.example/path/of/data.tar" # http or https is supported
```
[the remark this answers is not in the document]
awels: Good point, will remove.

pkg/controller/util.go
```diff
@@ -62,6 +80,45 @@ func getEndpoint(pvc *v1.PersistentVolumeClaim) (string, error) {
...
+}
+
+// returns the source string which determines the type of source. If no source or invalid source found, default to http
+func getContentType(pvc *v1.PersistentVolumeClaim) string {
```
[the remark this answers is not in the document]
awels: Yes, if it is not there or invalid, we treat it as kubevirt, this way we maintain the current behavior.

pkg/controller/util.go
```diff
@@ -62,6 +80,45 @@ func getEndpoint(pvc *v1.PersistentVolumeClaim) (string, error) {
...
+		ContentTypeKubevirt,
+		ContentTypeArchive:
+		glog.V(2).Infof("pvc content type annotation found for pvc \"%s/%s\", value %s\n", pvc.Namespace, pvc.Name, contentType)
+	default:
+		glog.V(2).Infof("No content type annotation found for pvc \"%s/%s\", default to kubevirt\n", pvc.Namespace, pvc.Name)
+		contentType = ContentTypeKubevirt
+	}
+	return contentType
+}
+
```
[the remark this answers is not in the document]
awels: Same answer.

pkg/controller/import-controller.go
awels: See the other answers, its to maintain backwards compatibility. Remember fairly soon we will have proper data sources on PVCs themselves and we won't maintain the backwards compatibility at that point.

pkg/controller/util.go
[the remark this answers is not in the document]
awels: Blank is SourceNone, so I didn't forget you. I figured I would put in the constants for you guys so you didn't have to. If you prefer to add them, I can remove the ones that are not supported at this point.

pkg/controller/util.go
```diff
@@ -260,10 +317,18 @@ func MakeImporterPodSpec(image, verbose, pullPolicy string, podEnvVar importPodE
 // return the Env portion for the importer container.
 func makeEnv(podEnvVar importPodEnvVar, uid types.UID) []v1.EnvVar {
 	env := []v1.EnvVar{
+		{
+			Name:  common.ImporterSource,
+			Value: podEnvVar.source,
+		},
 		{
 			Name:  common.ImporterEndpoint,
```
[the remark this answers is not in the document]
awels: The importer expects the environment variable to be set to something, which includes "". It should know how to handle a blank end point.

tools/cdi-func-test-file-host-init/main.go
```diff
@@ -55,7 +53,7 @@ type formatTable [][]string
 
 func (ft formatTable) initializeTestFiles(inFile, outDir string) error {
 	var wg sync.WaitGroup
-	errChan := make(chan error, len(ft))
+	errChan := make(chan error, 3)
```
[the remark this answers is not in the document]
awels: Yeah you are right, I can make that a separate PR, what is happening is that currently it starts 9 threads that all start hammering the FS at once. I wanted to reduce the number of concurrent hits to something more manageable.

pkg/controller/util.go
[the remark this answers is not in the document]
zvikorn: Of course. But I thought not setting env var if not needed (""). But it's minor thing....

pkg/controller/util.go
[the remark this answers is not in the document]
zvikorn: sure

cmd/cdi-importer/importer.go
awels: Right, but you have a method signature something like
```go
func BlankImage(size int64) error {
  ...
}
```
Then you can check
```go
if source == SourceNone && contentType == ContentTypeKubevirt {
   BlankImage(pvcRequestSize)
}
```

zvikorn: The question is how to pass to BlankImage() the common.ImporterWritePath argument without passing the whole dso object...

zvikorn: oh I see. what you saying is that I still need the  source and contentType...

zvikorn: maybe change the structure name as it's not relevant to datastream only?

awels: No the check is done in the importer main, based on the values of the env variables you either call blank image or go through the normal path. Since common.ImporterWritePath is a global const, there is no need to pass it around. The only reason its part of the struct is so that the tests can pass in something different. If you need it for the tests, you can add a parameter to the BlankImage function.

awels: No the structure is for the DataStream info only. Blank image is a very special case that goes outside the normal flow completely.

doc/annotations.md
```diff
@@ -0,0 +1,88 @@
...
+### Source
+Source describes the type of data source CDI will be collecting the data from. Based on the value of source, additional annotations may be required to successfully import the data. The full annotation for source is: cdi.kubevirt.io/storage.import.source. The following values are currently available:
+* http
+* S3
+* none (don't import, but create data based on the contentType annotation)
+
+#### http and s3
+The http and s3 sources require an additional annotation to describe the end point CDI needs to connect to. The annotation is cdi.kubevirt.io/storage.import.endpoint. If the end point requires authentication one can add an optional annotation to point to a Kubernetes Secret to get authentication information from. This annotation is: cdi.kubevirt.io/storage.import.secretName. If the source annotation is missing it will default to "http". There is an additional annotation that determines the content type of the source, the content type can be one of the following:
+* kubevirt (Virtual Machine image)
+* archive (archive)
```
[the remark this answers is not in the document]
awels: That is a good point, and I struggled with that a little, I originally had content type as a separate section to explain what it was. But then I realized that contentType actually has different meaning based on which source we have selected. Http/S3 source then archive and kubevirt both have meaning, but None source, then archive has no meaning and kubevirt does. Same for some of the other sources coming up.

I then decided to have a sub section for the content type for each source section. I guess I didn't make that part clear enough in the document, I will make some modifications to make it clearer.

cmd/cdi-importer/importer.go
zvikorn: okay.

doc/annotations.md
```diff
@@ -0,0 +1,94 @@
+# Sources and contentType
+All annotations associated with Containerized Data Importer (CDI) have a prefix of: cdi.kubevirt.io. This denotes that the annotations are meant to be consumed by the CDI controller.
+
+## Source
+Source describes the type of data source CDI will be collecting the data from. Based on the value of source, additional annotations may be required to successfully import the data. The full annotation for source is: cdi.kubevirt.io/storage.import.source. The following values are currently available:
+* http
```
danielerez: maybe worth referring to Source constants in util file, so we won't have to update this file when adding a new one (or we prefer to keep the documentation neutral from the code?)

pkg/importer/dataStream.go
```diff
@@ -59,22 +60,37 @@ var qemuOperations = image.NewQEMUOperations()
...
 	rdr     io.ReadCloser
 }
 
+// DataStreamOptions contains all the values needed for importing from a DataStream.
+type DataStreamOptions struct {
+	// Dest is the destination path of the contents of the stream.
+	Dest string
+	// Endpoint is the endpoint to get the data from for various Sources.
+	Endpoint string
+	// AccessKey is the access key for the endpoint, can be blank.
```
danielerez: worth mentioning here that it should be encoded in base64? or is it enough to rely on the readme?

doc/annotations.md
awels: We will have to update this file when we add new sources, we don't want users to go and dive into sources to figure out which sources are valid.

awels:
So I think I have addressed anything, let me know if you see something @zvikorn @rollandf @j-griffith

pkg/importer/dataStream.go
j-griffith: @danielerez makes a good point, it probably would be good to add the b64 encoding requirement to the godoc strings.

pkg/controller/util.go
```diff
@@ -62,6 +80,45 @@ func getEndpoint(pvc *v1.PersistentVolumeClaim) (string, error) {
...
+		contentType = ""
+	}
+	switch contentType {
+	case
+		ContentTypeKubevirt,
+		ContentTypeArchive:
+		glog.V(2).Infof("pvc content type annotation found for pvc \"%s/%s\", value %s\n", pvc.Namespace, pvc.Name, contentType)
+	default:
+		glog.V(2).Infof("No content type annotation found for pvc \"%s/%s\", default to kubevirt\n", pvc.Namespace, pvc.Name)
+		contentType = ContentTypeKubevirt
```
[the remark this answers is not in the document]
awels: So to keep backwards compatibility (of not having the annotation at all) we default to kubevirt, which is what the original behavior is anyway. Would you prefer I check for blank, and set it to ContentTypeKubevirt and return an error for invalid values? Its possible I am just trying to see what benefit that would get us from the default behavior of just assuming its a disk image. This code would strip out anything nefarious by converting it to ContentTypeKubevirt.

pkg/importer/dataStream.go
awels: Sure, done.

zvikorn:
@awels Isn't DataVolume-controller supposed to changed too?

awels:
@zvikorn technically yes, practically no, since everything is backwards compatible, whatever DV is doing is still working. I could have it put in the two annotations for http sources, if we really want.

PR #522, opened 2018-11-14
Specify image size

### LINKED ISSUE #482: Add option to resize disk.img to actual DataVolume size

 /kind enhancement
There should be an option allowing to resize `disk.img` to actual DataVolume size created for that image file. This would be beneficial in a situation when one wants to use the same VM image for different services/environments with different storage requirements (Ex.: service A need 50GB, service B 75GB, VMs on dev need 25GB, on prod 100GB, etc). Option could be trigered by adding `resize` flag to DataVolumeTemplate:
```
  dataVolumeTemplates:
      spec:
        resize: true
```
By default `resize` should be set to false.

**What happened**:

**Environment**:
- Install tools:
- Others:

awels:
Combined code from #489 and #490 by
@gites and
@danielerez
Added some tests and rebased on current master.

This PR calculates the needed image size based on the PVC request size (Not the actual PVC size), and resizes the image to match.

Fixes #482 

This should finally get the feature in CDI.

**Release note**:

```release-note
Disk images are resized to the PVC request size.
```

Co-authored-by: Tomasz Pawelczak <gites@users.noreply.github.com>
Co-authored-by: Daniel Erez <danielerez@users.noreply.github.com>

### DIFF
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -34,6 +34,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"github.com/golang/glog"
 	"github.com/minio/minio-go"
 	"github.com/pkg/errors"
@@ -81,11 +83,7 @@ type DataStreamOptions struct {
 	Dest string
 	// Endpoint is the endpoint to get the data from for various Sources.
 	Endpoint string
-<<<<<<< HEAD
 	// AccessKey is the access key for the endpoint, can be blank. This needs to be a base64 encoded string.
-=======
-	// AccessKey is the access key for the endpoint, can be blank.
->>>>>>> Add source and contentType annotations
 	AccessKey string
 	// SecKey is the security key needed for the endpoint, can be blank.
 	SecKey string
@@ -93,6 +91,8 @@ type DataStreamOptions struct {
 	Source string
 	// ContentType is the content type of the data.
 	ContentType string
+	// ImageSize is the size we want the resulting image to be.
+	ImageSize string
 }
 
 const (
@@ -132,6 +132,7 @@ func newDataStreamFromStream(stream io.ReadCloser) (*DataStream, error) {
 		"",
 		controller.SourceHTTP,
 		controller.ContentTypeKubevirt,
+		"", // Blank means don't resize
 	}, stream)
 }
 
@@ -309,6 +310,29 @@ func SaveStream(stream io.ReadCloser, dest string) (int64, error) {
 		return 0, errors.Wrap(err, "data stream copy failed")
 	}
 	return ds.Size, nil
+}
+
+// ResizeImage resizes the images to match the requested size. Sometimes provisioners misbehave and the available space
+// is not the same as the requested space. For those situations we compare the available space to the requested space and
+// use the smallest of the two values.
+func ResizeImage(dest, imageSize string) error {
+	info, err := qemuOperations.Info(dest)
+	if err != nil {
+		return err
+	}
+	currentImageSizeQuantity := resource.NewQuantity(info.VirtualSize, resource.BinarySI)
+	newImageSizeQuantity := resource.MustParse(imageSize)
+	minSizeQuantity := util.MinQuantity(resource.NewScaledQuantity(util.GetAvailableSpace(dest), 0), &newImageSizeQuantity)
+	if minSizeQuantity.Cmp(newImageSizeQuantity) != 0 {
+		// Available dest space is smaller than the size we want to resize to
+		glog.Warningf("Available space less than requested size, resizing image to available space %s.\n", minSizeQuantity.String())
+	}
+	if currentImageSizeQuantity.Cmp(minSizeQuantity) == 0 {
+		glog.V(1).Infof("No need to resize image. Requested size: %s, Image size: %d.\n", imageSize, info.VirtualSize)
+		return nil
+	}
+	glog.V(1).Infof("Expanding image size to: %s\n", minSizeQuantity.String())
+	return qemuOperations.Resize(dest, minSizeQuantity.String())
 }
 
 // Read the endpoint and determine the file composition (eg. .iso.tar.gz) based on the magic number in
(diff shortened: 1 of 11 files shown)


awels:
ci test please

awels:
ci test please

awels:
ci test please

pkg/util/util.go
```diff
@@ -70,3 +72,18 @@ func (r *CountingReader) Read(p []byte) (n int, err error) {
 func (r *CountingReader) Close() error {
 	return r.Reader.Close()
 }
+
+// GetAvailableSpace gets the amount of available space at the path specified.
+func GetAvailableSpace(path string) int64 {
+	var stat syscall.Statfs_t
+	syscall.Statfs(path, &stat)
+	return int64(stat.Bavail) * int64(stat.Bsize)
```
gites: It looks like you forgot about filesystem overhead. This is a good place to subtract the 5% overhead.

aglitke: Looks good but let's see a test for this function to make sure.  You could consume some storage space on the FS and ensure that this function returns the precise amount left.

awels: Done

pkg/importer/dataStream.go
```diff
@@ -307,6 +312,27 @@ func SaveStream(stream io.ReadCloser, dest string) (int64, error) {
...
+	info, err := qemuOperations.Info(dest)
+	if err != nil {
+		return err
+	}
+	currentImageSizeQuantity := resource.NewQuantity(info.VirtualSize, resource.BinarySI)
+	newImageSizeQuantity := resource.MustParse(imageSize)
+	minSizeQuantity := util.MinQuantity(resource.NewScaledQuantity(util.GetAvailableSpace(dest), 0), &newImageSizeQuantity)
+	if minSizeQuantity.Cmp(newImageSizeQuantity) != 0 {
+		// Available dest space is smaller than the size we want to resize to
+		glog.V(1).Info("[WARN] Available space less than requested size, resizing image to available space %s.\n", minSizeQuantity.String())
```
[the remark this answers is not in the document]
awels: There is no warn in glog.V(1), but there is a glog.Warning, so using that.

pkg/importer/dataStream.go
```diff
@@ -307,6 +312,27 @@ func SaveStream(stream io.ReadCloser, dest string) (int64, error) {
 	return ds.Size, nil
 }
 
+// ResizeImage resizes the images to match the requested size.
```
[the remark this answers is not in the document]
awels: Sure, done

pkg/importer/dataStream.go
```diff
@@ -307,6 +312,29 @@ func SaveStream(stream io.ReadCloser, dest string) (int64, error) {
 	return ds.Size, nil
 }
 
+// ResizeImage resizes the images to match the requested size. Sometimes provisioners misbehave and the available space
+// is not the same as the requested space. For those situations we compare the available space to the requested space and
+// use the smallest of the two values.
+func ResizeImage(dest, imageSize string) error {
```
zvikorn: Is this being called somewhere?

zvikorn:
@awels where do we do the filesystem overhead calculation and subtract from requested size?

pkg/importer/dataStream.go
awels: Whoops, you are right it is not, but obviously it should, making a patch to fix now.

awels:
ResizeImage does the calculation. The algorithm is: take whatever is passed from the controller (The request size) and then compare to the available space on the file system of the PVC that is mounted in the container. Pick the lesser of the two.

zvikorn:
Yea..... I guess I only need to take the minimum between
'util.GetAvailableSpace(dest)'
and t he requested size, so that's fine. Thanks!

zvikorn:
@awels so.... I cannot use the GetAvailableSpace as the PVC is not bound yet at that point. I will subtract 4% for FS from the requested image size, unless we have something else to think about...

awels:
Not sure what you mean, in the importer the PVC is by definition bound, and that is where you have to do the calculation of which one to pick.

zvikorn:
yea, I take my words back. I believe the PVC is bound at that point.
Anyway, I do not understand the connection between the GetAvailableSpace result and the FS overhead we need to take into account. Could you please elaborate about this a bit? 
By the way, GetAvailableSpace returns '0' for me. Did it worked for you in the Resize patch? Did you verify the image size after resize is not '0'?

awels:
So we don't have to take the FS overhead into account anymore, since its mounted with a FS on it, the available reported by the FS is what is available. We are just taking the minimum of that that value and the requested image size.

awels:
Also just verified that I am getting the requested information from GetAvailableSpace, however found a bug that its not actually calling resize image at the right time, will have a PR to fix that soon.

zvikorn:
Ok. Thanks. I tried getAvailableSpace in qemu.go in the createBlankImage method just right before I’m calling the qemu-img and it return ‘0’. I will try it again and also will try it in the importer.

zvikorn:
I am running the following lines, and getAvailableSpace '0'. Do you know why?

if source == controller.SourceNone && contentType == controller.ContentTypeKubevirt {
		requestedImageSizeQuantity := resource.MustParse(imageSize)
		glog.V(3).Infof("TZVIKA1 requestedImageSizeQuantity quantity is  %s ", requestedImageSizeQuantity)
		x := resource.NewScaledQuantity(util.GetAvailableSpace(common.ImporterWritePath), 0)
		glog.V(3).Infof("TZVIKA1 availablespaces in int64 is  %s ", x)
		minSizeQuantity := util.MinQuantity(x, &requestedImageSizeQuantity)
		glog.V(3).Infof("TZVIKA1 minSizeQuantity is  %s ", minSizeQuantity)

		//imageSize = minSizeQuantity.String()
		//imageSize = "tzvika"

		err := image.CreateBlankImage(common.ImporterWritePath, imageSize)
		if err != nil {
			glog.Errorf("%+v", err)
			os.Exit(1)
		}
	}

zakkg3:
> Also just verified that I am getting the requested information from GetAvailableSpace, however found a bug that its not actually calling resize image at the right time, will have a PR to fix that soon.

Is this fixed ?
i have a dataVolumeTemplates requesting 250Gi. the pvc is created with this size. then it clones the disk from a source pvc in another namespace (10G) but it never  resizes the image. so my vm sees 10Gb instead of the 250...
Iam running cdi v1.19.0

awels:
So this particular issue is talking about the virtual disk size. We do NOT resize the partition inside the VM because we don't know which (if there are multiple) partition you would like to have resized. So if you run something like fdisk inside the VM, it will show the disk is the size you requested, however the partitions will not be.

awels:
Oh i missed you were cloning, so there is a known issue https://github.com/kubevirt/containerized-data-importer/issues/930 about resizing after cloning. In particular since we can clone any persistent volume, we cannot be sure we cloned a virtual disk. That is why we are not resizing after a clone. It is a definite issue we need to solve at some point.

PR #526, opened 2018-11-15
Check orphaned DataVolume PVCs for PODs

### LINKED ISSUE #525: Referenced POD object not deleted if failing / in retry-loop
/kind bug

**What happened**:
When you delete a DataVolume and a POD is in a retry/crash loop, the PVC goes to Terminating and is stuck there forever as the POD is never deleted and just continues retrying forever.

You CAN address this by issuing a kubectl delete against the POD, on the next restart interval the update will be detected and the POD finally destroyed, leading to deletion of the PVC completing as well.

Normally referenced objects are deleted as part of the deletion of their parent, regardless of the retry loops on the POD I would expect it to be deleted, then the PVC deleted and then finally the DataVolume deleted. 

Create a DataVolume with an invalid URL, thereby causing the importer POD to go into it's restart-on-failure loop.

Now, issue a delete of the Data Volume ``kubectl delete dv my-data-volume``

You'll note that the DataVolume is removed, the PVC goes to terminating but the POD just keeps on doing it's restarts.

Related issues:
#523 

Also opening another issue to rework our retry strategy 

**Environment**:
- CDI version Current Master

j-griffith:
This PR fixes the issue of DataVolume deletion leaving associated PVCs and PODs behind on the system.  For now we're just adding an extra check to deleted Data Volume PVCs and explicitly deleting their associated PODs regardless of their phase.

Fixes #525

This is a temporary fix, until a larger change is implemented around retries

**Release note**:

```release-note
NONE
```


j-griffith:
Added hold label until I add unit tests

PR #538, opened 2018-11-25
Validate whether there is enough space in PVC before converting qcow2…

annastopel:
… image, stored locally, to raw format

Fixes #

**Release note**:

```release-note
Validate whether there is enough space in PVC before converting qcow2 image (stored locally) into raw.
The validation is performed in order to prevent the conversion in case there is not enough space.

```


pkg/importer/dataStream.go
```diff
@@ -693,6 +704,58 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
 	return nil
 }
 
+//Validates wheather there is enough space in PVC for both qcow2 image and converted raw image
+//Assumes it is possible to retrieve info from qcow2 file by means of qemu-img utility
+//Returns failure if either of the following happens
+//1. info cannot be treieved from qcow2 file
```
awels: 1. info cannot be retrieved from qcow2 file

pkg/importer/dataStream.go
```diff
@@ -693,6 +704,58 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+//2. Either ActrualSize or VirtualSize are 0 - not specifed
```
awels: 2. Either ActualSize or VirtualSize are 0 - not specified

pkg/importer/dataStream.go
```diff
@@ -693,6 +704,58 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+//3. ActualSize and VirtualSize together exceed available PVC Space
+func ValidateSpaceConstraint(qcow2Image string, destDir string, qemu bool) (bool, error) {
```
awels: The caller of ValidateSpaceConstraint should know if they are in qemu mode or not. I think it makes sense to have them decided based on that flag to call ValidateSpaceConstraint, instead of making this a no-op based on that flag. So I would remove the qemu flag and check in this function and instead do the check in dataStream.go something like
```go
if qemu {
  isEnoughSpace, err := ValidateSpaceConstraint(dest, filepath.Dir(out))
  if err != nil {
    return err
  }
  if !isEnoughSpace {
    return errors.Wrap(err, "Local qcow2 to raw conversion failed - there is not enough space for conversion")
  }
}
```

pkg/importer/dataStream.go
```diff
@@ -693,6 +704,52 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+	imgVirtualSize, succeed := resource.NewQuantity(info.VirtualSize, resource.BinarySI).AsInt64()
+	if !succeed {
+		return false, errors.Wrap(err, "Validate enough space failed - invalid format of Virtual Size value")
+	}
+	imgActualSize, succeed := resource.NewQuantity(info.ActualSize, resource.BinarySI).AsInt64()
+	if !succeed {
+		return false, errors.Wrap(err, "Validate enough space failed - invalid format of Actual Size value")
+	}
+
+	if imgVirtualSize == 0 ||
```
aglitke: I'd prefer to see both predicates on the same line since it is not very long.

pkg/importer/dataStream_test.go
```diff
@@ -215,6 +236,7 @@ var _ = Describe("Copy", func() {
 
 	table.DescribeTable("Image, with import source should", func(dest, endpt string, qemuOperations image.QEMUOperations, wantErr bool) {
 		By("Configuring endpoint")
+
```
aglitke: I don't think you meant to add this whitespace.

pkg/importer/dataStream.go
```diff
@@ -703,6 +714,51 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+
+	if err != nil {
+		return false, errors.Wrap(err, "Local image validation failed to retrieve image info")
+	}
+
+	if info == nil {
+		return false, errors.Wrap(err, "Local image validation failed to retrieve image info")
+	}
+
+	imgVirtualSize, succeed := resource.NewQuantity(info.VirtualSize, resource.BinarySI).AsInt64()
```
awels: AFAIK info.VirtualSize is an int64 (according to the Struct info), so why are we converting to Quantity and then back to int64. I think we can skip that step can't we, or am I missing something?

pkg/importer/dataStream.go
```diff
@@ -703,6 +714,51 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+	if !succeed {
+		return false, errors.Wrap(err, "Validate enough space failed - invalid format of Virtual Size value")
+	}
+	imgActualSize, succeed := resource.NewQuantity(info.ActualSize, resource.BinarySI).AsInt64()
```
awels: Same with info.ActualSize, that is int64 already as well.

pkg/importer/dataStream.go
```diff
@@ -703,6 +714,45 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+	}
+
+	if info == nil {
+		return false, errors.Wrap(err, "Local image validation failed to retrieve image info")
+	}
+
+	imgVirtualSize := info.VirtualSize
+	imgActualSize := info.ActualSize
+
+	if imgVirtualSize == 0 || imgActualSize == 0 {
```
awels: Any reason you can just do:
```go
if info.VirtualSize == 0 || info.ActualSize == 0 {
  ...
}

convRequiredSpace := info.VirtualSize + info.ActualSize
```
You already have the info object, this is purely a style thing, as what you have works as well.

pkg/importer/dataStream.go
```diff
@@ -703,6 +714,45 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+//Returns failure if either of the following happens
+//1. info cannot be retrieved from qcow2 file
+//2. Either ActualSize or VirtualSize are 0 - not specifed
+//3. ActualSize and VirtualSize together exceed available PVC Space
+func ValidateSpaceConstraint(qcow2Image string, destDir string) (bool, error) {
+
+	//Verify there is enough space in pvc before conversion
+	info, err := qemuOperations.Info(qcow2Image)
+
+	if err != nil {
```
awels: qemuOperations.Info guarantees that if err != nil then info == nil and if err == nil then info != nil. So you don't need both checks.

awels:
Looks like we have some conflicts in the tests now.

pkg/importer/dataStream.go
annastopel: if I check only that info != null then i assume that method Info(..) does not return error otherwise. Currently, this is the case, but if the method info(..) is changed so that err can be returned even though info is not nill, then there can be a bug in ValidateSpaceInfo and there will be no clear indication for that.

annastopel:
Conflicts in the tests resolved

pkg/importer/dataStream.go
awels: Fair enough.

PR #539, opened 2018-11-26
Fixes incorrect call of resize.

awels:

Resize was not being called in all paths, specifically the streaming path was not working.
Updated unit and functional tests to take resize into account.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -25,6 +25,8 @@ import (
 	"strconv"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
@@ -56,7 +58,7 @@ type ImgInfo struct {
 type QEMUOperations interface {
 	ConvertQcow2ToRaw(string, string) error
 	ConvertQcow2ToRawStream(*url.URL, string) error
-	Resize(string, string) error
+	Resize(string, resource.Quantity) error
 	Info(string) (*ImgInfo, error)
 	Validate(string, string) error
 }
@@ -111,12 +113,18 @@ func (o *qemuOperations) ConvertQcow2ToRawStream(url *url.URL, dest string) erro
 	return nil
 }
 
-func (o *qemuOperations) Resize(image, size string) error {
+// convertQuantityToQemuSize translates a quantity string into a Qemu compatible string.
+func convertQuantityToQemuSize(size resource.Quantity) string {
 	// size from k8s contains an upper case K, so we need to lower case it before we can pass it to qemu.
 	// Below is the message from qemu that explains what it expects.
 	// qemu-img: Parameter 'size' expects a non-negative number below 2^64 Optional suffix k, M, G, T, P or E means kilo-, mega-, giga-, tera-, peta-and exabytes, respectively.
-	size = strings.Replace(size, "K", "k", -1)
-	_, err := qemuExecFunction(qemuLimits, nil, "qemu-img", "resize", "-f", "raw", image, size)
+	stringSize := strings.Replace(size.String(), "K", "k", -1)
+	stringSize = strings.Replace(stringSize, "i", "", -1)
+	return stringSize
+}
+
+func (o *qemuOperations) Resize(image string, size resource.Quantity) error {
+	_, err := qemuExecFunction(qemuLimits, nil, "qemu-img", "resize", "-f", "raw", image, convertQuantityToQemuSize(size))
 	if err != nil {
 		return errors.Wrapf(err, "Error resizing image %s", image)
 	}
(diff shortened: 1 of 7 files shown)


awels:
ci test please

awels:
ci test please

pkg/image/qemu.go
```diff
@@ -111,12 +113,18 @@ func (o *qemuOperations) ConvertQcow2ToRawStream(url *url.URL, dest string) erro
...
 	// qemu-img: Parameter 'size' expects a non-negative number below 2^64 Optional suffix k, M, G, T, P or E means kilo-, mega-, giga-, tera-, peta-and exabytes, respectively.
-	size = strings.Replace(size, "K", "k", -1)
-	_, err := qemuExecFunction(qemuLimits, nil, "qemu-img", "resize", "-f", "raw", image, size)
+	stringSize := strings.Replace(size.String(), "K", "k", -1)
+	stringSize = strings.Replace(stringSize, "i", "", -1)
+	return stringSize
+}
+
+func (o *qemuOperations) Resize(image string, size resource.Quantity) error {
+	_, err := qemuExecFunction(qemuLimits, nil, "qemu-img", "resize", "-f", "raw", image, quantityToQemu(size))
```
j-griffith: Oh, this is the only place we're using that conversion call?  Mabye we don't even need the wrapper, but anyway... it would be nice to give it a better name IMO.

pkg/image/qemu_test.go
```diff
@@ -181,6 +183,15 @@ var _ = Describe("Report Progress", func() {
 	})
 })
 
+var _ = Describe("quantity to qemu", func() {
+	It("Should properly parse quantity to qemu", func() {
+		result := quantityToQemu(resource.MustParse("1Gi"))
```
j-griffith: Never hurts to test go's strings.Replace function I suppose.

pkg/importer/dataStream.go
```diff
@@ -320,19 +320,22 @@ func ResizeImage(dest, imageSize string) error {
...
-	newImageSizeQuantity := resource.MustParse(imageSize)
-	minSizeQuantity := util.MinQuantity(resource.NewScaledQuantity(util.GetAvailableSpace(dest), 0), &newImageSizeQuantity)
-	if minSizeQuantity.Cmp(newImageSizeQuantity) != 0 {
-		// Available dest space is smaller than the size we want to resize to
-		glog.Warningf("Available space less than requested size, resizing image to available space %s.\n", minSizeQuantity.String())
-	}
-	if currentImageSizeQuantity.Cmp(minSizeQuantity) == 0 {
-		glog.V(1).Infof("No need to resize image. Requested size: %s, Image size: %d.\n", imageSize, info.VirtualSize)
-		return nil
+	if imageSize != "" && info != nil {
```
j-griffith: can info be nil here?  It seems that would be an error in the qemu.Info function, the only reason I ask is the alternate return from this block would be incorrect then.

pkg/importer/dataStream.go
```diff
@@ -320,19 +320,22 @@ func ResizeImage(dest, imageSize string) error {
...
+		if currentImageSizeQuantity.Cmp(minSizeQuantity) == 0 {
+			glog.V(1).Infof("No need to resize image. Requested size: %s, Image size: %d.\n", imageSize, info.VirtualSize)
+			return nil
+		}
+		glog.V(1).Infof("Expanding image size to: %s\n", minSizeQuantity.String())
+		return qemuOperations.Resize(dest, minSizeQuantity)
 	}
-	glog.V(1).Infof("Expanding image size to: %s\n", minSizeQuantity.String())
-	return qemuOperations.Resize(dest, minSizeQuantity.String())
+	return errors.New("Image resize called with blank resize")
```
j-griffith: Or if "info == nil", if that's a possible thing

pkg/image/qemu.go
awels: Its the only place right now, we need it for instance for the blank image as well, and probably in the import from registry.

pkg/image/qemu.go
```diff
@@ -111,12 +113,18 @@ func (o *qemuOperations) ConvertQcow2ToRawStream(url *url.URL, dest string) erro
 	return nil
 }
 
-func (o *qemuOperations) Resize(image, size string) error {
+// QuantityToQemu translates a quantity string into a Qemu compatible string.
+func quantityToQemu(size resource.Quantity) string {
```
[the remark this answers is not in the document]
awels: Well its converting the quantity string to a qemy compatible string. Not sure if convertSizeSuffix is correct either.

pkg/image/qemu_test.go
awels: Well we are testing that I am correctly stripping i and lowercasing k.

pkg/importer/dataStream.go
awels: no, since err will be != nil if it is, and thus it will never get here.

pkg/importer/dataStream.go
awels: its not

pkg/importer/dataStream.go
```diff
@@ -648,7 +651,12 @@ func (d *DataStream) copy(dest string) error {
 		if err != nil {
 			return errors.Wrap(err, "Streaming qcow2 to raw conversion failed")
 		}
-
+		if d.ImageSize != "" {
```
awels: @j-griffith This was missing before, so it wouldn't resize in this path.

pkg/importer/dataStream.go
```diff
@@ -683,12 +691,14 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
 			return errors.Wrap(err, "Local qcow to raw conversion failed")
 		}
-		if imageSize != "" {
-			err = ResizeImage(dest, imageSize)
-		}
-		if err != nil {
-			return errors.Wrap(err, "Resize of image failed")
-		}
+		// Set the dest to out, since we just wrote that
+		dest = out
```
awels: This sets it so that once the file is downloaded and converted (non streaming path) that the actual output file is resized instead of the temp file.

pkg/importer/dataStream.go
```diff
@@ -683,12 +691,14 @@ func copy(r io.Reader, out string, qemu bool, imageSize string) error {
...
+	}
+	if imageSize != "" {
```
awels: This makes sure that converted ISO files also get resized.

awels:
@j-griffith I put some comments on the patch that show you where we were missing calls.

PR #545, opened 2018-11-28
Host-assisted cloning runs after PVC is already cloned #534

### LINKED ISSUE #534: Host-assisted cloning runs after PVC is already cloned
/kind bug

**What happened**:
Host-assisted cloning runs on a PVC which was already cloned by the gluster subvol provisioner.

Host-assisted cloning should not run on PVCs which already carry the CloneOf annotation.

- Begin with a development environment using ```make cluster-up cluster-sync```
- Create 10 PVCs and 10 Pods that use the PVCs in order to use up all PVs
-  Create PVC with CloneRequest annotation to clone one of the dummy PVCs
- The PVC is in pending state and clone pods are scheduled to run
- Simulate the completion of a clone by adding a CloneOf annotation to the PVC
- Delete one of the dummy PVCs
- Clone PVC binds to a PV and cloning pods run against the bound PVC

The above steps simulate the behavior without needing to deploy the gluster subvol provisioner.  I think the proper fix for this issue is to not schedule the clone pods unless the target PVC is either Bound or Pending/WaitFirstConsumer.  At that point you can check for the CloneOf annotation and if it's missing you will be confident that the provisioner will not actually do a clone itself.

**Environment**:
- Install tools:
- Others:

zvikorn:

This PR fixes issue #534 
Host-assisted cloning runs on a PVC which was already cloned by the gluster subvol provisioner.

Fixes #534 

To test the fix works (we test it with local storage since gluster subvol is not necessary for this test):
- Begin with a development environment using make cluster-up cluster-sync
- Create and deploy a storageclase with no volumeBindingMode field since the existing PV's contain 
  this field with the value 'WaitForFirstConsumer'.
- Create a PV with the storageclass name you just created. Do NOT deploy the PV yet.
- Create and deploy 10 PVCs to import disk image. 10 importer Pods will run and complete. All 10 
  PV's are bound to the 10 PVC's.
- Create and deploy a cloner PVC with CloneRequest annotation and storageclass with the name of 
   the storageclass you created above to clone one of the dummy PVCs. 
- Since there is no available PV, the PVC wil be in 'Pending' phase. Cloner pods will NOT be spawned 
  with this fix. This is the important thing to verify! Cloner pods are NOT spawned.
- Edit the deployed cloner PVC and add the "k8s.io/CloneOf: true" annotation to indicate that the clone 
  was done by external provisioner (smart cloning).
- Deploy the PV you created above to make the cloner PVC bound to this PV.
- VERIFY cloner pods are NOT spawned (this is because the CloneOf annotation).

```release-note
Fixing issue #534 
```


pkg/controller/clone-controller.go
```diff
@@ -230,6 +231,23 @@ func (cc *CloneController) syncPvc(key string) error {
 	if !checkPVC(pvc, AnnCloneRequest) {
 		return nil
 	}
+
+	pvcPhase := pvc.Status.Phase
+	glog.V(3).Infof("PVC phase for PVC \"%s/%s\" is ", pvc.Namespace, pvc.Name, pvcPhase)
```
awels: I think you are missing a %s in the log message.

pkg/controller/clone-controller.go
```diff
@@ -230,6 +231,23 @@ func (cc *CloneController) syncPvc(key string) error {
...
+	storageClassName := *pvc.Spec.StorageClassName
+	glog.V(3).Infof("storageClassName used by PVC \"%s/%s\" is \"%s\"", pvc.Namespace, pvc.Name, storageClassName)
+	storageclass, err := cc.clientset.StorageV1().StorageClasses().Get(storageClassName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	//Do not schedule the clone pods unless the target PVC is either Bound or Pending/WaitFirstConsumer.
+	if !(pvcPhase == v1.ClaimBound || (pvcPhase == v1.ClaimPending && *storageclass.VolumeBindingMode == storageV1.VolumeBindingWaitForFirstConsumer)) {
+		glog.V(3).Infof("PVC \"%s/%s\" is either not bound or is in pending phase and VolumeBindingModein is not VolumeBindingWaitForFirstConsumer."+
```
awels: ... and VolumeBindingMode is ...

pkg/controller/clone-controller.go
zvikorn: Thanks.

pkg/controller/clone-controller.go
zvikorn: added log

pkg/controller/clone-controller.go
```diff
@@ -230,6 +231,23 @@ func (cc *CloneController) syncPvc(key string) error {
 	if !checkPVC(pvc, AnnCloneRequest) {
 		return nil
 	}
+
+	pvcPhase := pvc.Status.Phase
+	glog.V(3).Infof("PVC phase for PVC \"%s/%s\" is ", pvc.Namespace, pvc.Name, pvcPhase)
+	storageClassName := *pvc.Spec.StorageClassName
```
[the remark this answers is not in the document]
zvikorn: Fixed. Ready for another review. Thank you!

PR #547, opened 2018-11-29
Import docker registry func test

danielerez:

Added a test for registry source: existence of the bin file in the image (i.e. /pvc/usr/bin/cdi-importer).
A complementary PR to https://github.com/kubevirt/containerized-data-importer/pull/517

Fixes #

**Release note**:

```release-note
None
```


awels:
Looks like the failure is actually related to your test?

danielerez:
@awls - yes, seems like it. But I'm not sure why it fails only on openshift and passes successfully on k8s. Do you have any idea what could be different between these envs? I'll try to test it locally on os env..

awels:
Not a 100% sure, but usually if it sometimes fails, its because something hasn't finished yet before we check the status. Usually calling a waitForPhase or status to make sure whatever we are running is finished will fix that. Or an Eventually will also work, there are lots of examples of us using those as well.

PR #553, opened 2018-12-02
skopeo: use util UnArchiveTar

danielerez:

Added UnArchiveLocalTar func to util for using in skoepo (instead of ExtractTar func).

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/image/skopeo.go
+++ b/pkg/image/skopeo.go
@@ -28,6 +28,7 @@ import (
 	"github.com/pkg/errors"
 
 	"kubevirt.io/containerized-data-importer/pkg/system"
+	"kubevirt.io/containerized-data-importer/pkg/util"
 )
 
 const dataTmpDir string = "/data_tmp"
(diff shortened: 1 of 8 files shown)


danielerez:
@igoihman - nope, this PR is about removing the ExtractTar func from skopeo by using UnArchiveTar in util. PR #556 is about fixing changes in tests introduced on PR #547.

pkg/image/skopeo.go
```diff
@@ -127,11 +119,12 @@ var extractImageLayers = func(dest string) error {
 			layerID = m.Digest
 		}
 		layer := strings.TrimPrefix(layerID, "sha256:")
-		file := fmt.Sprintf("%s%s/%s", dest, dataTmpDir, layer)
-		err := ExtractTar(file, dest)
-		if err == nil {
-			err = cleanWhiteoutFiles(dest)
+		filePath := fmt.Sprintf("%s%s/%s", dest, dataTmpDir, layer)
+
+		if err := util.UnArchiveLocalTar(filePath, dest, "z"); err != nil {
```
awels: Are we guaranteed that filePath will be local, I tried something similar, and it kept feeling when we were using a registry as  the part of the file

danielerez: We are using skopeo to download the tar layers into a local temp folder, then extract it into data folder. In which scenario did it fail for you?

PR #559, opened 2018-12-05
Blank images always being created with size of 0

mhenriks:

Blank image always size 0 because function that calculates free space was referencing a non existent file. 

Fixes #

**Release note**:

```release-note

```


mhenriks:
@awels @aglitke thinking should we backport this fix to release 1.4.0

awels:
So the actual fix is changing the writepath to the voumepath, because the writepath is returning 0 and thus min will also return 0?

mhenriks:
@awels yeah GetAvailableSpace() was returning 0 when passed `/data/disk.img`.

PR #573, opened 2018-12-17
Clone sometimes doesn't add k8s.io/CloneOf annotation to target PVC #560

### LINKED ISSUE #560: Clone sometimes doesn't add k8s.io/CloneOf annotation to target PVC

/kind bug

**What happened**:
Sometimes when cloning a PVC the cloneOf annotation does not get added to the target PVC. I believe the following sequence of events happens:
1. target PVC is created
2. Controller sees new PVC, and creates source/target pods. And puts Running annotation on target PVC.
3. The clone completes. And the target pod is cleaned up before the clone controller is allowed to run. The clone controller doesn't see the completed pod, and thus never updates the annotation on the target PVC.

The cleanup of the finished pod will trigger the annotation to be put on the target PVC.

Since this is a race condition (does the pod get cleaned before the clone controller gets a chance to run, its hard to reproduce). The CI does fail on it sometimes. I will attach a failed run log to this bug.

**Environment**:
master
all supported versions have the issue.
- Install tools:
- Others:

zvikorn:

Fixing issue #560 Clone sometimes doesn't add k8s.io/CloneOf annotation to target PVC

Fixes #560 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -96,6 +96,7 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	if err != nil {
 		return err
 	}
+
 	pvcKey, err := cache.MetaNamespaceKeyFunc(pvc)
 	if err != nil {
 		return err
@@ -158,7 +159,6 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	//add the following annotation only if the pod pahse is succeeded, meaning job is completed
 	if phase == string(v1.PodSucceeded) {
 		anno[AnnCloneOf] = "true"
-		defer cc.deleteClonePods(sourcePod.Namespace, sourcePod.Name, targetPod.Name)
 	}
 	var lab map[string]string
 	if !checkIfLabelExists(pvc, common.CDILabelKey, common.CDILabelValue) {
@@ -167,6 +167,8 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	pvc, err = updatePVC(cc.clientset, pvc, anno, lab)
 	if err != nil {
 		return errors.WithMessage(err, "could not update pvc %q annotation and/or label")
+	} else if pvc.Annotations[AnnCloneOf] == "true" {
+		cc.deleteClonePods(sourcePod.Namespace, sourcePod.Name, targetPod.Name)
 	}
 	return nil
 }


pkg/controller/clone-controller.go
```diff
@@ -158,7 +167,6 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	//add the following annotation only if the pod pahse is succeeded, meaning job is completed
 	if phase == string(v1.PodSucceeded) {
 		anno[AnnCloneOf] = "true"
-		defer cc.deleteClonePods(sourcePod.Namespace, sourcePod.Name, targetPod.Name)
```
mhenriks: I don't see how this change ensures that the "cloneof" annotation gets applied.  It just changes the order of operations from delete pods/update pvc to update pvc/delete pods.  The controller still needs to see the pod enter "succeeded" state for the annotation to be applied.

In a separate but possibly related issue, I think it is a mistake for to always initialize `anno` to an empty map here: https://github.com/kubevirt/containerized-data-importer/blob/64cc6e7068dd29dd45db8c3ad3966b1beb885c73/pkg/controller/clone-controller.go#L92

I wonder if it is possible that the controller is actually setting the annotation then removing it when the pod enters "terminating" state.   That is just speculation.  But a definite issue is that we may be precluding users from adding their own annotations to the PVC.

By suggestion is to simply change line 92 to `anno := pvc.Annotations` and see if that fixes the issue.  If not, something else is at play.

pkg/controller/clone-controller.go
```diff
@@ -158,7 +167,6 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
...
 	}
 	var lab map[string]string
```
mhenriks: Separate issue, but I think it would also be wise to initialize `lab` as `lab := pvc.Labels`

pkg/controller/clone-controller.go
j-griffith: @mhenriks I had the same challenge, but it sounded like the if condition was sometimes causing this to be missed?  By skipping that if clauses an deferring until exit of the function it seems to eliminate the race and deletion of the pod "before" the annotation is updated.

So to make sure my interpretation matches:
1. Pod goes to completed status
2. Pod gets deleted
3. Controller looks for pod and it's status to make annotation on PVC
4. Pod no longer exists so we never update the PVC

Although... what I'm not sure is how the deferred delete call always being called, vs in the if block (which is also deferred) changes this.

There is likely a better fix.

pkg/controller/clone-controller.go
```diff
@@ -91,11 +91,20 @@ func (cc *CloneController) findClonePodsFromCache(pvc *v1.PersistentVolumeClaim)
 func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	anno := map[string]string{}
 
+	deferFunc := func(cc *CloneController, anno map[string]string, sourcePod *v1.Pod, targetPod *v1.Pod) {
+		if anno[AnnCloneOf] == "true" {
```
j-griffith: This check in the deferred deletePod call is what fixes the bug.  We work the opposite direction and we don't delete the pod until the annotation is set, so this way subsequent updates don't delete the pod before the annotation is set.

This might cause problems in other places, but it does fix up the bug.

zvikorn:
The defer was executed immediately when we exit the if clause. So the pods were deleted before the pvc was updated. Now wit the fix, the defer will be executed only if the pvc was updated.

zvikorn:
Right!

zvikorn:
Hard for me to see where it can break other places :-)

pkg/controller/clone-controller.go
mhenriks: @j-griffith the controller deletes the pods and updates the pvc in the scope of the same function.  I don't see where your step 3 happens.

j-griffith:
> Hard for me to see where it can break other places :-)
> […](#)
> Sent from my iPhone
> On 17 Dec 2018, at 18:59, John Griffith ***@***.***> wrote: @j-griffith commented on this pull request. In pkg/controller/clone-controller.go: > @@ -91,11 +91,20 @@ func (cc *CloneController) findClonePodsFromCache(pvc *v1.PersistentVolumeClaim) func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error { anno := map[string]string{} +	deferFunc := func(cc *CloneController, anno map[string]string, sourcePod *v1.Pod, targetPod *v1.Pod) { +	if anno[AnnCloneOf] == "true" { This check in the deferred deletePod call is what fixes the bug. We work the opposite direction and we don't delete the pod until the annotation is set, so this way subsequent updates don't delete the pod before the annotation is set. This might cause problems in other places, but it does fix up the bug. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or mute the thread.

I'm not sure; but the concern is around things like failed clones; this change relies on the clone annotation to cleanup/delete pods so my concern there was around orphaned pods but I think we handle that in other places (existing fail path).

zvikorn:
ci test please

pkg/controller/clone-controller.go
mhenriks: I guess I don't follow.  I understand that the pods would now be deleted after the pvc is updated.  And that may helps if say the controller crashes after the pods are deleted (and the pvc never gets updated and another controller takes over somehow).  But the controller isn't crashing as far as I know.  And the conditions that set the "cloneof" annotation are the same.  I don't see how deleting the pods a couple miliseconds after the pvc helps us in the non-failure case.

zvikorn:
@j-griffith I thought about orphaned pods. This can happen only if the updatePVC fails and return error. In this case we will have a new set of cloning pods. However, the annotation itself assigned with 'CloneOf' (not the pvc) when the old pods complete! we check the annotation and then delete the pods. I hope I got you...

zvikorn:
@mhenriks @j-griffith Apparently I'm not following as well. So, I will describe the bug that Alexander opened and my code change to fix it. Please let me know which scenario can be missed/broken....
The bug is that when the clonning pods complete, they are deleted before the PVC gets updated with the 'CloneOf' annotation. Looking at the code, I figured out what causes this. The reason the pods get deleted before the PVC gets updated, is that there is defer call to delete the pods inside an 'if' clause before we update the PVC with the annotation. When the 'if' clause ends, the defer call is executed, and the pods are deleted (the scope of the defer call is the if clause). At that point, the pvc was not updated yet, and will not be updated as the pods do not exist anymore. The fix is, to move the defer call outside the 'if' clause and only when the PVC is updated. Before we delete the pods, we check if the 'CloneOf' annotation exists of course. The defer call will be executed at the END of the function. Meaning, the PVC will be updated with the 'CloneOf' annotation for sure, unless there is an error during the update.
In a case like that, the pods are completed anyway, and the defer call will be executed at the end of the function as the annotation map was assigned with the 'CloneOf' value (even though the PVC was not updated).
Could you please describe where this code change does not fix the bug?

j-griffith:
This was my understanding as well, and why I thought the changes fixed the problem:

> The bug is that when the clonning pods complete, they are deleted before the PVC gets updated with the 'CloneOf' annotation.

pkg/controller/clone-controller.go
j-griffith: @mhenriks Yeah, my understanding is that there's a race here and in a normal flow the pod is being deleted before the annotation is updated on the PVC, and that's what's being addressed here.

mhenriks:
@zvikorn @j-griffith  this code will delete the pods if the update of the PVC fails because the defered function checks the `anno` map and not the actual PVC returned by the call to update.

zvikorn:
@mhenriks @j-griffith correct. However, as far as I know, two new pods will be created as a result of the error that was returned

zvikorn:
@mhenriks @j-griffith changed the code.

pkg/controller/clone-controller.go
```diff
@@ -167,6 +167,10 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 	pvc, err = updatePVC(cc.clientset, pvc, anno, lab)
 	if err != nil {
 		return errors.WithMessage(err, "could not update pvc %q annotation and/or label")
+	} else {
+		if phase == string(v1.PodSucceeded) {
```
mhenriks: I think this is a step in the right direction.  Some style comments:

I think an `else if` would be better than separate statements.  I also think it maybe be better to change the condition to check `pvc.Annotations[AnnCloneOf] == "true"` or have a local `deletePods` bool that is set to true when `anno[AnnCloneOf] = "true"`.

pkg/controller/clone-controller.go
```diff
@@ -167,6 +167,10 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
...
+			cc.deleteClonePods(sourcePod.Namespace, sourcePod.Name, targetPod.Name)
```
mhenriks: maybe out of the scope for this, but if either of the pod deletes fails, they will never be retried.

zvikorn:
cli test please

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
ci test please

Issue #359, opened 2018-08-20
Common tags for ginkgo based tests, including functional(e2e) and unit tests


jeffvance:
This issue is here to start the definition of tags for all tests (functional and unit) that are based on the ginkgo framework. Ginkgo allows test filtering based on using regular expressions in the `skip=` and `focus=` test invocation args.  Eg. `ginkgo test ... -focus="\[e2e\]"`.

These tags can be simple/flat, eg _[e2e]_, _[importer]_, _[cloner]_, etc.  Or they can be hierarchical, eg: _[e2e/importer]_, _[e2e.clonerer]_, etc. However, there is no real need for hierarchical tags since multiple tags can be used in the ginko skip= and focus= args.

Proposal:
- [CLONER] - all cloner tests match this tag
- [DATAVOLUME] - all dataVolume tests match this tag.
- [E2E] - all functional tests match this tag
- [FLAKY] - tests that fail periodically usually for unknown reasons and require further debugging. Useful to skip these tests in CI. 
- [HW-DEP] - eg fast clone test would set this tag
- [IMPORTER] - all importer tests match this tag
- [SERIAL] - tests cannot be run in parallel
- [SLOW] - set of tests that have longer timeouts, sleep, or other factor that can result in slow completion times
- [SW-DEP] - tests that require specific libs installed on the node(s)
- [UNIT] - all unit tests match this tag
- [UPLOADER] - all uploader tests match this tag

An example of running only e2e cloner tests:
` ginkgo test ... -focus="\[E2E\]" -focus="\[CLONER\]"`

jeffvance:
@nellyc @shiywang @copejon @screeley44 @davidvossel @awels

nellyc:
Sounds good! 
Just 2 comments:
* I think that 'E2E' should be renamed to 'Functional'
* I dont really like the idea of a Flaky tests tag, as we may end up forgetting about them ;) in my POV a flaky test should be either fixed ASAP or redesigned if we cannot fix it (basically we dont get any value from such tests a)

@fabiand what do you say about similar tags for kubevirt? (excluding feature specific ones)

awels:
One more to add [OC] means its an openshift specific test, that requires either an open shift client or oc instead of kubectl.

To help polarion we would also like to put on context or it level another 'tag' that is more complex that contain a set of values to help polarion organize the tests. The format would look something like this:
[...]

PR #583, opened 2018-12-26
Add test attributes

nellyc:
Changed all test to have rfe id, test id, criticality etc attributes

based on issue
https://github.com/kubevirt/containerized-data-importer/issues/359

Updated most of the existing tests, so we will be able to auto create them in Polarion from the code.
There are 2 steps to it:
first, we create the tests without test ids (if they are new tests)
after running the script which creates the tests in Polarion (kubevirt/qe-tools/polarion-generator), each one gets its own ID,
and we can than add the test ids to the test code (this will allow reporting & changing the test in the future).

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/api_validation_test.go
+++ b/tests/api_validation_test.go
@@ -22,10 +22,10 @@ const (
 	destinationFile    = "/var/tmp/datavolume_test.yaml"
 )
 
-var _ = Describe("Validation tests", func() {
+var _ = Describe("[rfe_id:1130, crit:medium, vendor:cnv-qe@redhat.com, level:component]Validation tests", func() {
 	f := framework.NewFrameworkOrDie("api-validation-func-test")
 
-	Describe("Verify DataVolume validation", func() {
+	Describe("[posneg:negative]Verify DataVolume validation", func() {
 		Context("when creating Datavolume", func() {
 			dv := map[string]interface{}{}
 
@@ -70,12 +70,12 @@ var _ = Describe("Validation tests", func() {
 				}, timeout, pollingInterval).Should(BeTrue())
 
 			},
-				table.Entry("fail with http source with invalid url format", "http", invalidURLFormat),
-				table.Entry("fail with http source with empty url", "http", ""),
-				table.Entry("fail with s3 source with invalid url format", "s3", invalidURLFormat),
-				table.Entry("fail with s3 source with empty url", "s3", ""),
-				table.Entry("fail with empty PVC source namespace", "pvc", "", "test-pvc"),
-				table.Entry("fail with empty PVC source name", "pvc", "test", ""),
+				table.Entry("[test_id:1321]fail with http source with invalid url format", "http", invalidURLFormat),
+				table.Entry("[test_id:1322]fail with http source with empty url", "http", ""),
+				table.Entry("[test_id:1323, crit:low]fail with s3 source with invalid url format", "s3", invalidURLFormat),
+				table.Entry("[test_id:1324, crit:low]fail with s3 source with empty url", "s3", ""),
+				table.Entry("[test_id:1325]fail with empty PVC source namespace", "pvc", "", "test-pvc"),
+				table.Entry("[test_id:1326]fail with empty PVC source name", "pvc", "test", ""),
 			)
 
 			table.DescribeTable("with Datavolume PVC size should", func(size string) {
@@ -97,9 +97,9 @@ var _ = Describe("Validation tests", func() {
 					return false
 				}, timeout, pollingInterval).Should(BeTrue())
 			},
-				table.Entry("fail with zero PVC size", "0"),
-				table.Entry("fail with negative PVC size", "-500m"),
-				table.Entry("fail with invalid PVC size", "invalid_size"),
+				table.Entry("[test_id:1033]fail with zero PVC size", "0"),
+				table.Entry("[test_id:1327]fail with negative PVC size", "-500m"),
+				table.Entry("[test_id:1328]fail with invalid PVC size", "invalid_size"),
 			)
 
 		})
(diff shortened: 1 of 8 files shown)


igoihman:
please note that changing test levels to focus causes the CI to ignore the other tests.
Toy need to run tests with focus flag if you want to run a specific subset.
https://onsi.github.io/ginkgo/#focused-specs

nellyc:
> please note that changing test levels to focus causes the CI to ignore the other tests.
> You need to run tests with focus flag if you want to run a specific subset.
> https://onsi.github.io/ginkgo/#focused-specs

@igoihman yes, my intention is to update all existing tests with these attributes & set them to focused,
but wanted to get an ack first :)

igoihman:
@nellyc you don't have to use FDescribe for that.
Running them with -ginkgo.focus=<regex> is enough.

nellyc:
@igoihman I believe I removed all the 'F's, as you suggested

igoihman:
@nellyc please ignore my last comment

nellyc:
ci test please

j-griffith:
Hi @nellyc, I was curious:

> irst, we create the tests without test ids (if they are new tests)
> after running the script which creates the tests in Polarion (kubevirt/qe-tools/polarion-generator), each one gets its own ID,
> and we can than add the test ids to the test code (this will allow reporting & changing the test in the future).

So that implies a second/external step where we run through the RedHat internal test tool, and then manually use that data to update the source later?  I'm wondering if there's another way to do this, like perhaps generate/assign test-ids from CDI automatically?  Part of the reason is that a two step manual process is likely to not work great (lazy people like myself will not update things); but more importantly given that Polarion is an internal test tool that doesn't mean much to the upstream community does it make sense to add those ID's back in?  

Test ID's are definitely a useful addition, I just wonder if we should work from CDI outwards here.

One other question I had was regarding the annotations around "crit" etc, how are those determined?  What's their significance from the perspective of somebody that just does `git clone; make cluster-up/sync; make test-all` or something along those lines?

nellyc:
> Hi @nellyc, I was curious:
Hey!

> 
> > irst, we create the tests without test ids (if they are new tests)
> > after running the script which creates the tests in Polarion (kubevirt/qe-tools/polarion-generator), each one gets its own ID,
> > and we can than add the test ids to the test code (this will allow reporting & changing the test in the future).
> 
> So that implies a second/external step where we run through the RedHat internal test tool, and then manually use that data to update the source later? I'm wondering if there's another way to do this, like perhaps generate/assign test-ids from CDI automatically? Part of the reason is that a two step manual process is likely to not work great (lazy people like myself will not update things); but more importantly given that Polarion is an internal test tool that doesn't mean much to the upstream community does it make sense to add those ID's back in?

yes, that is correct, ATM its a manual step.
and you are actually raising an interesting point :)
i was thinking to add the execution of the script as part of the repo CI (in std-ci) and maybe later on, to even parse the response and create a script to auto update the IDs,
but your idea may be simpler. 

if we want, we can assign IDs and create the tests based on a pre-defined custom field.
the only thing we need to take care of, is making sure the ids we are putting are unique. 
how do you suggest to do that?

> 
> Test ID's are definitely a useful addition, I just wonder if we should work from CDI outwards here.
> 
> One other question I had was regarding the annotations around "crit" etc, how are those determined? What's their significance from the perspective of somebody that just does `git clone; make cluster-up/sync; make test-all` or something along those lines?

it has 2 purposes:
1. you can chose to run with Focus only high criticality tests, or by any other attribute
2. when creating tests, we must add these attributes to Polarion
how do we determine them? there should be a conversation between the developer and his qe :) so if qe are CRing the feature tests, we can give that feedback

tests/api_validation_test.go
```diff
@@ -97,9 +97,9 @@ var _ = Describe("Validation tests", func() {
 					return false
 				}, timeout, pollingInterval).Should(BeTrue())
 			},
-				table.Entry("fail with zero PVC size", "0"),
-				table.Entry("fail with negative PVC size", "-500m"),
-				table.Entry("fail with invalid PVC size", "invalid_size"),
+				table.Entry("[test_id:1033]fail with zero PVC size", "0"),
```
shiywang: for those cases, I have old case CNV-1027 to cover, are we sure we gonna use a new id instead ?

tests/api_validation_test.go
```diff
@@ -117,7 +117,7 @@ var _ = Describe("Validation tests", func() {
 			err := utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolumeName)
 			Expect(err).ToNot(HaveOccurred())
 		})
-		It("should fail creating an already existing DataVolume", func() {
+		It("[test_id:1030]should fail creating an already existing DataVolume", func() {
```
shiywang: Same, so I understand for new case, we can definitely use qe-tools to genereate, but for old case still same?

shiywang: Okay, after checking your another comment, I see, I think (1030-1033) are fully covered, we can mark my cases to be inactive and generate new one

nellyc: for old TCs we can just use the existing ID
so if now you are going to automated test case number 123,
just add the test_id attribute on your test, it will not create a new one when executing the qe-tools script, but instead will update this TC, with the test automation code (may result in rename, test steps will be overridden with whatever we have in 'by', link to rfe, criticality etc....)

nellyc: BTW, yes, we probably have some duplication now,
if you have a manual test in Polarion, which is actually already automated,
please add 'markfordeletion' tag to it,
i will do a clean up soon and delete all tests with this tag

awels:
@nellyc We just added one more test, which is causing a conflict.

PR #588, opened 2019-01-02
Fix some functional test issues running locally.

awels:

Fixes some functional tests that can fail intermittently.
Fixes path issues on local machine.

Fixes #

**Release note**:

```release-note
NONE
```


tests/api_validation_test.go
```diff
@@ -24,8 +26,23 @@ const (
...
+
 	f := framework.NewFrameworkOrDie("api-validation-func-test")
 
+	dir, err := os.Getwd()
+	if err != nil {
+		Fail("Unable to get current directory")
+	}
+
+	// If we are in the 'tests' directory already, go up a level. This happens in the dev environment.
+	if strings.HasSuffix(dir, "tests") {
```
mhenriks: Is there any way we can address in bash somewhere?  Like in `run-tests.sh`?  Thinking for the case when we are running in ci, we can cd to the tests dir?  This doesn't seem like a sustainable pattern.

awels: I looked at that first, the problem is that in CI we call 
make functional-test-ci which generates a tests binary, and it is run from the root of the project. But when running in dev environment we call go test tests/... which apparently cds into the tests directory and then runs the tests. Not sure how we would get out of that from bash.

mhenriks: Yeah I ran into this the other day, and here's how I got around it:  https://gist.github.com/mhenriks/d25916aadd369d1d30e362618ce3cde2

That got `make test-functional` working for me and should still be good for ci.  Maybe there's an even better way to do it?

awels: Actually that looks good to me, and avoids me having to do silly stuff like I am doing there.

awels:
ci test please

mhenriks:
ci test please

PR #589, opened 2019-01-03
Adding DataVolume support for unpacking imported archive

zvikorn:

Adding DataVolume support for unpacking imported archive

**Release note**:

```release-note
Adding DataVolume support for unpacking imported archive
```

### DIFF
--- a/pkg/apis/datavolumecontroller/v1alpha1/types.go
+++ b/pkg/apis/datavolumecontroller/v1alpha1/types.go
@@ -42,7 +42,19 @@ type DataVolumeSpec struct {
 	Source DataVolumeSource `json:"source"`
 	//PVC is a pointer to the PVC Spec we want to use
 	PVC *corev1.PersistentVolumeClaimSpec `json:"pvc"`
+	//DataVolumeContentType options: "kubevirt", "archive"
+	ContentType DataVolumeContentType `json:"contentType,omitempty"`
 }
+
+// DataVolumeContentType represents the types of the imported data
+type DataVolumeContentType string
+
+const (
+	// DataVolumeKubeVirt is the content-type of the imported file, defaults to kubevirt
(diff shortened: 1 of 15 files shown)


awels:
@zvikorn Looks like you need to make generate, to update some generated stuff?

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -42,6 +42,8 @@ type DataVolumeSpec struct {
 	Source DataVolumeSource `json:"source"`
 	//PVC is a pointer to the PVC Spec we want to use
 	PVC *corev1.PersistentVolumeClaimSpec `json:"pvc"`
+	//ContentType options: "kubevirt", "archive"
+	ContentType string `json:"contentType,omitempty"`
```
mhenriks: I think it would be better to create a new type here.  So something like this:

```
type DataVolumeContentType string

const (
    DataVolumeKubeVirt = "kubevirt"
    DataVolumeArchive = "archive"
)
```

zvikorn: @mhenriks so the following consts are defined in pkg/controller/util.go:
// ContentTypeKubevirt is the content-type of the import, defaults to kubevirt
ContentTypeKubevirt = "kubevirt"
// ContentTypeArchive is the content-type to specify if wanting to extract an archive
ContentTypeArchive = "archive"

and there are several references to these two, from other files. Should I remove these two consts from utils.go as well and use the new consts I'm going to add in types.go instead?

mhenriks: Yeah I think using api type definitions would be best.

zvikorn: Done

zvikorn:
ci test  please

PR #595, opened 2019-01-07
Update Datavolume status to include progress

awels:

This PR adds DataVolume status field for import/clone progress.

Fixes #

**Release note**:

```release-note
DataVolume status now has a progress field that is updated during import/clone.
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -17,15 +17,22 @@ limitations under the License.
 package controller
 
 import (
+	"crypto/tls"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"reflect"
+	"regexp"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
@@ -42,6 +49,7 @@ import (
 	cdischeme "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned/scheme"
 	informers "kubevirt.io/containerized-data-importer/pkg/client/informers/externalversions/core/v1alpha1"
 	listers "kubevirt.io/containerized-data-importer/pkg/client/listers/core/v1alpha1"
+	"kubevirt.io/containerized-data-importer/pkg/common"
 	expectations "kubevirt.io/containerized-data-importer/pkg/expectations"
 )
 
@@ -115,6 +123,8 @@ const (
 	// MessageUploadSucceeded provides a const to form upload has succeeded message
 	MessageUploadSucceeded = "Successfully uploaded into %s"
 )
+
+var httpClient *http.Client
 
 // DataVolumeController represents the CDI Data Volume Controller
 type DataVolumeController struct {
@@ -354,6 +364,9 @@ func (c *DataVolumeController) syncHandler(key string) error {
 			c.pvcExpectations.CreationObserved(key)
 			return err
 		}
+		if canUpdateProgress(newPvc.Annotations) {
+			go c.scheduleProgressUpdate(dataVolume.Name, dataVolume.Namespace, pvc.GetUID())
+		}
 	}
 
 	// Finally, we update the status block of the DataVolume resource to reflect the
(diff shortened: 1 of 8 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -363,6 +373,28 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
+		time.Sleep(2 * time.Second)
+		dataVolume, err := c.cdiClientSet.Cdi().DataVolumes(dataVolumeNamespace).Get(dataVolumeName, metav1.GetOptions{})
+		if k8serrors.IsNotFound(err) {
+			// Data volume is no longer there, or not found.
+			glog.V(3).Info("DV is gone, cancelling update thread.")
+			return
+		} else if err != nil {
+			glog.Errorf("error retrieving data volume %+v", err)
+		}
+		if dataVolume.Status.Phase == cdiv1.Succeeded {
```
j-griffith: I was thinking about our discussion around handling transient states on these, since we have this annoying case where the DV can go in and out of failed state, would it make sense to try and keep some kind of reference counter for that and then we can expire the go routine if it become hopeless?  

I'm not sure what the best answer would be, but something I just thought of (may or may not work) would be an annotation acting as a counter on the dv maybe?  Or  we do something like an annotation with a time-stamp (lastDetectedFailure?) where we see a volume has been in fail for more than some arbitrary amount of time we terminate the routine here?  If it goes to another state before expiration we clear the annotation

pkg/controller/datavolume-controller.go
```diff
@@ -363,6 +373,28 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
+			// Data volume completed progress, or failed, either way stop queueing the data volume.
+			glog.V(3).Info("**** completed imported, or failed, cancelling update thread: " + dataVolume.Status.Phase)
```
j-griffith: If we do something like above I won't have to guess which one it is :)

pkg/controller/datavolume-controller.go
```diff
@@ -524,6 +559,94 @@ func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolu
...
+		}
+	}
+}
+
+func (c *DataVolumeController) updateProgressUsingPod(dataVolumeCopy *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim, pod *corev1.Pod) {
+	httpClient := buildHTTPClient()
+	// Example value: import_progress{ownerUID="b856691e-1038-11e9-a5ab-525500d15501"} 13.45
+	var importRegExp = regexp.MustCompile("progress\\{ownerUID\\=\"" + string(dataVolumeCopy.UID) + "\"\\} (\\d{1,3}\\.?\\d*)")
+
+	url := fmt.Sprintf("https://%s:8443/metrics", pod.Status.PodIP)
```
j-griffith: This reminds me, should we look at making this port configurable?

pkg/apis/datavolumecontroller/v1alpha1/types.go
```diff
@@ -95,7 +95,8 @@ type DataVolumeSourceHTTP struct {
 // DataVolumeStatus provides the parameters to store the phase of the Data Volume
 type DataVolumeStatus struct {
 	//Phase is the current phase of the data volume
-	Phase DataVolumePhase `json:"phase,omitempty"`
+	Phase    DataVolumePhase    `json:"phase,omitempty"`
+	Progress DataVolumeProgress `json:"progress,omitempty"`
```
j-griffith: so when looking at the status if this is set I see something like:

> Progress:  15.03

It might be nice instead to use something like:

>PercentageDownloaded: 15.03

It's sort of interesting though, I have a failure for  URL not found, but according to the status I was at 20+ before it failed and started crash loop.

awels: This is why I threw it up here, I am not completely happy with it yet, but having trouble determining all the nifty edge cases. I think it would make sense to have a retry counter in the status as well. Should be relatively simple to add, simply count the crash loops.

As for the name, this will be used for cloning and other types as well, where 'Download' doesn't make sense, so I went with the more generic 'progress' I make it a float, but maybe string with a % at the end would make sense?

pkg/controller/datavolume-controller.go
awels: We can keep variables outside of the loop to keep some state, so there is no need to put things in annotations to keep state. The tricky part would be, how do we define what is 'long' enough in either counter or time. But those are interesting ideas, let me mull them over.

pkg/controller/datavolume-controller.go
awels: Well instead of hard coding, I should probably look it up from the service. In theory its already configurable from the service.

j-griffith: +1, yeah pulling it from the service is way more clear (for me at least)

pkg/apis/datavolumecontroller/v1alpha1/types.go
j-griffith: `PercentagePopulated: xxx` ?

Honestly not a big deal, just wasn't intuitive when I first saw it is all.

awels: I talked to Matt about it, and he honestly couldn't think of a better name besides progress due to the nature of the source of the data, it could be clone/import/upload/whatever. He did want me to make it a string and append % to make it clearer what it is.

j-griffith: fine by me

awels:
Yeah I know, once I get a chance to work on it again, I will rebase and update.

pkg/controller/datavolume-controller.go
```diff
@@ -517,6 +558,12 @@ func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolu
 		}
 	}
 
+	if pvc != nil && canUpdateProgress(pvc.Annotations) {
+		pod, err := c.getPodFromPvc(pvc.Namespace, pvc.GetUID())
+		if err == nil {
+			c.updateProgressUsingPod(dataVolumeCopy, pod)
```
mhenriks: Not sure it is wise to update the progress here as well as in `scheduleProgressUpdate()`.  Don't forget that every time `scheduleProgressUpdate()` updates the progress, that will trigger the DataVolume to be reconciled again.  So we'll basically be double updating the progress.

pkg/controller/datavolume-controller.go
```diff
@@ -367,6 +380,32 @@ func (c *DataVolumeController) syncHandler(key string) error {
 	return nil
 }
 
+func (c *DataVolumeController) scheduleProgressUpdate(dataVolumeName, dataVolumeNamespace string, pvcUID types.UID) {
+	for {
+		time.Sleep(2 * time.Second)
+		dataVolume, err := c.cdiClientSet.Cdi().DataVolumes(dataVolumeNamespace).Get(dataVolumeName, metav1.GetOptions{})
```
mhenriks: Should we maybe use the dataVolumesLister here instead?  So we're not always going to the API server?

pkg/controller/datavolume-controller.go
awels: Yeah I have been seeing the double updates, so that is a good point.

pkg/controller/datavolume-controller.go
awels: yeah good plan.

PR #600, opened 2019-01-09
generate cdi-controller.yaml from resources defined in code

mhenriks:

This is a stop-gap on the way to a functional cdi operator.  The operator works best with resources defined in code.  But we still need to get work done while the operator is completed/integrated.  So, let's generate cdi-controller.yaml from code.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/operator/resources/cluster/controler.go
```diff
@@ -0,0 +1,152 @@
+/*
```
aglitke: name this controller.go please.

pkg/operator/resources/cluster/crd.go
```diff
@@ -0,0 +1,160 @@
+/*
```
aglitke: Should this be named datavolume.go in case we will have other crds in the future?

pkg/operator/resources/cluster/crd.go
```diff
@@ -0,0 +1,160 @@
...
+												"name",
+											},
+										},
+										"upload": extv1beta1.JSONSchemaProps{},
+										"blank":  extv1beta1.JSONSchemaProps{},
+									},
+								},
+								"pvc": extv1beta1.JSONSchemaProps{
+									Properties: map[string]extv1beta1.JSONSchemaProps{
+										"resources": extv1beta1.JSONSchemaProps{
```
aglitke: Is there a reason that we are not using the real PVC types in here?

pkg/operator/resources/namespaced/apiserver.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func createExtensionAPIServerRoleBinding(serviceAccountNamespace string) *rbacv1beta1.RoleBinding {
+	roleBinding := createRoleBinding(
+		extensionAPIResourceName,
+		extensionAPIResourceName,
+		apiServerRessouceName,
+		serviceAccountNamespace,
+	)
+	roleBinding.Namespace = "kube-system"
```
aglitke: Should have a constant for this.  I'm hoping that with the operator we will adopt the 'cdi' namespace as our default rather than kube-system'.

pkg/operator/resources/namespaced/apiserver.go
```diff
@@ -0,0 +1,124 @@
...
+				IntVal: 8443,
+			},
+			Protocol: corev1.ProtocolTCP,
+		},
+	}
+	return service
+}
+
+func createAPIServerDeployment(repo, image, tag, verbosity, pullPolicy string) *appsv1.Deployment {
+	deployment := createDeployment(apiServerRessouceName, "cdi.kubevirt.io", apiServerRessouceName, apiServerRessouceName, 1)
```
aglitke: constant for cdi.kubevirt.io

tools/manifest-generator/manifest-generator.go
```diff
@@ -66,15 +82,102 @@ func main() {
...
-	tmpl := template.Must(template.ParseFiles(*templFile))
+	tmpl := template.Must(template.ParseFiles(templFile))
 	err = tmpl.Execute(os.Stdout, data)
 	if err != nil {
 		glog.Fatalf("Error executing template: %v\n", err)
 	}
 }
+
+func generateFromCode(codeGroup string) {
+	var resources []interface{}
```
aglitke: This is to suspend type checking on the array right?  Why aren't all members of this array a Resource?

pkg/operator/resources/cluster/controler.go
mhenriks: oops

pkg/operator/resources/cluster/crd.go
mhenriks: This is open api validation stuff which works at the json level.  Maybe @igoihman can provide more context

pkg/operator/resources/namespaced/apiserver.go
mhenriks: no this has to be kube-system.  need to read a specific configmap that lives there

tools/manifest-generator/manifest-generator.go
mhenriks: at this layer we don't really care about the resource types (which are different for cluster or namespaced).  we just want to marshall the objects to json.  By converting to interface{} we can put cluster and namespaced resources in the same slice.

pkg/operator/resources/cluster/crd.go
mhenriks: eh, maybe hopefully other crds won't be so heinous

pkg/operator/resources/namespaced/apiserver.go
awels: common.CDIComponentLabel is the constant name.

pkg/operator/resources/cluster/apiserver.go
```diff
@@ -0,0 +1,87 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cluster
+
+import (
+	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
```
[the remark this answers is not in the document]
mhenriks: I was just bringing over the exact resources from the manifest.  I'm not sure if there are any differences between the versions for the resources we use in this package.  But if you think it would be wise, I can bump the versions.

pkg/operator/resources/cluster/common.go
```diff
@@ -0,0 +1,64 @@
+/*
```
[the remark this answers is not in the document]
mhenriks: Yeah, I think roles.go would be fine for this specific file.

pkg/operator/resources/namespaced/common.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var commonLabels = map[string]string{
+	"cdi.kubevirt.io": "",
```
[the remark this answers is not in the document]
mhenriks: I share your concern about kitchen sink files, but I personally find too many files to be distracting.  The goal of this file is to provide a bunch of utility functions for creating k8s resources.  I could break it up into rbac.go/deployment.go/service.go/labels.go.  But that seems a little much to me.  What would you suggest?

glide.yaml
```diff
@@ -30,3 +30,5 @@ import:
   version: ^1.0.0
 - package: sigs.k8s.io/controller-runtime
   version: v0.1.9
+- package: k8s.io/apiextensions-apiserver
+  version: kubernetes-1.12.4
```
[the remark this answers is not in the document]
mhenriks: yeah, the operator scaffolding updated the rest of the client libraries to 12.4.

PR #605, opened 2019-01-13
refactor data stream copy 

igoihman:
We have a lot of code duplicity in data stream and the code is messy.
Organized copy function so that we can add more import types to this function (registry for example).

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -682,80 +682,83 @@ func (d *DataStream) isHTTPQcow2() bool {
 		len(d.Readers) == 2
 }
 
+func (d *DataStream) convertQcow2ToRawStream(dest string) error {
+	glog.V(3).Infoln("Validating qcow2 file")
+	err := qemuOperations.Validate(d.url.String(), "qcow2")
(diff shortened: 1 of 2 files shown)


pkg/importer/dataStream.go
```diff
@@ -682,29 +682,84 @@ func (d *DataStream) isHTTPQcow2() bool {
...
+	glog.V(3).Infoln("Doing streaming qcow2 to raw conversion")
+	err = qemuOperations.ConvertQcow2ToRawStream(d.url, dest)
+	if err != nil {
+		return errors.Wrap(err, "Streaming qcow2 to raw conversion failed")
+	}
+
+	return nil
+}
+
+func (d *DataStream) convertRawToQcow2(src, dest string) error {
```
zvikorn: @igoihman  Should not these methods be in utils as generic methods and not relevant only to datastream?
Convert images
validate images
Resize images...

igoihman: they are wrappers for other functions and specific for datastream, they are not being used elsewhere.

PR #607, opened 2019-01-15
Add a new custom resource for CDI user config

igoihman:
Add a new Custom Resource defining user configuration for CDI.
For now it's used for holding upload proxy URL details, making it easier for users to get this details without having to access ingress/route details.

CDIConfig is created when CDI is deployed.
config controller is watching routes, ingresses and CDIConfig resources.
In case we create a new ingress or route, config status will store upload proxy URL.

Fixes #

**Release note**:

```release-note
Added custom resource for CDI user config, currently holding upload proxy url.
```


awels:
Did you run make generate?

pkg/apis/core/v1alpha1/types.go
```diff
@@ -186,3 +186,18 @@ type CDIList struct {
 	// Items provides a list of CDIs
 	Items []CDI `json:"items"`
 }
+
+//CDIUserConfig provides a user configuration for CDI
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type CDIUserConfig struct {
```
mhenriks: Maybe `CDIConfig` is a better name?

pkg/apis/core/v1alpha1/types.go
```diff
@@ -186,3 +186,18 @@ type CDIList struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec CDIUserConfigSpec `json:"spec"`
```
mhenriks: Should have `Spec` and a `Status` members.  Maybe look something like this:

```
type CDIConfigSpec struct {
     UploadProxyURLOverride *string //if controller can't derive the URL or it's incorrect, user can override
}

type CDIConfigStatus struct {
     UploadProxyURL *string //this is the field users look at to find the upload proxy url.  Controller sets it to UploadProxyURLOverride if set, otherwise attempts to discover it
}
```

pkg/apis/core/v1alpha1/types.go
```diff
@@ -186,3 +186,33 @@ type CDIList struct {
...
+
+//CDIConfig provides a user configuration for CDI
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type CDIConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec CDIConfigSpec `json:"spec"`
+}
```
zvikorn: You forgot to add the 'CDIConfigStatus' member ...

pkg/operator/resources/cluster/cdiconfig.go
```diff
@@ -0,0 +1,54 @@
+package cluster
+
+import (
+	extv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func createcDIConfigCRD() *extv1beta1.CustomResourceDefinition {
```
zvikorn: Typo: createcDIConfigCRD

awels:
@igoihman I think you forgot to run make generate, travis is failing on the verification of that.

awels:
ci test please

cmd/cdi-controller/controller.go
```diff
@@ -101,10 +111,16 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
 	serviceInformerFactory := k8sinformers.NewFilteredSharedInformerFactory(client, common.DefaultResyncPeriod, "", func(options *v1.ListOptions) {
 		options.LabelSelector = common.CDILabelSelector
 	})
+	ingressInformerFactory := k8sinformers.NewFilteredSharedInformerFactory(client, common.DefaultResyncPeriod, "", func(options *v1.ListOptions) {
+		options.LabelSelector = common.CDILabelSelector
+	})
+	routeInformerFactory := routeinformers.NewSharedInformerFactory(openshiftClient, common.DefaultResyncPeriod)
```
zvikorn: Will this work in case openshiftClient is nil?

cmd/cdi-controller/controller.go
```diff
@@ -142,6 +161,10 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
 	go pvcInformerFactory.Start(stopCh)
 	go podInformerFactory.Start(stopCh)
 	go serviceInformerFactory.Start(stopCh)
+	go ingressInformerFactory.Start(stopCh)
+	if isOpenshift:= controller.IsOpenshift(client) ; isOpenshift {
+		go routeInformerFactory.Start(stopCh)
```
zvikorn: if we start this only if it's openshift, why don't we set routeInformerFactory,  routeInformer also conditionally?

pkg/controller/upload-controller.go
```diff
@@ -254,6 +320,48 @@ func (c *UploadController) initCerts() error {
...
+			break
+		}
+	}
+	namespaces, err := getAllSupportedNamespaces(c.client)
+	if err != nil {
+		return err
+	}
+	if url == "" {
+		for _, namespace := range namespaces {
+			ingresses, err := c.client.ExtensionsV1beta1().Ingresses(namespace).List(metav1.ListOptions{
```
awels: Isn't it possible to have multiple ingresses? For different namespaces. This code takes the first it finds and finishes. Shouldn't we pass a list of ingress URLs to CreateCDIConfig with all the results found, or create one config for each namespace?

cmd/cdi-controller/controller.go
zvikorn: Don't we want to call isOpenShift before setting this,as you call it before you start routeInformerfactory later on...

cmd/cdi-controller/controller.go
```diff
@@ -101,10 +111,16 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
...
+	ingressInformerFactory := k8sinformers.NewFilteredSharedInformerFactory(client, common.DefaultResyncPeriod, "", func(options *v1.ListOptions) {
+		options.LabelSelector = common.CDILabelSelector
+	})
+	routeInformerFactory := routeinformers.NewSharedInformerFactory(openshiftClient, common.DefaultResyncPeriod)
 
 	pvcInformer := pvcInformerFactory.Core().V1().PersistentVolumeClaims()
 	podInformer := podInformerFactory.Core().V1().Pods()
 	serviceInformer := serviceInformerFactory.Core().V1().Services()
+	ingressInformer := ingressInformerFactory.Extensions().V1beta1().Ingresses()
+	routeInformer := routeInformerFactory.Route().V1().Routes()
```
zvikorn: same here...

cmd/cdi-controller/controller.go
```diff
@@ -142,6 +161,10 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
 	go pvcInformerFactory.Start(stopCh)
 	go podInformerFactory.Start(stopCh)
 	go serviceInformerFactory.Start(stopCh)
+	go ingressInformerFactory.Start(stopCh)
```
zvikorn: Do we always need to start this? even if it's openshift and we use routeInformerFactory?

pkg/controller/util.go
```diff
@@ -861,6 +868,62 @@ func MakeUploadServiceSpec(name string, pvc *v1.PersistentVolumeClaim) *v1.Servi
 	return service
 }
 
+// CreateCDIConfig creates cdi config manifest and sends to server
+func CreateCDIConfig(cdiClient clientset.Interface, name, url string) (*cdiv1.CDIConfig, error) {
```
awels: So I think we can have multiple urls (one for each namespace) and thus we have to either pass in a namespace and have a config per namespace or have the config in the cdi namespace (like we have here) and pass in multiple urls (maybe a map with ns/url pairs?) and store that in the config.

pkg/controller/upload-controller.go
mhenriks: Assuming CDI is only running in a single namespace, there should not be ingresses in any other namespaces.  There may be multiple in the same namespace though but prob unlikely.

pkg/controller/util.go
mhenriks: There should only be one CDI config with a single url

cmd/cdi-controller/controller.go
igoihman: it doesn't work until we run the go routine

cmd/cdi-controller/controller.go
igoihman: yes, we support both ingress and route on openshift

igoihman:
@awels I ran `make generate`.
The cause for the failure was missing `openshift/client-go` package

pkg/controller/upload-controller.go
```diff
@@ -21,18 +21,29 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
+
```
mhenriks: I kind of hate to bring this up at this point but I wonder if it makes more sense to have a dedicated cdiconfig controller rather than rolling this into the upload controller?  I could imagine exposing more stuff through cdiconfig.  Could also allow users to set configuration params as well.  What do others think?

igoihman: I tend to agree with you. We took the upload controller approach since we only have upload proxy related stuff but I think it needs its own controller.

cmd/cdi-controller/controller.go
zvikorn: I know. My intention was why we set it unconditionally, where we might not execute the go routine...

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,313 @@
...
+	c.handleObject(obj, "add")
+}
+func (c *ConfigController) handleObjUpdate(obj interface{}) {
+	c.handleObject(obj, "update")
+}
+func (c *ConfigController) handleObjDelete(obj interface{}) {
+	c.handleObject(obj, "delete")
+}
+
+func (c *ConfigController) handleObject(obj interface{}, verb string) {
```
mhenriks: These callback functions really just decide whether the Route/Ingress/whatever is relevant to us (refers to our service) and if so, enqueue something to be processed a worker thread.  Everything else is ignored.

I don't actually see any queuing happening anywhere in this file so `processNextWorkItem()` must never be getting called

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,313 @@
...
+
+func (c *ConfigController) syncHandler(key string) error {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(errors.Errorf("invalid resource key: %s", key))
+		return nil
+	}
+
+	// Get the CDIConfig resource with this namespace/name
+	config, err := c.configLister.CDIConfigs(namespace).Get(name)
```
mhenriks: You have to figure out what object(s) you will be queueing.  Is it the CDIConfig or or the Ingress/Routes?

If you queue the CDIConfig, you fill have to go through all the routes/ingresses here.  If you queue the ingress/routes, I think you have to have multiple handler functions/queues.

pkg/operator/resources/cluster/cdiconfig.go
```diff
@@ -0,0 +1,31 @@
...
+		},
+		Spec: extv1beta1.CustomResourceDefinitionSpec{
+			Group: "cdi.kubevirt.io",
+			Names: extv1beta1.CustomResourceDefinitionNames{
+				Kind:     "CDIConfig",
+				Plural:   "cdiconfigs",
+				Singular: "cdiconfig",
+			},
+			Version: "v1alpha1",
+			Scope:   "Namespaced",
```
mhenriks: Should be cluster scoped

mmazur:
From the pov of code that interfaces with the cdi upload proxy, two configuration details are required:
1. URL to upload to
2. CA cert (location) for httpS connectivity

Currently I've basically hardcoded `verify_ssl: false` into my examples of cdi upload proxy usage. I'm hoping to drop that eventually. Is making that happen a part of this PR?

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,376 @@
...
+		go wait.Until(c.runWorker, time.Second, stopCh)
+	}
+
+	glog.Info("Started workers")
+	<-stopCh
+	glog.Info("Shutting down workers")
+	return nil
+}
+
+func (c *ConfigController) createCDIConfig(name string) error {
```
mhenriks: I think this function only has to create a empty/default CDIConfig if one doesn't exist.  Updating the object will happen in the reconcile loop.  I would try running this controller for awhile and having it pick up a route.  Then stop the controller and delete/recreate the route with a different name.  Then restart the controller.  Pretty sure the change will get picked up.

pkg/controller/util.go
```diff
@@ -861,6 +868,67 @@ func MakeUploadServiceSpec(name string, pvc *v1.PersistentVolumeClaim) *v1.Servi
...
+	ns := util.GetNamespace()
+	cfg := MakeCDIConfigSpec(name, ns, url)
+
+	cfg, err := cdiClient.CdiV1alpha1().CDIConfigs(ns).Create(cfg)
+	if err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			cfg, err = cdiClient.CdiV1alpha1().CDIConfigs(ns).Get(name, metav1.GetOptions{})
```
mhenriks: Should this be a `Get` or an `Update`?

igoihman:
@mmazur let's discuss it and I will add it in a separate PR

mmazur:
@igoihman well, it comes down to the fact that without a way to verify the connection, the PVC upload is open to (a) intercepting the data by a third party and (b) modifying it by a third party while getting uploaded, neither of which is acceptable in a production environment.

To avoid that, downstream code when obtaining the upload url needs to be able to also (**securely**) retrieve a certificate used for verifying the https connection (aka the CA cert). Because if it's unable to do so (which is currently the case), it needs to assume that whatever tls credentials it gets from the https server are to be trusted. Even if it says "Issued by Hackers, Inc.", which it might if the connection gets intercepted with a man in the middle attack.

I haven't dived into the code of the upload proxy pod, but I'm assuming it generated itself a fresh new tls cert for the https server when starting up. Those things are text files on the order of a few kilobytes tops. I don't know if that's too much to embed directly in CDIConfig or not – ideally they would be, so that it'd contain both an url and the cert for verification.

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,376 @@
...
+		object, ok = tombstone.Obj.(metav1.Object)
+		if !ok {
+			runtime.HandleError(errors.Errorf("error decoding object tombstone, invalid type"))
+			return
+		}
+		glog.V(3).Infof("Recovered deleted object '%s' from tombstone", object.GetName())
+	}
+	glog.V(3).Infof("Processing object: %s", object.GetName())
+
+	config, err := c.cdiClientSet.CdiV1alpha1().CDIConfigs(util.GetNamespace()).Get("config", metav1.GetOptions{})
```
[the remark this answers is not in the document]
igoihman: done

pkg/controller/util.go
igoihman: correct, changed to `Update`

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{})
+	if err != nil {
+		return err
+	}
+
+	// Get the CDIConfig resource with this namespace/name
+	var config *cdiv1.CDIConfig
+	configs, err := c.configLister.CDIConfigs(namespace).List(selector)
```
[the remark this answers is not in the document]
igoihman: done

mhenriks:
Hi @mmazur, great suggestion!  Off topic for this PR, but here are my thoughts:

Currently, the uploadproxy tls key+cert is stored in a secret called  `cdi-upload-proxy-server-key` in the `cdi` namespace.  The config controller could watch that resource and add the cert to the CDIConfig object.  

However, I think there has to be some work for tls hostname validation to work correctly end to end.  Here's the situation:

The upload controller creates a cert for the upload proxy when it starts up for the first time.  The cert has hostnames that work fine within the cluster like `cdi-uplodproxy`, `cdi-uploadproxy.cdi`, etc.  But this may not work for a client coming from outside the cluster connecting to `cdi-uplodproxy.mycluster.com` or whatever via Ingress/Routes.  So how we we deal with this?

The easiest option (for us) is to have a cluster administrator create a key/cert for the cdi-uploadproxy to use and assign it to the `cdi-upload-proxy-server-key` secret.  I am going to create documentation on how to do this.  But it obviously sucks for the administrator.

The upload controller could watch CDIConfig and regenerate uploadproxy certs and restart the uploadproxy whenever the hostname changes.  Major downside here is the restart of uploadproxy could disrupt clients.

This is just off the top of my head.  Maybe other things to consider.

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,376 @@
...
+			if url != "" {
+				break
+			}
+		}
+	}
+	//if user overrode the upload proxy url, ignore all routes and ingresses
+	if config.Spec.UploadProxyURLOverride != nil && config.Spec.UploadProxyURLOverride != config.Status.UploadProxyURL {
+		url = *config.Spec.UploadProxyURLOverride
+	}
+	err = updateCDIConfig(c.cdiClientSet, config, url)
```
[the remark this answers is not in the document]
igoihman: done

pkg/controller/config-controller.go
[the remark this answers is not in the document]
igoihman: I used `List` since `Get` from empty/all namespaces doesn't work.

manifests/example/example-cdiconfig.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: CDIConfig
+metadata:
+  name: config
+spec:
+  uploadProxyURLOverride: "www.example.com"
```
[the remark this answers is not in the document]
igoihman: done

igoihman:
@mhenriks @awels I updated the PR. Reviews are welcome!

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,396 @@
...
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	extensioninformers "k8s.io/client-go/informers/extensions/v1beta1"
+	"k8s.io/client-go/kubernetes"
+	extensionlisters "k8s.io/client-go/listers/extensions/v1beta1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+	cdiclientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
+	clientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
```
mhenriks: "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned" imported twice.

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,396 @@
...
+		}
+		for _, ing := range ingresses {
+			url = getURLFromIngress(ing, c.uploadProxyServiceName)
+			if url != "" {
+				break
+			}
+		}
+	}
+	//if user overrode the upload proxy url, ignore all routes and ingresses
+	if config.Spec.UploadProxyURLOverride != nil && config.Spec.UploadProxyURLOverride != config.Status.UploadProxyURL {
```
mhenriks: Not sure I understand this part of `if` condition `config.Spec.UploadProxyURLOverride != config.Status.UploadProxyURL`

Is it your intention to compare to pointer values?

Also, wouldn't you always want to assign `url` to `*config.Spec.UploadProxyURLOverride` if it is set?

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,396 @@
...
+		url = *config.Spec.UploadProxyURLOverride
+	}
+	newConfig := config.DeepCopy()
+	// mutate newConfig
+	newConfig.TypeMeta.Kind = "CDIConfig"
+	newConfig.TypeMeta.APIVersion = "cdi.kubevirt.io/v1alpha1"
```
mhenriks: Kind and APIVersion should be set during deep copy.  Assuming the source has the properties set.  And if the source doesn't have the properties set, I'm not sure the destination does either.

pkg/controller/config-controller.go
mhenriks: Curious if you ran the test I described earlier?  Because you essentially have duplicate logic here and in `syncHandler`.  If the logic cannot live in a single place, I think the two pieces of code should share a function.

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,396 @@
...
+		return err
+	}
+	for _, route := range routes {
+		url = getURLFromRoute(route, c.uploadProxyServiceName)
+		if url != "" {
+			break
+		}
+	}
+	if url == "" {
+		ingresses, err := c.client.ExtensionsV1beta1().Ingresses(metav1.NamespaceAll).List(metav1.ListOptions{
```
mhenriks: Can't use the ingress lister here like in `syncHandler`?

pkg/controller/config-controller.go
igoihman: I had an issue with empty fields when I called `Get`.  Now when I use `List` all these fields are set.

pkg/controller/config-controller.go
igoihman: correct, thanks!

pkg/controller/config-controller.go
igoihman: done

pkg/controller/config-controller.go
mhenriks: Can we get rid of one of these imports?

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,384 @@
...
+				if url != "" {
+					break
+				}
+			}
+		}
+	}
+
+	newConfig := config.DeepCopy()
+	// mutate newConfig
+	newConfig.Status.UploadProxyURL = &url
```
mhenriks: looks like it is possible to get here and have `url == ""`.  If so, `newConfig.Status.UploadProxyURL` should be `nil`.  Not the address of `url`.

Also, if the url did not change, should we still do an update?

pkg/controller/util.go
```diff
@@ -959,3 +1022,54 @@ func getCertConfigMap(client kubernetes.Interface, pvc *v1.PersistentVolumeClaim
...
+		}
+	}
+	return ""
+
+}
+
+func getURLFromRoute(route *routev1.Route, uploadProxyServiceName string) string {
+	if route.Spec.To.Name == uploadProxyServiceName {
+		if route.Spec.Host != "" {
+			return route.Spec.Host
```
mhenriks: I wonder about getting this value from the `spec` of the Route.  I would think that something from `status` would be more appropriate.  Like `status.ingress[].host`?

pkg/controller/config-controller.go
igoihman: done - I was sure I got rid of it

pkg/controller/util.go
igoihman: done

igoihman:
@mhenriks good insights. I did several changes according to your comments.

pkg/controller/config-controller.go
```diff
@@ -0,0 +1,352 @@
...
+			}
+			for _, ing := range ingresses {
+				url = getURLFromIngress(ing, c.uploadProxyServiceName)
+				if url != "" {
+					break
+				}
+			}
+		}
+	}
+	if config.Status.UploadProxyURL != nil && url == *config.Status.UploadProxyURL {
```
mhenriks: Is it possible for this `config.Status.UploadProxyURL = nil && url == ""`?  Possible but unlikely?

igoihman: It's possible, I've updated the condition.

PR #608, opened 2019-01-16
Import container disk image from registry

annastopel:

Fixes #

**Release note**:

```release-note
CDI import from registry supports the same container image format as  kubevirt containerdisks
```

### DIFF
--- a/doc/image-from-registry.md
+++ b/doc/image-from-registry.md
@@ -2,47 +2,50 @@
 The purpose of this document is to show how to create registry image containing a Virtual Machine image that can be imported into a PV.
 
 ## Prerequisites
-The VM disk image should be called 'disk.img' and placed in the root folder.
+Import from registry should be able to consume the same container images as [containerDisk](https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md).
+Thus the VM disk image file to be consumed must be located under /disk directory in the container image. The file can be in any of the supported formats : qcow2, raw, archived image file. There are no special naming constraints for the VM disk file.
 
-## Create an image with Buildah
+## Import VM disk image file from existing containerDisk images in kubevirt repository 
+For example vmidisks/fedora25:latest as described in [containerDisk](https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md)
 
+## Create a container image with Buildah
 Buildah is a tool that facilitates building Open Container Initiative (OCI) container images.
-
 More information is available here: [Buildah tutorial](https://github.com/containers/buildah/blob/master/docs/tutorials/02-registries-repositories.md).
 
-Create a new container image from scratch and keep its name in a variable:
+Create a new directory `/tmp/vmdisk` with the following Docker file and a vm image file (ex: `fedora28.qcow2`)
+Create a new container image with the following docker file 
 
 ```bash
-$ newcontainer=$(buildah from scratch)
+cat << END > Dockerfile
+FROM kubevirt/container-disk-v1alpha
+ADD fedora28.qcow2 /disk
+END
 ```
-
-Add the VM disk to the container image and commit:
+Build and push image to a registry. 
+Note: In development environment you can push to 
+1. A cluster local `cdi-docker-registry-host` which hosts docker registry and is accessible within the cluster via `cdi-docker-registry-host.cdi`. The registry is initialized from `cluster-sync` flow and is used for functional tests purposes. 
+2. Globally accessible registry that is used for image caching and is accessible via `registry:5000` host name
 
 ```bash
-$ buildah copy $newcontainer ./disk.img /
-buildah commit $newcontainer my-vm-img
-```
-
-To push an image from your local Buildah container storage, check the image name, then push it using the buildah push command. Remember to identify both the local image name and a new name that includes the location (<registry>:5000, in this case):
-```
-buildah push --tls-verify=false my-vm-img:latest <registry>:5000/my-vm-img:latest
-```
-
-## Create an image with Docker
-
-Create a Dockerfile with the following content:
+$buildah bud -t vmidisk/fedora28:latest /tmp/vmdisk
+$buildah push --tls-verify=false vmidisk/fedora28:latest docker://cdi-docker-registry-host.cdi/fedora28:latest
 
 ```
-FROM scratch
-ADD disk.img /
+## Create a container image with Docker
+
+Create a Dockerfile with the following content in a new directory /tmp/vmdisk. Add an image file to the same directory (for example fedora28.qcow2)
+
+```
+FROM kubevirt/container-disk-v1alpha
+ADD fedora28.qcow2 /disk
 ```
 
 Build, tag and push the image:
 
 ```bash
-$ docker build . -t my-vm-img
-$ docker tag my-vm-img <registry>:5000/my-vm-img
-$ docker push <registry>:5000/my-vm-img
+$docker build -t vmdisks/fedora28:latest /tmp/vmdisk
+$docker push vmdisks/fedora28:latest
+
 ```
 
 # Import the registry image into a PVC
@@ -52,7 +55,7 @@ Use the following annotations in the PVC yaml:
 ...
 annotations:
     cdi.kubevirt.io/storage.import.source: "registry"
-    cdi.kubevirt.io/storage.import.endpoint: "docker://<registry>:5000/my-vm-img"
+    cdi.kubevirt.io/storage.import.endpoint: "docker://docker.io/kubevirt/cirros-registry-disk-demo"
 ...
 ```
 
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -18,6 +18,7 @@ package importer
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"context"
@@ -61,16 +62,24 @@ var _ DataStreamInterface = &DataStream{}
 
 var qemuOperations = image.NewQEMUOperations()
 
+//DataContext - allows cleanup of temporary data created by specific import method
+type DataContext interface {
+	Cleanup() error
+}
+
 // DataStream implements the ReadCloser interface
 type DataStream struct {
 	*DataStreamOptions
-	url     *url.URL
-	Readers []reader
-	buf     []byte // holds file headers
-	qemu    bool
-	Size    int64
-	ctx     context.Context
-	cancel  context.CancelFunc
+	url         *url.URL
+	Readers     []reader
+	buf         []byte // holds file headers
+	qemu        bool
+	archived    bool
+	tmpDataPath string
+	Size        int64
+	ctx         context.Context
+	cancel      context.CancelFunc
+	dataCtxt    DataContext
 }
 
 type reader struct {
@@ -82,6 +91,8 @@ type reader struct {
 type DataStreamOptions struct {
 	// Dest is the destination path of the contents of the stream.
 	Dest string
+	// DataDir is the destination path where data is stored
+	DataDir string
 	// Endpoint is the endpoint to get the data from for various Sources.
 	Endpoint string
 	// AccessKey is the access key for the endpoint, can be blank. This needs to be a base64 encoded string.
@@ -121,6 +132,25 @@ var rdrTypM = map[string]int{
 	"stream": rdrStream,
 }
 
+//Data stream helper methods
+func (d *DataStream) isQemuStored() bool {
+	//if it is unarchived qcow file that is already stored
+	//then no copy to tmp is required
+	if d.qemu && d.isStored() && !d.archived {
+		return true
+	}
+	return false
+}
+
+func (d *DataStream) isMoveable() bool {
+	//basically raw file that is already stored
+	return d.isStored() && !d.qemu && !d.archived
+}
+
+func (d *DataStream) isStored() bool {
+	return len(d.tmpDataPath) > 0
+}
+
 // NewDataStream returns a DataStream object after validating the endpoint and constructing the reader/closer chain.
 // Note: the caller must close the `Readers` in reverse order. See Close().
 func NewDataStream(dso *DataStreamOptions) (*DataStream, error) {
@@ -130,6 +160,7 @@ func NewDataStream(dso *DataStreamOptions) (*DataStream, error) {
 func newDataStreamFromStream(stream io.ReadCloser) (*DataStream, error) {
 	return newDataStream(&DataStreamOptions{
 		common.ImporterWritePath,
+		"",
 		"stream://data",
 		"",
 		"",
@@ -146,7 +177,8 @@ func newDataStream(dso *DataStreamOptions, stream io.ReadCloser) (*DataStream, e
 	}
 	var ep *url.URL
 	var err error
-	if dso.Source == controller.SourceHTTP || dso.Source == controller.SourceS3 || dso.Source == controller.SourceGlance {
+	if dso.Source == controller.SourceHTTP || dso.Source == controller.SourceS3 ||
+		dso.Source == controller.SourceGlance || dso.Source == controller.SourceRegistry {
 		ep, err = ParseEndpoint(dso.Endpoint)
 		if err != nil {
 			return nil, errors.Wrapf(err, fmt.Sprintf("unable to parse endpoint %q", dso.Endpoint))
@@ -160,6 +192,10 @@ func newDataStream(dso *DataStreamOptions, stream io.ReadCloser) (*DataStream, e
 		buf:               make([]byte, image.MaxExpectedHdrSize),
 		ctx:               ctx,
 		cancel:            cancel,
+		tmpDataPath:       "",
+		qemu:              false,
+		archived:          false,
+		dataCtxt:          nil,
 	}
 
 	// establish readers for endpoint's formats and do initial calc of size of raw endpt
@@ -192,22 +228,33 @@ func (d *DataStream) Close() error {
 	if d.cancel != nil {
 		d.cancel()
 	}
+	if d.dataCtxt != nil {
+		dataCtxtErr := d.dataCtxt.Cleanup()
+		if dataCtxtErr != nil {
+			glog.Errorf(dataCtxtErr.Error(), "DataCtxt cleanup failed")
+		}
+	}
 	return err
 }
 
 // Based on the endpoint scheme, append the scheme-specific reader to the receiver's
 // reader stack.
-func (d *DataStream) dataStreamSelector() (err error) {
+func (d *DataStream) dataStreamSelector() error {
 	var r io.Reader
 	scheme := d.url.Scheme
+	var err error
 	switch scheme {
 	case "s3":
 		r, err = d.s3()
 	case "http", "https":
 		r, err = d.http()
+	case "docker", "oci":
+		r, err = d.registry()
 	default:
+		glog.V(1).Infoln("Error in dataStream Selector - invalid url scheme")
 		return errors.Errorf("invalid url scheme: %q", scheme)
 	}
+
 	if err == nil && r != nil {
 		d.appendReader(rdrTypM[scheme], r)
 	}
@@ -269,6 +316,128 @@ func (d *DataStream) http() (io.ReadCloser, error) {
 	return countingReader, nil
 }
 
+//registryData - temporary data used by import from registry flow
+//is deleted as part of dataStream.Close()
+type registryData struct {
+	dataDir string
+	file    *os.File
+}
+
+func (r registryData) Cleanup() error {
+	glog.V(1).Infof("registryData - deleting all the data")
+	if r.file != nil {
+		r.file.Close()
+	}
+	err := os.RemoveAll(r.dataDir)
+	if err != nil {
+		glog.Errorf(" Failed removing directory ", err.Error())
+	}
+	return nil
+}
+
+const (
+	//ContainerDiskImageDir - Expected disk image location in container image as described in
+	//https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md
+	ContainerDiskImageDir = "disk"
+	//TempContainerDiskDir - Temporary location for pulled containerImage
+	TempContainerDiskDir = "tmp"
+)
+
+//This import source downloads specified container image from registry location
+//Then it extracts the image to a temporary location and expects an image file to be located under /disk directory
+//If such exists it creates a Reader on it and returns it for further processing
+func (d *DataStream) registry() (io.ReadCloser, error) {
+
+	tmpData := registryData{"", nil}
+	tmpData.dataDir = filepath.Join(d.DataDir, TempContainerDiskDir)
+
+	imageDir := filepath.Join(tmpData.dataDir, ContainerDiskImageDir)
+
+	//1. create temporary directory if does not exist to which all the data will be extracted
+	if _, err := os.Stat(tmpData.dataDir); os.IsNotExist(err) {
+		err := os.Mkdir(tmpData.dataDir, os.ModeDir|os.ModePerm)
+		if err != nil {
+			glog.Errorf("Failed to create temporary directory")
+			return nil, errors.Wrapf(err, fmt.Sprintf("Failed to create tempdirectory %s", tmpData.dataDir))
+		}
+	}
+
+	//cleanup in case of failure
+	cleanup := true
+	defer func() {
+		if cleanup {
+			tmpData.Cleanup()
+		}
+	}()
+
+	//2. copy image from registry to the temporary location
+	glog.V(1).Infof("using skopeo to copy from registry")
+	err := image.CopyRegistryImage(d.Endpoint, tmpData.dataDir, ContainerDiskImageDir, d.AccessKey, d.SecKey)
+	if err != nil {
+		glog.Errorf("Failed to read data from registry")
+		return nil, errors.Wrapf(err, fmt.Sprintf("Failed ro read from registry"))
+	}
+
+	//3. Search for file in /disk directory - if not found - failure
+	imageFile, err := getImageFileName(imageDir)
+	if err != nil {
+		glog.Errorf("Error getting Image file from imageDirectory")
+		return nil, errors.Wrapf(err, fmt.Sprintf("Cannot locate image file"))
+	}
+
+	// 4. If found - Create a reader that will read this file and attach it to the dataStream
+	tmpData.file, err = os.Open(filepath.Join(imageDir, imageFile))
+	if err != nil {
+		glog.Errorf("Failed to open image file")
+		return nil, errors.Wrapf(err, fmt.Sprintf("Fail to create data stream from image file"))
+	}
+
+	//got this far - do not cleanup
+	cleanup = false
+
+	d.tmpDataPath = filepath.Join(imageDir, imageFile)
+	d.dataCtxt = tmpData
+	glog.V(3).Infof("Sucecssfully found file. VM disk image filename is %s", imageFile)
+
+	return ioutil.NopCloser(bufio.NewReader(tmpData.file)), nil
+}
+
+func getImageFileName(dir string) (string, error) {
+
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		glog.Errorf("image directory does not exist")
+		return "", errors.Errorf(fmt.Sprintf("image directory does not exist "))
+	}
+
+	entries, err := ioutil.ReadDir(dir)
+	if err != nil {
+		glog.Errorf("Error reading directory")
+		return "", errors.Wrapf(err, fmt.Sprintf("image file does not exist in image directory"))
+	}
+
+	if len(entries) == 0 {
+		glog.Errorf("image file does not exist in image directory - directory is empty ")
+		return "", errors.Errorf(fmt.Sprintf("image file does not exist in image directory - directory is empty"))
+	}
+
+	fileinfo := entries[len(entries)-1]
+	if fileinfo.IsDir() {
+		glog.Errorf("image file does not exist in image directory contains another directory ")
+		return "", errors.Errorf(fmt.Sprintf("image file does not exist in image directory"))
+	}
+
+	filename := fileinfo.Name()
+
+	if len(strings.TrimSpace(filename)) == 0 {
+		glog.Errorf("image file does not exist in image directory - file has no name ")
+		return "", errors.Errorf(fmt.Sprintf("image file does not exist in image directory"))
+	}
+
+	glog.V(1).Infof("VM disk image filename is %s", filename)
+
+	return filename, nil
+}
+
 func (d *DataStream) pollProgress(reader *util.CountingReader, idleTime, pollInterval time.Duration) {
 	count := reader.Current
 	lastUpdate := time.Now()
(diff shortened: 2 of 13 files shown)


doc/image-from-registry.md
```diff
@@ -2,47 +2,50 @@
 The purpose of this document is to show how to create registry image containing a Virtual Machine image that can be imported into a PV.
 
 ## Prerequisites
-The VM disk image should be called 'disk.img' and placed in the root folder.
+Import from registry should be able to consume the same container images as containerDisk (https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md).
+Thus the VM disk image file to be consumed must be located under /disk directory in the container image. The file can be in any of the supported formats : qcow2, raw, archived. There is no any special naming contraints for the VM disk file.
 
-## Create an image with Buildah
+## Import from existing containerDisk images from kubevirt repository
```
igoihman: Import existing containerDisk images from kubevirt repository

doc/image-from-registry.md
```diff
@@ -2,47 +2,50 @@
 The purpose of this document is to show how to create registry image containing a Virtual Machine image that can be imported into a PV.
 
 ## Prerequisites
-The VM disk image should be called 'disk.img' and placed in the root folder.
+Import from registry should be able to consume the same container images as containerDisk (https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md).
```
igoihman: should be in this format:
``` [containerDisk](https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md) ```

File: doc/image-from-registry.md [hunk 3]
```diff
@@ -2,47 +2,50 @@
...
+## Import from existing containerDisk images from kubevirt repository 
+For example vmidisks/fedora25:latest as described in https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md
 
+## Create a container image with Buildah
 Buildah is a tool that facilitates building Open Container Initiative (OCI) container images.
-
 More information is available here: [Buildah tutorial](https://github.com/containers/buildah/blob/master/docs/tutorials/02-registries-repositories.md).
 
-Create a new container image from scratch and keep its name in a variable:
+Create a new directory /tmp/vmdisk with the following Docker file and a vm image file (ex: fedora28.qcow2)
```
igoihman: style: put `/tmp/vmdisk ` inside backticks.

doc/image-from-registry.md
```diff
@@ -2,47 +2,50 @@
...
+cat << END > Dockerfile
+FROM kubevirt/container-disk-v1alpha
+ADD fedora28.qcow2 /disk
+END
 ```
-
-Add the VM disk to the container image and commit:
+Build and push image to a registry. 
+Note: In development environment you can push to 
+1. A cluster local cdi-docker-registry-host which hosts docker registry and is accessible within the cluster via cdi-docker-registry-host.kube-system. The registry is initialized from cluster-sync flow and is used for functional tests purposes.
```
igoihman: `cdi-docker-registry-host`, `cdi-docker-registry-host.kube-system` and `cluster-sync` inside backticks.

doc/image-from-registry.md
```diff
@@ -2,47 +2,50 @@
...
+FROM kubevirt/container-disk-v1alpha
+ADD fedora28.qcow2 /disk
+END
 ```
-
-Add the VM disk to the container image and commit:
+Build and push image to a registry. 
+Note: In development environment you can push to 
+1. A cluster local cdi-docker-registry-host which hosts docker registry and is accessible within the cluster via cdi-docker-registry-host.kube-system. The registry is initialized from cluster-sync flow and is used for functional tests purposes. 
+2. Globally accessible registry that is used for image caching and is accessible via registry:5000 host name
```
igoihman: same for `registry:5000`

doc/image-from-registry.md
```diff
@@ -2,47 +2,50 @@
...
-To push an image from your local Buildah container storage, check the image name, then push it using the buildah push command. Remember to identify both the local image name and a new name that includes the location (<registry>:5000, in this case):
 ```
-buildah push --tls-verify=false my-vm-img:latest <registry>:5000/my-vm-img:latest
-```
-
-## Create an image with Docker
+## Create a container image with Docker
 
-Create a Dockerfile with the following content:
+Create a Dockerfile with the following content in a new directory /tmp/vmdisk. Add an image file to the same directory (for example fedora28.qcow2)
```
igoihman: ` /tmp/vmdisk`

pkg/importer/dataStream.go
```diff
@@ -266,6 +270,83 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+//If such exists it creates  e Reader on it and returns it for further processing
+func (d *DataStream) registry() (io.ReadCloser, error) {
+	//generate random name for temporary directory
+	tmpDest := randTmpName(TempContainerDiskDir)
+
+	defer func(f string) {
+		os.Remove(f)
+	}(tmpDest)
+
+	imageDir := tmpDest + ContainerDiskImageDir
```
igoihman: better use `filepath.join`

pkg/importer/dataStream.go
```diff
@@ -266,6 +270,83 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	}
+
+	//Search for file in /disk directory - if not found - failure
+	imageFile, err := getImageFileName(imageDir)
+	if err != nil {
+		return nil, errors.Wrapf(err, fmt.Sprintf("Cannot locate image file"))
+	}
+
+	// 3. If found - Create a reader that will read this file and attach it to the dataStream
+	file, err := os.Open(imageDir + "/" + imageFile)
```
igoihman: `filepath.join`

pkg/importer/dataStream.go
```diff
@@ -266,6 +270,83 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+
+	imageDir, err := os.Open(dir)
+
+	if err != nil {
+		return "", errors.Wrapf(err, fmt.Sprintf("Cannot read from image directory"))
+	}
+
+	defer imageDir.Close()
+
+	list, err := imageDir.Readdirnames(0) // 0 to read all files and folders
```
igoihman: better use `fileList` - `list` is confusing.

doc/image-from-registry.md
```diff
@@ -2,47 +2,50 @@
 The purpose of this document is to show how to create registry image containing a Virtual Machine image that can be imported into a PV.
 
 ## Prerequisites
-The VM disk image should be called 'disk.img' and placed in the root folder.
+Import from registry should be able to consume the same container images as [containerDisk](https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md).
+Thus the VM disk image file to be consumed must be located under /disk directory in the container image. The file can be in any of the supported formats : qcow2, raw, archived image file. There are no any special naming contraints for the VM disk file.
```
awels: There are no special naming constraints for the VM disk.

Extra 'any' and constraints is misspelled.

aglitke:
Looks good to me.  I think this will be ready once you address @awels comments and CI passes.

tests/transport_test.go
```diff
@@ -123,14 +119,14 @@ var _ = Describe("Transport Tests", func() {
...
 		Entry("should not connect to http endpoint with invalid credentials", httpAuthEp, targetFile, "gopats", "bradyisthegoat", controller.SourceHTTP, false),
 		Entry("should connect to QCOW http endpoint without credentials", httpNoAuthEp, targetQCOWFile, "", "", controller.SourceHTTP, true),
 		Entry("should connect to QCOW http endpoint with credentials", httpAuthEp, targetQCOWFile, utils.AccessKeyValue, utils.SecretKeyValue, controller.SourceHTTP, true),
-		Entry("should connect to registry endpoint without credentials", registryNoAuthEp, "registry", "", "", controller.SourceRegistry, true),
-		//		Entry("should not connect to registry endpoint with invalid credentials", registryNoAuthEp, "registry", "gopats", "bradyisthegoat", controller.SourceRegistry, false),
+		Entry("should connect to registry endpoint without credentials", registryNoAuthEp, targetQCOWImage, "", "", controller.SourceRegistry, true),
+		Entry("should fail to import from registry endpoint when container image is invalid", registryNoAuthEp, targetRegistryInvalidImage, "", "", controller.SourceRegistry, false),
```
shiywang: I think the negtive case we want to conver is here: https://github.com/kubevirt/containerized-data-importer/pull/608/files#diff-9cbc435c7c64940bbba1090c17662234R309

should add a test about:
`should fail to import from registry endpoint when container image is not putting in  /disk directory` sth like this.

aglitke:
@annastopel Have you figured out why this is not passing CI on openshift?

annastopel:
@aglitke Import fails with os since cri-o containers have limited privileges. Specifically, tar fails to execute mknod and thus the entire sequence fails. I am working on a solution

annastopel:
ci please

annastopel:
ci test please

annastopel:
ci test please

pkg/importer/dataStream.go
```diff
@@ -266,6 +270,86 @@ func (d *DataStream) http() (io.ReadCloser, error) {
 	return countingReader, nil
 }
 
+const (
+	//ContainerDiskImageDir - Expected disk image location in contianer image as described in
+	//https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md
+	ContainerDiskImageDir = "disk"
+	//TempContainerDiskDir - Temporary location for pulled containerImage
+	TempContainerDiskDir = "/tmp"
```
mhenriks: So, there is no way to use the PVC for temp space?  fakeroot did not work?

annastopel:
I changed the implementation after our talk - now it writes to PVC.
Hopefully will push it today

pkg/importer/dataStream.go
annastopel: Now storing on PVC in /data/tmp directory which is deleted at the end of the flow

annastopel:
ci test please

awels:
travis failed here: https://travis-ci.org/kubevirt/containerized-data-importer/builds/486877012#L4886

annastopel:
> travis failed here: https://travis-ci.org/kubevirt/containerized-data-importer/builds/486877012#L4886

I am not sure this is the reason, since for https://github.com/kubevirt/containerized-data-importer/pull/628 the same error had occured and travis called it passed

awels:
Good point, I didn't see anything remotely looking strange, so I figured that was the problem.

annastopel:
ci test please

pkg/importer/dataStream.go
```diff
@@ -292,9 +403,6 @@ func (d *DataStream) pollProgress(reader *util.CountingReader, idleTime, pollInt
 func CopyData(dso *DataStreamOptions) error {
 	glog.V(1).Infof("copying %q to %q...\n", dso.Endpoint, dso.Dest)
 	switch dso.Source {
-	case controller.SourceRegistry:
-		glog.V(1).Infof("using skopeo to copy from registry")
-		return image.CopyRegistryImage(dso.Endpoint, dso.Dest, dso.AccessKey, dso.SecKey)
 	default:
```
mhenriks: Switch statement with only default case

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
 }
 
+type registryData struct {
+	dataDir string
+	file    *os.File
+}
+
+//_regData - global variable that stores temporary data of registy flow
+// to be deleted when teh flow ends
+var _regData = registryData{"", nil}
```
mhenriks: Why can't this data belong to the datastream struct?

pkg/image/skopeo.go
```diff
@@ -95,6 +95,61 @@ func CopyRegistryImage(url, dest, accessKey, secKey string) error {
...
+func CopyDirFromRegistryImage(url, dest, VMimagePath, accessKey, secKey string) error {
+	skopeoDest := "dir:" + dest + dataTmpDir
+	err := SkopeoInterface.CopyImage(url, skopeoDest, accessKey, secKey)
+	if err != nil {
+		os.RemoveAll(dest + dataTmpDir)
+		return errors.Wrap(err, "Failed to download from registry")
+	}
+	err = extractVMDiskImageDir(dest, VMimagePath)
+	if err != nil {
+		return errors.Wrap(err, "Failed to extract image layers")
```
mhenriks: no cleaning up if we error here.

pkg/image/skopeo.go
```diff
@@ -95,6 +95,61 @@ func CopyRegistryImage(url, dest, accessKey, secKey string) error {
...
+	}
+
+	// Clean temp folder
+	os.RemoveAll(dest + dataTmpDir)
```
mhenriks: should probably just make this deferred.

pkg/importer/dataStream.go
```diff
@@ -202,6 +207,8 @@ func (d *DataStream) dataStreamSelector() (err error) {
 		r, err = d.s3()
 	case "http", "https":
 		r, err = d.http()
+	case "docker", "oci":
+		r, err = d.registry()
```
mhenriks: unlike the other functions here like `http()`, `registry()` does a lot of stuff.  It downloads the container and extracts everything.  The bulk of the work.  Doesn't do much in `copy()` which is I believe is supposed to be the workhorse.

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	// 3. If found - Create a reader that will read this file and attach it to the dataStream
+	_regData.file, err = os.Open(filepath.Join(imageDir, imageFile))
+	if err != nil {
+		os.RemoveAll(_regData.dataDir)
+		return nil, errors.Wrapf(err, fmt.Sprintf("Fail to create data stream from image file"))
+	}
+
+	glog.V(1).Infof("VM disk image filename is %s", imageFile)
+
+	return RegistryDataCloser(bufio.NewReader(_regData.file)), nil
```
mhenriks: this seems really inefficient and could cause issues with PVC capacity.  Basically, the PVC has to be at least twice the size of the image because you are doing a copy.

pkg/importer/dataStream.go
annastopel: This data is relevant only to import from registry flow

pkg/image/skopeo.go
annastopel: Agreed

pkg/importer/dataStream.go
annastopel: First of all, I think we should discuss how is it better to solve temporary space issue required by all import flows.

Regarding the specific flow:
The VM disk image image we import can be in several formats:
- raw - (no redundancy see below)
- qcow2 - agree 1 redundant copy  
- qcow archived - (no redundancy see below)

- In case of raw - there is no redundant copy  - we extract the disk image file from container image and then copy it to /disk/disk.img - 
**tmpPVC** space = sizeof(containerImage)+sizeof(/disk directory in the image ~  sizeof(VM disk image file) in typical registry Image) 
**PVC space** = sizeof (raw VM disk image)

- In case of qcow2 archived - there is no redundant copy - we extract archived qcow2 file from container image and then in StreamDataToFile we unarchive it and then convert from qcow2 to raw.
**tmpPVC space** = sizeof(containerImage) + sizeof(~ archived VM disk image file)  + sizeof (unarchived VM disk image file)
**PVC space** = sizeof (raw VM disk image file)

In case of unarchive qcow2 - I agree that one copy is redundant (I am handling it) - currently 
**tmpPVC space** = sizeof(containerImage) + 2*sizeof(~  VM disk image file in qcow2 format)
**PVC space** = sizeof (raw VM disk image file)

After the fix space consumption is:
**tmpPVC space** = sizeof(containerImage) + sizeof(~  VM disk image file in qcow2 format)
**PVC space** = sizeof (raw VM disk image file)

pkg/image/skopeo.go
annastopel: Good catch!

pkg/importer/dataStream.go
annastopel: Import form registry is different from other imports, so the handling is different. But as the rest it obtains a reader on VM image file for further processing such as (inspecting image file, its size validations and extracting it if it is archived ) Do you see a problem with such implementation?

pkg/importer/dataStream.go
mhenriks: I still think it would be preferable to have members (even if they are unused) in the datastream.  By using global variables, we are precluding ourselves from ever doing simultaneous registry imports in the same process.

pkg/importer/dataStream.go
mhenriks: Yeah, I think it should do as little as possible.  I don't think it has to go to the network at all, actually.  The other methods only read as little of the file as necessary to construct the appropriate readers.  The whole file is not downloaded until copy().  The registry implementation should be similar for consistency sake but also, I could imagine a case where we would want to initialize a datastream but not actually move anything until later.

mhenriks: Okay, I see, you have to download everything in `registry()` in order to see if the file is qcow2 so that it gets converted like an http download.  I understand now, but I'm still not sure we want to do this much work upfront

pkg/importer/dataStream.go
mhenriks: I see redundancy in all cases.  Here's how I see it, ignoring downloading of the container layers which obliviously adds its own overhead.

Raw:

0.  Image extracted from image to `/data/disk/myimage.raw`
1.  Image COPIED from `/data/disk/myimage.raw` to `/data/disk.img`
3.  `/data/disk` directory deleted.

At step 1, there are two copies of the image on the PVC.

qcow2:

0.  Image extracted from image to `/data/disk/myimage.qcow2`
1.  Image copied to `/data/<temp dir>/myimage.qcow2`
2.  Conversion to raw creates `/data/disk.img`
3.  `/data/<temp dir>` deleted
4.  `/data/disk` deleted

At step 2, there are two copies of the qcow2 image on the pvc

qcow2 archived is similar to above except replace step 0 with:

0.  Image extracted from image to `/data/disk/myimage.qcow2.tar.gz`

pkg/image/skopeo.go
```diff
@@ -95,6 +95,61 @@ func CopyRegistryImage(url, dest, accessKey, secKey string) error {
...
+
+	return err
+}
+
+//extractVMDiskImageDir - from each layer extracts only disk/ directory
+//Since this is the oly one that is in intereset right now
```
danielerez: * typos: 'oly' -> 'only', 'intereset' - 'interest'
* Maybe worth keeping the signature more generalised. I.e. 'extractImageDir = func(dest, imagePath string)'.

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
 	return countingReader, nil
 }
 
+type registryData struct {
+	dataDir string
+	file    *os.File
+}
+
+//_regData - global variable that stores temporary data of registy flow
+// to be deleted when teh flow ends
```
danielerez: s/teh/the

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+//This import source downloads specified container image from registry location
+//Then it extracts the image to a temporary location and expects an image file to be located under /disk directory
+//If such exists it creates a Reader on it and returns it for further processing
+func (d *DataStream) registry() (io.ReadCloser, error) {
+	//generate random name for temporary directory
+	_regData.dataDir = filepath.Join(d.DataDir, TempContainerDiskDir)
+
+	imageDir := filepath.Join(_regData.dataDir, ContainerDiskImageDir)
+
+	//create temporary directory if note exists to which all the data will be extracted
```
danielerez: note exists -> doesn't exist

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	}
+
+	//Search for file in /disk directory - if not found - failure
+	imageFile, err := getImageFileName(imageDir)
+	if err != nil {
+		os.RemoveAll(_regData.dataDir)
+		return nil, errors.Wrapf(err, fmt.Sprintf("Cannot locate image file"))
+	}
+
+	// 3. If found - Create a reader that will read this file and attach it to the dataStream
```
danielerez: add '1.' in line 326, and '2.' in line 334 :)

pkg/util/util.go
```diff
@@ -113,6 +113,38 @@ func UnArchiveTar(reader io.Reader, destDir string, arg ...string) error {
 	return nil
 }
 
+// ExrtactSingleDirTar unarchives a tar file and streams its files
+// using the specified io.Reader to the specified destination.
+func ExrtactSingleDirTar(reader io.Reader, destDir string, extractDir string) error {
```
danielerez: maybe we could modify 'UnArchiveTar' function and reuse it here, instead of copy the logic. could be done later on..

pkg/util/util.go
```diff
@@ -113,6 +113,38 @@ func UnArchiveTar(reader io.Reader, destDir string, arg ...string) error {
...
+	err = untar.Wait()
+	if err != nil {
+		glog.V(3).Infof("%s\n", string(errBuf.Bytes()))
+		return err
+	}
+
+	return nil
+}
+
+// UnArchiveSingleDirTar extracts a  extractDir from local tar file to the specified destination.
```
danielerez: maybe rephrase: 'extracts a specified extractDir..'

pkg/util/util.go
```diff
@@ -113,6 +113,38 @@ func UnArchiveTar(reader io.Reader, destDir string, arg ...string) error {
...
+func UnArchiveSingleDirTar(filePath, destDir string, extractDir string) error {
```
danielerez: maybe call it 'UnArchiveSingleDirFromLocalTar' to make it more explicit and correlated with UnArchiveLocalTar

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	//copy image from registry to the temporary location
+	glog.V(1).Infof("using skopeo to copy from registry")
+	err := image.CopyDirFromRegistryImage(d.Endpoint, _regData.dataDir, ContainerDiskImageDir, d.AccessKey, d.SecKey)
+	if err != nil {
+		os.RemoveAll(_regData.dataDir)
+		return nil, errors.Wrapf(err, fmt.Sprintf("Failed ro read from registry"))
+	}
+
+	//Search for file in /disk directory - if not found - failure
+	imageFile, err := getImageFileName(imageDir)
```
mhenriks: Is it our intention that import from registry only supports the `kubevirt` content type?

annastopel: yes

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return "", errors.Wrapf(err, fmt.Sprintf("image directory does not exist "))
+	}
+
+	entries, err := ioutil.ReadDir(dir)
+	if len(entries) == 0 || err != nil {
+		return "", errors.Wrapf(err, fmt.Sprintf("image file does not exist in image directory"))
+	}
+
+	fileinfo := entries[len(entries)-1]
```
mhenriks: Should we maybe add a little more intelligence here if there are multiple files?  Like, if there happens to be multiple files, but one is named `disk.img` use that?  Or maybe inspect content types?  Or maybe it's an error?

annastopel: this logic is the same as in containerDisk kubevirt/cmd/container-disk-v1alpha/entry-point.sh. Our intention is to have the same logic. we shall not assume that the disk image file has specific name or format

pkg/image/skopeo.go
```diff
@@ -95,6 +95,61 @@ func CopyRegistryImage(url, dest, accessKey, secKey string) error {
 	return err
 }
 
+// CopyDirFromRegistryImage download image from registry with skopeo and extracts a specific directory from it
+func CopyDirFromRegistryImage(url, dest, VMimagePath, accessKey, secKey string) error {
```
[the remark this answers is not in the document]
annastopel: agreed

pkg/image/skopeo.go
annastopel: agreed

pkg/importer/dataStream.go
annastopel: removed

pkg/image/skopeo.go
awels: if you defer this above the error checks, then you don't need to worry about it if any of the errors happen since it will get cleaned up anyway.

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	io.Reader
+}
+
+func (registryDataCloser) Close() error {
+	_regData.file.Close()
+	os.RemoveAll(_regData.dataDir)
+	return nil
+}
+
+// RegistryDataCloser returns a ReadCloser with a Close method wrapping delettion
```
awels: // RegistryDataCloser returns a ReadCloser with a Close method wrapping deletion

pkg/importer/dataStream.go
```diff
@@ -266,6 +273,110 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+// of temporary data of registry flow
+func RegistryDataCloser(r io.Reader) io.ReadCloser {
+	return registryDataCloser{r}
+}
+
+const (
+	//ContainerDiskImageDir - Expected disk image location in contianer image as described in
```
awels: contianer -> container

pkg/importer/dataStream.go
awels: Michael has a good point, at this point there no reason to have switch statement if we just have default and nothing else.

pkg/util/util.go
awels: There is also a typo in the name it should be ExtractSingleDirTar

pkg/importer/dataStream.go
annastopel: Currently leave it as is

pkg/importer/dataStream.go
annastopel: optimized in case of qcow2 file - now it is not copied to additional tmp location, but converted to raw from the location registry flow extracted it to

annastopel:
ci test please

pkg/importer/dataStream.go
```diff
@@ -119,6 +130,32 @@ var rdrTypM = map[string]int{
...
+func (d *DataStream) isCopyToTmpRequired() bool {
+	//if it is unarchived qcow file that is already stored
+	//then no copy to tmp is required
+	if d.isQemu() && d.isStored() && !d.isArchived() {
+		return false
+	}
+	return true
+}
+
+func (d *DataStream) getTmpDataPath() string {
```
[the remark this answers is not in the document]
mhenriks: I smell a java programmer!

pkg/importer/dataStream.go
```diff
@@ -732,21 +910,26 @@ func (d *DataStream) copy(dest string) error {
...
 		glog.V(2).Infof("copying image file to %q", dest)
 		tmpDest := dest
-		if d.qemu {
-			// copy to tmp; qemu conversion will write to passed-in destination
-			tmpDest = randTmpName(dest)
-			glog.V(3).Infof("Copy: temp file for qcow2 conversion: %q", tmpDest)
-			defer func(f string) {
-				os.Remove(f)
-			}(tmpDest)
+		if d.isCopyToTmpRequired() {
```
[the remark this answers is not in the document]
mhenriks: Please, please, please rename this function

pkg/importer/dataStream.go
```diff
@@ -734,31 +920,51 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
+				// copy to tmp; qemu conversion will write to passed-in destination
+				tmpDest = randTmpName(dest)
+				glog.V(3).Infof("Copy: temp file for qcow2 conversion: %q", tmpDest)
+				defer func(f string) {
+					os.Remove(f)
+				}(tmpDest)
+			}
+			if d.isMovble() {
+				//before moving has to close all readers
+				closeReaders(d.Readers)
```
mhenriks: Shouldn't be necessary

pkg/importer/dataStream.go
```diff
@@ -734,31 +920,51 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
+				err = MoveFile(d.getTmpDataPath(), tmpDest)
+				if err != nil {
+					//Failback to copy since move failed
```
mhenriks: I what situations would this code succeed if a move failed?

pkg/importer/dataStream.go
```diff
@@ -734,31 +920,51 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
+					file, fileerr := os.Open(d.getTmpDataPath())
```
mhenriks: You wouldn't have to bother reopening this file if you didn't close readers above

pkg/importer/dataStream.go
annastopel: At this stage there are Readers on the file,either Rename will fail or the file handle sill become stale and Close in dataStream.Close() will throw an error

pkg/importer/dataStream.go
annastopel: If source and dest belong to different devices/partitions (tmpPVC vs PVC) Rename will fail with the following error "invalid cross-device link". Copy however, does succeed, since it will actually create a new file on the destination file system

pkg/importer/dataStream.go
annastopel: Yes, but then again - Rename will fail due to opened file handles

pkg/importer/dataStream.go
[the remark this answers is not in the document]
annastopel: Done

pkg/importer/dataStream.go
mhenriks: That is very surprising to me.  Renaming an open file is totally fine at the POSIX level so I wonder where/how this error is triggered.

pkg/importer/dataStream.go
mhenriks: oh, got it

pkg/importer/dataStream.go
```diff
@@ -734,31 +920,51 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
+					if fileerr != nil {
+						glog.Errorf(fileerr.Error(), "Failed to open image file ", d.getTmpDataPath)
+						return errors.WithMessage(fileerr, fmt.Sprintf("failed to open image file when failing back to copy"))
+					}
+					d.appendReader(rdrFile, bufio.NewReader(file))
+					err = StreamDataToFile(d.topReader(), tmpDest)
```
mhenriks: I think it would be cleaner to just do `io.Copy` or `ioutil.WriteFile` here.

pkg/importer/dataStream.go
annastopel: You are absolutely right!
 i got confused with src opened file vs dest opened file. Pushing a fix for that

pkg/importer/dataStream.go
annastopel: You are right! pushing the fix

annastopel:
ci test please

annastopel:
ci test please

pkg/importer/dataStream.go
```diff
@@ -121,6 +132,37 @@ var rdrTypM = map[string]int{
...
+func (d *DataStream) isQemuStored() bool {
+	//if it is unarchived qcow file that is already stored
+	//then no copy to tmp is required
+	if d.isQemu() && d.isStored() && !d.isArchived() {
+		return true
+	}
+	return false
+}
+
+func (d *DataStream) isMovble() bool {
```
awels: isMoveable()

pkg/importer/dataStream.go
```diff
@@ -269,6 +328,128 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
+	}
+
+	filename := fileinfo.Name()
+
+	if len(strings.TrimSpace(filename)) == 0 {
+		glog.Errorf("image file does not exist in image directory - file has no name ")
+		return "", errors.Errorf(fmt.Sprintf("image file does not exist in image directory"))
+	}
+
+	glog.V(1).Infof("VM disk image filename is %s", filename)
```
awels: I don't think this is a verbosity 1 level message, its more debugging, so a 3 is more appropriate IMO.

pkg/importer/dataStream.go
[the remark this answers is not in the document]
awels: I am going to have to agree with Michael here, the extra functions that simply return a field value, have little value. isStored actually does a calculation so I have no issues with that one, but isQemu, and isArchived don't give you any value. d.qemu vs d.isQemu() and d.archived vs d.isArchived() aren't that helpful.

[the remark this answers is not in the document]
annastopel: done

pkg/importer/dataStream.go
annastopel: done

pkg/importer/dataStream.go
annastopel: done

annastopel:
ci test please

annastopel:
ci test please

PR #618, opened 2019-01-18
fix apiserver unittests

mhenriks:

adding the owner reference to the api service broke unit tests.  this should fix.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/api.go
+++ b/pkg/operator/api.go
@@ -19,6 +19,9 @@ package operator
 import (
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/api.go
```diff
@@ -19,6 +19,9 @@ package operator
 import (
 	"fmt"
 
+	// use glog her because all callers currently using that
```
awels: // use glog here, because

PR #622, opened 2019-01-22
Updated the released files to include operator and current binaries.

awels:

Update the files that are on the release page to include latest binaries and examples.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,11 +68,15 @@ deploy:
   skip_cleanup: true
   file_glob: true
   file:
-  - _out/cmd/cdi-controller/cdi-controller
-  - _out/cmd/cdi-importer/cdi-importer
-  - _out/cmd/cdi-cloner/cloner_startup.sh
-  - manifests/generated/cdi-controller.yaml
-  - manifests/generated/cdi-controller.yaml.j2
+  - _out/cmd/cdi-controller/*
+  - _out/cmd/cdi-importer/*
+  - _out/cmd/cdi-apiserver/*
+  - _out/cmd/cdi-cloner/*
+  - _out/cmd/cdi-operator/*
+  - _out/cmd/cdi-uploadproxy/*
+  - _out/cmd/cdi-uploadserver/*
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -67,11 +67,15 @@ deploy:
...
-  - manifests/generated/cdi-controller.yaml
-  - manifests/generated/cdi-controller.yaml.j2
+  - _out/cmd/cdi-controller/*
+  - _out/cmd/cdi-importer/*
+  - _out/cmd/cdi-apiserver/*
+  - _out/cmd/cdi-cloner/*
+  - _out/cmd/cdi-operator/*
+  - _out/cmd/cdi-uploadproxy/*
+  - _out/cmd/cdi-uploadserver/*
+  - manifests/generated/*.yaml
```
mhenriks: Not sure the functional testing manifests should be included (file-host.yaml and registry-host.yaml).  Also wonder if we should ship cdi-controller.yaml.  Maybe deprecate it now to be removed later?  If we do include it, we should probably  add creation of the 'cdi' namespace to the manifest.

awels: Well I would like to avoid having to manually add each new file we create to this list. Because someone will forget to add it at some point. When/if we remove cdi-controller.yaml, doing it like this will automatically remove it from the release as well.

awels:
ci test please

hack/build/build-manifests.sh
```diff
@@ -72,8 +81,16 @@ done
...
-) 1>"${MANIFEST_GENERATED_DIR}/cdi-controller.yaml"
+) 1>>"${MANIFEST_GENERATED_DIR}/cdi-controller.yaml"
 
+cat > "${MANIFEST_GENERATED_DIR}/cdi-controller.yaml.j2" << EOF
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    cdi.kubevirt.io: ""
+  name: cdi
```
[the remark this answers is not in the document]
awels: Oops.

PR #624, opened 2019-01-23
operator unit tests

mhenriks:

Gotta eat your veggies.  Also changed the "Ready" condition a bit.  Now checks that spec.numReplicas == status.numReplicas.  Also, rather than delete condition when not ready, set status to False.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
Looks like it dropped coverage though :(

mhenriks:
@awels something must be up with coveralls.  Before this, there were zero unit tests for pkg/operator/controller.  Does coveralls only include packages with tests?

awels:
Yeah it only includes the packages we have in the unit tests cmd/... pkg/... and now the operator stuff

awels:
@mhenriks if you look at the goverals site, you can see what parts of the code have low coverage. For instance operator/controller/util.go should be relatively easy to get 100% coverage (some missing labels/annotation cases).

awels:
I know, its a pain in the ass, but we committed to maintaining a certain level of coverage, from unit and functional tests. I think 70% should be a reasonable level. Unit testing boiler plate is not on the top of my list of fun things to do either.

PR #625, opened 2019-01-24
Add validation of content type and unit tests.

awels:

Adds validation of content type in data volumes.
Adds unit tests to verify the validation.
Fixes #

**Release note**:

```release-note
Added validation of content type in data volumes.
```

### DIFF
--- a/pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
+++ b/pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
@@ -124,6 +124,27 @@ func validateDataVolumeSpec(field *k8sfield.Path, spec *cdicorev1alpha1.DataVolu
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


awels:
ci test please

awels:
Well some of the point of this exercise is to actually apply manifests to a cluster and see what happens. Instead of programatically generating the manifest (which is not what a human would do). We have the unit tests that handle that aspect of the testing.

mhenriks:
@awels I was not suggesting that we generate manifests from code.  There are a ton of places in the functional tests where resources are created by client-go.  This is the only test that uses manifests and kubectl.  Was wondering why this test special.

awels:
Because for this particular test, we explicitly wanted manually created manifests that are called by kubectl. To test what actual humans would do instead of some go client. To catch things like the status field missing causing errors instead of what the client generated (it might have seen the required on that field, and generated an empty field instead of bombing like a manual manifest would have done)

PR #629, opened 2019-01-30
Adding support to clone disk image from one Raw Block PV to another 

zvikorn:

This is WIP pr for adding support to clone disk image from one Block PV to another.

**Release note**:

```release-note
Adding support to clone disk image from one raw block PV to another
```

### DIFF
--- a/cmd/cdi-cloner/cloner_startup.sh
+++ b/cmd/cdi-cloner/cloner_startup.sh
@@ -16,15 +16,22 @@
 
 set -euo pipefail
 
-if [ $# != 2 ]; then
-    echo "cloner: 2 args are supported: source|target and socket name"
+if [ $# != 3 ]; then
+    echo "cloner: 3 args are supported: source|target, socket name, block|fs pv"
     exit 1
 fi
+
 obj="$1"      # source|target
 rand_dir="$2" # part of socket path
+volumeMode="$3" # FS|Block
 
 pipe_dir="/tmp/clone/socket/$rand_dir/pipe"
 image_dir="/tmp/clone/image"
+if [ "$volumeMode" == "block" ]; then
+    image_dir="/dev/blockDevice"
+fi
+echo "image_dir is" $image_dir
+
 retries=0
 max_retries=20
 sleep_time=3
@@ -33,15 +40,27 @@ if [ "$obj" == "source" ]; then
     echo "cloner: Starting clone source"
     echo "cloner: creating fifo pipe"
     mkfifo $pipe_dir
-    echo "cloner: creating tarball of the image and redirecting it to $pipe_dir"
-    pushd $image_dir
-	#figure out the size of content in the directory
-    size=$(du -sb . | cut -f1)
-    echo $size
-    #Write the size to the pipe so the other end can read it.
-    echo "$size" >$pipe_dir
-    tar cv --sparse . >$pipe_dir
-    popd
+    
+    echo $volumeMode
+    if [ "$volumeMode" == "FS" ]; then
+        echo "cloner: creating tarball of the image and redirecting it to $pipe_dir"
+        pushd $image_dir
+        #figure out the size of content in the directory
+        size=$(du -sb . | cut -f1)
+        echo size is $size
+        #Write the size to the pipe so the other end can read it.
+        echo "$size" >$pipe_dir
+        tar cv --sparse . >$pipe_dir
+        popd
+    elif [ "$volumeMode" == "block" ]; then
+        #figure out the size of the block device
+        size=$(lsblk -n -b -o SIZE $image_dir) 
+        echo size is $size
+        #Write the size to the pipe so the other end can read it.
+        echo "$size" >$pipe_dir
+        echo "cloner: writing the image to $pipe_dir"
+        dd if=$image_dir bs=64K of=$pipe_dir 
+    fi
     echo "cloner: finished writing image to $pipe_dir"
     exit 0
 fi
@@ -51,10 +70,14 @@ if [ "$obj" == "target" ]; then
     while true; do
         echo "cloner: check if the fifo pipe was created by the cloning source pod"
         if [ -e "$pipe_dir" ]; then
-            pushd $image_dir
+            if [ "$volumeMode" != "block" ]; then
+                pushd $image_dir
+            fi
             echo "cloner: extract the image from $pipe_dir into $image_dir directory"
             /usr/bin/cdi-cloner -pipedir $pipe_dir -alsologtostderr -v=3
-            popd
+            if [ "$volumeMode" != "block" ]; then
+                popd
+            fi
         	if [ "$?" != "0" ]; then
         		echo "cloner: failed with exit code $?"
         		exit 1
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -80,10 +80,8 @@ const (
 
 	// UploadServerCDILabel is the label applied to upload server resources
 	UploadServerCDILabel = "cdi-upload-server"
-
 	// UploadServerPodname is name of the upload server pod container
 	UploadServerPodname = UploadServerCDILabel
-
 	// UploadServerDataDir is the destination directoryfor uploads
 	UploadServerDataDir = ImporterDataDir
 	// UploadServerServiceLabel is the label selector for upload server services
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -634,33 +634,43 @@ func MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName string, pvc *v1.Per
 			},
 		},
 		Spec: v1.PodSpec{
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
+			// The pod has to be privileged as it has to have access to the hostPath in the node.
+			// However, currently there is a bug that we cannot attach block device to the pod if the pod (container)
+			// is privileged:
+			//https://github.com/kubernetes/kubernetes/issues/58251
+			//https://github.com/kubernetes/kubernetes/issues/62560
+			// As a result of that, instead setting the SecurityContext field as Privileged, in the main container
+			// where we need to have access to  the hostPath, we set the SecurityContext field in the initContainer,
+			// and that will do it.
+			InitContainers: []v1.Container{
 				{
-					Name:            common.ClonerSourcePodName,
-					Image:           image,
-					ImagePullPolicy: v1.PullPolicy(pullPolicy),
-					SecurityContext: &v1.SecurityContext{
-						Privileged: &[]bool{true}[0],
-						RunAsUser:  &[]int64{0}[0],
-					},
-
+					Name:  "init",
+					Image: image,
 					VolumeMounts: []v1.VolumeMount{
-						{
-							Name:      ImagePathName,
-							MountPath: common.ClonerImagePath,
-						},
 						{
 							Name:      socketPathName,
 							MountPath: common.ClonerSocketPath + "/" + id,
 						},
 					},
-					Args: []string{"source", id},
+					SecurityContext: &v1.SecurityContext{
+						Privileged: &[]bool{true}[0],
+						RunAsUser:  &[]int64{0}[0],
+					},
+					Command: []string{"sh", "-c", "echo setting the pod as privileged"},
+				},
+			},
+			Containers: []v1.Container{
+				{
+					Name:            common.ClonerSourcePodName,
+					Image:           image,
+					ImagePullPolicy: v1.PullPolicy(pullPolicy),
 				},
 			},
 			RestartPolicy: v1.RestartPolicyNever,
 			Volumes: []v1.Volume{
 				{
-					Name: ImagePathName,
+					Name: DataVolName,
 					VolumeSource: v1.VolumeSource{
 						PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
 							ClaimName: sourcePvcName,
@@ -679,7 +689,52 @@ func MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName string, pvc *v1.Per
 			},
 		},
 	}
+
+	var volumeMode v1.PersistentVolumeMode
+	if pvc.Spec.VolumeMode != nil {
+		volumeMode = *pvc.Spec.VolumeMode
+	} else {
+		volumeMode = v1.PersistentVolumeFilesystem
+	}
+	if volumeMode == v1.PersistentVolumeBlock {
+		pod.Spec.Containers[0].VolumeDevices = addVolumeDevices()
+		pod.Spec.Containers[0].VolumeMounts = addCloneVolumeMounts("Block", id)
+		pod.Spec.Containers[0].Args = addArgs("source", id, "block")
+	} else {
+		pod.Spec.Containers[0].VolumeMounts = addCloneVolumeMounts("FS", id)
+		pod.Spec.Containers[0].Args = addArgs("source", id, "FS")
+	}
 	return pod
+}
+
+func addArgs(source, id, volumeMode string) []string {
+	args := []string{
+		source, id, volumeMode,
+	}
+	return args
+}
+
+func addCloneVolumeMounts(volumeMode, id string) []v1.VolumeMount {
+	if volumeMode == "Block" {
+		volumeMounts := []v1.VolumeMount{
+			{
+				Name:      socketPathName,
+				MountPath: common.ClonerSocketPath + "/" + id,
+			},
+		}
+		return volumeMounts
+	}
+	volumeMounts := []v1.VolumeMount{
+		{
+			Name:      DataVolName,
+			MountPath: common.ClonerImagePath,
+		},
+		{
+			Name:      socketPathName,
+			MountPath: common.ClonerSocketPath + "/" + id,
+		},
+	}
+	return volumeMounts
 }
 
 // CreateCloneTargetPod creates our cloning tgt pod which will be used for out of band cloning to write the contents of the tgt PVC
@@ -757,27 +812,38 @@ func MakeCloneTargetPodSpec(image, pullPolicy, podAffinityNamespace string, pvc 
 					},
 				},
 			},
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
+			// The pod has to be privileged as it has to have access to the hostPath in the node.
+			// However, currently there is a bug that we cannot attach block device to the pod if the pod (container)
+			// is privileged:
+			//https://github.com/kubernetes/kubernetes/issues/58251
+			//https://github.com/kubernetes/kubernetes/issues/62560
+			// As a result of that, instead setting the SecurityContext field as Privileged, in the main container
+			// where we need to have access to  the hostPath, we set the SecurityContext field in the initContainer,
+			// and that will do it.
+			InitContainers: []v1.Container{
 				{
-					Name:            common.ClonerTargetPodName,
-					Image:           image,
-					ImagePullPolicy: v1.PullPolicy(pullPolicy),
-					SecurityContext: &v1.SecurityContext{
-						Privileged: &[]bool{true}[0],
-						RunAsUser:  &[]int64{0}[0],
-					},
-
+					Name:  "init",
+					Image: image,
 					VolumeMounts: []v1.VolumeMount{
-						{
-							Name:      ImagePathName,
-							MountPath: common.ClonerImagePath,
-						},
 						{
 							Name:      socketPathName,
 							MountPath: common.ClonerSocketPath + "/" + id,
 						},
 					},
-					Args: []string{"target", id},
+					SecurityContext: &v1.SecurityContext{
+						Privileged: &[]bool{true}[0],
+						RunAsUser:  &[]int64{0}[0],
+					},
+					Command: []string{"sh", "-c", "echo setting the pod as privileged"},
+				},
+			},
+
+			Containers: []v1.Container{
+				{
+					Name:            common.ClonerTargetPodName,
+					Image:           image,
+					ImagePullPolicy: v1.PullPolicy(pullPolicy),
 					Ports: []v1.ContainerPort{
 						{
 							Name:          "metrics",
@@ -796,7 +862,7 @@ func MakeCloneTargetPodSpec(image, pullPolicy, podAffinityNamespace string, pvc 
 			RestartPolicy: v1.RestartPolicyNever,
 			Volumes: []v1.Volume{
 				{
-					Name: ImagePathName,
+					Name: DataVolName,
 					VolumeSource: v1.VolumeSource{
 						PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
 							ClaimName: pvc.Name,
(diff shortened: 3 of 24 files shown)


cmd/cdi-cloner/cloner_startup.sh
```diff
@@ -33,15 +40,24 @@ if [ "$obj" == "source" ]; then
...
-    echo $size
-    #Write the size to the pipe so the other end can read it.
-    echo "$size" >$pipe_dir
-    tar cv --sparse . >$pipe_dir
-    popd
+    if [ "$volumeMode" != "block" ]; then
+        echo "cloner: creating tarball of the image and redirecting it to $pipe_dir"
+        pushd $image_dir
+	    #figure out the size of content in the directory
+        size=$(du -sb . | cut -f1)
```
zvikorn: @awles Can you please point me where we use the size we echo into the pipe, in the target pod? I guess it's somewhere in target_cloner.go but I can't find it. Also I  assume we cannot use this approach with block device. Can we? Do you have an idea what can be done instead?

awels: We use the size to determine the total to calculate the progress of the clone. That is the only reason. It is used in the CountingReader its transparent to the writer of the file.

zvikorn: but size=$(du -sb . | cut -f1) won't work on /data/clone which is block device. right?

zvikorn:
@awels ready for review. especially clonetarget.go where we use prometheus

pkg/controller/util.go
```diff
@@ -633,33 +633,43 @@ func MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName string, pvc *v1.Per
...
-					Name:            common.ClonerSourcePodName,
-					Image:           image,
-					ImagePullPolicy: v1.PullPolicy(pullPolicy),
-					SecurityContext: &v1.SecurityContext{
-						Privileged: &[]bool{true}[0],
-						RunAsUser:  &[]int64{0}[0],
-					},
-
+					Name:  "myinit",
+					Image: "registry.hub.docker.com/library/busybox:latest",
```
zvikorn: Any other suggestions for a better image?

zvikorn:
@aglitke @awels @igoihman @mhenriks @j-griffith @rollandf @danielerez @annastopel 
Still missing functional tests, but it would be great if you can start reviewing this. Thanks!

zvikorn:
ci test please

pkg/controller/util.go
mhenriks: You can use the same image as the cloner pod.  Just set the container Command/Args params appropriately.

I'd also recommend you name the container `init` instead of `myinit`.

cmd/cdi-cloner/cloner_startup.sh
awels: no but you know how to get the size of the block device with lsblk, we do it in the importer.

cmd/cdi-cloner/cloner_startup.sh
```diff
@@ -33,15 +40,24 @@ if [ "$obj" == "source" ]; then
...
+        echo "going to echo size to pipe dir"
+        #Write the size to the pipe so the other end can read it.
+        echo "$size" >$pipe_dir
+        echo "before tar"
+        tar cv --sparse . >$pipe_dir
+        echo "tar completed"
+        popd
+    else
+        echo "cloner: writing the image to $pipe_dir"
+        dd if=$image_dir bs=4M > $pipe_dir
```
awels: you can do: dd if=$image_dir bs=4M of=$pipe_dir should save you one shell invocation I think.

cmd/cdi-cloner/cloner_startup.sh
```diff
@@ -51,10 +67,14 @@ if [ "$obj" == "target" ]; then
     while true; do
         echo "cloner: check if the fifo pipe was created by the cloning source pod"
         if [ -e "$pipe_dir" ]; then
-            pushd $image_dir
+            if [ "$volumeMode" != "block" ]; then
```
awels: The check if we created the pipe is relevant in all cases no?

pkg/controller/util.go
```diff
@@ -633,33 +633,43 @@ func MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName string, pvc *v1.Per
...
 		},
 		Spec: v1.PodSpec{
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
+			// The pod has to be privileged as it has to have access to the hostPath in the node.
+			// However, currently there is a bug that we cannot attach block device to the pod if the pod (container)
+			// is privileged:
+			//https://github.com/kubernetes/kubernetes/issues/58251
+			//https://github.com/kubernetes/kubernetes/issues/62560
+			// As a reulst of that, instread setting the SecurityContext field as Privileged, in the main conatiner
```
awels: As a result of that, instead
in the main container

pkg/controller/util.go
```diff
@@ -756,27 +816,38 @@ func MakeCloneTargetPodSpec(image, pullPolicy, podAffinityNamespace string, pvc
...
 				},
 			},
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
+			// The pod has to be privileged as it has to have access to the hostPath in the node.
+			// However, currently there is a bug that we cannot attach block device to the pod if the pod (container)
+			// is privileged:
+			//https://github.com/kubernetes/kubernetes/issues/58251
+			//https://github.com/kubernetes/kubernetes/issues/62560
+			// As a reulst of that, instread setting the SecurityContext field as Privileged, in the main conatiner
```
awels: same typos

pkg/controller/util_test.go
```diff
@@ -1506,8 +1519,8 @@ func createCloneController(pvcSpec *v1.PersistentVolumeClaim, sourcePodSpec *v1.
 
 func createSourcePod(pvc *v1.PersistentVolumeClaim, id string) *v1.Pod {
 	_, sourcePvcName := ParseSourcePvcAnnotation(pvc.GetAnnotations()[AnnCloneRequest], "/")
-	// source pod name contains the pvc name
-	podName := fmt.Sprintf("%s-", ClonerSourcePodName)
+	podName := fmt.Sprintf("%s-", common.ClonerSourcePodName)
+	//id := string(pvc.GetUID())
```
awels: id is commented out here?

pkg/controller/util_test.go
```diff
@@ -1540,32 +1553,43 @@ func createSourcePod(pvc *v1.PersistentVolumeClaim, id string) *v1.Pod {
...
 		},
 		Spec: v1.PodSpec{
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
+			// The pod has to be privileged as it has to have access to the hostPath in the node.
+			// However, currently there is a bug that we cannot attach block device to the pod if the pod (container)
+			// is privileged:
+			//https://github.com/kubernetes/kubernetes/issues/58251
+			//https://github.com/kubernetes/kubernetes/issues/62560
+			// As a reulst of that, instread setting the SecurityContext field as Privileged, in the main conatiner
```
awels: same typo

pkg/controller/util_test.go
```diff
@@ -1577,22 +1601,42 @@ func createSourcePod(pvc *v1.PersistentVolumeClaim, id string) *v1.Pod {
...
+		pod.Spec.Containers[0].VolumeMounts = addCloneVolumeMounts("Block", id)
+		pod.Spec.Containers[0].Args = addArgs("source", id, "block")
+	} else {
+		pod.Spec.Containers[0].VolumeMounts = addCloneVolumeMounts("FS", id)
+		pod.Spec.Containers[0].Args = addArgs("source", id, "FS")
+	}
 	return pod
 }
 
+/*func createSourcePodBlockPvc(pvc *v1.PersistentVolumeClaim, id string) *v1.Pod {
```
awels: Commented out?

pkg/controller/util_test.go
```diff
@@ -1577,22 +1601,42 @@ func createSourcePod(pvc *v1.PersistentVolumeClaim, id string) *v1.Pod {
...
+	pod.Spec.Containers[0].VolumeMounts = nil
+	pod.Spec.Containers[0].VolumeDevices = addVolumeDevices()
+	return pod
+}*/
+
 func createTargetPod(pvc *v1.PersistentVolumeClaim, id, podAffinityNamespace string) *v1.Pod {
-	// target pod name contains the pvc name
-	podName := fmt.Sprintf("%s-", ClonerTargetPodName)
+	podName := fmt.Sprintf("%s-", common.ClonerTargetPodName)
+	//id := string(pvc.GetUID())
```
awels: commented out?

pkg/controller/util_test.go
```diff
@@ -1630,38 +1674,50 @@ func createTargetPod(pvc *v1.PersistentVolumeClaim, id, podAffinityNamespace str
...
 				},
 			},
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
+			// The pod has to be privileged as it has to have access to the hostPath in the node.
+			// However, currently there is a bug that we cannot attach block device to the pod if the pod (container)
+			// is privileged:
+			//https://github.com/kubernetes/kubernetes/issues/58251
+			//https://github.com/kubernetes/kubernetes/issues/62560
+			// As a reulst of that, instread setting the SecurityContext field as Privileged, in the main conatiner
```
awels: same typo

pkg/controller/util_test.go
```diff
@@ -1692,16 +1748,33 @@ func createTargetPod(pvc *v1.PersistentVolumeClaim, id, podAffinityNamespace str
...
+		pod.Spec.Containers[0].Args = addArgs("target", id, "block")
+		pod.Spec.Containers[0].Env = addEnvVar("block")
+	} else {
+		pod.Spec.Containers[0].VolumeMounts = addCloneVolumeMounts("FS", id)
+		pod.Spec.Containers[0].Args = addArgs("target", id, "FS")
+	}
 	return pod
 }
 
+/*func createTargetPodBlockPvc(pvc *v1.PersistentVolumeClaim, id, podAffinityNamespace string) *v1.Pod {
```
awels: commented out?

cmd/cdi-cloner/cloner_startup.sh
zvikorn: we do check if the pipe exists. This check is only done before we pushd the directory. As I'm thinking about it now, I do not get exactly why we do this (pushd/popd). The only code we execute between the pushd and popd is '/usr/bin/cdi-cloner -pipedir $pipe_dir -alsologtostderr -v=3'. isn't it leftover from earlier design?

zvikorn:
@awels can you also please review the functional test?
I updated the code. ms5sum of source and clone loopback files are different. Do you see something wrong/missing? Does the clonetarget.go with the prometheus part look ok? Thanks!

zvikorn:
ci test please

cmd/cdi-cloner/cloner_startup.sh
```diff
@@ -33,15 +40,24 @@ if [ "$obj" == "source" ]; then
...
-    echo "cloner: creating tarball of the image and redirecting it to $pipe_dir"
-    pushd $image_dir
-	#figure out the size of content in the directory
-    size=$(du -sb . | cut -f1)
-    echo $size
-    #Write the size to the pipe so the other end can read it.
-    echo "$size" >$pipe_dir
-    tar cv --sparse . >$pipe_dir
-    popd
+    if [ "$volumeMode" != "block" ]; then
```
awels: this should be:
```bash
    if [ "$volumeMode" == "fs" ]; then
        pushd $image_dir
        #figure out the size of content in the directory
        size=$(du -sb . | cut -f1)
        echo $size
        #Write the size to the pipe so the other end can read it.
        echo "$size" >$pipe_dir
        tar cv --sparse . >$pipe_dir
        popd
    elif [ "$volumeMode" == "block" ]; then
        #figure out the size of the block device
        size=$(lsblk -n -b -o SIZE $image_dir) 
        echo $size
        #Write the size to the pipe so the other end can read it.
        echo "$size" >$pipe_dir
        echo "cloner: writing the image to $pipe_dir"
        dd if=$image_dir bs=4M of=$pipe_dir 
    fi
```
This way you are explicitly checking for the right mode, and for BOTH modes you are passing in the size in first, and then you are writing the actual data. Also the block size of 4M seems awfully large. Now we can have a huge long debate over the best size of a block with DD, but going a quick google search it appears the optimal size to be somewhere in the 32-128k range. So I guess 64k seems like a decent middle ground.

cmd/cdi-cloner/cloner_startup.sh
awels: I was wrong, the pushd and popd are for switching to the target directory, which is not relevant for block devices.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -44,6 +44,8 @@ type DataVolumeSpec struct {
 	PVC *corev1.PersistentVolumeClaimSpec `json:"pvc"`
 	//DataVolumeContentType options: "kubevirt", "archive"
 	ContentType DataVolumeContentType `json:"contentType,omitempty"`
+	//VolumeMode indicates whether the PV is a block PV or not
```
awels: This is not needed, the volumeMode is defined in the PVC section.

cmd/cdi-cloner/cloner_startup.sh
zvikorn: Won't 'dd if=$image_dir bs=4M of=$pipe_dir' overwrite the existing $pipe_dir that already contains the size? will it append?

pkg/common/common.go
```diff
@@ -109,4 +107,7 @@ const (
 
 	// ScratchSpaceNeededExitCode is the exit code that indicates the importer pod requires scratch space to function properly.
 	ScratchSpaceNeededExitCode = 42
+
+	// VolumeMode is environmet variable indicating the volume mode of the PV (block/FS)
```
awels: Just like with the importer, we don't have to pass the volume mode to the pod because the pod can determine by checking the existence of the block device. So we don't have to set the environment variable at all.

pkg/controller/util.go
```diff
@@ -678,9 +688,59 @@ func MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName string, pvc *v1.Per
...
+		pod.Spec.Containers[0].Args = addArgs("source", id, "block")
+	} else {
+		pod.Spec.Containers[0].VolumeMounts = addCloneVolumeMounts("FS", id)
+		pod.Spec.Containers[0].Args = addArgs("source", id, "FS")
+	}
 	return pod
 }
 
+// return the Env portion for the importer container.
+func addEnvVar(volumeMode string) []v1.EnvVar {
```
awels: Per comment above, don't need this.

cmd/cdi-cloner/cloner_startup.sh
awels: Its a pipe, it acts like a stream, anything that is written to it, will get appended.

zvikorn:
@awels updated the patch. Cloned is good now. Still need to update the FT, but it would be great if you review the changes I have made.

doc/target-local-block-pv.md
```diff
@@ -0,0 +1,78 @@
+# How to clone an image from one block raw PV to another one
```
awels: So I am trying to understand what is going on in this document, and I think we are missing the point a little bit. We already have a document explaining how to create a block PV, so we don't need to repeat that. I think the point we should be trying to make is the following:
1. Given a block PVC source, and an existing block PV, what do we need to put in the target block PVC yaml to make it work. Note we should not be mixing block and non block while cloning.

manifests/example/clone-block-pv.yaml
```diff
@@ -0,0 +1,24 @@
+kind: PersistentVolume
```
awels: I don't think we need multiple examples of how to create a block PV. The one from the importer should be sufficient.

doc/target-local-block-pv.md
zvikorn: so in that case, the user needs also to assume that there is another blockDevice (/dev/loop11) already defined on the host (except the clone PV with the other local path). Is that ok with you? Not to show how to define the block Device on the host?

tests/framework/pvc.go
```diff
@@ -61,13 +61,38 @@ func (f *Framework) CreateAndPopulateSourcePVC(pvcName string, podName string, f
 	return sourcePvc
 }
 
+// CreateAndPopulateSourcePVCForBlockPV Creates and populates a PVC using the provided POD and command for block PV
+func (f *Framework) CreateAndPopulateSourcePVCForBlockPV(pvcName string, podName string, fillCommand string) *k8sv1.PersistentVolumeClaim {
```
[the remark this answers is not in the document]
zvikorn: I did not get you. Your example here still shows another method (CreateAndPopulateSourcePVCForBlockPV), just with pvcDef parameter instead pvcName. so we remain with two similar methods.  There is also a call to utils.NewPodWithBlockPVC vs utils.NewPodWithPVC
If we want to have one method instead of two, we will have to create either the FS or Block pvc def, and either the FS or Block Pod, in advance, and then call CreateAndPopulateSourcePVCForPV with pvcDef and podDef as parameters.

pkg/controller/util_test.go
```diff
@@ -1349,6 +1353,15 @@ func createClonePvc(name, ns string, annotations, labels map[string]string) *v1.
 			},
 		},
 	}
+	pvc.Spec.VolumeMode = &VolumeMode
```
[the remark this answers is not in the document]
zvikorn: I know. However, I got an error without it (which is weird, I agree). I am removing it and let's see the result.

pkg/util/util.go
```diff
@@ -189,3 +189,19 @@ func MoveFileAcrossFs(src, dst string) error {
 	}
 	return os.Remove(src)
 }
+
+// CopyBlock reads the disk image from the named pipe (FIFO) and writes it to a block device destination
+func CopyBlock(reader io.Reader, destDir string, arg ...string) error {
```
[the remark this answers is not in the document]
zvikorn: done

pkg/controller/util_test.go
```diff
@@ -1631,38 +1667,50 @@ func createTargetPod(pvc *v1.PersistentVolumeClaim, pvcUID, podAffinityNamespace
...
 							},
 							Namespaces:  []string{podAffinityNamespace}, //the scheduler looks for the namespace of the source pod
-							TopologyKey: CloningTopologyKey,
+							TopologyKey: common.CloningTopologyKey,
 						},
 					},
 				},
 			},
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
```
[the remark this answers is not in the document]
zvikorn: done

pkg/controller/util_test.go
```diff
@@ -1541,32 +1553,43 @@ func createSourcePod(pvc *v1.PersistentVolumeClaim, pvcUID string) *v1.Pod {
 			},
 		},
 		Spec: v1.PodSpec{
-			Containers: []v1.Container{
+			// We create initContainer just to set the pod as privileged.
```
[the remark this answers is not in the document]
zvikorn: done

pkg/controller/util_test.go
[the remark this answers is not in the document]
zvikorn: This is what I get:
https://travis-ci.org/kubevirt/containerized-data-importer/builds/518249481#L4932

tests/framework/pvc.go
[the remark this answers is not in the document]
awels: I copied the name wrong, I meant to take away the block part. Basically we end up with 1 function, and when you call it you pass in the definition you want.

doc/target-local-block-pv.md
awels: The manual creation of the block PVs is done for the benefit of our CI system so we know some block PVs are available. But in general we can assume that if a user wants a block PV they have some mechanism available to create it for them. Since we have already shown the user how to manually create one block PV creating another one should be obvious.

pkg/util/util.go
```diff
@@ -189,3 +189,25 @@ func MoveFileAcrossFs(src, dst string) error {
 	}
 	return os.Remove(src)
 }
+
+// StreamDataToFile provides a function to stream the specified io.Reader to the specified local file
```
awels: I meant USE pkg/importer/util.go version of StreamDataToFile, not copy it.

pkg/controller/util_test.go
```diff
@@ -1700,6 +1724,15 @@ func createTargetPod(pvc *v1.PersistentVolumeClaim, pvcUID, podAffinityNamespace
 			},
 		},
 	}
+	volumeMode := pvc.Spec.VolumeMode
```
awels: This is where you are getting your npe.
So instead of
```go
volumeMode := pvc.Spec.VolumeMode
```
Do
```go
var volumeMode v1.PersistentVolumeMode
if pvc.Spec.VolumeMode != nil {
   volumeMode = *pvc.Spec.VolumeMode
} else {
    volumeMode = v1.PersistentVolumeFilesystem
}

if volumeMode == v1.PersistentVolumeBlock {
...
}
```

And do this for all 4 places where you have this problem, the test will pass.

zvikorn:
I wanted to do that, but you said that the volumeMode is by default filesystem and not nil...

awels:
it is by default file system, if it is not there

doc/target-local-block-pv.md
zvikorn: ok

tests/framework/pvc.go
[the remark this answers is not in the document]
zvikorn: Is it right that to do so, I need to create the pvc def and pod def before we call this function? Meaning, changing all places we call this function in cloner_test.go and datavolume_test.go?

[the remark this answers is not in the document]
awels: Yes, that would be an improvement on the current state and we should always try to improve the existing code as well, if it is something we are touching.

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
ci test please

doc/target-local-block-pv.md
awels: So any reason not to remove this file, we have basically the identical example for import?

zvikorn: This is the only yaml to include clone dv example with volumeMode:Block.

awels: Oh I was looking at the name of the file, I thought it was describing the creation of a block PV. If you could rename the file to better reflect the purpose of the file

zvikorn: clone-block-datavolume.md?

awels: Sounds good to me, that way the name of the file explains what is in it.

tests/cloner_test.go
```diff
@@ -30,9 +36,77 @@ var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
 	var sourcePvc *v1.PersistentVolumeClaim
+	var targetPvc *v1.PersistentVolumeClaim
+	AfterEach(func() {
+		if sourcePvc != nil {
+			By("[AfterEach] Clean up target PVC")
+			err := f.DeletePVC(sourcePvc)
+			Expect(err).ToNot(HaveOccurred())
+		}
+		if targetPvc != nil {
+			By("[AfterEach] Clean up target Block PVC")
```
awels: Oops forgot to change the message to not include block here, should be Clean up target PVC

zvikorn:
done

PR #631, opened 2019-01-31
Operator cleanup

mhenriks:
Renamed the operator leader configmap.  Also make sure any secrets that don't have an owner are owned by the CDI object so they'll get cleaned up when CDI is delete.  To do all this I had to move some variables/packages around to avoid circular deps.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/apiserver/apiserver_test.go
```diff
@@ -157,6 +157,16 @@ func tlsSecretGetAction() core.Action {
 		apiCertSecretName)
 }
 
+func cdiConfigGetAction() core.Action {
```
igoihman: maybe change to `cdiConfigMapGetAction`?
I'm adding `CDIConfig` CR now so naming is confusing.

pkg/keys/keystore.go
```diff
@@ -25,14 +25,15 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
igoihman: is there a reason for this change? this is the same package name

awels: His IDE automatically does that, I wondered the same a while ago.

pkg/apiserver/apiserver_test.go
mhenriks: Sounds good

PR #639, opened 2019-02-06
BugFix: cdi importer fails to import from registry when run in unpriv…

annastopel:
…iledged container

Fixes #

Now import from registry will succeed on images with content type kubevirt if the image contains VM disk image file in the root of the container. The filename has to be disk.img.
The following errors are reported by the importer:
- On container image that does not contain VM disk image file on its root  - "Failed to find VM disk image file in the container image"
- On import PVC with registry source and contentType archive - "Content type \"archive\" is currently not supported when importing from registry"
**Release note**:

```release-note
Bug Fix: cdi import from registry fails when runs in unprivileged mode.
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -51,6 +51,12 @@ func main() {
 	contentType, _ := util.ParseEnvVar(common.ImporterContentType, false)
 	imageSize, _ := util.ParseEnvVar(common.ImporterImageSize, false)
 
+	//Registry import currently support only kubevirt content type
+	if contentType != string(cdiv1.DataVolumeKubeVirt) && source == controller.SourceRegistry {
+		glog.Errorf("Unsupported content type %s when importing from registry", contentType)
+		os.Exit(1)
+	}
+
 	dest := common.ImporterWritePath
 	if contentType == string(cdiv1.DataVolumeArchive) || source == controller.SourceRegistry {
 		dest = common.ImporterVolumePath
--- a/pkg/image/skopeo.go
+++ b/pkg/image/skopeo.go
@@ -77,25 +77,32 @@ func (o *skopeoOperations) CopyImage(url, dest, accessKey, secKey string) error 
 }
 
 // CopyRegistryImage download image from registry with skopeo
-func CopyRegistryImage(url, dest, accessKey, secKey string) error {
+func CopyRegistryImage(url, dest, destFile, accessKey, secKey string) error {
 	skopeoDest := "dir:" + dest + dataTmpDir
 	err := SkopeoInterface.CopyImage(url, skopeoDest, accessKey, secKey)
 	if err != nil {
 		os.RemoveAll(dest + dataTmpDir)
 		return errors.Wrap(err, "Failed to download from registry")
 	}
-	err = extractImageLayers(dest)
+	err = extractImageLayers(dest, destFile)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


pkg/image/skopeo.go
```diff
@@ -77,25 +77,32 @@ func (o *skopeoOperations) CopyImage(url, dest, accessKey, secKey string) error
...
 		return errors.Wrap(err, "Failed to download from registry")
 	}
-	err = extractImageLayers(dest)
+	err = extractImageLayers(dest, destFile)
 	if err != nil {
 		return errors.Wrap(err, "Failed to extract image layers")
 	}
 
+	//If a specifc file was requested verify it exists, if not - fail
+	if len(destFile) > 0 {
```
awels: if len(destFile) > 0 && err != nil {

Otherwise we can potentially overwrite an existing error.

pkg/image/skopeo.go
```diff
@@ -120,9 +127,16 @@ var extractImageLayers = func(dest string) error {
 		layer := strings.TrimPrefix(layerID, "sha256:")
 		filePath := filepath.Join(dest, dataTmpDir, layer)
 
-		if err := util.UnArchiveLocalTar(filePath, dest, "z"); err != nil {
-			return errors.Wrap(err, "could not extract layer tar")
+		//prepend z option to the beggining of untar arguments
+		args := append([]string{"z"}, arg...)
+
+		if err := util.UnArchiveLocalTar(filePath, dest, args...); err != nil {
+			//ignore errors if specific file extract was requested - we validate wheather the file was extracted at the end of the sequenec
```
awels: //ignore errors if specific file extract was requested - we validate whether the file was extracted at the end of the sequence

just some small typos.

pkg/util/util.go
```diff
@@ -122,3 +129,15 @@ func UnArchiveLocalTar(filePath, destDir string, arg ...string) error {
 	fileReader := bufio.NewReader(file)
 	return UnArchiveTar(fileReader, destDir, arg...)
 }
+
+/*
```
awels: This is commented out, can we just remove it, if we are not using it?

awels:
in general looks good to me, just a few minor comments.

cmd/cdi-importer/importer.go
```diff
@@ -51,6 +51,12 @@ func main() {
 	contentType, _ := util.ParseEnvVar(common.ImporterContentType, false)
 	imageSize, _ := util.ParseEnvVar(common.ImporterImageSize, false)
 
+	//Registry import currently support only kubevirt content type
+	if contentType == string(cdiv1.DataVolumeArchive) && source == controller.SourceRegistry {
```
aglitke: Might it be better to check if content-type is ! kubevirt ?

cmd/cdi-importer/importer.go
```diff
@@ -51,6 +51,12 @@ func main() {
...
+		glog.Errorf("Content type \"archive\" is currently not supported when importing from registry")
```
aglitke: Unsupported content type: %s

pkg/image/skopeo.go
annastopel: if there was an error it was returned in line 89

awels:
ci test please

pkg/image/skopeo.go
awels: Good point, I missed the return.

PR #495, opened 2018-10-18
Import pvc 'cdi.kubevirt.io/storage.pod.phase' annotation valu…

zvikorn:
…e is wrong when import fails. Issue #494

Signed-off-by: tavni <tavni@redhat.com>

**What this PR does / why we need it**:
Fix for Issue #494 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #494 

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #644, opened 2019-02-08
Import pod status should not be 'Failed' when starting up

mhenriks:
The import controller was marking import pods as failed when they were starting up.  This was not playing well with kubevirt as it deletes/recreates dataolumes on when they enter failed state.  Basically, datavolumes were getting deleted before they were populated.

For now, just making sure we don't mark failure when starting up but bigger conversation is necessary to determine exactly if/when an import pod should be marked as failed.

#646 and #495 related

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -2,9 +2,10 @@ package controller
 
 import (
 	"fmt"
+
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


awels:
ci test please

awels:
ci test please

PR #646, opened 2019-02-08
Update README

j-griffith:

Make the README a little more concise, but still get the main points
across.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,265 +1,80 @@
 # Containerized Data Importer
+
+[![Build Status](https://travis-ci.org/kubevirt/containerized-data-importer.svg?branch=master)](https://travis-ci.org/kubevirt/containerized-data-importer)
 [![Go Report Card](https://goreportcard.com/badge/github.com/kubevirt/containerized-data-importer)](https://goreportcard.com/report/github.com/kubevirt/containerized-data-importer)
 [![Coverage Status](https://img.shields.io/coveralls/kubevirt/containerized-data-importer/master.svg)](https://coveralls.io/github/kubevirt/containerized-data-importer?branch=master)
+[![Licensed under Apache License version 2.0](https://img.shields.io/github/license/kubevirt/containerized-data-importer.svg)](https://www.apache.org/licenses/LICENSE-2.0)
 
-A declarative Kubernetes utility to import Virtual Machine images for use with [Kubevirt](https://github.com/kubevirt/kubevirt). At a high level, a persistent volume claim (PVC), which defines VM-suitable storage via a storage class, is created. A custom controller watches for importer specific claims, and when discovered, starts an import/copy process. The status of the import process is reflected in the same claim, and when the copy completes Kubevirt can create the VM based on the just-imported image.
-In Addition, the Containerized Data Importer gives the option to clone the imported VM image from one PVC to another one across two different namespaces (A.K.A Host-Assisted cloning). 
+**Containerized-Data-Importer (CDI)** is a persistent storage management add-on for Kubernetes.
+It's primary goal is to provide a declarative way to build Virtual Machine Disks on PVCs for [Kubevirt](https://github.com/kubevirt/kubevirt) VMs
 
-1. [Purpose](#purpose)
-1. [Versions](#versions)
-1. [Design](/doc/design.md#design)
-1. [Running the CDI Controller](#deploying-cdi)
-1. [Cloning VM Images](#cloning-vm-images)
-1. [Hacking (WIP)](hack/README.md#getting-started-for-developers)
-1. [Security Configurations](#security-configurations)
-
-## Overview
-
-### Purpose
-
-This project is designed with Kubevirt in mind and provides a declarative method for importing VM images into a Kuberenetes cluster.
-Kubevirt detects when the VM image copy is complete and, using the same PVC that triggered the import process, creates the VM.
-
-This approach supports two main use-cases:
--  a cluster administrator can build an abstract registry of immutable images (referred to as "Golden Images") which can be cloned and later consumed by Kubevirt, or
--  an ad-hoc user (granted access) can import a VM image into their own namespace and feed this image directly to Kubevirt, bypassing the cloning step.
+CDI works with standard core Kubernetes resources and is storage device agnostic, while its primary focus is to build disk images for Kubevirt, it's also useful outside of a Kubevirt context to use for initializing your Kubernetes Volumes with data.
 
 
-For an in depth look at the system and workflow, see the [Design](/doc/design.md#design) documentation.
+# Introduction
 
-### Versions
+## Kubernetes extension to populate PVCs with VM images
+CDI provides the ability to populate PVCs with VM images upon creation.  Multiple image formats and sources are supported:
 
-CDI follows the common semantic version scheme defined at semver.org in the "vMajor.Minor.Patch" pattern.  These are defined as:
+### Current supported Image formats
+* .tar
+* .gz
+* .xz
+* .img
+* .iso
+* .qcow2
 
-- Major: API or other changes that will break existing CDI deployments.  These changes are expected to require users to alter the way they interact with CDI.  Major versions are released at the end a development cycle (every 2 weeks) in leiu of a Minor version.
+### Current supported image endpoints
+* http
+* S3
+* local directory
 
-- Minor: Backwards compatible changes within the current Major version.  These changes represent the products of a 2 week developement cycle and contain bug fixes and new features.  By releasing merged work at the end of the cycle, users are able to closely track the project's progress and report issues or bugs soon after they are introduced.  At the same time, it should be easy to roll back to the previous Minor version if the release blocks the user's workflow.
+## DataVolumes
+CDI also includes a CRD, that provides an object of type DataVolume.  The DataVolume is an abstraction on top of the standard Kubernetes PVC and can be used to automate creation and population of a PVC for consumption in a Kubevirt VM.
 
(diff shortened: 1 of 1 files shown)


jekader:
I was doing CI maintenance and the build running checks for this PR was aborted. Please re-trigger to get proper results.

README.md
```diff
@@ -2,264 +2,65 @@
...
 
-A declarative Kubernetes utility to import Virtual Machine images for use with [Kubevirt](https://github.com/kubevirt/kubevirt). At a high level, a persistent volume claim (PVC), which defines VM-suitable storage via a storage class, is created. A custom controller watches for importer specific claims, and when discovered, starts an import/copy process. The status of the import process is reflected in the same claim, and when the copy completes Kubevirt can create the VM based on the just-imported image.
-In Addition, the Containerized Data Importer gives the option to clone the imported VM image from one PVC to another one across two different namespaces (A.K.A Host-Assisted cloning). 
+The Containerized Data Importer (CDI) is an open source, Apache licensed, Kubernetes Controller to import data from various sources to a Kubernetes Volume (PVC) in a declarative manner. CDI includes a CRD Object Type to introduce DataVolumes to automate the creation and population of PVCs.  Users can access the PVCs created by the DataVolumes directly, or when using Kubevirt, you can use the DataVolumes in Kubevirt directly.
```
karmab: when using Kubevirt, can use the DataVolumes

README.md
```diff
@@ -2,264 +2,65 @@
...
 
-1. [Purpose](#purpose)
-1. [Versions](#versions)
-1. [Design](/doc/design.md#design)
-1. [Running the CDI Controller](#deploying-cdi)
-1. [Cloning VM Images](#cloning-vm-images)
-1. [Hacking (WIP)](hack/README.md#getting-started-for-developers)
-1. [Security Configurations](#security-configurations)
+CDI is part of the [Kubevirt project](http://kubevirt.io/), and it's primary purpose is to provide an easy, declarative way to build Virtual Machine Disks on PVCs for Kubevirt VMs.  CDI handles multiple data source types, including HTTPS, Local Files, S3 and existing PVCs, and not only handles retrieving data from various sources, but also expands or converts that data as necessary (ie expand tgz files, or convert qcow to raw).
```
karmab: its primary

README.md
```diff
@@ -2,264 +2,65 @@
...
 
-## Overview
+CDI works with standard core Kubernetes resources and is backend agnostic, while it's primary focus is to build disk images for Kubevirt, it's also useful outside of a Kubevirt context to use for intiializing your Kubernetes Volumes with data.
```
karmab: initializing

README.md
```diff
@@ -2,264 +2,65 @@
...
-- An optional "golden" namespace acting as the image registry. The `default` namespace is fine for tire kicking.
-
-### Download CDI
-
-`$ go get kubevirt.io/containerized-data-importer`
-
-*Or*
-
-`$ git clone https://github.com/kubevirt/containerized-data-importer.git $GOPATH/src/kubevirt.io/containerized-data-importer`
+Deploying the CDI controller is straight forward. In this document the _default_ namespace is used, but in a production setup a [protected namespace](#protecting-the-golden-image-namespace) that is inaccessible to regular users should be used instead.
```
karmab: straightforward

README.md
```diff
@@ -2,264 +2,65 @@
...
 
-*Or download only the yamls:*
+  ```
```
karmab: i think that starting from 1.5.2, we want users to deploy from the operator yaml (as it also bundles the required scc and the configmap)

README.md
```diff
@@ -2,264 +2,65 @@
...
-Deploying the CDI controller is straight forward. In this document the _default_ namespace is used, but in a production setup a [protected namespace](#protecting-the-golden-image-namespace) that is inaccessible to regular users should be used instead.
+## Use it
 
-1. Ensure that the cdi-sa service account has proper authority to run privileged containers, typically in a kube environment this is true by default.
-   If you are running an openshift variation of kubernetes you may need to enable privileged containers in the security context:
+Create a DataVolume (Kubevirt Custom Object) which will provide a PVC with a Cirros disk image
 
 ```
-$ oc adm policy add-scc-to-user privileged -z cdi-sa
+$ kubectl create -f https://raw.githubusercontent.com/kubevirt/containerized-data-importer/release-v1.5/manifests/example/datavolume.yaml
```
karmab: do we want the main workflow to be based on datavolumes ?

README.md
awels: Agreed, we should only support deploying with the operator, so the main example should use it as well.

README.md
j-griffith: I assumed we wanted that as the primary use case, do you disagree?  Or are you suggesting we omit the direct PVC use case?

karmab: what i mean is that since datavolumes are feature gated, maybe they shouldnt be the primary use case shown

awels: Data volumes are feature gated in kubevirt, not in CDI.

README.md
```diff
@@ -1,265 +1,85 @@
...
 
 ```
-apiVersion: v1
-kind: ResourceQuota
-metadata:
-  name: protect-mynamespace
-spec:
-  hard:
-    <STORAGE-CLASS-NAME>.storageclass.storage.k8s.io/requests.storage: "0"
+$ git clone https://github.com/kubevirt/containerized-data-importer && cd containerized-data-importer
```
awels: Don't forget to mention the clone needs to happen in $GOPATH/src/kubevirt.io otherwise the cluster commands won't work since the compile will fail.

README.md
```diff
@@ -1,265 +1,85 @@
...
-> Note: Only set these values if the file server requires authentication credentials.
-
-1. `metadata.name:` Arbitrary name of the secret. Must match the PVC's `cdi.kubevirt.io/storage.import.secretName:`
-
-1.  `accessKeyId:` Contains the endpoint's key and/or user name. This value **must be base64 encoded** with no extraneous linefeeds. Use `echo -n "xyzzy" | base64` or `printf "xyzzy" | base64` to avoid a trailing linefeed
-
-1.  `secretKey:` the endpoint's secret or password, again **base64 encoded** with no extraneous linefeeds.
-
-### Deploy the API Objects
+There are quite a few examples in the [example manifests](https://github.com/kubevirt/containerized-data-importer/tree/release-v1.5/manifests/example), check them out as a reference to create DataVolumes from additional sources like registries, S3 and your local system.
```
awels: I would remove the release part in the path. Something like 
```
[example manifests](./manifests/example)
```
 should do the trick.

README.md
```diff
@@ -1,265 +1,85 @@
...
-```
+## Use it
 
-2. Deploy the controller from the release manifest:
+Create a PVC and load data from an http source on to it (processes a tgz file for you)
 
 ```
-$ VERSION=<cdi version>
-$ kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-controller.yaml
+$ kubectl create -f https://raw.githubusercontent.com/kubevirt/containerized-data-importer/release-v1.5/manifests/example/import-archive-pvc.yaml
```
awels: I just released 1.6, so this is already out of date, I am not particularly fond of the way it was before, but that didn't lock us into a particular version to keep up with.

README.md
```diff
@@ -1,265 +1,85 @@
...
 
-- An optional "golden" namespace acting as the image registry. The `default` namespace is fine for tire kicking.
+  ```
+  $ VERSION=<cdi version>
+  $ kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-operator.yaml
+  $ kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-operator-cr.yaml
+  ```
 
-### Download CDI
+If you're running OpenShift, make sure you enable privileged containers
```
awels: I don't think this is needed anymore due to the operator @mhenriks keep me honest.

mhenriks: correcct

README.md
j-griffith: Added it in the steps, thanks!

README.md
j-griffith: Agreed, updated.

README.md
j-griffith: I'll use the same idea of just going to the local directory instead to avoid that; the other option I considered was $VERSION; but maybe it's reasonable to consider this README being geared towards working from source locally?

My only concern is that we are introducing dependencies in the README for changes to those files in the future, which setting a version solves; but meh.

awels: Well if I understand the intent of that particular example is to show this working for a released version of CDI, and thus should not be pointing to master, but the latest release. I am not sure if there is a good way to not paint ourselves into a particular version corner. I know github has a rest endpoint that allows you to discover the latest release, so with some create curl command you should be able to get the latest version from that endpoint, and then use that to grab the example. Not sure.

README.md
```diff
@@ -1,265 +1,80 @@
...
-```
+## Use it
 
-2. Deploy the controller from the release manifest:
+Create a DataVolume and populate it with data from an http source
 
 ```
-$ VERSION=<cdi version>
-$ kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-controller.yaml
+$ kubectl create -f https://raw.githubusercontent.com/kubevirt/containerized-data-importer/release-v1.6/manifests/example/datavolume.yaml
```
awels: Since you update the install example to use $VERSION, maybe this one as well?

PR #647, opened 2019-02-10
reject DataVolume creation with already exisiting destination PVC

### LINKED ISSUE #627: Disallow creation of DV when PVC with same name already exists.

>
/kind bug
> /kind enhancement

**What happened**:
When trying to create a data volume with the same name as an existing PVC the creation of the data volume object succeeds, but the matching PVC fails because of a naming conflict.

The creation of the DV to fail with an appropriate error message such as "Error when creating .... a PVC with the same name already exists.

- Create a PVC with name "test"
- Create a DV with name "test"
Note the error in the data volume event list, but the creation of the data volume 'succeeded'

**Environment**:
- CDI version 1.5.1
- Kubernetes version 1.11
- Install tools:
- Others:

igoihman:

Fixes #627 

**Release note**:

```release-note
Disallow creation of DataVolume when PVC with same name already exists.
```


pkg/apiserver/webhooks/validating-webhook/util.go
```diff
@@ -0,0 +1,22 @@
+package validatingwebhook
+
+import (
+	"sync"
+
+	"k8s.io/client-go/kubernetes"
+)
+
+var kubeClient *kubernetes.Interface
```
mhenriks: You sure you mean for this to be a pointer to an interface?  I think a regular interface will be fine.

pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -196,6 +196,21 @@ func admitDVs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 		return toAdmissionResponseError(err)
 	}
 
+	client := GetClient()
+	if client != nil {
+		_, err := client.CoreV1().PersistentVolumeClaims(dv.GetNamespace()).Get(dv.GetName(), metav1.GetOptions{})
+		if err == nil {
```
zvikorn: Is it possible that PVC does not exist and still err=nil since there was no explicit error in CoreV1().PersistentVolumeClaims....what happens if this changes in the future?
I recommend doing 'pvc, err' := client.CoreV1().PersistentVolumeClaims(dv.GetNamespace()).Get(dv.GetName(), metav1.GetOptions{})
and then check that pvc is not nil. This way, if  tomorrow  CoreV1().PersistentVolumeClaims is changed and the logic is - if there is no pvc, returns nil,nil (and not nil,err) since there was not error, and not because it could not find pvc, you are good.

igoihman: OK, it seems like pvc is not `nil` if it doesn't exist and we get an empty object.
I changed it to get list of pvcs in the same namespace and fail if I find pvc with the same destination name.

pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -196,6 +196,26 @@ func admitDVs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
...
 	}
 
+	client := GetClient()
+	if client != nil {
+		pvcs, err := client.CoreV1().PersistentVolumeClaims(dv.GetNamespace()).List(metav1.ListOptions{})
+		if err != nil {
+			return toAdmissionResponseError(err)
+		}
+		for _, pvc := range pvcs.Items {
+			if pvc.Name == dv.GetName() {
```
zvikorn: Are we validating that pvc.name = dv.name somewhere? Just in case someone changes the code to create another name for the PVC, we should not allow this.

pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
zvikorn: +1. just a little question....

pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
zvikorn: Actually, isn't better to check the name of the ownerReferences in the pvc?

igoihman: I'm not sure I got you, we want to block the option of creating dv with already existing pvc name in the same namespace.

awels: This check is explicitly for when a pvc not owned by a dv already exists, so there won't be an owner reference.

zvikorn: yeah... I was confused for a second and thought I delete this comment. you two definitely right.

PR #650, opened 2019-02-11
Delete importer pod on pvc delete request

### LINKED ISSUE #525: Referenced POD object not deleted if failing / in retry-loop

### LINKED ISSUE #649: Can't delete PVC when importer pods are running/restarting
/kind bug

**What happened**:
Related to issue #525 we have a similar situation when not using DataVolumes.  The user requests a PVC using the importer, if the Importer POD is in a retry loop and a user requests deletion of the PVC the PVC will be stuck in terminating until the user forcibly deletes the PODs that are retrying forever.

This is what I expect, but it is not ideal from a usability perspective.  Instead, we should go ahead and delete the retrying PODs for the user when we detect the PVC delete request.

1. Create a PVC that uses the importer, but provide an invalid URL 
2. Wait until the POD enters the retry/restart loop
3. Issue a PVC delete no the importer created PVC

j-griffith:
Related to issue #525, we have a similar situation when using PVCs
without DataVolumes.  The user requests a PVC with CDI Importer
annotations, if the Importer POD is in a fail/retry loop and the user
requests deletion of the PVC the PVC will be stuck in terminating
waiting for the user to forcibly delete the running Import POD.

This PR adds a check for PVC delete in the import controller and
will delete the POD explicitly if it sees a PVC delete request.  This
will also be picked up for Import DataVolumes, so we can remove the
extra checks that were previously added in the DV controller.

fixes #649
fixes #525

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1,6 +1,8 @@
 package controller
 
 import (
+	"time"
+
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"k8s.io/api/core/v1"
@@ -13,7 +15,6 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"kubevirt.io/containerized-data-importer/pkg/expectations"
-	"time"
 )
 
 const (
@@ -66,6 +67,7 @@ func NewController(client kubernetes.Interface,
 		UpdateFunc: func(old, new interface{}) {
 			c.enqueuePVC(new)
 		},
+		DeleteFunc: c.enqueuePVC,
 	})
 
 	// Bind the pod SharedIndexInformer to the pod queue
@@ -81,7 +83,6 @@ func NewController(client kubernetes.Interface,
 			}
 			c.handlePodUpdate(new)
 		},
-
 		DeleteFunc: c.handlePodDelete,
 	})
 
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -581,27 +581,9 @@ func (c *DataVolumeController) handleObject(obj interface{}, verb string) {
 			return
 		}
 
-		// BUG: GH Issue #523, currently you can delete a DV and the object will be removed before it's referenced objects are actually
-		// removed (ie POD in a retry loop).  So we need to deal with that by cleaning up any PODs associated with the PVC so that it
-		// can actually be deleted.  The trick here is that we may not have a DV any longer, but still have a PVC and a POD, so deal with it
 		dataVolume, err := c.dataVolumesLister.DataVolumes(object.GetNamespace()).Get(ownerRef.Name)
 		if err != nil {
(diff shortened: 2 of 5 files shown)


j-griffith:
> So far so good but I have a couple questions:
> 
> The controller is not watching for PVC delete events so I think it may be possible to miss deletes. Fine for kubectl because it does a foreground delete which sets the deletion timestamp and adds a finalizer 

True, I was focused on the reported issue of the user not "liking" the default behavior of kubectl, wondering what non-kubectl use case we'd have here?  Anything done by our controllers can set the delete policy to do what we want explicitly to avoid any issues.

> but the object is not really deleted until all the children are. So I wonder if we have an issue with background deletes (which are the default)? If so, cloe/uplod controllers are also affected.

Yes, this focuses on just import for now, but we'd need the same changes to be added to clone and upload as well if we agree on an approach here.

> I believe the code in the datavolume controller that deletes pods should be removed along with this.

Agreed, I'll look at that; I think it needs to be modified; not completely removed. I believe we'll need to do something like delete the PVC so then the other controllers with this addition will clean everything up properly.

pkg/controller/controller.go
```diff
@@ -81,8 +83,6 @@ func NewController(client kubernetes.Interface,
 			}
 			c.handlePodUpdate(new)
 		},
-
-		DeleteFunc: c.handlePodDelete,
```
mhenriks: Why get rid of this?  If some jerk deletes a pod, shouldn't the controller recreate it?

j-griffith: Oops!  That was accidental, nice catch!

j-griffith:
Rebase is KILLING ME

awels:
> Rebase is KILLING ME

Welcome to software engineering?

j-griffith:
Looks like something upstream in coveralls went wonky:  
https://github.com/lemurheavy/coveralls-public/issues/1264 (and a number of duplicates)

j-griffith:
Seeing these same failures on master currently as well:  https://hastebin.com/mepulajima.cpp Looking

j-griffith:
ci test please

PR #653, opened 2019-02-13
Imported iso image file should not be resized

zvikorn:

An ISO file should not be resized to the requested PVC size as an iso file is likely an installation media and resizing it doesn't make sense.

**Release note**:

```release-note
Imported iso image file should not be resized
```

### DIFF
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -80,6 +80,7 @@ type DataStream struct {
 	ctx         context.Context
 	cancel      context.CancelFunc
 	dataCtxt    DataContext
+	isIsoImage  bool
 }
 
 type reader struct {
@@ -210,6 +211,10 @@ func newDataStream(dso *DataStreamOptions, stream io.ReadCloser) (*DataStream, e
 		ds.Size, err = ds.isoSize()
 		if err != nil {
 			return nil, errors.Wrapf(err, "unable to calculate iso file size")
+		}
+		// at that point, only if ds.size != 0 we know for sure that this is an iso file
+		if ds.Size != 0 {
+			ds.isIsoImage = true
 		}
 	}
 	glog.V(3).Infof("NewDataStream: endpoint %q's computed byte size: %d", ep, ds.Size)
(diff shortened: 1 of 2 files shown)


pkg/importer/dataStream.go
```diff
@@ -175,6 +176,10 @@ func newDataStream(dso *DataStreamOptions, stream io.ReadCloser) (*DataStream, e
 		if err != nil {
 			return nil, errors.Wrapf(err, "unable to calculate iso file size")
 		}
+		// at that point, only if ds.size != 0 we know for sure that this is an iso file
+		if ds.Size != 0 {
```
igoihman: I must say I hate this assumption, I think we should check the actual format and not rely on the size value

zvikorn: @igoihman the call to ds.isoSize (L.175) checks if this is iso file and if not return 0. So the logic exists there (look also L.628)

igoihman: what about the calls in `fileFormatSelector`? can't it be changed in the future?
Maybe we should change `ds.isoSize()` call and check if it's iso first.

zvikorn: yeah.... so this is currently the implementation to decide whether it's an iso file or not (day one).
I wonder what other think? if we want to change this, it should be in a different PR as this one is about not resizing iso file only. I do not think the decision if it's an iso or not should be here.

awels: Yeah I don't understand why ds.isoSize() doesn't return an error when the input is not an iso. They have the check in there, but it simply logs the error and returns 0 size. I guess they wanted to differentiate between a fatal error (unable to open the stream) vs a non fatal one, unable to determine that the endpoint is an iso.

So for now I think simply checking is the size > 0 is enough.

awels:
Looks like the e2e test failures are because we are using some iso files, and the test is expecting them to get resized, which they are not longer (which is good, but the test needs to be adjusted to not expect the resize)

zvikorn:
@awels 
Adjusted e2e tests to pass. Thanks!

PR #658, opened 2019-02-14
Fix validate space constraints to properly validate the space needed.

awels:

This PR fixes two bugs, one where the space validation is incorrect and another where the error is not being propagated properly back to the called.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -893,12 +893,9 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
 	}
 
 	//Verify there is enough space in pvc before conversion
-	isEnoughSpace, err := ValidateSpaceConstraint(src, filepath.Dir(dest))
+	err = ValidateSpaceConstraint(src, filepath.Dir(dest))
 	if err != nil {
 		return err
-	}
-	if !isEnoughSpace {
-		return errors.Wrap(err, "Local qcow2 to raw conversion failed - there is not enough space for conversion")
 	}
 
 	glog.V(2).Infoln("converting qcow2 image")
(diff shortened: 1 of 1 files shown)


pkg/importer/dataStream.go
```diff
@@ -973,35 +970,34 @@ func (d *DataStream) copy(dest string) (err error) {
...
 	if info.VirtualSize == 0 || info.ActualSize == 0 {
-		return false, errors.Wrap(err, "Local image validation failed - no image size info is provided")
+		return errors.New("Local image validation failed - no image size info is provided")
 	}
 
-	convRequiredSpace := info.VirtualSize + info.ActualSize
-
+	// The converted file is guaranteed not the be larger than the virtual size, so we need to compare current available space (which is total space
+	// - temporary file actual size) against the worst case scenario which is the conversion actual size equals the virtual size.
+	convRequiredSpace := info.VirtualSize
```
mhenriks: Can we assume that this function will always be called when the qcow2 file has been downloaded to the pvc?  For example, it may make sense to call this before the qcow2 file is downloaded.  Which we could do with http qcow2.

awels: Yes we can, this function is called from convertQcow2ToRaw, which is only called in the copy function, in the path where we are guaranteed that that the temporary file has been copied to the temporary location. Then convertQcow2ToRaw is called to convert the existing temporary file to the target format in the target location. This method is not called in the streaming path.

mhenriks: I understand that this is the case now, but wouldn't it be better if we could validate before we downloaded?  And we can in some cases.  I'm not saying we should do that now.  But maybe it would be better if this function could somehow handle both cases by adding an additional arg or something.

awels: Well I would prefer if this whole method simply went away, and it can with the scratch space initiative. Since we will store the temporary files in the scratch space, this whole calculation is no longer needed at that point. And I just pushed the first WIP PR for scratch space #659

mhenriks: Okay, we'll see how this works out in your PR.

awels:
ci test please

PR #633, opened 2019-02-05
Data stream refactor

annastopel:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
In this refactor all dataStream selection logic is taken out of dataStream.go file and placed in dedicated file with respect to endpoint type. 
**Release note**:

```release-note
None
```

PR #659, opened 2019-02-15
Scratch space retry framework logic.

awels:

This PR contains the framework to create a scratch PVC for temporary storage that is needed for certain import/upload operations. There needs to be some temporary storage that allows the operation to complete. After the operation is finished this storage needs to be made available again.

Fixes #

Currently this is only implemented fully for registry import, I just want some feedback on the overal design.

**Release note**:

```release-note
When required scratch space is created for import/upload operations.
```


pkg/controller/import-controller.go
```diff
@@ -72,6 +93,56 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+func (ic *ImportController) importerRequiresScratchSpace(pvc *v1.PersistentVolumeClaim) bool {
+	contentType := getContentType(pvc)
+	// All archive requires scratch space.
+	if contentType == "archive" {
+		glog.V(3).Infof("Archive content type, requires scratch space\n")
+		return true
+	}
+	source := getSource(pvc)
+	// Invalid Sources will have gotten rejected by the web hook, so this is fine.
+	if source != SourceHTTP && source != SourceNone {
```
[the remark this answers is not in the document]
awels: As does http with specific archiving as well. Thing is, we don't know ahead of time, so we launch the importer, it sees it requires scratch space, and exits with the special exit code, which the controller sees, and it will relaunch the pod with scratch space. This check is simply to short circuit that loop for ones we know ahead of time will require scratch space.

pkg/controller/import-controller.go
```diff
@@ -72,6 +93,56 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+// volume claim. The order of preference is the following:
+// 1. Defined value in CDI config map.
+// 2. If 1 is not available use the 'default' storage class.
+// 3. If 2 is not available use the storage class name of the pvc that will own the scratch pvc.
+// 4. If none of those are available, fail with an error.
+func (ic *ImportController) getScratchPvcStorageClass(pvc *v1.PersistentVolumeClaim) (string, error) {
+	var storageClassName = ""
+	// TODO: Read config map to see if it has a storageClassName field in it.
+
+	if storageClassName == "" {
```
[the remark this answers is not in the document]
awels: I am not a fan of functions with multiple exit points. So this function is constructed that as soon as storageClassName has a value, all the subsequent if statements will fail and it will get to the exit point at the bottom of the function.

pkg/controller/import-controller.go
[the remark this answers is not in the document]
zvikorn: got you.

pkg/controller/import-controller.go
```diff
@@ -72,6 +93,56 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
 	return podList[0], nil
 }
 
+func (ic *ImportController) importerRequiresScratchSpace(pvc *v1.PersistentVolumeClaim) bool {
```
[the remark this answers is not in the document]
awels: I agree the name is a little too long and verbose (doesn't need importer, since we are in the importer controller), but I don't think that a 'requiresX' is less clear than shouldUse, requires is much more explicit about the actual hard need, vs shouldUse is more of a yeah we should but we can probably get away with not doing whatever it is.

pkg/controller/import-controller.go
```diff
@@ -72,6 +93,56 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+		return true
+	}
+	return false
+}
+
+// getScratchPvcStorageClass tries to determine which storage class to use for use with a scratch persistent
+// volume claim. The order of preference is the following:
+// 1. Defined value in CDI config map.
+// 2. If 1 is not available use the 'default' storage class.
+// 3. If 2 is not available use the storage class name of the pvc that will own the scratch pvc.
```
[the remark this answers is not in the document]
awels: k

pkg/importer/dataStream.go
```diff
@@ -324,121 +306,73 @@ func (d *DataStream) http() (io.ReadCloser, error) {
 	return countingReader, nil
 }
 
-//registryData - temporary data used by import from registry flow
-//is deleted as part of dataStream.Close()
-type registryData struct {
-	dataDir string
```
annastopel: This is temporary directory where registry flow stores its data, it should just point to tmpPvc location. This is a data context which cleanup is called on datastream.Close(). So all registry related temprary data is removed by the end of the flow.

pkg/importer/dataStream.go
```diff
@@ -324,121 +306,73 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
-	}
-	return nil
-}
-
-const (
-	//ContainerDiskImageDir - Expected disk image location in container image as described in
-	//https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md
-	ContainerDiskImageDir = "disk"
-	//TempContainerDiskDir - Temporary location for pulled containerImage
-	TempContainerDiskDir = "tmp"
```
annastopel: Why to remove? the only change here must be that TempContainerDikDur must be set to scratch space location

pkg/importer/dataStream.go
```diff
@@ -915,51 +845,38 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
+		if !hasScratch {
+			//Need scratch space but none provided.
+			return true, errors.New("Scratch space required, and none found")
+		}
+		// Replace /data/target name with scratch path/target name
+		tmpDest := filepath.Join(scratchDataDir, filepath.Base(dest))
+
+		err := StreamDataToFile(d.topReader(), tmpDest)
+		if err != nil {
+			return false, err
```
annastopel: In this case CleanDir is not called - this will leave files on /data. 
Why the existing solution in registry code has to be removed - it handles all theses cases. Please see above

pkg/controller/import-controller.go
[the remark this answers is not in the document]
awels: Good point, and I am doing something similar above, I should be consistent.

pkg/image/skopeo.go
```diff
@@ -80,35 +80,42 @@ func (o *skopeoOperations) CopyImage(url, dest, accessKey, secKey, certDir strin
 }
 
 // CopyRegistryImage download image from registry with skopeo
-func CopyRegistryImage(url, dest, destFile, accessKey, secKey, certDir string) error {
-	skopeoDest := "dir:" + dest + dataTmpDir
+// url: source registry url.
+// dest: the destination to write the unaltered disk image to
+// imageFile: the name of the disk image in the container image.
```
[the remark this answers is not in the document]
awels: Right I got that confused, I got confused about a few other things as well, re-looking at what I did here.

pkg/image/skopeo.go
```diff
@@ -80,35 +80,42 @@ func (o *skopeoOperations) CopyImage(url, dest, accessKey, secKey, certDir strin
...
+// tmpDest: The location to write the container image to before extracting the layers.
+// accessKey: accessKey for the registry described in url.
+// secKey: secretKey for the registry decribed in url.
+// certDir: directory public CA keys are stored for registry identity verification
+func CopyRegistryImage(url, dest, imageFile, tmpDest, accessKey, secKey string, certDir string) error {
+	skopeoDest := "dir:" + tmpDest + dataTmpDir
 	err := SkopeoInterface.CopyImage(url, skopeoDest, accessKey, secKey, certDir)
 	if err != nil {
 		os.RemoveAll(dest + dataTmpDir)
```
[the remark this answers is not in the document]
awels: I am reverting this change, I was trying to be slick by using both the temp storage and actual storage for storage during the extraction process. But that assumed the actual storage has a file system on it, which is not always the case.

pkg/importer/dataStream.go
```diff
@@ -324,121 +306,73 @@ func (d *DataStream) http() (io.ReadCloser, error) {
...
 	}
 
-	//cleanup in case of failure
-	cleanup := true
-	defer func() {
-		if cleanup {
-			tmpData.Cleanup()
-		}
-	}()
+	imageDir := filepath.Join(d.DataDir, ContainerDiskImageDir)
```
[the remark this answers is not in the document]
awels: Yes, so the thought was the following flow:
1. use skopeo to get the container image in /scratch.
2. Extract the image layers to /data (clean up /scratch).
3. Find the disk image and return the reader to it.
4. In data stream, write the contents of the reader back to /scratch (clean up /data)
5. Use the qemu convert to write the end result to /data

This way there is never a point in which I have the image data on either pvc twice. However this assumes that /data is available which is not always the case (if block pvc). I wanted to avoid doing everything in scratch because I will have 2 copies of the data at some point (container image AND extracted layers). But I don't think I can avoid that.

pkg/importer/dataStream.go
```diff
@@ -915,51 +845,38 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
+		tmpDest := filepath.Join(scratchDataDir, filepath.Base(dest))
+
+		err := StreamDataToFile(d.topReader(), tmpDest)
+		if err != nil {
+			return false, err
+		}
+
+		// Make sure the destination is empty
+		glog.V(3).Infoln("Cleaning up my mess.")
+		err = cleanDir(d.DataDir)
```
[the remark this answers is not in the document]
awels: The problem is you return the reader to data stream, and the reader is just an open file to the unconverted disk image, but the copy code tries to write that reader to the temporary space, but its already there with the same name so we can't do that.

I don't like the way its solved right now by some flags that say don't copy or don't move. That gives a very complex flow that makes it hard to extend in the future. It should be fairly simple:

if canStreamConvert {
  do streaming conversion
} else {
 write contents of reader to temporary location
  do non streaming conversion (although technically the difference between streaming and non streaming is the intermediate write to scratch, so we could write first, and then alter the url to a file:///path, and do the conversion from there)
}
resize if needed.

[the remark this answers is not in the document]
annastopel: What we have prior to scratch space enhancement is:
1. registry()
2. create directory /data/tmp
3. download container layers to /data/tmp/datatmp
4. extract only disk directory to /data/tmp/
5. obtain file reader on /data/tmp/imageifle
6. copy :
7. In case the file is qcow2 - do not copy to tmp - since it is already Stored -> just convert o raw and store converted under /data/disk.img
8. In case the file is raw - move it (this is an optimization inside the same fs, itr may fail if src and dst are not in the same fs -> failback to copy) from, /data/tmp/disk/imagefile to /data/disk.img
9. In all other cases (archived qcow2 etc.) - copy the file to /data/tmp<mangling>/ and then convert to raw and store in /data/disk.img
10. On ds.Close() call Clenaup of dataCtxt received from registry() flow - it will cleanuo the entire /data/tmp directory

So in fact, the entire /data/tmp dircetory and all that is there can be siwtched to be on scrtach space - this is a minor change of TempContainerDiskDir. 

I agree that when working with scratch space - there is no need in Moveable() is isStored() etc. If we can afford that we should keep it simple:
All tmp data on scratch space and only the resulting converted disk.img on /data/disk.img

awels:
ci test please

pkg/controller/import-controller.go
```diff
@@ -74,6 +95,56 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+func (ic *ImportController) importerRequiresScratchSpace(pvc *v1.PersistentVolumeClaim) bool {
+	contentType := getContentType(pvc)
+	// All archive requires scratch space.
+	if contentType == "archive" {
+		glog.V(3).Infof("Archive content type, requires scratch space\n")
+		return true
+	}
+	source := getSource(pvc)
+	// Invalid Sources will have gotten rejected by the web hook, so this is fine.
+	if source != SourceHTTP && source != SourceNone {
```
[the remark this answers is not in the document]
annastopel: So, what about the switch case? Without it - it is easy to avoid explicit consideration regarding need for scratch space in particular import and always ask for scratch PVC

pkg/util/util.go
```diff
@@ -129,3 +132,33 @@ func UnArchiveLocalTar(filePath, destDir string, arg ...string) error {
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
```
annastopel: Is it ok that Close() is called twice for out (return statement and defer)?

pkg/util/util.go
```diff
@@ -129,3 +132,33 @@ func UnArchiveLocalTar(filePath, destDir string, arg ...string) error {
...
+}
+
+// MoveFile moves a file by doing a copy and then removing the source file. This works across file system which os.Rename won't
+func MoveFile(src, dst string) error {
+	err := CopyFile(src, dst)
```
annastopel: For files that are located in the same file system we would like to have actual Rename and not copy - much faster.
So, it is worth executing rename and fallback to copy if fails

awels: This is a helper function explicitly for when the files are not on the same file system, I guess I should name the function more explicit that this is the case. We don't really want to encourage people calling this unless they know for a fact that os.Rename doesn't work.

pkg/util/util.go
awels: Yes you can call close multiple times on files. In general though the behaviour is undefined calling close twice on a Closer. For files it works fine.

pkg/importer/dataStream.go
```diff
@@ -485,24 +400,46 @@ func CopyData(dso *DataStreamOptions) error {
...
+	dso.RequiresScratch = requiresScratch
+	return err
 }
 
 // SaveStream reads from a stream and saves data to dest
-func SaveStream(stream io.ReadCloser, dest string) (int64, error) {
+func SaveStream(stream io.ReadCloser, dest string, diskImageFileName, dataPath, scratchPath string) (int64, error) {
 	glog.V(1).Infof("Saving stream to %q...\n", dest)
-	ds, err := newDataStreamFromStream(stream)
+	ds, err := newDataStream(&DataStreamOptions{
```
mhenriks: I know you moved this from elsewhere, but how about using property name constructor or whatever it's called

pkg/importer/dataStream.go
```diff
@@ -112,6 +108,12 @@ type DataStreamOptions struct {
 	CertDir string
 	// InsecureTLS is it okay to skip TLS verification
 	InsecureTLS bool
+	// HasScratchAvailable indicates if we have an attached scratch space PVC to write to.
+	HasScratchAvailable bool
+	// ScratchDataDir is the path to the scratch space inside the container.
+	ScratchDataDir string
```
mhenriks: How about getting rid of `HasScratchAvailable` and inferring with `ScratchDataDir != ""`?

pkg/importer/dataStream.go
```diff
@@ -485,24 +400,46 @@ func CopyData(dso *DataStreamOptions) error {
...
+		"stream://data",
+		"",
+		"",
+		controller.SourceHTTP,
+		string(cdiv1.DataVolumeKubeVirt),
+		"", // Blank means don't resize
+		util.GetAvailableSpace(dataPath),
+		"",
+		false,
+		util.GetAvailableSpace(scratchPath) != int64(-1),
```
mhenriks: should this be `> 0`?

pkg/importer/dataStream.go
```diff
@@ -918,51 +849,39 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
 }
 
 // Copy endpoint to dest based on passed-in reader.
-func (d *DataStream) copy(dest string) (err error) {
+// if the error is not nil and true is return, we require scratch space to complete the copy.
+func (d *DataStream) copy(dest string) (bool, error) {
+	if d.HasScratchAvailable {
+		defer CleanDir(d.ScratchDataDir)
```
mhenriks: If the scratch PVC has a distinct lifecycle from the pod, may make sense to preemptively clean up the scratch dir as well just in case.

cmd/cdi-importer/importer.go
```diff
@@ -98,6 +101,9 @@ func main() {
 		err = importer.CopyData(dso)
 		if err != nil {
 			glog.Errorf("%+v", err)
+			if dso.RequiresScratch {
```
mhenriks: Maybe have `CopyData` somehow return whether scratch is required?  Maybe it return `(bool, error)` or `err` is some sort of specific error type?

cmd/cdi-importer/importer.go
```diff
@@ -68,17 +68,20 @@ func main() {
...
+		Endpoint:            ep,
+		AccessKey:           acc,
+		SecKey:              sec,
+		Source:              source,
+		ContentType:         contentType,
+		ImageSize:           imageSize,
+		AvailableSpace:      util.GetAvailableSpace(common.ImporterVolumePath),
+		CertDir:             certDir,
 		InsecureTLS:    insecureTLS,
+		HasScratchAvailable: util.GetAvailableSpace(common.ScratchDataDir) != int64(-1),
```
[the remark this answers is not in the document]
awels: GetAvailableSpace will return -1 if there is an error statting the directory passed in. I suppose you could argue that if we have a 0 sized PVC this will return true for has scratch space available, and with it being 0 sized we don't have any. But I wrote this with the assumption, if the pvc is not mounted, I don't have scratch space, and thus statting the path will return -1, in which case this is correct.

cmd/cdi-importer/importer.go
awels: So I didn't want to get into parsing error messages, but I can just make a type and have it implement the error interface and then check the type instead of this clumsy setup, I will change.

cmd/cdi-importer/importer.go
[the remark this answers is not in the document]
mhenriks: `> 0` makes more sense to me

pkg/controller/controller.go
```diff
@@ -37,6 +39,7 @@ type Controller struct {
 	pullPolicy               string // Options: IfNotPresent, Always, or Never
 	verbose                  string // verbose levels: 1, 2, ...
 	podExpectations          *expectations.UIDTrackingControllerExpectations
+	pvcExpectations          *expectations.UIDTrackingControllerExpectations
```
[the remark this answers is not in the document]
awels: Good point, was trying to mimic the pod things we do.

cmd/cdi-importer/importer.go
mhenriks: Was thinking something like this:

```golang
var RequiresScratchSpaceError = fmt.Errorf("need scratch space")

// CopyData may return RequiresScratchSpaceError

// then check

if err == RequiresScratchSpaceError {

}

```

But if you want to to include more data in the error obviously a struct is the way to go

cmd/cdi-importer/importer.go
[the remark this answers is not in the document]
awels: I am not married to the != -1, the result is the same.

cmd/cdi-importer/importer.go
awels: Well not right now, but if we want more info in the future we can always switch to a struct then.

pkg/controller/import-controller.go
```diff
@@ -57,6 +60,25 @@ func NewImportController(client kubernetes.Interface,
 	return c
 }
 
+func (ic *ImportController) findScratchPvcFromCache(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
```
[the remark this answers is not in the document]
awels: I guess I can just do a .Get instead.

pkg/controller/import-controller.go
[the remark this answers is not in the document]
awels: With a switch we explicitly note which ones require scratch space vs which ones don't, so adding a new source type requires altering this switch if it requires scratch space vs. this which is the reverse, if the new one doesn't require scratch space we have to alter the if statement. Its a matter of preference I guess, and I have no objection to either, so I will put in the switch.

pkg/controller/import-controller.go
```diff
@@ -76,6 +98,29 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+	}
+	source := getSource(pvc)
+	// Invalid Sources will have gotten rejected by the web hook, so this is fine.
+	if source != SourceHTTP && source != SourceNone {
+		glog.V(3).Infof("Source type %s, requires scratch space\n", source)
+		return true
+	}
+
+	value, ok := pvc.Annotations[AnnRequiresScratch]
+	if ok && value == "true" {
```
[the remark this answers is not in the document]
awels: Sure

pkg/controller/import-controller.go
```diff
@@ -105,35 +150,36 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
...
-		return nil
+	if pod == nil && needsSync {
+		return ic.createImporterPod(pvc, pvcKey)
 	}
 
 	// update pvc with importer pod name and optional cdi label
 	anno := map[string]string{}
 	if pod != nil {
+		scratchExitCode := false
+		if pod.Status.Phase != v1.PodPending && pod.Status.ContainerStatuses != nil && pod.Status.ContainerStatuses[0].LastTerminationState.Terminated != nil &&
```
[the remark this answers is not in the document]
awels: Yes, that recently got added by you with PR #644 to avoid setting the data volume in failed while the pod was starting up.

pkg/controller/import-controller.go
```diff
@@ -143,6 +189,26 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
...
+
+		// Pod finished, if we used a scratch pvc remove it.
+		if pod.Status.Phase == v1.PodSucceeded {
+			scratchPvc, err := ic.findScratchPvcFromCache(pvc)
+			if err != nil {
+				glog.Errorf("error finding scratch pvc in cache: %v", err)
+				return err
+			}
+			if scratchPvc != nil {
+				// Scratch PVC found, lets clean it up.
```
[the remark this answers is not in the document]
awels: What happens in the following scenario?

1. PVC with annotation is created.
2. Pod is created
3. Pod figured out it needs scratch and kills itself.
4. scratch pvc is created
5. before next cycle to create new pod, original pvc is deleted.
6. new pod is never created and now I am stuck with the scratch pvc (this is the reason I made the original PVC the owner of the scratch pvc).

Maybe I should change the owner of the scratch pvc to the pod after the pod is created. That way if the original pvc is removed before the pod starts, the scratch is cleaned up due to the owner being original pvc, then once the pod is finished it cleans up the scratch automatically.

pkg/controller/import-controller.go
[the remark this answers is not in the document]
mhenriks: Yeah I remember that.   And it is needed for the condition below but  II susect this kind of makes in unnecessary: `pod.Status.ContainerStatuses[0].LastTerminationState.Terminated != nil `

pkg/controller/import-controller.go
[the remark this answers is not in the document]
mhenriks: Create the pod before the PVC.

pkg/controller/import-controller.go
```diff
@@ -76,6 +84,32 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+	case SourceRegistry:
+		return true
+	}
+
+	value, ok := pvc.Annotations[AnnRequiresScratch]
+	if ok {
+		boolVal, _ := strconv.ParseBool(value)
+		return boolVal
+	}
+	glog.V(3).Infof("No scratch space\n")
```
j-griffith: /nit
"no scratch space needed for import of PVC: %s", pvc.Name

j-griffith: /nit
I don't know how detailed we want to be in our logging, but if we want to follow the pattern you have for logging scratch not needed; might conisder collapsing all of the returns into a return value, then before leaving the routine have a single log statements for scratch or !scratch.  So something like:

```
scratchRequired := false
if contenteType == "archive"{
  scratchRquire = true
}
else {
  switch getSource(pvc){
    case SourceS3:
      scratchRequired = true
.......
........
glog.V(3).Infof("import pvc %s scratch requirement is %v", pvc.Name, scratchRequired)
return scratchRequired
```

I think the if blocks and cases could probably be optimized a bit but maybe not, that's not a big deal IMO

pkg/controller/upload-controller.go
```diff
@@ -436,6 +436,22 @@ func (c *UploadController) syncHandler(key string) error {
 		if err != nil {
 			return errors.Wrapf(err, "Error deleting upload pod for pvc: %s", key)
 		}
+		// TODO: This doesn't get called if the pod deletion fails. Should we try to make sure to delete the scratch pvc if the pod deletion fails?
+		scratchPvc, err := c.pvcLister.PersistentVolumeClaims(namespace).Get(name + "-scratch")
```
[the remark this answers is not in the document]
awels: Not needed anymore since owner is now pod.

pkg/controller/import-controller.go
awels: Well I am expecting some changes to this in the near future as narrow down what needs and what does not need scratch space. I will update the logging.

pkg/controller/import-controller.go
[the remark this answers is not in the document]
mhenriks: Still think we don't want this: `pod.Status.Phase != v1.PodPending`

pkg/controller/import-controller.go
```diff
@@ -105,35 +138,36 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
...
+	if pod == nil && needsSync {
+		return ic.createImporterPod(pvc, pvcKey)
 	}
 
 	// update pvc with importer pod name and optional cdi label
 	anno := map[string]string{}
 	if pod != nil {
+		scratchExitCode := false
+		if pod.Status.Phase != v1.PodPending && pod.Status.ContainerStatuses != nil && pod.Status.ContainerStatuses[0].LastTerminationState.Terminated != nil &&
+			pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode > 0 {
```
mhenriks: Did you look into using a pod termination message?  See here:  https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/

This way, we could pass obviously pass more information back to the controller.  Like, maybe (in the future) the pod could figure out exactly how big the scratch space needs to be.  Some sort of json termination message would be cool.

pkg/controller/import-controller.go
```diff
@@ -157,6 +191,49 @@ func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error
...
+			return err
+		}
+		if scratchPvc == nil {
+			storageClassName, err := GetScratchPvcStorageClass(ic.clientset, pvc)
+			if err != nil {
+				return err
+			}
+			// Scratch PVC doesn't exist yet, create it. Determine which storage class to use.
+			scratchPvc, err = CreateScratchPersistentVolumeClaim(ic.clientset, pvc, pod, storageClassName)
+			if err != nil {
```
mhenriks: I believe that if this fails, the pod will be pending forever.  Somewhere you you have to make sure that the scratch pvc exists so long as it is needed and the pod exists.

I initially thought you could simply delete the pod if creating the pvc fails.  That may be nice but is not sufficient.  The controller could die between operations.

pkg/controller/upload-controller.go
```diff
@@ -456,16 +455,42 @@ func (c *UploadController) getOrCreateUploadPod(pvc *v1.PersistentVolumeClaim, n
 	pod, err := c.podLister.Pods(pvc.Namespace).Get(name)
 
 	if k8serrors.IsNotFound(err) {
-		pod, err = CreateUploadPod(c.client, c.serverCAKeyPair, c.clientCAKeyPair.Cert, c.uploadServiceImage, c.verbose, c.pullPolicy, name, pvc)
+		pod, err = CreateUploadPod(c.client, c.serverCAKeyPair, c.clientCAKeyPair.Cert, c.uploadServiceImage, c.verbose, c.pullPolicy, name, pvc, pvc.Name+"-scratch")
+		_, err = c.createScratchPvc(pvc, pod)
```
mhenriks: similar to import controller, we have to handle pvc creation failure

pkg/controller/import-controller.go
awels: I did, but decided it was over engineering at this point in time. I purposely kept this part pretty simple so its easy to change out to something more complicated in the future.

pkg/importer/dataStream.go
awels: Done

pkg/importer/dataStream.go
awels: I am not sure I follow, we linked the pod and scratch pvc lifecycle at this point. This code is to clean up the scratch space after we are done with it.

pkg/controller/import-controller.go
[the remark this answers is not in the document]
awels: Actually the thing you added was somewhere else, so yeah don't think we need it.

pkg/controller/import-controller.go
awels: How about in the main loop, I check if pod exists and is in pending, and scratch PVC doesn't exist, I try to create it again there

awels:
ci test please

awels:
ci test please

pkg/controller/import-controller.go
```diff
@@ -76,6 +84,31 @@ func (ic *ImportController) findImportPodFromCache(pvc *v1.PersistentVolumeClaim
...
+func (ic *ImportController) requiresScratchSpace(pvc *v1.PersistentVolumeClaim) bool {
+	scratchRequired := false
+	contentType := getContentType(pvc)
+	// All archive requires scratch space.
+	if contentType == "archive" {
+		scratchRequired = true
+	} else {
+		switch getSource(pvc) {
+		case SourceS3:
+			scratchRequired = true
```
mhenriks: I don't think scratch is strictly required for S3

awels: No you are right should use the same rules as http, fixed.

awels:
> @awels Why to use Move method that calls copy inside and not just copy? Isn't scratch space cleaned anyway in CleanDir?

You are right, I could use copy instead of move, but the cleanup will have to happen some place, if its cleaned up during the process and we do another round of making sure at the end, the second round won't have to do anything.

> In fact (as mentioned before in code review) For raw images there is no need to copy to temp location and there is no need in scratch

At what point do we know an image is raw? After we have downloaded the container image to the scratch space, and after we have extracted the layers and identified the image. At that point we can determine if it is raw or not. But at that point it doesn't matter because the only operation left is to write/convert the image to the target pvc.

> 
> In addition, if both scratch and actual pvc have the same storage class, the actual Move will succeed, providing performance benefit

Actually that makes no difference, from the container perspective, the scratch pvc and target pvc are two different file systems, and thus we can't call os.Rename, and thus we have to copy.

zvikorn:
We can check if it's raw when we call copy , and when we check d.isHTTPQcow2 we can check in addition if it's raw

awels:
I am interested to know how we know the image is raw, before we have even identified the name of the image file.

zvikorn:
qemu-img info?

zvikorn:
Something like:
if !isRaw(image){
// Replace /data/target name with scratch path/target name
		tmpDest := filepath.Join(d.ScratchDataDir, filepath.Base(dest))
}

func isRaw(image){
qemu-img info (url)
if format == raw return true

}

awels:
We can run qemu-img info after we have identified the name of image from the container image. In order to find the file name we have to first download the container image using skopeo, and then extract the image layers and find the file. Then we can do qemu-img info, but at that point all that is left to do is either do a qemu-img convert or a straight copy. So the info if a file is raw doesn't matter. If we call qemu-img convert with a raw, its just a straight copy by qemu-img.

zvikorn:
convertQcow2ToRawStream checks info on url. why cant we do the same for raw?>

awels:
The '-f qcow2' in the parameter list in ConvertQcow2ToRaw doesn't actually mean much (https://linux.die.net/man/1/qemu-img) qemu-img detects the input format and does a conversion based on that.

annastopel:
But we are not talking only about import from registry, right? The same
code is  executed also when importing raw image from http target.

awels:
But if we are importing a raw image from an http source, it should set the isHTTPQcow2 flag to true when it is inspecting the headers, and it should stream directly to the target PVC. The name is misleading at this point, it should really be called 'canStream' or something of that nature.

zvikorn:
So it’s not :-(

awels:
Okay, that is not good, it should, but right now doesn't matter. I am working on the refactor of data stream and I will address that during that time. For now it can be a little less efficient.

zvikorn:
For the block PV, I will have to set tmpDest=dest (or something) as I do not want  to create /scratch/blockDevice for iso http

awels:
Why would you create /scratch/blockDevice? Your block device should be /blockdevice or something like that. It has no relationship with the scratch space.

zvikorn:
I would not. This is what happens currently. It goes in the 'scratch path' by default. So I end up having the /scratch/blockDevice.

awels:
I don't follow what currently happens with an iso (or anything that we think needs scratch space)
1. tmpDest is set to /scratch/disk.img (dest should be /data/disk.img, it takes the last part of that). For block PVs dest is something else, doesn't really matter what it is called, it will take the last part and create the image file that name in /scratch, so if dest is /blockdevice it will create /scratch/blockdevice (the file) which is image from the source. If we want to be clearer, might want to not re-use dest with a block device, or check dataDir existing to change the name of the file we create in /scratch
2. The data is streamed to tmpDest from the source reader (if its http, it comes from a URL, if its registry it comes from a file, namely /scratch/disk.img, so nothing happens)
3. If d.qemu is true we convert from from /scratch/disk.img to dest (either /data/disk.img) or whatever the block device is. If it is false we do a standard copy (which should only happen in the case of archives)

awels:
Actually I have an idea to make things less confusing, since it doesn't matter what the name of the file is in /scratch, lets just call it tmpfile So change the line

```go
tmpDest := filepath.Join(d.ScratchDataDir, filepath.Base(dest))
```
to
```go
tmpDest := filepath.Join(d.ScratchDataDir, "tmpfile")
```
Then you can leave the rest alone and it is clear that what is in scratch is just a temporary intermediate file before we convert/copy to the target

zvikorn:
Sent from my iPhone

> On 10 Mar 2019, at 21:21, Alexander Wels <notifications@github.com> wrote:
> 
> I don't follow what currently happens with an iso (or anything that we think needs scratch space)
> 
> tmpDest is set to /scratch/disk.img (dest should be /data/disk.img, it takes the last part of that). For block PVs dest is something else, doesn't really matter what it is called, it will take the last part and create the image file that name in /scratch, so if dest is /blockdevice it will create /scratch/blockdevice (the file) which is image from the source. If we want to be clearer, might want to not re-use dest with a block device, or check dataDir existing to change the name of the file we create in /scratch
Why do I need to change /blockDevice to /scratch/blockDevice if I do no need temporary space in case of iso?

awels:
But you do need temporary space for an ISO, otherwise d.isHTTPQcow2() would be true and it would call that path.

awels:
If you are 100% sure you can stream directly to the target when you have an iso, then set ds.qemu to true on the same line you set ds.isIsoImage to true.

annastopel:
ISO is raw, right? So for iso isHTTPQCow2() is false, right?

awels:
Apparently ISOs are raw yes, didn't know that, so yeah you should be able to stream it directly to the target. So for the http source ISO you can set ds.qemu=true, which in turn will make d.isHTTPQcow2() true, which then falls into the streaming path.

pkg/importer/dataStream.go
```diff
@@ -970,49 +895,36 @@ func (d *DataStream) convertQcow2ToRaw(src, dest string) error {
...
-				err = StreamDataToFile(d.topReader(), tmpDest)
-			}
-			if err != nil {
-				return errors.WithMessage(err, fmt.Sprintf("unable to stream data to file %q", dest))
-			}
-		} else {
-			//It is a stored qemu file - taking a location where it is stored
-			klog.V(3).Infof("Streaming file to %s", tmpDest)
-			tmpDest = d.tmpDataPath
+		if util.GetAvailableSpace(d.ScratchDataDir) <= int64(0) {
```
annastopel: @awels  may be during refactoring it is worth considering to place the logic that decides whether scratch space is required or not with import method logic, otherwise we have  several places that check and verify weather scratch space is required and is provided, which is confusing IMO.

For example it is enough to decide per import method that Scratch is required and then use it inside that method (http/s3/registry etc.) Scratch can be assigned to dataStreamer (http/s3/registry) and in code that is common to all streamers (as copy) the scratch can be taken from concrete streamer if it is required. In such case all the logic is in one place per importing technique (see https://github.com/kubevirt/containerized-data-importer/pull/633 to see what I mean)

PR #681, opened 2019-02-28
Remove duplicate API validation tests

igoihman:

I found a few duplicate tests, same logic exists somewhere else in the file.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
So the point of those manifest tests was to have the test run by using actual manifests instead of code generated manifests. The actual logic tested might be the same, but the source of the test is something an actual user would do vs our generated manifests.

igoihman:
Yeah I get you point but it's almost the same. I'm converting yaml to JSON, changing the fields and then converting back to yaml and saving the file.
The base is an actual manifest.
But if you decide to keep it we can close this PR.

PR #684, opened 2019-03-02
tech debt: make cleanup

annastopel:
- remove stale make targets
       - add make target for cleanup docker cache on clsuter-down cluster-down-purge
       - minor fixes
       - update hack/build/README.md

Fixes #

Content:
       - remove stale make targets
       - add make target for cleanup docker cache on clsuter-down cluster-down-purge
       - minor fixes
       - update hack/build/README.md

**Release note**:

```release-note
None
```


annastopel:
ci test please

igoihman:
@annastopel These are great changes! Can you split it to a few commits? one for every change? It will make the review much more easy.

hack/build/cleanup_docker.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+REPO=
```
awels: You can replace the whole setTag function with this:

REPO=${DOCKER_REPO:-"localhost"}

That basically means REPO = ${DOCKER_REPO} if not found, use "localhost"

hack/build/cleanup_docker.sh
```diff
@@ -0,0 +1,34 @@
...
+
+function setTag {
+  if [ "$1" == "" ]; then
+      REPO="localhost"
+  else
+      REPO=$1
+  fi
+}
+
+function dockerCleanup {
```
awels: What does this do that "docker system prune -a" doesn't do?

annastopel:
Prune will remove everything including k8s images, which can be reused in
opposite to cdi images that are built every cluster-sync and are set in
repo associated with generated local registry port of this specific cluster
instance

hack/build/cleanup_docker.sh
annastopel: > What does this do that "docker system prune -a" doesn't do?

Prune will remove everything including k8s images, which can be reused in
opposite to cdi images that are built every cluster-sync and are set in
repo associated with generated local registry port of this specific cluster
instance

hack/build/cleanup_docker.sh
annastopel: > You can replace the whole setTag function with this:
> 
> REPO=${DOCKER_REPO:-"localhost"}
> 
> That basically means REPO = ${DOCKER_REPO} if not found, use "localhost"

Nice! Updated

hack/build/cleanup_docker.sh
mhenriks: this is a very cool feature.  But I think it should only prune images for the current (soon to be exiting) instance of the of the docker registry.  Can call gocli to get the registry port.  See here:  https://github.com/kubevirt/containerized-data-importer/blob/master/cluster/sync.sh#L16-L17

annastopel: > this is a very cool feature. But I think it should only prune images for the current (soon to be exiting) instance of the of the docker registry. Can call gocli to get the registry port. See here: https://github.com/kubevirt/containerized-data-importer/blob/master/cluster/sync.sh#L16-L17

Done. Please review

PR #686, opened 2019-03-04
https cert configuration

mhenriks:

We added `storage.import.certConfigMap` for registry import, so why not use it for https import as well

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/clean.sh
+++ b/cluster/clean.sh
@@ -19,28 +19,25 @@ if [ -f "${OPERATOR_MANIFEST}" ]; then
     ./cluster/kubectl.sh delete --ignore-not-found -f "${OPERATOR_MANIFEST}"
 fi
 
-# Work around https://github.com/kubernetes/kubernetes/issues/33517
-./cluster/kubectl.sh delete ds -l 'operator.cdi.kubevirt.io' -l "cdi.kubevirt.io" -n ${NAMESPACE} --cascade=false --grace-period 0 2>/dev/null || :
-
-# Delete all traces of kubevirt
-namespaces=(default kube-system ${NAMESPACE})
+# Everything should be deleted by now, but just to be sure
+namespaces=(default kube-system)
 for i in ${namespaces[@]}; do
     ./cluster/kubectl.sh -n ${i} delete deployment -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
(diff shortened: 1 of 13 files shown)


cluster/clean.sh
```diff
@@ -19,29 +19,26 @@ if [ -f "${OPERATOR_MANIFEST}" ]; then
     ./cluster/kubectl.sh delete --ignore-not-found -f "${OPERATOR_MANIFEST}"
 fi
 
-# Work around https://github.com/kubernetes/kubernetes/issues/33517
-./cluster/kubectl.sh delete ds -l 'operator.cdi.kubevirt.io' -l "cdi.kubevirt.io" -n ${NAMESPACE} --cascade=false --grace-period 0 2>/dev/null || :
-
-# Delete all traces of kubevirt
-namespaces=(default kube-system ${NAMESPACE})
+# Ecerything should be deleted by now, but just to be sure
```
awels: Everything

tests/utils/configmaps.go
```diff
@@ -1,50 +1,66 @@
...
-func CopyConfigMap(client kubernetes.Interface, srcNamespace, srcName, destNamespace, destName string) error {
+func CopyConfigMap(client kubernetes.Interface, srcNamespace, srcName, destNamespace, destName string) (string, error) {
 	src, err := client.CoreV1().ConfigMaps(srcNamespace).Get(srcName, metav1.GetOptions{})
 	if err != nil {
-		return err
+		return "", err
+	}
+
+	if destName == "" {
+		destName = srcName + "-" + strings.ToLower(util.RandAlphaNum(8))
```
awels: Maybe I am missing something, why are we creating a config map with a partially random name? Its getting copied to a new namespace which already has a randon name, so what is the reason to have the same thing for the config map?

mhenriks: Not strictly necessary given that tests are running serially.  But I like the idea of each test having its own distinct data.  Which will be necessary if we ever run tests in parallel (which would be nice)

awels: But each test creates its own namespace, which will be unique, even the ones in a 'table'

mhenriks: Okay, I was confused.  I thought that a single namespace was created for all of the transport tests.  Did not see that the framework adds a before/after each calback that creates/deletes the namespace.

I still think this randomization is a nice to have.

awels: It doesn't bother me that much, just wanted to point it out.

mhenriks:
ci test please

PR #700, opened 2019-03-10
Use CDIConfig for scratch storage class.

awels:

Allow admins to specify which storage class to use for scratch space using the CDIConfig object.

Fixes #

**Release note**:

```release-note
Scratch storage class can be specified using the CDIConfig CRD
```

### DIFF
--- a/doc/scratch-space.md
+++ b/doc/scratch-space.md
@@ -1,0 +1,20 @@
+# CDI Scratch space
+Containerized Data Importer(CDI) requires scratch space for certain operations to complete, this temporary space needs to be obtained from somewhere. Kubernetes has some options available to get temporary space like emptyDir volumes, however that space is shared among pods and it is uncertain how much space is available or what the node behavior will be if CDI fills up that space with a large image. For this and other reasons CDI will create scratch space from available PVs using a storage class. This scratch space will then be used to process the data before writing it to the target PVC.
+
+CDI uses the following mechanism to determine which storage class to use:
+
+1. Read the CDI config field _scratchSpaceStorageClass_ if that field exists, and the value matches one of the storage classes in the cluster, it will be used to create scratch space.
+2. If 1 is empty or didn't match a storage class, look up the _default_ storage class in the cluster, if that exists use that storage class.
+3. If there is no default storage class in the cluster, then use the storage class of the PersistentVolumeClaim(PVC) that is backing the DataVolume(DV) that started the CDI operation.
+
+If none of those exist, then CDI will be unable to create scratch space. This means that none of the operations that require scratch space will work, however operations that do not require scratch space will continue to operate normally.
+
+Operations that require scratch space are:
+
+| Type | Reason|
+|------|-------|
+| Registry imports | In order to import from registry container images, CDI has to first download the image to a scratch space, extract the layers to find the image file, and then pass that image file to QEMU-IMG for conversion to a raw disk |
+| Upload image | Because QEMU-IMG does not accept inputs from stdin yet, we cannot stream the upload directly to QEMU-IMG, so we have to save the upload to a scratch space first and then pass it to QEMU-IMG for conversion |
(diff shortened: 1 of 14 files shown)


doc/scratch-space.md
```diff
@@ -0,0 +1,10 @@
...
+
+CDI uses the following mechanism to determine which storage class to use:
+
+1. Read the CDI config field _scratchSpaceStorageClass_ if that field exists, and the value matches one of the storage classes in the cluster, it will be used to create scratch space.
+2. If 1 is empty or didn't match a storage class, look up the _default_ storage class in the cluster, if that exists use that storage class.
+3. If there is no default storage class in the cluster, then use the storage class of the PersistentVolumeClaim(PVC) that is backing the DataVolume(DV) that started the CDI operation.
+
+If none of those exist, then CDI will be unable to create scratch space. This means that none of the operations that require scratch space will work, however operations that do not require scratch space will continue to operate normally.
```
annastopel: Can we specify explicitly which operations require scratch and which are not?

awels: We can, that will probably help @aglitke with his task to document all the valid permutations of formats we support. Let me update the PR to include a grid with operations that require scratch space.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -256,7 +256,8 @@ type CDIConfig struct {
 
 //CDIConfigSpec defines specification for user configuration
 type CDIConfigSpec struct {
-	UploadProxyURLOverride *string `json:"uploadProxyURLOverride,omitempty"`
+	UploadProxyURLOverride   *string `json:"uploadProxyURLOverride,omitempty"`
+	ScratchSpaceStorageClass *string `json:"scratchSpaceStorageClass,omitempty"`
```
mhenriks: I think this should be handled similarly to the upload proxy URL.  There should be a value in the status struct that has whatever the effective value of the scratch space storage class.  It will be either the user provided value in the spec struct or a computed value if the user provided value does not exist.  Clients simply use whatever value is in the status.  Basically move all the logic of determining the scratch space storage class to the config controller.

awels: Good point, I will make that change.

doc/scratch-space.md
awels: Updated.

pkg/apis/core/v1alpha1/types.go
awels: Updated.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -256,12 +256,14 @@ type CDIConfig struct {
...
 }
 
 //CDIConfigStatus provides
 type CDIConfigStatus struct {
-	UploadProxyURL *string `json:"uploadProxyURL,omitempty"`
+	UploadProxyURL           *string `json:"uploadProxyURL,omitempty"`
+	ScratchSpaceStorageClass *string `json:"scratchSpaceStorageClass,omitempty"`
```
mhenriks: I think shis should be `string` instead of `*string`.  There should always be a value.  '""' is a fine value.

pkg/controller/config-controller.go
```diff
@@ -267,27 +268,65 @@ func (c *ConfigController) syncHandler(key string) error {
...
+		if url == "" {
+			newConfig.Status.UploadProxyURL = nil
+		} else {
+			newConfig.Status.UploadProxyURL = &url
+		}
+	}
+
+	storageClass, err := c.scratchSpaceStorageClassStatus(config)
+
+	if storageClass == config.Status.ScratchSpaceStorageClass {
```
mhenriks: comparing pointer values here not string values so will never be equal.

Recommend `config.Status.ScratchSpaceStorageClass` be `string` and `scratchSpaceStorageClassStatus` returns `(string, error)`

pkg/apis/core/v1alpha1/types.go
awels: Done

pkg/controller/config-controller.go
awels: Done

PR #706, opened 2019-03-13
Add type metadata to cdi-insecure registries ConfigMap

booxter:
Without TypeMeta, controller manifest generated by manifest-generator
lacks Kind attribute for the ConfigMap, which makes `oc` fail as
follows:

error validating "cdi-controller.yaml": error validating data:
[apiVersion not set, kind not set]; if you choose to ignore these
errors, turn validation off with --validate=false

```release-note
NONE
```


igoihman:
ci test please

PR #707, opened 2019-03-13
Fix wrong value being passed to validation.

awels:

This PR fixes an issue where validating how much space was available the wrong value was being passed to the validation, and thus it would fail in several instances. In particular it would fail for upload validation checking.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/dataStream.go
+++ b/pkg/importer/dataStream.go
@@ -104,7 +104,7 @@ type DataStreamOptions struct {
 	// ImageSize is the size we want the resulting image to be.
 	ImageSize string
 	// Available space is the available space before downloading the image
-	AvailableSpace int64
+	AvailableDestSpace int64
 	// CertDir is a directory containing tls certs
 	CertDir string
 	// InsecureTLS is it okay to skip TLS verification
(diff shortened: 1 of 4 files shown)


awels:
ci test please

pkg/importer/dataStream.go
```diff
@@ -853,7 +853,7 @@ func (d *DataStream) isHTTPQcow2() bool {
 }
 
 func (d *DataStream) calculateTargetSize(dest string) int64 {
-	targetQuantity := resource.NewScaledQuantity(util.GetAvailableSpace(filepath.Dir(dest)), 0)
+	targetQuantity := resource.NewScaledQuantity(d.AvailableSpace, 0)
```
mhenriks: Maybe this is something I missed in the scratch space PR.  But why is available space being cached?  My concerns are:
1.  One more thing to set when initializing a data stream
2.  If we forget to do above, the value will be zero, so nothing will work
3.  The name is ambiguous.  Free space where?

awels: The variable has been there for a while, its the space available in the target pvc. We added this variable when we didn't have scratch space and we calculated the availabe space after we wrote the file there already but before the resize, so we would not resize to the correct size. This is why we set it before doing anything (cached)

mhenriks: Maybe rename the variable in your datastream cleanup?

awels: I can rename it in this patch, as well, not a big deal not many places use it.

awels:
@mhenriks I updated the name of the variable to reflects its purpose better.

PR #713, opened 2019-03-18
dump logs when functional tests fail

mhenriks:
**What this PR does / why we need it**:

Looks like all the registry related functional tests have been failing for awhile.  Maybe this long overdue change will help us figure out what's going on.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #714, opened 2019-03-19
bug fix: func test - registry buildah failure on outdated /etc/contai…

annastopel:
…ners/storage.conf

Fixes #

in func-test-registry-host pod, registry-populate container was failing to build and push images.
it utilizes buildah tool for that purpose.  The error was 
```bash
 'overlay' is not supported over xfs at "/var/lib/containers/storage/overlay" 
kernel does not support overlay fs: 'overlay' is not supported over xfs at "/var/lib/containers/storage/overlay": backing file system is unsupported for this graph driver
```

Solution:
The /etc/container/storage.conf file is outdated
removing it solves the issue for buildah tool

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/build/docker/cdi-func-test-registry-populate/populate-registry.sh
+++ b/hack/build/docker/cdi-func-test-registry-populate/populate-registry.sh
@@ -110,6 +110,8 @@ function pushImages {
    done
 }
 
+#remove storage.conf if exists 
+rm -rf /etc/containers/storage.conf
 
 #start health beat
 health $HEALTH_PATH $HEALTH_PERIOD &


awels:
I did a little research on this, and in some cases an old conf file would mess with things, something about the old configuration using camel case, if that is all it takes, I am all for it.

mhenriks:
Can't believe builds are taking over three hours now.  @awels any idea what may be going on there?

awels:
Looking at the log timestamps, it spends 30 minutes running actual tests, and 1:30 setting up the environment, and the rest is waiting for a runner. https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_containerized-data-importer_standard-check-pr/204/execution/node/182/log/?consoleFull

Why it is taking 1:30 to setup the environment I do not know.

awels:
Aww man, the k8s one failed because the environment never came up.
ci test please

awels:
ci test please

mhenriks:
https://github.com/kubevirt/containerized-data-importer/pull/713 has been going for 5+ hours

mhenriks:
Anyway back to this PR:  I'm not crazy about the fix since I don't think we know exactly why it's failing but since this is only for functional tests I think I'd okay with whatever works for now.  PRs are starting to back up.

awels:
I am fine with it as well, of course CI is just bonkers slow right now. But I am willing to merge.

annastopel:
> Anyway back to this PR: I'm not crazy about the fix since I don't think we know exactly why it's failing but since this is only for functional tests I think I'd okay with whatever works for now. PRs are starting to back up.

@mhenriks according to issue https://github.com/containers/buildah/issues/1332  when working on fedora28 and buildah v1.7 local configuration that exist on fedora28 is not good anymore. 
Removing this outdated storage.conf  resolves the issue, since there is a global configuration that is being used then.

I know this does not pinpoint the issue of what exactly not good in fedora28 storage.conf file (apprently overlay driver configuration, if I switch driver to vfs and comment out other few configurations the issue is also resolved, but if there is a global config that works, why not to use it), but as you mentioned it is a func test host, so I did the first thing that works

Thank You

PR #717, opened 2019-03-20
registry func test host fixes

annastopel:
1. remove unecessary pvc and volume mount
2. bug fix in populate-registry error handling

Fixes #

FIxes:
1. Remove unnecessary pvc in registry pod configuration
2. fix error handling in populate_registry.sh script 
**Release note**:

```release-note
None
```

### DIFF
--- a/hack/build/docker/cdi-func-test-registry-populate/populate-registry.sh
+++ b/hack/build/docker/cdi-func-test-registry-populate/populate-registry.sh
@@ -93,7 +93,6 @@ function pushImages {
    registry_tls=$4
    registry=$registry_host":"$registry_port
    
-   retval=$?
    shopt -s nullglob
    for IMAGEDIR in *$DIR; do
         cd $IMAGEDIR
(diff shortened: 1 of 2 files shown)


hack/build/docker/cdi-func-test-registry-populate/populate-registry.sh
```diff
@@ -102,10 +101,10 @@ function pushImages {
         IMAGENAME=${FILE//.}
         echo "building image "$IMAGENAME
         buildah bud -t $IMAGENAME":latest" $images"/"$IMAGEDIR"/"
-	error $retval	
+	error $?
```
mhenriks: Curious why using error function instead of `set -e`?

awels: Does it matter that much? I am perfectly happy either way

mhenriks: Well, the script does a bunch of other stuff with no error checking.  And I think it would be preferable for the container to exit with an error than fail silently (that way functional tests will never run, right?).

Obviously, this is an improvement, and that's great.  And if you want to merge it now, that is a-ok with me.

awels: Okay, @annastopel could you do an additional PR with set -e?

PR #721, opened 2019-03-21
data stream refactor.

awels:

This PR removes the dataStream struct and associated file and replaces it with a data processer and a number of data sources. This makes for a much cleaner data pipe line that is easier to understand and maintain. It is also easier to write unit tests for each individual files.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/importer/data_processor.go
```diff
@@ -0,0 +1,211 @@
...
+	var err error
+	defer dp.provider.Close()
+	if util.GetAvailableSpace(dp.scratchDataDir) > int64(0) {
+		defer CleanDir(dp.scratchDataDir)
+	}
+	for dp.currentPhase != Complete {
+		switch dp.currentPhase {
+		case Info:
+			dp.currentPhase, err = dp.provider.Info()
+		case Transfer:
```
annastopel: Suggest to unite Transfer and TransferTarget into Transfer and pass both dataDir and SratchDataDir to each data_source file, so each specific dataSource will decide how and when to use dataDir or ScratchDataDir. 
In additiona, each data source can cleanup after itself on provider.Close(), so no need to call CleanDir() when it is not used

pkg/importer/http_data_source.go
```diff
@@ -0,0 +1,271 @@
...
+	for {
+		if count < reader.Current {
+			// Some progress was made, reset now.
+			lastUpdate = time.Now()
+			count = reader.Current
+		}
+
+		if time.Until(lastUpdate.Add(idleTime)).Nanoseconds() < 0 {
+			// No progress for the idle time, cancel http client.
+			dp.cancel() // This will trigger dp.ctx.Done()
```
annastopel: What happens we reach this line and dp.cancel() is called? Shall dp.cancel be set to nil in order  for dp.cancel() not to be called second time on provider.Close()?

Can here be a race between Polling thread (that reaches this line) and main thread that runs provider.Close()?

pkg/importer/http_data_source.go
```diff
@@ -0,0 +1,271 @@
...
+
+// GetURL returns the URI that the data processor can use when converting the data.
+func (dp *HTTPDataSource) GetURL() *url.URL {
+	return dp.url
+}
+
+// Close all readers.
+func (dp *HTTPDataSource) Close() error {
+	var err error
+	if dp.readers != nil {
```
annastopel: Need to close dp.httpReader as well, right?

pkg/importer/s3_data_source.go
```diff
@@ -0,0 +1,130 @@
...
+// GetURL returns the url that the data processor can use when converting the data.
+func (sd *S3DataSource) GetURL() *url.URL {
+	return sd.url
+}
+
+// Close closes any readers or other open resources.
+func (sd *S3DataSource) Close() error {
+	var err error
+	if sd.readers != nil {
+		err = sd.readers.Close()
```
annastopel: Shall close sd.s3Reader as well, right?

pkg/importer/s3_data_source.go
```diff
@@ -0,0 +1,130 @@
...
+		return Error, ErrRequiresScratchSpace
+	}
+	file := filepath.Join(path, tempFile)
+	err := StreamDataToFile(sd.readers.TopReader(), file)
+	if err != nil {
+		return Error, err
+	}
+	// If streaming succeeded, then parsing the file into URL will also succeed, no need to check error status
+	sd.url, _ = url.Parse(file)
+	return Process, nil
```
annastopel: Same comment as for http_data_source, shall we support tar.gz files?

pkg/importer/http_data_source.go
```diff
@@ -0,0 +1,271 @@
...
+		file := filepath.Join(path, tempFile)
+		err := StreamDataToFile(dp.readers.TopReader(), file)
+		if err != nil {
+			return Error, err
+		}
+		// If we successfully wrote to the file, then the parse will succeed.
+		dp.url, _ = url.Parse(file)
+		return Process, nil
+	} else if dp.contentType == cdiv1.DataVolumeArchive {
+		if err := util.UnArchiveTar(dp.readers.TopReader(), path); err != nil {
```
annastopel: Sorry, for keep asking the same question - don't we want to support tar.gz as well?

pkg/importer/data_processor.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// DataProcessor holds the fields needed to process data from a data provider.
+type DataProcessor struct {
+	// currentPhase is the phase the processing is in currently.
+	currentPhase ProcessingPhase
+	// provider provides the data for processing.
+	provider DataProviderInterface
+	// destination file. will be DataDir/disk.img if file system, or a block device (if a block device, then DataDir will not exist).
+	dest string
```
annastopel: may be dataFile instead dest?

cmd/cdi-importer/importer.go
```diff
@@ -104,7 +88,28 @@ func main() {
...
+			if err != nil {
+				klog.Errorf("%+v", err)
+				os.Exit(1)
+			}
+		default:
+			klog.Errorf("Unknown source type %s\n", source)
+			os.Exit(1)
+		}
+		processor := importer.NewDataProcessor(dp, dest, dataDir, common.ScratchDataDir, imageSize)
+		err = processor.ProcessData()
```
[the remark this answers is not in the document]
awels: Okay I did miss a few things here, one I forgot to rename the interface to be DataSourceInterface, I will change that. I don't feel that strongly about calling close from inside the processor vs outside. But I see the argument of whomever created the resource, should close the resource. I will change.

pkg/importer/data_processor.go
awels: That would be consistent with the terminology we are using elsewhere, agreed will change.

pkg/importer/data_processor.go
awels: I disagree, there is no need for 2 parameters since we are either transferring to the scratch space, or to the target space but not both. So if I pass 2 parameters one will always be empty. I also don't really want the data source to worry too much about where to write the data. The data source job is to collect the data, and then write it to where the process wants it to write.

pkg/importer/data_processor.go
```diff
@@ -0,0 +1,211 @@
...
+			klog.Errorf("%v", err)
+			return err
+		}
+		klog.V(1).Infof("New phase: %s\n", dp.currentPhase)
+	}
+	return err
+}
+
+// convert is called when convert the image from the url to a RAW disk image. Source formats include RAW/QCOW2 (Raw to raw conversion is a copy)
+func (dp *DataProcessor) convert(url *url.URL) (ProcessingPhase, error) {
```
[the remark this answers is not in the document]
awels: I believe we have had this discussion before. Yes technically raw doesn't need to be converted, because its raw already. But the added complexity in code to detect raw and then go into a special code path just for that is simply not worth it. This keeps the logic of the code much simpler. Unless there are significant benefits to a special code path I don't want to make the code more complex.

pkg/importer/http_data_source.go
```diff
@@ -0,0 +1,271 @@
...
+			return Error, ErrRequiresScratchSpace
+		}
+		file := filepath.Join(path, tempFile)
+		err := StreamDataToFile(dp.readers.TopReader(), file)
+		if err != nil {
+			return Error, err
+		}
+		// If we successfully wrote to the file, then the parse will succeed.
+		dp.url, _ = url.Parse(file)
+		return Process, nil
```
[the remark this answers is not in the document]
awels: No that should work, the format readers are a stack each reader will extract one format and pass the result to the next reader. so for imagefile.qcow2.tar.gz it will create a tar and a gz format reader, which will result in the qcow2 file being written to scratch space, then coverted to raw during the convert phase.

pkg/importer/http_data_source.go
awels: We didn't before, we only supported just tar files in this mode, we could potentially want to support tar.gz or tar.xz as well, but that would be an enhancement.

pkg/importer/http_data_source.go
awels: No because the httpReader is part of the reader stack

```go
dp.readers, err = NewFormatReaders(dp.httpReader, dp.contentType)
```
Its the bottom of the stack after that call.

pkg/importer/data_processor.go
annastopel: 1. in all data sources scratch space is either required or not, but all of them eventually will copy the data to /data/disk.img
2. In fact in all those data sources where scratch is required there are validations that scratch space is provided, and if not provided then a dedicated error ErrRequiresScratchSpace is returned. So in fact data source is aware of which space it needs and when it uses it. It might be more consistent to validate scratch space in one place (either data_processor or specific data_source. I personally vote for data_source) By the way, scratch space path can be provided on data_source creation and not on Transfer, in such case it is passed only to those data_sources that can potentially require it.

pkg/importer/http_data_source.go
awels: I think you are right, there is a potential race between the two. I also should set dp.cancel to nil so the other can't be called and vice versa, when I call the other one I should set it to nil there as well. I will add a lock to this code so there won't be a problem.

pkg/importer/registry_data_source.go
```diff
@@ -0,0 +1,146 @@
...
+func (rd *RegistryDataSource) Process() (ProcessingPhase, error) {
+	imageFile, err := getImageFileName(rd.imageDir)
+	if err != nil {
+		return Error, errors.Wrapf(err, "Cannot locate image file")
+	}
+
+	// imageFile and rd.imageDir are both valid, thus the Join will be valid, and the parse will work, no need to check for parse errors
+	rd.url, _ = url.Parse(filepath.Join(rd.imageDir, imageFile))
+	klog.V(3).Infof("Successfully found file. VM disk image filename is %s", rd.url.String())
+	return Convert, nil
```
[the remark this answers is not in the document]
awels: According https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md, which is the format we support, the only format stored in a registry disk should be RAW (otherwise kubevirt can't boot it). So I specifically didn't run the registry import through the format readers. Technically we should fail the convert if we find a qcow2 disk image as that also doesn't conform to the format outlined in that document. But then we get back to the should we add more complex code (to detect qcow2 and reject it specifically) vs simpler code.

pkg/importer/s3_data_source.go
awels: Same answer, the format reader stack takes care of that.

pkg/importer/s3_data_source.go
awels: Same answer, the s3reader is part of the stack (the bottom in fact) and thus shall be closed as part of the readers.close

annastopel: You are right, did not notice before that the s3 reader is added to readers in the first line of constructReaders

pkg/importer/s3_data_source.go
annastopel: You mean we do not intend to support tar.gz file from s3 source as well?

pkg/importer/http_data_source.go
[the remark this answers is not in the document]
annastopel: Do we have a unit test that checks with case?

pkg/importer/data_processor.go
awels: 1. Not true, in particular archive content type will write to the /data directory one or more files of whatever name is in the archive. Also in the happy paths scratch space is not required, in fact the data source doesn't even write any data anywhere, it all happens in the conversion phase.
2. You are right the validation for scratch space is knowledge the data source shouldn't have, it should check if the path provided is valid or not, and return an error based on that, then the data processor should determine based on what it provided the transfer function, if it should return the scratch space error. I will fix that, I took a short cur when I shouldn't have.

As for setting the data/scratch space when creating the data source, yes I can do that, but why would I want to. I want the data source to concentrate on its purpose, getting the data. The data processor decides where that data goes.

pkg/importer/s3_data_source.go
awels: No I mean since we are using the format readers for s3 as well, that will be supported through them.

pkg/importer/data_processor.go
```diff
@@ -0,0 +1,217 @@
...
+	// Calculate available space before doing anything.
+	dp.availableSpace = dp.calculateTargetSize()
+	return dp
+}
+
+// ProcessData is the main processing loop.
+func (dp *DataProcessor) ProcessData() error {
+	var err error
+	if util.GetAvailableSpace(dp.scratchDataDir) > int64(0) {
+		defer CleanDir(dp.scratchDataDir)
```
mhenriks: Great that we are cleaning up, but why not do a preemptive clean as well?  An earlier one could have failed.

pkg/importer/upload_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+// Info is called to get initial information about the data.
+func (ud *UploadDataSource) Info() (ProcessingPhase, error) {
+	var err error
+	// Hardcoded to only accept kubevirt content type.
+	ud.readers, err = NewFormatReaders(ud.stream, cdiv1.DataVolumeKubeVirt)
+	if err != nil {
+		klog.Errorf("Error creating readers: %v", err)
+		return Error, err
+	}
+	return Transfer, nil
```
mhenriks: Can this function return `TransferToTarget` if we have detected that the user is not uploading a `qcow2` file?  This could potentially speed things up greatly.  Remember that there may be a real human being at the other side of this upload.  And if that human happens to be one of our qa people in they have the slowest storage in the world.

pkg/importer/http_data_source.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	// The readers now contain all the information needed to determine if we can stream directly or if we need scratch space to download
+	// the file to, before converting.
+	if dp.readers.Qemu && !dp.readers.Archived && !dp.customCA {
+		// We can pass straight to conversion from the endpoint. No scratch required.
+		dp.url = dp.endpoint
+		return Convert, nil
+	}
+	if dp.contentType == cdiv1.DataVolumeArchive {
+		return TransferTarget, nil
```
mhenriks: Can we also return `TransferTarget` for raw image?

mhenriks:
Another general comment - `_` vs `-` in filenames.  I prefer dashes.  And that's what we do for controllers.  We should be consistent.

awels:
No problem, I can change the file names

pkg/importer/http_data_source.go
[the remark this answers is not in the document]
awels: That is not a unit test instance, we have a functional test that checks though.

pkg/importer/data_processor.go
awels: Right that is sort of the intention, even though we don't have any of that right now. But a good example of something we could potentially do is. Lets say we always give each pod emptyDir volume (host space). And we determine we need scratch space, if the info phase decided that the size of the incoming data is small enough, the data processor might decided to go directly to transfer with the emptyDir instead of killing the pod with the scratch space exit code.

This is just some stuff I am thinking of of the top of my head, but I think there is value in having the items separated between the processor and the data source.

pkg/importer/data_processor.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+var qemuOperations = image.NewQEMUOperations()
+
+// ProcessingPhase is the current phase being processed.
+type ProcessingPhase string
+
+const (
+	// Info is the first phase, durlng this phase the provider obtains information needed to determine which phase to go to next.
+	Info ProcessingPhase = "Info"
```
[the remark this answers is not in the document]
awels: Yeah, they are a little generic, I can rename

pkg/importer/data_processor.go
awels: Sure

pkg/importer/http_data_source.go
awels: Done

pkg/importer/upload_data_source.go
awels: Done

pkg/importer/data_processor.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (dp *DataProcessor) calculateTargetSize() int64 {
+	var targetQuantity *resource.Quantity
+	if util.GetAvailableSpace(dp.dataDir) > int64(0) {
+		// File system volume.
+		targetQuantity = resource.NewScaledQuantity(util.GetAvailableSpace(dp.dataDir), 0)
+	} else {
+		// Block volume.
+		targetQuantity = resource.NewScaledQuantity(util.GetBlockFileSize(dp.dataFile), 0)
```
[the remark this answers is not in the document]
awels: Did an explicit check now.

PR #723, opened 2019-03-25
Orchestrate smart cloning with DataVolumes

danielerez:
This PR is introducing support for smart-clone PVC using DataVolumes.
As a CNV user I want to clone volumes as efficiently as possible and I want the system to automatically select the best method available based on my request.

Currently CSI direct cloning is not implemented by kubernetes so we will need to use a snapshots based approach.  This will require orchestration which could probably be handled by the CDI DataVolume controller.  A possible flow could be:

- DataVolume is created with a pvc source (clone request)
- Check if smart cloning is possible
  - The source and target pvcs must be in the same namespace
  - The source and target pvcs must be in the same storage class
  - There must be a snapshotclass associated with the storage class
- If smart cloning is enabled:
  - Create a snapshot of the source pvc
  - Create a pvc from the created snapshot
  - Delete the snapshot
- If smart cloning is not enabled:
  - Trigger a host-assisted clone

* Code Changes:
  * Added volumesnapshot api [1] and generated clientset/informers/listers using k8s code-generator [2].
  
  * datavolume-controller:
    - When clone pvc is requested and smart-clone is applicable -> Create a new VolumeSnapshot.

  * Introduced smart-clone-controller:
    - Listens to VolumeSnapshot changes and create a new PVC based on the snapshot.
    - Upon PVC successful creation, deletes the snapshot for cleaning up.

Requirements:
* k8s >= 1.13 with VolumeSnapshotDataSource feature-gate enabled (for snapshots support)
* VolumeSnapshotClass (associated with StorageClass)
* Storage driver the supports snapshots should be installed.

[1] https://github.com/kubernetes-csi/external-snapshotter/tree/master/pkg/apis/volumesnapshot/v1alpha1

[2] https://github.com/kubevirt/containerized-data-importer/blob/master/vendor/k8s.io/code-generator/generate-groups.sh

Fixes #

**Release note**:
```release-note
Data Volumes (DV) can be created also by specifying a PVC as an input source. It will trigger a clone of the original PVC.

CDI implementation of cloning a PVC is done with host assisted cloning by streaming the data from the source PVC and write to the new PVC.

In order to improve the performance of the cloning process, a Smart-Cloning flow using snapshots is introduced.

CDI use the feature of creating a PVC from snapshot in order to clone PVCs more efficiently when a CSI plugin with snapshot capabilities is available.

The yaml structure and annotations of the DV are not changed.
```


pkg/apis/volumesnapshot/v1alpha1/register.go
```diff
@@ -0,0 +1,58 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name use in this package.
+const GroupName = "snapshot.storage.k8s.io"
```
awels: You just defined this const in the volumesnapshot package, can you use that one?

pkg/client/clientset/versioned/clientset.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The CDI Authors.
+Copyright The Kubernetes Authors.
```
awels: I am fairly certain this should remain the CDI authors, not the kubernetes authors.

danielerez: Yeah, I wondered how I could prevent the k8s code-generator to change it. Any idea how a custom copyright can be configured?

pkg/apis/volumesnapshot/v1alpha1/register.go
danielerez: Sure. Missed it as we simply copied the file from: https://github.com/kubernetes-csi/external-snapshotter/tree/master/pkg/apis/volumesnapshot/v1alpha1

pkg/client/clientset/versioned/clientset.go
danielerez: ok, used 'make generate' instead (which using the 'custom-boilerplate.go.txt')

pkg/apis/upload/v1alpha1/openapi_generated.go
```diff
@@ -23,149 +23,9 @@ limitations under the License.
 package v1alpha1
 
 import (
-	spec "github.com/go-openapi/spec"
 	common "k8s.io/kube-openapi/pkg/common"
 )
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
-	return map[string]common.OpenAPIDefinition{
```
mhenriks: Appears that you did not use `make generate' to do code generation.

pkg/apis/volumesnapshot/register.go
```diff
@@ -0,0 +1,6 @@
+package volumesnapshot
+
+const (
+	// GroupName to hold the string name for the cdi uploads
+	GroupName = "snapshot.storage.k8s.io"
```
mhenriks: Why are we not just vendoring the API in from https://github.com/kubernetes-csi/external-snapshotter?

pkg/operator/resources/cluster/factory.go
```diff
@@ -39,6 +39,9 @@ func createCRDResources(args *FactoryArgs) []runtime.Object {
 	return []runtime.Object{
 		createDataVolumeCRD(),
 		createCDIConfigCRD(),
+		createVolumeSnapshotClassCRD(),
```
mhenriks: Do we really want to be creating these CRDs if they don't already exist?  I don't think so.  We can detect if the VolumeSnapshot api does not exist and always do host assisted cloning.

rollandf: What about the case that CDI is installed before the CSI driver?

danielerez: The smart-cloner-controller is registered in the cmd/controller. I.e. we listen to the snapshot informer events after deployment. So if we don't create the CRDs, it means that we must rely on them being already installed before deployment. Thus, if not installed, we need to stop the smart-cloner-controller? So we'll have to detect an installation of the CRDs afterwards to run the controller again?

pkg/apis/volumesnapshot/register.go
danielerez: We can try that, but just thought that these types should be in the client-go later on any way. So instead of relying on another dependency, we can just copy them for now and remove once added to client-go. What do you think?

pkg/operator/resources/cluster/factory.go
awels: Can't we have the operator (or deployment) controller listen for CRDs being created, and if it finds the right CRDs to create smart cloner controller, or if the CRDs are removed it removes the smart cloner controller.

pkg/apis/volumesnapshot/register.go
mhenriks: Much cleaner to vendor this stuff in.  Also be easier to get rid of later.

pkg/operator/resources/cluster/factory.go
mhenriks: Yeah, I think the operator should watch CRDs.  If VolumeSnapshot exists either a dedicated deployment for the smart clone controller will be started or the current controller deployment can be started/restarted with a different set of args

pkg/apis/upload/v1alpha1/openapi_generated.go
aglitke: Can we add a Jenkins validation that can detect this automatically?

mhenriks: @aglitke we already have that.  This build didn't get that far.  Travis build appears to have got wedged sometime before that.

awels: Travis is currently having issues starting linux VMs, if you look at their status page you will see they are investigating. Also STDCI is not working right, I have engaged the maintainers to see what is going on.

pkg/apis/volumesnapshot/register.go
rollandf: Done

glide.yaml
```diff
@@ -38,3 +38,5 @@ import:
   version: release-3.11
 - package: github.com/openshift/client-go
   version: release-3.11
+- package: github.com/kubernetes-csi/external-snapshotter
+  version: ^v1.0.1
```
mhenriks: May be wise to lock this at a specific version (no `^`)

Also no newline at the end of the file

hack/build/config.sh
```diff
@@ -49,7 +49,7 @@ VERBOSITY=${VERBOSITY:-1}
 PULL_POLICY=${PULL_POLICY:-IfNotPresent}
 NAMESPACE=${NAMESPACE:-cdi}
 
-KUBERNETES_IMAGE="k8s-1.11.0@sha256:e02ec414c7673a3644b5ba742b550a124c9195eaacb280151406bf9e8201a95f"
+KUBERNETES_IMAGE="k8s-multus-1.13.3"
```
mhenriks: We need multus?  Or 1.13?  Or both?

awels: The only available 1.13 image is the multus one. Not sure why.

mhenriks: I assume that was the case before https://github.com/kubevirt/containerized-data-importer/pull/724?

awels: Yeah, you are right, they added a non multus one in kubevirtci at this point, so once rebased on master that should be correct now.

pkg/operator/resources/cluster/factory.go
danielerez: Sounds like a major change in the operator behaviour? Can we do it in a consecutive PR later on and just manually create the CRDs for now?

glide.yaml
rollandf: Done

hack/build/config.sh
rollandf: Done.

rollandf:
ci test please

pkg/controller/datavolume-controller.go
```diff
@@ -460,11 +603,13 @@ func (c *DataVolumeController) updateUploadStatusPhase(pvc *corev1.PersistentVol
 }
 
 func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) error {
+	klog.V(3).Infof("!!! DataVolumeController updateDataVolumeStatus dataVolume name: %s", dataVolume.Name)
 	dataVolumeCopy := dataVolume.DeepCopy()
 	var err error
 	var event DataVolumeEvent
 
 	curPhase := dataVolumeCopy.Status.Phase
+
```
igoihman: redundant

pkg/controller/datavolume-controller.go
```diff
@@ -460,11 +603,13 @@ func (c *DataVolumeController) updateUploadStatusPhase(pvc *corev1.PersistentVol
 }
 
 func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) error {
+	klog.V(3).Infof("!!! DataVolumeController updateDataVolumeStatus dataVolume name: %s", dataVolume.Name)
```
igoihman: redundant

pkg/controller/datavolume-controller.go
```diff
@@ -478,6 +623,11 @@ func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolu
 		}
 
 	} else {
+
+		klog.V(3).Infof("!!! DataVolumeController updateDataVolumeStatus pvc pvc.Status.Phase: %s", pvc.Status.Phase)
+		klog.V(3).Infof("!!! DataVolumeController updateDataVolumeStatus pvc name: %s", pvc.Name)
+		klog.V(3).Infof("!!! DataVolumeController updateDataVolumeStatus curPhase: %s", curPhase)
+
```
igoihman: redundant

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+	snapshotsinformers "kubevirt.io/containerized-data-importer/pkg/snapshot-client/informers/externalversions/volumesnapshot/v1alpha1"
+	snapshotslisters "kubevirt.io/containerized-data-importer/pkg/snapshot-client/listers/volumesnapshot/v1alpha1"
+)
+
+const (
+	//AnnSmartCloneRequest sets our expected annotation for a CloneRequest
+	AnnSmartCloneRequest = "k8s.io/SmartCloneRequest"
+)
+
+// SmartCloneController represents the CDI Clone Controller
```
igoihman: `CDI SmartClone Controller`

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+	snapshotInformer snapshotsinformers.VolumeSnapshotInformer,
+	dataVolumeInformer informers.DataVolumeInformer,
+	image string,
+	pullPolicy string,
+	verbose string) *SmartCloneController {
+	c := &SmartCloneController{
+		Controller:       *NewController(client, pvcInformer, podInformer, image, pullPolicy, verbose),
+		cdiClientSet:     cdiClientSet,
+		csiClientSet:     csiClientSet,
+		snapshotsLister:  snapshotInformer.Lister(),
```
igoihman: missing: `snapshotsSynced: snapshotInformer.Informer().HasSynced,`

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+
+// SmartCloneController represents the CDI Clone Controller
+type SmartCloneController struct {
+	Controller
+	cdiClientSet clientset.Interface
+	csiClientSet csiclientset.Interface
+
+	snapshotsLister  snapshotslisters.VolumeSnapshotLister
+	dataVolumeLister listers.DataVolumeLister
+	snapshotsSynced  cache.InformerSynced
```
igoihman: missing: dataVolumeSynced

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+	defer c.queue.Done(key)
+
+	ns, name, err := cache.SplitMetaNamespaceKey(key.(string))
+	if err != nil {
+		runtime.HandleError(errors.Errorf("invalid resource key: %s", key))
+		return false
+	}
+
+	pvc, err := c.pvcLister.PersistentVolumeClaims(ns).Get(name)
+	snapshot, err := c.snapshotsLister.VolumeSnapshots(ns).Get(name)
```
igoihman: you forgot to handle the errors

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+
+	if pvc != nil && pvc.Status.Phase == corev1.ClaimBound {
+		klog.V(3).Infof("!!! smart-clone ProcessNextItem pvc.Name: %s", pvc.Name)
+		klog.V(3).Infof("!!! smart-clone ProcessNextItem pvc.ObjectMeta.Annotations: %s", pvc.ObjectMeta.Annotations[AnnSmartCloneRequest])
+		if pvc.ObjectMeta.Annotations[AnnSmartCloneRequest] == "true" {
+			snapshotName := pvc.Spec.DataSource.Name
+			snapshotToDelete, err := c.snapshotsLister.VolumeSnapshots(ns).Get(snapshotName)
```
igoihman: error handling

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+		//set Datavolume phase to Scheduled
+		return nil
+	}
+
+	return c.processSnapshotItem(snapshot)
+}
+
+// Create the cloning source and target pods based the pvc. The pvc is checked (again) to ensure that we are not already
+// processing this pvc, which would result in multiple pods for the same pvc.
+func (c *SmartCloneController) processSnapshotItem(snapshot *csisnapshotv1.VolumeSnapshot) error {
```
igoihman: `processNextSnapshotItem`

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,308 @@
...
+		runtime.HandleError(err)
+		return
+	}
+	c.queue.AddRateLimited(key)
+}
+
+func newPvcFromSnapshot(snapshot *csisnapshotv1.VolumeSnapshot, dataVolume *cdiv1.DataVolume) *corev1.PersistentVolumeClaim {
+	labels := map[string]string{
+		"cdi-controller": snapshot.Name,
+		"app":            "containerized-data-importer",
```
igoihman: ```common.CDILabelKey:       common.CDILabelValue ```

pkg/controller/smart-clone-controller.go
rollandf: Done

pkg/controller/smart-clone-controller.go
rollandf: Done

pkg/controller/smart-clone-controller.go
rollandf: Done

pkg/controller/smart-clone-controller.go
rollandf: Done

pkg/controller/controller.go
```diff
@@ -31,6 +31,7 @@ type Controller struct {
 	clientset                kubernetes.Interface
 	queue                    workqueue.RateLimitingInterface
 	pvcInformer, podInformer cache.SharedIndexInformer
+	snapshotInformer         cache.SharedIndexInformer
```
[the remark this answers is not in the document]
rollandf: Done, moved to smartclone-controller

pkg/controller/datavolume-controller.go
```diff
@@ -297,7 +311,7 @@ func (c *DataVolumeController) processNextWorkItem() bool {
 // converge the two. It then updates the Status block of the DataVolume resource
 // with the current status of the resource.
 func (c *DataVolumeController) syncHandler(key string) error {
-
+	klog.V(3).Infof("!!! DataVolumeController syncHandler key: %s", key)
```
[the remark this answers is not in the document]
rollandf: Done

pkg/controller/datavolume-controller.go
```diff
@@ -342,17 +356,34 @@ func (c *DataVolumeController) syncHandler(key string) error {
 		return errors.Errorf(msg)
 	}
 
+	// expectations prevent us from creating multiple pods. An expectation forces
+	// us to observe a pod's creation in the cache.
 	needsSync := c.pvcExpectations.SatisfiedExpectations(key)
+	klog.V(3).Infof("!!! DataVolumeController needsSync: %t", needsSync)
```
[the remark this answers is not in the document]
rollandf: Done

pkg/controller/datavolume-controller.go
```diff
@@ -342,17 +356,34 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
+	klog.V(3).Infof("!!! DataVolumeController exists: %t", exists)
```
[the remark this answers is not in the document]
rollandf: Done

pkg/controller/datavolume-controller.go
```diff
@@ -342,17 +356,34 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
-		if err != nil {
-			return err
-		}
-		c.pvcExpectations.ExpectCreations(key, 1)
-		pvc, err = c.kubeclientset.CoreV1().PersistentVolumeClaims(dataVolume.Namespace).Create(newPvc)
-		if err != nil {
-			c.pvcExpectations.CreationObserved(key)
-			return err
+		smartCloneApplicable, snapshotClassName := c.isSmartCloneApplicable(dataVolume)
+		klog.V(3).Infof("!!! smartCloneApplicable: %t", smartCloneApplicable)
```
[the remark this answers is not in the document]
rollandf: Done

pkg/controller/datavolume-controller.go
```diff
@@ -367,6 +398,118 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
+	pvc, err := c.pvcLister.PersistentVolumeClaims(dataVolume.Spec.Source.PVC.Namespace).Get(dataVolume.Spec.Source.PVC.Name)
+	if err != nil {
+		runtime.HandleError(err)
+		return false, nil
+	}
+	if pvc == nil {
+		klog.V(3).Infof("Source PVC is missing: %s/%s", dataVolume.Spec.Source.PVC.Namespace, dataVolume.Spec.Source.PVC.Name)
+		return false, nil
+	}
+	klog.V(3).Infof("!!! pvc.Namespace %s", pvc.Namespace)
```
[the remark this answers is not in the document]
rollandf: Done

pkg/controller/datavolume-controller.go
```diff
@@ -367,6 +398,118 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
+			return false, nil
+		}
+		for _, storageClass := range storageclasses.Items {
+			if storageClass.Annotations["storageclass.kubernetes.io/is-default-class"] == "true" {
+				targetPvcStorageClassName = &storageClass.Name
+				break
+			}
+		}
+	}
+	klog.V(3).Infof("!!! (source) pvc.Spec.StorageClassName %s", *pvc.Spec.StorageClassName)
```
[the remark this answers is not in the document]
rollandf: Done

pkg/controller/datavolume-controller.go
rollandf: Done

pkg/controller/datavolume-controller.go
rollandf: Done

tests/smartclone_test.go
```diff
@@ -0,0 +1,97 @@
...
+				dataVolume = utils.NewCloningDataVolume(dataVolumeName, "1Gi", sourcePvc)
+				dataVolume.Spec.PVC.StorageClassName = &rbdStorageClassName
+			}
+
+			By(fmt.Sprintf("creating new datavolume %s", dataVolume.Name))
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("waiting for datavolume to match phase %s", string(phase)))
+			utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, phase, dataVolume.Name)
```
awels: You didn't assign err, so the following check is against the err above.

tests/smartclone_test.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests", func() {
+
+	var sourcePvc *v1.PersistentVolumeClaim
+
+	fillData := "123456789012345678901234567890123456789012345678901234567890"
+	testFile := utils.DefaultPvcMountPath + "/source.txt"
+	fillCommand := "echo \"" + fillData + "\" >> " + testFile
+	rbdStorageClassName := "csi-rbd"
```
awels: We cannot assume that a particular storage class name exists in the environment the test is running in, yes in OUR CI it is true, but someone else can try to run the test in their environment that has a different name for the storage class.

The usual pattern to solve a problem like this is pass the storage class name that has a particular property into the tests as an argument, store it in the framework struct, then use it during the tests. There are several examples of this in framework.go, in particular look at the init function.

I would call it something like 'snapshotSCName' in the framework, that way we know that argument to the test provides a storage class that can do snapshots.

tests/smartclone_test.go
rollandf: Removed the Expect.
We have the same issue here:
https://github.com/kubevirt/containerized-data-importer/blob/master/tests/datavolume_test.go#L91

pkg/controller/datavolume-controller.go
```diff
@@ -225,7 +238,7 @@ func (c *DataVolumeController) Run(threadiness int, stopCh <-chan struct{}) erro
 
 	// Wait for the caches to be synced before starting workers
 	klog.V(2).Info("Waiting for informer caches to sync")
-	if ok := cache.WaitForCacheSync(stopCh, c.pvcsSynced, c.dataVolumesSynced); !ok {
+	if ok := cache.WaitForCacheSync(stopCh, c.pvcsSynced, c.dataVolumesSynced, c.snapshotClassSynced); !ok {
```
awels: Are we guaranteed to always have the snapshotClass available, so the sync can happen. We had an issue in the config-controller, where it was waiting for Routes in a k8s environment, and that would cause the controller to not initialize. I can see the same thing happening here in environments that don't have snapshots available. In the config-controller I solved it by checking for openshift and passing a slice to the WaitForCacheSync that had the RouteSync only if we are in openshift. Not sure if we can detect if the snapshots are available here.

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,329 @@
...
+		UpdateFunc: func(old, new interface{}) {
+			c.enqueuePvc(new)
+		},
+		DeleteFunc: c.enqueuePvc,
+	})
+
+	return c
+}
+
+func (c *SmartCloneController) initializeExpectations(pvcKey string) error {
```
awels: Do we need these two? I don't think pods are created in this controller.

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,329 @@
...
+	// anno := map[string]string{}
+
+	snapshotKey, err := cache.MetaNamespaceKeyFunc(snapshot)
+	if err != nil {
+		return err
+	}
+
+	// expectations prevent us from creating multiple pods. An expectation forces
+	// us to observe a pod's creation in the cache.
+	needsSync := c.podExpectations.SatisfiedExpectations(snapshotKey)
```
awels: Oh I see, we need them, they are just named podExpectations, instead of snapshotExpectations.

pkg/controller/datavolume-controller.go
j-griffith: No, we are not guaranteed to have the snapshotClass... snapshot is a feature gate, and it's  an optional capability depending on the CSI plugin, and it's an external service.  You must try and reach the service and act accordingly based on the response.

tests/smartclone_test.go
rollandf: Done

pkg/controller/smart-clone-controller.go
rollandf: Done, renamed.

pkg/controller/smart-clone-controller.go
rollandf: Renamed

pkg/controller/datavolume-controller.go
rollandf: What do you by "have the snapshotClass available"?
We create the CRDs in createVolumeSnapshotClassCRD, though Michael commented that it is maybe not the right way.

rollandf:
ci test please

pkg/controller/datavolume-controller.go
j-griffith: The Snapshotter endpoint may not even be on the cluster (Snapshot capability in the first place) and the plugin may or may not have snapshot capability at all.

pkg/controller/datavolume-controller.go
```diff
@@ -406,6 +433,123 @@ func (c *DataVolumeController) scheduleProgressUpdate(dataVolumeName, dataVolume
...
+		return ""
+	}
+
+	targetPvcStorageClassName := dataVolume.Spec.PVC.StorageClassName
+
+	// Handle unspecified storage class name, fallback to default storage class
+	if targetPvcStorageClassName == nil {
+		storageclasses, err := c.kubeclientset.StorageV1().StorageClasses().List(metav1.ListOptions{})
+		if err != nil {
+			runtime.HandleError(err)
```
zvikorn: So if we do not have any storage class at all, we should fall back to host assisted cloning. right?

tests/smartclone_test.go
```diff
@@ -0,0 +1,96 @@
...
+			Eventually(func() bool {
+				events, err := RunKubectlCommand(f, "get", "events", "-n", dataVolume.Namespace)
+				if err == nil {
+					fmt.Fprintf(GinkgoWriter, "%s", events)
+					return strings.Contains(events, eventReason)
+				}
+				fmt.Fprintf(GinkgoWriter, "ERROR: %s\n", err.Error())
+				return false
+			}, timeout, pollingInterval).Should(BeTrue())
+
```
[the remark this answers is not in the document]
rollandf: Done

zvikorn:
can you guys add .md file under doc folder?

pkg/controller/datavolume-controller.go
awels: My main concern for this comment is, will WaitForCacheSync block forever waiting for a cache sync on a CRD that doesn't exist. I wasn't even worried about what John is talking about.

pkg/controller/datavolume-controller.go
```diff
@@ -352,20 +364,35 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
-			c.pvcExpectations.CreationObserved(key)
-			return err
-		}
-		if canUpdateProgress(newPvc.Annotations) {
-			go c.scheduleProgressUpdate(dataVolume.Name, dataVolume.Namespace, pvc.GetUID())
+		snapshotClassName := c.getSnapshotClassForSmartClone(dataVolume)
+		if snapshotClassName != "" {
+			klog.V(3).Infof("Smart-Clone via Snapshot is available with Volume Snapshot Class: %s", snapshotClassName)
+			newSnapshot := newSnapshot(dataVolume, snapshotClassName)
+			c.pvcExpectations.ExpectCreations(key, 1)
```
awels: So I don't think you need the expectations at all, let me explain why. Expectations are like a lock counter, they are needed to account for the request for creation of pods and them actually being created. But they are only needed if the pods names are not unique, like in the case of an importer or cloner pod. However the snapshot name is derived from the data volume name, which by itself is guaranteed to be unique, and thus the snapshot is guaranteed to be unique, so we don't need the expectations to count if the snapshots are being created. I had the same thing when I worked on the scratch space and @mhenriks rightfully pointed out I didn't need them. I think the same is true for snapshots.

tests/smartclone_test.go
```diff
@@ -0,0 +1,113 @@
...
+		if sourcePvc != nil {
+			By("[AfterEach] Clean up source PVC")
+			err := f.DeletePVC(sourcePvc)
+			Expect(err).ToNot(HaveOccurred())
+			sourcePvc = nil
+		}
+	})
+
+	Describe("Verify DataVolume Smart Cloning", func() {
+		FIt("succeed creating smart-clone dv", func() {
```
awels: Think you forgot to remove "F" from it, so now its just focused on the one test.

pkg/controller/datavolume-controller.go
danielerez: We are currently creating the CRDs in 'factory.go -> createCRDResources', so it should guarantee we won't block on WaitForCacheSync. 
Regarding the snapshotClass availability, it should be validated in 'datavolume-controller -> getSnapshotClassForSmartClone' func. Do you think it's good enough for now?

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,325 @@
...
+		return nil
+	}
+
+	return c.processNextSnapshotItem(snapshot)
+}
+
+// Create the cloning source and target pods based the pvc. The pvc is checked (again) to ensure that we are not already
+// processing this pvc, which would result in multiple pods for the same pvc.
+func (c *SmartCloneController) processNextSnapshotItem(snapshot *csisnapshotv1.VolumeSnapshot) error {
+	// anno := map[string]string{}
```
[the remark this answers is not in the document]
danielerez: sure. done.

tests/smartclone_test.go
danielerez: oops. done. :)

pkg/controller/datavolume-controller.go
danielerez: ok, got it. removed.

rollandf:
> can you guys add .md file under doc folder?
Done

cmd/cdi-controller/controller.go
```diff
@@ -113,7 +115,14 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
 		klog.Fatalf("Error building example clientset: %s", err.Error())
```
awels: This is not your issue, but its funny no one noticed the 'example' part of the message.

pkg/controller/datavolume-controller.go
```diff
@@ -352,20 +374,37 @@ func (c *DataVolumeController) syncHandler(key string) error {
 		return errors.Errorf(msg)
 	}
 
+	// expectations prevent us from creating multiple pods. An expectation forces
```
awels: Don't need this comment anymore I think, since not using expectations.

pkg/controller/datavolume-controller_test.go
```diff
@@ -88,6 +93,10 @@ func newImportDataVolume(name string) *cdiv1.DataVolume {
 }
 
 func newCloneDataVolume(name string) *cdiv1.DataVolume {
+	return newCloneDataVolumeWithPVCNS(name, "default")
+}
+
+func newCloneDataVolumeWithPVCNS(name string, pvcNamespace string) *cdiv1.DataVolume {
```
awels: I would call the pvcNamespace just namespace to be consistent with everywhere else. it is implied that it is used with a PVC.

rollandf: Done

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,325 @@
...
+	if err != nil {
+		return err
+	}
+
+	newPvc := newPvcFromSnapshot(snapshot, dataVolume)
+	if newPvc == nil {
+		klog.Errorf("error creating new pvc from snapshot object")
+		return nil
+	}
+	//set Datavolume phase to InProgress
```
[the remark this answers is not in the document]
rollandf: Removed

pkg/controller/datavolume-controller.go
mhenriks: I still don't think we should create the CRDs.  And the controller (or operator) should handle the situation where they do not exist.  Maybe put the CRDs in a separate manifest that can be installed by users/CI but I feel pretty strongly that the operator should not install them.

tests/smartclone_test.go
```diff
@@ -0,0 +1,139 @@
...
+			By("[AfterEach] Clean up source PVC")
+			err := f.DeletePVC(sourcePvc)
+			Expect(err).ToNot(HaveOccurred())
+			sourcePvc = nil
+		}
+	})
+
+	Describe("Verify DataVolume Smart Cloning - Positive flow", func() {
+		It("succeed creating smart-clone dv", func() {
+			if controller.IsOpenshift(f.K8sClient) {
```
awels: So I know I told you guys to use IsOpenshift to skip the test, but I think we can make it more robust by checking if the snapshot storage class we passed in as a parameter exists. That way if the tests are run on a k8s cluster that doesn't have a storage class with csi snapshotting enabled, this will get skipped, and if at some point openshift gets the support needed, we won't have to change this code.

The check can be as simple as taking the class name passed in, and doing a get on the storage class

```go
func (f *Framework) IsSnapshotStorageClassAvailable() bool {
  sc, err := f.K8sClient.StorageV1().StorageClasses().Get(f.snapshotSCName, metav1.GetOptions{})
  if err != nil {
    return false
  }
  return sc.Name == f.snapshotSCName
}

File: pkg/controller/datavolume-controller.go [hunk 40]
```diff
@@ -406,6 +441,129 @@ func (c *DataVolumeController) scheduleProgressUpdate(dataVolumeName, dataVolume
...
+
+func newSnapshot(dataVolume *cdiv1.DataVolume, snapshotClassName string) *csisnapshotv1.VolumeSnapshot {
+	annotations := make(map[string]string)
+	annotations[AnnSmartCloneRequest] = "true"
+	className := snapshotClassName
+	snapshot := &csisnapshotv1.VolumeSnapshot{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        dataVolume.Name,
+			Namespace:   dataVolume.Namespace,
+			Annotations: annotations,
```
mhenriks: some labels here, like `common.CDIComponentLabel`

pkg/controller/datavolume-controller.go
```diff
@@ -352,20 +370,37 @@ func (c *DataVolumeController) syncHandler(key string) error {
...
-			go c.scheduleProgressUpdate(dataVolume.Name, dataVolume.Namespace, pvc.GetUID())
+		snapshotClassName := c.getSnapshotClassForSmartClone(dataVolume)
+		if snapshotClassName != "" {
+			klog.V(3).Infof("Smart-Clone via Snapshot is available with Volume Snapshot Class: %s", snapshotClassName)
+			newSnapshot := newSnapshot(dataVolume, snapshotClassName)
+			_, err := c.csiClientSet.SnapshotV1alpha1().VolumeSnapshots(newSnapshot.Namespace).Create(newSnapshot)
+			if err != nil {
+				return err
+			}
+			err = c.updateSmartCloneStatusPhase(cdiv1.SnapshotForSmartCloneInProgress, dataVolume)
```
mhenriks: Looks like the phase is updated in the smart clone controller as well.  Can we keep it in one place?  I think it would be nice and consistent to do in the datavoume controller.  But would be more work as would have to watch snapshots.

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,387 @@
...
+
+	newPvc := newPvcFromSnapshot(snapshot, dataVolume)
+	if newPvc == nil {
+		klog.Errorf("error creating new pvc from snapshot object")
+		return nil
+	}
+
+	c.snapshotExpectations.ExpectCreations(snapshotKey, 1)
+	_, err = c.clientset.CoreV1().PersistentVolumeClaims(snapshot.Namespace).Create(newPvc)
+	if err != nil {
```
mhenriks: We appear to be assuming that the PVC definitely does not exist.  But I think that is not guaranteed to be the case.  I think it is possible for our code to try to create twice (pvc exists but is not bound).  Also, what if the pvc already happens to exist by some other means ?  Do we delete/recreate?  Or put the datavolume in a failed state?  Or just keep retrying until it does not exist (current situation).

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,387 @@
...
+		return err
+	}
+	return nil
+}
+
+func newPvcFromSnapshot(snapshot *csisnapshotv1.VolumeSnapshot, dataVolume *cdiv1.DataVolume) *corev1.PersistentVolumeClaim {
+	labels := map[string]string{
+		"cdi-controller":   snapshot.Name,
+		common.CDILabelKey: common.CDILabelValue,
+	}
```
mhenriks: `common.CDIComponentLabel` as well?

cmd/cdi-controller/controller.go
```diff
@@ -113,7 +120,18 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
...
+		klog.Fatalf("Error building csi clientset: %s", err.Error())
+	}
+
+	extClient, err := extclientset.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatalf("Error building extClient: %s", err.Error())
+	}
+
 	cdiInformerFactory := informers.NewSharedInformerFactory(cdiClient, common.DefaultResyncPeriod)
+	csiInformerFactory := csiinformers.NewSharedInformerFactory(csiClient, common.DefaultResyncPeriod)
```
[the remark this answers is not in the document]
mhenriks: We only care about Snapshots that are created by us, right?  So we can label them.  Then if we add a labelselector to the informer, this controller will only receive callbacks for "our" snapshots

pkg/controller/datavolume-controller.go
mhenriks: Later is fine

pkg/controller/smart-clone-controller.go
```diff
@@ -0,0 +1,384 @@
...
+			}
+		}
+	} else if snapshot != nil {
+		err := c.syncSnapshot(key.(string))
+		if err != nil {
+			klog.Errorf("error processing snapshot %q: %v", key, err)
+			return true
+		}
+	}
+	return c.forgetKey(key, fmt.Sprintf("ProcessNextSnapshotItem: processing pvc/snapshot %q completed", key))
```
mhenriks: Wrong function name is string

pkg/controller/smart-clone-controller.go
mhenriks: I don't know exactly when the snapshot status transitions to/from `ReadyToUse` but it appears that `syncSnapShot` may called when the pvc exists but is not yet bound.  Is the snapshot status guaranteed to not be `ReadyToUse` at that point?

cmd/cdi-controller/controller.go
```diff
@@ -290,3 +329,40 @@ func handleSignals() <-chan struct{} {
 	}()
 	return stopCh
 }
+
+func addCrdInformerEventHandlers(crdInformer cache.SharedIndexInformer, extClient extclientset.Interface,
+	csiInformerFactory csiinformers.SharedInformerFactory, smartCloneController *controller.SmartCloneController,
+	stopCh <-chan struct{}) {
+	crdInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
```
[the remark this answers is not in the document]
mhenriks: You stop a controller by closing the channel that is passed to the controller `Run` function.  

Not sure it's that important to handle delete.  But may be interesting to see what happens when the CRD is deleted.

hack/update-codegen.sh
```diff
@@ -38,6 +39,23 @@ ${CODEGEN_PKG}/generate-groups.sh "client,informer,lister" \
...
 
+
+# Workaround due to missing groupName in doc.go for CSI external snapshotter
+# See https://github.com/kubernetes-csi/external-snapshotter/issues/112
+set +e
+doc_go="${SCRIPT_ROOT}/vendor/github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1alpha1/doc.go"
+status=`grep groupName $doc_go`
+
+if [[ -z "$status" ]]; then
+  sed -i '/^package/i \/\/ +groupName=snapshot.storage.k8s.io' $doc_go
```
mhenriks: I think it would be best to check in the modified file otherwise running `make generate` will leave give changes in working directory.  Alternatively, the github issue above has been closed.  So maybe can vendor in a more recent version?

rollandf: The fix is currently in master only, not on a released version.
If we check in the modified file, after "make deps-update" it will be overridden, and eventually checked in back without the fix.
Possible solutions:
1 - undo the change after generating in the bash script
2 - use master version of the external-snapshot
3 - keep current solution until we get an updated version of the CSI library.

What do you think?

mhenriks: I think I am okay with 1) or 2).  If you go with 2, pin to a specific commit.  What do you think @awels ?

awels: I am fine with 2 or 3, we just somehow need to make sure we update it to a released version as soon as it is released.  I think I prefer 2 with a pinned commit, to avoid leaving stuff around in certain cases.

rollandf: OK, pinned to commit and removed ugly hack.

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
mhenriks: Yeah, you would have to create a new channel and a goroutine that closes the new channel if the other channel is closed.  No big deal.  But can be done later.

mhenriks:
`make cluster-sync` timed out on openshift.  I think @awels planning fix for k8s tests.

mhenriks:
@danielerez @rollandf Looks like the operator needs volumesnapshot permissions.

awels:
> @danielerez @rollandf Looks like the operator needs volumesnapshot permissions.

Is this what is causing the cluster sync failure while waiting for it to be ready?

mhenriks:
> Is this what is causing the cluster sync failure while waiting for it to be ready?

Yeah, `make cluster-sync` with os-3.11 provider and look at the operator log.  Complaints about privilege escalation

awels:
Hmm I thought this got merged, maybe I am losing it.

rollandf:
Added permissions to operator

rollandf:
ci test please

rollandf:
ci test please

mhenriks:
Okay!!  Let's merge this thing!  Thanks @rollandf and @danielerez!!

PR #724, opened 2019-03-26
update k8s image to 1.13.3 and enable ceph

igoihman:

Changed k8s version to 1.13.3, updated gocli image and enabled Ceph.

Fixes #

**Release note**:

```release-note
NONE
```


j-griffith:
This is great, one quick question; just to verify.  Is there any value in keeping the storage option selectable?  So like by default we go with the ceph work you added to the CI but still allow a variable to override or fall back to local storage if we need to?  We can always do it as follow up work but wanted to throw it out there.  I'm happy to see this merge as is.

igoihman:
@j-griffith This change just changes the image and now we will have static provisiong with local volume and dynamic provisioning using Ceph CSI provisioner.
In order to actually start using Ceph in our tests we need to indicate its storage class (I think local storage class is still the default).
But, we need to pay attention to QE, they are running our functional tests in their system with different storage classes and we should avoid making assumptions of the storage class names - Maybe just test on all available storage classes?

igoihman:
ci test please

rollandf:
:+1:

awels:
> @j-griffith This PR just changes the image and now we will haveboth static provisioning with local storage and dynamic provisioning using Ceph CSI provisioner.
> In order to actually start using Ceph in our tests we need to indicate its storage class (I think local storage class is still the default).
> But, we need to pay attention to QE, they are running our functional tests in their system with different storage classes and we should avoid making assumptions of the storage class names - Maybe just test all available storage classes?

Maybe add a tests_args parameter that specifies the ceph/dynamic/block storage class name, so QE can override it with whatever they have set up? Which then ends up as a flag in the test suite, and our tests can simply reference that flag for the storage class name for tests that require that kind of storage.

igoihman:
@awels sounds like a good idea

PR #725, opened 2019-03-26
operator creates upload proxy route

mhenriks:

Operator creates route for upload proxy (must remove from kubevirt-ansible).

While I was in there, I also did some cleanup:  

Make sure we were watching securitycontextconstraints
Don't watch configmaps/secrets since they may be mutated

Fixes #

**Release note**:

```release-note
Operator creates UploadProxy Route for Openshift deployments
```

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -273,7 +273,6 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI) 
 
 func (r *ReconcileCDI) isMutable(obj runtime.Object) bool {
 	switch obj.(type) {
-	case *corev1.Secret:
 	case *corev1.ConfigMap:
 		return true
 	}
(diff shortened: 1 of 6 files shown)


pkg/operator/controller/controller.go
```diff
@@ -259,20 +257,27 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI)
...
+		}
+	}
+
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileCDI) shouldUpdateObject(obj runtime.Object) bool {
+func (r *ReconcileCDI) isMutable(obj runtime.Object) bool {
 	switch obj.(type) {
 	case *corev1.Secret:
```
awels: So golang doesn't by default fall through to the next statement like a lot of other languages, so the secret will return false, and the configmap will return true, but its a little confusing from the code to know which one you intended, because I would leave Secret out of the switch statement if I wanted to return false.

mhenriks: Oh man, great catch!  Clearly I was confused.  Fortunately, the operator doesn't create any secrets at this point so wasn't an issue.  But will remove the secret case for clarity.

Issue #726, opened 2019-03-26
Scratch space storage not created on minishift


mhenriks:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
> /kind enhancement

**What happened**:

Tried to use virtctl image-upload with release 1.7.0 and it failed.  Here's the controller log:

```
E0326 14:18:03.438889       1 upload-controller.go:347] error syncing 'default/upload-cirros-pvc-https-re': Error creating upload pod for pvc: default/upload-cirros-pvc-https-re: Unable to create scratch space for upload: Unable to determine storage class to use for creating scratch space
E0326 14:18:03.440531       1 upload-controller.go:347] error syncing 'default/upload-cirros-pvc-https-re': Error creating upload pod for pvc: default/upload-cirros-pvc-https-re: Unable to create scratch space for upload: Unable to determine storage class to use for creating scratch space
E0326 14:21:53.238003       1 upload-controller.go:347] error syncing 'default/upload-cirros-pvc-https-re': Error creating upload pod for pvc: default/upload-cirros-pvc-https-re: Unable to create scratch space for upload: Unable to determine storage class to use for creating scratch space
W0326 14:26:34.000439       1 reflector.go:270] github.com/openshift/client-go/route/informers/externalversions/factory.go:101: watch of *v1.Route ended with: The resourceVersion for the provided watch is too old.
E0326 14:31:53.394942       1 upload-controller.go:347] error syncing 'default/upload-cirros-pvc-https-re': Error creating upload pod for pvc: default/upload-cirros-pvc-https-re: Unable to create scratch space for upload: Unable to determine storage class to use for creating scratch space
```

Here's the command I executed:

```
virtctl image-upload --uploadproxy-url=https://$(oc get route cdi-uploadproxy -n cdi -o=jsonpath='{.status.ingress[0].host}') --pvc-name=upload-cirros-pvc-https-re --pvc-size=1Gi --image-path=./tests/images/cirros-qcow2.img
```

**What you expected to happen**:

The scratch PVC should be created successfully.  Do not raise an error if storageclass is '""`

Scratch space should be successfully created.  

**How to reproduce it (as minimally and precisely as possible)**:

See command above

**Anything else we need to know?**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 1.7.0
- Kubernetes version (use `kubectl version`):
[...]

PR #727, opened 2019-03-26
Allow missing storage class for scratch space, if none exists, create scratch PVC without SC

awels:

Allow scratch space without storage class.

Fixes #
#726

**Release note**:

```release-note
Allow scratch space without requiring storage class.
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -253,7 +253,7 @@ func newScratchPersistentVolumeClaimSpec(pvc *v1.PersistentVolumeClaim, pod *v1.
 		LabelImportPvc:   pvc.Name,
 	}
 
-	return &v1.PersistentVolumeClaim{
+	pvcDef := &v1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      pvc.Name + "-scratch",
 			Namespace: pvc.Namespace,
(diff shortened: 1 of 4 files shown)


pkg/controller/util.go
```diff
@@ -292,8 +295,8 @@ func CreateScratchPersistentVolumeClaim(client kubernetes.Interface, pvc *v1.Per
 // 1. Defined value in CDI config map.
 // 2. If 1 is not available use the 'default' storage class.
 // 3. If 2 is not available use the storage class name of the original pvc that will own the scratch pvc.
-// 4. If none of those are available, fail with an error.
-func GetScratchPvcStorageClass(client kubernetes.Interface, cdiclient clientset.Interface, pvc *v1.PersistentVolumeClaim) (string, error) {
+// 4. If none of those are available, return blank.
+func GetScratchPvcStorageClass(client kubernetes.Interface, cdiclient clientset.Interface, pvc *v1.PersistentVolumeClaim) string {
```
mhenriks: Hmmm, I think it may still be a good idea to return an error if we can't get the CDIConfig object.

awels: I considered that option, but whomever calls this will ignore that, since a blank is just as good as an actual value. Not being able to find the storage classes is the same as not having a storage class from the perspective of this function, so why make it return something that will just get ignored anyway.

mhenriks: I wouldn't say that a blank with no error is just as meaningful as a blank with error.  But maybe we are splitting hairs at this point?

awels: I am not that passionate either way, if you really want an error returned as well, I can make that happen. Personally I don't think its that significant either way.

mhenriks: NBD either way, but maybe just log an error if we can't get the CDIConfig?

awels: Sure I can do that.

PR #729, opened 2019-03-28
Fix issue with the cdi config controller hanging in k8s

awels:

The cdi config controller would hang waiting for a cache sync on routes that will never happen in k8s, thus it would not update the CDIConfig object and the scratch storage class would not be correctly found.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
We should, but since we have only one storage class right now, there is nothing to override. Once the ceph stuff gets in we can add another with an override.

PR #733, opened 2019-03-31
Adding support to upload disk image to a Raw Block PV

zvikorn:

Adding support to upload disk image to a Raw Block PV

**Release note**:

```release-note
Adding support to upload disk image to a Raw Block PV
```

### DIFF
--- a/manifests/example/upload-block-datavolume.yaml
+++ b/manifests/example/upload-block-datavolume.yaml
@@ -1,0 +1,14 @@
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: DataVolume
+metadata:
+  name: upload-block-datavolume
+spec:
+  source:
+      upload: {}
+  pvc:
+    volumeMode: Block
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 500Mi
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -12,6 +12,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	v1 "k8s.io/api/core/v1"
+	k8sv1 "k8s.io/api/storage/v1"
 
 	"github.com/onsi/ginkgo/extensions/table"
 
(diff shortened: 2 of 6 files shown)


manifests/example/upload-block-datavolume.yaml
```diff
@@ -0,0 +1,15 @@
+# This example assumes you are using a default storage class
```
awels: if you are using a block device its usually not the default storage class. then again it might be.

awels:
Are you going to add functional tests for this as well?

zvikorn:
@awels of course

doc/upload-block-pv.md
```diff
@@ -0,0 +1,115 @@
+# How to upload an image to a block raw PV User Guide
+The purpose of this document is to show how to uploadan image from a URL (http/s3) without authentication.
```
igoihman: upload an

doc/upload-block-pv.md
```diff
@@ -0,0 +1,115 @@
...
+- You have a Kubernetes cluster up and running with CDI installed and at least one PersistentVolume is available.
+- Feature-Gate 'BlockVolume' is enabled.
+
+## Upload an image to a Local raw block PV
+In case you do not upload an image to a local PV, please skip to the next section (Upload an image with DataVolume manifest).
+There are a few steps to follow, to be able to upload an image to a local raw block PV.
+
+First, create a local PV with the volumeMode field set to 'Block':
+
+```bash
```
rollandf:
```yaml

File: doc/upload-block-pv.md [hunk 4]
```diff
@@ -0,0 +1,115 @@
...
+```
+
+Note: the path you create on the node has to be the same path that you define in the PV.
+
+
+## Deploy the DataVolume manifest
+
+Create the following DataVolume manifest with volumeMode field set to Block (upload-block-datavolume.yaml):
+
+```bash
```
rollandf:
```yaml

File: doc/upload-block-pv.md [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+kubectl create -f upload-block-datavolume.yaml
+```
+
+An uploader pod will be spawned and a new upload PVC will be bound to the created PV.
+
+## Create an upload token 
+
+Create the following upload token manifest (upload-block-datavolume-token.yaml)
+```bash
```
rollandf:
```yaml

File: doc/upload-block-pv.md (hunk 2 above)
Comment (zvikorn) [reply doc/upload-block-pv.md]: yeah.... the entire line is wrong.

File: tests/upload_test.go [hunk 6]
```diff
@@ -99,21 +141,24 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
 		} else {
 			By("Verify PVC empty")
 			_, err = framework.VerifyPVCIsEmpty(f, pvc)
 			Expect(err).ToNot(HaveOccurred())
 		}
 	},
-		table.Entry("[test_id:1368]succeed given a valid token", true, http.StatusOK),
-		table.Entry("[posneg:negative][test_id:1369]fail given an invalid token", false, http.StatusUnauthorized),
+		table.Entry("[test_id:1368]succeed given a valid token", true, http.StatusOK, false),
+		table.Entry("[test_id:1368]succeed given a valid token", true, http.StatusOK, true),
```
rollandf: Maybe remove the test_id so that QE will update accordingly

tests/upload_test.go
```diff
@@ -99,21 +141,24 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
 			Expect(found).To(BeTrue())
 
-			By("Verify content")
-			same, err := f.VerifyTargetPVCContentMD5(f.Namespace, pvc, utils.DefaultImagePath, utils.UploadFileMD5, false)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(same).To(BeTrue())
-			fileSize, err := f.RunCommandAndCaptureOutput(pvc, "stat -c \"%s\" /pvc/disk.img")
-			Expect(err).ToNot(HaveOccurred())
-			Expect(fileSize).To(Equal("1073741824")) // 1G
+			if !isBlockVolume {
```
rollandf: No way for now to check content on block?

doc/upload-block-pv.md
rollandf: yaml tag instead of bash

doc/upload-block-pv.md
rollandf: yaml tag instead of bash

doc/upload-block-pv.md
rollandf: yaml tag instead of bash

doc/upload-block-pv.md
zvikorn: hmmm..... This is what we are using in all other .md's.
What's the difference?

tests/upload_test.go
zvikorn: yeah. forgot... 
the first line is an old one. just added a new parameter.

tests/upload_test.go
zvikorn: still working on this...

zvikorn:
ci test please

tests/upload_test.go
zvikorn: done

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
ci test please

zvikorn:
ci test please

doc/upload-block-pv.md
awels: The highlighting of the text block. Since you are representing some yaml here, use yaml and github will use yaml highlighting.

tests/upload_test.go
```diff
@@ -68,10 +81,39 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		Expect(deleted).To(BeTrue())
 	})
 
-	table.DescribeTable("should", func(validToken bool, expectedStatus int) {
+	table.DescribeTable("should", func(validToken bool, expectedStatus int, isBlockVolume bool) {
+
+		var pod *v1.Pod
+		var err error
+
+		if isBlockVolume {
```
awels: So instead of making the DescribeTable a complicated message based on several booleans, it would make more sense to simply make a separate
```go
var _ = Describe("Upload block volume", func() {
})
```
Where in the before each you create the block PVCs. 

DescribeTables are meant for the case where you have a few parameters to a function, and the expected output changes based on those parameters. Its easy to pass in the different parameters and expected output. But if the test changes completely with certain parameters then it is much better to write a separate test for those parameters.

doc/upload-block-pv.md
zvikorn: done

zvikorn:
ci test please

doc/upload-block-datavolume.md
```diff
@@ -0,0 +1,72 @@
+# How to upload an image to a block raw PV User Guide
```
igoihman: I think it's almost identical to the original upload doc. Maybe just add there a section about block mode?

zvikorn: Sounds good. I will remove this file and open a new PR to update doc/Upload.md

zvikorn:
I will add something for import/clone/upload in datavolume.md

awels:
I think you need to rebase and fix some arguments the CI failing on unable to compile

zvikorn:
ci test please

zvikorn:
ci test please

tests/upload_test.go
```diff
@@ -165,3 +166,129 @@ func uploadImage(portForwardURL, token string, expectedStatus int) error {
...
+	f := framework.NewFrameworkOrDie(namespacePrefix)
+
+	BeforeEach(func() {
+		pod, err := utils.FindPodByPrefix(f.K8sClient, "cdi", "cdi-block-device", "kubevirt.io=cdi-block-device")
+		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Unable to get pod %q", "cdi"+"/"+"cdi-block-device"))
+
+		nodeName := pod.Spec.NodeName
+
+		By(fmt.Sprintf("Creating storageClass for Block PV"))
+		_, err = f.CreateStorageClassFromDefinition(utils.NewStorageClassForBlockPVDefinition("manual"))
```
awels: you are not capturing the storage class in storageClass so it doesn't get cleaned up.

tests/upload_test.go
```diff
@@ -165,3 +166,129 @@ func uploadImage(portForwardURL, token string, expectedStatus int) error {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		By(fmt.Sprintf("Creating Block PV"))
+		pv, err = f.CreatePVFromDefinition(utils.NewBlockPVDefinition("local-volume", "1G", nil, "manual", nodeName))
```
awels: The block PVs real size is 500M not 1G

tests/upload_test.go
```diff
@@ -165,3 +166,129 @@ func uploadImage(portForwardURL, token string, expectedStatus int) error {
...
+		if validToken {
+			By("Verify PVC status annotation says succeeded")
+			found, err := utils.WaitPVCPodStatusSucceeded(f.K8sClient, pvc)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue())
+
+			same, err := f.VerifyTargetPVCContentMD5(f.Namespace, pvc, utils.DefaultPvcMountPath, utils.UploadBlockDeviceMD5)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(same).To(BeTrue())
+			fileSize, err := f.RunCommandAndCaptureOutput(pvc, "lsblk -n -b -o SIZE \"%s\" /pvc")
```
awels: It doesn't like this command, the %s looks suspicious to me, since that a golang string substitution, but that is not what is happening here.

tests/upload_test.go
```diff
@@ -165,3 +166,129 @@ func uploadImage(portForwardURL, token string, expectedStatus int) error {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(fileSize).To(Equal("1073741824")) // 1G
```
awels: This should never be true since the real size is 500M.

tests/upload_test.go
zvikorn: :-(

tests/upload_test.go
zvikorn: fixed

tests/upload_test.go
zvikorn: changed it

PR #762, opened 2019-04-25
Fix issues with size header in cloning sometimes reading beginning of…

awels:
… data.

Fixes cloning sometimes missing the first block when cloning.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-cloner/clonertarget.go
+++ b/cmd/cdi-cloner/clonertarget.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bufio"
 	"flag"
 	"fmt"
 	"io"
@@ -10,6 +9,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	v1 "k8s.io/api/core/v1"
@@ -21,7 +21,7 @@ import (
 
 type prometheusProgressReader struct {
 	util.CountingReader
-	total int64
+	total uint64
 }
 
 const (
@@ -79,7 +79,7 @@ func main() {
 	}
 
 	//re-open pipe with fresh start.
-	out, err := os.OpenFile(*namedPipe, os.O_RDONLY, 0600)
+	out, err := os.OpenFile(*namedPipe, os.O_RDONLY, os.ModeNamedPipe)
 	if err != nil {
 		klog.Errorf("%+v", err)
 		os.Exit(1)
@@ -114,14 +114,14 @@ func main() {
 	klog.V(1).Infoln("clone complete")
 }
 
-func collectTotalSize() (int64, error) {
+func collectTotalSize() (uint64, error) {
 	klog.V(3).Infoln("Reading total size")
-	out, err := os.OpenFile(*namedPipe, os.O_RDONLY, 0600)
+	out, err := os.OpenFile(*namedPipe, os.O_RDONLY, os.ModeNamedPipe)
 	if err != nil {
-		return int64(-1), err
+		return uint64(0), err
 	}
 	defer out.Close()
-	return readTotal(out), nil
+	return readTotal(out)
 }
 
 func (r *prometheusProgressReader) timedUpdateProgress() {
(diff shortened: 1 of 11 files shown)


cmd/cdi-cloner/clonertarget.go
```diff
@@ -145,18 +145,19 @@ func (r *prometheusProgressReader) updateProgress() {
...
-		return -1
+		return -1, err
+	}
+	if n != len(b) {
+		// Didn't read all 8 bytes..
+		return -1, errors.New("Didn't read all bytes for size header")
 	}
-	klog.V(1).Infoln(fmt.Sprintf("total size: %s", totalText))
-	return total
+	size := int64(binary.LittleEndian.Uint64(b))
```
aglitke: Is 8 bytes enough to hold the maximum size you want to support?  The largest uint64 is 18446744073709551616 which would take 20 ascii bytes to represent.  

Also, You're parsing a 64 bit unsigned integer and converting it to a 64 bit signed integer.  Aren't you losing half of your precision?  Will the size 2^63+1 (9223372036854775809) be handled correctly?

tests/framework/pv.go
```diff
@@ -15,3 +15,28 @@ func (f *Framework) CreatePVFromDefinition(def *k8sv1.PersistentVolume) (*k8sv1.
 func (f *Framework) WaitTimeoutForPVReady(pvName string, timeout time.Duration) error {
 	return utils.WaitTimeoutForPVReady(f.K8sClient, pvName, timeout)
 }
+
+// ResetBlockPVtoZero resets the device to the initial state and wipes any junk left on it.
+func (f *Framework) ResetBlockPVtoZero() error {
```
aglitke: How about naming this clearBlockPV?

cmd/cdi-cloner/clonertarget.go
aglitke: Best we add some testing to this code that to ensure that numbers on the periphery work as expected.  ie. 2^64, 2^32+1, 0, -1, etc.

awels: int64 by definition is 8 bytes 8x8=64, I am not sending ascii characters, but the actual bytes, so 8 bytes is enough. Yeah there is no reason I can't use uint64 all the way, let me fix that.

Adding tests to catch edge cases.

tests/framework/pv.go
awels: Sure

awels:
ci test please

awels:
ci test please

cmd/cdi-cloner/header/sizeheader.go
```diff
@@ -0,0 +1,61 @@
...
+	out, err := os.OpenFile(*namedPipe, os.O_WRONLY, os.ModeNamedPipe)
+	if err != nil {
+		klog.Errorf("Unable to open named pipe: %+v", err)
+		os.Exit(1)
+	}
+	defer out.Close()
+
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, size)
+	out.Write(b)
```
mhenriks: I think it would be wise to check the values returned by this function

awels: Switched to writing with print, so this whole file goes away.

awels:
> Wondering why the go program is necessary? Can't we do this in bash?
> 
> On the source, do: `printf "%016x" $size >$pipe_dir`
> 
> On the target, make a string from the first 16 bytes and `strconv.ParseUint(zeroPaddedString, 16, 64)`

Was trying to get everything in 8 bytes instead of 16, but this is much simpler.

tests/cloner_test.go
```diff
@@ -32,7 +32,7 @@ const (
 	controllerSkipPVCCompleteTimeout = 90 * time.Second
 )
 
-var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:component]Cloner Test Suite", func() {
+var _ = FDescribe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:component]Cloner Test Suite", func() {
```
mhenriks: FDescribe

PR #773, opened 2019-04-29
Enable use of external provider

### LINKED ISSUE #772: Enable use of external provider
/kind enhancement
As a developer, I would like to be able to make modifications to the CDI code and test them in my own cluster with my own storage rather than relying upon the provided CDI images.

Ideally we'd have two different possibilities here:
1. An existing test cluster
- Build images for CDI based on my changes and push them to my own Image Repo
- Use the CDI Makefile/Workflow to deploy my changes to my running Kubernetes Cluster

2. An existing dev cluster from Kubernetes source with my custom changes
- Same as item 1, but requires slightly different implementation

j-griffith:

As a developer, I would like to be able to make
modifications to the CDI code and test them in
my own cluster with my own storage rather than
relying upon the provided CDI images.

This change adds an external provider and makes some slight
modifications to enable it's use.

Assuming an existing kubernetes cluster with a properly configured
kubectl client on your system:

```bash
export KUBEVIRT_PROVIDER=external
export DOCKER_REPO=quay.io/barney-rubble
```

Then running `make cluster-sync` will build the CDI Images and push them
to the specified registry, generate the deployment manifests with the
appropriate registry and tag, and finally deploy the CDI operator to
your existing cluster.

The existing workflow is unchanged, the default `make cluster-up && make
cluster-sync` will work the same as it has in the past.

Partial Fix #772 (handles the external provider, but not local-cluster-up.sh use case)

**Release note**:

```release-note
Adds the ability to build locally and deploy to an existing Kubernetes cluster.
```


j-griffith:
Running a few more tests and I'll get a release note and rebase.

cluster/ephemeral_provider.sh
```diff
@@ -0,0 +1,28 @@
...
+  container=""
+  container_alias=""
+  images="${@:-${DOCKER_IMAGES}}"
+  for arg in $images; do
+      name=$(basename $arg)
+      container="${container} registry:5000/${name}:latest"
+  done
+
+  # We don't need to seed the nodes, but in the case of the default dev setup we'll just leave this here
+  for i in $(seq 1 ${KUBEVIRT_NUM_NODES}); do
```
awels: Just and FYI why this code was in here in the first place. This is to make sure the nodes have the latest version of the containers in case the pull policy is not set to always pull.

cluster/ephemeral_provider.sh
```diff
@@ -0,0 +1,28 @@
...
+      echo "node$(printf "%02d" ${i})" "echo \"${container}\" | xargs \-\-max-args=1 sudo docker pull"
+      ./cluster/cli.sh ssh "node$(printf "%02d" ${i})" "echo \"${container}\" | xargs \-\-max-args=1 sudo docker pull"
+      # Temporary until image is updated with provisioner that sets this field
+      # This field is required by buildah tool
+      ./cluster/cli.sh ssh "node$(printf "%02d" ${i})" "sudo sysctl -w user.max_user_namespaces=1024"
```
awels: This is here for the registry container so our functional tests will run on it. To test for instance import from registry.

cluster/external/provider.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+function _kubectl(){
```
awels: Any particular reason you didn't modify cluster/kubectl.sh to check for external provider and point it to the right kubectl in that case?

cluster/ephemeral_provider.sh
j-griffith: ack, I'll likely point that out in the comment for future John :)

cluster/external/provider.sh
j-griffith: That's another approach (maybe better, not sure); I went this route because it was easy to just convert to a function and stuff uniqueness into the appropriate provider script and keep them extendable and independent.  I can look at the other possibility though and see if it works better.  First to figure out what I broke :)

mhenriks:
@j-griffith you tried this on a mac?  If not, I'll give it a shot.

j-griffith:
> @j-griffith you tried this on a mac? If not, I'll give it a shot.

Negative, that would be great though.  I don't use my mac much any more.

cluster/clean.sh
```diff
@@ -3,6 +3,11 @@
 source ./cluster/gocli.sh
 source ./hack/build/config.sh
 
+if [ "${KUBEVIRT_PROVIDER}" == "external" ]; then
+  echo "No clean up implemented for external providers!"
```
mhenriks: Why no clean?

j-griffith: cuz I'm a slacker :)  I ran into an issue trying to implement it:  https://github.com/kubevirt/containerized-data-importer/issues/774

I'll take another look and maybe do another iteration.

j-griffith: Added clean functionality

j-griffith:
ci test please

j-griffith:
ci test please

cluster/sync.sh
```diff
@@ -6,45 +6,32 @@ cdi="${cdi##*/}"
...
+source ./cluster/${KUBEVIRT_PROVIDER}/provider.sh
 
 CDI_NAMESPACE=${CDI_NAMESPACE:-cdi}
 
 # Set controller verbosity to 3 for functional tests.
 export VERBOSITY=3
 
 registry_port=$($gocli ports registry | tr -d '\r')
-registry=localhost:$registry_port
+registry=${IMAGE_REGISTRY:-localhost:$registry_port}
```
mhenriks: I think this should be `DOCKER_REPO` not `IMAGE_REGISTRY`

j-griffith: Oh!  Yes!  I was thinking about changing away from DOCKER_REPO and just use a generic image registry variable (ie quay.io etc).  I decided to do that as a follow up but didn't clean that last part out of there.

j-griffith: I think I've got that fixed up at lest a little better than what I had.  Let me know if that looks ok to you.  We can always work out some things in more detail later too.

j-griffith:
ci test please

PR #777, opened 2019-04-30
Remove tar from the readers

awels:

This PR removes the ability to tar a kubevirt disk image, as that makes no sense since its a single file.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
ci test please

mhenriks:
I understand that disk images are single files.  And tarring a single file doesn't make much sense.  But I thought that CDI was also supposed to be a more general purpose way to just put crap on a PVC?  If so, then I could imagine a tar reader being useful.

awels:
Right, and it is still there for 'archive' content type, just not kubevirt content type.

mhenriks:
@awels I must be confused.  Can you point we to where/how tar is handled for archive with http source?  Do we have a functional test for that?

awels:
So tar works IF the following is true: contentType: archive, which will then send the data to tar for extraction. I am like and the test should be in transport_test.go. But I looked, and its not there. Let me add it there, since we really should be testing archive content type as well

awels:
ci test please

awels:
ci test please

awels:
ci test please

PR #779, opened 2019-05-01
Taking into account sparseness when cloning from one raw block pv to another

### LINKED ISSUE #778: Cloning block PV does not take into account sparseness
/kind enhancement

**What happened**:
When cloning from one FS PV to another FS PV, we take into account sparseness whereas for block PV we do not.

zvikorn:

see $title
Fixes #778 

```release-note
Taking into account sparseness when cloning from one raw block pv to another
```

### DIFF
--- a/cmd/cdi-cloner/cloner_startup.sh
+++ b/cmd/cdi-cloner/cloner_startup.sh
@@ -59,7 +59,7 @@ if [ "$obj" == "source" ]; then
         #Write the size to the pipe so the other end can read it.
         printf "%016x" $size >$pipe_dir
         echo "cloner: writing the image to $pipe_dir"
-	    dd if=$image_dir bs=64K of=$pipe_dir
+	    dd conv=sparse if=$image_dir bs=64K of=$pipe_dir
     fi
     echo "cloner: finished writing image to $pipe_dir"
     exit 0


cmd/cdi-cloner/cloner_startup.sh
```diff
@@ -59,7 +59,7 @@ if [ "$obj" == "source" ]; then
         #Write the size to the pipe so the other end can read it.
         printf "%016x" $size >$pipe_dir
         echo "cloner: writing the image to $pipe_dir"
-	    dd if=$image_dir bs=64K of=$pipe_dir
+	    dd conv=sparse if=$image_dir bs=64K of=$pipe_dir
```
mhenriks: Pretty sure this setting only makes sense for files and could actually be dangerous for a block device.  Unlike with files on a file system, unwritten to regions of a block device are not guaranteed to be zero'd out.  So we actually want to write the zeros from the source.

zvikorn: @aglitke Is this something we did not think about?

zvikorn: @mhenriks can we zero out the entire block device before we copy? does it make sense?

mhenriks: I think that would make the overall operation take longer than omitting `conv=sparse`

aglitke: I agree with @mhenriks that we can only safely write selectively to the device if we know it has been zeroed beforehand.  Also, I am not sure this will actually work anyway.  conv=sparse is most likely implemented using seek and you cannot use seek with a fifo.

zvikorn: @mhenriks @aglitke so sparseness is relevant for FS more than to a block device? No worthy way to remove the 'holes"? should I close this PR?

awels:
Based on comments closing.

PR #785, opened 2019-05-03
Verify accessModes in DV spec.

awels:

Validate that the accessModes we pass in as part of the DV, are actually valid.
Fixes #

**Release note**:

```release-note
BUGFIX: Validate DataVolume AccessModes
```


j-griffith:
ci test please

pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -206,6 +207,24 @@ func validateDataVolumeSpec(field *k8sfield.Path, spec *cdicorev1alpha1.DataVolu
 		return causes
 	}
 
+	accessModes := spec.PVC.AccessModes
+	if len(accessModes) > 1 {
```
awels: Note kubernetes itself will return an error if there are no access modes, so no need to check for it here.

PR #792, opened 2019-05-08
Added some basic storage information

### LINKED ISSUE #776: Provide example docs for (dynamic ?) PV storage creation for DataVolume's PVCs
It looks like the design of DataVolumes relies on some dynamic mechanism for binding PVCs to PVs, or on the cluster admin to have already provided a suitable PV.

For instance https://kubevirt.io/2018/CDI-DataVolumes.html relies on OpenShift providing "dynamic provisioning for storage running OpenShift Cloud Storage (GlusterFS)"...

However it is not really obvious to deploy such PVs with a minimal Kubernetes cluster (plain kubeadm init on single node Linux machine), by just reading KubeVirt / CDI docs, IMHO.

I think it would be great to provide a minimal demo/instructions to give hints on either how to install a dynamic PV allocator (local hostpath kind), or just "trivial" PV creation instructions.

Thanks in advance.

awels:

Adds some documentation on how to provision storage for kubernetes.

Fixes #776 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/doc/basic_pv_pvc_dv.md
+++ b/doc/basic_pv_pvc_dv.md
@@ -1,0 +1,200 @@
+## Kubernetes Storage Overview
+
+As a Containerized Data Importer user, I need to have a basic understanding of Kubernetes storage. The [Kubernetes storage documentation](https://kubernetes.io/docs/concepts/storage/) is very in depth, but we would like to highlight the basic building blocks of storage in Kubernetes.  The basic building block of storage in Kubernetes is Persistent Volumes ([PV](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)) which are a unit of storage. PVs define their size and other information needed to define the storage. When a user wants to use storage they define a Persistent Volume Claim ([PVC](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)), which specifies the size of storage they want, and some other optional information. Kubernetes then tries to match an available PV to the PVC. If one is found, then the PVC is bound to the PV and the PVC becomes available. A pod can now mount the PVC using the pod spec when it starts.
+
+In CDI we have seen three types of storage being used for different use cases.
+* Manual hostPath based storage
+* Local volume based storage
+* Dynamic provisioned storage
+
+Each one is useful for different use cases, which we will explain below. Note this document will likely become out of date as storage progresses at a rapid rate in Kubernetes.
+
+### Manual hostPath based storage
+Manual hostPath based storage is basically a mapping between a path on the host to a PV. This is the simplest form of storage and should be a good starting point. If you are trying CDI with [minikube](https://github.com/kubernetes/minikube) we will provide an example to get started.
+
+#### minikube example
+The contents here are directly taken from the [minikube persistent volume documentation](https://github.com/kubernetes/minikube/blob/master/docs/persistent_volumes.md) and we encourage you to read and understand that document.
+
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: pv0001
+spec:
+  accessModes:
+    - ReadWriteOnce
+  capacity:
+    storage: 5Gi
+  hostPath:
+    path: /data/pv0001/
+```
+
+This will create a 5Gi sized PV, and now you are ready to apply one of the many [examples](../manifests/example) to this PV. If you run 
+
+```bash
+kubectl create -f  https://github.com/kubevirt/containerized-data-importer/blob/master/manifests/example/import-kubevirt-datavolume.yaml
+```
+
+It will create a Data Volume (DV) and import a cirros virtual disk image to that DV. If you try some of the other examples you might find that some of them require [scratch space](scratch-space.md). To create scratch space you simply need another PV available and CDI will automatically make use of it while importing. For cloning you will also need two PVs, one that contains the source data, and one that will be the target.
+
+### Local volume based storage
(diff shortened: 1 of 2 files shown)


awels:
ci test please

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,219 @@
+# Storage
+
+Storage can be a tricky subject in kubernetes so we would like to give some basic guidance on how to implement storage for use with CDI. Storage in kubernetes is defined by Persistent Volumes ([PV](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)) which are a unit of storage. PVs define their size and other information needed to define the storage. When a user wants to use storage they define a Persistent Volume Claim ([PVC](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)), which specifies the size of storage they want, and some other optional information. Kubernetes then tries to match an available PV to the PVC. If one is found, then the PVC is bound to the PV and the PVC becomes available. A pod can now mount the PVC using the pod spec when it starts.
```
j-griffith: I'd leave out the "tricky" part (even if it's somewhat true) and instead just provide a high level description of storage in Kubernetes, mostly rely on the upstream docs you already have linked.  Maybe break out this first section you have into some sub-sections... somethign like:

```
## Kubernetes Storage Overview
As a user of CDI it's important to have a clear understanding of the storage objects provided by Kubernetes.  Be sure you have a thorough understanding of [P(https://kubernetes.io/docs/concepts/storage/persistent-volumes/))ersistentVoumes[ and [PersistentVolumeClaims](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)).  

## Types of Storage 
There are three main categories of provisioners available in Kubernetes:
1. Manual hostPath
2. Local Disk
3. Dynamic Provisioners
```

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,219 @@
...
+
+Cluster administrators can group PVs using storage classes. A [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/) is a way for administrators to describe the classes of storage they offer. You can specify which storage class to use when you request a PVC, this allows you to control which storage is used.
```
j-griffith: I don't know if I'd go into this level of detail (storage classes, PVs etc) in our docs; as opposed to pointing to K8s docs as part of a K8s storage overview piece type thing.  I don't think we want to be an addendum to Kubernetes documentation, just mention and point to it perhaps.

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,219 @@
...
+
+In Containerized Data Importer (CDI) we have seen three types of storage being used for different use cases.
```
j-griffith: See note above, replacing this with a "Types of Storage" section.  Again, I'd keep it pretty brief and instead rely on Kubernetes docs to do the work for you.

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,219 @@
...
+* Manual hostPath based storage
+* local volume based storage
+* Dynamic provisioned storage
+
+Each one is useful for different use cases, which we will explain below. Note this document will likely become out of date as storage progresses at a rapid rate in kubernetes.
+
+### Manual storage
+We see this when someone just started with kubernetes and [kubevirt](https://github.com/kubevirt/kubevirt) and are using something like [minikube]( https://kubernetes.io/docs/setup/minikube/) or [minishift](https://github.com/minishift/minishift). These do not come preconfigured with storage and it is somewhat unclear on how to use local storage for some testing purposes. We will describe how to manually create a PV, and a matching PVC for use with CDI. Note a storage class is not defined nor required.
```
j-griffith: So maybe now switch the doc up a bit... So instead of walking through these thing like this; instead just convert this to a section of "Example of how to use local storage in a Minikube deployment".

Then have a section that just points out that everything works the same if you already have a provisioner configured etc.

awels:
@j-griffith Okay, tried to improve the document based on your comments, hopefully I got what you were saying.

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,200 @@
+## Kubernetes Storage Overview
```
j-griffith: Thanks for reworking that, this section is better IMO.  Personally I'd leave most of the detail out of it, but I know you want to include it and I think others do as well so that makes sense.

It gets a little tricky because of dynamic-hostpath, hostpath, local, dynamic-local-provisioner, the CSI plugins etc etc etc

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,200 @@
...
+
+In CDI we have seen three types of storage being used for different use cases.
+* Manual hostPath based storage
+* Local volume based storage
+* Dynamic provisioned storage
+
+Each one is useful for different use cases, which we will explain below. Note this document will likely become out of date as storage progresses at a rapid rate in Kubernetes.
```
j-griffith: Save yourself :)

doc/basic_pv_pvc_dv.md
awels: The goal is to have something for people to kick the tires with, storage is complicated, and the amount of permutations is huge, so a couple of 'simple' examples to get folks going is all I am aiming for.

doc/basic_pv_pvc_dv.md
awels: Indeed

doc/basic_pv_pvc_dv.md
```diff
@@ -0,0 +1,200 @@
...
+        - name: provisioner-config
+          configMap:
+            name: local-storage-config
+        - name: local-storage
+          hostPath:
+            path: /mnt/local-storage
+```
+
+### Dynamic provisioning storage
+The difference between dynamic provisioning and static provisioning is that PVs are created on demand. If a user requests a 5Gi PVC, a matching PV is created (if possible) and bound to the PVC. There are many dynamic provisioned storage solutions out there like gluster and ceph (there are also hostPath based dynamic provisioning projects out there).
```
rollandf: Maybe rephrase this:
There are many dynamic provisioned storage solutions out there like gluster and ceph (there are also hostPath based dynamic provisioning projects out there).

There are many dynamic provisioned storage solutions available like gluster,  ceph and hostPath based dynamic provisioning.

awels: Sounds good to me, let me make that change.

PR #794, opened 2019-05-08
Fix cluster-clean

### LINKED ISSUE #774: Delete CDI Namespace hangs forever
/kind bug

**What happened**:

After deploying CDI to my Kubernetes 1.14 cluster, I wanted to quickly remove it so I just used `kubectl delete ns cdi`.  All of the objects were in fact cleaned up succesfully, however the CDI namespace itself appears to hang in terminating state indefinitely.

I believe this also occurs with Kubevirt as well, perhaps something in the operator implementation, or maybe a side effect of running on external clusters?

I expect all CDI resources to be removed, including the CDI Namespace

Deploy CDI to a cluster, wait for it to finish deployment, then run `kubectl delete ns cdi`

**Environment**:

Running from master against an external 1.14 cluster

j-griffith:
We were missing some resources in the cleanup on make cluster-clean
(when running against an external cluster).  We also weren't patching
any of the finalizers so the result is a delete of the cdi namespace
would hang in terminating state forever.

This change explicitly checks all resources on the cluster for a cdi
label and it also patches the cdi objects to clear out the finalizer.

Fixes #774

```release-note
NONE
```

### DIFF
--- a/cluster/clean.sh
+++ b/cluster/clean.sh
@@ -8,6 +8,12 @@ echo "Cleaning up ..."
 
 OPERATOR_CR_MANIFEST=./_out/manifests/release/cdi-cr.yaml
 OPERATOR_MANIFEST=./_out/manifests/release/cdi-operator.yaml
+LABELS=("operator.cdi.kubevirt.io" "cdi.kubevirt.io")
+NAMESPACES=(default cdi)
+
+set +e
+_kubectl patch cdi cdi --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
+set -e
 
 if [ -f "${OPERATOR_CR_MANIFEST}" ]; then
     if _kubectl get crd cdis.cdi.kubevirt.io ; then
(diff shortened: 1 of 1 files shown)


cluster/clean.sh
```diff
@@ -21,25 +27,49 @@ if [ -f "${OPERATOR_MANIFEST}" ]; then
...
-    _kubectl -n ${i} delete secrets -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
-    _kubectl -n ${i} delete configmaps -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
-    _kubectl -n ${i} delete pvc -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
-    _kubectl -n ${i} delete pods -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
-    _kubectl -n ${i} delete rolebinding -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
-    _kubectl -n ${i} delete roles -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
-    _kubectl -n ${i} delete serviceaccounts -l 'operator.cdi.kubevirt.io' -l 'cdi.kubevirt.io'
+for n in ${NAMESPACES[@]}; do
+  for label in ${labels[@]}; do
+    _kubectl -n ${n} delete deployment -l ${lael}
```
awels: you have a typo, should be label not lael

j-griffith: Done, thanks!

awels:
ci test please

awels:
ci test please

PR #795, opened 2019-05-10
Enable return of CDI objects from `get all`

j-griffith:

This adds the `all` category to the CDI objects.  Currently CDI objects
aren't returned from a get all request, adding the category all
parameter to the objects fixes that.

With this change a `get all` will include things like a users DataVolumes.

Note, this is NOT allowed for tokens, so we don't set that on upload
tokens.

**Release note**:

```release-note
Make CDI objects visible to `get all` queries.
```


j-griffith:
ci test please

awels:
ci test please

PR #805, opened 2019-05-16
Test using Ceph-RBD CSI plugin as default SC

j-griffith:
Testing this locally and in the CI pipeline to see if we can use a more
realistic storage class as our default for functional testing.

We wouldn't actually use this change, instead we'd modify the CI image; but for now as a sort of canary we can run this config a bit and see how things go.


awels:
@j-griffith You this will 100% fail all the openshift tests right since we don't have csi rbd in that image.

j-griffith:
> @j-griffith You know this will 100% fail all the openshift tests right since we don't have csi rbd in that image.

Getting to that.. and some other things.

j-griffith:
Seems @awels has another approach to this in #815

PR #809, opened 2019-05-21
Adding code to not allow importing from registry when contenType is not kubevirt

### LINKED ISSUE #808: The check not to allow import from registry if contentType is not 'kubevirt' was removed
/kind bug

**What happened**:
The check not to allow import from registry if contentType is not 'kubevirt' was removed

getting error when trying to import from registry with contentType = archive. Instead the import is completed.

zvikorn:

fixing a bug where we allow import from registry when contentType is not kubevirt
Fixes #808 

**Release note**:

```release-note
BugFix: The check not to allow import from registry if contentType is not 'kubevirt' was removed #808 
```

### DIFF
--- a/pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
+++ b/pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
@@ -147,6 +147,16 @@ func validateDataVolumeSpec(field *k8sfield.Path, spec *cdicorev1alpha1.DataVolu
 		return causes
 	}
 
+	if spec.Source.Registry != nil && spec.ContentType != "" && string(spec.ContentType) != string(cdicorev1alpha1.DataVolumeKubeVirt) {
+		sourceType = field.Child("contentType").String()
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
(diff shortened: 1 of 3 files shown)


zvikorn:
Done

pkg/apiserver/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -147,6 +147,16 @@ func validateDataVolumeSpec(field *k8sfield.Path, spec *cdicorev1alpha1.DataVolu
 		return causes
 	}
 
+	if spec.Source.Registry != nil && spec.ContentType != "" && string(spec.ContentType) != string(cdicorev1alpha1.DataVolumeKubeVirt) {
+		sourceType = field.Child("contentType").String()
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("ContentType must be KubeVirt when SourceType is Registry"),
```
awels: "ContentType must be KubeVirt when Source is Registry"

zvikorn: I think this is what we decided. For now, I do not believe we support special 'archive' case for registry.

zvikorn: :-) done

PR #813, opened 2019-05-24
Add a datavolume test to use csi-rbd block mode

j-griffith:
This adds a simple enhancement to the existing datavolume functional
test by adding a context for csi-rbd storage class when available.

This only adds another iteration to the dv block mode test for now, more
enhancements to follow later.

This just runs a second iteration of the existing DV importer block mode functional test with the csi-rbd storage class if it's available.

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/utils/storageclass.go
+++ b/tests/utils/storageclass.go
@@ -57,3 +57,21 @@ func DeleteStorageClass(clientSet *kubernetes.Clientset, storageClass *storageV1
 		return false, err
 	})
 }
+
+// GetStorageClass uses the provided client to attempt to get the specified storage class by name
+func GetStorageClass(clientSet *kubernetes.Clientset, scName string) (*storageV1.StorageClass, error) {
+	sc, err := clientSet.StorageV1().StorageClasses().Get(scName, metav1.GetOptions{})
+	return sc, err
+}
+
+// StorageClassExists is a simple check to see if the given SC exists or not
+func StorageClassExists(clientSet *kubernetes.Clientset, scName string) (bool, error) {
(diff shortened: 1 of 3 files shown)


tests/utils/storageclass.go
```diff
@@ -57,3 +59,24 @@ func DeleteStorageClass(clientSet *kubernetes.Clientset, storageClass *storageV1
 		return false, err
 	})
 }
+
+// GetStorageClass uses the provided client to attempt to get the specified storage class by name
+func GetStorageClass(clientSet *kubernetes.Clientset, scName string) (*storageV1.StorageClass, error) {
+	sc, err := clientSet.StorageV1().StorageClasses().Get(scName, metav1.GetOptions{})
+	if err != nil {
+		fmt.Printf("\nOk... I expected this error so wth: %v\n", err)
```
awels: I would do an errors.Wrap here, instead of a Printf that may or may not end up somewhere.

awels:
So I am currently working on a PR that will allow you to specify the storage class to use for your block device tests. And I am basically removing the hookey manual block device things.

tests/utils/storageclass.go
j-griffith: Oops, that's actually a debug line I had in there; never meant for you to  see it :)

j-griffith:
> So I am currently working on a PR that will allow you to specify the storage class to use for your block device tests. And I am basically removing the hookey manual block device things.

So that makes this obsolete then?

awels:
Yes likely, look at #815 for details.

j-griffith:
Closing since @awels is working the same sort of thing in #815

PR #815, opened 2019-05-24
Allow specifying block storage class.

awels:

Allow one to specify which storage class to use for block devices.

Fixes #

**Release note**:

```release-note
NONE
```


hack/build/run-functional-tests.sh
```diff
@@ -29,6 +29,8 @@ KUBECONFIG=${KUBECONFIG:-${CDI_DIR}/cluster/.kubeconfig}
 GOCLI=${GOCLI:-${CDI_DIR}/cluster/cli.sh}
 KUBE_MASTER_URL=${KUBE_MASTER_URL:-""}
 CDI_NAMESPACE=${CDI_NAMESPACE:-cdi}
+SNAPSHOT_SC=${SNAPSHOT_SC:-csi-rbd}
```
j-griffith: I'm wondering if there's a better way to tackle this than specifying each SC by feature.  For now it's not horrible to have two; but I see this growing and getting unwieldy over time.  We could introduce *special* requirements for functional testing storage classes whereby we include annotations in the sc itself on what features it supports.

It's not great either, maybe there's a better solution; but I think dumping in more env variables to call the test with may be ugly after a while.

I do agree with basically making all of these default to the csi-rbd sc since we're deploying that in the dev environment.

awels:
ci test please

awels:
ci test please

awels:
@j-griffith So is there something specific I am missing in this PR or you don't like the general premise?

awels:
ci test please

awels:
ci test please

hack/build/run-functional-tests.sh
aglitke: The problem with annotating storage classes is how do we handle the case where multiple storage classes carry the same annotations?

tests/smartclone_test.go
```diff
@@ -62,7 +62,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
 
 	Describe("Verify DataVolume Smart Cloning - Negative flow", func() {
 		It("verify inapplicable smart-clone dv", func() {
-			if controller.IsOpenshift(f.K8sClient) {
+			if !f.IsSnapshotStorageClassAvailable() {
```
aglitke: I'm confused why this check switched from checking for openshift to checking for the presence of a storage class.

awels: Basically with this PR, instead of checking for openshift (which at some point will have snapshotting as well) we check if the passed in snapshot storage class exists. If so, then we run the test.

This allows us to have the check by feature instead of openshift vs kubernetes where kubernetes is simply ahead by feature for a short period of time. This makes the test more robust.

hack/build/run-functional-tests.sh
awels: So I think it should be manageable, since certain features will be mainlined and every storage class will have them at some point, then we can remove that feature parameter. So I am expecting a maybe up to half a dozen or so features, that will exist in alpha/beta for a while at a time.

I think @j-griffith is fine with this at this point since we discussed it offline a little more.

awels:
ci test please

awels:
ci test please

PR #816, opened 2019-05-26
functional tests for malicious qcow2 images

danielerez:
Added functional tests to ensure that malicious qcow2 images are being rejected. 
I.e. the creation of import datavolume should fail when converted to raw.
The test images were taken from: https://bugs.launchpad.net/ossa/+bug/1449062

**Release note**:

```release-note
NONE
```


tests/datavolume_test.go
```diff
@@ -122,6 +123,10 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			table.Entry("[rfe_id:138][crit:high][test_id:1362]succeed creating upload dv", "upload", "", "", "upload-dv", controller.UploadReady, cdiv1.Succeeded),
 			table.Entry("[rfe_id:1115][crit:high][test_id:1478]succeed creating import dv with given valid registry url", "import-registry", "", tinyCoreIsoRegistryURL, "dv-phase-test-4", controller.ImportSucceeded, cdiv1.Succeeded),
 			table.Entry("[rfe_id:1115][crit:high][test_id:1379]succeed creating import dv with given valid url (https)", "import-https", "", httpsTinyCoreIsoURL, "dv-phase-test-1", controller.ImportSucceeded, cdiv1.Succeeded),
+			table.Entry("[rfe_id:2139][crit:high][posneg:negative][test_id:2139]fail creating import dv: invalid qcow large size", "import-http", "", InvalidQcowImagesURL+"invalid-qcow-large-size.img", "dv-invalid-qcow-large-size", controller.ImportFailed, cdiv1.Failed),
```
annastopel: @danielerez can you please in few words decribe what are those images are and which "invalidity" is actually tested?

tests/datavolume_test.go
```diff
@@ -122,6 +123,10 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			table.Entry("[rfe_id:2139][crit:high][posneg:negative][test_id:2139]fail creating import dv: invalid qcow large json", "import-http", "", InvalidQcowImagesURL+"invalid-qcow-large-json.img", "dv-invalid-qcow-large-json", controller.ImportFailed, cdiv1.Failed),
+			table.Entry("[rfe_id:2139][crit:high][posneg:negative][test_id:2139]fail creating import dv: invalid qcow large memory", "import-http", "", InvalidQcowImagesURL+"invalid-qcow-large-memory.img", "dv-invalid-qcow-large-memory", controller.ImportFailed, cdiv1.Failed),
+			table.Entry("[rfe_id:2139][crit:high][posneg:negative][test_id:2139]fail creating import dv: invalid qcow backing file", "import-http", "", InvalidQcowImagesURL+"invalid-qcow-backing-file.img", "dv-invalid-qcow-backing-file", controller.ImportFailed, cdiv1.Failed),
```
annastopel: The same as above for each one of the invalid qcow2 files

rollandf:
ci test please

tests/datavolume_test.go
awels: Been going back and forth in my head on this comment. We are currently testing if the import fails with known bad images. I am wondering if just the fail/not failed test is enough, or if we care about why the import failed (resource usage exceeded when called qemu-info info). I believe the importer pod logs should have some specifics in there.

The question is, do we care enough about that? We have unit tests that test for that specific issue. Let me know what you think.

rollandf: The purpose of these tests is to check a functional flow of importing of malicious QCOW and check that the import fails, the same way that a "user" will see it.
The unit-tests are good for checking our logics when qemu fails. However, these are using mocks to simulate the behaviour of qemu.
It is important to have also functional test to check the whole flow.

Regarding the reason of why it fails, parsing the log can be problematic, as the importer pod is restarted on failures.
The best way would be to have specific events on the DV that can be checked.
Since we don't have it, and not sure if we plan/want to have them, the current test that check if the import fails is probably best for now.

pkg/util/util.go
```diff
@@ -206,3 +206,41 @@ func CopyFile(src, dst string) error {
...
+	if err != nil {
+		return err
+	}
+
+	directory, _ := os.Open(source)
+	objects, err := directory.Readdir(-1)
+
+	for _, obj := range objects {
+		src := source + "/" + obj.Name()
+		dst := dest + "/" + obj.Name()
```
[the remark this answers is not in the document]
danielerez: Sure, done.

tests/datavolume_test.go
danielerez: Added explanations for the images inside the test.

tests/datavolume_test.go
awels: PR #825 adds the messages to the PVC event log. If we can review and merge that, we can improve the tests here.

danielerez: Added specific error message validations to the tests (using the infra introduces in PR #825 and PR #836).

PR #831, opened 2019-05-29
Refactored provider infra structure

awels:
Refactored cluster up to allow for adding more providers easily.

Refactored make cluster-up to use kubevirtci cluster-up
Added OKD 4.1 provider from kubevirtci.
Added local storage to OKD 4.1 provider.
Fixed some problems with external provider.

Fixes #

**Release note**:

```release-note
NONE
```


cluster/external/provider.sh
```diff
@@ -7,3 +7,7 @@ function _kubectl(){
 function seed_images(){
   echo "seed_images is a noop for external provider"
 }
+
+function up() {
+  echo "using external cluster"
+}
```
j-griffith: You've got an artifact here that's causing you issues

cluster/up.sh
```diff
@@ -4,54 +4,8 @@ source ./cluster/gocli.sh
...
-
-echo 'Wait until all nodes are ready'
-until [[ $(_kubectl get nodes --no-headers | wc -l) -eq $(_kubectl get nodes --no-headers | grep " Ready" | wc -l) ]]; do
-    sleep 1
-done
-echo "cluster node are ready!"
+echo "Starting cluster"
+up
 
+verify
```
j-griffith: If you want to add this like this; then you need to also make sure you add it to the external/provider.sh file.  Going forward we'd also need to add it to local if/when we add that.

awels: I don't since I am sourcing the ephemeral-provider.sh which defines the function. So if a specific provider doesn't care about it, and wants to use the one from the ephemeral-provider.sh it doesn't have to add it.

cluster/external/provider.sh
awels: I am not sure I follow, make cluster-up for an external cluster is basically a noop, but we are echoing that we are using an external cluster to let the user know. What am I missing?

cluster/up.sh
j-griffith: Sure, but that breaks the external provider; because the externa/provider.sh file does not have a verify function; and of course you can't use the ephemeral provider scripts for the external provider.

Just add a verify function to the external provider and you're good.

j-griffith: All you need to do is add a function to the external/provider.sh file:

```
function verify(){
  echo "no verification performed on external providers"
}
```

j-griffith: One other nit on this PR... so the title is "Updated k8s image to latest"; yes that's true, but at the same time there's a refactor of the provider scripts for deploying clusters.  That's ok; but I think you're changing enough here (inheritance model closer to kubevirts) that it warrants its own pr.  Not a big deal, but going forward I think it's worth considering.

cluster/external/provider.sh
j-griffith: ```function verify()```  https://github.com/kubevirt/containerized-data-importer/pull/831/files#diff-11a543d943c2b3d40534dccbf45d4449R11

cluster up call "verify()" at L#11, but the external provider doesn't have a verify function; as a result cluster-up in that configuration fails.  See my notes in the up.sh file, easy to fix.

cluster/up.sh
awels: I changed the title of the PR to better convey what it is doing, not the commit message is more descriptive.

awels: I thought I was sourcing the ephemeral provider in the external one, and I wasn't so yes I am missing the verify function, I will add.

awels:
ci test please

awels:
ci test please

awels:
ci test please

awels:
ci test please

cluster-up/cluster/external/README.md
```diff
@@ -0,0 +1,36 @@
...
+export DOCKER_PREFIX=myregistry:5000/kubevirt
+export KUBECONFIG=mycluster.conf
+export IMAGE_PULL_POLICY=Always
+make cluster-up
+```
+
+## Building and pushing to the registry
+
+```bash
+make cluster-build
```
j-griffith: There's now "cluster-build" directive in the Makefile, this should always fail, and the external provider is now useless because there's no build directive to properly build/push or set the images.

cluster-up/kubectl.sh
```diff
@@ -0,0 +1,38 @@
...
+        echo "$(pwd)/"
+    )"
+fi
+
+source ${KUBEVIRTCI_PATH}/hack/common.sh
+source ${KUBEVIRTCI_PATH}/cluster/$KUBEVIRT_PROVIDER/provider.sh
+source ${KUBEVIRTCI_PATH}/hack/config.sh
+
+if [ "$1" == "console" ] || [ "$1" == "vnc" ]; then
+    ${KUBEVIRTCI_PATH}/_out/cmd/virtctl/virtctl --kubeconfig=${kubeconfig} "$@"
```
j-griffith: What's the point of virtctl from CDI perspective?  We don't load a kubevirt service so is this useful for anything?

cluster-up/cluster/external/README.md
awels: actually, we don't have a cluster-build directive in the Makefile, we have an equivalent though, so I can change it so the push happens.

cluster-up/kubectl.sh
awels: Nope, but see my long comment in the commit.

awels:
> Why the directory changes? You moved everything out of ./cluster and into ./cluster-up
> 
> If you like it better that way for some reason that seems "ok" but then what's the point of the old "cluster" directory which is now just an empty dir?
> 
> You also missed a number of references to the "old" dir structure in hac/kREADME
> 
> The base README file appears to need updated as well... things like "./cluster/kubect.sh" are no longer valid.
> 
> It also looks like the external provider is hosed (even if I hack out the build step), the KUBEVIRT_PROVIDER variable seems to be stuck at default during build; but that could be my hack that's causing that, not sure.

So a couple of things:
1. I emptied the cluster directory but forgot to delete it, it needs to be removed obviously.
2. The whole point of this, is that we are vendoring in the cluster-up directory from the kubevirtci project. The update-codegen rebuilds that directory from the kubevirtci source based on the git hash in that file. This allows us to directly consume (and run) 100% the same cluster as all the other kubevirt projects. Before, I had hacked and trimmed it just for CDI. Now if new providers become available we just regenerate cluster-up from kubevirtci repo and voila it should work.
3. We can argue that the virtctl.sh should not be in the cluster-up code, but right now it is, and thus we have it as well.
4. I need to make a round of checks on the documentation to make sure I have updated all the links.

cluster-up/cluster/external/README.md
j-griffith: > actually, we don't have a cluster-build directive in the Makefile, we have an equivalent though, so I can change it so the push happens.

Actually that's exactly my point s/now/no/  you need to add it; if we're going to copy over the kubevirt directives and model (which I think is a good idea) we should do the whole thing.

cluster-up/kubectl.sh
j-griffith: I don't see a long comment in the commit explaining this

cluster-up/cluster/external/README.md
awels: The point is, the cluster-up is common among the different projects, but each project has different requirements for syncing their stuff into the cluster. The kubevirt model is to put the sync stuff in hack/cluster/something afaict, which is what I did here. hack/cluster/<provider> is where the sync code lives, and cluster-up/<provider> if where the cluster up stuff lives.

j-griffith:
Latest change fixes up the registry issue on my side (with external clusters).  I'm not crazy about having:

DOCKER_PREFIX and DOCKER_REPO variables, would be nice to consolidate or see if we just want to move to "IMAGE_REGISTRY" or whatever naming.  The main thing IMO is we should only have to set one.

Thanks for looking into the points I raised and getting this squared away!

j-griffith:
Ok, we've chatted offline; but for history/record-keeping:

My concern is that now that I've shown the patch didn't work for external provisioners you've fixed that with the DOCKER_REPO variable in the Makefile, so that's awesome.  I would argue that the cluster-up files should now have DOCKER_PREFIX in them any more.  If they aren't using them remove them, if they need to be substituted with DOCKER_REPO then s/DOCKER_PREFIX/DOCKER_REPO/

Bottom line is even if you tell me that it's irrelevant, as a user or new developer trying to deploy a build to an external provider, if I look at the README it won't work and there's no value in having unused variables there IMO:

```
[jdg@doublehaul containerized-data-importer] ➜   git:(pr-831) ✗  grep -r DOCKER_PREFIX
cluster-up/hack/config-default.sh:docker_prefix=${DOCKER_PREFIX:-kubevirt}
cluster-up/cluster/external/provider.sh:docker_prefix=${DOCKER_PREFIX}
cluster-up/cluster/external/provider.sh:manifest_docker_prefix=${DOCKER_PREFIX}
cluster-up/cluster/external/README.md:export DOCKER_PREFIX=myregistry:5000/kubevirt
```
If they're not used, or as you say "irrelevant' then they shouldn't be there IMO, but if you don't agree then I guess "ok".

awels:
Update I replaced DOCKER_REPO with DOCKER_PREFIX to match what we vendor in, so only DOCKER_PREFIX will be used from now on.

README.md
```diff
@@ -61,7 +61,7 @@ $ mkdir $GOPATH/src/kubevirt.io && cd $GOPATH/src/kubevirt.io
 $ git clone https://github.com/kubevirt/containerized-data-importer && cd containerized-data-importer
 $ make cluster-up
 $ make cluster-sync
-$ ./cluster/kubect.sh .....
+$ ./cluster-up/kubectl.sh .....
```
j-griffith: Might want to create a section for this and point out something like "The development tools include a version of kubectl that you can use to communicate with the cluster.  The wrapper script to communicate with your CI cluster can be invoked using `./cluster-up/kubectl.sh`.

Probably not worth going through and describing how to add your own kubectl; if people want to go that route they can figure it out.

awels:
Once the CI finishes the OKD lane, I will update with the requested README changes. I just want to see how it does with the okd lane. Locally all the tests passed using the okd provider.

PR #832, opened 2019-05-29
fix host:port not allowed in insecure registry configmap

### LINKED ISSUE #833: host:port not allowed in cdi-insecure-registries

>
/kind bug
> /kind enhancement

**What happened**:  can't add key `my-registry-host:5000` to `cdi-insecure-registries`

**What you expected to happen**: You can

`kubectl patch configmap cdi-insecure-registries -n cdi \
  --type merge -p '{"data":{"my-private-registry-host:5000": ""}}'`

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 1.9.1
- Kubernetes version (use `kubectl version`): 1.13
- Install tools:
- Others:

mhenriks:
**What this PR does / why we need it**:  Turns out you can't have `:` in a configmap key.  Which means no host:port combos in the cdi-insecure-registries configmap.  Which is probably necessary for many internal registries.

Didn't catch this in testing because our functional test registry listens on port 443.

Fixes #833

Fixed by using the `value` fields of the configmap.  Keys can be anything. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/doc/image-from-registry.md
+++ b/doc/image-from-registry.md
@@ -125,5 +125,5 @@ Add the registry to the `cdi-insecure-registries` `ConfigMap` in the `cdi` names
 
 ```bash
 kubectl patch configmap cdi-insecure-registries -n cdi \
-  --type merge -p '{"data":{"my-private-registry-host": ""}}'
+  --type merge -p '{"data":{"mykey": "my-private-registry-host:5000"}}'
 ```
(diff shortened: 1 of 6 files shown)


awels:
There seems to be a unit test failure in the pkg/controller package, but I can't tell which unit test is failing since its not reporting on that.

PR #839, opened 2019-05-31
Make sure we are using mirror.

awels:
Signed-off-by: Alexander Wels <awels@redhat.com>

**What this PR does / why we need it**:
Try a different way to have the CI system use its mirrors.
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #840, opened 2019-05-31
Change name of new deployment in leader election test.

awels:

Change what we call the new deployment in the leader election test so it doesn't match the check to see if we have multiple deployments in pre test checks. Sometimes it takes a little bit before the test deployment is removed, and this causes test failures that aren't actually failures.

Fixes #

**Release note**:

```release-note
NONE
```


mhenriks:
@awels I think these test failures are unrelated?

awels:
Yes this and #839 should help improve stability of the tests.

PR #841, opened 2019-06-01
validating webhook for clone operations

mhenriks:
**What this PR does / why we need it**: It is currently possible for a user with no RBAC privileges in the  source namespace to create a clone of a PVC.  Obvious security concern.

Fixes #

What privileges should the user have in the source namespace?  Currently going with "create pod."  Alternatives?  Anything else?   

**Release note**:

```release-note
Creating a clone now requires "create pod" permission in source namespace.
```


awels:
So wouldn't a get and/or list on the source pvc be more appropriate?

mhenriks:
@awels I think create pod is more appropriate than get/list PVC because get/list PVC doesn't actually give you access to the data which is what we care about.

awels:
The question really is, which rbac permission gives you the ability to mount a PVC to a POD in the namespace. You need create pod to create the pod, but what else?

mhenriks:
> The question really is, which rbac permission gives you the ability to mount a PVC to a POD in the namespace. You need create pod to create the pod, but what else?

If someone already knows the pvc name, I think only create pod is required.

We may want to do something different though.

mhenriks:
@aglitke @awels @j-griffith hold cancelled.  What do you say we go with this PR for now so we get something out there quickly?  Drawback being, we are not protected when cdi-apiserver is not available.  So now we have 2 replicas.  Overall better but not perfect.

mhenriks:
Going another route, so will do another PR.

PR #866, opened 2019-06-25
Turn ceph-enabled back on for ephemeral providers

j-griffith:
The move to the common cluster-up model dropped the ceph-enabled flag, this pr adds it back in for ephemeral providers.

Also, since we're adding it in, might as well make it the default SC at
the same time.

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -2,6 +2,7 @@
 
 set -e
 
+KUBEVIRT_PROVIDER_EXTRA_ARGS="${KUBEVIRT_PROVIDER_EXTRA_ARGS} --enable-ceph"
 _cli_container="kubevirtci/gocli@sha256:caf1c3f63dc5f3137795084e492a48c802a7ef2e7f7fbe1d67e5cff684d376a3"
 _cli_with_tty="docker run --privileged --net=host --rm -t -v /var/run/docker.sock:/var/run/docker.sock ${_cli_container}"
 _cli="docker run --privileged --net=host --rm ${USE_TTY} -v /var/run/docker.sock:/var/run/docker.sock ${_cli_container}"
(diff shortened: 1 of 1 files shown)


j-griffith:
Not sure if ceph landed in the ocp image, I believe it has but need confirmation for that.

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -2,6 +2,7 @@
 
 set -e
 
+KUBEVIRT_PROVIDER_EXTRA_ARGS="${KUBEVIRT_PROVIDER_EXTRA_ARGS} --enable-ceph"
```
awels: ceph should only be turned on for the k8s provider, AFAIC neither open shift provider has it enabled. I have an open task to enable ceph for OKD-4.1.0

Also cluster-up/* will be overriden when we update the vendoring, so any changes we make (like making certain storage classes the default storage class, should happen in cluster-sync). Now the ceph enabling NEEDS to happen in cluster-up, which is why I sort of punted and told people to set their environment variable prior to calling make cluster-up to enable it.

If we want to enable it all the time, we should create a PR in kubevirtci to enable it there since it got disabled there.

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -24,6 +25,10 @@ gocli=${BASE_PATH}/../cluster-up/cli.sh
 docker_prefix=localhost:$(_port registry)/kubevirt
 manifest_docker_prefix=registry:5000/kubevirt
 EOF
+
+_kubectl patch storageclass local -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
+_kubectl patch storageclass csi-rbd -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
```
awels: CI was not at all happy making the ceph provider the default storage class, if you take a look at the k8s CI run you will see a ton of failures, and all of them look like a failure when trying clean up some stuff in the pvc.

cluster-up/cluster/ephemeral-provider-common.sh
j-griffith: Thanks

PR #856, opened 2019-06-17
Added leader election functional test

awels:
Signed-off-by: Alexander Wels <awels@redhat.com>

**What this PR does / why we need it**:
Add a functional test that verifies that switching leader during an import will not interrupt the import.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #867, opened 2019-06-26
Revert "Added leader election functional test"

j-griffith:
Reverts kubevirt/containerized-data-importer#856, sorry I didn't think we needed to wait for qe review.  I'll go ahead and revert so this can be reviewed by qe prior to merge.  I think it's fine as is, but we have a process.


awels:
Its fine, if QE has issues we can always add a new PR.

PR #868, opened 2019-06-27
Set fsGroup to match user that is running in pod

awels:

This PR makes it so that the mounted volumes are accessible by the user we are running as.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -438,6 +438,7 @@ func MakeImporterPodSpec(image, verbose, pullPolicy string, podEnvVar *importPod
 		pod.Spec.SecurityContext = &v1.PodSecurityContext{
 			RunAsNonRoot: &[]bool{true}[0],
 			RunAsUser:    &[]int64{1001}[0],
+			FSGroup:      &[]int64{1001}[0],
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


awels:
Yes, if we are using a csi storage provider (in particular the csi hostpath I am dabbling with) the pvc is mounted as root by default. This stops us from writing to the pvc because we don't have the right permissions. This changes the mount to be the user we are running as.

mhenriks:
I'm wondering if/how setting these permissions will affect subsequent users of the PVC?  I'm thinking that maybe it would be best to not explicitly set user id for clone/import/upload pods?

awels:
Well as long as subsequent users are either root or 1001 they should have no issues reading the data. If we don't set the user they would have to be root only because the user writing would be root no?

mhenriks:
> Well as long as subsequent users are either root or 1001 they should have no issues reading the data. If we don't set the user they would have to be root only because the user writing would be root no?

With all the different combinations of k8s and storage providers I have no idea what the resulting permissions will be for each.  Given that you've found a storage provider that behaves differently than any other ones we've tested previously, I'm thinking it may be wise to just go with defaults.

 Setting user id to 1001 is kind of a hack and will probably have to get rid of it when we have PodSecurityPolicy support.  So I'm thinking it may make more sense to revert setting the user id for clone/import/upload pods.  Can keep it for our long running pods though.  What do you think @j-griffith?

awels:
@mhenriks @j-griffith So just abandon this, and revert setting the user PR?

mhenriks:
> @mhenriks @j-griffith So just abandon this, and revert setting the user PR?

I am not opposed to that

awels:
Okay so #897 does a revert, closing this.

PR #869, opened 2019-06-28
Validating webhook and token authorization for PVC cloning

mhenriks:

The goal of the PR is to make it more secure to do cross-namespace cloning.  We do this by checking user credentials in a validating webhook on DataVolumes.  If the user has "appropriate" permission (more on that later), a token is added as an annotation on the DataVolume.  The DataVolume controller puts that annotation on a new PVC where it is validated by the clone controller.   

For DataVolumes with  PVC sources, the webhook will check that the user has `create` permission on `datavolumes/clone-init` in the apigroup `cdi.kubevirt.io` in the source PVC namespace.   If you can think of a better name than `datavolumes/clone-init` let me know!  Adding this permission to a regular user may look like this:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: cdi-clone-source
rules:
- apiGroups: ["cdi.kubevirt.io"]
  resources: ["datavolumes/source"]
  verbs: ["create"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: mike-clone-init
  namespace: mikesproject
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cdi-clone-source
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: mike
```

Mike can now create clones of PVCs in namespace `mikesproject`

Fixes #

Notice that webhook package refactored/restructured to better share code.

The JWT token is now shared with Upload has it's own package that was expanded.  

**Release note**:

```release-note
Validating webhook and token authorization for PVC cloning
```


doc/RBAC.md
```diff
@@ -0,0 +1,164 @@
...
+rules:
+- apiGroups: ["cdi.kubevirt.io"]
+  resources: ["datavolumes/source"]
+  verbs: ["create"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: joe-cdi-cloner
+  namespace: project1
```
awels: Should this not be 'golden-images'?

pkg/apiserver/apiserver.go
```diff
@@ -162,16 +170,23 @@ func NewCdiAPIServer(bindAddress string,
 
 	})
 
-	validatingwebhook.SetClient(client)
+	err = app.createValidatingWebhook()
+	if err != nil {
+		return nil, errors.Errorf("failed to create validating webhook: %s", err)
```
awels: Should it be a %v for an err, or err.Error() if you use %s?

pkg/apiserver/apiserver.go
```diff
@@ -162,16 +170,23 @@ func NewCdiAPIServer(bindAddress string,
...
+	}
 
-	err = app.createWebhook()
+	err = app.createMutatingWebhook()
 	if err != nil {
-		return nil, errors.Errorf("failed to create webhook: %s", err)
+		return nil, errors.Errorf("failed to create mutating webhook: %s", err)
```
awels: Should it be a %v for an err, or err.Error() if you use %s?

pkg/apiserver/webhooks/datavolume-mutate_test.go
```diff
@@ -0,0 +1,186 @@
...
+var _ = Describe("Mutating DataVolume Webhook", func() {
+	Context("with DataVolume admission review", func() {
+		key, _ := rsa.GenerateKey(rand.Reader, 2048)
+
+		It("should reject review without request", func() {
+			ar := &v1beta1.AdmissionReview{}
+
+			resp := mutateDVs(key, ar, true)
+			Expect(resp.Allowed).To(BeFalse())
+			Expect(resp.Result.Message).ShouldNot(BeEmpty())
```
awels: Do we care about what the message is? And do we know what it should be if it is rejected?

pkg/apiserver/webhooks/handler.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func validateDataVolumeResource(ar v1beta1.AdmissionReview) error {
+	resource := metav1.GroupVersionResource{
+		Group:    cdicorev1alpha1.SchemeGroupVersion.Group,
+		Version:  cdicorev1alpha1.SchemeGroupVersion.Version,
+		Resource: "datavolumes",
+	}
+	if ar.Request.Resource != resource {
+		klog.Errorf("resource is %s but request is: %s", resource, ar.Request.Resource)
```
awels: These are GroupVersionResources, I don't think you can pass them as %s parameters. Should it be %v or resource.Resource as that is the name string.

pkg/apiserver/webhooks/handler.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	// verify the content type is accurate
+	contentType := r.Header.Get("Content-Type")
+	if contentType != "application/json" {
+		klog.Errorf("contentType=%s, expect application/json", contentType)
+		return
+	}
+
+	klog.V(2).Info(fmt.Sprintf("handling request: %s", body))
```
awels: body is a byte[], you need to cast to string before printing.

pkg/apiserver/webhooks/handler.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		toAdmissionResponseError(err)
+	}
+
+	pb, err := json.Marshal(patches)
+	if err != nil {
+		return toAdmissionResponseError(err)
+	}
+
+	klog.V(3).Infof("sending patches\n%s", pb)
```
awels: pb is a byte[] need to cast to string before printing.

pkg/apiserver/apiserver.go
mhenriks: %s is okay for error type but not all types

pkg/apiserver/webhooks/handler.go
mhenriks: GroupVersionResources has a `String()` method

pkg/apiserver/webhooks/handler.go
mhenriks: this is fine too:  https://play.golang.org/p/4Kyup47Lmcg

mhenriks:
FWIW - all tests passed but deleting the test namespace failed for some reason.

awels:
Interesting, so if the struct has a String() you can use %s instead of %v? Sort of like toString in java.

PR #872, opened 2019-07-01
fix cluster-sync for okd

mhenriks:

Creating LocalVolume failing because api version updated

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/ssh.sh
+++ b/cluster-up/ssh.sh
@@ -26,7 +26,7 @@ if [ -z "$node" ]; then
     exit 1
 fi
 
-if [[ $provider_prefix =~ okd.* ]]; then
+if [[ $KUBEVIRT_PROVIDER =~ okd.* ]]; then
     ports=$(${KUBEVIRTCI_PATH}cli.sh --prefix $provider_prefix ports --container-name cluster)
 
     if [[ $node =~ worker-0.* ]]; then
--- a/cluster-sync/okd-4.1.0/create-local-storage-cr.yaml
+++ b/cluster-sync/okd-4.1.0/create-local-storage-cr.yaml
@@ -1,10 +1,10 @@
-apiVersion: "local.storage.openshift.io/v1alpha1"
-kind: "LocalVolume"
+apiVersion: local.storage.openshift.io/v1
+kind: LocalVolume
 metadata:
-  name: "local-disks"
+  name: local-disks
 spec:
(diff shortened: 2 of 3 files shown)


cluster-up/ssh.sh
```diff
@@ -40,7 +40,7 @@ if [[ $provider_prefix =~ okd.* ]]; then
         exit 1
     fi
     shift
-    ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -q -lcore -p $port core@127.0.0.1 -i ${ssh_key} $@
+    ssh -o UserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no -q -lcore -p $port core@127.0.0.1 -i ${ssh_key} $@
```
awels: I have an unpublished PR that uses the latest kubevirtci, and the ssh.sh in there already adds the StrictHostKeyChecking to it. Maybe just wait for that?

mhenriks: Let's see how this run goes had to fix `ssh.sh`

mhenriks:
@awels okd tests running now.  Interestingly, fewer failures than os.

awels:
The failures are due to the cloning issue in os-3.11.

mhenriks:
> The failures are due to the cloning issue in os-3.11.
I know.  My point was that there are fewer failures on okd and that is kind of interesting.

awels:
Yeah seems that whatever broke OS-3.11, didn't break 4.1

awels:
I think #873 includes the contents of this already. And the cluster-up ssh.sh fixes as well.

mhenriks:
> I think #873 includes the contents of this already. And the cluster-up ssh.sh fixes as well.

It does not include the fix to `ssh.sh`

awels:
https://github.com/kubevirt/containerized-data-importer/blob/master/cluster-up/ssh.sh#L43 has the fix, for the ssh already, not sure when that made it in.

mhenriks:
> https://github.com/kubevirt/containerized-data-importer/blob/master/cluster-up/ssh.sh#L43 has the fix, for the ssh already, not sure when that made it in.

No, that is not the fix I've included here.  This line is broken in CI:  https://github.com/kubevirt/containerized-data-importer/blob/master/cluster-up/ssh.sh#L29

PR #881, opened 2019-07-11
Add the option to change the operator name

irosenzw:
Since the operator name is different on
downstream (cnv-cdi-operator), adding the option
to change it when creating the deployment manifest.

This is important for automating the build process
upstream and downstream for the HCO.

Signed-off-by: Ido Rosenzwig <irosenzw@redhat.com>


annastopel:
ci test please

PR #883, opened 2019-07-15
Test commit

j-griffith:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note

```

PR #886, opened 2019-07-16
* make file fix: all target requires manifests 

annastopel:
make docker failswith error, since it expects olm manifests
since olm catalog source support, docker target in Makefile requires manifests to be prepared
Fixes #

since olm catalog source support, docker target in Makefile requires manifests to be prepared
**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ else
 DO=eval
 endif
 
-all: docker
+all: manifests docker
 
 clean:
 	${DO} "./hack/build/build-go.sh clean; rm -rf bin/* _out/* manifests/generated/* .coverprofile release-announcement"


Makefile
```diff
@@ -31,7 +31,7 @@ else
 DO=eval
 endif
 
-all: docker
+all: manifests docker
```
mhenriks: I think it would be better to have the docker target depend on manifests

j-griffith: > I think it would be better to have the docker target depend on manifests

I agree, but oddly enough when I tried that here:  https://github.com/kubevirt/containerized-data-importer/pull/883 it "broke all the things".  Need to look into that and understand why; @annastopel mentioned something about env variables not being set, I didn't investigate yet.

annastopel:
ci test please

Makefile
annastopel: We do want to be able to call make doker and make manifests in separate commands, to make that docker registry trick we have. 

```
QUAY_NAMESPACE=$QUAY_NAMESPACE DOCKER_PREFIX=$MANIFEST_REGISTRY PULL_POLICY=$(getTestPullPolicy) make manifests
DOCKER_PREFIX=$DOCKER_PREFIX make docker push
  
```

annastopel: That is why I intend to set manifests on all target. It is not clear to me though why there is a failure on travis on ``` make goveralls```

annastopel: the error is 
``` Ran 4 of 4 Specs in 0.000 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestClonerTarget (0.00s)
PASS
coverage: 56.0% of statements
ok  	kubevirt.io/containerized-data-importer/pkg/util/prometheus	0.011s	coverage: 56.0% of statements
?   	kubevirt.io/containerized-data-importer/pkg/version	[no test files]
?   	kubevirt.io/containerized-data-importer/pkg/version/verflag	[no test files]
make: *** [goveralls] Error 1
The command "make goveralls" exited with 2.
```

PR #899, opened 2019-07-30
Pass annotations from DV to PVC.

awels:

Pass annotations from Data Volume to PVC.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -939,6 +939,10 @@ func newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume) (*corev1.PersistentV
 
 	annotations := make(map[string]string)
 
+	for k, v := range dataVolume.ObjectMeta.Annotations {
+		annotations[k] = v
+	}
+
 	if dataVolume.Spec.Source.HTTP != nil {
 		annotations[AnnEndpoint] = dataVolume.Spec.Source.HTTP.URL
 		annotations[AnnSource] = SourceHTTP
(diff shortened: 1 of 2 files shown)


awels:
Done

aglitke:
LGTM

pkg/controller/datavolume-controller.go
```diff
@@ -939,6 +939,10 @@ func newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume) (*corev1.PersistentV
 
 	annotations := make(map[string]string)
 
+	for k, v := range dataVolume.ObjectMeta.Annotations {
```
j-griffith: This is fine, but I would suggest at least considering a PVC-Annotations section on the DV.  Passing through is certainly easy and works; but it's not always applicable and at some point you may run the risk of collisions.  I'm good either way, but I do think it should be mentioned.

awels: Agreed, but that would cause us to have to be vendored into kubevirt, this is a fairly minimal change that doesn't require vendoring. And yes the collisions is a concern, but we have some order guarantee, if we collide with annotations created by the controller, those annotations will win.

j-griffith: Not sure about that vs an annotations key, but regardless; I'm fine with this for now.

j-griffith: Ok, I guess we won't wait for CI to run

awels: We did wait, actually, but we have known failures that happened, I verified that those are the only failures.

PR #902, opened 2019-07-30
Initial upgrade support

annastopel:
* - Detect from reconcile loop that it is uograde flow
* - Set ObeservedVersion to target when upgrade is finished
* - Delete unused objects at the end of upgrade

This PR provides initial upgrade support of CDI:
- Detection of upgrade from operator Reconcile loop  - done
- Cleanup of unused resources at the end of upgrade
- Implement unit tests for upgrade flows of controller of cdi operator

Fixes #

This PR provides initial upgrade support of CDI:
- Detection of upgrade from operator Reconcile loop  - done
- Cleanup of unused resources at the end of upgrade
- Implement unit tests for upgrade flows of controller of cdi operator

**Release note**:

```release-note
Initial cdi upgrade detection and handling
```

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -226,6 +226,10 @@ func shouldTakeUpdatePath(targetVersion, currentVersion string) (bool, error) {
 		return false, nil
 	}
 
+	if targetVersion == currentVersion {
+		return false, nil
+	}
+
 	// semver doesn't like the 'v' prefix
 	targetVersion = strings.TrimPrefix(targetVersion, "v")
 	currentVersion = strings.TrimPrefix(currentVersion, "v")
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -156,6 +156,10 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 
 	// mid delete
 	if cr.DeletionTimestamp != nil {
+		if r.isUpgrading(cr) {
+			//CDI is marked to be deleteted while upgrade flow is in process
+			reqLogger.Info("Deleting  CDI during upgrade")
+		}
 		reqLogger.Info("Doing reconcile delete")
 		return r.reconcileDelete(reqLogger, cr)
 	}
@@ -377,6 +381,9 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI) 
 			previousVersion := cr.Status.ObservedVersion
 			cr.Status.ObservedVersion = r.namespacedArgs.DockerTag
 			cr.Status.OperatorVersion = r.namespacedArgs.DockerTag
+
+			//Is there a possible race if cr is marked as deleted during upgrade?
+			//we want to set cr in DeployedPhase, but may be it is being marked as Deleted already
 			if err = r.crUpdate(cdiv1alpha1.CDIPhaseDeployed, cr); err != nil {
 				return reconcile.Result{}, err
 			}
@@ -454,6 +461,7 @@ func (r *ReconcileCDI) isMutable(obj runtime.Object) bool {
 
 // I hate that this function exists, but major refactoring required to make CDI CR the owner of all the things
 func (r *ReconcileCDI) reconcileDelete(logger logr.Logger, cr *cdiv1alpha1.CDI) (reconcile.Result, error) {
+
 	i := -1
 	for j, f := range cr.Finalizers {
 		if f == finalizerName {
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -55,6 +55,8 @@ import (
 
 const (
 	finalizerName = "operator.cdi.kubevirt.io"
+
+	maxTypeCallBacks = 5
 )
 
 var log = logf.Log.WithName("cdi-operator")
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -381,7 +381,7 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI) 
 				return reconcile.Result{}, err
 			}
 
-			logger.Info("Successfully finished upgrade from %v to %v and entered Deployed state", previousVersion, cr.Status.ObservedVersion)
+			logger.Info("Successfully finished Upgrade from %s to %s and entered Deployed state", previousVersion, cr.Status.ObservedVersion)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/controller.go
```diff
@@ -111,6 +116,19 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling CDI")
 
+	//TODO:
+	/*
+		Here for upgrade purpose:
+		* if cr vresion observed and target are not the same upgrade is taking place:
```
mhenriks: maybe a stupid question: where/how is target version set?

pkg/operator/controller/controller.go
```diff
@@ -266,11 +328,259 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI)
...
+		return err
+	}
+
+	lo := &client.ListOptions{}
+	// maybe use different selectors?
+	lo.SetLabelSelector("cdi.kubevirt.io")
+
+	// remove unused CRDs
+	{
+		objs := &extv1beta1.CustomResourceDefinitionList{}
```
mhenriks: What's with the extra nesting here?

pkg/operator/controller/controller.go
```diff
@@ -266,11 +328,259 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI)
...
+	if err != nil {
+		return err
+	}
+
+	lo := &client.ListOptions{}
+	// maybe use different selectors?
+	lo.SetLabelSelector("cdi.kubevirt.io")
+
+	// remove unused CRDs
+	{
```
mhenriks: There's got to be a way to use reflection/type info to reduce LOC here.  

Thinking in could start off like this:

```golang
for _, listType := range []runtime.Object{&corev1.ServiceList{}, &appsv1.DeploymentList{}, ...}

``` 

But need a function that will use reflection to call `.Items` and convert result to []runtime.Object.  Then can call GetName() and GetNamespace() on that.

This vexed me before (dealing with list types in a generic way) and I think this would be a fun thing to figure out.

pkg/operator/controller/controller.go
annastopel: TargetVersion is set initially on crInit  this is in case Get of CRD CR did not find the object
``` func (r *ReconcileCDI) crInit(cr *cdiv1alpha1.CDI) error {
	cr.Finalizers = append(cr.Finalizers, finalizerName)
	cr.Status.OperatorVersion = r.namespacedArgs.DockerTag
	cr.Status.TargetVersion = r.namespacedArgs.DockerTag
	return r.crUpdate(cdiv1alpha1.CDIPhaseDeploying, cr)
}
``` 
In case CT was found - in Reconcile method we check whether it is upgrade by comparing cr.ObeservedVersion to DockerTag environmental variable. If they are not the same - cr is updated with cr.TargetVersion = DockerTag and from that point isUpgrading method of reconcile object will return true.
When current operator is deploied and ready, cr.Observed is set to cr.Target

pkg/operator/controller/controller.go
annastopel: yeah, the scope to perform easy copy paste and to keep all the loops the same. I was hoping to find out later a way to make a method out of this loop

pkg/operator/controller/controller.go
mhenriks: This code can be simplified quite a bit see here for inspiration:  https://github.com/kubevirt/containerized-data-importer/blob/24741566f3d8215356e8d9d8860f09a0373598f7/pkg/operator/controller/callbacks.go#L241-L268

pkg/operator/api.go
```diff
@@ -30,6 +30,18 @@ const (
...
+	//ManagedByLabelOperatorValue This label will be set on all resources created by the operator
+	ManagedByLabelOperatorValue = "cdi-operator"
+	//InstallStrategyVersionAnnotation This annotation represents the kubevirt version for an install strategy configmap.
+	InstallStrategyVersionAnnotation = "cdi.kubevirt.io/install-strategy-version"
+	//InstallStrategyRegistryAnnotation This annotation represents the kubevirt registry used for an install strategy configmap.
+	InstallStrategyRegistryAnnotation = "cdi.kubevirt.io/install-strategy-registry"
+	//EphemeralBackupObject This annotation represents that this object is for temporary use during updates
+	EphemeralBackupObject = "cdi.kubevirt.io/ephemeral-backup-object"
+	//InstallStrategyLabel This label indicates the object is a part of the install strategy retrieval process.
+	InstallStrategyLabel = "cdi.kubevirt.io/install-strategy"
```
mhenriks: None of these referenced anywhere yet.  I possible see a use for `ManagedByLabel` and `ManagedByLabelOperatorValue` but do we need the rest?

pkg/operator/controller/callbacks.go
```diff
@@ -0,0 +1,209 @@
+/*
```
mhenriks: I assume all this will be replaced by #911?

pkg/operator/controller/controller.go
```diff
@@ -125,6 +151,24 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 		return reconcile.Result{}, err
 	}
 
+	//compare target and observed
+	//Retriveed CR will contain previous version ImageTag and ImageRegistry
+	//while namespaced arguments will contain new version
+	//if versions are not the same: update TargetVersion in CDI cr and move to implementation
+	isUpgrade, err := shouldTakeUpdatePath(r.namespacedArgs.DockerTag, cr.Status.ObservedVersion)
```
mhenriks: maybe check if we are upgrading after we check `cr.DeletionTimestamp`?

pkg/operator/controller/controller_test.go
```diff
@@ -520,6 +1222,34 @@ func getDeployment(client realClient.Client, deployment *appsv1.Deployment) (*ap
 	return result.(*appsv1.Deployment), nil
 }
 
+func equalObjects(a, b runtime.Object) (bool, error) {
```
mhenriks: is this used/necessary?  what's wrong with `reflect.DeepEqual`?

pkg/operator/controller/controller_test.go
```diff
@@ -534,6 +1264,46 @@ func getObject(client realClient.Client, obj runtime.Object) (runtime.Object, er
...
 
+func updateObject(client realClient.Client, obj runtime.Object) error {
+	if err := client.Update(context.TODO(), obj); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getObjectFromResources(reconciler *ReconcileCDI, obj runtime.Object) (runtime.Object, error) {
```
mhenriks: can't use `ReconcileCDI.getAllResources` for part of this?

tools/util/marshaller.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/json"
+	"io"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// MarshallObject marshalls an object to yaml appropriate for kubectl
+func MarshallObject(obj interface{}, writer io.Writer) error {
```
mhenriks: where will this be used?

pkg/operator/controller/controller.go
annastopel: fixed

pkg/operator/controller/controller.go
annastopel: fixed

pkg/operator/api.go
annastopel: removed

pkg/operator/controller/controller_test.go
annastopel: nothing wrong (was not aware of that) - fixed

pkg/operator/controller/controller_test.go
annastopel: removed - not needed

tools/util/marshaller.go
annastopel: was required for installStrategy. Now removed, since this approach was not taken

pkg/operator/controller/controller.go
annastopel: ok, I switched between Deletion check and upgrade, but this is racy

mhenriks: I think we have to handle delete during upgrade.  The possible issue I see is that we have to make sure to delete orphaned resources from N-1 install.

pkg/operator/controller/controller.go
```diff
@@ -302,11 +374,83 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI)
...
+
+		sv := reflect.ValueOf(lt).Elem()
+		iv := sv.FieldByName("Items")
+
+		for i := 0; i < iv.Len(); i++ {
+			obj := iv.Index(i).Addr().Interface().(runtime.Object)
+			found := false
+			for _, target := range targetStrategy {
+				if reflect.TypeOf(obj) == reflect.TypeOf(target) {
+					if obj.(metav1.Object).GetName() == target.(metav1.Object).GetName() {
```
mhenriks: Should we also check namespace just to be safe?

pkg/operator/controller/controller.go
```diff
@@ -302,11 +374,83 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI)
...
+						found = true
+						break
+					}
+				}
+			}
+			if !found {
+				logger.Info("Deleting  ", "type", reflect.TypeOf(obj), "Name", obj.(metav1.Object).GetName())
+				if err = r.client.Delete(context.TODO(), obj); err != nil {
```
mhenriks: Have to invoke callbacks here, state is `ReconcileStatePostDelete` desiredObject is `nil` and currentObject is `obj`

pkg/operator/controller/controller.go
annastopel: @mhenriks added logic that call lcleanupUnusedResources if DeletionTimestamp is detectd during upgrade

pkg/operator/controller/controller.go
annastopel: done

pkg/operator/controller/controller.go
annastopel: done

annastopel:
ci test please

annastopel:
ci test please

mhenriks:
ci test please

pkg/operator/controller/controller.go
```diff
@@ -50,6 +55,8 @@ import (
 
 const (
 	finalizerName = "operator.cdi.kubevirt.io"
+
+	maxTypeCallBacks = 5
```
mhenriks: What is this?

pkg/operator/controller/controller.go
```diff
@@ -136,10 +156,38 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 	// mid delete
 	if cr.DeletionTimestamp != nil {
+		if r.isUpgrading(cr) {
+			//CDI is marked to be deleteted while upgrade flow is in process
+			reqLogger.Info("Deleting  CDI during upgrade")
+			//clenaup all resources of previous installation first
+			err := r.cleanupUnusedResources(reqLogger, cr)
+			if err != nil {
+				reqLogger.Info("Failed to cleanupUnused resource prior to CDI cr deletion during upgrade")
+				return reconcile.Result{}, nil
```
mhenriks: Should be `return reconcile.Result{}, err`?

pkg/operator/controller/controller.go
```diff
@@ -136,10 +156,38 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 
 	// mid delete
 	if cr.DeletionTimestamp != nil {
+		if r.isUpgrading(cr) {
+			//CDI is marked to be deleteted while upgrade flow is in process
+			reqLogger.Info("Deleting  CDI during upgrade")
+			//clenaup all resources of previous installation first
+			err := r.cleanupUnusedResources(reqLogger, cr)
```
mhenriks: hmmm, actually is this necessary if the resources are owned by the CDI CR?

mhenriks:
ci test please

pkg/operator/controller/callbacks.go
annastopel: yes

pkg/operator/controller/controller.go
annastopel: invokeDeleteCDICallbacks is based on getAllResources, that is if resource is not in the list (resource of N-1 version) then it is not deleted.
cleanupUnusedResources iterates over all resource that do not belong to the list returned bu getAllResources method

pkg/operator/controller/controller.go
annastopel: of course, fixed

pkg/operator/controller/controller.go
annastopel: leftover - removed

pkg/operator/controller/controller.go
mhenriks: If an N-1 resource no longer exists in version N but it is still owned by the CDI CR the resource should get deleted when the CDI CR is deleted.

mhenriks:
continuing in #929

pkg/operator/controller/controller.go
bmozaffa: This is a common pattern and looks good.

pkg/operator/controller/controller.go
bmozaffa: Agree, resources with owner references will be cleaned up when the owner is removed.

pkg/operator/controller/controller.go
bmozaffa: What type of race condition are you concerned about?

PR #903, opened 2019-07-31
Adding webhook validation for length of DataVolume name

### LINKED ISSUE #895: Verify datavolumes names are no longer than 63 characters

>
> /kind bug
/kind enhancement

**What happened**:

A datavolume with a too-long name was created, resulting in the controller erroring with a 
```
E0729 09:01:47.337668       1 datavolume-controller.go:299] error syncing 'kube-system/ubuntu-1-14-0-kubevirt-1154768504474308608-745745b667-w6ktw-data-volume-1': PersistentVolumeClaim "ubuntu-1-14-0-kubevirt-1154768504474308608-745745b667-w6ktw-data-volume-1" is invalid: metadata.labels: Invalid value: "ubuntu-1-14-0-kubevirt-1154768504474308608-745745b667-w6ktw-data-volume-1": must be no more than 63 character
```

Would be nice if we catch that during validation 

**Environment**:
- Install tools:
- Others:

zvikorn:

Fix #895 
Fixes #895 

**Release note**:

```release-note
Adding webhook validation for length of DataVolume name
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-validate.go
+++ b/pkg/apiserver/webhooks/datavolume-validate.go
@@ -54,6 +54,19 @@ func validateSourceURL(sourceURL string) string {
 	return ""
 }
 
+func validateDataVolumeName(name string) []metav1.StatusCause {
+	var causes []metav1.StatusCause
+	// name of data volume cannot be more than 55 characters (not including '-scratch')
+	if len(name) > 55 {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("Name of data volume cannot be more than 55 characters"),
+			Field:   "",
+		})
+	}
+	return causes
+}
+
 func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *v1beta1.AdmissionRequest, field *k8sfield.Path, spec *cdicorev1alpha1.DataVolumeSpec) []metav1.StatusCause {
 	var causes []metav1.StatusCause
 	var url string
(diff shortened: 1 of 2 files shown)


pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -54,6 +54,19 @@ func validateSourceURL(sourceURL string) string {
 	return ""
 }
 
+func (wh *dataVolumeValidatingWebhook) validateDataVolumeName(request *v1beta1.AdmissionRequest, name string) []metav1.StatusCause {
```
awels: There is no reason to make this function part of the struct, it doesn't use the struct in any way. validateDataVolumeSpec does use the wh struct, which is why it is part of it.

zvikorn: you are right. Fixing that.

awels:
lgtm

mhenriks:
How about some tests?

Also, I think the name has to be shorter to accommodate scratch pvc naming.

awels:
I am thinking maybe I need to change the scratch space naming convention, and just make it a random value or something. To avoid the length issue.

zvikorn:
@awels isn't the code I added to pkg/apiserver/webhooks/datavolume-validate_test.go sufficient?

awels:
@zvikorn yes it is, I missed you added it already.

zvikorn:
Do you want me to restrict it to less than 57 (scratch)?

awels:
Actually we add '-scratch' which is 8 characters, so it would be less than 63-8=55.

PR #908, opened 2019-08-04
Requiring replica 2 for CDI deployment

rollandf:
In order to increase CDI availability in case of node failure, the number of replicas for cdi-deployment, cdi-apiserver, cdi-uploadproxy is updated to "2".

```release-note
NONE
```

### DIFF
--- a/pkg/operator/resources/namespaced/apiserver.go
+++ b/pkg/operator/resources/namespaced/apiserver.go
@@ -127,7 +127,7 @@ func createAPIServerService() *corev1.Service {
 }
 
 func createAPIServerDeployment(repo, image, tag, verbosity, pullPolicy string) *appsv1.Deployment {
-	deployment := utils.CreateDeployment(apiServerRessouceName, cdiLabel, apiServerRessouceName, apiServerRessouceName, 1)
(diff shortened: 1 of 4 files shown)


awels:
Looks like tons of functional tests are broken here.

rollandf:
Yep, I need to see how to handle it.
There is a check that only one instance of the pods should be up in the tests.

awels:
@rollandf I think there needs to be a rebase on this and something to fix the broken tests?

maya-r:
I tried to rebase this patch.

FindPodByPrefix will error if there's more than one match, but we can adapt it to return true for multiple matches.  
But after that, I run into failures in tests/leaderelection_test.go and I'm not sure if that's a problem with the test or with the behaviour.

awels:
So you can't use find pod by prefix if we have multiple controller pods to find those pods, exactly for the reason you stated. At the same time the leader election test assumes there is only one, it then creates another pod, kills the original one and checks if the new one is now the leader. If we have 2 pods to start with, this makes the test somewhat easier. We just need to find both pods, kill the active one, and see if the stand by becomes the leader.

PR #910, opened 2019-08-06
Added conditions to match the HCO requirements.

awels:

Implement HCO Conditions based on https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md

Fixes #

**Release note**:

```release-note
Implemented HCO Conditions in CR object
```


pkg/operator/controller/controller.go
```diff
@@ -393,21 +406,45 @@ func (r *ReconcileCDI) checkReady(logger logr.Logger, cr *cdiv1alpha1.CDI) (bool
...
 		}
-
 	}
 
-	logger.Info("CDI Ready check", "Status", readyCond.Status)
+	logger.Info("CDI degraded check ", "Degraded", degraded)
 
-	if err = r.conditionUpdate(readyCond, cr); err != nil {
-		return false, err
+	if degraded {
```
mhenriks: can this block be cleaned up somehow?  try to minimize the calls to `r.crUpdate`

pkg/operator/controller/util.go
```diff
@@ -46,3 +50,155 @@ func mergeLabelsAndAnnotations(src, dest metav1.Object) {
 func deployClusterResources() bool {
 	return strings.ToLower(os.Getenv("DEPLOY_CLUSTER_RESOURCES")) != "false"
 }
+
+// MarkCrHealthyMessage marks the passed in CR as healthy. The CR object needs to be updated by the caller afterwards.
```
mhenriks: I think this stuff belongs in cr.go because conditions are part of the cr.  Or make a conditions.go.

pkg/operator/controller/util.go
```diff
@@ -46,3 +50,155 @@ func mergeLabelsAndAnnotations(src, dest metav1.Object) {
...
+		Status: corev1.ConditionFalse,
+	})
+}
+
+// MarkCrHealthy marks the passed in CR as healthy. The CR object needs to be updated by the caller afterwards.
+// Healthy means the following status conditions are set:
+// ApplicationAvailable: true
+// Progressing: false
+// Degraded: false
+func MarkCrHealthy(cr *cdiv1alpha1.CDI) {
```
mhenriks: why not have this call `MarkCrHealthyMessage`?

pkg/apis/core/v1alpha1/types.go
```diff
@@ -204,11 +206,11 @@ type CDIPhase string
 
 // CDIStatus defines the status of the CDI installation
 type CDIStatus struct {
-	Phase           CDIPhase       `json:"phase,omitempty"`
-	Conditions      []CDICondition `json:"conditions,omitempty" optional:"true"`
```
mhenriks: Changing the type of a field is something to avoid when possible.  The right thing to do when necessary is to bump the version of the CRD.  Are we not at all concerned about instances of this CR in the wild?

Also, Kubevirt did not migrate to this new Conditions struct/package.  Why are we? `CDICondition` also pretty much matches `KubeVirtCondition`.

awels: I pulled all of it from here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md Not sure why KubeVirt decided not to follow that.

awels: As for the type, https://github.com/openshift/custom-resource-status/blob/master/conditions/README.md describes what it should be.

pkg/operator/controller/util.go
awels: I couldn't figure out how to make it update the status but not overwrite the message any other way.

pkg/operator/controller/util.go
awels: Sure, don't care that much where it goes, if you prefer cr.go, that is where I will put it. Or a conditions.go doesn't matter much.

pkg/operator/controller/controller.go
awels: Just making sure I understand what you are asking here. You want me to make sure I try not to call r.crUpdate if no status/conditions actually changed?

mhenriks: No, there are three places in that block that do this:
```golang
if err = r.crUpdate(<some phase value>, cr); err != nil {
	return degraded, err
}
```
Can that be reduced?  If not, can the code be extracted to a helper function?

pkg/operator/controller/util.go
mhenriks: The way I read the code, the reason/message will be set to `""` wheen `MarkCrHealthy` is called.  But maybe I'm wrong.

pkg/apis/core/v1alpha1/types.go
mhenriks: If we have to switch condition types it may be worth the exercise of bumping the CRD version.  I'm wondering what would happen if the operator is updated when old CR exists.  At a minimum, the `LastProbeTime` of `CDICondition` will be lost which is probably not the end of the world but I'm not sure what else would happen.

annastopel:
When crd is updated and version is advanced via crd versions, api server
allows to access the existing cr via both apis. When object is updated it
will be stored under new api

annastopel:
Operator can update the cr to match  new api. As another option - a
dedicated conversion webhook can be implemented to convert the existing crs.

mhenriks:
@annastopel I was wondering what would happen if we do not bump the CRD version

annastopel:
I did not try it but from what I understood from kubevirt (who tried that)
it is possible to update spec (add fields) without modifying the version.
On get we will get golang default values. We can set them and they are
stored

annastopel:
However, not sure what happens when a field is deleted

mhenriks:
@annastopel, yes you can add fields safely to a CR without bumping the version.  But we are changing the type of the field can be complicated.  In this case , the serialized version of the resource is largely compatible.  The differences being `LastProbeTime` vs `LastHeartbeatTime` and the different values for `Type`.  Again, probably not the biggest deal in the world.  But I wonder if it is worth the time to do the right thing here.

PR #911, opened 2019-08-06
Operator refactoring to support callback model

mhenriks:

Make the operator reconcile loop more generic with callbacks.  Add SCC update/route creation to callbacks.  Also refactor code that cleans up after CDI is deleted to callbacks.

Fixes #

Somewhat related and could be considered a prereq for #910 and #902

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/controller/callbacks.go
+++ b/pkg/operator/controller/callbacks.go
@@ -17,12 +17,26 @@ limitations under the License.
 package controller
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"reflect"
+
 	"github.com/go-logr/logr"
 
+	secv1 "github.com/openshift/api/security/v1"
+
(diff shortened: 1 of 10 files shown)


pkg/operator/controller/callbacks.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// ReconcileStateCDIDelete is called during CDI finalizer
+	ReconcileStateCDIDelete ReconcileState = "CDI_DELETE"
+)
+
+// ReconcileState is the current state of the reconcile for a particuar resource
+type ReconcileState string
+
+// ReconcileCallbackArgs contains the data of a ReconcileCallback
+type ReconcileCallbackArgs struct {
```
mhenriks: @annastopel you may want to add an `IsUpgrading` member here in the future

PR #912, opened 2019-08-07
func-tests: support hostpath-provisioner

danielerez:
Added support for hostpath-provisioner in functional tests:
* Introduced 'AddProvisionOnNodeToAnn' func for adding
  'kubevirt.io/provisionOnNode' annotaion when hostpath-provisioner
  is the default storage class.
* pvc -> NewPVCDefinition and datavolume -> CreateDataVolumeFromDefinition:
  invoke 'AddProvisionOnNodeToAnn' when needed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/test.go
+++ b/tests/test.go
@@ -18,3 +18,13 @@
  */
 
 package tests
+
+import (
+	"kubevirt.io/containerized-data-importer/tests/framework"
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+func init() {
+	f := framework.NewFrameworkOrDie("tests")
+	utils.CacheTestsData(f.K8sClient)
+}
--- a/tests/utils/common.go
+++ b/tests/utils/common.go
@@ -1,5 +1,12 @@
 package utils
 
+import (
+	"github.com/onsi/ginkgo"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
 // cdi-file-host pod/service relative values
 const (
 	//RegistryHostName provides a deploymnet and service name for registry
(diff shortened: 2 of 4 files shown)


aglitke:
@awels another option would be to use something like snapshotSC and blockSC to configure a storage class that wants DV node annotations.

awels:
@aglitke I don't think we care, do we? If we just slap the annotation on there in the test, if we are running this on a cluster that has the provisioner, it will work, and if we run it on a cluster that doesn't it will still work since it gets ignored. The hardest part at that point will be deciding on the node name (either detect, or pass in)

aglitke:
@awels yeah that's fine albeit a bit messy

j-griffith:
I see the logic in how you approached this, and it makes sense, but checking every time we create a pvc def and then calling the client every time we want to use our hostpath driver seems like it's going to make things brittle.

The node problem is easy to solve, we could just set that up in the init section of each test; I think we could also use that same strategy to identify the storage class and set up annotations once as part of a test init.  Another option is you could possibly use the framework init function to do this.  What would be nice about that is you only have to add it to the framework and you gain the info you need in any/all functional tests.  What do you think?

awels:
@aglitke I agree it is a bit messy, but since this is for the tests and not a general approach, I think it would be fine. But that is why I asked for opinions as I am not married to the concept. And I think @j-griffith wrote basically the same thing I did but longer.

j-griffith:
> Two suggestions that will hopefully make this a lot smaller.
> 
> 1. You don't need to look up the nodes each time, the nodes will not change during a run, so run the look up once at startup, and store the value you need somewhere that is accessible everywhere. Maybe even make it a test parameter and if not specified set it.

Yeah, I think we could do this on init in the framework and make things pretty easy.

> 2. This one is a little dirty, but there is no reason not to simply always put that annotation on the DV/PVC in the tests, if it is running with another provisioner, that annotation is simply ignored and thus there is no harm in having it.

yeah, kinda ugly to do that; initial glance I think this could be handled in the framework as well; determine things like nodes, storage-classes etc and then just provide a var for the provision annotation.

We'd still need special checks (or a special test) to check things like the annotation itself, but it would make this more flexible.

j-griffith:
> @awels another option would be to use something like snapshotSC and blockSC to configure a storage class that wants DV node annotations.

Yeah, that's kinda what an init model could do; I wouldn't do it explicitly like we do with snapshot and block currently (I personally wish we did that differently), but the framework init could basically determine this all in an automated fashion based on the deployment so it wouldn't require any special config or deploy steps.

danielerez:
As suggested by @awels and @j-griffith, simplified by fetching the data once and caching in a common location.

aglitke:
Simple enough, looks good to me.

PR #915, opened 2019-08-08
Fallback to Fedora29 minimal for images

### LINKED ISSUE #914: qemu-img 3.1 failing when trying to do stream conversions of cirros image
/kind bug

**What happened**:

When attempting to use the CDI importer to create a DataVolume (or a PVC directly) and specifying a cirros image, the conversion process hangs and never completes.

The conversion process should either finish or fail, terminating the importer pod and providing me with a PVC that has the expected image on it (upon success).

Create a DV using the follwing yaml:

```
kind: DataVolume
metadata:
  name: "fedora-datavol"
  annotations:
    kubevirt.io/provisionOnNode: kube-1
spec:
  source:
      http:
         url: "http://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img"
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: "1Gi"
```
Then tail the logs on the importer pod and note the hang

j-griffith:
The newer version of qemu-img (3.1.0) that ships in Fedora 30 has issues
with streaming conversions whereby it hangs during the conversion
process.  Rather than try to force a downgrade of qemu when building
images instead, let's try reverting back to Fedora29 minimal which has
qemu-img version  3.0.0.

This issue has also exposed the fact that we aren't performing any
testing against streaming external images, we're only testing against
images that we've pulled in to our test infrastructure.  So add a
functional test that builds a DataVolume directly from the cirros site.

It's been discovered that qemu-img version 3.1.0 which ships in Fedora-30 hangs when trying to do a streaming conversion of the external cirros image.  It also turns out that we weren't detecting this via our test automation because we're running in an air gapped env with our own VM images.

Fixes #914 

**Release note**:

```release-note
Downgrade base images from Fedora30 to Fedora29
```

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora-minimal:30
+FROM registry.fedoraproject.org/fedora-minimal:29
 MAINTAINER "The KubeVirt Project" <kubevirt-dev@googlegroups.com>
 
 COPY fedora.repo /tmp/fedora_ci.dnf.repo
--- a/hack/build/docker/builder/fedora.repo
+++ b/hack/build/docker/builder/fedora.repo
@@ -1,4 +1,4 @@
-[fedora-base-fc30]
+[fedora-base-fc29]
 clean_requirements_on_remove=True
 tsflags=nodocs
 name=Fedora $releasever - $basearch
@@ -13,7 +13,7 @@ gpgcheck=1
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
 skip_if_unavailable=False
 
-[fedora-updates-fc30]
+[fedora-updates-fc29]
 clean_requirements_on_remove=True
 tsflags=nodocs
 name=Fedora $releasever - $basearch - Updates
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -41,6 +41,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 	tinyCoreIsoRegistryURL := fmt.Sprintf(utils.TinyCoreIsoRegistryURL, f.CdiInstallNs)
 	tarArchiveURL := fmt.Sprintf(utils.TarArchiveURL, f.CdiInstallNs)
 	InvalidQcowImagesURL := fmt.Sprintf(utils.InvalidQcowImagesURL, f.CdiInstallNs)
+	cirrosURL := fmt.Sprintf(utils.CirrosURL, f.CdiInstallNs)
 
 	// Invalid (malicious) QCOW images:
 	// An image that causes qemu-img to allocate 152T (original image is 516 bytes)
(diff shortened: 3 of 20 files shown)


awels:
Do we really have to downgrade fedora to 29 for all the containers? The only one that really needs it is the importer container. The rest should be fine.

j-griffith:
> Do we really have to downgrade fedora to 29 for all the containers? The only one that really needs it is the importer container. The rest should be fine.

Well, personally I like consistency and I didn't see any downside to downgrading.  If there's some advantage to using F30 that I'm not thinking of I'm fine either way.  I just wanted to be as consistent and vanilla as possible and NOT have to manage another corner case going forward, we have enough of those already.

tests/utils/datavolume.go
```diff
@@ -36,6 +36,8 @@ const (
 	InvalidQcowImagesURL = "http://cdi-file-host.%s/invalid_qcow_images/"
 	// TarArchiveURL provides a test url for a tar achive file
 	TarArchiveURL = "http://cdi-file-host.%s/archive.tar"
+	// CirrosExternalURL provides the standard cirros image download url external from our test env
+	CirrosExternalURL = "http://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img"
```
awels: I don't think I want to depend on an external source. from CDIs perspective there is no difference in going to the internet and going to an http server hosted in the cluster (besides maybe which network to go on). We already have the cirros image in the repo tests/images/cirros.qcow2 we just need to copy it into the http container in the right directory. /tmp/shared/images

j-griffith: No problem, I can change that if you think it's a safe assumption.

tests/datavolume_test.go
```diff
@@ -41,6 +41,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 	tinyCoreIsoRegistryURL := fmt.Sprintf(utils.TinyCoreIsoRegistryURL, f.CdiInstallNs)
 	tarArchiveURL := fmt.Sprintf(utils.TarArchiveURL, f.CdiInstallNs)
 	InvalidQcowImagesURL := fmt.Sprintf(utils.InvalidQcowImagesURL, f.CdiInstallNs)
+	externalCirrosURL := utils.InvalidQcowImagesURL
```
awels: Should this not be: CirrosExternalURL

j-griffith: Yes, it should; just noticed that in the failure thanks

aglitke:
@j-griffith Have you created a bug against either qemu or fedora for the root cause?  We can hold back on Fedora for a little while but we should make sure to get the underlying issue resolved.

mhenriks:
@aglitke the issue does not happen with qemu-img 4.0.0 I build locally

aglitke:
Ok thanks @mhenriks .

hack/build/docker/builder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora-minimal:30
+FROM registry.fedoraproject.org/fedora-minimal:29
```
aglitke: Can we find a good place to comment why we are stuck on 29 so that we will remember later and can know when to upgarade to 31?

awels: We are adding a functional test that imports the image that is giving us trouble. During the course of normal development we will attempt to upgrade the images, and if grab a version that is broken it should fail that test.

hack/config-provider-external.sh
```diff
@@ -0,0 +1,4 @@
+docker_tag=devel
```
awels: I don't think we want this in here.

test-pod.yaml
```diff
@@ -0,0 +1,11 @@
+kind: Pod
```
awels: I don't think this is needed is it?

awels:
You also need to copy the file from the 'out' into the container when building. Add this to the Dockerfile of cdi-func-test-file-host-init
COPY cirros.qcow2 /tmp/source/cirros-qcow2.img

test-pod.yaml
j-griffith: No, it's not

hack/config-provider-external.sh
j-griffith: Nope, you don't

j-griffith:
> Also need to edit tools/cdi-func-test-file-host-init/main.go to copy from /tmp/source to the actual target. Add something like this:
> 

@awels Ahh, thanks!  I think that's the thing that I was missing.

awels:
@j-griffith there are some linting errors in here.

hack/build/docker/builder/fedora.repo
```diff
@@ -1,4 +1,4 @@
-[fedora-base-fc30]
+[fedora-base-fc29]
```
awels: Need to do this for automation/check-patch.yumrepos as well. That is a symlink to all the other fedora.repos that are using in the dockerfiles.

j-griffith: I have that here no?  https://github.com/kubevirt/containerized-data-importer/pull/915/commits/52ad4935b02fb2a3f41d8aea41ce06cb4ffb0394#diff-42d85b9ddb8c0d2cec8713a2713f107cR1  Is there another hidden tgem

j-griffith:
@awels Hmm... wonder if this fixes that too large test?  Causing it to now pass and thus fail?

awels:
There that should work.

awels:
ci test please

awels:
ci test please

awels:
ci test please

awels:
Trying one more time, but at least the k8s run worked last time.

j-griffith:
ci test please

awels:
ci test please

awels:
ci test please

j-griffith:
ci test please

awels:
ci test please

PR #916, opened 2019-08-09
Update kubevirtci k8s version to 1.15.1

awels:

Updated the kubevirtci k8s version to 1.15.1
Updated OKD image to latest available from: https://github.com/kubevirt/kubevirtci/pull/126 once that is merged we need to update our usage of cluster-up to match.

Fixes #

**Release note**:

```release-note
NONE
```


mhenriks:
What version is kubevirt at?  Since they vendor in our clients, code has to be compatible.   Did you run `make generate` and there were no changes?

Also, I think it would be a good idea to update `openapi-gen` in the builder Dockerfile

awels:
All of this is just the kubevirtci image versions, none of the code actually changed

awels:
ci test please

awels:
ci test please

awels:
ci test please

mhenriks:
If we can only test one k8s version, I think it should be 1.13.  If we can do both, that's great.

awels:
We can do how ever many we want, we can even do two separate k8s-1.13.3 if we want (one with ceph and one without for instance). The problem is, that will simply increase the amount of time we are running tests and increases the chance that a flaky test will result in a test failure.

I am trying to understand what the problem is with k8s 1.15.1, since 1.13 is getting old at this point.

awels:
Lets discuss the k8s-1.13 need further in a different PR.

j-griffith:
I'm good with jumping to 1.15; adding additional versions seems like a great idea too.  I'm just thrilled to see a clean-green output!  I'll hold off on merging until others get a chance to discuss the 1.13 question.

j-griffith:
Let's merge this and add in other versions so we can get things moving along again.  Thanks @awels !

PR #917, opened 2019-08-09
Downgrade to 4.1.0 version, 4.2.0 is broken.

awels:

Set testing olm catalog container version to 4.1.0.

Fixes #

**Release note**:

```release-note
NONE
```


j-griffith:
Just to clarify, I'm not sure 4.2 is "broken" I suspect includes new OLM requirements that we're not compatible with yet.  Regardless, I suspect we should definitely be targeting the same version that we're deploying :)

/lgtm

awels:
Lets see how CI likes it. Well its broken for 'us'

PR #919, opened 2019-08-12
* bug fix - make OLM manifests be set per provider

annastopel:
* when installing via OLM, for all providers k8s manifests were used
* Now each provider when chose to install via OLM should specify path for catalog-source manifest and for sunsription manifest via variables:
- CDI_OLM_MANIFESTS_CATALOG_SRC
- CDI_OLM_MANIFESTS_SUBSCRIPTION

Fixes #

* relevant for ci
* when installing via OLM, for all providers k8s manifests were used
* Now each provider when chose to install via OLM should specify path for catalog-source manifest and for sunsription manifest via variables:
- CDI_OLM_MANIFESTS_CATALOG_SRC
- CDI_OLM_MANIFESTS_SUBSCRIPTION

**Release note**:

```release-note
None
```


cluster-up/hack/common.sh
```diff
@@ -17,6 +17,8 @@ fi
 
 CDI_INSTALL_OPERATOR="install-operator"
 CDI_INSTALL_OLM="install-olm"
+CDI_OLM_MANIFESTS_CATALOG_SRC=""
```
awels: We really can't change anything in the 'cluster-up' directory as that is wiped and pulled when you we do a make generate.

mhenriks: Oh good call.

awels: We currently have 4 fields in there already that need to go.

annastopel: @awels good call indeed! I moved all the variables to cluster-sync/install-config.sh. Please take  a look

awels:
coveralls is currently down which is what is causing the failure on travis.

PR #927, opened 2019-08-19
added CRD open API schema validation and unit tests

redhatHameed:
Signed-off-by: Abdul Hameed <ahameed@redhat.com>

CDI operator CRD validation was missing, added schema validation and unit tests to validate CRD and CR file

Fixes #

I added new go file for unit tests, under pkg/operator/resources/operator/operator_test.go, maybe we can move this file somewhere under, tests folder, but function ```createCDIListCRD``` is private, as it begins with a lowercase, so I create a unit test in that same directory.

**Release note**:

```release-note
NONE
```


ovirt-infra:

2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically. ( keep in mind this list might be overwritten if the job XML is refreshed, for permanent whitelisting, please follow #3 option )

awels:
Looks like some formatting issues. If you could run gofmt -w pkg/operator/resources/operator/operator.go pkg/operator/resources/operator/operator_test.go to fix. You can run make test-lint to check if it got fixed. Also make test-unit will run the unit tests.

redhatHameed:
> Looks like some formatting issues. If you could run gofmt -w pkg/operator/resources/operator/operator.go pkg/operator/resources/operator/operator_test.go to fix. You can run make test-lint to check if it got fixed. Also make test-unit will run the unit tests.

@awels updated and ```make test-lint``` and ```make test-unit``` running successfully. additional vendor packages files also included in a commit.

mhenriks:
Doesn't appear that vendored files were added correctly.  Run `make deps-update`.

redhatHameed:
> Doesn't appear that vendored files were added correctly. Run `make deps-update`.

updated now.

pkg/operator/resources/operator/operator.go
```diff
@@ -528,6 +527,40 @@ func createCDIListCRD() *extv1beta1.CustomResourceDefinition {
...
+								},
+								"imagePullPolicy": {
+									Type: "string",
+								},
+							},
+							Type: "object",
+						},
+					},
+					Required: []string{
+						"spec",
```
mhenriks: Not sure spec should be required if all spec params are optional

pkg/operator/resources/operator/operator_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"strings"
+	"testing"
+)
+
+var crdTypeMap = map[string]interface{}{
+	OperatorCdiCRD: &cdiv1alpha1.CDI{},
+}
+
+func TestCRDSchemas(t *testing.T) {
```
mhenriks: ginkgo/gomega tests would be preferred

pkg/operator/resources/operator/operator_test.go
```diff
@@ -0,0 +1,62 @@
...
+				assert.Fail(t, "Discrepancy between CRD and Struct",
+					"Missing or incorrect schema validation at %v, expected type %v  in CRD file %v", missing.Path, missing.Type, crdFileName)
+			}
+		}
+	}
+}
+
+func TestSampleCustomResources(t *testing.T) {
+
+	var crFileName, crdFileName string = "cdi-cr.yaml", OperatorCdiCRD
```
mhenriks: How about some negative tests?  And rather than storing a bunch of manifests in files, how about creating a bunch of CDIs in code and serializing them to yaml?

mhenriks: Oh yeah CDIs in code prob won’t work for negative test cases would have to use something like k8s unstructured or raw strings

pkg/operator/resources/operator/operator_test.go
```diff
@@ -0,0 +1,62 @@
+package operator
+
+import (
+	"fmt"
+	"github.com/RHsyseng/operator-utils/pkg/validation"
```
mhenriks: Since this is explicitly imported, may want to explicitly add to glide.yaml with version info

pkg/operator/resources/operator/operator.go
```diff
@@ -528,6 +527,40 @@ func createCDIListCRD() *extv1beta1.CustomResourceDefinition {
...
+						"spec": {
+							Properties: map[string]extv1beta1.JSONSchemaProps{
+								"imageRegistry": {
+									Type: "string",
+								},
+								"imageTag": {
+									Type: "string",
+								},
+								"imagePullPolicy": {
+									Type: "string",
```
mhenriks: Is it possible to limit to valid values for k8s image pull policy (Always, IfNotPresent, Never)?

pkg/operator/resources/operator/operator.go
redhatHameed: > Not sure spec should be required if all spec params are optional

ImagePullPolicy param mentioned as required in types.go  https://github.com/kubevirt/containerized-data-importer/blob/b3bf364b238217750be5bd0d1c499630fbfb751b/pkg/apis/core/v1alpha1/types.go#L199

So we can mark ImagePullPolicy as required in CRD validation section and also I think Spec is mandatory root entry for CRD.

mhenriks: No, ImagePullPolicy should required.  `valid:"required"` means that the value (if set) cannot be empty string

spec should not be required until it has a member that is required.

redhatHameed: okay, In that case, I will remove the required from Spec.

redhatHameed: removed. the required from Spec.

pkg/operator/resources/operator/operator_test.go
redhatHameed: > Since this is explicitly imported, may want to explicitly add to glide.yaml with version info

added entry into glide.yaml

pkg/operator/resources/operator/operator_test.go
redhatHameed: > Oh yeah CDIs in code prob won’t work for negative test cases would have to use something like k8s unstructured or raw strings

added a negative test case for CR validation.

pkg/operator/resources/operator/operator.go
redhatHameed: > Is it possible to limit to valid values for k8s image pull policy (Always, IfNotPresent, Never)?

added the Enum which limits image pull policy values ```Always, IfNotPresent, Never.```

mhenriks:
ci test please

awels:
So the failures we are seeing in k8s are race conditions in the test. And I am not sure how to fix them. They are all negative tests, so we expect the import pod to fail and then we start a verifier pod to check the contents of the PVC. but if the importer pod restarted before the verifier pod could get a chance to grab the PVC, it will remain pending forever, like we are seeing in the failures here. I am not sure how we can force the verifier pod to get the PVC instead of the importer pod. Would manually deleting the importer pod give the verifier pod enough time to grab the PVC before a new importer pod is created by the import controller?

redhatHameed:
> @redhatHameed can you remove the file '0' from the root dir. It is generated by the build and should not be checked in.

@mhenriks removed file name '0' from the root dir..

awels:
ci test please

awels:
@mhenriks I think this is good to go no? The one failure is due to kubevirt ci image certificate issues. A rebase will fix.

awels:
ci test please

hack/custom-boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The CDI Authors.
+Copyright YEAR The Kubernetes Authors.
```
awels: So I think this change is causing travis to fail.

redhatHameed: @awels Reverted that change.

awels:
ci test please

PR #938, opened 2019-08-22
Resize disk image after cloning if we cloned a virtual disk.

### LINKED ISSUE #930: Dynamically resize virtual disk image based on size of cloned PVC

>
/kind enhancement

**What happened**:

A VM started from a larger PVC (source: pvc) that was cloned from a smaller "original" PVC (source: http) doesn't see the increased space on the root filesystem. 

VM has X+Y space on the root drive when started from a PVC of size X+Y.

Download OS onto PVC of size X. Clone datavolume using source: pvc to new PVC of size X + Y. Start VM from new PVC. Run `lsblk`

This is useful for dynamic VM flavours. The use case is an administrator creates a golden PVC of minimal size, and users can clone the PVC with a clone size >= golden size giving their VMs root filesystem as much or as little space as they require.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.9.5
- Kubernetes version (use `kubectl version`): 1.14.0
- Cloud provider or hardware configuration: Bare metal
- Install tools: kubeadm
- Others:

awels:

Add the ability to resize a clone after the clone finishes and we have cloned a virtual disk image.

Fixes #930 

**Release note**:

```release-note
Resize cloned virtual disk images.
```

### DIFF
--- a/pkg/uploadserver/uploadserver.go
+++ b/pkg/uploadserver/uploadserver.go
@@ -265,11 +265,7 @@ func (app *uploadServerApp) uploadHandler(w http.ResponseWriter, r *http.Request
 
 func newUploadStreamProcessor(stream io.ReadCloser, dest, imageSize, contentType string) error {
 	if contentType == FilesystemCloneContentType {
-		return filesystemCloneProcessor(stream, common.ImporterVolumePath)
-	}
-
-	if contentType == BlockdeviceCloneContentType {
-
+		return filesystemCloneProcessor(stream, common.ImporterVolumePath, imageSize)
 	}
 
 	uds := importer.NewUploadDataSource(stream)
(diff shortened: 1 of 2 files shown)


awels:
Think we need a functional test for this? clone a 1G virtual disk into a 2G, the resulting virtual disk should have size 2G

pkg/uploadserver/uploadserver.go
```diff
@@ -265,44 +265,62 @@ func (app *uploadServerApp) uploadHandler(w http.ResponseWriter, r *http.Request
 
 func newUploadStreamProcessor(stream io.ReadCloser, dest, imageSize, contentType string) error {
 	if contentType == FilesystemCloneContentType {
-		return filesystemCloneProcessor(stream, common.ImporterVolumePath)
-	}
-
-	if contentType == BlockdeviceCloneContentType {
```
mhenriks: Jeepers, I can't believe this was an empty `if` block.  My intention was obviously to call `blockdeviceCloneProcessor()` which is now gone.  Should I not be surprised that the data processor is doing the right thing?  It's not going to use scratch space, right?

awels: It doing the right thing, and no it should not be using scratch space, since the pod didn't ask for it. Yeah I was boggled by the empty if, and started implementing it, but then I went, wait a minute, this works, we have tests that explicitly test this for block devices, and they work.

awels:
This won't work for snapshot cloning.

awels:
Need a different approach for all cloning options.

PR #948, opened 2019-09-03
Set conditions properly while deploying.

awels:

The conditions were not properly being set during deployment (Available too soon). This fixes the conditions to match the following:
Start deploy:
Available: false
Progression: true
Degraded: false

During deploy stay on that.
Once deploy is finished and all pods are available:
Available: true
Progression: false
Degraded: false

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/controller/controller_test.go
+++ b/pkg/operator/controller/controller_test.go
@@ -156,6 +156,7 @@ var _ = Describe("Controller", func() {
 			It("should get deployed", func() {
 				args := createArgs()
 				doReconcile(args)
+				setDeploymentsReady(args)
 
 				Expect(args.cdi.Status.OperatorVersion).Should(Equal(version))
 				Expect(args.cdi.Status.TargetVersion).Should(Equal(version))
@@ -164,8 +165,7 @@ var _ = Describe("Controller", func() {
 				Expect(args.cdi.Status.Conditions).Should(HaveLen(3))
 				Expect(conditions.IsStatusConditionTrue(args.cdi.Status.Conditions, conditions.ConditionAvailable)).To(BeTrue())
 				Expect(conditions.IsStatusConditionFalse(args.cdi.Status.Conditions, conditions.ConditionProgressing)).To(BeTrue())
-				// We will expect degraded status, because in the test the deployment.status.replicas will not be 1, when the desired is 1.
-				Expect(conditions.IsStatusConditionTrue(args.cdi.Status.Conditions, conditions.ConditionDegraded)).To(BeTrue())
+				Expect(conditions.IsStatusConditionFalse(args.cdi.Status.Conditions, conditions.ConditionDegraded)).To(BeTrue())
 
 				Expect(args.cdi.Finalizers).Should(HaveLen(1))
 			})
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/controller_test.go
```diff
@@ -216,69 +216,9 @@ var _ = Describe("Controller", func() {
...
-
-					if len(args.cdi.Status.Conditions) == 3 &&
-						conditions.IsStatusConditionTrue(args.cdi.Status.Conditions, conditions.ConditionAvailable) &&
-						conditions.IsStatusConditionFalse(args.cdi.Status.Conditions, conditions.ConditionProgressing) &&
-						conditions.IsStatusConditionFalse(args.cdi.Status.Conditions, conditions.ConditionDegraded) {
-						running = true
-					}
-				}
-
-				Expect(running).To(BeTrue())
```
mhenriks: lot of red here.  not sure how this is all useless/invalid now.

awels: That is a duplicate of setDeploymentsReady(args), the code is identical. So instead of duplicating it, I just replaced it.

mhenriks:
>Start deploy:
Available: false
Progression: true
Degraded: false

Why not degraded here as well?  Under what conditions are we available and degraded?  And how is that different than not available and not degraded?

pkg/operator/controller/controller_test.go
mhenriks: part of it is identical.  not the whole thing.

pkg/operator/controller/controller_test.go
```diff
@@ -1314,6 +1264,29 @@ func setDeploymentsReady(args *args) bool {
...
+		Expect(err).ToNot(HaveOccurred())
+		if d.Spec.Replicas != nil {
+			d.Status.Replicas = int32(0)
+			d.Status.ReadyReplicas = d.Status.Replicas
+			err = args.client.Update(context.TODO(), d)
+			Expect(err).ToNot(HaveOccurred())
+		}
+
+	}
+	doReconcile(args)
```
mhenriks: should conditions be verified here?

awels:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md has all the combinations of conditions. Deploying is not degraded. But we can be degraded and available after we have been deployed and a lets say one of our deployment pods goes down, while its restarting the pod. If for instance the upload proxy goes down most of CDI is available except upload. So available and degraded.

pkg/operator/controller/controller_test.go
awels: I am assuming you are asking if I should verify if the degraded condition is true. The problem is depending on the state of the CR, degraded can be either true or false (if I am deploying degraded is always false, but if I am deployed already, then degraded will be true). I guess I can check if I am deploying or not and check based on that.

pkg/operator/controller/controller_test.go
awels: Actually you are right, first it setting the Status.Replicas to a valid number, and then it checks what the conditions are, then it sets the ReadyReplicas and checks the conditions again. Where setDeploymentsReady sets both at once and then checks the conditions. If we are interested in the intermediate check, I can put it back.

mhenriks:
> https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md has all the combinations of conditions. Deploying is not degraded. But we can be degraded and available after we have been deployed and a lets say one of our deployment pods goes down, while its restarting the pod. If for instance the upload proxy goes down most of CDI is available except upload. So available and degraded.

Okay, according to the doc, there are 6 possible condition value combinations.  Do we support all of them?  And if so, do we have tests for each?

awels:
> > https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md has all the combinations of conditions. Deploying is not degraded. But we can be degraded and available after we have been deployed and a lets say one of our deployment pods goes down, while its restarting the pod. If for instance the upload proxy goes down most of CDI is available except upload. So available and degraded.
> 
> Okay, according to the doc, there are 6 possible condition value combinations. Do we support all of them? And if so, do we have tests for each?

I put in helper functions to each combination (cr.go), and one of them was never called so I removed it. So technically we don't 'support' all 6 since we never encounter one combination.

As for the tests, the combinations mapped relatively easily onto our existing conditions, so I just modified those unit tests to check for the new conditions.

PR #949, opened 2019-09-03
Fix leader election test with import by not scaling.

awels:

Updated import while leader changes test to not use scale up as it is not reliable in all situations.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/leaderelection_test.go
+++ b/tests/leaderelection_test.go
@@ -26,7 +26,7 @@ const (
 	newDeploymentName    = "cdi-new-deployment"
 
 	pollingInterval = 2 * time.Second
-	timeout         = 90 * time.Second
+	timeout         = 180 * time.Second
 )
 
 var (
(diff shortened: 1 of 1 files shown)


tests/leaderelection_test.go
```diff
@@ -133,57 +133,69 @@ var _ = Describe("[rfe_id:1250][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+	var (
+		leaderPodName string
+		newDeployment *appsv1.Deployment
+		err           error
+	)
 
 	BeforeEach(func() {
-		By("Unsubscribing from OLM, so we can stop the operator")
-		// TODO once we have olm integration completed.
+		By("Check only one CDI controller deployment/pod")
```
mhenriks: I like the beforeeach/aftereach for this test.  Does it make sense to do the same for the other test?  Maybe have some shared setup/teardown functions?  If not, it's fine.

awels: it probably does make sense for a future us, so we aren't copying and pasting the same code over and over and over again. Let me fix it up so our code is better.

awels:
ci test please

PR #962, opened 2019-09-18
Creating a blank disk on a block device

zvikorn:

Creating a blank disk on a block device

```release-note
Creating a blank disk on a block device
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -710,6 +710,14 @@ func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolu
 		switch pvc.Status.Phase {
 		case corev1.ClaimPending:
 			dataVolumeCopy.Status.Phase = cdiv1.Pending
+			// the following check is for a case where the request is to create a blank disk for a block device.
+			// in that case, we do not create a pod as there is no need to create a blank image.
+			// instead, we just mark the DV phase as 'Succeeded' so any consumer will be able to use it.
+			phase, _ := pvc.Annotations[AnnPodPhase]
+			if phase == string(cdiv1.Succeeded) {
+				dataVolumeCopy.Status.Phase = cdiv1.Succeeded
+				c.updateImportStatusPhase(pvc, dataVolumeCopy, &event)
+			}
 		case corev1.ClaimBound:
 			switch dataVolumeCopy.Status.Phase {
 			case cdiv1.Pending:
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -133,6 +133,24 @@ func (ic *ImportController) requiresScratchSpace(pvc *v1.PersistentVolumeClaim) 
 // the importer pod as env vars. The pvc is checked (again) to ensure that we are not already
 // processing this pvc, which would result in multiple importer pods for the same pvc.
 func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
+	var lab map[string]string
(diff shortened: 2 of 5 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -710,6 +710,11 @@ func (c *DataVolumeController) updateDataVolumeStatus(dataVolume *cdiv1.DataVolu
 		switch pvc.Status.Phase {
 		case corev1.ClaimPending:
 			dataVolumeCopy.Status.Phase = cdiv1.Pending
```
aglitke: I think it would be helpful to comment here why a pending PVC could get the succeeded annotation.  It's not obvious what special case we're dealing with here.

pkg/controller/import-controller.go
```diff
@@ -133,6 +133,24 @@ func (ic *ImportController) requiresScratchSpace(pvc *v1.PersistentVolumeClaim)
...
 // processing this pvc, which would result in multiple importer pods for the same pvc.
 func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
+	var lab map[string]string
+	if !checkIfLabelExists(pvc, common.CDILabelKey, common.CDILabelValue) {
+		lab = map[string]string{common.CDILabelKey: common.CDILabelValue}
+	}
+	anno := map[string]string{}
+
+	//In case this is a request to create a blank disk on a block device, we do not create a pod.
+	//we just matk the DV as successful
```
aglitke: sp matk -> mark

pkg/controller/import-controller.go
```diff
@@ -133,6 +133,24 @@ func (ic *ImportController) requiresScratchSpace(pvc *v1.PersistentVolumeClaim)
 // the importer pod as env vars. The pvc is checked (again) to ensure that we are not already
 // processing this pvc, which would result in multiple importer pods for the same pvc.
 func (ic *ImportController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
+	var lab map[string]string
+	if !checkIfLabelExists(pvc, common.CDILabelKey, common.CDILabelValue) {
+		lab = map[string]string{common.CDILabelKey: common.CDILabelValue}
+	}
+	anno := map[string]string{}
+
+	//In case this is a request to create a blank disk on a block device, we do not create a pod.
```
aglitke: Space between // and text please.

pkg/controller/datavolume-controller.go
zvikorn: Done

pkg/controller/import-controller.go
zvikorn: Done

pkg/controller/import-controller.go
zvikorn: Done

tests/import_test.go
```diff
@@ -264,6 +264,25 @@ var _ = Describe("Importer Test Suite-Block_device", func() {
 		Expect(same).To(BeTrue())
 
 	})
+
+	It("Should create blank raw image for block PV", func() {
+		if !f.IsBlockVolumeStorageClassAvailable() {
+			Skip("Storage Class for block volume is not available")
+		}
+		dv := utils.NewDataVolumeForBlankRawImage("create-blank-image-to-block-pvc", "500Mi")
```
awels: So NewDataVolumeForBlankRawImage doesn't appear to set block volume mode. So that should not trigger the event and this test should fail.

awels: Maybe add a NewDataVolumeForBlankRawImageBlock that takes storage class and volume mode parameters.

zvikorn: yea... I missed that.

zvikorn: Done

PR #966, opened 2019-09-19
Sparsify clone tarring

awels:
Make xfs default fs for ceph in k8s provider.

Fix issue when cloning in XFS the target would take more space than the source due to not being sparsified.
Update default FS on ceph to be xfs instead of ext4
Fixes #

**Release note**:

```release-note
BugFix: cloning using xfs file system to same size PVC
```

### DIFF
--- a/cluster-sync/k8s-1.15.1/provider.sh
+++ b/cluster-sync/k8s-1.15.1/provider.sh
@@ -12,8 +12,32 @@ fi
 
 if [[ $KUBEVIRT_PROVIDER_EXTRA_ARGS =~ "--enable-ceph" ]]; then
   echo "Switching default storage class to ceph"
-  #Switch the default storage class to ceph
+  # Switch the default storage class to ceph
   _kubectl patch storageclass local -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
-  _kubectl patch storageclass csi-rbd -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
+  # Make sure the SC is configured to use xfs instead of ext4
+  _kubectl delete storageclass csi-rbd
+  cat <<EOF | _kubectl apply -f -
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+  name: csi-rbd
+parameters:
+  adminid: admin
+  csi.storage.k8s.io/node-publish-secret-name: csi-rbd-secret
+  csi.storage.k8s.io/node-publish-secret-namespace: default
+  csi.storage.k8s.io/provisioner-secret-name: csi-rbd-secret
+  csi.storage.k8s.io/provisioner-secret-namespace: default
+  imageFeatures: layering
+  imageFormat: "2"
+  monitors: 192.168.66.2
+  multiNodeWritable: enabled
+  pool: rbd
+  fsType: xfs
+provisioner: csi-rbdplugin
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+EOF
 fi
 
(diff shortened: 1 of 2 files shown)


cluster-sync/k8s-1.15.1/provider.sh
```diff
@@ -12,8 +12,32 @@ fi
...
   echo "Switching default storage class to ceph"
-  #Switch the default storage class to ceph
+  # Switch the default storage class to ceph
   _kubectl patch storageclass local -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
-  _kubectl patch storageclass csi-rbd -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
+  # Make sure the SC is configured to use xfs instead of ext4
+  _kubectl delete storageclass csi-rbd
+  cat <<EOF | _kubectl apply -f -
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
```
mhenriks: I think this is fine but wonder if if would be better to do in kubevirtci?

awels: I would actually prefer that kubevirtci doesn't install any storage providers at all, maybe provide some scripts to active the ones you want. But at this point I would like to minimize the storage modifications in kubevirtci.

mhenriks:
@awels I agree with @aglitke.  Resize expands the image to fill the entire disk leaving no room for metadata.  By default, XFS may be configured to use up to 25% of available space for metadata.  Since we will only have a single file on the filesystem (but it may be a big file with a big block map), I don't think we'll need that much metadata space.  But I think it would be wise to reserve somewhere between 1% and 5%.

awels:
@mhenriks  So you are suggesting we resize to the max - a certain percentage? So we would be adjusting the code for a very specific provider/fs combination in code that should be agnostic to the underlying storage... And that combination is lying about how much space is available to write to.

mhenriks:
@awels xfs is not lying.  The problem is that a non sparse file requires more metadata than a sparse file and we should attempt to accommodate for that.  Otherwise, the filesystem could run out of space at some point when the sparse file is written to and requires more metadata.  Which is what is happening when we clone without passing --sparse to tar.

My recommendation is for the resize to leave some percentage of the filesystem free to accommodate the metadata that will be required as the sparse file is written to.

awels:
So I understand that XFS has some overhead, so here is what I am seeing:
```
[root@node01 vagrant]# xfs_info /dev/rbd0 
meta-data=/dev/rbd0              isize=512    agcount=8, agsize=16384 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=0 spinodes=0
data     =                       bsize=4096   blocks=128000, imaxpct=25
         =                       sunit=1024   swidth=1024 blks
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
log      =internal               bsize=4096   blocks=624, version=2
         =                       sectsz=512   sunit=8 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
```
So I have 128000 blocks of 4k, 624 are used for journaling, which leaves me 127376 blocks available to store data. This matches exactly with what I am seeing inside the pod when I run df
```
bash-4.4# df -h --block-size=4k
Filesystem     4K-blocks    Used Available Use% Mounted on
overlay         10480385 3006758   7473627  29% /
tmpfs             614018       0    614018   0% /dev
tmpfs             614018       0    614018   0% /sys/fs/cgroup
/dev/rbd0         127376  127371         5 100% /pvctest <----
/dev/vda1       10480385 3006758   7473627  29% /etc/hosts
shm                16384       0     16384   0% /dev/shm
tmpfs             614018       3    614015   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs             614018       0    614018   0% /proc/acpi
tmpfs             614018       0    614018   0% /proc/scsi
tmpfs             614018       0    614018   0% /sys/firmware
```
So when I write data to fill up the disk I end up with the result above. 127376 total, 127371 used, and 5 available, but not really since I can't actually write more. What happened to the last 5 blocks. Those are not journaling blocks as that is already accounted for.

PR #981, opened 2019-09-27
Wait for SC to be available before making it default

awels:

The OKD sometimes take a bit to get the local storage class available, this PR adds a wait to make sure the local SC is available before setting it as default.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/okd-4.1/provider.sh
+++ b/cluster-sync/okd-4.1/provider.sh
@@ -7,7 +7,22 @@ function seed_images(){
 }
 
 function configure_local_storage() {
-  #Set the default storage class.
+  retry_counter=0
+  sc=`_kubectl get sc local -o=jsonpath="{.metadata.name}"`
+  all_sc=`_kubectl get sc`
+  echo "All storage classes:"
+  echo "${all_sc}"
+  while [[ $retry_counter -lt 60 ]] && [ "$sc" != "local" ]; do
+    sc=`_kubectl get sc local -o=jsonpath="{.metadata.name}"`
+    retry_counter=$((retry_counter + 1))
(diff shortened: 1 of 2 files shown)


awels:
I could remove the 'if' that would cause it to always try setting the sc as the default and if it is not there it would fail and abort the CI run.

cluster-sync/okd-4.1/provider.sh
```diff
@@ -7,7 +7,15 @@ function seed_images(){
 }
 
 function configure_local_storage() {
-  #Set the default storage class.
+  retry_counter=0
+  while [[ $retry_counter -lt 60 ]] && [ "$sc" != "local" ]; do
```
mhenriks: sc not defined

awels: Right, so $sc will not match "local" and it gets defined in the loop.

awels: I guess I can save us 5 seconds if I copy
```bash
sc=`_kubectl get sc local -o=jsonpath="{.metadata.name}"`
```
outside of the loop, if it exists already, the loop doesn't get entered.

mhenriks: Yeah nbd because not using "set -u"

awels: I likely don't need retry_counter=0 either, but I wasn't sure what would happen if I added one to an undefined value. If it would figure out hey it should be a number and thus I will initalize it to 0 and add 1.

awels:
ci test please

awels:
ci test please

awels:
ci test please

awels:
ci test please

PR #982, opened 2019-09-30
Remove OLM integration as we aren't using it.

awels:
Removed OLM code since it is not being used in CDI deployment.

Removes OLM catalog and CSV generation as we aren't using it and it is causing problems.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
/hold
Need to update hack/README.md and create a new deploy readme in docs.

awels:
If we keep the CSV generation stuff, then we need to keep the parameters passed into the generator. And we have to keep them at some value. And I prefer to not keep cruft around if we aren't using it. We have all the code in git, so if we need to retrieve it, its easy to find.

awels:
@mhenriks Checked with Ryan and David and yes they do use the csv generator, so I put that code back.

hack/build/build-copy-artifacts.sh
```diff
@@ -31,12 +31,6 @@ for tgt in ${targets}; do
     if [[ "${bin_name}" == "${CLONER}" ]]; then
         mkdir -p "${CMD_OUT_DIR}/${bin_name}/"
         cp -f "${CDI_DIR}/cmd/${bin_name}/cloner_startup.sh" "${dest_dir}"
-    elif [[ "${bin_name}" == "${OPERATOR}" ]]; then
-        # This container also needs the csv-generator file, so it needs to be
-        # copied into the operator's directory for docker build to source
-        # csv-generator correctly
-        mkdir -p "${CMD_OUT_DIR}/${bin_name}/"
-        cp ${OUT_DIR}/tools/${CSV_GENERATOR}/${CSV_GENERATOR} ${OUT_DIR}/cmd/${OPERATOR}/${CSV_GENERATOR}
```
mhenriks: doesn't this have to stay?

hack/build/config.sh
```diff
@@ -19,15 +19,13 @@ APISERVER="cdi-apiserver"
 UPLOADPROXY="cdi-uploadproxy"
 UPLOADSERVER="cdi-uploadserver"
 OPERATOR="cdi-operator"
-CDI_OLM_CATALOG="cdi-olm-catalog"
 FUNC_TEST_INIT="cdi-func-test-file-host-init"
 FUNC_TEST_HTTP="cdi-func-test-file-host-http"
 FUNC_TEST_REGISTRY="cdi-func-test-registry"
 FUNC_TEST_REGISTRY_POPULATE="cdi-func-test-registry-populate"
 FUNC_TEST_REGISTRY_INIT="cdi-func-test-registry-init"
-CSV_GENERATOR="csv-generator"
```
mhenriks: should this stay?

hack/build/docker/cdi-operator/Dockerfile
```diff
@@ -17,6 +17,5 @@ WORKDIR /home/cdi-operator
 USER 1001
 
 COPY ./cdi-operator /usr/bin/cdi-operator
-COPY ./csv-generator /usr/bin/csv-generator
```
mhenriks: Should this stay?

mhenriks:
Should we add a functional test that executes the csv-generator?

hack/build/build-copy-artifacts.sh
awels: Well only the downstream version is used, and the downstream build still has the information needed to compile and copy the binary into the downstream container. They shouldn't be using the upstream csv generator to create the HCO CSV.

hack/build/config.sh
awels: Same answer, if we don't care about upstream CSV generator, then we don't need to compile and install it in the container.

hack/build/docker/cdi-operator/Dockerfile
awels: Same answer, if we don't care about upstream CSV generator, then we don't need to compile and install it in the container.

awels: Only reason I can see it remaining here, is to stop the Dockerfile check downstream from triggering that something changed.

PR #987, opened 2019-10-06
Unique client cert per clone source pod

mhenriks:

Clone source pods currently use the same TLS client key/cert as uploadproxy.  This is a potential security concern because an attacker could do the following:

1) Create clone in namespace they have admin/edit access to
2) Get TLS key/cert from clone source pod environment
3) Use key/cert to upload to any uploadserver pod in the cluster

Addressing this by: 

1) Assigning unique CN for each source/target pod combo
2) Uploadserver pod validates CN

Fixes #

**Release note**:

```release-note
Unique TLS key/cert per clone source pod
```

### DIFF
--- a/pkg/controller/upload-controller_test.go
+++ b/pkg/controller/upload-controller_test.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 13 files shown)


pkg/controller/upload-controller_test.go
```diff
@@ -43,6 +43,14 @@ const (
 	cloneRequestAnnotation  = "k8s.io/CloneRequest"
 )
 
+var (
+	yyytestUploadServerCASecret     *corev1.Secret
```
awels: Can you fix the variable names here to not include yyy/zzz/etc?

mhenriks: DONE DEAL

rmohr:
/test pull-containerized-data-importer-e2e-k8s-1.15.1-ceph
/test pull-containerized-data-importer-e2e-okd-4.1

@mhenriks I don't expect them to pass, but I want to see what happens :)

PR #995, opened 2019-10-17
[skip ci] Enable bazel cache

awels:

Enable the bazel cache to work with kubevirts cache

Fixes #

**Release note**:

```release-note
NONE
```


automation/test.sh
```diff
@@ -30,6 +30,7 @@ export CDI_NAMESPACE=$NAMESPACE
 echo "namespace: ${NAMESPACE}, cdi-namespace: ${CDI_NAMESPACE}"
 
 readonly ARTIFACTS_PATH="exported-artifacts"
+readonly BAZEL_CACHE="${BAZEL_CACHE:-http://bazel-cache.kubevirt-prow.svc.cluster.local:8080/kubevirt.io/kubevirt}"
```
rmohr: the path of the url should be `kubevirt.io/containerized-data-importer`.

awels: Fixed

rmohr:
Looks good to me. Follow-ups to get all features from Prow:

* Save your junit ginkgo junit file like this, so that Prow can pick it up: `--junit-output=$ARTIFACTS_PATH/junit.functest.xml`.
* Copy and wire this reporter into your testsuite: https://github.com/kubevirt/kubevirt/tree/master/tests/reporter. This will collect logs per test-failure and uploads them to GCS.

Once you have these two changes in place, we can also enable flakefinder for you.

rmohr:
Oh and ARTIFACTS_PATH should point to $ARTIFACTS (that is the environment variable from prow, which denotes the location for artifacts to upload to GCS).

rmohr:
/lgtm

I am unsure if you want to disable the stdci webhook, or just delete the automation folder. I am mostly thinking about backports here.

awels:
So I think for now I want to remove the automation folder, in case we have a back port we can still run CI on that.

awels:
The failures appear to all be because of snapshot cloning failing in one way or another. Needs more investigation, but proves the artifact generation works properly. So I am happy at this point.

PR #997, opened 2019-10-21
Use an existing bazel-builder image instead of building it every time

aglitke:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/config.sh
+++ b/hack/build/config.sh
@@ -24,6 +24,7 @@ FUNC_TEST_HTTP="cdi-func-test-file-host-http"
 FUNC_TEST_REGISTRY="cdi-func-test-registry"
 FUNC_TEST_REGISTRY_POPULATE="cdi-func-test-registry-populate"
 FUNC_TEST_REGISTRY_INIT="cdi-func-test-registry-init"
+BUILDER_TAG=${BUILDER_TAG:-kubevirt/kubevirt-cdi-bazel-builder@sha256:38e449bb8f5e2e3dc8ae241323c69ce37e72bd98a109d2032955dd9bd0e0a545}
 
 BINARIES="cmd/${OPERATOR} cmd/${CONTROLLER} cmd/${IMPORTER} cmd/${CLONER} cmd/${APISERVER} cmd/${UPLOADPROXY} cmd/${UPLOADSERVER} cmd/${OPERATOR} tools/${FUNC_TEST_INIT} tools/${FUNC_TEST_REGISTRY_INIT}"
 CDI_PKGS="cmd/ pkg/ test/"
--- a/hack/build/in-docker.sh
+++ b/hack/build/in-docker.sh
@@ -18,13 +18,9 @@ set -e
 
 script_dir="$(readlink -f $(dirname $0))"
 source "${script_dir}"/common.sh
+source "${script_dir}"/config.sh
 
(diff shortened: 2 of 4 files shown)


rmohr:
> So this is going to be a chicken and egg problem, we need to push the builder to dockerhub first, and then we can use it, but not before. I see you 'solved' it by using your own. kubevirt has a separate 'push' for just that one image outside of bazel. Something to look into?

This is indeed tricky. We have a pretty complex solution right now. We automatically push the base image after every merge for every branch. So if something is merged against `master`, we push to the `:master` tag.

The tests in prow just pull dumb the latest one of the target branch of a PR.

I see this system as a technical debt. Most of the things which we install in the build container are still coming from the pre-bazel time, or are easy to replace. At the end we should pretty much only need bazel and rsync in there.

hack/build/config.sh
```diff
@@ -24,6 +24,7 @@ FUNC_TEST_HTTP="cdi-func-test-file-host-http"
 FUNC_TEST_REGISTRY="cdi-func-test-registry"
 FUNC_TEST_REGISTRY_POPULATE="cdi-func-test-registry-populate"
 FUNC_TEST_REGISTRY_INIT="cdi-func-test-registry-init"
+BUILDER_TAG=${BUILDER_TAG:="kubevirt/kubevirt-cdi-bazel-builder"}
```
awels: Can you make this:
```
BUILDER_TAG=${BUILDER_TAG:-kubevirt-cdi-bazel-builder@sha256:38e449bb8f5e2e3dc8ae241323c69ce37e72bd98a109d2032955dd9bd0e0a545}
```

hack/build/config.sh
```diff
@@ -24,6 +24,7 @@ FUNC_TEST_HTTP="cdi-func-test-file-host-http"
 FUNC_TEST_REGISTRY="cdi-func-test-registry"
 FUNC_TEST_REGISTRY_POPULATE="cdi-func-test-registry-populate"
 FUNC_TEST_REGISTRY_INIT="cdi-func-test-registry-init"
+BUILDER_TAG=${BUILDER_TAG:-kubevirt-cdi-bazel-builder@sha256:38e449bb8f5e2e3dc8ae241323c69ce37e72bd98a109d2032955dd9bd0e0a545}
```
awels: Okay I messed up when I gave the info, it should be:
```
BUILDER_TAG=${BUILDER_TAG:-kubevirt/kubevirt-cdi-bazel-builder@sha256:38e449bb8f5e2e3dc8ae241323c69ce37e72bd98a109d2032955dd9bd0e0a545}
```
Note the extra 'kubevirt'

hack/build/in-docker.sh
```diff
@@ -18,10 +18,10 @@ set -e
 
 script_dir="$(readlink -f $(dirname $0))"
 source "${script_dir}"/common.sh
+source "${script_dir}"/config.sh
 
 WORK_DIR="/go/src/kubevirt.io/containerized-data-importer"
 BUILDER_SPEC="${BUILD_DIR}/docker/builder"
-BUILDER_TAG="kubevirt-cdi-bazel-builder"
 
 # Build the encapsulated compile and test container
```
awels: These two lines need to be removed as well, its still trying to build these.

PR #999, opened 2019-10-27
Add node annotation on PVS for HPP only in Immediate Binding Mode

rollandf:
Signed-off-by: Fred Rolland <frolland@redhat.com>

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/common.go
+++ b/tests/utils/common.go
@@ -65,12 +65,14 @@ func getDefaultStorageClass(client *kubernetes.Clientset) *storagev1.StorageClas
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


danielerez:
> I would just remove the code that explicity sets nodes, once the hpp (looking at that) is updated, then we don't need explicit binding at all.

We still want to support the old behaviour, so we need to keep the 'provisionOnNode' annotation handling.

awels:
I think we can abandon this PR. The latest hostpath provisioner, supports the annotation in both volumeBindingModes. So there is no need to check for Immediate.

rollandf:
Closing as HPP supports the annotation on both modes (waitforfirstcustomre/immediate)

PR #1010, opened 2019-11-07
Replace current ceph install with operator based rook-ceph.

awels:
Add reporting of DVs on test failures.

Update ceph to the latest version available.
Implement DV dumping on test failures.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/smart-clone-controller.go
+++ b/pkg/controller/smart-clone-controller.go
@@ -116,12 +116,58 @@ func (c *SmartCloneController) ProcessNextItem() bool {
 	if shutdown {
 		return false
 	}
-	defer c.queue.Done(key)
 
-	ns, name, err := cache.SplitMetaNamespaceKey(key.(string))
+	// We wrap this block in a func so we can defer c.workqueue.Done.
+	err := func(obj interface{}) error {
+		// We call Done here so the workqueue knows we have finished
+		// processing this item. We also must remember to call Forget if we
+		// do not want this work item being re-queued. For example, we do
+		// not call Forget if a transient error occurs, instead the item is
+		// put back on the workqueue and attempted again after a back-off
+		// period.
+		defer c.queue.Done(obj)
+		var key string
+		var ok bool
+		// We expect strings to come off the workqueue. These are of the
+		// form namespace/name. We do this as the delayed nature of the
+		// workqueue means the items in the informer cache may actually be
+		// more up to date that when the item was initially put onto the
+		// workqueue.
+		if key, ok = obj.(string); !ok {
+			// As the item in the workqueue is actually invalid, we call
+			// Forget here else we'd go into a loop of attempting to
+			// process a work item that is invalid.
+			c.queue.Forget(obj)
+			runtime.HandleError(fmt.Errorf("expected string in workqueue but got %#v", obj))
+			return nil
+		}
+		// Run the syncHandler, passing it the namespace/name string of the data volume
+		// to be synced.
+		if err := c.syncHandler(key); err != nil {
+			// Put the item back on the workqueue to handle any transient errors.
+			c.queue.AddRateLimited(key)
+			return fmt.Errorf("error syncing '%s': %s, requeuing", key, err.Error())
+		}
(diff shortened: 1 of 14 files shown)


awels:
So the problem was that datavolume controller would sync the data volume, right before the smart controller would write succeeded to the data volume, now they are out of sync, and the smart controller write would fail. This PR re-reads the data volume and tries again in that case.

pkg/controller/smart-clone-controller.go
```diff
@@ -335,16 +343,30 @@ func (c *SmartCloneController) updateSmartCloneStatusPhase(phase cdiv1.DataVolum
...
 
-func (c *SmartCloneController) emitEvent(dataVolume *cdiv1.DataVolume, dataVolumeCopy *cdiv1.DataVolume, event *DataVolumeEvent) error {
+func (c *SmartCloneController) emitEvent(dataVolume *cdiv1.DataVolume, dataVolumeCopy *cdiv1.DataVolume, event *DataVolumeEvent, newPVC *corev1.PersistentVolumeClaim) error {
 	// Only update the object if something actually changed in the status.
 	if !reflect.DeepEqual(dataVolume.Status, dataVolumeCopy.Status) {
 		_, err := c.cdiClientSet.CdiV1alpha1().DataVolumes(dataVolume.Namespace).Update(dataVolumeCopy)
-		// Emit the event only when the status change happens, not every time
-		if event.eventType != "" {
-			c.recorder.Event(dataVolume, event.eventType, event.reason, event.message)
+		if err != nil {
```
mhenriks: What was going on here?  Why not okay to just return the error?

pkg/controller/smart-clone-controller.go
```diff
@@ -335,16 +343,30 @@ func (c *SmartCloneController) updateSmartCloneStatusPhase(phase cdiv1.DataVolum
...
+			if k8serrors.IsConflict(err) {
+				// Something else updated before we could. Refresh data volume and try again
+				klog.V(3).Info("Something updated the data volume, fetching fresh one and trying again.")
+				refreshDataVolume, err := c.dataVolumeLister.DataVolumes(dataVolume.Namespace).Get(dataVolume.Name)
+				if err != nil {
+					klog.Errorf("error getting data volume: %v", err)
+					return err
+				}
+				return c.updateSmartCloneStatusPhase(dataVolumeCopy.Status.Phase, refreshDataVolume, newPVC)
```
mhenriks: emitEvent is called by `updateSmartCloneStatusPhase` so recursing here.  Forever it keeps on failing.

awels: Yes I debated putting a limit on the recursion here. Btw it only recurses IF the error returned is a conflict error. So it would only recurse forever if each time in between us refreshing the object and attempting to write the update, something else has updated the object.

pkg/controller/smart-clone-controller.go
awels: So what happens if we return the error, is that the data volume phase is not updated. So that will requeue the data volume. Both the data volume and smart controller will grab it, and the exact same thing will happen, right before we update in the smart controller, the data volume controller will finish syncing and we would fail again, and thus requeue again, etc, etc. So instead of always returning the error, I check for a conflict, do a recursive refresh the data volume, then update it while the data volume controller is done with this data volume and thus is not updating it.

mhenriks: I understand how there can be contention between the two controllers but I don't see how a failure here causes the datavolume to be "requeued" in the datavolume controller.  If the update succeeds, then yes.  But not on failure. 
 Controllers should have independent queues.  And failures don't cause items to be "requeued" rather on success keys are removed from the queue.

pkg/controller/smart-clone-controller.go
awels: Removed this, and added a c.queue.Add instead, so no recursion.

pkg/controller/smart-clone-controller.go
awels: So it turns out to not be re-queuing on failure. If I force a re-queue it works.

PR #1012, opened 2019-11-10
cut down operator permissions

mhenriks:
Signed-off-by: Michael Henriksen <mhenriks@redhat.com>

**What this PR does / why we need it**:

Minimize cluster role permissions.  Add namespaced permissions.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #1013, opened 2019-11-11
Fix typo in operator cluster roles

aglitke:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/resources/operator/operator.go
+++ b/pkg/operator/resources/operator/operator.go
@@ -182,7 +182,7 @@ func getOperatorClusterRules() *[]rbacv1.PolicyRule {
 			Resources: []string{
 				"deployments",
 				"deployments/finalizers",
-				"daemonstes",
+				"daemonsets",
 			},
 			Verbs: []string{
 				"create",


aglitke:
@mhenriks Why wouldn't this typo be breaking our deployment?

awels:
> @mhenriks Why wouldn't this typo be breaking our deployment?

I don't think we are using daemonsets in CDI, so there is no need to have this permission?

mhenriks:
@aglitke there are no daemonsets in CDI also, daemonsets are not included in #1012 so this PR should prob be closed

MarSik:
@mhenriks Well, if you do not need the permission then you should probably remove it completely. #1012 does not seem to be touching this one.

awels:
https://github.com/kubevirt/containerized-data-importer/pull/1012/files#diff-afdd9efbf2b4dce118add9994e36486fL185 removes it as far as I can tell.

MarSik:
@awels You are right.. I could not Ctrl-F it in my browser for some reason..

awels:
It was hiding by default because of the big change in that particular file. That is why ctrl-f didn't find it.

awels:
But I think we can close this in favor of #1012 now.

aglitke:
Agreed.

PR #1023, opened 2019-11-14
Provide a more verbose error.

maya-r:
Use the term that appears in YAML files, volumeMode, and print
which one it is (may be an implicit value picked from the default)

```release-note
Improve volumeMode mismatch error reports
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -619,7 +619,8 @@ func ValidateCanCloneSourceAndTargetSpec(sourceSpec, targetSpec *v1.PersistentVo
 		targetVolumeMode = v1.PersistentVolumeBlock
 	}
 	if sourceVolumeMode != targetVolumeMode {
-		return errors.New("source and target volume modes do not match")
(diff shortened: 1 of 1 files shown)


pkg/controller/util.go
```diff
@@ -619,7 +619,8 @@ func ValidateCanCloneSourceAndTargetSpec(sourceSpec, targetSpec *v1.PersistentVo
 		targetVolumeMode = v1.PersistentVolumeBlock
 	}
 	if sourceVolumeMode != targetVolumeMode {
-		return errors.New("source and target volume modes do not match")
+		return fmt.Errorf("source volumeMode (%v) and target volumeMode (%v) do not match",
```
awels: So I would use %s instead of %v here. %s prints a string, while %v prints an entire object. In this case the type is actually a string, so %v and %s will return the same result. But if you have a more complex struct or something, then %v will print everything in the struct. Depending on what you are showing this may or may not be what you want.

https://golang.org/pkg/fmt/ will explain it better than I can.

PR #1031, opened 2019-11-18
make dataVolume wait for its PVC to be deleted in status terminating

maya-r:
```release-note
NONE
```


maya-r:
thanks for the review

PR #1034, opened 2019-11-22
Use controller library for config controller

### LINKED ISSUE #913: CDIConfig not updated when default storage class changes

>
/kind bug
> /kind enhancement

**What happened**:
I switched the default storage class in my cluster, and the CDIConfig did not have a specific storage class defined for usage as scratch space. Normally the resolved scratch space SC will be the default, but the default storage class changed, and scratch space SC didn't change to match.

If I change the default storage class AND I don't have a specific scratch space storage class, I expect the resolved on to be the new default.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): master
- Install tools:
- Others:

awels:

Update the cdi config controller to use the kubernetes controller library.
Add new functional tests
Update the unit tests to use the controller library.

Fixes #913 

**Release note**:

```release-note
NONE
```


cmd/cdi-controller/controller.go
```diff
@@ -256,18 +256,17 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
...
-	go func() {
-		err = configController.Run(1, stopCh)
-		if err != nil {
-			klog.Fatalf("Error running config controller: %+v", err)
-		}
-	}()
-
 	startSmartController(extClient, csiInformerFactory, smartCloneController, stopCh)
+
+	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
```
mhenriks: signal handlers are set elsewhere in this file as well.  Should prob only do it in one place.  I think moving leader election and signal handling to controller runtime may make sense.

pkg/controller/runtime-util.go
```diff
@@ -0,0 +1,121 @@
...
+				common.CDIComponentLabel: "",
+			},
+		},
+	}
+	return config
+
+}
+
+// SetOwnerToOperator sets the object owner to the same owner as the cdi config map.
+func SetOwnerToOperator(client *client.Client, object metav1.Object, namespace string) error {
```
mhenriks: `client.Client` is an interface so client param prob shouldn't be a pointer.  But I propose getting rid of this function entirely and using `SetOwner` in pkg/operator.  The config controller can have a client-go client in addition to controller runtime client.

pkg/controller/config-controller.go
```diff
@@ -1,377 +1,190 @@
...
+func (r *CDIConfigReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+	log := r.Log.WithValues("CDIConfig", req.NamespacedName)
+	log.Info("reconciling CDIConfig")
 
-	var config *cdiv1.CDIConfig
-	config, err := c.configLister.Get(c.configName)
-	if err != nil {
-		runtime.HandleError(errors.Errorf("error getting CDI config: %s", err))
-		return
+	var config cdiv1.CDIConfig = *MakeEmptyCDIConfigSpec(r.ConfigName)
```
mhenriks: May make more sense for this variable to be a pointer

pkg/controller/config-controller.go
```diff
@@ -1,377 +1,190 @@
...
-		defer c.queue.Done(obj)
-
-		var key string
-		var ok bool
-
-		if key, ok = obj.(string); !ok {
-			c.queue.Forget(obj)
-			runtime.HandleError(errors.Errorf("expected string in workqueue but got %#v", obj))
-			return nil
+	if !reflect.DeepEqual(currentConfigCopy, config) {
```
mhenriks: as is `currentConfigCopy` is interface with a pointer receiver.  config is a struct (not pointer).  Does `DeepEqual` do the right thing given that?

cmd/cdi-controller/controller.go
```diff
@@ -192,15 +186,30 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
...
-		ingressInformer,
-		routeInformer,
-		configInformer,
-		uploadProxyServiceName,
-		configName,
-		pullPolicy,
-		verbose)
+	mgr, err := manager.New(config.GetConfigOrDie(), manager.Options{
+		// Restrict to only watching namespaced scoped objects in cdi namespace, will also watch cluster scoped objects.
+		Namespace: util.GetNamespace(),
```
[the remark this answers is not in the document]
awels: We could, but I don't think that should be part of this particular PR. But yes that thought crossed my mind as well.

cmd/cdi-controller/controller.go
awels: Until everything is converted, we will have some duality and duplication between the two mechanisms, one of which is this, as we move controllers that will solve itself.

cmd/cdi-controller/controller.go
```diff
@@ -192,15 +186,30 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
...
+	})
+	if err != nil {
+		klog.Errorf("Unable to setup controller manager: %v", err)
+		os.Exit(1)
+	}
+
+	configController, err := sigcontroller.New("config-controller", mgr, sigcontroller.Options{
```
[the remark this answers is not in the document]
awels: Yeah, I already moved the watches in there, makes sense to move the instantiation as well.

pkg/controller/config-controller.go
awels: Good question, need to double check.

pkg/controller/config-controller.go
awels: yeah you are right.

pkg/controller/runtime-util.go
awels: let me check.

pkg/controller/config-controller.go
awels: Since I am making config a pointer, then deep equal will do the right thing.

awels:
So lets treat Ingress the same as Route, its an optional extra, so if it is not there when this controller becomes active, it won't react to it like with a Route.

pkg/controller/runtime-util.go
awels: Switched the create to use the client-go clients and made an init just because I could.

pkg/controller/config-controller.go
```diff
@@ -1,377 +1,238 @@
...
+	}
+	if err := configController.Watch(&source.Kind{Type: &storagev1.StorageClass{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
+	if err := configController.Watch(&source.Kind{Type: &extensionsv1beta1.Ingress{}}, &handler.EnqueueRequestForObject{}); IgnoreIsNoMatchError(err) != nil {
+		return err
+	}
+	if err := configController.Watch(&source.Kind{Type: &routev1.Route{}}, &handler.EnqueueRequestForObject{}); IgnoreIsNoMatchError(err) != nil {
+		return err
+	}
```
mhenriks: something to consider:  I think it would be more efficient to use `handler.EnqueueRequestsFromMapFunc` for routes/ingress/storageclasses.  It will return a `reconcile.Request` for the singleton config object.  We don't care about the resource that actually triggered the reconcile.  And by always reconciling on the same resource, we should reconcile less frequently when there are bursts of activity.  Not that we expect updates to be that frequent.

The operator uses `handler.EnqueueRequestsFromMapFunc` here but this implementation can be much simpler (don't have to look up anything):  https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/operator/controller/scc.go#L39-L58 

May also want to filter out routes/ingress that do not refer to cdi-uploadproxy.

cmd/cdi-controller/controller.go
```diff
@@ -192,37 +194,26 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
...
-		pullPolicy,
-		verbose)
+	_, err = controller.NewConfigController(mgr, cdiClient, client, log, uploadProxyServiceName, configName)
+	if err != nil {
+		klog.Errorf("Unable to setup config controller: %v", err)
+		os.Exit(1)
+	}
 
+	klog.V(1).Infoln("Init CDI Config object")
+	controller.CreateCDIConfig(cdiClient, client, configName, util.GetNamespace())
```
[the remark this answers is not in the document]
mhenriks: also why not have generic `Init` function instead to keep this layer more generic?

cmd/cdi-controller/controller.go
```diff
@@ -53,7 +59,6 @@ var (
 //   specified we do an in-cluster config. For testing it's easiest to export KUBECONFIG.
 func init() {
 	// flags
-	flag.StringVar(&configPath, "kubeconfig", os.Getenv("KUBECONFIG"), "(Optional) Overrides $KUBECONFIG")
```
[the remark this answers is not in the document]
awels: Early in development it was causing some issue, but that has been resolved, I forgot to undo it.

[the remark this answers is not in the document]
awels: Actually now I remember, 
/usr/bin/cdi-controller flag redefined: kubeconfig
panic: /usr/bin/cdi-controller flag redefined: kubeconfig

Looks like the runtime-controller already specifies the flag, so now we are redefining it. So better to let the runtime controller library define it, we should still be able to use it.

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
awels: Well what should I do if the create cdi config object fails? The reconciler will get a chance later to create it again. So if it is not a fatal error not much I can do with the error except log it, and move on. I will do that. As for Init, init is a special function in Go and I don't like having a function that is named almost the same, could cause confusion. But I see uploadController has an init already, let me rework the function to be part of the controller.

cmd/cdi-controller/controller.go
awels: I am open to suggestions here, not sure how to make it cleaner.

pkg/controller/runtime-util.go
```diff
@@ -0,0 +1,45 @@
...
+	return &cdiv1.CDIConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Labels: map[string]string{
+				common.CDILabelKey:       common.CDILabelValue,
+				common.CDIComponentLabel: "",
+			},
+		},
+		Spec:   cdiv1.CDIConfigSpec{},
+		Status: cdiv1.CDIConfigStatus{},
```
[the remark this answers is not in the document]
awels: good point. removed.

pkg/operator/resources/cluster/factory.go
```diff
@@ -44,6 +44,7 @@ func createCRDResources(args *FactoryArgs) []runtime.Object {
 	return []runtime.Object{
 		createDataVolumeCRD(),
 		createCDIConfigCRD(),
+		//		createCDIConfigInstance(),
```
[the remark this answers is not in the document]
awels: Done

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
mhenriks: Then get rid of the configPath variable if it's not needed.  Right now it is referenced here https://github.com/kubevirt/containerized-data-importer/blob/e098d397fef14678feaa8db54320787753fc1dc8/cmd/cdi-controller/controller.go#L259 and will always be empty

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
mhenriks: The process should exit. Remember, reconciler will not run if there are no route/ingress/storageclasses.  That's the whole reason we are creating it here.

Package `init()` is not exportable and will prob not be confused with `.Init()` member.

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
awels: Done

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
awels: Good point, that will simply cause the pod to restart and try again if it was an intermittent failure. done.

pkg/controller/config-controller.go
```diff
@@ -1,377 +1,238 @@
...
+			Client:                 mgr.GetClient(),
+			Scheme:                 mgr.GetScheme(),
+			CdiClient:              cdiClient,
+			K8sClient:              k8sClient,
+			Log:                    log,
+			UploadProxyServiceName: uploadProxyServiceName,
+			ConfigName:             configName,
+			CDINamespace:           util.GetNamespace(),
+		},
+		MaxConcurrentReconciles: 2, //Equivalent to 'threadiness' in original configuration
```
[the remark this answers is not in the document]
awels: Done

pkg/controller/config-controller.go
awels: Done

pkg/controller/config-controller.go
awels: Types should match now, so deep equal will do the right thing.

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
mhenriks: I see that you got rid of the variable but I assumed that you would somehow be fetching the value from the flag defined in controller runtime per your comment above.  Not just hardcoding "".

This probably isn't the biggest deal in the world.  I've never used the -kubeconfig flag.  But this does seem to be a regression.  Workaround is to use KUBECONFIG env var I guess.

go.mod
```diff
@@ -43,21 +41,19 @@ require (
...
+	k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655
 	k8s.io/client-go v11.0.0+incompatible
 	k8s.io/cluster-bootstrap v0.0.0-20190228181738-e96ff33745e4
 	k8s.io/klog v0.3.1
 	k8s.io/kube-aggregator v0.0.0-20190404125450-f5e124c822d6
 	k8s.io/kube-openapi v0.0.0-20190709113604-33be087ad058
 	kubevirt.io/qe-tools v0.1.3
-	sigs.k8s.io/controller-runtime v0.1.10
-	sigs.k8s.io/testing_frameworks v0.0.0-20190821092519-b6c33f574b58 // indirect
+	sigs.k8s.io/controller-runtime v0.4.0
```
mhenriks: wait, what?  How/why was this file updated?  This is the most recent version of controller runtime.

awels: Don't know, I didn't touch it.

awels:
Ugh of course the OS one failed, no networking ingress available there, since that didn't become active until k8s 1.14. I will have to skip it for OS 3.11

Issue #939, opened 2019-08-23
Block volumes do not smart clone with supported CSI driver


iExalt:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

Block volumes do not smart clone even when all prerequisites are met:
```
Check if Smart-Cloning is possible:
- The source and target PVCs must be in the same namespace
- The source and target PVCs must be in the same Storage Class
- There must be a Snapshot Class associated with the Storage Class
```
(Using the Ceph CSI driver)
However, a standard filesystem volume successfully smart clones.

Filesystem volume:
```
kubectl describe datavolume clear
Name:         clear
Namespace:    default
Labels:       <none>
Annotations:  cdi.kubevirt.io/storage.clone.token:
                eyJhbGciOiJQUzI1NiIsImtpZCI6IiJ9.eyJleHAiOjE1NjY1MjgxMzYsImlhdCI6MTU2NjUyNzgzNiwiaXNzIjoiY2RpLWFwaXNlcnZlciIsIm5hbWUiOiJjbGVhci10ZW1wbGF0Z...
              kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"cdi.kubevirt.io/v1alpha1","kind":"DataVolume","metadata":{"annotations":{},"name":"clear","namespace":"default"},"spec":{"p...
API Version:  cdi.kubevirt.io/v1alpha1
Kind:         DataVolume
Metadata:
  Creation Timestamp:  2019-08-23T02:37:17Z
  Generation:          5
  Resource Version:    52887
  Self Link:           /apis/cdi.kubevirt.io/v1alpha1/namespaces/default/datavolumes/clear
  UID:                 84b2d804-3e5b-449b-b347-ee7f81853fb9
Spec:
  Pvc:
    Access Modes:
      ReadWriteOnce
    Data Source:  <nil>
    Resources:
      Requests:
        Storage:         20Gi
    Storage Class Name:  csi-rbd
  Source:
    Pvc:
      Name:       clear-template
      Namespace:  default
Status:
  Phase:     Succeeded
  Progress:  100.0%
Events:
  Type    Reason           Age                   From                   Message
  ----    ------           ----                  ----                   -------
  Normal  CloneInProgress  10m                   datavolume-controller  Cloning from default/clear-template into default/clear in progress
  Normal  CloneSucceeded   10m                   datavolume-controller  Successfully cloned from default/clear-template into default/clear
[...]

PR #1035, opened 2019-11-24
Test that smart clone is used also in volumeMode: block

maya-r:
Inspired by issue #939
```release-note
NONE
```

### DIFF
--- a/tests/framework/pvc.go
+++ b/tests/framework/pvc.go
@@ -86,32 +86,6 @@ func (f *Framework) CreateAndPopulateSourcePVC(pvcDef *k8sv1.PersistentVolumeCla
 	return sourcePvc
 }
 
-// VerifyTargetPVCContent is used to check the contents of a PVC and ensure it matches the provided expected data
-func (f *Framework) VerifyTargetPVCContent(namespace *k8sv1.Namespace, pvc *k8sv1.PersistentVolumeClaim,
-	expectedData, testBaseDir, testFile string) (bool, error) {
-
-	var dest string
-	var err error
-	var executorPod *k8sv1.Pod
-
-	executorPod, err = utils.CreateExecutorPodWithPVC(f.K8sClient, "verify-pvc-content", namespace.Name, pvc)
-	volumeMode := pvc.Spec.VolumeMode
-	if volumeMode != nil && *volumeMode == k8sv1.PersistentVolumeBlock {
-		dest = testBaseDir
-	} else {
-		dest = testBaseDir + testFile
-	}
-
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
-	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
-	output, err := f.ExecShellInPod(executorPod.Name, namespace.Name, "cat "+dest)
-	if err != nil {
-		return false, err
-	}
-	return strings.Compare(expectedData, output) == 0, nil
-}
-
 // VerifyTargetPVCContentMD5 provides a function to check the md5 of data on a PVC and ensure it matches that which is provided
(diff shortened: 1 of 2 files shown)


tests/framework/pvc.go
```diff
@@ -109,7 +109,8 @@ func (f *Framework) VerifyTargetPVCContent(namespace *k8sv1.Namespace, pvc *k8sv
 	if err != nil {
 		return false, err
 	}
-	return strings.Compare(expectedData, output) == 0, nil
+	outputWithoutNullBytes := strings.Trim(output, "\x00")
```
awels: My problem with this is that we have to store all the \x00 in memory after getting it from the pod, and then we trim it before testing. So if the PVC is large, we will have to store a large number of bytes for no reason. Instead since we know how many bytes the actual data is, we should just grab the actual number of bytes we are interested in, and compare that. Everywhere else in the testing framework we have switched over to VerifyTargetPVCContentMD5 and we pass in the number of bytes we want to check.

Second, it doesn't look like the modified tests actually are calling VerifyTargetContent at all,

tests/smartclone_test.go
```diff
@@ -53,7 +54,29 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
 			// Wait for operation Succeeded
 			waitForDvPhase(cdiv1.Succeeded, dataVolume, f)
 			verifyEvent(controller.CloneSucceeded, dataVolume.Namespace, f)
+			// Cleanup
```
awels: Maybe verify the content is correct here?

tests/smartclone_test.go
```diff
@@ -53,7 +54,29 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
...
+			verifyEvent(controller.SnapshotForSmartCloneInProgress, dataVolume.Namespace, f)
+			// Wait for PVC creation to start
+			waitForDvPhase(cdiv1.SmartClonePVCInProgress, dataVolume, f)
+			verifyEvent(controller.SmartClonePVCInProgress, dataVolume.Namespace, f)
+			// Verify PVC's content
+			verifyPVC(dataVolume, f)
+			// Wait for operation Succeeded
+			waitForDvPhase(cdiv1.Succeeded, dataVolume, f)
+			verifyEvent(controller.CloneSucceeded, dataVolume.Namespace, f)
 			// Cleanup
```
awels: Maybe verify the content is correct here.

tests/smartclone_test.go
```diff
@@ -53,7 +54,29 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
 			// Wait for operation Succeeded
 			waitForDvPhase(cdiv1.Succeeded, dataVolume, f)
 			verifyEvent(controller.CloneSucceeded, dataVolume.Namespace, f)
+			// Cleanup
+			err := utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolume.Name)
```
awels: Since we are improving the tests, lets take the cleanup out of the test itself, and put it in the AfterEach. Right now if the test fails above after creating the DV is not cleaned up. (It will get cleaned up when namespace is deleted by the framework, but lets not count on that working all the time)

1. Add var datavolume *cdiv1.DataVolume or put both datavolume and pvc inside a var ( ... ) block
2. In the AfterEach, check if datavolume != nil and delete it like here.
3. Set it to nil

tests/framework/pvc.go
maya-r: Switched the remaining use, too.

maya-r:
The newer commits should address most of the comments:

- The verification happens in "verifyPVC".
- I didn't switch to NewBlockPVCDefinition, it doesn't seem to gain much (the function is used for both filesystem and block modes right now)

awels:
looks like you forgot the linter? make test-lint

maya-r:
Sorry about that. Should be fixed.

PR #1036, opened 2019-11-25
Split unit and functional test into own script,

awels:
The unit and functional tests were called the same script in the end, and we had an if statement determining which one to call. There really is no reason for it, and now we can combine test-functional and test-functional-ci into the same call, and run that to functional tests.

It combines functional-test and functional-test-ci into one call, and splits unit-test into its own script. This will help with making the unit tests be called from bazel as well as give immediate output during functional test runs.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,13 +58,11 @@ test: test-unit test-functional test-lint
 
 test-unit: WHAT = ./pkg/... ./cmd/...
 test-unit:
-	${DO} "./hack/build/run-tests.sh ${WHAT}"
+	${DO} "./hack/build/run-unit-tests.sh ${WHAT}"
 
 test-functional:  WHAT = ./tests/...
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -58,14 +58,12 @@ test: test-unit test-functional test-lint
 
 test-unit: WHAT = ./pkg/... ./cmd/...
 test-unit:
-	${DO} "./hack/build/run-tests.sh ${WHAT}"
+	${DO} "./hack/build/run-unit-tests.sh ${WHAT}"
 
 test-functional:  WHAT = ./tests/...
-test-functional:
+test-functional: build-functest
```
mhenriks: not important now but I wonder if we can have bazel build the functest binary?

PR #1037, opened 2019-11-25
Push builder image to dockerhub

awels:

This is the first step in using a prebuild builder images. This puts in the code to build and push the container image on merge. Then #997 can be modified the use the builder image.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ before_deploy:
 deploy:
   # Release :latest images
 - provider: script
-  script: make publish
+  script: make builder-push && make publish
   skip_cleanup: true
   on:
     branch: master
--- a/Makefile
+++ b/Makefile
@@ -121,3 +121,6 @@ bazel-push-images: bazel-cdi-generate bazel-build
 
 push: bazel-push-images
 
+builder-push:
+	./hack/build/bazel-build-builder.sh
+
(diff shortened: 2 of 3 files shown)


mhenriks:
Will #997 use a tag or a hash?  How will updates to the builder image be handled?

awels:
#997 It will use a tag. So updates to the builder image will require a new PR, once the PR is merged it will update the container image in dockerhub. The change here in .travis.yml will ensure that a new image is pushed after each PR merge.

mhenriks:
What is the scheme for naming the tags of new builder images?

awels:
That is a good question, I have debated if we should version the builder image or not. I guess if we use shas then we don't care, we just use the sha.

awels:
Thinking about it some more, I think we should go with a hash, that way we can go back to earlier versions if we want, and we don't have to worry about releasing versions for the builder. But that would have to be done in #997 and we will make that a requirement for it to be merged. This is just getting it into dockerhub so #997 can consume it.

mhenriks:
Just so we are on the same page, this is the process for updating the builder image:

0.  PR with changes limited to builder image merged
1.  Get hash of new builder image from dockerhub when travis build completes
2.  Create new PR which uses image from last step

awels:
Yes, so this PR is 0, Adams PR will look up the hash on dockerhub and reference it.

awels:
digest: sha256:38e449bb8f5e2e3dc8ae241323c69ce37e72bd98a109d2032955dd9bd0e0a545 for the builder, but I can't find it in dockerhub for some reason.

awels:
https://hub.docker.com/r/kubevirt/kubevirt-cdi-bazel-builder/tags

PR #1040, opened 2019-11-26
Exclude master and tainted nodes from being used as node for hostpath…

awels:
… provisioner

Exclude master and tainted nodes from the nodes that we can use for annotations used by the hostpath provisioner.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/common.go
+++ b/tests/utils/common.go
@@ -1,7 +1,9 @@
 package utils
 
 import (
+	"fmt"
 	"github.com/onsi/ginkgo"
+	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


mhenriks:
I get excluding NoSchedule, but why masters?  Does this mean you can't use the provisioner with minikube/etc?

awels:
Because in openshift, masters are magically tainting the daemonset set as well [0]. So if my first node is a master, all the tests will fail because its going to try to create the PVCs on the master node which is not running the hostpath provisioner

[0] https://support.coreos.com/hc/en-us/articles/115000625513-Daemonset-only-deploys-to-worker-nodes-

awels:
Note this only affects the functional tests run against a hostpath provisioner.

tests/utils/common.go
```diff
@@ -75,6 +96,9 @@ func IsHostpathProvisioner() bool {
 
 // AddProvisionOnNodeToAnn adds 'kubevirt.io/provisionOnNode' annotaion
 func AddProvisionOnNodeToAnn(pvcAnn map[string]string) {
+	if DefaultNodeName == "" {
+		ginkgo.Skip("No scheduleable nodes available")
```
mhenriks: Do you think it is cleaner to add the skip call to the test function itself rather than a helper?  Is that too cumbersome?

awels: So this is called by 40 or so tests I believe, so doing it individually on each is going to be cumbersome and somewhat error prone.

PR #1044, opened 2019-12-04
Fix DataVolumeSpec's PVCSpec as not pointer

WoohyungHan:
Current generated openapi file(swagger.json) has following lines in DataVolumeSpec, but the PersistentVolumeAccessMode is empty object. It has to be array of string, and of course if we generate a java code with this swagger file, get an error with no PersistentVolumeAccessMode object.

```
   "v1.PersistentVolumeAccessMode": {},
   "v1.PersistentVolumeClaimSpec": {
    "description": "PersistentVolumeClaimSpec describes the common attributes of storage devices and allows a Source for provider-specific attributes",
    "required": [
     "dataSource"
    ],
    "properties": {
     "accessModes": {
      "description": "AccessModes contains the desired access modes the volume should have. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#access-modes-1",
      "type": "array",
      "items": {
       "$ref": "#/definitions/v1.PersistentVolumeAccessMode"
      }
     },
```

This is because our DataVolumeSpec has PVCSpec with a pointer. Fix DataVolumeSpec PVCSpec as not a pointer

NONE

NONE

**Release note**:
```release-note
Fix OpenAPI generated file has empty DataVolumeSpec object
```


mhenriks:
Hi @WoohyungHan thanks for the PR!  I understand that it may make more sense for that member to not be a pointer.  But I'd first like to better understand the problem.  

Your description seems to indicate an issue with java code generated by `swagger.json` but there appear to be no meaningful changes there, just description text.  Which leads me to believe that you are seeing errors on the server side.  Is that correct?  Can you provide example json that demonstrates the issue?  And the context in which it is created?

Regarding a possible solution, changing the PVC member from a pointer to a struct is probably a big enough change to require adding a new api version.  There is existing code that vendors in the DataVolume type.

WoohyungHan:
Hi @mhenriks, Thanks for your review :) I realized that there's no change in swagger.json now, it was my mistake. I think we could solve other direction. and about the issue, I apologize my explanation. We will post more concrete issue soon. Thanks again.

PR #1054, opened 2019-12-12
Update Import controller to use runtime controller library

awels:

Refactor the import controller to use the runtime library.
Fix a permission issue where we want to patch events (to include Nx same) in an event.
Stop using generate name for importer pods as they are guaranteed to be unique in a ns.

Fixes #

**Release note**:

```release-note
Importer pod naming is no longer generated.
```


pkg/controller/import-controller.go
```diff
@@ -44,315 +51,334 @@ const (
...
+			return false
+		},
+	}); err != nil {
+		return err
 	}
-	value, ok := pvc.Annotations[AnnRequiresScratch]
-	if ok {
-		boolVal, _ := strconv.ParseBool(value)
-		scratchRequired = scratchRequired || boolVal
+	if err := importController.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
```
mhenriks: The PVCs that own these pods may not be "interesting"

awels: So as far as I could tell, something that triggers the Pod  watch, which is then translated to a PVC, then goes through the PVC watch, so it will do the check. Maybe I should rename isPVCInteresting, to shouldReconcilePVC?

awels: But I will verify this is the case

awels: You are right, it bypasses the check. I added another check in the reconcile loop

cmd/cdi-controller/controller.go
```diff
@@ -252,7 +242,12 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
 
 func main() {
 	defer klog.Flush()
-	logf.SetLogger(logf.ZapLogger(false))
+	debug := false
+	if i, err := strconv.Atoi(verbose); err == nil && i > 1 {
+		debug = true
+	}
+	logf.SetLogger(logf.ZapLogger(debug))
```
mhenriks: nice

pkg/controller/import-controller.go
mhenriks: Now it is checked in both places.  Maybe it is clearer to just do it in the reconcile loop?

Also, I'm not sure that the check that was added to the reconcile loop is in the right place.  Why is it not called right after getting the PVC, like maybe here:  https://github.com/kubevirt/containerized-data-importer/blob/c9b741974233f120f8eef35649cbe75c96b334fc/pkg/controller/import-controller.go#L158

pkg/controller/util.go
```diff
@@ -384,7 +384,8 @@ func MakeImporterPodSpec(image, verbose, pullPolicy string, podEnvVar *importPod
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: podName,
```
mhenriks: If no longer using GenerateName, then `findImporterPod` should be refactored to simply do a get rather than list

pkg/controller/import-controller.go
awels: Let me take another look at what I am doing in the reconcile loop, there is some trickiness when I delete a pod triggering a reconcile and causing it to create a new importer pod which results in an endless loop of create/delete pods. But that was happening before I did the check. So I had to delete in the next reconcile loop, but if I have the check I can do both in the same loop.

pkg/controller/util.go
awels: Good point done.

pkg/controller/import-controller.go
awels: So the problem turned out to be in checkPVC, it would return false if the termination was not nil, which works in most cases, except where the PVC is removed externally and then it fails. So once fixed, I was to make a clean reconcile loop.

pkg/controller/clone-controller.go
```diff
@@ -240,7 +240,7 @@ func (cc *CloneController) syncPvc(key string) error {
 		return nil
 	}
 
-	if !checkPVC(pvc, AnnCloneRequest) || metav1.HasAnnotation(pvc.ObjectMeta, AnnCloneOf) {
+	if !checkPVC(pvc, AnnCloneRequest) || pvc.DeletionTimestamp != nil || metav1.HasAnnotation(pvc.ObjectMeta, AnnCloneOf) {
```
mhenriks: How did this change to clone controller come up?

pkg/controller/import-controller.go
```diff
@@ -44,315 +53,536 @@ const (
...
+		OwnerType:    &corev1.PersistentVolumeClaim{},
+		IsController: true,
+	}, predicate.Funcs{
+		DeleteFunc: func(e event.DeleteEvent) bool {
```
mhenriks: Shouldn't we be handling deletes?  If someone deletes an import pod, we want to recreate it, right?

pkg/controller/clone-controller.go
awels: Because I changed checkPVC to not include the check for the timestamp, I didn't want to alter the (potentially broken) behavior of the clone, so I had to add that back in this call.

pkg/controller/import-controller.go
awels: Good point, I think I added that to stop the loop that would keep creating a new importer pod once it finished, because we would delete the completed pod. It should not be needed anymore now that I have solved that.

awels: Yeah don't need this, removed.

pkg/controller/import-controller.go
```diff
@@ -44,315 +51,532 @@ const (
...
+	return importController, nil
+}
 
-	if len(podList) == 0 {
-		return nil, nil
-	} else if len(podList) > 1 {
-		return nil, errors.Errorf("multiple pods found for import PVC %s/%s", pvc.Namespace, pvc.Name)
+func addImportControllerWatches(mgr manager.Manager, importController controller.Controller) error {
+	// Add schemes.
+	if err := corev1.AddToScheme(mgr.GetScheme()); err != nil {
```
mhenriks: I don't think this is necessary.

pkg/controller/import-controller.go
```diff
@@ -44,315 +51,532 @@ const (
...
-		return errors.Errorf("found pod %s/%s not owned by pvc %s/%s", pod.Namespace, pod.Name, pvc.Namespace, pvc.Name)
-	}
+	return r.reconcilePvc(pvc, pod)
+}
 
-	// expectations prevent us from creating multiple pods. An expectation forces
-	// us to observe a pod's creation in the cache.
-	needsSync := ic.podExpectations.SatisfiedExpectations(pvcKey)
+func (r *ImportReconciler) findImporterPod(pvc *corev1.PersistentVolumeClaim) (*corev1.Pod, error) {
+	podName := fmt.Sprintf("%s-%s", common.ImporterPodName, pvc.Name)
```
mhenriks: pod name is constructed in this manner in a couple spots so should probably be a function (that takes pvc as param).  Should prob do the same for scratch pvc name.

pkg/controller/import-controller.go
```diff
@@ -44,315 +51,532 @@ const (
...
-	if exists && (previousPhase == string(v1.PodSucceeded)) {
-		needsSync = false
+	if k8serrors.IsNotFound(err) {
+		return nil, nil
 	}
 
-	if pod == nil && needsSync {
-		return ic.createImporterPod(pvc, pvcKey)
+	// Check if the PVC owns the pod.
+	pvcOwns := false
```
mhenriks: maybe call `metav1.IsControlledBy` here?

pkg/controller/import-controller.go
```diff
@@ -44,315 +51,532 @@ const (
...
-			} else {
-				ic.recorder.Event(pvc, v1.EventTypeWarning, ErrImportFailedPVC, pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Message)
+func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, pod *corev1.Pod) (reconcile.Result, error) {
+	if pod == nil {
+		if isPVCComplete(pvc) {
+			// Don't create the POD if the PVC is completed already
+			r.Log.V(1).Info("PVC is already complete")
+		} else {
+			// Create importer pod, make sure the PVC owns it.
+			if err := r.createImporterPod(pvc); err != nil {
```
mhenriks: what if `pvc.DeletionTimestamp != nil`?

pkg/controller/import-controller.go
```diff
@@ -44,315 +51,532 @@ const (
...
-	if err != nil {
-		return errors.WithMessage(err, "could not update pvc %q annotation and/or label")
+	if !reflect.DeepEqual(currentPvcCopy, pvc) {
+		if err := r.updatePVC(pvc); err != nil {
+			return err
+		}
+		r.Log.V(1).Info("Updated PVC", "pvc.anno.Phase", anno[AnnPodPhase])
+	}
+
+	if pvc.GetAnnotations()[AnnPodPhase] == string(corev1.PodSucceeded) || scratchExitCode {
```
mhenriks: maybe call `isPVCComplete()` instead of `pvc.GetAnnotations()[AnnPodPhase] == string(corev1.PodSucceeded)`

awels: done

pkg/controller/import-controller.go
awels: Good point, shouldn't be creating pods if the pvc is getting ready to be deleted.

pkg/controller/import-controller.go
awels: Was copying what was there, but this is better.

pkg/controller/import-controller.go
awels: Good suggestion, done.

pkg/controller/import-controller.go
```diff
@@ -44,315 +51,532 @@ const (
...
+		pvc.GetAnnotations()[AnnPodPhase] = string(corev1.PodSucceeded)
+		if err := r.updatePVC(pvc); err != nil {
+			return reconcile.Result{}, errors.WithMessage(err, fmt.Sprintf("could not update pvc %q annotation and/or label", pvc.Name))
+		}
+		return reconcile.Result{}, nil
 	}
-	pvcKey, err := cache.MetaNamespaceKeyFunc(pvc)
+
+	// See if we have a pod associated with the PVC, we know the PVC has the needed annotations.
+	pod, err := r.findImporterPod(pvc)
```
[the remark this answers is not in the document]
awels: Yeah, artefact of me trying to break up the large function into smaller pieces. Done

pkg/controller/clone-controller.go
awels: Note, I am looking at rewriting the clone controller as well, and in my initial investigation, I think that will get refactored away. Trying to figure out how to delete the source pod, once the upload server has deleted the target pod.

pkg/controller/import-controller.go
awels: You are right.

mhenriks:
/lgtm
/approve
👍

PR #1055, opened 2019-12-13
Fix to create temporary pod with default resource quota

### LINKED ISSUE #1043: Creating a DataVolume in the namespace having resource quota

/kind bug
/kind enhancement

--------------------
**What happened**:

When I create dataVolume in the specific namespace, which has resourceQuota limits, the dataVolume created, but the requested image does not imported. (i tested images with spec.source.registy & spec.source.http both.)

I found error messages occurred in 'cdi-deployment pod log', which said:
```
importer-controller.go:296] error processing pvc "{$pvcName}" : importer pod API create errored: pods "importer-{$pvcName}-{$arbitraryCharacters}" is forbidden: failed quota: {$resourceQuotaName}: must specify limts.cpu, limits.memory
```
In such namespace and resourceQuota, I specified limits.cpu and limits.memory.

By inspecting source codes, i found that the function "CreateImporterPod" use pvc.Namespace to make Importer pod.
However, in the pod.spec section, it seems that there isn't any part inserting resourceQuota and its limits.
But, when creating any pod in namespace having some resourcequota, the section "spec.containers.resources.limits" is required.

Thus, the importer pod didn't created and also pvc-scratch didn't created.
Therefore importing image didn't succeeded, so the Status.Phase of DataVolume remained in "PVCBound", not "Succeeded".

--------------------
Is this situation does not considered?
or Is there any solution that i'm missing to make importer-pod and datavolume in namespace having resourcequota?

--------------------
1. create namespace
2. create resourcequota in that namespace
3. create datavolume in that namespace

--------------------

--------------------
**Environment**:
  - Tested on both version:
    - v1.11.0
    - v1.10.9
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
- Install tools:
- Others:
  - OS version : ubuntu 18.04.3 LTS

woohhan:
If we create a temporary pod in the namespace which has resource quota, `must specify quota` error occurs. Fix to create temporary pod with default resource quota.

Fixes #1043

for now, I just added for the ImporterPod only. Could you please take a look? After the review, I will add for the CloneSourcePod and UploadPod.

**Release note**:
```release-note
- Fix to apply default quota if namespace does not have a quota
```

### DIFF
--- a/pkg/apis/core/v1alpha1/types.go
+++ b/pkg/apis/core/v1alpha1/types.go
@@ -256,14 +256,16 @@ type CDIConfig struct {
 
 //CDIConfigSpec defines specification for user configuration
 type CDIConfigSpec struct {
-	UploadProxyURLOverride   *string `json:"uploadProxyURLOverride,omitempty"`
-	ScratchSpaceStorageClass *string `json:"scratchSpaceStorageClass,omitempty"`
+	UploadProxyURLOverride   *string                      `json:"uploadProxyURLOverride,omitempty"`
+	ScratchSpaceStorageClass *string                      `json:"scratchSpaceStorageClass,omitempty"`
+	PodResourceRequirements  *corev1.ResourceRequirements `json:"podResourceRequirements,omitempty"`
 }
 
 //CDIConfigStatus provides
 type CDIConfigStatus struct {
-	UploadProxyURL           *string `json:"uploadProxyURL,omitempty"`
-	ScratchSpaceStorageClass string  `json:"scratchSpaceStorageClass,omitempty"`
+	UploadProxyURL                 *string                      `json:"uploadProxyURL,omitempty"`
+	ScratchSpaceStorageClass       string                       `json:"scratchSpaceStorageClass,omitempty"`
+	DefaultPodResourceRequirements *corev1.ResourceRequirements `json:"defaultPodResourceRequirements,omitempty"`
 }
 
 //CDIConfigList provides the needed parameters to do request a list of CDIConfigs from the system
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"crypto/rsa"
 	"fmt"
+	clientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"strconv"
 	"time"
 
@@ -54,11 +55,13 @@ type CloneController struct {
 	Controller
 	recorder       record.EventRecorder
 	tokenValidator token.Validator
+	cdiClient      clientset.Interface
 }
 
 // NewCloneController sets up a Clone Controller, and returns a pointer to
 // to the newly created Controller
 func NewCloneController(client kubernetes.Interface,
+	cdiClientSet clientset.Interface,
 	pvcInformer coreinformers.PersistentVolumeClaimInformer,
 	podInformer coreinformers.PodInformer,
 	image string,
@@ -77,6 +80,7 @@ func NewCloneController(client kubernetes.Interface,
 		Controller:     *NewController(client, pvcInformer, podInformer, image, pullPolicy, verbose),
 		recorder:       recorder,
 		tokenValidator: newCloneTokenValidator(apiServerKey),
+		cdiClient:      cdiClientSet,
 	}
 	return c
 }
@@ -161,7 +165,8 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 		}
 
 		cc.raisePodCreate(pvcKey)
-		sourcePod, err = CreateCloneSourcePod(cc.clientset, cc.image, cc.pullPolicy, clientName, pvc)
+
+		sourcePod, err = CreateCloneSourcePod(cc.clientset, cc.cdiClient, cc.image, cc.pullPolicy, clientName, pvc)
 		if err != nil {
 			cc.observePodCreate(pvcKey)
 			return err
--- a/pkg/controller/config-controller.go
+++ b/pkg/controller/config-controller.go
@@ -2,6 +2,8 @@ package controller
 
 import (
 	"context"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"reflect"
 
 	routev1 "github.com/openshift/api/route/v1"
@@ -72,6 +74,10 @@ func (r *CDIConfigReconciler) Reconcile(req reconcile.Request) (reconcile.Result
 	}
 
 	if err := r.reconcileStorageClass(config); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err := r.reconcileDefaultPodResourceRequirements(config); err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -153,6 +159,41 @@ func (r *CDIConfigReconciler) reconcileStorageClass(config *cdiv1.CDIConfig) err
 	log.Info("No default storage class found, setting scratch space to blank")
 	// No storage class found, blank it out.
 	config.Status.ScratchSpaceStorageClass = ""
+	return nil
+}
+
+func (r *CDIConfigReconciler) reconcileDefaultPodResourceRequirements(config *cdiv1.CDIConfig) error {
+	config.Status.DefaultPodResourceRequirements = &v1.ResourceRequirements{
+		Limits: map[v1.ResourceName]resource.Quantity{
+			v1.ResourceCPU:    *resource.NewQuantity(0, resource.DecimalSI),
+			v1.ResourceMemory: *resource.NewQuantity(0, resource.DecimalSI)},
+		Requests: map[v1.ResourceName]resource.Quantity{
+			v1.ResourceCPU:    *resource.NewQuantity(0, resource.DecimalSI),
+			v1.ResourceMemory: *resource.NewQuantity(0, resource.DecimalSI)},
+	}
+
+	if config.Spec.PodResourceRequirements != nil {
+		if config.Spec.PodResourceRequirements.Limits != nil {
+			if cpu, exist := config.Spec.PodResourceRequirements.Limits[v1.ResourceCPU]; exist {
+				config.Status.DefaultPodResourceRequirements.Limits[v1.ResourceCPU] = cpu
+			}
+
+			if memory, exist := config.Spec.PodResourceRequirements.Limits[v1.ResourceMemory]; exist {
+				config.Status.DefaultPodResourceRequirements.Limits[v1.ResourceMemory] = memory
+			}
+		}
+
+		if config.Spec.PodResourceRequirements.Requests != nil {
+			if cpu, exist := config.Spec.PodResourceRequirements.Requests[v1.ResourceCPU]; exist {
+				config.Status.DefaultPodResourceRequirements.Requests[v1.ResourceCPU] = cpu
+			}
+
+			if memory, exist := config.Spec.PodResourceRequirements.Requests[v1.ResourceMemory]; exist {
+				config.Status.DefaultPodResourceRequirements.Requests[v1.ResourceMemory] = memory
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -303,7 +303,8 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 	}
 
 	// all checks passed, let's create the importer pod!
-	pod, err := createImporterPod(r.Log, r.Client, r.Image, r.Verbose, r.PullPolicy, podEnvVar, pvc, scratchPvcName)
+	pod, err := createImporterPod(r.Log, r.Client, r.CdiClient, r.Image, r.Verbose, r.PullPolicy, podEnvVar, pvc, scratchPvcName)
+
 	if err != nil {
 		return err
 	}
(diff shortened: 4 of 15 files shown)


cluster-up/cluster/okd-4.3/provider.sh
```diff
@@ -41,7 +41,7 @@ function up() {
     # Copy k8s config and kubectl
     cluster_container_id=$(docker ps -f "name=$provider_prefix-cluster" --format "{{.ID}}")
 
-    _install_from_cluster $cluster_container_id /usr/bin/oc 0755 .kubectl
+    _install_from_cluster $cluster_container_id /usr/local/bin/oc 0755 .kubectl
```
awels: I don't think this is part of what you are changing, it probably got generated when you ran make deps-update. Its a bug in kubevirtci because the oc executable in in /usr/bin in the okd 4.3 provider.

pkg/controller/util.go
```diff
@@ -331,12 +332,58 @@ func GetScratchPvcStorageClass(client kubernetes.Interface, cdiclient clientset.
...
+// If the ResourceRequirements values are specified in CDIConfig, these values are taken and used.
+// If not, 0 is used as default values.
+func GetPodResourceRequirements(cdiclient clientset.Interface) (*v1.ResourceRequirements, error) {
+	config, err := cdiclient.CdiV1alpha1().CDIConfigs().Get(common.ConfigName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Unable to find CDI configuration, %v\n", err)
+		return nil, err
+	}
+
+	cpuLimit := resource.NewQuantity(0, resource.DecimalSI)
```
awels: Am I missing something here, can't you simply return config.Status.DefaultPodResourceRequirements (once that is added)?
If you don't want to return the same section of memory and want a copy you can do:
```go
if config.Spec.PodResourceRequirements != nil {
  return &config.Spec.PodResourceRequirements.DeepCopy(), nil
} else {
  return nil, nil
}
```

pkg/controller/util.go
```diff
@@ -420,6 +467,7 @@ func MakeImporterPodSpec(image, verbose, pullPolicy string, podEnvVar *importPod
 							Protocol:      v1.ProtocolTCP,
 						},
 					},
+					Resources: *resourceRequirements,
```
awels: This is a potential null pointer, I would add a section below like this:
```go
if resourceRequirements != nil {
  pod.Spec.Containers[0].Resources = *resourceRequirements
}
```
This way we will only add it, if we now that it is not nil. and we will only dereference it if we know it is not nil.

pkg/controller/util_test.go
```diff
@@ -690,14 +691,14 @@ func TestCreateImporterPod(t *testing.T) {
 	}{
 		{
 			name:    "expect pod to be created for PVC with VolumeMode Filesystem",
-			args:    args{k8sfake.NewSimpleClientset(pvc), "test/image", "-v=5", "Always", &importPodEnvVar{"", "", "", "", "1G", "", false}, pvc},
-			want:    MakeImporterPodSpec("test/image", "-v=5", "Always", &importPodEnvVar{"", "", "", "", "1G", "", false}, pvc, nil),
+			args:    args{k8sfake.NewSimpleClientset(pvc), "test/image", "-v=5", "Always", &importPodEnvVar{"", "", "", "", "1G", "", false}, pvc, nil},
+			want:    MakeImporterPodSpec("test/image", "-v=5", "Always", &importPodEnvVar{"", "", "", "", "1G", "", false}, pvc, nil, nil),
```
awels: You are passing nil as the resourceRequirements here, this will cause the null pointer error I described above. Also we have test MakeImporterPodSpec that needs to be updated to match.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -258,6 +258,7 @@ type CDIConfig struct {
 type CDIConfigSpec struct {
 	UploadProxyURLOverride   *string `json:"uploadProxyURLOverride,omitempty"`
 	ScratchSpaceStorageClass *string `json:"scratchSpaceStorageClass,omitempty"`
+	PodResourceRequirements *corev1.ResourceRequirements `json:"podResourceRequirements,omitempty"`
```
awels: So this will be the override right? The pattern we have been using in the CDIConfig object is the following.

The status reflects what the import/upload/clone controllers will use when creating the pods. The spec contains optional overrides. So I am missing a 
'DefaultPodResourceRequirements' in the CDIConfigStatus struct. then we need to add something to the config-controller reconcile loop, that checks if the PodResourceRequirements exists, if it does, copy that to the status.DefaultPodResourceRequirements, if it doesn't then the status.DefaultPodResourceRequirements should be our 'reasonable default values'.

cluster-up/cluster/okd-4.3/provider.sh
anencore94: Yes, you are right. we didn't change this file intentionally. Thanks for your explanation, we will fix it.

pkg/apis/core/v1alpha1/types.go
anencore94: >So this will be the override right?

Yes, that's the object we want to use for creating temporary pods.

>The status reflects what the import/upload/clone controllers will use when creating the pods. 

Before your review, we thought that importer/upload/clone controllers could use the values directly from CDIConfigSpec, not CDIConfigStatus.

So if I understand right, you mean that
1. There should be some functions that insert 'reasonable default values' in status.DefaultPodResourceRequirements when deploying cdi modules.
2. Also, in config-controller reconcile loop, when spec.PodResourceRequirements is filled, loop always copy that to the status.DefaultPodResourceRequirements.
3. At 'pkg/controller/util.go', we only use the values in status.DefaultPodResourceRequirements.

Is it right?

pkg/controller/util.go
anencore94: We decided to use Status.DefaultPodResourceRequirements.Limits.Cpu(),  Status.DefaultPodResourceRequirements.Limits.Memory(), Status.DefaultPodResourceRequirements.Requests.Cpu(), Status.DefaultPodResourceRequirements.Requests.Memory() respectively
rather than to use Status.DefaultPodResourceRequirements.DeepCopy() directly.

Since someone could insert only limits.cpu in Spec.PodResourceRequirements, when the ResourceQuota requires both limits.cpu and limits.memory. Then importer pod won't create as we want.
Thus we thought we should nil-check for every PodResourceRequirements.XXX.XXX() components. Is there any idea that I'm missing?

awels: So, that is why we want you to use cdiconfig.status.PodResourceRequirements, since the status is generated by the config-controller (from the spec.PodResourceRequirements and default values encoded in the controller), it will always have all values set to something, and thus you can always copy it because it is guaranteed to always be fully populated.

pkg/apis/core/v1alpha1/types.go
awels: > So if I understand right, you mean that
> 
>     1. There should be some functions that insert 'reasonable default values' in status.DefaultPodResourceRequirements when deploying cdi modules.
> 
This will be the config controller I think. We do something similar for the scratchspace and proxyurl already. If those don't exist in the spec, do some default behavior.


>     2. Also, in config-controller reconcile loop, when spec.PodResourceRequirements is filled, loop always copy that to the status.DefaultPodResourceRequirements.
> 

Yes, if the spec if filled copy those values into the status, and if there is a partial fill, then the rest will be the default values.

>     3. At 'pkg/controller/util.go', we only use the values in status.DefaultPodResourceRequirements.

Correct.
> 
> 
> Is it right?

pkg/controller/util.go
anencore94: If we use cdiconfig.status.PodResourceRequirements, which is always be fully populated, then *resourceRequirements will not be a null pointer, right?

pkg/controller/util.go
anencore94: Ok i got it. Thanks :)

pkg/controller/util.go
awels: Since we cannot guarantee that whomever calls MakeImporterPodSpec will always set that to something, IMO its better to be a bit defensive and deal with the case where someone didn't set it. But you are right in the context of this patch, that should never be nil.

pkg/controller/util_test.go
WoohyungHan: DONE

pkg/controller/util.go
WoohyungHan: DONE

WoohyungHan:
@awels We updated the commit with including your suggestion. Could you please re-check? :)

pkg/controller/clone-controller.go
```diff
@@ -161,7 +163,13 @@ func (cc *CloneController) processPvcItem(pvc *v1.PersistentVolumeClaim) error {
 		}
 
 		cc.raisePodCreate(pvcKey)
-		sourcePod, err = CreateCloneSourcePod(cc.clientset, cc.image, cc.pullPolicy, clientName, pvc)
+
+		podResourceRequirements, err := GetDefaultPodResourceRequirements(cc.cdiClient)
```
awels: So this is more a stylistic thing than a code problem. podResourceRequirements is looked up here, just so we can pass it to CreateCloneSourcePod. I think it makes sense to look up the podResourceRequirements in CreateCloneSourcePod, but that will require us to pass cc.cdiClient to CreateCloneSourcePod.

pkg/controller/config-controller.go
```diff
@@ -156,6 +162,51 @@ func (r *CDIConfigReconciler) reconcileStorageClass(config *cdiv1.CDIConfig) err
...
+		Requests: map[v1.ResourceName]resource.Quantity{
+			v1.ResourceCPU:    *resource.NewQuantity(0, resource.DecimalSI),
+			v1.ResourceMemory: *resource.NewQuantity(0, resource.DecimalSI)},
+	}
+
+	if config.Spec.PodResourceRequirements != nil {
+		// TODO: Override
+		log.Info("Overriding default pod resource requirements from spec if spec is not null")
+		if config.Spec.PodResourceRequirements.Limits != nil {
+			cpu, exist := config.Spec.PodResourceRequirements.Limits[v1.ResourceCPU]
```
awels: Another stylistic comment, code is fine:
```go
if cpu, exist := config.Spec.PodResourceRequirements.Limits[v1.ResourceCPU]; exists {
  config.Status.DefaultPodResourceRequirements.Limits[v1.ResourceCPU] = cpu
}
```
We tend to like the if err := something; err != nil { pattern. Against its stylistic, code is fine.

The other 3 have the same.

pkg/controller/import-controller.go
```diff
@@ -250,9 +250,14 @@ func (ic *ImportController) createImporterPod(pvc *v1.PersistentVolumeClaim, pvc
 		return err
 	}
 
+	podResourceRequirements, err := GetDefaultPodResourceRequirements(ic.cdiClient)
```
awels: Same as clone controller, I think it makes sense to look up the pod requirements in the function that builds the spec.

pkg/controller/upload-controller.go
```diff
@@ -476,7 +476,12 @@ func (c *UploadController) getOrCreateUploadPod(pvc *v1.PersistentVolumeClaim, p
 			ClientName:     clientName,
 		}
 
-		pod, err = CreateUploadPod(args)
+		podResourceRequirements, err := GetDefaultPodResourceRequirements(c.cdiClient)
```
awels: Same as clone controller, I think it makes sense to look up the pod requirements in the function that builds the spec.

pkg/controller/clone-controller.go
anencore94: Thanks for your comments :) We updated commit applying this style

pkg/controller/config-controller.go
anencore94: Thanks for your comments :) We updated commit applying this style

pkg/controller/import-controller.go
anencore94: DONE

pkg/controller/upload-controller.go
anencore94: DONE

woohhan:
We fixed the problem with the test :) @awels

pkg/controller/config-controller.go
```diff
@@ -156,6 +162,47 @@ func (r *CDIConfigReconciler) reconcileStorageClass(config *cdiv1.CDIConfig) err
 	return nil
 }
 
+func (r *CDIConfigReconciler) reconcileDefaultPodResourceRequirements(config *cdiv1.CDIConfig) error {
+	log := r.Log.WithName("CDIconfig").WithName("DefaultPodResourceRequirements")
+
+	// TODO: set as default value (0)
```
awels: I think this TODO is done, I also think the log is not needed as its normal operation.

pkg/controller/config-controller.go
```diff
@@ -156,6 +162,47 @@ func (r *CDIConfigReconciler) reconcileStorageClass(config *cdiv1.CDIConfig) err
...
+		Limits: map[v1.ResourceName]resource.Quantity{
+			v1.ResourceCPU:    *resource.NewQuantity(0, resource.DecimalSI),
+			v1.ResourceMemory: *resource.NewQuantity(0, resource.DecimalSI)},
+		Requests: map[v1.ResourceName]resource.Quantity{
+			v1.ResourceCPU:    *resource.NewQuantity(0, resource.DecimalSI),
+			v1.ResourceMemory: *resource.NewQuantity(0, resource.DecimalSI)},
+	}
+
+	if config.Spec.PodResourceRequirements != nil {
+		// TODO: Override
```
awels: I think this TODO is done, I also think the log is not needed as its normal operation.

pkg/controller/config-controller.go
woohhan: Done!

pkg/controller/config-controller.go
woohhan: Done!

awels:
/lgtm
/approve
Verified this is working with some functional tests that I will create a PR for.

PR #1061, opened 2019-12-20
Support custom authentication headers

pkliczewski:
Till now http source supported only basic authentication with this PR
we can provide authentication header name. With this approach we enable
importing images from glance using `X-Auth-Token` header name.

Signed-off-by: Piotr Kliczewski <piotr.kliczewski@gmail.com>

### DIFF
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -68,16 +68,18 @@ type HTTPDataSource struct {
 	customCA bool
 	// the content length reported by the http server.
 	contentLength uint64
+	// custom authentication header
+	headerName string
 }
 
 // NewHTTPDataSource creates a new instance of the http data provider.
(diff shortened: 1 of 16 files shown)


pkg/importer/http-datasource.go
```diff
@@ -112,6 +115,10 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 		klog.Errorf("Error creating readers: %v", err)
 		return ProcessingPhaseError, err
 	}
+	// special case to support glance
```
mdbooth: This isn't actually a workaround to support glance: it's applicable to any HTTP source which requires a custom header, because we can't pass that header to qemu-img, and therefore we need to download directly with curl.

This is the same whether glance supports or doesn't support byte ranges (which it doesn't).

pkliczewski: You are right will remove this comment. TNX

pkliczewski:
Let's close it for now

PR #1075, opened 2020-01-08
Update clone controller to use runtime controler library.

awels:

Update the clone controller to use the controller runtime library

Fixes #

**Release note**:

```release-note
NONE
```


awels:
/hold
Need to add some more unit tests.

mhenriks:
I will wait until #1079 is merged to fully review but so far so good.

Can we get rid of `pkg/controller/controller.go` now?

awels:
Agreed, I will likely need to fix a few things with the new version of the libraries. Also good question on the pkg/controller/controller.go our current unit test coverage on that is 0 at this point. I am not sure if any of the remaining controllers (datavolume/smart clone/upload) even use that code at all.

awels:
So there were a few constants in the controller.go file, but nothing that couldn't be moved, similar in the test file, a couple of things that got moved.

pkg/controller/clone-controller.go
```diff
@@ -203,108 +230,382 @@ func (cc *CloneController) waitTargetPodRunningOrSucceeded(pvc *v1.PersistentVol
...
+	id := string(pvc.GetUID())
+	url := GetUploadServerURL(pvc.Namespace, pvc.Name)
+	pvcOwner := metav1.GetControllerOf(pvc)
+	if pvcOwner != nil && pvcOwner.Kind == "DataVolume" {
+		ownerID = string(pvcOwner.UID)
+	}
+
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: podName,
```
mhenriks: I believe using GenerateName without the protection of something like expectations can get you multiple pods.   But I'm not exactly sure how.  Maybe it's safe.  Otherwise, some thought would have to go into how to get a deterministic name here.  Not as simple as import/upload case because there could theoretically be multiple concurrent clones with the same source.  Which is something we should probably be testing if we aren't.

pkg/controller/clone-controller.go
```diff
@@ -45,112 +54,126 @@ const (
...
 	}
+	return cloneController, nil
+}
 
-	if len(podList) > 1 {
-		return nil, errors.Errorf("multiple source pods found for clone PVC %s/%s", pvc.Namespace, pvc.Name)
+// addConfigControllerWatches sets up the watches used by the config controller.
+func addCloneControllerWatches(mgr manager.Manager, cloneController controller.Controller) error {
+	// Setup watches
+	if err := cloneController.Watch(&source.Kind{Type: &corev1.PersistentVolumeClaim{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
mhenriks: you forgot to watch pods

awels: https://github.com/kubevirt/containerized-data-importer/pull/1075/files/d3d13953a54c55c2ebb0bc61e0b7a8a3d4d20dfb#diff-c604e18ed9a6be7674e4881621e1cf6bR109 is where I am watching pods.

pkg/controller/clone-controller.go
awels: We do have a functional test for that: https://github.com/kubevirt/containerized-data-importer/blob/master/tests/cloner_test.go#L395
When we look up the source pod, we have a label selector that uses the UID of the TARGET pvc, and thus it doesn't matter if we have multiple source pods, the lookup will find only the one we want based on the label selector.

mhenriks: I understand how that works but I think it is possible to create a pod then list pods and not get the pod you just created.  Especially with controller runtime client which is cached.

awels: I get what you are saying, it is possible to list (for instance in a different reconcile loop) before the create completed, and thus you won't find the pod, and create another one. The only way I can think of to get a unique deterministic name is to use the target pvc name and target pvc namesapce. If we just do source-<tgt_name>-<tgs_ns> its likely we will run into the name length limit of the pod pretty quickly. So I would prefer to do something else, maybe a hash of those two?

awels: Decided to do a hash, I updated the PR.

pkg/controller/clone-controller.go
```diff
@@ -45,112 +56,126 @@ const (
 	// APIServerPublicKeyPath is the path to the apiserver public key
 	APIServerPublicKeyPath = APIServerPublicKeyDir + "/id_rsa.pub"
 
+	// CloneSucceededPVC provides a const to indicate an import to the PVC failed
```
mhenriks: I think this comment needs some work

pkg/controller/clone-controller.go
```diff
@@ -45,112 +56,126 @@ const (
...
 )
 
-// CloneController represents the CDI Clone Controller
-type CloneController struct {
-	Controller
+// CloneReconciler members
+type CloneReconciler struct {
+	Client         client.Client
+	Scheme         *runtime.Scheme
+	K8sClient      kubernetes.Interface
```
mhenriks: not opposed to having these client-go clients around but may want to refactor the code that requires them

awels: So my plan is to remove the client-go clients once all the controllers have been converted, and I can refactor any utility functions that use them to use the runtime lib client.

pkg/controller/clone-controller.go
awels: Yes, fixed it.

pkg/controller/clone-controller.go
```diff
@@ -234,108 +245,376 @@ func (cc *CloneController) waitTargetPodRunningOrSucceeded(pvc *v1.PersistentVol
...
-	return ValidateCanCloneSourceAndTargetSpec(&sourcePvc.Spec, &targetPvc.Spec)
+	log.V(1).Info("cloning source pod (image) created\n", "pod.Namespace", pod.Namespace, "pod.Name", pod.Name, "image", image)
+
+	return pod, nil
 }
 
-//Run is being called from cdi-controller (cmd)
-func (cc *CloneController) Run(threadiness int, stopCh <-chan struct{}) error {
-	cc.Controller.run(threadiness, stopCh, cc.runPVCWorkers)
+func getCloneSourcePodName(pvc *corev1.PersistentVolumeClaim) string {
```
mhenriks: You are attempting to create a unique name for the source pod but `findCloneSourcePod` is still looking up by label value.  I initially thought this was not ideal.  But it is good for backward compatibility.

I'm personally not crazy about hashes/UIDs being the the entirety of a resource name.  See the names of replicasets created by deployments for example.  Hash is only part of the name.  But if you want to do something quick and dirty here, I think I would prefer the target PVC UID in place of the hash.  Basically use the `CloneUniqueID` label value.

awels: Okay, the goal was a deterministic unique label that would not overflow the max name length. If you prefer the CloneUniqueID over this, that works for me. I will make this function generate that string.

pkg/controller/clone-controller_test.go
```diff
@@ -37,236 +46,549 @@ func (cg *fakeCertGenerator) MakeServerCert(namespace, service string, duration
...
+	It("Should update the cloneof when complete, block mode", func() {
+		testPvc := createBlockPvc("testPvc1", "default", map[string]string{
+			AnnCloneRequest: "default/source", AnnPodReady: "true", AnnCloneToken: "foobaz", AnnUploadClientName: "uploadclient"}, nil)
+		reconciler = createCloneReconciler(testPvc, createBlockPvc("source", "default", map[string]string{}, nil))
+		By("Setting up the match token")
+		reconciler.tokenValidator.(*FakeValidator).match = "foobaz"
+		reconciler.tokenValidator.(*FakeValidator).Name = "source"
+		reconciler.tokenValidator.(*FakeValidator).Namespace = "default"
+		reconciler.tokenValidator.(*FakeValidator).Params["targetNamespace"] = "default"
+		reconciler.tokenValidator.(*FakeValidator).Params["targetName"] = "testPvc1"
```
aglitke: I've lost count how many times I've seen these exact same 5 lines.  In a follow-up PR can you please refactor and reuse this code?

aglitke:
@awels Please see my request for a follow-up PR.

PR #1076, opened 2020-01-09
Added functional tests for ResourceRequirements feature.

awels:
Added documentation for ResourceQuota.

Add functional tests for the Resource Request/Limits feature.
Add documentation.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/doc/quota.md
+++ b/doc/quota.md
@@ -1,0 +1,27 @@
+# CDI ResourceQuota
+Containerized Data Importer(CDI) needs to work well in environments that have [ResourceQuota](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/) defined in a namespace. CDI launches pods for various operations suchs as import, clone, and upload. These pods will fail to start in a namespace with a ResourceQuota if the pod does not define proper resource requests and limits.
+
+By default all pods created by CDI will have a cpu/memory request/limit of 0, which basically tells kubernetes to do it best but there are no particular requirements on the pods. This will ensure that the pods created by CDI have the best chance of starting and completing in a ResourceQuota limited environment.
+
+## Override defaults
+There might be circumstances where the admin wants to override this behavior and configure requests and limits of a particular value for CDI. The admin can update the spec section of the cdi config to override the defaults.
+Example CDIConfig yaml
+```yaml
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: CDIConfig
+metadata:
+  labels:
+    app: containerized-data-importer
+    cdi.kubevirt.io: ""
+  name: config
+spec: {
+    PodResourceRequirements:
+    limits:
+      cpu: "4"
+      memory: "1Gi"
+    requests:
+      cpu: "1"
+      memory: "250Mi"
+}
+```
+Once the CDIConfig object is updated, the status section of the object will reflect that values that will be used to pass to the pods. [limits and requests](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#motivation-for-default-memory-limits-and-requests) are explained in the kubernetes documentation.
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -185,9 +185,11 @@ var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
 		Expect(err).ToNot(HaveOccurred())
 
-		targetPvc, err := utils.WaitForPVC(f.K8sClient, dataVolume.Namespace, dataVolume.Name)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(targetPvc.Spec.VolumeName).To(Equal(targetPV.Name))
+		Eventually(func() bool {
+			targetPvc, err = utils.WaitForPVC(f.K8sClient, dataVolume.Namespace, dataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+			return targetPvc.Spec.VolumeName == targetPV.Name
+		}, 60, 1).Should(BeTrue())
 
 		fmt.Fprintf(GinkgoWriter, "INFO: wait for PVC claim phase: %s\n", targetPvc.Name)
 		utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, f.Namespace.Name, v1.ClaimBound, targetPvc.Name)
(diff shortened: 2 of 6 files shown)


awels:
@woohhan Here are some documenation and functional tests for the ResourceQuota feature.

woohhan:
Hey @awels, Thanks for keep taking a look and writing functional tests and documents!

tests/cloner_test.go
```diff
@@ -521,6 +521,107 @@ var _ = Describe("Block PV Cloner Test", func() {
...
+		sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile)
+		// Create targetPvc in new NS.
+		targetDV := utils.NewCloningDataVolume("target-dv", "1G", pvcDef)
+		dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = utils.WaitForPVC(f.K8sClient, dataVolume.Namespace, dataVolume.Name)
+		Expect(err).ToNot(HaveOccurred())
+		By("Verify Quota was exceeded in logs")
+		matchString := fmt.Sprintf("upload pod API create errored: pods \"cdi-upload-target-dv\" is forbidden: exceeded quota: test-quota, requested: requests.cpu=2, used: requests.cpu=0, limited: requests.cpu=1")
```
maya-r: It might bite us in the future if we're hard-coding the full error string, if they change it even slightly

doc/quota.md
```diff
@@ -0,0 +1,27 @@
+# CDI ResourceQuota
```
maya-r: Alternative suggestion:

Containerized Data Importer (CDI)'s utility pods request resource limits of 0 ("unlimited").  
This means that they will fail to run in namespaces that have a [ResourceQuota](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/).  

In order to run in such environments, or impose limits on CDI for other reasons, an admin can adjust the spec section of the CDIConfig:  
-- yaml --

tests/cloner_test.go
awels: So its the test that is creating the name that we are checking for. I like to have a different mechanism (in this case hard coding) than what the actual code is doing for checking value against. That way if there is an issue with the mechanism that generates the value, the test doesn't pass because both result and match used the same mechanism to generate the value.

tests/framework/framework.go
```diff
@@ -440,6 +441,67 @@ func (f *Framework) CreatePrometheusServiceInNs(namespace string) (*v1.Service,
...
+		Limits: map[v1.ResourceName]resource.Quantity{
+			v1.ResourceCPU:    *resource.NewQuantity(limitsCPU, resource.DecimalSI),
+			v1.ResourceMemory: *resource.NewQuantity(limitsMemory, resource.DecimalSI)},
+	}
+	_, err = f.CdiClient.CdiV1alpha1().CDIConfigs().Update(config)
+	if err != nil {
+		return err
+	}
+	time.Sleep(1 * time.Second)
+	// Make sure the CDIConfig is updated before continuing
```
aglitke: Wow, I really wish the configmap update api was synchronous!

tests/utils/pod.go
```diff
@@ -79,6 +81,14 @@ func NewPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1
 					Name:    "runner",
 					Image:   "kubevirt/cdi-importer:latest",
 					Command: []string{"/bin/sh", "-c", cmd},
+					Resources: k8sv1.ResourceRequirements{
```
aglitke: This doesn't consult CDIConfig.  Is that going to be a problem?

tests/framework/framework.go
awels: This sleep might not actually be enough, it depends on how well the cluster is behaving, but the 1 seconds seems enough in most circumstances. Definitely potential to be flake.

tests/utils/pod.go
awels: This is for verifier pods and things of that nature, its not actually part of CDI.

tests/upload_test.go
```diff
@@ -281,3 +282,146 @@ var _ = Describe("Block PV upload Test", func() {
...
+		f.CreateQuotaInNs(int64(1), int64(1024*1024*1024), int64(2), int64(2*1024*1024*1024))
+		By("Creating PVC with upload target annotation")
+		pvc, err = f.CreatePVCFromDefinition(utils.UploadPVCDefinition())
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Set up port forwarding")
+		uploadProxyURL, portForwardCmd, err = startUploadProxyPortForward(f)
+		Expect(err).ToNot(HaveOccurred())
+		By("Verify PVC annotation says ready")
+		found, err := utils.WaitPVCPodStatusReady(f.K8sClient, pvc)
```
mhenriks: The quota only effects whether the pod gets created, right?  So why do the rest of this?

awels: True, I just wanted it to do a complete operation while inside the quota, but you are right technically we just need to verify the pod is created.

tests/cloner_test.go
```diff
@@ -521,6 +521,107 @@ var _ = Describe("Block PV Cloner Test", func() {
...
+	})
+
+	It("Should create clone in namespace with quota", func() {
+		f.CreateQuotaInNs(int64(1), int64(1024*1024*1024), int64(2), int64(2*1024*1024*1024))
+		smartApplicable := f.IsSnapshotStorageClassAvailable()
+		sc, err := f.K8sClient.StorageV1().StorageClasses().Get(f.SnapshotSCName, metav1.GetOptions{})
+		if err == nil {
+			value, ok := sc.Annotations["storageclass.kubernetes.io/is-default-class"]
+			if smartApplicable && ok && strings.Compare(value, "true") == 0 {
+				Skip("Cannot test host assisted cloning for within namespace when all pvcs are smart clone capable.")
```
mhenriks: Smart/dumb cloning should be irrelevant to quota, right?

tests/cloner_test.go
```diff
@@ -521,6 +521,107 @@ var _ = Describe("Block PV Cloner Test", func() {
...
+			}
+		}
+
+		pvcDef := utils.NewPVCDefinition(sourcePVCName, "1G", nil, nil)
+		pvcDef.Namespace = f.Namespace.Name
+		sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile)
+		doFileBasedCloneTest(f, pvcDef, f.Namespace)
+	})
+
+	It("Should fail to clone in namespace with quota when pods have higher requirements", func() {
```
mhenriks: Should we consider the case where we cannot create a pod in one of source/target namespace when cross namespace cloning?

tests/cloner_test.go
awels: No because smart cloning doesn't spawn pods. So we can only test host assisted cloning.

tests/cloner_test.go
awels: Good question, might be an interesting test

tests/cloner_test.go
mhenriks: But the new PVC may cause a quota violation, right?  Is that tested?

mhenriks:
Do we support the following flow and if so, do we have a functional test anywhere?

1.  Import/clone/upload get in error state because of pod/pvc quota issue
2.  Quota adjusted
3.  Operation succeeds

tests/cloner_test.go
awels: Not on the memory or cpu limits that we are talking about here. There might be storage quotas but those are out of scope for this.

awels:
We should have a test for that, I don't think I included that here though, but basically the pod will remain pending until there is enough quota.

tests/framework/framework.go
awels: So to detail the problem:
Above we update the resource requirements in the CDIConfig object, and below we attempt to verify those changes are actually there. However it appears to be cached because the below will immediately return success, but in reality the CDIConfig object might not have been updated yet. This will cause the resource request/limit to be 0 on the pod, and thus it will successfully launch when it should fail. The timeout is to give the cluster some time to actually propagate the values to the actual object.

PR #1080, opened 2020-01-13
Don't run raw images through qemu-img as it sometimes has issues tran…

awels:
…sferring files.

qemu-img sometimes has issues transferring data, don't call qemu-img if we are transferring a raw file.
Fixes #

**Release note**:

```release-note
Don't call QEMU-IMG on transferring a raw disk image.
```

### DIFF
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -114,7 +114,7 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	}
 	// The readers now contain all the information needed to determine if we can stream directly or if we need scratch space to download
 	// the file to, before converting.
-	if !hs.readers.Archived && !hs.customCA {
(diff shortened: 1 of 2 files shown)


mhenriks:
Will this cause scratch space to be used for importing raw images?

What do we know about the "issues" seen with qemu-img?  Would it be better to look into updating qemu-img?  I know we had an issue with a more recent version.  But maybe it's fixed?

awels:
> Will this cause scratch space to be used for importing raw images?
> 
> What do we know about the "issues" seen with qemu-img? Would it be better to look into updating qemu-img? I know we had an issue with a more recent version. But maybe it's fixed?

No, this will stream a RAW image directly to the target file or block device. No scratch space and no qemu-img

awels:
Oh and as an added bonus it will now also work with http sources that don't support range.

PR #1081, opened 2020-01-13
Explain better what is going on when trying to allocate maximum space…

awels:
… available in PVC.

Increased number of retries for larger PVCs.

The information provided when fallocating resizing the disk image was not clear in what was going on. Also the number of retries was insufficient for most cases.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -137,8 +137,8 @@ func RetryBackoffSize(dest string, size resource.Quantity, sizeFunc func(string,
 		retryCount++
 		// Reduce by 25 blocks, and try again. How did we end up with 25 blocks? Well we need to leave some blocks for the fs on the pvc.
 		// Then we need to leave some extra blocks so it can mount inside the container.
-		reduceSize.Sub(*resource.NewScaledQuantity(blockSize*int64(25), 0))
-		klog.V(3).Infof("Trying new size: %s\n", reduceSize.String())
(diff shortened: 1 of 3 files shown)


pkg/util/util.go
```diff
@@ -133,12 +133,12 @@ func RetryBackoffSize(dest string, size resource.Quantity, sizeFunc func(string,
 	err = sizeFunc(dest, size)
 	retryCount := 0
 	reduceSize := *size.Copy()
-	for err != nil && retryCount < 10 {
+	for err != nil && retryCount < 100 {
```
mhenriks: Kind of not a fan of "if you don't succeed, try, try again."  Especially when retrying on any error and not something specific.

I also like being "more safe than sorry."   Think we're still cutting things a little too close here.  I would personally be inclined to just expand the image up to a certain threshold, like 90-95% of the disk.  Expect that the user gives us reasonably sized disks.  If there is extra space, help them out up to a point.  I'm afraid of shooting them in the foot later on.

Assuming you want to stay with this method, maybe an exponential backoff would be better?  Start with 25 or whatever, then double it after each failure?

awels: If we do exponential, we don't need this many tries. I am just trying to make sure we give the user the most amount of available space in the guest.

mhenriks: I care more about avoiding failures and I personally think we should still be more conservative here.  But maybe @aglitke or others feel differently.

awels:
Super seeded by #1140

PR #1090, opened 2020-01-23
Update datavolume controller to use runtime controller library

awels:

Update the data volume controller to use the controller runtime library.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/controller/datavolume-controller.go
```diff
@@ -797,48 +649,45 @@ func (c *DataVolumeController) getPodFromPvc(namespace string, pvcUID types.UID)
...
-func (c *DataVolumeController) updateProgressUsingPod(dataVolumeCopy *cdiv1.DataVolume, pod *corev1.Pod) {
+func updateProgressUsingPod(dataVolumeCopy *cdiv1.DataVolume, pod *corev1.Pod) error {
 	httpClient := buildHTTPClient()
 	// Example value: import_progress{ownerUID="b856691e-1038-11e9-a5ab-525500d15501"} 13.45
 	var importRegExp = regexp.MustCompile("progress\\{ownerUID\\=\"" + string(dataVolumeCopy.UID) + "\"\\} (\\d{1,3}\\.?\\d*)")
 
-	port, err := c.getPodMetricsPort(pod)
+	port, err := getPodMetricsPort(pod)
 	if err == nil {
 		url := fmt.Sprintf("https://%s:%d/metrics", pod.Status.PodIP, port)
```
mhenriks: while we're here:  maybe we should not attempt to get status if `pod.Status.PodIP == ""`.  This spams the logs and confuses QA.

awels: Good point, I will add that check to only attempt if we have a pod ip.

pkg/controller/datavolume-controller.go
```diff
@@ -682,6 +686,10 @@ func updateProgressUsingPod(dataVolumeCopy *cdiv1.DataVolume, pod *corev1.Pod) e
 	return err
 }
 
+func errConnectionRefused(err error) bool {
+	return strings.Contains(err.Error(), "connection refused")
```
mhenriks: Spent a bit of time looking for a better way to check this specific error but I think this is the way to go.

PR #1091, opened 2020-01-26
cert rotation

mhenriks:

Currently certs expire after a year.  This fixes that by having the operator create/refresh certs.  The apiserver and uploadproxy mount certs as secrets and watch for updates to update TLSConfig appropriately.

The operator also creates/manages CA certs for upload server/client.

Fixes #

The apiserver doesn't support non disruptive update of their CA certs.  This would be a pretty easy add on, though.  Have to watch CA bundles and reregister apiserver when it changes.

**Release note**:

```release-note
Cert rotation
```


mhenriks:
/hold

Still want to think on a couple things

maya-r:
Having the private and public keys files in the same directory and accessing it by mounting the directory probably means that the pods verifying keys can also impersonate that key.
Is this a concern?

maya-r:
It would be nice to write more of the design somewhere, at least in the commit message
for example:

keys are stored on the master node's filesystem in
/var/run/certs/cdi-uploadproxy-server-cert
... several other places ...
readers watch them for change using fsnotify
writers modify the file

mhenriks:
>(I might be misreading the code...) Having the private and public keys files in the same directory and accessing it by mounting the directory probably means that the pods verifying keys can also impersonate that key.
Is this a concern?

@maya-r Not quite sure I understand your concern.  But the paths we are protecting are as follows.  And in no case does the client have direct access to the server key.  In ll cases, verification is done via the CA bundles which are stored in configmaps and separate from the keys/certs stored in secrets.

upload proxy -> upload server: upload proxy has upload server CA bundle configmap mounted
clone source -> upload server:  clone source has upload server CA bundle passed in as env var
kube apiserver -> cdi apiserver:  kube apiserver get apiserver CA bundle in apiserver and webhook registrations
openshift outer -> upload proxy: upload proxy CA bundle is included Route resource

go.mod
```diff
@@ -61,6 +64,7 @@ require (
 replace (
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20191219222812-2987a591a72c
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191125132246-f6563a70e19a
+	github.com/openshift/library-go => github.com/mhenriks/library-go v0.0.0-20200116194830-9fcc1a687a9d
```
mhenriks: using fork until this is merged:  https://github.com/openshift/library-go/pull/540

aglitke: If we merge this could you create a gh issue so we don't forget to back to the trunk?

pkg/operator/controller/predicate.go
```diff
@@ -0,0 +1,79 @@
...
+
+package controller
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// LeaderAlectionAnnotation is the annotatation used on resources for leader election
```
aglitke: LeaderElectionAnnotation?

mhenriks: lol yup

PR #1093, opened 2020-01-30
Add NFS as KUBEVIRT_STORAGE to cluster-sync

awels:

Adds NFS storage to the CI test matrix.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/nfs/nfs-server.yaml
+++ b/cluster-sync/nfs/nfs-server.yaml
@@ -1,0 +1,51 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: nfs-data
+spec:
+  storageClassName: local
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 30Gi
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nfs-server
+  labels:
+    app: "nfs-server"
+spec:
+  containers:
+  - image: itsthenetwork/nfs-server-alpine:12
+    imagePullPolicy: IfNotPresent
+    name: nfs-server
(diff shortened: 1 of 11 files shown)


cluster-sync/nfs/nfs-server.yaml
```diff
@@ -0,0 +1,53 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nfs-server
+  namespace: cdi
+  labels:
+    app: "nfs-server"
+spec:
+  containers:
+  - image: itsthenetwork/nfs-server-alpine:latest
```
aglitke: Do we want to select a specific version to avoid being unexpectedly broken?

tests/framework/nfs-utils.go
```diff
@@ -0,0 +1,60 @@
...
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+const (
+	timeout         = time.Second * 90
+	pollingInterval = time.Second
+)
+
+func createNFSPVs(client *kubernetes.Clientset, cdiNs string) error {
+	for i := 1; i <= 5; i++ {
```
aglitke: 5 should be a constant probably.

cluster-sync/nfs/nfs-server.yaml
awels: I can do a specific tag, but there is always a chance we will get broken unexpectedly due to changes to the tag. But let me change it to the specific tag.

tests/framework/nfs-utils.go
awels: Done.

aglitke:
lgtm

PR #1102, opened 2020-02-06
Disable caching when converting qcow2 images

aglitke:
When converting images the qemu-img command uses a writeback cache mode
by default.  This means that writes to a block device go through the
host page cache and are lazily flushed to the storage device.  When
using this cache mode, the process can exit before all writes have
reached storage and our DataVolume could appear ready to use before all
I/O has completed.  This becomes a problem with shared storage because a
second host does not have the benefit of the first host's page cache
when reading.  To prevent this problem we must perform I/O directly to
the storage device.  This behavior is selected by passing "-t" "none" to
qemu-img.

This problem will not arise with "create", "info", or "resize" qemu-img
commands because those are metadata operations.

The following document provides some good background on qemu cache modes:
https://documentation.suse.com/sles/11-SP4/html/SLES-kvm4zseries/cha-qemu-cachemodes.html

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -99,7 +99,7 @@ func NewQEMUOperations() QEMUOperations {
 }
 
 func convertToRaw(src, dest string) error {
-	_, err := qemuExecFunction(nil, nil, "qemu-img", "convert", "-p", "-O", "raw", "-o", "preallocation=falloc", src, dest)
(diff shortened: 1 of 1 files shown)


mhenriks:
Cache should really be disabled at the RBD level for RWX volumes.  https://docs.ceph.com/docs/master/rbd/rbd-config-ref/#cache-settings

No idea how to enable that with rook, though

PR #1104, opened 2020-02-10
Add opt-in protection against deleting CDI CR if DVs exist

maya-r:
It isn't immediately obvious to users that deleting the CDI CR will delete all their data volumes, and we want to protect them from this case.
    
This behaviour requires opting in, for compatibility sake.

**Release note**:
```release-note
Add opt-in protection against deleting CDI CR if DVs exist, opt-in by adding the following to the CDI CR:
    spec:
      uninstallStrategy: BlockUninstallIfWorkloadsExist
```


pkg/operator/controller/controller.go
```diff
@@ -93,8 +95,15 @@ func newReconciler(mgr manager.Manager) (*ReconcileCDI, error) {
...
 
+	cfg := mgr.GetConfig()
+	cdiClient, err := cdiClientset.NewForConfig(cfg)
+	if err != nil {
+		log.Error(err, "Error building cdi clientset")
+	}
+
 	r := &ReconcileCDI{
 		client:         mgr.GetClient(),
+		cdiClientset:   cdiClient,
```
awels: You don't need the cdiClientSet, the runtime library client will register the CDI types, and you can use that in the operator. We currently have a duality with using 2 separate clients:
1. client-go (which requirest the cdiClientset
2. controller runtime-library (used in operator, cdiconfig, and I have several PRs pending to convert the other controllers)

So we are converging on 2, and thus I would encourage you to use the runtime library client instead of the client-go one. This will also safe you from having to create a fake cdiClientSet in the tests.

pkg/operator/controller/controller.go
```diff
@@ -160,6 +170,16 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 
 	// mid delete
 	if cr.DeletionTimestamp != nil {
+		waitForDVs, err := r.waitForDVs(reqLogger, cr)
+		if err != nil {
+			return reconcile.Result{Requeue: true, RequeueAfter: 5 * time.Second}, err
```
awels: Any error will automatically requeue the object. So no need to say requeue after 5 seconds.

pkg/operator/controller/controller.go
```diff
@@ -160,6 +170,16 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+		}
+
+		if waitForDVs {
+			reqLogger.Info("Not deleting as long as DVs exist")
+			return reconcile.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil
```
awels: Just so I understand, we are rechecking every 5 seconds if all the dvs are gone, once the CR is marked for deletion?

pkg/operator/controller/controller.go
```diff
@@ -519,11 +539,47 @@ func (r *ReconcileCDI) isMutable(obj runtime.Object) bool {
...
+			break
+		}
+	}
+
+	if i < 0 {
+		// No finalizer, not protecting against DV deletion
+		return false, nil
+	}
+
+	dv, err := r.cdiClientset.CdiV1alpha1().DataVolumes(metav1.NamespaceAll).List(metav1.ListOptions{Limit: 2})
```
awels: Same using runtime library
```go
dvList := &cdiv1.DataVolumeList{}
if err := r.Client.List(context.TODO(), dvList, &client.ListOptions{Limit: 2}); err != nil {
  logger.Error(err, "Error finding datavolumes")
  return true, err
}
if len(dv.Items) > 0 {
  return true, nil
}
```

pkg/operator/controller/controller.go
```diff
@@ -57,7 +58,8 @@ import (
 )
 
 const (
-	finalizerName = "operator.cdi.kubevirt.io"
+	cleanupResourcesFinalizer = "operator.cdi.kubevirt.io"
```
awels: Any particular reason you renamed this?

mhenriks:
Finalizers are only meant to be added by controllers.  Users should opt into this feature by setting a field in the spec of the CDI CR.

pkg/operator/controller/controller.go
mhenriks: Yeah this comment got me thinking.  Turns out that it is not possible to list all datavolumes in all namespaces with the operator client.  As of #1091 It has a cache bound to the install namespace.  This broke something.  But I fixed it in #1106.  Maya should use the uncached client created in that PR.

pkg/operator/controller/controller.go
maya-r: Yes

pkg/operator/controller/controller.go
maya-r: I am adding a second finalizer, and "finalizerName" is a name that could refer to both. I thought it'd be more clear this way.

maya-r: After changes, this change was not necessary.

maya-r:
I rebased on top of #1107.
I'm still in the progress of writing the functional test awels suggested, but figured I'd share the code in case anyone has more suggestions for changes.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -195,8 +195,22 @@ type CDI struct {
 // CDISpec defines our specification for the CDI installation
 type CDISpec struct {
 	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty" valid:"required"`
+
+	// Specifies if CDI can be deleted if workloads are still present.
+	// This is mainly a precaution to avoid accidental data loss
+	UninstallStrategy CDIUninstallStrategy `json:"uninstallStrategy,omitempty"`
```
mhenriks: This should be a `*CDIUninstallStrategy` as it is optional

pkg/operator/controller/controller.go
awels: You don't need the Requeue part, because you are returning an error:
```go
```

maya-r:
This still doesn't contain the functional test, but otherwise should be good.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -195,8 +195,22 @@ type CDI struct {
...
+	// This is mainly a precaution to avoid accidental data loss
+	UninstallStrategy *CDIUninstallStrategy `json:"uninstallStrategy,omitempty"`
 }
 
+// CDIUninstallStrategy defines whether to protect from deleting data in DataVolumes
+type CDIUninstallStrategy string
+
+const (
+	// CDIUninstallStrategyRemoveWorkloads means that DataVolumes may be deleted during CDI removal
+	CDIUninstallStrategyRemoveWorkloads CDIUninstallStrategy = "RemoveWorkloads"
```
mhenriks: `Workloads` seems like an odd word to use here.  Do we mention "workloads" anywhere else?

pkg/apis/core/v1alpha1/types.go
```diff
@@ -195,8 +195,22 @@ type CDI struct {
...
+	// CDIUninstallStrategyBlockUninstallIfWorkloadsExist means that CDI deletion will block if DataVolumes exist
+	CDIUninstallStrategyBlockUninstallIfWorkloadsExist CDIUninstallStrategy = "BlockUninstallIfWorkloadsExist"
```
mhenriks: `BlockUninstallIfWorkloadsExist` is a bit of a mouthful.  I wonder if there are more simple phrases we can use.  Maybe have a `dataVolumeRetentionPolicy` property with `Delete` and `Block` as values?

awels:
What about a webhook to reject delete when the protection is enabled. As discussed in the mailing list without the webhook, we are just delaying the delete, not preventing it.

mhenriks:
The more I think about it, the more I don't like the state the operator gets into here while waiting for DataVolumes to be deleted.  All CDI components are still running but not protected by the operator.  For example, someone could delete a webhook registration and it would not get recreated.  Also, what if someone deletes CDI by mistake and wants to reinstall?  They'll have to delete all their DataVolumes first.

Has the option of just leaving the DataVolume CRD around been discussed?  The operator simply has to remove the owner reference.  And it should already be able to deal with an existing DataVolume CRD in a reinstall scenario.

maya-r:
Since we're second to the party I was following along with the ideas in https://github.com/kubevirt/kubevirt/pull/2976 for consistency sake.

maya-r:
Rebased since #1107 was merged (I previously had it as one of the commits).

maya-r:
@rmohr / @fabiand have you discussed the option of removing the ownerReferences on kubevirt CRDs that we don't want deleted as an alternative to "uninstallStrategy"?

> The more I think about it, the more I don't like the state the operator gets into here while waiting for DataVolumes to be deleted. All CDI components are still running but not protected by the operator. For example, someone could delete a webhook registration and it would not get recreated. Also, what if someone deletes CDI by mistake and wants to reinstall? They'll have to delete all their DataVolumes first.
> 
> Has the option of just leaving the DataVolume CRD around been discussed? The operator simply has to remove the owner reference. And it should already be able to deal with an existing DataVolume CRD in a reinstall scenario.

rmohr:
> @rmohr / @fabiand have you discussed the option of removing the ownerReferences on kubevirt CRDs that we don't want deleted as an alternative to "uninstallStrategy"?

At least  in kubevirt/kubevirt we use a DELETE webhook, which ensures that the deployment does not get deleted, to exactly not have the problems discussed here. If workloads exist, we simply don't ally uninstallation in contrast to delaying it.

rmohr:
> @rmohr / @fabiand have you discussed the option of removing the ownerReferences on kubevirt CRDs that we don't want deleted as an alternative to "uninstallStrategy"?

Sorry did not answer the question with the owner reference. For insance in kubevirt/kubevirt we don't use owner references. But yes, an "orphan delete" would be interesting. However, it would basically make the installation unmonitored and you will have to add rediscover-logic, if you were relying on the references before. You  would also have the issue that over time you installation may degrade in functionality if you don't install the operator again. I agree that this coud be an option, however since the only reason we do the protection, is to not by accident destroy workloads or create unwanted downtime, I personally would vote for the delete webhook, since it seems to best fit the requirements.

mhenriks:
I find the webhook approach preferable to what is currently implemented in this PR.

maya-r:
Abandoning in favour of #1118 which uses a webhook.

PR #1109, opened 2020-02-11
Copy kubevirt.io annotations to scratch space pvc

awels:
but NOT cdi.kubevirt.io as those are what triggers import/upload/clone.

In general we do not want to copy over annotations from the original PVC to the scratch space PVC. For instance if provisioners use annotations do to special things to a PVC. However in certain situations we would like to copy them if the annotations are part of kubevirt.io. This PR copies kubevirt.io annotations from the PVC to the scratch space PVC, but not cdi.kubevirt.io because those are the annotations that trigger import/upload/clone and thus would cause the scratch space to be imported/uploaded/cloned to.

We want to copy the kubevirt.io/provisionOn annotation for the host path provisioner, so we can guarantee that the scratch space is on the same node if that annotation has been used. Otherwise there is a race condition between binding a PVC and scheduling a pod to determine which node the pod will run on. 

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	crdv1alpha1 "github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/controller/util.go
```diff
@@ -240,11 +240,22 @@ func newScratchPersistentVolumeClaimSpec(pvc *v1.PersistentVolumeClaim, pod *v1.
 		LabelImportPvc:   pvc.Name,
 	}
 
+	annotations := make(map[string]string, 0)
+	// Copy kubevirt.io annotations, but NOT the CDI annotations as those will trigger another import/upload/clone on the scratchspace
+	// pvc.
+	if len(pvc.GetAnnotations()) > 0 {
+		for k, v := range pvc.GetAnnotations() {
+			if strings.Contains(k, common.KubeVirtAnnKey) && !strings.Contains(k, common.CDIAnnKey) {
```
mhenriks: This is probably fine, but I just wonder:  if we only really care about `kubevirt.io/provisionOn` why not just limit to that?

PR #1111, opened 2020-02-11
keep go.mod and vendor up to date

mhenriks:

We are using `go mod` for dependency management but not using straight `go build` for building.  If we were, `go.mod` would be updated whenever we build.  So simulating that with this PR.

Fixes #

If `kubevirt-cdi-volume-bazel-server` container is running it, kill it.  The first build will take awhile as all deps are downloaded.  But future builds will be fast.

**Release note**:

```release-note
NONE
```


mhenriks:
/hold

Have to think about implications on CI build.  Have to make sure not including anything from GOPATH

PR #1118, opened 2020-02-19
Uninstall strategy and blocking webhook

mhenriks:

Quick and dirty CDI implementation of what KubeVirt is doing here:  https://github.com/kubevirt/kubevirt/pull/2976

The webhook should really live in cdi-operator.  But I guess this mostly works because the webhook only handles deletes.

Fixes #

**Release note**:

```release-note
Add uninstallStrategy to CDI resource and webhook to block CDI deletion when appropriate.
```


pkg/apis/core/v1alpha1/types.go
```diff
@@ -227,6 +240,9 @@ const (
 
 	// CDIPhaseUpgrading signals that the CDI resources are being deployed
 	CDIPhaseUpgrading CDIPhase = "Upgrading"
+
+	// CDIPhaseEmpty is an unitinialized phase
```
maya-r: uninitialized

maya-r:
With the release note I feel that it's unclear whether this feature is on by default or not. Perhaps we can use the release note in #1104 
````
Add opt-in protection against deleting CDI CR if DVs exist, opt-in by adding the following to the CDI CR:
    spec:
      uninstallStrategy: BlockUninstallIfWorkloadsExist
````
/lgtm
/retest

File: pkg/operator/resources/cluster/apiserver.go [hunk 2]
```diff
@@ -96,10 +97,10 @@ func getAPIServerClusterPolicyRules() []rbacv1.PolicyRule {
 				"cdi.kubevirt.io",
 			},
 			Resources: []string{
-				"*",
+				"datavolumes",
 			},
 			Verbs: []string{
-				"*",
+				"list",
 			},
```
maya-r: We intend to backport the commit to a release immediately, and I'm slightly concerned about unintended effects from this change. Perhaps we should have it as a separate change that isn't backported.

mhenriks: Not too concerned about this as the cdi client was not even imported before this PR.

maya-r:
Seems reasonable for me, but I am hesitating to be the one to approve it.
- What's weird in the weird resource in pkg/apiserver/webhooks/cdi-validate_test.go?
- having to pass cdiClient reminds me of your first comment about how we should be able to use the uncached client for this purpose, and I don't need a separate cdiClient
 How come this doesn't hold here?

mhenriks:
> What's weird in the weird resource in pkg/apiserver/webhooks/cdi-validate_test.go?

The wehook is expecting a CDI not a DataVolume.  Should never happen in practice

>having to pass cdiClient reminds me of your first comment about how we should be able to use the uncached client for this purpose, and I don't need a separate cdiClient
How come this doesn't hold here?

Not quite sure I understand the question.  The client that was added here is in cdi-apiserver.  I think you are confusing it with cdi-operator.

maya-r:
This failure looks like a real issue. All the tests fail after the "should block CDI delete" test fails on CDI being missing.

PR #1119, opened 2020-02-20
Move base images to fedora 31.

awels:

We current have our base images as fedora 29, which is no longer maintained, moving to more recent 31.
Fixes #

**Release note**:

```release-note
Container base  images are now fedora 31
```


awels:
Pushed the image to be on quay.io now. Hopefully fewer 503s.

awels:
Need to investigate the one failure, its always the same test that fails on NFS. Locally it works fine.

awels:
/hold
Need to figure out what to do with the one disabled test that always fails on NFS. @aglitke any thoughts. Its the registry tar one that only fails on NFS.

PR #1122, opened 2020-02-24
Eliminate the need for HEAD requests in the Go HTTP download client.

maya-r:
Some servers reject such requests, and if Content-Length is available,
it's going to be sent in GET requests as well.

Some servers reject HEAD requests but allow GET, this causes 403 Failed errors.
Example: https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img fails (this is also listed as one of our examples).

This is one step towards a solution but doesn't actually solve the problem.
It's not complete because later we invoke qemu-img on the same URL and it suffers from the same limitation.

A complete solution will probably involve saving the file to disk and passing the saved file to qemu-img, but I don't want to block others until I finish writing that.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -271,7 +271,7 @@ func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certD
 		return nil, uint64(0), true, errors.Errorf("expected status code 200, got %d. Status: %s", resp.StatusCode, resp.Status)
 	}
 
-	if (total == 0) {
+	if total == 0 {
 		// The total seems bogus. Let's try the GET Content-Length header
 		total = parseHTTPHeader(resp)
 	}
@@ -342,7 +342,7 @@ func getContentLength(client *http.Client, ep *url.URL, accessKey, secKey string
 	return total, nil
 }
 
-func parseHTTPHeader(resp *http.Response) (uint64) {
+func parseHTTPHeader(resp *http.Response) uint64 {
 	var err error
 	total := uint64(0)
 	if val, ok := resp.Header["Content-Length"]; ok {
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -348,7 +348,7 @@ func parseHTTPHeader(resp *http.Response) (uint64) {
 	if val, ok := resp.Header["Content-Length"]; ok {
 		total, err = strconv.ParseUint(val[0], 10, 64)
 		if err != nil {
-			klog.Errorf("could not convert content length, got %d", err)
+			klog.Errorf("could not convert content length, got %v", err)
 		}
 		klog.V(3).Infof("Content length: %d\n", total)
 	}
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -66,6 +66,8 @@ type HTTPDataSource struct {
 	url *url.URL
 	// true if we are using a custom CA (and thus have to use scratch storage)
 	customCA bool
+	// true if we know `qemu-img` will fail to download this
+	brokenForQemuImg bool
 	// the content length reported by the http server.
 	contentLength uint64
 }
@@ -77,7 +79,7 @@ func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType 
 		return nil, errors.Wrapf(err, fmt.Sprintf("unable to parse endpoint %q", endpoint))
 	}
 	ctx, cancel := context.WithCancel(context.Background())
-	httpReader, contentLength, err := createHTTPReader(ctx, ep, accessKey, secKey, certDir)
+	httpReader, contentLength, brokenForQemuImg, err := createHTTPReader(ctx, ep, accessKey, secKey, certDir)
 	if err != nil {
 		cancel()
 		return nil, err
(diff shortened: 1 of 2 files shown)


maya-r:
I went with not failing if Content-Length doesn't exist.
It felt like a wiser idea to have nil rather than 0 as the value for "content length unknown" so did some reshuffling for that.

The test is using a garbage content length rather than not adding one - it turns out Go will add a Content-Length even if you don't ask for it.

awels:
> I went with not failing if Content-Length doesn't exist.
> It felt like a wiser idea to have nil rather than 0 as the value for "content length unknown" so did some reshuffling for that.
> 
Can you elaborate on that? A total of 0 is a clear indication of we don't know the total. With the pointer we now have to deal with potential null pointers vs always having some kind of value.

I also think we should keep the HEAD attempt because if it fails, we know that doing a streaming conversion from qemu-img will also fail, because that does a HEAD as well? With the HEAD attempt we know it failed, and can switch to using scratch space instead of a streaming conversion. If we just get the length from the GET, we do not have that information and as you said the streaming conversion will still fail.

So I would extract the code that looks at the header of the response into its own method, and call that method from both the HEAD attempt and the GET if we don't have the content length yet. I would also take the knowledge we gain from the failed HEAD and use that to direct to use scratch space instead of a streaming conversion.

maya-r:
> > I went with not failing if Content-Length doesn't exist.
> > It felt like a wiser idea to have nil rather than 0 as the value for "content length unknown" so did some reshuffling for that.
> 
> Can you elaborate on that? A total of 0 is a clear indication of we don't know the total. With the pointer we now have to deal with potential null pointers vs always having some kind of value.
> 

I got the impression that we prefer pointers for "Maybe value" from [this discussion](https://github.com/kubevirt/containerized-data-importer/pull/1104#discussion_r377776605).

> I also think we should keep the HEAD attempt because if it fails, we know that doing a streaming conversion from qemu-img will also fail, because that does a HEAD as well? With the HEAD attempt we know it failed, and can switch to using scratch space instead of a streaming conversion. If we just get the length from the GET, we do not have that information and as you said the streaming conversion will still fail.
> 
> So I would extract the code that looks at the header of the response into its own method, and call that method from both the HEAD attempt and the GET if we don't have the content length yet. I would also take the knowledge we gain from the failed HEAD and use that to direct to use scratch space instead of a streaming conversion.

This is a self-contained change that doesn't harm any fetches and is a net reduction in the lines of code despite an addition of a test case. I don't think the bar for accepting this change needs to be very high (that is, also fix the entire issue with HEAD rejection. It's only mentioned as a motivating factor).

I'm trying to practice "commit early, commit often", I find that this usually helps working effectively in a team.

awels:
So my problem with this PR is that it removed the HEAD checking code, which is important as I explained above. And by your own admission this PR doesn't actually solve the HEAD problem in the case it is a streaming conversion because qemu-img will try and do a HEAD as well. So I would prefer to keep the HEAD code, handle the error gracefully, and then in addition do what this PR does and check the GET. Then as a follow up we can do the code to not call qemu-img streaming conversion if the HEAD failed originally.

maya-r:
/hold
until there's a full fix for the HEAD issue.

pkg/importer/http-datasource.go
```diff
@@ -248,7 +251,7 @@ func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certD
 
 	total, err := getContentLength(client, ep, accessKey, secKey)
 	if err != nil {
-		return nil, total, err
+		total = 0
```
awels, line 254: if an error happens total is already 0, so no need to explicitly set it here. But we do need to capture the head failed (so we can set brokenForQemu). We also need to communicate that back to the caller of this function, so we need an extra return parameter because below we can potentially find a content length.

maya-r:
Let me know if this is the flow you meant
1. Try HEAD
2. If it fails, mark broken for qemu-img
3. get Content-Length from HEAD
4. If it looks bogus, try Content-Length from GET too

awels:
That flow looks fine to me, I would expect that if the HEAD content length is bogus, the GET one would be bogus as well. But there is no harm in trying.

maya-r:
PR still needs an explicit approval to go through :)
(More than happy to make any changes requested).

pkg/importer/http-datasource.go
```diff
@@ -324,18 +333,25 @@ func getContentLength(client *http.Client, ep *url.URL, accessKey, secKey string
...
+}
+
+func parseHTTPHeader(resp *http.Response) (uint64) {
+	var err error
 	total := uint64(0)
 	if val, ok := resp.Header["Content-Length"]; ok {
 		total, err = strconv.ParseUint(val[0], 10, 64)
 		if err != nil {
-			return uint64(0), errors.Wrap(err, "could not convert content length")
+			klog.Errorf("could not convert content length, got %d", err)
```
awels, line 351: %d is not an err compatible replacement, use %v for complex types. %v prints values in a 'default' format. You can also use %s and do err.Error()

pkg/importer/http-datasource_test.go
```diff
@@ -313,21 +313,43 @@ var _ = Describe("Http reader", func() {
...
+		ep, err := url.Parse(ts.URL)
+		Expect(err).ToNot(HaveOccurred())
+		r, total, _, err := createHTTPReader(context.Background(), ep, "", "", "")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(uint64(0)).To(Equal(total))
+		err = r.Close()
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+
```
awels, line 344: I am missing a test that checks that if the HEAD fails on something that the return value of the boolean return is true.

awels:
Looks like you have a linter failure in travis on some files. make test-lint will let you know which ones.

PR #1129, opened 2020-03-03
Remove error log from ExecWithLimits, let the caller handle this

brybacki:

Removes unnecessary log message. The system.ExecWithLimits did log error message and return error, and even when the caller ignored errors, the message was logged - this was confusing for users. Since it returns err, there is no need to log it.

Example: When resizing an image to available space (util.RetryBackoffSize) - after a failure it retries with new parameters until it succeeds (max 10 times), and it ignores errors until max retries is reached. When it fails on first try and then succeeds it does not return error. With the old behavior an error was logged, a user could see a misleading message like this:
```
I0114 13:38:50.390318 1 data-processor.go:246] Expanding image size to: 1003917312
E0114 13:38:50.405630 1 prlimit.go:164] qemu-img failed output is:
E0114 13:38:50.406281 1 prlimit.go:165] qemu-img: Could not preallocate new data: No
space left on device
```

Fixes #

**Release note**:

```release-note
Removed unnecessary error message when resizing a disk image. 
```

### DIFF
--- a/pkg/system/prlimit.go
+++ b/pkg/system/prlimit.go
@@ -108,6 +108,15 @@ func processScanner(scanner *bufio.Scanner, buf *bytes.Buffer, done chan bool, c
 
 // ExecWithLimits executes a command with process limits
 func ExecWithLimits(limits *ProcessLimitValues, callback func(string), command string, args ...string) ([]byte, error) {
+	return executeWithLimits(limits, callback, true, command, args...)
+}
+
+// ExecWithLimitsSilently executes a command with process limits and do not print output on error
+func ExecWithLimitsSilently(limits *ProcessLimitValues, callback func(string), command string, args ...string) ([]byte, error) {
+	return executeWithLimits(limits, callback, false, command, args...)
+}
+
+func executeWithLimits(limits *ProcessLimitValues, callback func(string), logErr bool, command string, args ...string) ([]byte, error) {
 	// Args can potentially contain sensitive information, make sure NOT to write args to the logs.
 	var buf bytes.Buffer
 	var cmd *exec.Cmd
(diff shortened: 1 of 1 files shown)


mhenriks:
as-is, I think we will regret making this change.  The output of the command could be very helpful in debugging errors.  It is only a problem, er confusing, is in the "if you don't succeed, try, try again" resize case.  I took a quick look and it doesn't appear that any of the callers of this function print the output on error.  So maybe we should change that?  Ideally, the output will only be silenced in cases where errors are okay.

mhenriks:
hold to discuss

awels:
So we should probably identify all the callers of that function and see if they properly print the output on an error, or if that is something we want to do in that case. I believe the context of when/if to surpress the error output should be left up to the caller of the ExecWithLimits function.

brybacki:
I agree, at least the callers should log it. I am on it now.

brybacki:
I have tried to analyze the code, and from what I see none of the direct callers of the system.ExecWithLimits are logging the information. Most of the calls in the CDI originate from processor.ProcessDataWithPause() (data-processor.go) which prints the error with stack trace. 

After removing the error log from ExecWithLimits, there is no information about command output anywhere `prlimit.go:165		klog.Warningf("%s\n", string(output))`, so it would be good to either add the output to error returned, or to log it as a warning or some kind of debug info. 

system.ExecWithLimits is directly used in 7 places in CDI. Now I see three ways to solve the problem:
1. In all 7 cases handle the error and print some details (but then it would be good to have the metioned `command output` in the error or at least logged in ExecWithLimits)
2. just do nothing since all usage in CDI originates from processor.ProcessDataWithPause() and it is logged there (but without command output, so maybe we can improve that by logging as warn, or adding output to error stack trace)
3. restore the error log (do not merge this) - but then the log is misleading, no one expects seeing errors in normal successful execution

brybacki:
I have created additional function ExecWithLimitsSilently instead of addressing all the call sites, so now it is optionally possible to silence the log. 

But on the other hand, if this is now not used anywhere (since the original retry logic was removed in another change) maybe we do not need to do anything there? 

@awels @mhenriks What do you think?

PR #1134, opened 2020-03-06
Runtime lib smartclone controller

awels:
Update the smart clone controller the user the controller runtime library.
Add different storage providers for k8s-1.17.0 provider.
Fix some copy right notices.

Fixes #

**Release note**:

```release-note
NONE
```


cmd/cdi-controller/controller.go
```diff
@@ -135,21 +129,14 @@ func start(cfg *rest.Config, stopCh <-chan struct{}) {
 	}
 
 	cdiInformerFactory := informers.NewSharedInformerFactory(cdiClient, common.DefaultResyncPeriod)
-	csiInformerFactory := csiinformers.NewFilteredSharedInformerFactory(csiClient, common.DefaultResyncPeriod, "", func(options *v1.ListOptions) {
-		options.LabelSelector = common.CDILabelSelector
-	})
+	crdInformerFactory := crdinformers.NewSharedInformerFactory(extClient, common.DefaultResyncPeriod)
 	pvcInformerFactory := k8sinformers.NewSharedInformerFactory(client, common.DefaultResyncPeriod)
```
mhenriks: Can we get rid of unused informer factories?  I think there are a few.

pkg/controller/smart-clone-controller.go
```diff
@@ -1,377 +1,234 @@
...
-	snapshot, err := c.snapshotsLister.VolumeSnapshots(ns).Get(name)
-	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			snapshot = nil
-		} else {
-			// Error getting Snapshot - return
-			return err
-		}
+func shouldReconcilePvc(pvc *corev1.PersistentVolumeClaim) bool {
+	if pvc.Status.Phase != corev1.ClaimBound {
```
mhenriks: PVCs created by snapshots never waitforfirstconsumer?

pkg/controller/smart-clone-controller.go
```diff
@@ -1,377 +1,234 @@
...
+	}
+	return smartCloneController, nil
 }
 
-//ProcessNextItem ...
-func (c *SmartCloneController) ProcessNextItem() bool {
-	key, shutdown := c.queue.Get()
-	if shutdown {
-		return false
+func addSmartCloneControllerWatches(mgr manager.Manager, importController controller.Controller) error {
```
mhenriks: `importController`

pkg/controller/smart-clone-controller.go
```diff
@@ -1,377 +1,234 @@
...
+		return reconcile.Result{}, err
 	}
 
-	klog.V(3).Infof("Creating PVC \"%s/%s\" from snapshot", newPvc.Namespace, newPvc.Name)
-	_, err = c.clientset.CoreV1().PersistentVolumeClaims(snapshot.Namespace).Create(newPvc)
-	if err != nil {
-		klog.Errorf("error creating pvc from snapshot: %v", err)
-		return err
+	if err := r.Client.Delete(context.TODO(), snapshotToDelete); err != nil {
+		log.Error(err, "error deleting snapshot for smart-clone")
```
mhenriks: why not handle notfound here as well

cmd/cdi-controller/controller.go
awels: Yes planning on doing that when I get rid of most of the client-go clients as well. In a follow up PR.

pkg/controller/smart-clone-controller.go
awels: They can't be, we have written data to it, so if they were WaitForFirstConsumer because they are local storage, then the node has been picked. If it is a topology zone, we have at a minimum written to the storage in that zone, and thus still can't have WaitForFirstConsumer.

pkg/controller/smart-clone-controller.go
awels: Whoops, I thought I had gotten all of those. Fixed.

pkg/controller/smart-clone-controller.go
awels: I check for not found right above, but you are right it is possible that in between those two statements the snapshot got deleted, so I should not error if the delete fails on a not found.

pkg/controller/smart-clone-controller.go
mhenriks: I was able to simulate this easily in in my local rook ceph configuration by making the storageclass waitforfirstconsumer.

```
➜  kubevirt git:(vmclone) ✗ k get volumesnapshot
NAME            READYTOUSE   SOURCEPVC   SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS             SNAPSHOTCONTENT                                    CREATIONTIME   AGE
test-snapshot   true         my-pvc                              500Mi         csi-rbdplugin-snapclass   snapcontent-36a79976-79f3-4c5c-825b-90c9b38d5776   7s             9s
```

```
➜  kubevirt git:(vmclone) ✗ k get pvc my-pvc-from-snap -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  creationTimestamp: "2020-03-16T14:03:29Z"
  finalizers:
  name: my-pvc-from-snap
  namespace: default
  resourceVersion: "2406"
  selfLink: /api/v1/namespaces/default/persistentvolumeclaims/my-pvc-from-snap
  uid: 7f80ccab-1068-4f67-b715-673f3150f2fd
spec:
  accessModes:
  - ReadWriteOnce
  dataSource:
    apiGroup: snapshot.storage.k8s.io
    kind: VolumeSnapshot
    name: test-snapshot
  resources:
    requests:
      storage: 500Mi
  volumeMode: Filesystem
status:
  phase: Pending
```

awels: Interesting, I didn't think it would be possible. I guess I can add pending to the status that we can reconcile.

pkg/controller/smart-clone-controller.go
```diff
@@ -132,7 +132,7 @@ func shouldReconcileSnapshot(snapshot *csiv1.VolumeSnapshot) bool {
 }
 
 func shouldReconcilePvc(pvc *corev1.PersistentVolumeClaim) bool {
-	if pvc.Status.Phase != corev1.ClaimBound {
+	if pvc.Status.Phase == corev1.ClaimLost {
```
mhenriks: Correct me if I'm wrong, but the code will now delete the `VolumeSnapshot` when the PVC is pending?  I think that is technically okay because it has a finalizer.  But I'm not sure.  May not want to depend on that.  Maybe it would be better to keep it around until the PVC is bound?

awels: So we have 3 possible values for the PVC state, Pending, Bound, and Lost. If due to WaitForFirstConsumer we have to handle pending, and we handle bound, then what are our options?

PR #1136, opened 2020-03-09
Looks like gradle is no longer available in fedora 31, and the build …

awels:
…image build is failing due to it.

Builder is not completing and releases are failing due to it.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -15,7 +15,6 @@ RUN sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
 	python3-pip \
 	python3-devel \
 	mercurial \
-	gradle \
 	gcc \
     gcc-c++ \
     glibc-devel \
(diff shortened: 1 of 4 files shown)


mhenriks:
looks like we need gradle here:  https://github.com/kubevirt/containerized-data-importer/blob/7db4c01fccc0afa7df910036c34f3d1b874bf9c4/hack/gen-swagger-doc/gen-swagger-docs.sh#L41

maya-r:
would this do the trick
```
@@ -32,6 +31,9 @@ RUN sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
        rmdir /etc/yum.repos.d/old
 
 RUN pip3 install j2cli && pip3 install operator-courier
+RUN curl -sL https://services.gradle.org/distributions/gradle-6.2.2-bin.zip && \
+    unzip gradle-6.2.2-bin.zip && \
+    ln -s ./gradle-6.2.2/bin/gradle /usr/local/bin/gradle
 
 ENV GIMME_GO_VERSION=1.12.14 GOPATH="/go" KUBEBUILDER_VERSION="2.2.0" ARCH="amd64"
 ENV BAZEL_PYTHON=/usr/bin/python3
````

**Comment by awels** [comment]
I wonder why gradle is no longer available in fedora, but we can manually grab it. I will update the PR.

**Comment by mhenriks** [comment]
I pulled this branch, built the image, updated BUILDER_IMAGE, and ran `make apidocs`.  Got the following error:

```
ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH.

Please set the JAVA_HOME variable in your environment to match the
location of your Java installation.

make: *** [Makefile:53: apidocs] Error 1
```

Java does appear to be installed, though

awels:
So it appears with gradle 6.2.2 the build will fail, so I have to use 4.3.1 like before which requires java 8 instead of 11.

PR #1143, opened 2020-03-12
Add ember csi lane

awels:

Add ember csi as a storage provider for make cluster-sync, this involves:
- Adding a loopback device container that creates a new loopback device. The path and size are parameters that can be changed.
- Adding a new item to the list of storage providers, ember_lvm. It can be enabled by setting the KUBEVIRT_STORAGE environment variable to ember_lvm
- A script to identify the host the loopback container is running on and running some pvcreate and vgcreate commands to create the needed lvm objects.
- Yaml to create a new ember-csi-lvm namespace and creating the ember csi objects in that namespace.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/ember/ember-csi-lvm.yaml
+++ b/cluster-sync/ember/ember-csi-lvm.yaml
@@ -67,8 +67,6 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: false
-  # volumeLifecycleModes: # added in Kubernetes 1.16, we use default
-  # - Persistent
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
@@ -162,8 +160,7 @@ spec:
               mountPath: /registration
         - name: csi-driver
           image: "embercsi/ember-csi:master"
-          #command: ["tail"]
-          #args: ["-f", "/dev/null"]
+          # Priviledged needed for access to lvm backend
           securityContext:
             privileged: true
             allowPrivilegeEscalation: true
--- a/cluster-sync/ember/ember-csi-lvm.yaml
+++ b/cluster-sync/ember/ember-csi-lvm.yaml
@@ -1,0 +1,299 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: csi-controller-sa
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: csi-controller-cr
+rules:
+  # Allow managing ember resources
+  - apiGroups: ['ember-csi.io']
+    resources: ['*']
+    verbs: ['*']
+  # Allow listing and creating CRDs
+  - apiGroups: ['apiextensions.k8s.io']
+    resources: ['customresourcedefinitions']
+    verbs: ['list', 'create']
+  - apiGroups: ['']
+    resources: ['persistentvolumes']
+    verbs: ['create', 'delete', 'get', 'list', 'watch', 'update', 'patch']
+  - apiGroups: ['']
+    resources: ['secrets']
+    verbs: ['get', 'list']
+  - apiGroups: ['']
+    resources: ['persistentvolumeclaims']
+    verbs: ['get', 'list', 'watch', 'update']
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims/status"]
+    verbs: ["update", "patch"]
+  - apiGroups: ['']
+    resources: ['nodes']
+    verbs: ['get', 'list', 'watch']
+  - apiGroups: ['storage.k8s.io']
+    resources: ['volumeattachments']
+    verbs: ['get', 'list', 'watch', 'update', 'patch']
+  - apiGroups: ['storage.k8s.io']
+    resources: ['storageclasses']
+    verbs: ['get', 'list', 'watch']
+  - apiGroups: ['csi.storage.k8s.io']
+    resources: ['csidrivers']
+    verbs: ['get', 'list', 'watch', 'update', 'create']
+  - apiGroups: ['']
+    resources: ['events']
+    verbs: ['list', 'watch', 'create', 'update', 'patch']
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotcontents"]
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots/status"]
+    verbs: ["update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["csinodes"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: storage.k8s.io/v1beta1
+kind: CSIDriver
+metadata:
+  name: ember-csi.io
+spec:
+  attachRequired: true
+  podInfoOnMount: false
+  # volumeLifecycleModes: # added in Kubernetes 1.16, we use default
+  # - Persistent
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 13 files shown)


maya-r:
Would be nice if the commit message said,
- This adds ember CSI support
- Why we had to bump sizes

If the reason for changing the smartclone tests is that we can't do it yet with ember CSI, maybe we can comment it out and add a comment like
// TODO: smartclone using ember CSI

awels:
thats why I put the hold on here. We don't want to bump the sizes for instance because that is a bug in ember. I just put it up here so other people could review. I am aware the PR message is lacking and is missing a lot of detail.

cluster-sync/ember/ember-csi-lvm.yaml
```diff
@@ -0,0 +1,299 @@
...
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - mountPath: /var/lib/ember-csi
+              name: ember-csi-data
+            - name: registration-dir
+              mountPath: /registration
+        - name: csi-driver
+          image: "embercsi/ember-csi:master"
+          #command: ["tail"]
+          #args: ["-f", "/dev/null"]
```
aglitke: Remove the commented lines.

cluster-sync/ember/ember-csi-lvm.yaml
```diff
@@ -0,0 +1,299 @@
...
+          securityContext:
+            privileged: true
+            allowPrivilegeEscalation: true
```
aglitke: Perhaps it's worth noting that this is here for the lvm backend setup.

hack/build/run-tests.sh
```diff
@@ -23,11 +23,11 @@ source hack/build/common.sh
 parseTestOpts "${@}"
 
 if [ -f "${TESTS_OUT_DIR}/tests.test" ]; then
-    test_command="${TESTS_OUT_DIR}/tests.test -test.timeout 180m ${test_args}"
+    test_command="${TESTS_OUT_DIR}/tests.test -test.timeout 270m ${test_args}"
```
aglitke: Hmm, ember is that much slower?

cluster-sync/ember/ember-csi-lvm.yaml
```diff
@@ -0,0 +1,299 @@
...
+  - apiGroups: ['storage.k8s.io']
+    resources: ['storageclasses']
+    verbs: ['get', 'list', 'watch']
+  - apiGroups: ['csi.storage.k8s.io']
+    resources: ['csidrivers']
+    verbs: ['get', 'list', 'watch', 'update', 'create']
+  - apiGroups: ['']
+    resources: ['events']
+    verbs: ['list', 'watch', 'create', 'update', 'patch']
+  - apiGroups: ["snapshot.storage.k8s.io"]
```
aglitke: Any concerns with the versions of snapshot apis here?  I assume no but just checking.

hack/build/run-tests.sh
awels: Yes, it appears that mounting the volume in the pod is fairly slow for some reason. And of course we have a block device backed by file on another file system. But a successful run takes almost 3 hours, any failures will put it over the top, so the increase is just so we get some output on failures.

cluster-sync/ember/ember-csi-lvm.yaml
awels: No, but I expect that to change as the new version gets adopted, but for now this is sufficient.

cluster-sync/ember/ember-csi-lvm.yaml
awels: Done

cluster-sync/ember/ember-csi-lvm.yaml
awels: Done

cluster-sync/ember/ember-csi-lvm.yaml
mhenriks: beta snapshots have the same resource names so I don't expect anything to have to change here.

cluster-sync/ember/ember-csi-lvm.yaml
```diff
@@ -0,0 +1,296 @@
...
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - mountPath: /var/lib/ember-csi
+              name: ember-csi-data
+            - name: registration-dir
+              mountPath: /registration
+        - name: csi-driver
+          image: "embercsi/ember-csi:master"
```
mhenriks: should we be using the `master` tag?

tools/loop-back-lvm/BUILD.bazel
```diff
@@ -0,0 +1,63 @@
+load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
```
mhenriks: is `tools` the right place for this?  I think of `tools` as a place for utility programs (executables).  Would it be better to have this in a directory alongside the builder docker image?

cluster-sync/ember/ember-csi-lvm.yaml
awels: The last release is from Jun 2019, and it is missing a bunch of stuff we need.

tools/loop-back-lvm/BUILD.bazel
awels: I looked at this as a utility container to enable ember to work. But I have no strong feelings on where to put it.

cluster-sync/ember/loop_back.yaml
```diff
@@ -0,0 +1,73 @@
...
+spec:
+  storageClassName: local
+  accessModes:
+     - ReadWriteOnce
+  resources:
+    requests:
+      storage: 25Gi
+---
+apiVersion: apps/v1
+kind: Deployment
```
mhenriks: Would a Job/Pod be better?  As-is, deleting the deployment pod puts it in permanent Error/CrashloopBackoff state.

tools/loop-back-lvm/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
...
+
+cp ./create_lvm.sh /host/create_lvm.sh
+chroot /host ./create_lvm.sh
+rm /host/create_lvm.sh
+
+# let the monitoring script know we're done
+echo "done" >/ready
+echo "ready"
+while true; do
+  sleep 60
```
mhenriks: FYI:  the script cannot receive signals when `sleep` is executing.  So, when docker/kubernetes tries to stop this container, it may hang for awhile and eventually get force killed

mhenriks: So `sleep 1` is better IMO

cluster-sync/ember/ember-csi-lvm.yaml
mhenriks: We want to have a consistent image.  So why not grab one that works and push it to your personal/kubevirt dockerhub?

cluster-sync/ember/loop_back.yaml
awels: Yeah deleting the pod will mess with all kinds of things on the host. So it doesn't really make sense to do a deployment, might as well be a pod.

tools/loop-back-lvm/entrypoint.sh
awels: Sure.

cluster-sync/ember/ember-csi-lvm.yaml
awels: Okay, I can do that.

tools/loop-back-lvm/entrypoint.sh
```diff
@@ -10,5 +10,5 @@ rm /host/create_lvm.sh
 echo "done" >/ready
 echo "ready"
 while true; do
-  sleep 60
+  sleep 1
```
mhenriks: Any reason to keep this running not that it's a pod?  I see no issue with keeping a completed pod around

awels: Not quite sure why, but if I do an exit 0 to make a clean exit, it ends up in a crash loop...

mhenriks: Maybe try removing the readiness probe?

mhenriks: And maybe restart policy of Never?

awels: Yeah that appears to work fine. Pushed an update.

PR #1145, opened 2020-03-12
Enable IPv6 on cdi http file host.

awels:

Enable ipv6 for the cdi file hosts in case of an ipv6 only cluster for testing. The registry file host is already ipv6 enabled from what I could tell.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/cdi-func-test-file-host-init/nginx.conf
+++ b/tools/cdi-func-test-file-host-init/nginx.conf
@@ -20,6 +20,7 @@ http {
         server_name localhost;
 
         listen 80;
+        listen [::]:80;
 
         root /tmp/shared/images;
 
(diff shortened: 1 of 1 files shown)


tools/cdi-func-test-file-host-init/nginx.conf
```diff
@@ -34,6 +35,7 @@ http {
         server_name localhost;
 
         listen 81;
+        listen [::]:82;
```
mhenriks: should this be `[::]:82`?

awels: Yes... D'oh!

awels:
Not sure why https://github.com/kubevirt/containerized-data-importer/pull/1130 was closed, but this PR is the same.

PR #1148, opened 2020-03-13
Set secondary GID of user on import and upload pods to be 107 (qemu) …

awels:
…for kubevirt content type.

Attempt to set the group of the disk.img to be qemu (107) so that pods using the PVCs can run as qemu and still read the disk image. If a pod wants to override the group id of the file they can change the fsGroup of the POD consuming the PVC.

Fixes #

**Release note**:

```release-note
Enhancement: Kubevirt content type disk.img files now have a group id of qemu (107) when imported/cloned/uploaded.
```

### DIFF
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -503,6 +503,7 @@ func GetUploadServerURL(namespace, pvc, uploadPath string) string {
 
 func (r *UploadReconciler) makeUploadPodSpec(args UploadPodArgs, resourceRequirements *v1.ResourceRequirements) *v1.Pod {
 	requestImageSize, _ := getRequestedImageSize(args.PVC)
+	fsGroup := common.QemuSubGid
 	pod := &v1.Pod{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Pod",
@@ -583,6 +584,10 @@ func (r *UploadReconciler) makeUploadPodSpec(args UploadPodArgs, resourceRequire
 				},
 			},
 		},
+	}
+
+	if !checkPVC(args.PVC, AnnCloneRequest) {
+		pod.Spec.SecurityContext.FSGroup = &fsGroup
 	}
 
 	if resourceRequirements != nil {
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -67,6 +67,45 @@ var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		doFileBasedCloneTest(f, pvcDef, f.Namespace)
 	})
 
+	It("Should clone imported data within same namespace and preserve fsGroup", func() {
+		diskImagePath := filepath.Join(testBaseDir, testFile)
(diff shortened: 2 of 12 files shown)


maya-r:
Sounds like our NFS issue might be https://github.com/kubernetes/examples/issues/260

awels:
Hmm all three failures are the same test failing...

pkg/controller/upload-controller.go
```diff
@@ -526,6 +527,7 @@ func (r *UploadReconciler) makeUploadPodSpec(args UploadPodArgs, resourceRequire
 		Spec: v1.PodSpec{
 			SecurityContext: &v1.PodSecurityContext{
 				RunAsUser: &[]int64{0}[0],
+				FSGroup:   &fsGroup,
```
mhenriks: Wondering if we should not set this for clone targets?  Assuming that cloning preserves ownership.  If not, it probably should, right?

And we should probably only be doing this for "kubevirt" `DataVolumeContentType`, right?

awels: You make very good points. Let me fix that.

tests/cloner_test.go
```diff
@@ -716,10 +721,14 @@ func doFileBasedCloneTest(f *framework.Framework, srcPVCDef *v1.PersistentVolume
...
-	completeClone(f, targetNs, targetPvc, filepath.Join(testBaseDir, testFile), fillDataFSMD5sum)
+	sourcePvcDiskGroup, err := f.GetDiskGroup(f.Namespace, srcPVCDef)
+	fmt.Fprintf(GinkgoWriter, "INFO: %s\n", sourcePvcDiskGroup)
+	Expect(err).ToNot(HaveOccurred())
+
+	completeClone(f, targetNs, targetPvc, filepath.Join(testBaseDir, testFile), fillDataFSMD5sum, sourcePvcDiskGroup)
 }
 
-func completeClone(f *framework.Framework, targetNs *v1.Namespace, targetPvc *v1.PersistentVolumeClaim, filePath, expectedMD5 string) {
+func completeClone(f *framework.Framework, targetNs *v1.Namespace, targetPvc *v1.PersistentVolumeClaim, filePath, expectedMD5, sourcePvcDiskGroup string) {
```
mhenriks: I might be missing it, but is there a test that imports a disk image, then clones it (including the fsgroup)?

awels: No but we haven't modified the clone behaviour in this PR.

mhenriks: I understand, but to me, the goal of this PR is to ensure that all disk images have the appropriate fs group.  Cloning from an import seems like a pretty common use case.  Should we not confirm that it works correctly?

awels: There is no specific test, but at the end of completeClone, I compare the fsGroup of the clone against the fsGroup of the source. And that is called in a bunch of the clone tests.

awels:
> Is this going across namespace or not? If same namespace, will it do a smart or dumb clone?

Same namespace, and depending on the storage class it will do either one. That is why I changed the 'success' check to look at the data volume instead of the PVC, so a smart clone will also work.

PR #1152, opened 2020-03-23
Add critical addons toleration to cdi deployment

danielBelenky:
By toleratin the CriticalAddonsOnly taint, we ensure that our components
will get a reserved spot for critical addons in case they are being
evicted. This will reduce of chance of CDI component/s starving for a
place on a node and permanently unavailable.

Also added a functional test to verify this behaviour.

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>

**Release note**:

```release-note
CDI components now tolerate the CriticalAddonsOnly taint
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -2,6 +2,7 @@ package tests_test
 
 import (
 	"fmt"
+	"reflect"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -14,6 +15,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 
 	cdiv1alpha1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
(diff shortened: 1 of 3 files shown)


danielBelenky:
> What about the API server and the upload proxy, should those have the tolerations as well?

All deployments are being created through the `CreateDeployment` function... so they all get this toleration.
https://github.com/kubevirt/containerized-data-importer/pull/1152/files#diff-2a23fc371bce9ff719c7cca34e83a89aR222

This is also being verified in the the functional test... we only exclude test components so in case one of the components will not have this toleration, it will be evicted and the test will fail.

tests/operator_test.go
```diff
@@ -206,4 +209,91 @@ var _ = Describe("Operator delete CDI tests", func() {
...
+				_, err = f.K8sClient.CoreV1().Nodes().Update(cdiNodeCopy)
+				Expect(err).ShouldNot(HaveOccurred(), "failed setting taint on node")
+			}
+		})
+
+		defer func() {
+			var errors []error
+			By("restoring nodes")
+			for _, nodeSpec := range cdiNodes {
+				_, err := f.K8sClient.CoreV1().Nodes().Update(nodeSpec)
```
aglitke: Wouldn't it be better to just remove the taint instead of overwriting the whole spec?  Could there be unwanted/unexpected changes as a result of this approach?

danielBelenky: Theoretically, yes, in a multi-user/multi-threaded environment, we could expose ourselves to a race. But in this case, it's a functional test that can't run in a multi-threaded environment anyway so I don't see a risk here.

If you think that patching is better here, I don't mind to use a patch instead of an update.

tests/operator_test.go
```diff
@@ -206,4 +209,91 @@ var _ = Describe("Operator delete CDI tests", func() {
...
+				if err != nil {
+					errors = append(errors, err)
+				}
+			}
+			Expect(errors).Should(BeEmpty(), "failed restoring one or more nodes")
+		}()
+
+		By("watching for terminated cdi pods", func() {
```
mhenriks: Shouldn't we technically be watching before the taint is added?  Is there not a possibility of missing an event otherwise?

tests/operator_test.go
```diff
@@ -206,4 +209,91 @@ var _ = Describe("Operator delete CDI tests", func() {
...
+							terminatedPodsCn <- true
+							return
+						}
+					}
+				}
+			}
+			stopCn := make(chan bool, 1)
+			terminatedPodsCn := make(chan bool, 1)
+			go signalTerminatedPods(stopCn, lw.ResultChan(), terminatedPodsCn)
+			Consistently(terminatedPodsCn, 5*time.Second).
```
mhenriks: It sucks increasing timeouts when waiting for things to NOT happen but may want to consider here given overloaded CI environment.

tests/operator_test.go
mhenriks: I think update is fine.

tests/operator_test.go
danielBelenky: not sure I understand what you mean... do you want me to increase the timeout?

tests/operator_test.go
danielBelenky: Technically, it is possible... since it takes the pod controller some time (seconds) to actually delete pods after they were marked with the deletion timestamp, we catch them.

But I agree with you, it's better to start the watching routine before we taint the nodes... I've updated the pr. Thanks :smile:

awels:
Note #1154 should fix some of the NFS failures we are seeing.

awels:
Hmm could it be that the NFS server pod is being evicted and that is causing NFS failures?

danielBelenky:
@awels I've rebased the PR on top of your fix.

maya-r:
I am assuming github doesn't notify for it: the PR has conflicts currently

maya-r:
I wonder if this pull request can be served by #1346 which allows specifying tolerations for all things created by CDI

awels:
obsoleted by #1361

PR #1155, opened 2020-03-24
Add a field to DataVolume to track the number of retries/pod restarts

brybacki:

Add a field to DataVolume to track the number of retries. 

Some background: A DataVolume will create a PVC with a specific annotation, that will trigger a CDI controller to create a pod (import/clone/upload) that will populate the PVC with a disk image. Things can potentially go wrong during the import (or upload/clone, I will talk only import from now on) causing the pod to fail and restart. Kubernetes when you get a pod will list the number or restarts:
```
$ k get pods
```
```
NAME                             READY   STATUS   RESTARTS   AGE
local-volume-provisioner-tgktc   1/1     Running  0          116m
local-volume-provisioner-w6dkt   1/1     Running  0          115m
```
This change adds RESTARTS column to DataVolume: 
```
k get dv
```
```
NAME                  PHASE             PROGRESS   RESTARTS   AGE
dv-test-1g-cloned     CloneInProgress   N/A        3          76s
dv-test-1g-qcow2-xz   Succeeded         100.0%                117s
```
DataVolumes abstract this whole process and unless you know that you need to look at the pods you can't tell if your data volume is progressing or not. 

Fixes #

Work In Progress, needs tests.

**Release note**:

```release-note
Add a RESTARTS field to DataVolume that tracks a pod(import/clone/upload) retries. 
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -217,6 +217,16 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 		pvc.Annotations[AnnCloneOf] = "true"
 		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, "Clone Successful")
 	}
+	if sourcePod.Status.ContainerStatuses != nil {
+		// update pvc annotation tracking pod restarts only if the source pod restarts are greater or equal
+		// see the same in upload-controller
+		annPodRestarts, _ := strconv.Atoi(pvc.Annotations[AnnPodRestarts])
+		podRestarts := int(sourcePod.Status.ContainerStatuses[0].RestartCount)
+		if podRestarts > annPodRestarts {
+			pvc.Annotations[AnnPodRestarts] = strconv.Itoa(podRestarts)
+		}
+	}
+
 	if !reflect.DeepEqual(currentPvcCopy, pvc) {
 		return r.updatePVC(pvc)
 	}
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -218,7 +218,7 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, "Clone Successful")
 	}
 	if sourcePod != nil && sourcePod.Status.ContainerStatuses != nil {
-		// update pvc annotation tracking pod restarts only if the source pod restarts are greater or equal
+		// update pvc annotation tracking pod restarts only if the source pod restart count is greater
 		// see the same in upload-controller
 		annPodRestarts, _ := strconv.Atoi(pvc.Annotations[AnnPodRestarts])
 		podRestarts := int(sourcePod.Status.ContainerStatuses[0].RestartCount)
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -217,7 +217,7 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 		pvc.Annotations[AnnCloneOf] = "true"
 		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, "Clone Successful")
 	}
-	if sourcePod.Status.ContainerStatuses != nil {
+	if sourcePod != nil && sourcePod.Status.ContainerStatuses != nil {
 		// update pvc annotation tracking pod restarts only if the source pod restarts are greater or equal
 		// see the same in upload-controller
 		annPodRestarts, _ := strconv.Atoi(pvc.Annotations[AnnPodRestarts])
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -605,7 +605,7 @@ func (r *DatavolumeReconciler) reconcileDataVolumeStatus(dataVolume *cdiv1.DataV
 	}
 
 	if pvc != nil {
-		if i, err := strconv.Atoi(pvc.Annotations[AnnPodRestarts]); err == nil && i > 1 {
+		if i, err := strconv.Atoi(pvc.Annotations[AnnPodRestarts]); err == nil && i >= 0 {
 			dataVolumeCopy.Status.RestartCount = int32(i)
 		}
 	}
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -277,6 +277,7 @@ func (r *DatavolumeReconciler) reconcileProgressUpdate(datavolume *cdiv1.DataVol
 	if datavolume.Status.Progress == "" {
 		datavolume.Status.Progress = "N/A"
 	}
+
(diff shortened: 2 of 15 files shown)


pkg/controller/datavolume-controller_test.go
```diff
@@ -133,6 +134,33 @@ var _ = Describe("Datavolume controller reconcile loop", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		pvc := &corev1.PersistentVolumeClaim{}
+		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, pvc)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(pvc.Name).To(Equal("test-dv"))
+
+		dv := &cdiv1.DataVolume{}
+		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, dv)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(dv.Status.RestartCount).To(BeEquivalentTo(0))
```
awels: BeEquivalenTo is used when comparing two types that are derived from each other but are not the same. You will that often with string and another type. If you are comparing the same types, use Equals.

pkg/controller/datavolume-controller_test.go
```diff
@@ -107,6 +107,7 @@ var _ = Describe("Datavolume controller reconcile loop", func() {
 	})
 
 	It("Should follow the phase of the created PVC", func() {
+		//todo
```
awels: This still need to be here?, you have tests below.

pkg/controller/datavolume-controller_test.go
brybacki: right, thanks

pkg/controller/datavolume-controller_test.go
brybacki: removed

tests/utils/datavolume.go
```diff
@@ -40,6 +40,8 @@ const (
 	CirrosURL = "http://cdi-file-host.%s/cirros-qcow2.img"
 	// ImageioURL provides URL of oVirt engine hosting imageio
 	ImageioURL = "https://imageio.%s:12346/ovirt-engine/api"
+	// NoImage404URL provides a fake broken url - a 404
+	NoImage404URL = "http://cdi-file-host:82/nonexistent.qcow2"
```
awels: We already have a constant with an invalid url: InvalidQcowImagesURL, please use that instead of making a new one.

pkg/controller/clone-controller.go
```diff
@@ -217,6 +217,16 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 		pvc.Annotations[AnnCloneOf] = "true"
 		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, "Clone Successful")
 	}
+	if sourcePod != nil && sourcePod.Status.ContainerStatuses != nil {
+		// update pvc annotation tracking pod restarts only if the source pod restarts are greater or equal
```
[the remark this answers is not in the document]
brybacki: ok, updated here and in the upload-controller.go

pkg/controller/datavolume-controller.go
```diff
@@ -603,6 +604,11 @@ func (r *DatavolumeReconciler) reconcileDataVolumeStatus(dataVolume *cdiv1.DataV
 		}
 	}
 
+	if pvc != nil {
+		if i, err := strconv.Atoi(pvc.Annotations[AnnPodRestarts]); err == nil && i > 1 {
```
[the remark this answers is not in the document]
brybacki: yes

pkg/controller/upload-controller.go
```diff
@@ -164,6 +164,16 @@ func (r *UploadReconciler) reconcilePVC(log logr.Logger, pvc *corev1.PersistentV
 	pvcCopy.Annotations[AnnPodPhase] = string(podPhase)
 	pvcCopy.Annotations[AnnPodReady] = strconv.FormatBool(isPodReady(pod))
 
+	if pod.Status.ContainerStatuses != nil {
+		// update pvc annotation tracking pod restarts only if the upload pod restarts are greater or equal
```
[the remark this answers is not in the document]
brybacki: done

pkg/operator/resources/cluster/datavolume.go
```diff
@@ -165,6 +165,12 @@ func createDataVolumeCRD() *extv1beta1.CustomResourceDefinition {
 					Description: "Transfer progress in percentage if known, N/A otherwise",
 					JSONPath:    ".status.progress",
 				},
+				{
+					Name:        "Restarts",
+					Type:        "integer",
+					Description: "The number of times the containers in this pod have been restarted.",
```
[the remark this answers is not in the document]
brybacki: makes sens

tests/datavolume_test.go
```diff
@@ -361,4 +361,37 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			if err != nil {
+				PrintControllerLog(f)
+				dv, dverr := f.CdiClient.CdiV1alpha1().DataVolumes(f.Namespace.Name).Get(dataVolume.Name, metav1.GetOptions{})
+				if dverr != nil {
+					Fail(fmt.Sprintf("datavolume %s phase %s", dv.Name, dv.Status.Phase))
+				}
+			}
+
+			By("Verify dv")
+			println("aaaaaaaaaaaa")
```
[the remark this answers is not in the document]
brybacki: thanks for info, done

tests/utils/datavolume.go
brybacki: good catch, I have missed that,

tests/datavolume_test.go
```diff
@@ -361,4 +361,35 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
 	})
+
+	Describe("Restarts reporting on import datavolume", func() {
+		It("Should report restarts on failed import", func() {
+			dataVolume := utils.NewDataVolumeWithHTTPImport(dataVolumeName, "1Gi", utils.InvalidQcowImagesURL)
+			By(fmt.Sprintf("creating new datavolume %s", dataVolume.Name))
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
+			Expect(err).ToNot(HaveOccurred())
+
+			//Due to the rate limit, this will take a while, so we can expect the phase to be in progress.
```
awels: I think you can remove the lines from 372 to 382. I am assuming you got most of this code from other tests, so here is some information:
We have 3 ports on the file host nginx container:
80 http, no rate limit
81 https, with self signed cert, you can get the certificate from the config map in the cdi namespace, there are existing functions that copy it into the testing namespace to be usable.
82 http, rate limited port. Same as port 80, but with a rate limit. This if for tests where we need it to take while for some resource in the import pod to become available (like the progress reporting). If the pod completes too fast we can't see if it is working.

In this case we are using port 80, so mention of the rate limit port in the comments confused me at first. But then I realized for this particular test (making sure the counter increases, we aren't actually interested in the phase of the data volume, so there is no need to make sure it is a certain value).

Long story short, you can likely remove the lines I mention above.

PR #1157, opened 2020-03-25
Don't annotate pvc/dvs with provisionOnNode as it causes a race with …

awels:
…the kubernetes scheduler.

The hostpath provisioner didn't support WaitForFirstConsumer before, so we managed that by always annotating DV/PVCs with provisionOnNode. However now that the hostpath provisioner supports WaitForFirstConsumer there is a potential race between the provisioner and the kubernetes scheduler.

In the time it takes the provisioner to create a PV and bind to the PVC, it is possible the scheduler schedules the pod that uses it on another node, leaving a pod that cannot be scheduled. Since we no longer need the annotation in our tests, this PR removes the annotation causing the race.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/datavolume.go
+++ b/tests/utils/datavolume.go
@@ -44,12 +44,6 @@ const (
 
 // CreateDataVolumeFromDefinition is used by tests to create a testable Data Volume
 func CreateDataVolumeFromDefinition(clientSet *cdiclientset.Clientset, namespace string, def *cdiv1.DataVolume) (*cdiv1.DataVolume, error) {
-	if IsHostpathProvisioner() {
-		if def.ObjectMeta.Annotations == nil {
-			def.ObjectMeta.Annotations = map[string]string{}
-		}
-		AddProvisionOnNodeToAnn(def.ObjectMeta.Annotations)
-	}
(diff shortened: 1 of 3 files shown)


tests/utils/datavolume.go
```diff
@@ -44,12 +44,6 @@ const (
 
 // CreateDataVolumeFromDefinition is used by tests to create a testable Data Volume
 func CreateDataVolumeFromDefinition(clientSet *cdiclientset.Clientset, namespace string, def *cdiv1.DataVolume) (*cdiv1.DataVolume, error) {
-	if IsHostpathProvisioner() {
-		if def.ObjectMeta.Annotations == nil {
-			def.ObjectMeta.Annotations = map[string]string{}
-		}
-		AddProvisionOnNodeToAnn(def.ObjectMeta.Annotations)
```
aglitke: It looks like you don't need this helper function anymore.

awels: Good point, don't need the DefaultNodeName either.

awels:
@maya-r had to rebase to fix conflict.

PR #1160, opened 2020-03-26
Read timeout from environment variable instead of hard coding it.

awels:

The timeout to wait for cdi to be available is hard coded, the os-3.11-crio lane seems to take a lot longer in certain cases. This PR allows us to read the timeout from an environment variable which we can then set in the lane configuration.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/sync.sh
+++ b/cluster-sync/sync.sh
@@ -14,6 +14,7 @@ source ./cluster-sync/${KUBEVIRT_PROVIDER}/provider.sh
 CDI_INSTALL="install-operator"
 CDI_NAMESPACE=${CDI_NAMESPACE:-cdi}
 CDI_INSTALL_TIMEOUT=${CDI_INSTALL_TIMEOUT:-120}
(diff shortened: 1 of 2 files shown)


cluster-sync/sync.sh
```diff
@@ -14,6 +14,7 @@ source ./cluster-sync/${KUBEVIRT_PROVIDER}/provider.sh
 CDI_INSTALL="install-operator"
 CDI_NAMESPACE=${CDI_NAMESPACE:-cdi}
 CDI_INSTALL_TIMEOUT=${CDI_INSTALL_TIMEOUT:-120}
+CDI_AVAILABLE_TIMEOUT=${CDI_AVAILABLE_TIMEOUT:-480}
```
mhenriks: how can we make this longer for the crio lane?

awels: See https://github.com/kubevirt/project-infra/pull/401 we increase it just for the CRIO lane.

PR #1163, opened 2020-03-31
Clean up client-go left overs after converting to runtime library.

awels:
Clean unused functions from controller utils.go
Turn util_test.go into gingko test.
Moved functions from util to proper controller if only used by that controller.

Clean up most remnants of the client-go objects in the CDI controller code.
Cleanup pkg/controller/util.go and remove unused functions and structs.
Switch util_test.go to use gingko framework.
Explicitly did not change leader election mechanism because of known bug in the version of the library we are using.
Fixes #

**Release note**:

```release-note
NONE
```


pkg/controller/config-controller.go
```diff
@@ -63,6 +61,7 @@ func (r *CDIConfigReconciler) Reconcile(req reconcile.Request) (reconcile.Result
 		log.Error(err, "Unable to create CDIConfig")
 		return reconcile.Result{}, err
 	}
+	log.Info("CDIConfig", "config", config)
```
maya-r, line 64: Leftover debug output?

pkg/controller/import-controller_test.go
```diff
@@ -409,11 +409,152 @@ var _ = Describe("Import test env", func() {
...
+		table.Entry("return true on endpoint with no port, and cdi configmap exists, and host defined", endpointNoPort, true, host, true),
+		table.Entry("return true on endpoint with port, and cdi configmap exists, and host with port", endpointWithPort, true, hostWithPort, true),
+		table.Entry("return false on endpoint with no port, and cdi configmap exists, and host with port", endpointNoPort, true, hostWithPort, false),
+		table.Entry("return false on endpoint with port, and cdi configmap exists, and host defined", endpointWithPort, true, host, false),
+		table.Entry("return false on endpoint with no port, and no cdi configmap exists, and blank host", endpointNoPort, false, "", false),
+		table.Entry("return false on blank endpoint, and cdi configmap exists, and host defined", "", true, host, false),
+	)
+})
+
+var _ = Describe("getInsecureTLS", func() {
```
maya-r, line 495: Do you mean getContentType?
(this test doesn't do anything TLS-related)

pkg/controller/util_test.go
```diff
@@ -14,881 +17,149 @@ import (
...
-	}
-
-	type args struct {
-		pvc    *corev1.PersistentVolumeClaim
-		objs   []runtime.Object
-		result string
-	}
-
-	for _, arg := range []args{
-		{&corev1.PersistentVolumeClaim{}, nil, ""},
```
maya-r, line 653: In the interest of completeness, it looks like this case was dropped in the new getCertConfigMap tests (2 cases vs. 3).

maya-r:
otherwise lgtm

pkg/controller/config-controller.go
awels: yes thanks for spotting that.

pkg/controller/import-controller_test.go
awels: I did mean that, copy and paste strikes again.

pkg/controller/util_test.go
awels: You are right, its not a very interesting case, but we should keep it for completeness.

awels:
/hold
need to fix a bunch of tests

pkg/controller/config-controller.go
```diff
@@ -36,9 +34,9 @@ import (
 
 // CDIConfigReconciler members
 type CDIConfigReconciler struct {
-	Client                 client.Client
-	CdiClient              cdiclientset.Interface
-	K8sClient              kubernetes.Interface
+	Client client.Client
```
mhenriks, line 37: Do these fields need to be exported?

awels: Nope, good point, fixed.

awels:
Looks like some GH or PROW issue, it was merged, so closing.

PR #1168, opened 2020-04-06
Convert remaining tests to ginkgo

maya-r:
Let me know if there are some preferences missing in the conversion.
I know Alexander want to get rid of client-go as well, but the commit is pretty noisy as it is.
I'd rather have that change done separately.

**Release note**:
```release-note
NONE
```


pkg/apiserver/auth-config_test.go
```diff
@@ -39,158 +41,153 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/util"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/triple"
+	"kubevirt.io/containerized-data-importer/tests/reporters"
 )
 
-func generateCACert(t *testing.T) string {
+func TestAuthConfig(t *testing.T) {
```
awels, line 47: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: <package>_suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/apiserver/auth-config_test.go
```diff
@@ -39,158 +41,153 @@ import (
...
-	aggregatorClient := aggregatorapifake.NewSimpleClientset()
-	cdiClient := cdiclientfake.NewSimpleClientset()
-	authorizer := &testAuthorizer{}
-	acw := NewAuthConfigWatcher(client, ch).(*authConfigWatcher)
-
-	server, err := NewCdiAPIServer("0.0.0.0", 0, client, aggregatorClient, cdiClient, authorizer, acw, nil)
-	if err != nil {
-		t.Errorf("Upload api server creation failed: %+v", err)
-	}
+		req, _ := http.NewRequest("GET", "/apis", nil)
```
awels, line 128: I am assuming the missing result is the err, and we are expecting err to be nil, so we can do usual 'Expect(err).ToNot(HaveOccurred())'

pkg/apiserver/apiserver_test.go
```diff
@@ -49,6 +52,11 @@ type testAuthorizer struct {
 	err     error
 }
 
+func TestApiServer(t *testing.T) {
```
awels, line 55: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: _suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/apiserver/apiserver_test.go
```diff
@@ -159,159 +147,182 @@ func generateTestKey() (*rsa.PrivateKey, error) {
...
-	}
+func doGetRequest(url string) *httptest.ResponseRecorder {
+	app := &cdiAPIApp{}
+	app.composeUploadTokenAPI()
 
-	signingKeySecret, err := keystest.NewPrivateKeySecret("cdi", apiSigningKeySecretName, signingKey)
-	if err != nil {
-		t.Errorf("error creating secret: %v", err)
-	}
+	req, _ := http.NewRequest("GET", url, nil)
```
awels, line 154: Missing param is err, which we can easily check with Expect(err).ToNot(HaveOccurred())

pkg/apiserver/apiserver_test.go
```diff
@@ -159,159 +147,182 @@ func generateTestKey() (*rsa.PrivateKey, error) {
...
 
-	checkActions(actions, client.Actions(), t)
-}
+		checkActions(actions, client.Actions())
+	})
 
-func TestGetAPIResouceList(t *testing.T) {
-	rr := doGetRequest(t, "/apis/upload.cdi.kubevirt.io/v1alpha1")
+	It("Should generate certs and key on first run", func() {
+		kubeobjects := []runtime.Object{}
```
awels, line 218: Don't think we need the kubeobjects here as we are not adding anything, you can just call: client := k8sfake.NewSimpleClientset() instead

pkg/apiserver/authorizer_test.go
```diff
@@ -30,8 +32,14 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
+	"kubevirt.io/containerized-data-importer/tests/reporters"
 )
 
+func TestAuthorizer(t *testing.T) {
```
awels, line 38: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: _suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/image/filefmt_test.go
```diff
@@ -1,34 +1,28 @@
...
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			if got := CopyKnownHdrs(); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("CopyKnownHdrs() = %v, want %v", got, tt.want)
-			}
-		})
-	}
+func TestFileFmt(t *testing.T) {
```
awels, line 13: This is not needed we already have a suite for this package.

pkg/image/filefmt_test.go
```diff
@@ -1,34 +1,28 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecsWithDefaultAndCustomReporters(t, "File format Suite", reporters.NewReporters())
 }
 
-func TestHeader_Match(t *testing.T) {
+var _ = Describe("File format tests", func() {
+	table.DescribeTable("Copy known headers", func(want Headers, wantErr bool) {
```
awels, line 19: Seems a little overkill to do a table for a single test.

pkg/image/filefmt_test.go
```diff
@@ -46,118 +40,72 @@ func TestHeader_Match(t *testing.T) {
...
+			mgOffset:    fields.mgOffset,
+			SizeOff:     fields.SizeOff,
+			SizeLen:     fields.SizeLen,
+		}
+		got := h.Match(args.b)
+		Expect(got).To(Equal(want))
+	},
+		table.Entry("match gz",
+			fields{"gz", []byte{0x1F, 0x8B}, 0, 0, 0},
+			args{[]byte{0x1F, 0x8B}},
```
awels, line 57: args is a struct with a single byte array as a field, why not just pass the byte array?

pkg/image/filefmt_test.go
```diff
@@ -46,118 +40,72 @@ func TestHeader_Match(t *testing.T) {
...
+		if wantErr {
+			Expect(err).To(HaveOccurred())
+		} else {
+			Expect(err).ToNot(HaveOccurred())
+		}
+		Expect(got).To(Equal(want))
+	},
+		table.Entry("get size of qcow2",
+			fields{"qcow2", []byte{'Q', 'F', 'I', 0xfb}, 0, 24, 8},
+			args{qcowbyte},
```
awels, line 102: same, seems we should just pass the byte array as parameter, instead of an args struct.

pkg/operator/resources/operator/operator_test.go
```diff
@@ -17,75 +19,80 @@ var crdTypeMap = map[string]interface{}{
 	"operator-crd": &cdiv1alpha1.CDI{},
 }
 
-func TestCRDSchemas(t *testing.T) {
-	for crdFileName, cdiObjType := range crdTypeMap {
+func TestOperatorResource(t *testing.T) {
```
awels, line 22: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: _suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/operator/resources/operator/operator_test.go
```diff
@@ -17,75 +19,80 @@ var crdTypeMap = map[string]interface{}{
...
+		  "metadata": {
+		    "name":"cdi",
+		    "namespace":"cdi"
+		  },
+		  "spec": {
+		    "imagePullPolicy":"noValue"
+		  }
+		}`)
+		crFileName, err := yaml.JSONToYAML(crFileName)
+		if err != nil {
```
awels, line 71: Expect(err).ToNot(HaveOccured()) should print the error if it finds an error.

pkg/operator/resources/operator/operator_test.go
```diff
@@ -17,75 +19,80 @@ var crdTypeMap = map[string]interface{}{
...
-
-		t.Log(err)
-	}
-	assert.Errorf(t, err, "File %v does not validate against the CRD schema", crFileName)
-}
+		var input map[string]interface{}
+		err = yaml.Unmarshal([]byte(crFileName), &input)
+		Expect(err).ToNot(HaveOccurred())
+		err = schema.Validate(input)
+		By(fmt.Sprintf("err: %d, File %v does not validate against the CRD schema", err, crFileName))
```
awels, line 82: Missing Expect(err).ToNot(HaveOccurred()), also the By only gets printed when the test fails, if it is the last statement of the test, it will never get printed.

pkg/token/token_test.go
```diff
@@ -38,156 +41,126 @@ func generateTestKey() (*rsa.PrivateKey, error) {
...
-	payload, err := validator.Validate(signedToken)
-
-	if err != nil {
-		t.Errorf("unable to verify token: %v", err)
-	}
-
-	if !reflect.DeepEqual(tokenData, payload) {
-		t.Errorf("invalid token payload")
-	}
+	RegisterFailHandler(Fail)
```
awels, line 44: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: _suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/token/token_test.go
```diff
@@ -38,156 +41,126 @@ func generateTestKey() (*rsa.PrivateKey, error) {
...
-		Name:      "fakepvc",
-		Namespace: "fakenamespace",
-		Resource: metav1.GroupVersionResource{
-			Group:    "",
-			Version:  "v1",
-			Resource: "persistentvolumeclaims",
-		},
-	}
+var _ = Describe("Token test", func() {
+	It("Token", func() {
```
awels, line 49: These 4 are a perfect example of using a DescribeTable in the correct way. Each one basically calls all the same functions but with some different parameters, if you make this a DescribeTable with 4 entries you will reduce the amount of code by like 50%.

pkg/uploadproxy/uploadproxy_test.go
```diff
@@ -31,6 +35,11 @@ type validateSuccess struct{}
 
 type validateFailure struct{}
 
+func TestUploadProxy(t *testing.T) {
```
awels, line 38: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: _suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/uploadserver/uploadserver_test.go
```diff
@@ -35,28 +38,28 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/importer"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/triple"
+	"kubevirt.io/containerized-data-importer/tests/reporters"
 )
 
+func TestUploadServer(t *testing.T) {
```
awels, line 44: In general we try to have a single 'suite' file that has the 'Testing.T' in it. We name it: _suite_test.go with the package here being apiserver. It also allows us to set things like logger in the before suite. A decent example is in pkg/controller/controller_suite_test.go

pkg/uploadserver/uploadserver_test.go
```diff
@@ -204,209 +200,175 @@ func replaceAsyncProcessorFunc(replacement func(io.ReadCloser, string, string, s
...
-			t.Errorf("handler returned wrong status code: got %v want %v",
-				status, http.StatusServiceUnavailable)
-		}
 	})
-}
 
-func TestInProcessUnavailableAsync(t *testing.T) {
-	withProcessorSuccess(func() {
-		req := newAsyncRequest(t)
+	It("Process unavailable, async", func() {
```
awels, line 251: Only difference between this one and the one above is the request type, can use a table with the request as a parameter to reduce code.

pkg/uploadserver/uploadserver_test.go
```diff
@@ -204,209 +200,175 @@ func replaceAsyncProcessorFunc(replacement func(io.ReadCloser, string, string, s
...
+			status := rr.Code
+			Expect(status).To(Equal(http.StatusConflict))
+		})
 	})
-}
 
-func TestCompletedConflictAsync(t *testing.T) {
-	withProcessorSuccess(func() {
-		req := newAsyncRequest(t)
+	It("Completion conflict, async", func() {
```
awels, line 282: Only difference between this one and the one above is the request type, can use a table with the request as a parameter to reduce code. Maybe combine the above 4 into a single table if you pass uploading and done values as a parameter to the table function.

pkg/uploadserver/uploadserver_test.go
```diff
@@ -204,209 +200,175 @@ func replaceAsyncProcessorFunc(replacement func(io.ReadCloser, string, string, s
...
+			Expect(status).To(Equal(http.StatusInternalServerError))
+		})
 	})
-}
 
-func TestStreamFailAsync(t *testing.T) {
-	withAsyncProcessorFailure(func() {
-		req := newAsyncRequest(t)
+	It("stream fail, async", func() {
+		withAsyncProcessorFailure(func() {
```
awels, line 354: Only difference between this one and the one above is the request type, can use a table with the request as a parameter to reduce code.

pkg/uploadserver/uploadserver_test.go
```diff
@@ -204,209 +200,175 @@ func replaceAsyncProcessorFunc(replacement func(io.ReadCloser, string, string, s
...
+			Expect(status).To(Equal(http.StatusOK))
+		})
 	})
-}
 
-func TestSuccessHeadAsync(t *testing.T) {
-	withAsyncProcessorSuccess(func() {
-		req := newAsyncHeadRequest(t)
+	It("Success HEAD, async", func() {
+		withAsyncProcessorSuccess(func() {
```
awels, line 326: Only difference between this one and the one above is the request type, can use a table with the request as a parameter to reduce code.

pkg/token/token_test.go
maya-r: I've avoided doing this because it felt like I'm about to make an error doing that

awels: I am not sure why you feel that way, as you can tell all this code is basically duplicated 4 times with some different parameters. You converted the structs in other places without issues, I don't see why this would be any different.

PR #1175, opened 2020-04-14
Added GetPodName function used to correctly generate pod names in con…

brybacki:
…trollers

CDI creates resources with new names - like short term controllers and "scratch" resources (the name generation has to be deterministic because it is also used to find the resource). Previously the name generation could make too long names. The PR fixes this problem. 

Fixes #

I am not yet 100% sure about the correct const to use as a length for Pod or Pvc in the  pkg/util/util.go: kvalidation.DNS1035LabelMaxLength or  kvalidation.DNS1123SubdomainMaxLength

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -43,6 +43,7 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/fetcher"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/generator"
+	"kubevirt.io/containerized-data-importer/pkg/util/naming"
 )
 
 const (
@@ -141,8 +142,7 @@ func (r *UploadReconciler) reconcilePVC(log logr.Logger, pvc *corev1.PersistentV
 	} else {
 		uploadClientName = uploadServerClientName
 
-		// TODO revisit naming, could overflow
-		scratchPVCName = pvc.Name + "-scratch"
+		scratchPVCName = getScratchPvcName(pvc.Name)
 	}
 
 	resourceName := getUploadResourceName(pvc.Name)
@@ -486,10 +486,14 @@ func addUploadControllerWatches(mgr manager.Manager, importController controller
 	return nil
 }
 
+// getScratchPvcName returns the name given to scratch pvc
+func getScratchPvcName(name string) string {
+	return naming.GetResourceName(name, common.ScratchNameSuffix)
+}
+
 // getUploadResourceName returns the name given to upload resources
 func getUploadResourceName(name string) string {
-	// TODO revisit naming, could overflow
-	return "cdi-upload-" + name
+	return naming.GetResourceName("cdi-upload", name)
 }
 
 // UploadPossibleForPVC is called by the api server to see whether to return an upload token
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -118,6 +118,9 @@ const (
 	// ScratchSpaceNeededExitCode is the exit code that indicates the importer pod requires scratch space to function properly.
 	ScratchSpaceNeededExitCode = 42
 
+	// ScratchNameSuffix (controller pkg only)
+	ScratchNameSuffix = "scratch"
+
 	// UploadTokenIssuer is the JWT issuer of upload tokens
 	UploadTokenIssuer = "cdi-apiserver"
 
--- a/pkg/controller/BUILD.bazel
+++ b/pkg/controller/BUILD.bazel
@@ -23,6 +23,7 @@ go_library(
         "//pkg/util/cert:go_default_library",
         "//pkg/util/cert/fetcher:go_default_library",
         "//pkg/util/cert/generator:go_default_library",
+        "//pkg/util/naming:go_default_library",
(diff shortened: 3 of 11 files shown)


brybacki:
/hold 
there are still few details to be solved before it is ready for merge

pkg/util/util.go
```diff
@@ -266,3 +268,74 @@ func CopyDir(source string, dest string) (err error) {
...
+func GetPodNameWithSuffix(base, suffix string) string {
+	return getName(base, suffix, kvalidation.DNS1123SubdomainMaxLength, false)
+}
+
+// GetPodNameWithPrefix returns a name given a base and a suffix
+func GetPodNameWithPrefix(base, prefix string) string {
+	return getName(base, prefix, kvalidation.DNS1123SubdomainMaxLength, true)
+}
+
+func getName(base, affix string, maxLength int, asPrefix bool) string {
```
mhenriks, line 287: why not import this in from openshift library-go?

If not importing in, may I suggest a new package called `naming`?

pkg/controller/upload-controller.go
```diff
@@ -486,10 +486,14 @@ func addUploadControllerWatches(mgr manager.Manager, importController controller
...
+// getScratchPvcName returns the name given to scratch pvc
+func getScratchPvcName(name string) string {
+	return util.GetPvcNameWithSuffix(name, "scratch")
+}
+
 // getUploadResourceName returns the name given to upload resources
 func getUploadResourceName(name string) string {
-	// TODO revisit naming, could overflow
-	return "cdi-upload-" + name
+	return util.GetPodNameWithPrefix(name, "cdi-upload")
```
mhenriks, line 496: In general, I'd think the "base" part of the name would be the const part (in this case `cdi-upload`) and would go first.

Specifically for upload, I think we should try to have the naming be compatible (there may be an upload pod running during an upgrade) or store the pod name in an annotation of the PVC once calculated.  If the annotation doesn't exist, use the old scheme to determine pod name.

brybacki: Here `GetPodNameWithPrefix` takes name and prefix. Prefix goes first and is not changed (unless it is too long then it is dropped). If name is too long it is shortened. Do you think the order of arguments should be changed here to improve readability?

The thing with backwards compatibility is really important I need to check this. But if the name is not too long then current method returns exactly the same results as previous ("cdi-upload-${name}"). I wonder what happened in previous version if name was too long. That might be not compatible.

pkg/util/util.go
brybacki: Because I had to change it to keep our naming style, for import/upload pods we do add prefix:
"importer-${NAME}" and "cdi-upload-${NAME}", 
and for scratch we use 'scratch' as suffix:
"${NAME}-scratch".

Source pod is named differently - "UID-source-pod", so it was not changed in this PR.

New package sounds reasonable.

brybacki:
> Suggestion: import/upload/clone controllers reconcile on PVCs. So when a pod name is computed for the first time, store it in an annotation in the PVC. And whenever the pod name is needed, check the annotation. If the annotation does not exist (pod created by old version of the controller) use the old naming scheme to look it up.
> 
> Same for scratch PVC.

This sounds like a good idea, to use names just as names and not as a search term, then we can probably even use k8s way of naming (or maybe not, if we want to have user friendly names). It will also be ease to do it in a backwards compatible way. 

> As far as kvalidation.DNS1035LabelMaxLength vs kvalidation.DNS1123SubdomainMaxLength: Pods/PVCs can be kvalidation.DNS1123SubdomainMaxLength. The problem we have is that we stick the names of those things into labels sometimes. And labels can only be kvalidation.DNS1035LabelMaxLength. So we have to stop doing that. Or name things appropriately.
> 
> Exit criteria for this should be a set of unit tests that exercises each of our datavolume operations with max pvc name.

This is something I need to check. Do you know where and why do we use resource names in labels, is it documented somewhere as a way for users to query for cdi resources? Now I see, it might be better to take a step back and rethink what exactly do we want to do with this issue.

pkg/util/util.go
awels: We used uid-source-pod, with the uid being the uid of the target PVC to have a reproducible consistent name for the source pod, if we have a better way of naming it now, I am all for doing that.

mhenriks:
here is one place where the resource name is a label value:  https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/controller/datavolume-controller.go#L736

You may be able to just get rid of that label.  But there may be others.

One way to tackle this may be to remove the restriction on datavolume name length in the webhook.  Then for each operation create datavolumes with max length and see where things break.

pkg/controller/upload-controller.go
brybacki: I have done it as you suggest, so there is now need to write our own namer(I have used openshift/library-go) , and also naming is backwards compatible.

pkg/util/util.go
brybacki: I have update to use openshift/library-go. 
I have left uid-source-pod not changed, since it is well within limits and is good acording to naming rules. The change to this and other names might be a "nice to do" thing - later, and will require special attention for backwards compatibility (like the usage of additional annotation).

brybacki:
This change only covers pvc/pod names.

I am now working on label/service names in another change, will open new PR.

PR #1182, opened 2020-04-16
Add events to operator

awels:

Add events to CDI CR when the operator changes a condition, as well when the operator creates/modifies/deletes an object
Fixes #

No functional test as there is no reliable way to record a particular event as the api server aggregates similar events into a single line.

**Release note**:

```release-note
Enhancement: The operator now emits events on condition changes as well as object CRUD operations
```


pkg/operator/controller/controller_test.go
```diff
@@ -1459,3 +1474,148 @@ func createReconciler(client realClient.Client) *ReconcileCDI {
...
+	return match
+}
+
+func createNotReadyEventValidationMap() map[string]bool {
+	// match is map of strings and if we observed the event.
+	// We are not interested in the order of the events, just that the events happen at least once.
+	match := make(map[string]bool)
+	match["Normal DeployStarted Started Deployment"] = false
+	match["Normal CreateResourceStart Started creation of resource *v1.ClusterRole cdi-apiserver"] = false
+	match["Normal CreateResourceSuccess Successfully created resource *v1.ClusterRole cdi-apiserver"] = false
```
arnongilboa, line 1531: can we use const string prefixes for easier read/maint? e.g.
const normalCreateResourceStart string = "Normal CreateResourceStart Started creation of resource"

pkg/operator/controller/controller_test.go
```diff
@@ -1459,3 +1474,148 @@ func createReconciler(client realClient.Client) *ReconcileCDI {
...
+func createDeleteCDIEventValidationMap() map[string]bool {
+	match := createNotReadyEventValidationMap()
+	match["Normal DeleteResourceStart Started deleting deployment cdi-deployment"] = false
+	match["Normal DeleteResourceSuccess Deleted deployment cdi-deployment successfully"] = false
+	match["Normal DeleteResourceStart Started deleting worker resources"] = false
+	match["Normal DeleteResourceSuccess Deleted worker resources successfully"] = false
+	return match
+}
+
+func createErrorCdiEventValidationMap() map[string]bool {
```
arnongilboa, line 1511: choose either Cdi or CDI for all funcs

pkg/operator/controller/controller_test.go
```diff
@@ -160,6 +161,8 @@ var _ = Describe("Controller", func() {
 				Expect(conditions.IsStatusConditionFalse(args.cdi.Status.Conditions, conditions.ConditionDegraded)).To(BeTrue())
 
 				Expect(args.cdi.Finalizers).Should(HaveLen(1))
+
+				validateEvents(args.reconciler.recorder.(*record.FakeRecorder).Events, createReadyEventValidationMap())
```
mhenriks, line 165: maybe change the signature of validateEvents to take args or args.reconciler to avoid the ugly cast in every invocation?

pkg/operator/controller/controller.go
```diff
@@ -423,30 +445,35 @@ func (r *ReconcileCDI) reconcileUpdate(logger logr.Logger, cr *cdiv1alpha1.CDI)
 			}
 
 			if !reflect.DeepEqual(currentRuntimeObjCopy, currentRuntimeObj) {
+				r.recorder.Event(cr, corev1.EventTypeNormal, updateResourceStart, fmt.Sprintf("Started update of resource %T %s", desiredMetaObj, desiredMetaObj.GetName()))
```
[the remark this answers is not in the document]
awels: During install, it will fire around 50ish events, as you can see from the unit testing.

pkg/operator/controller/controller.go
```diff
@@ -562,8 +589,10 @@ func (r *ReconcileCDI) cleanupUnusedResources(logger logr.Logger, cr *cdiv1alpha
 			}
 
 			if !found && metav1.IsControlledBy(observedMetaObj, cr) {
+				r.recorder.Event(cr, corev1.EventTypeNormal, deleteResourceStart, fmt.Sprintf("Start to delete resource %T %s", observedMetaObj, observedMetaObj.GetName()))
```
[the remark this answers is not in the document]
awels: Well the thought behind this is that it sometimes takes a long time for a particular resource to get deleted. Maybe it has a finalizer on it, and we want to be able to tell if this is the case.

pkg/operator/controller/controller_test.go
awels: Done

pkg/operator/controller/controller_test.go
awels: Good point, fixed.

pkg/operator/controller/controller.go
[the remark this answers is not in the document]
awels: So #1237 reduces the noise significantly, and we are down to large number to start, and only if something really changed do we get some events.

pkg/operator/controller/controller_test.go
brybacki: BTW: the "The Go Programming Language" book suggests all caps for acronyms `CDI` or  all lowercase `cdi`, never mix

pkg/operator/controller/controller.go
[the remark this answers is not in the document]
mhenriks: How about we get rid of `updateResourceStart` and keep failed/success?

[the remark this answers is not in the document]
awels: I am not opposed to doing that. Let me address that.

PR #1183, opened 2020-04-16
Allow for "pre initialized" DataVolumes.

mhenriks:
These are basically no-op DataVolumes for PVCs that already exist.

I see this mostly as an internal thing to facilitate VM restore from snapshot

This is basically an internal API to support VM restore from snapshot.  For that case, we want to have a "no op" DataVolume that references a PVC that is populated from a snapshot.  If the PVC does happen to get deleted, then it will get re-initialized with the import/clone/upload specified by the DataVolume. 

Fixes #

**Release note**:

```release-note
NONE
```


pkg/controller/datavolume-controller.go
```diff
@@ -616,7 +637,7 @@ func (r *DatavolumeReconciler) reconcileDataVolumeStatus(dataVolume *cdiv1.DataV
 
 func (r *DatavolumeReconciler) emitEvent(dataVolume *cdiv1.DataVolume, dataVolumeCopy *cdiv1.DataVolume, curPhase cdiv1.DataVolumePhase, event *DataVolumeEvent) error {
 	// Only update the object if something actually changed in the status.
-	if !reflect.DeepEqual(dataVolume.Status, dataVolumeCopy.Status) {
+	if !reflect.DeepEqual(dataVolume, dataVolumeCopy) {
```
awels, line 640: Won't this cause unwanted updates now?

mhenriks: The controller shouldn't be updating the spec anywhere.  And if it is, it would be great to know and stop it from doing that.

PR #1185, opened 2020-04-17
Wait for CDI config to be available if not found during functional tests

awels:

Sometimes if the quota tests would run right after the CDI operator delete tests, the CDI Config object would not exist yet, and the test would fail. This PR modified the test to be tolerant of the CDI config object not existing yet, and waiting for it to become available.
Fixes #

**Release note**:

```release-note
NONE
```


mhenriks:
Would it make more sense to wait for the CDIConfig here?

https://github.com/kubevirt/containerized-data-importer/blob/master/tests/operator_test.go#L144-L154

awels:
Yeah, I am fixing the cases we know about, but there might be more, and yours would fix them all.

PR #1186, opened 2020-04-17
Add DV restarts test for upload controller

brybacki:

New functional tests for new RESTARTS column in DataVolume (done in PR
https://github.com/kubevirt/containerized-data-importer/pull/1155). 

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -770,7 +770,7 @@ var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		}
 	})
 
-	It("[test_id:3999] Create a data volume importing an image and then clone it and verify retry count", func() {
+	It("[test_id:3999] Create a data volume and then clone it and verify retry count", func() {
 		pvcDef := utils.NewPVCDefinition(sourcePVCName, "1G", nil, nil)
 		pvcDef.Namespace = f.Namespace.Name
 		sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile)
@@ -796,7 +796,7 @@ var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		Expect(dv.Status.RestartCount).To(BeNumerically("==", 0))
 	})
 
-	It("[test_id:4000] Create a data volume importing an image and then clone it while killing the container and verify retry count", func() {
+	It("[test_id:4000] Create a data volume and then clone it while killing the container and verify retry count", func() {
 		By("Prepare source PVC")
 		pvcDef := utils.NewPVCDefinition(sourcePVCName, "1G", nil, nil)
 		pvcDef.Namespace = f.Namespace.Name
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"fmt"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


tests/framework/exec_util.go
```diff
@@ -143,8 +143,8 @@ func (f *Framework) ExecShellInPod(podName, namespace string, cmd string) (strin
 }
 
 // ExecShellInPodWithFullOutput provides a function to execute a shell cmd in a running pod and to capture its output
-func (f *Framework) ExecShellInPodWithFullOutput(podName string, cmd string) (string, string, error) {
-	return f.ExecCommandInPodWithFullOutput(podName, "/bin/sh", "-c", cmd)
+func (f *Framework) ExecShellInPodWithFullOutput(namespce, podName string, cmd string) (string, string, error) {
```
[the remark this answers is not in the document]
brybacki: fixed

tests/cloner_test.go
```diff
@@ -750,6 +751,104 @@ var _ = Describe("Namespace with quota", func() {
...
+			err := f.DeletePVC(targetPvc)
+			Expect(err).ToNot(HaveOccurred())
+		}
+	})
+
+	It("[test_id:3999] Create a data volume importing an image and then clone it and verify retry count", func() {
+		pvcDef := utils.NewPVCDefinition(sourcePVCName, "1G", nil, nil)
+		pvcDef.Namespace = f.Namespace.Name
+		sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile)
+		targetNs, err := f.CreateNamespace(f.NsPrefix, map[string]string{
```
awels, line 777: This is more an FYI than an issue. The targetNS and addNamespaceToDelete are for when you are cloning across namespaces. This particular test we don't care about that and could have used the same namespace and saved us writing some code.

tests/cloner_test.go
```diff
@@ -750,6 +751,104 @@ var _ = Describe("Namespace with quota", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+		}
+		if targetPvc != nil {
+			By("[AfterEach] Clean up target PVC")
+			err := f.DeletePVC(targetPvc)
+			Expect(err).ToNot(HaveOccurred())
+		}
+	})
+
+	It("[test_id:3999] Create a data volume importing an image and then clone it and verify retry count", func() {
```
awels, line 773: Technically we are not importing an image first here, we are filling it with some data using the filler pod. Also where did the test_id come from?

tests/cloner_test.go
```diff
@@ -750,6 +751,104 @@ var _ = Describe("Namespace with quota", func() {
...
+		Expect(status).To(BeTrue())
+		Expect(restartsValue).To(Equal("0"))
+
+		By("Verify the number of retries on the datavolume")
+		dv, err := f.CdiClient.CdiV1alpha1().DataVolumes(targetNs.Name).Get(targetDvName, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(dv.Status.RestartCount).To(BeNumerically("==", 0))
+	})
+
+	It("[test_id:4000] Create a data volume importing an image and then clone it while killing the container and verify retry count", func() {
```
awels, line 799: Technically we are not importing an image first here, we are filling it with some data using the filler pod. Also where did the test_id come from?

tests/cloner_test.go
```diff
@@ -750,6 +751,104 @@ var _ = Describe("Namespace with quota", func() {
...
+		By("Prepare source PVC")
+		pvcDef := utils.NewPVCDefinition(sourcePVCName, "1G", nil, nil)
+		pvcDef.Namespace = f.Namespace.Name
+		sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile)
+
+		By("Create clone DV")
+		targetNs, err := f.CreateNamespace(f.NsPrefix, map[string]string{
```
awels, line 806: This is more an FYI than an issue. The targetNS and addNamespaceToDelete are for when you are cloning across namespaces. This particular test we don't care about that and could have used the same namespace and saved us writing some code.

tests/import_test.go
```diff
@@ -513,3 +515,102 @@ var _ = Describe("Namespace with quota", func() {
...
+
+	var (
+		pvc                  *v1.PersistentVolumeClaim
+		dataVolume           *cdiv1.DataVolume
+		err                  error
+		tinyCoreIsoURL       = fmt.Sprintf(utils.TinyCoreIsoURL, f.CdiInstallNs)
+		invalidQcowImagesURL = fmt.Sprintf(utils.InvalidQcowImagesURL, f.CdiInstallNs)
+	)
+
+	BeforeEach(func() {
```
awels, line 530: Don't really need an empty before each.

tests/cloner_test.go
brybacki: I thought that smart clone might be used if I do not use two distinct ns, and then I'll not be able to break the pod.

But probably not.

awels: That is a good thought, but we have lanes that do not use smart clone and we would be testing in those lanes. But this test would fail in the lanes that do use smart clone, so we would have to skip the test in those lanes. We have tests we skip already in that particular case.

tests/cloner_test.go
brybacki: right, 
removing the "importing" part from name, 
test_id comes from test plan

tests/cloner_test.go
brybacki: the same as in previous test

tests/import_test.go
brybacki: good catch, a leftover after some changes

PR #1193, opened 2020-05-03
Implement local webserver to mimic flawed servers (for functional tests)

maya-r:
It's a functional test follow-up to #1122 
We've seen servers in the wild return errors on HEAD requests, having our own makes us certain that we handle this situation correctly. 
I've used the opportunity to also test HTTP import from flaky web servers.

**Release note**:
```release-note
NONE
```


maya-r:
/test pull-containerized-data-importer-e2e-k8s-1.16.2-nfs
checking to see if the CI issues were magically resolved

tools/cdi-func-test-bad-webserver/main.go
```diff
@@ -0,0 +1,48 @@
...
+	redirect(w, r)
+}
+
+func flaky(w http.ResponseWriter, r *http.Request) {
+	random, err := rand.Int(rand.Reader, big.NewInt(20))
+	if err != nil {
+		panic(err)
+	}
+	if random.Cmp(big.NewInt(0)) == 0 {
+		// 1-in-20 odds of success
```
awels, line 26: 1 in 20 odds of success, but if I can only attempt 10 tries the allotted time, then the odds of not getting success are pretty high. This will result in a flaky test. I think its better to always return success after X number of tries, where we know X will be within the retry limit of a certain period (crash looping pods have exponential backoff). Also do we want to make sure the first try always fails. With this code there is a 1 in 20 chance the first try will succeed and if the test is expecting a failure that will be a problem.

awels:
Looks like the bad server container is crash looping in the OS-3.11 lane.

maya-r:
/hold
one small change to go

tools/cdi-func-test-bad-webserver/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+func initCounter() {
+	newCounterBuf := make([]byte, binary.MaxVarintLen64)
+	_ = binary.PutUvarint(newCounterBuf, 0)
+
+	ioutil.WriteFile("state", newCounterBuf, 0644)
+}
+
+func getCounter() uint64 {
+	counterBuf, err := ioutil.ReadFile("state")
```
awels, line 47: So I think there is a potential race condition here:
The http server can get 2 requests at the same time, and the counter can get messed up potentially. However I don't think it is an issue in this particular case, just wanted to make you aware of it. A lock and a flush should ensure the values are always incremented correctly.

PR #1194, opened 2020-05-04
Add conditions to DataVolume and add condition annotation to PVCs

awels:

Add conditions to data volumes so we can derive the status of the datavolume from them.
There are 3 conditions:
1. ReadyCondition, if true, the DataVolume is ready.
2. BoundCondition, if true, the underlying PVC has been bound.
3. RunningCondition, if true, the population step of the DataVolume is still running. This usually means a pod is working on import/upload/clone.

All 3 conditions can go from true to false and vice versa based on what is happening in the cluster. The running and ready conditions are mutually exclusive. You cannot be ready and running at the same time.

Each condition has 6 fields:
- Type
- Status
- Reason
- Message
- LastTransitionTime
- LastHeartBeatTime

Fixes #

**Release note**:

```release-note
Enhancement: DataVolumes now have 3 conditions, Ready, Bound, and Running.
```


awels:
/hold
I need to add documentation and reason codes.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -141,12 +143,25 @@ type DataVolumeList struct {
 	Items []DataVolume `json:"items"`
 }
 
+// DataVolumeCondition represents the state of a data volume condition.
+type DataVolumeCondition struct {
```
mhenriks, line 147: The timestamp fields on most conditions are missing `lastTransitionTime` and `lastProbeTime`/`LastHeartbeatTime`

pkg/apis/core/v1alpha1/types.go
```diff
@@ -141,12 +143,25 @@ type DataVolumeList struct {
...
+	Type   DataVolumeConditionType `json:"type" description:"type of condition ie. Ready|Bound|Running."`
+	Status corev1.ConditionStatus  `json:"status" description:"status of the condition, one of True, False, Unknown"`
+	// +optional
+	Reason string `json:"reason,omitempty" description:"one-word CamelCase reason for the condition's last transition"`
```
mhenriks, line 151: Why only one word?

pkg/apis/core/v1alpha1/types.go
awels: I purposely didn't add transition/lastprobe/heartbeart/other time. After doing some research I noticed basically the only conditions that used that was the conditions we put in for HCO. The time didn't add much/any value so why bother adding it?

pkg/apis/core/v1alpha1/types.go
awels: So I can easily copy to/from event reasons without having to modify anything.

pkg/apis/core/v1alpha1/types.go
mhenriks: I think there is value.  At least in transition time.  And I've never encountered conditions without timestamps.  Here are some places that do:

https://github.com/kubevirt/kubevirt/blob/master/staging/src/kubevirt.io/client-go/api/v1/types.go#L266-L284

https://github.com/kubevirt/kubevirt/blob/master/staging/src/kubevirt.io/client-go/api/v1/types.go#L664-L671

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L448-L460

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2250-L2262

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L4009-L4021

pkg/apis/core/v1alpha1/types.go
mhenriks: That sounds easy but maybe more detailed info would be helpful?

awels: Well the message has more detail.

pkg/apis/core/v1alpha1/types.go
awels: Okay, I have added transition time.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -129,6 +129,8 @@ type DataVolumeStatus struct {
 	Phase        DataVolumePhase    `json:"phase,omitempty"`
 	Progress     DataVolumeProgress `json:"progress,omitempty"`
 	RestartCount int32              `json:"restartCount"`
+	// +listType=set
+	Conditions []*DataVolumeCondition `json:"conditions,omitempty" optional:"true"`
```
mhenriks, line 133: Unusual to have slice of pointers in k8s api.  `Conditions []DataVolumeCondition` preferred

awels: Yeah, was having issue with getting to the backing data of the slice, for some reason this:
```go
for _, cond := range conditions {
  return &cond
}
```
doesn't give me the backing data, but
```go
for i, cond := range conditions {
  return &conditions[i]
}
```
does

mhenriks: In your first example, `cond` is a copy of the slice item

awels: Yeah I had initially not realized this, that is why I did the pointer, that way the pointer was the copy and I was able to modify the actual struct. Once I realized its a copy, and I could return the pointer in the second way there was no reason for the pointer slice.

pkg/controller/datavolume-conditions.go
```diff
@@ -0,0 +1,158 @@
...
+			Type: conditionType,
+		})
+		condition = findConditionByType(conditionType, conditions)
+	}
+	if condition.Status != status {
+		condition.LastTransitionTime = metav1.Now()
+	}
+	condition.Status = status
+	condition.Message = message
+	condition.Reason = reason
```
mhenriks, line 58: I think we should also only update message/reason when the status changes

pkg/controller/datavolume-conditions.go
```diff
@@ -0,0 +1,158 @@
...
+		conditions = append(conditions, cdiv1.DataVolumeCondition{
+			Type: cdiv1.DataVolumeRunning,
+		})
+		condition = findConditionByType(cdiv1.DataVolumeRunning, conditions)
+	}
+	if val, ok := anno[AnnRunningConditionMessage]; ok {
+		condition.Message = val
+	} else {
+		condition.Message = ""
+	}
```
mhenriks, line 74: above if/else is equivalent to `val := anno[AnnRunningConditionMessage]`

awels: Hmm I thought it would error if the key didn't exist, but it just return blank. will fix.

pkg/controller/datavolume-conditions.go
awels: I went back and forth on this, and decided on always updating because its possible to stay at the same condition but have a different message/reason, I thought it was important to always have the latest message/reason, but if I do that, I should have the last transition time also be dependent on reason/message being different.

pkg/controller/datavolume-conditions.go
```diff
@@ -0,0 +1,158 @@
...
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+)
+
+const (
+	boundFalse      = "BoundChangeFalse"
+	transferRunning = "TransferRunning"
+	pvcBound        = "PVCBound"
+	pvcPending      = "PVCPending"
+	claimLost       = "ClaimLost"
+	notFound        = "NotFound"
```
mhenriks, line 33: Not clear to me why for example `PVCPending` is preferable to `PVC is pending` for condition reason

pkg/controller/datavolume-conditions.go
mhenriks: I think the original reason is more important and don't want to lose it

pkg/controller/import-controller.go
```diff
@@ -58,6 +58,12 @@ const (
 	AnnRequiresScratch = AnnAPIGroup + "/storage.import.requiresScratch"
 	// AnnDiskID provides a const for our PVC diskId annotation
 	AnnDiskID = AnnAPIGroup + "/storage.import.diskId"
+	// AnnRunningCondition provides a const for the running condition
+	AnnRunningCondition = AnnAPIGroup + "/storage.condition.running"
+	// AnnRunningConditionMessage provides a const for the running condition
+	AnnRunningConditionMessage = AnnAPIGroup + "/storage.condition.running.message"
```
mhenriks, line 64: maybe `AnnLastTerminationMessage = AnnAPIGroup + "/storage.condition.termination.message"` makes more sense?

pkg/controller/import-controller.go
```diff
@@ -243,12 +254,12 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
...
 			r.recorder.Event(pvc, corev1.EventTypeWarning, ErrImportFailedPVC, pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Message)
+			anno[AnnRunningCondition] = "false"
+			anno[AnnRunningConditionMessage] = pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Message
+			anno[AnnRunningConditionReason] = pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Reason
 		}
 	}
 
-	if pod.Status.ContainerStatuses != nil {
-		anno[AnnPodRestarts] = strconv.Itoa(int(pod.Status.ContainerStatuses[0].RestartCount))
-	}
```
mhenriks, line 251: Why was this deleted?

pkg/controller/smart-clone-controller.go
```diff
@@ -239,11 +239,17 @@ func (r *SmartCloneReconciler) updateSmartCloneStatusPhase(phase cdiv1.DataVolum
...
+		dataVolume.Status.Conditions = updateBoundCondition(dataVolume.Status.Conditions, newPVC)
+		dataVolume.Status.Conditions = updateReadyCondition(dataVolume.Status.Conditions, corev1.ConditionFalse, "", "")
+		dataVolume.Status.Conditions = updateCondition(dataVolume.Status.Conditions, cdiv1.DataVolumeRunning, corev1.ConditionTrue, MessageSmartClonePVCInProgress, SmartClonePVCInProgress)
 	case cdiv1.Succeeded:
 		dataVolumeCopy.Status.Phase = cdiv1.Succeeded
 		event.eventType = corev1.EventTypeNormal
 		event.reason = CloneSucceeded
 		event.message = fmt.Sprintf(MessageCloneSucceeded, dataVolumeCopy.Spec.Source.PVC.Namespace, dataVolumeCopy.Spec.Source.PVC.Name, newPVC.Namespace, newPVC.Name)
+		dataVolume.Status.Conditions = updateBoundCondition(dataVolume.Status.Conditions, newPVC)
+		dataVolume.Status.Conditions = updateReadyCondition(dataVolume.Status.Conditions, corev1.ConditionTrue, "", "")
```
mhenriks, line 251: Why no reason/message for ready?

pkg/controller/util.go
```diff
@@ -296,3 +300,23 @@ func podPhaseFromPVC(pvc *v1.PersistentVolumeClaim) v1.PodPhase {
 func podSucceededFromPVC(pvc *v1.PersistentVolumeClaim) bool {
 	return (podPhaseFromPVC(pvc) == v1.PodSucceeded)
 }
+
+func setConditionFromPod(anno map[string]string, pod *v1.Pod) {
```
mhenriks, line 304: This cannot be used for upload?

pkg/controller/import-controller.go
awels: Sounds good

pkg/controller/import-controller.go
awels: its part of setConditionFromPod now

pkg/controller/smart-clone-controller.go
awels: Because the Ready Condition is mutually exclusive with the Running Condition, and I wanted the messages to come from the running condition.

pkg/controller/util.go
awels: Its because I moved
```
anno[AnnPodRestarts] = strconv.Itoa(int(pod.Status.ContainerStatuses[0].RestartCount))
```
In here, which is different for upload. But I think that is a mistake, if I leave it out I can re-use the rest in upload

pkg/controller/import-controller.go
awels: Which is probably not a good plan, so reverting that.

pkg/controller/import-controller.go
```diff
@@ -59,6 +59,12 @@ const (
 	AnnRequiresScratch = AnnAPIGroup + "/storage.import.requiresScratch"
 	// AnnDiskID provides a const for our PVC diskId annotation
 	AnnDiskID = AnnAPIGroup + "/storage.import.diskId"
+	// AnnRunningCondition provides a const for the running condition
+	AnnRunningCondition = AnnAPIGroup + "/storage.condition.running"
+	// AnnLastTerminationMessage provides a const for the running condition
+	AnnLastTerminationMessage = AnnAPIGroup + "/storage.condition.running.message"
+	// AnnRunningConditionReason provides a const for the running condition
+	AnnRunningConditionReason = AnnAPIGroup + "/storage.condition.termination.message"
```
mhenriks, line 67:
```	
AnnLastTerminationMessage = AnnAPIGroup + "/storage.condition.termination.message"

AnnLastTerminationReason = AnnAPIGroup + "/storage.condition.termination.reason"
```

awels: Done

pkg/controller/import-controller.go
```diff
@@ -244,12 +255,16 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 			anno[AnnRequiresScratch] = "true"
 		} else {
 			r.recorder.Event(pvc, corev1.EventTypeWarning, ErrImportFailedPVC, pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Message)
+			anno[AnnRunningCondition] = "false"
+			anno[AnnLastTerminationMessage] = pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Message
+			anno[AnnLastTerminationReason] = pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Reason
```
mhenriks, line 260: is this redundant with `setConditionFromPod` called above?

pkg/controller/import-controller.go
```diff
@@ -261,6 +276,9 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 				return err
 			}
 		}
+		anno[AnnRunningCondition] = "false"
+		anno[AnnLastTerminationMessage] = "Creating scratch space"
+		anno[AnnLastTerminationReason] = creatingScratch
```
mhenriks, line 281: These are set when the pod is pending and needs scratch space.  But when the pod is no longer pending, it will display the real message/reason.  Is that okay?

Also, do any import types start out with scratch space?  If so it will never make sense to do this.

pkg/controller/upload-controller.go
```diff
@@ -289,12 +291,18 @@ func (r *UploadReconciler) getOrCreateUploadPod(pvc *v1.PersistentVolumeClaim, p
...
 	scratchPvc := &corev1.PersistentVolumeClaim{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: pvc.Namespace}, scratchPvc); err != nil {
 		if !k8serrors.IsNotFound(err) {
 			return nil, errors.Wrap(err, "error getting scratch PVC")
 		}
 
+		anno[AnnRunningCondition] = "false"
+		anno[AnnLastTerminationMessage] = "Creating scratch space"
+		anno[AnnLastTerminationReason] = creatingScratch
+
```
mhenriks, line 305: Not sure this ever makes sense.  Uploads should always have a scratch PVC and clones should never, right?

pkg/controller/import-controller.go
awels: There are certain import types we know will have scratch space, in particular registry import.

pkg/controller/util.go
```diff
@@ -296,3 +320,47 @@ func podPhaseFromPVC(pvc *v1.PersistentVolumeClaim) v1.PodPhase {
 func podSucceededFromPVC(pvc *v1.PersistentVolumeClaim) bool {
 	return (podPhaseFromPVC(pvc) == v1.PodSucceeded)
 }
+
+func setConditionFromPodWithPrefix(anno map[string]string, prefix string, pod *v1.Pod) {
```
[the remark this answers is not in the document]
awels: Done

pkg/controller/clone-controller.go
```diff
@@ -297,7 +300,13 @@ func (r *CloneReconciler) validateSourceAndTarget(targetPvc *corev1.PersistentVo
 		return err
 	}
 
-	return ValidateCanCloneSourceAndTargetSpec(&sourcePvc.Spec, &targetPvc.Spec)
+	err = ValidateCanCloneSourceAndTargetSpec(&sourcePvc.Spec, &targetPvc.Spec)
+	if err == nil {
+		// Validation complete, put source PVC bound status in annotation
+		setBoundConditionFromPVC(targetPvc.GetAnnotations(), AnnSourceRunningCondition, sourcePvc)
```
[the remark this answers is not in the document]
awels: That is a mistake should be AnnBoundCondition

pkg/controller/datavolume-controller_test.go
```diff
@@ -351,6 +373,39 @@ var _ = Describe("Reconcile Datavolume status", func() {
...
+func boundMessageByPVCPhase(pvcPhase corev1.PersistentVolumeClaimPhase) string {
+	if pvcPhase == corev1.ClaimBound {
+		return "PVC Bound"
+	} else if pvcPhase == corev1.ClaimPending {
+		return "PVC Pending"
+	} else if pvcPhase == corev1.ClaimLost {
+		return "Claim Lost"
+	}
+	return "No PVC found"
+}
```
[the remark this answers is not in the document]
awels: Done

pkg/controller/datavolume-controller.go
```diff
@@ -631,9 +631,38 @@ func (r *DatavolumeReconciler) reconcileDataVolumeStatus(dataVolume *cdiv1.DataV
...
+	}
+
+	if pvc != nil {
+		anno = pvc.Annotations
+	} else {
+		anno = make(map[string]string, 0)
+	}
+
+	readyStatus := corev1.ConditionUnknown
+	switch phase := dataVolume.Status.Phase; phase {
```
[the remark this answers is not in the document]
awels: No clue, probably copy and paste from somewhere else.

pkg/controller/datavolume-conditions.go
```diff
@@ -0,0 +1,159 @@
...
+		conditions = updateCondition(conditions, cdiv1.DataVolumeRunning, corev1.ConditionUnknown, anno[AnnRunningConditionMessage], anno[AnnRunningConditionReason])
+	}
+	return conditions
+}
+
+func updateReadyCondition(conditions []cdiv1.DataVolumeCondition, status corev1.ConditionStatus, message, reason string) []cdiv1.DataVolumeCondition {
+	return updateCondition(conditions, cdiv1.DataVolumeReady, status, message, reason)
+}
+
+func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.PersistentVolumeClaim) []cdiv1.DataVolumeCondition {
```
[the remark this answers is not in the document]
awels: Well the annotation can also be the bound status of the source PVC, if we are cloning. But it might be a good idea to put the logic in the respective controller instead of the data volume controller, and have the data volume controller just read the annotation.

pkg/controller/datavolume-conditions.go
```diff
@@ -0,0 +1,159 @@
...
+		condition.Message = message
+		condition.Reason = reason
+		condition.LastHeartbeatTime = metav1.Now()
+	}
+	condition.Status = status
+	return conditions
+}
+
+func updateRunningCondition(conditions []cdiv1.DataVolumeCondition, anno map[string]string) []cdiv1.DataVolumeCondition {
+	if val, ok := anno[AnnRunningCondition]; ok {
```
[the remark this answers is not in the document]
awels: Well I can replace some if/else with a switch to make it more readable, but I have to calculate combinations, I guess I can split it out into several functions to make it clearer?

pkg/controller/import-controller.go
```diff
@@ -261,7 +267,16 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 				return err
 			}
 		}
+		anno[AnnBoundCondition] = "false"
```
mhenriks, line 270: Isn't it kind of weird that we don't actually check that the PVC is bound?  Rather assume based on the fact that the pod is pending and scratch space is required?  Could be pending for other reasons.

pkg/controller/upload-controller.go
```diff
@@ -297,11 +301,18 @@ func (r *UploadReconciler) getOrCreateScratchPvc(pvc *v1.PersistentVolumeClaim,
...
+		anno[AnnBoundConditionReason] = creatingScratch
 		// Scratch PVC doesn't exist yet, create it.
 		scratchPvc, err = CreateScratchPersistentVolumeClaim(r.client, pvc, pod, name, storageClassName)
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		delete(anno, AnnBoundCondition)
+		delete(anno, AnnBoundConditionMessage)
+		delete(anno, AnnBoundConditionReason)
```
mhenriks, line 315: Not actually checking that the PVC is bound or not only it's existence

pkg/controller/import-controller.go
awels: Good point, updated to check scratch PVC status, and put this condition in a different function.

pkg/controller/upload-controller.go
awels: Good point fixed.

PR #1200, opened 2020-05-08
Generating label names

brybacki:

Removes unused labels holding resource names. If any label is used then its name will be shortened by provided function in package naming. 

Fixes #

**Release note**:

```release-note
NONE
```


awels:
In general looks good, but can we think of any functional tests that we can do here? I don't want to merge again and having to add functional tests after the fact.

brybacki:
Datavolume webhook still guards names (len <=55) so I can only test import using pvc for this PR. 
But when https://github.com/kubevirt/containerized-data-importer/pull/1201 is finished, we can decide if we want to lift the name length limit, and follow with tests for import/upload/clone (to test if pod/pvc/label/service names are ok)

awels:
Looks like a unit test failure: https://travis-ci.org/github/kubevirt/containerized-data-importer/jobs/689329082#L7732

brybacki:
pushed 'Focused' test by mistake, and now after turning off the focus I have found a problem to fix

pkg/controller/datavolume-controller.go
```diff
@@ -790,8 +790,7 @@ func buildHTTPClient() *http.Client {
 // that 'owns' it.
 func newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume) (*corev1.PersistentVolumeClaim, error) {
 	labels := map[string]string{
-		"cdi-controller": dataVolume.Name,
```
awels, line 793: So I know these don't look like they are being used, but we need them if we want to change the PVC names to not match the DV names. Tomasz's PR is using these.

brybacki: If he is using this from import/upload/clone controller to find dv then annotation would be easier/better (or even the owner reference)

but if he needs it in datavolume controller to find pvc, then it is not so simple, I need to check that PR

awels: He needs it in the datavolume controller to find the PVC, and in some other places like tests. So definitely coordinate because that is currently my issue with this PR, other than that looks pretty good.

awels: Okay, we discussed, and we are fine with removing these labels.

pkg/controller/import-controller.go
```diff
@@ -247,6 +285,7 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 			log.V(1).Info("Pod requires scratch space, terminating pod, and restarting with scratch space", "pod.Name", pod.Name)
 			scratchExitCode = true
 			anno[AnnRequiresScratch] = "true"
+			anno[AnnScratchPvc] = scratchNameFromPvc(pvc)
```
awels, line 288: This one really needed? Looks like we create it above when we determine that we need scratch space.

pkg/controller/import-controller.go
```diff
@@ -181,14 +183,14 @@ func (r *ImportReconciler) Reconcile(req reconcile.Request) (reconcile.Result, e
...
-	podName := importPodNameFromPvc(pvc)
+	podName := getImportPodNameFromPvc(pvc)
 	pod := &corev1.Pod{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: pvc.GetNamespace()}, pod)
 
 	if k8serrors.IsNotFound(err) {
 		return nil, nil
 	}
-
+	/// TODO: what about other errors? // check upload
```
awels, line 193: Not sure what you mean by this comment.

pkg/controller/import-controller.go
```diff
@@ -208,9 +210,18 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 			// Don't create the POD if the PVC is completed already
 			log.V(1).Info("PVC is already complete")
 		} else if pvc.DeletionTimestamp == nil {
-			// Create importer pod, make sure the PVC owns it.
-			if err := r.createImporterPod(pvc); err != nil {
-				return reconcile.Result{}, err
+
+			_, ok := pvc.Annotations[AnnImportPod]
```
awels, line 214: This is more a stylistic thing than a code problem, but I would write this as:
```go
if _, ok := pvc.Annotations[AnnImportPod]; ok {
...
} else {
...
}
```
Second, what is the difference between initPvcPodName vs createImporterPod. I see initPvcPodName is creating the name and adding it to the annotations, but can't we do that in createImporterPod as well? I am trying to understand why this is split.

pkg/controller/import-controller.go
```diff
@@ -332,10 +371,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
 		return err
 	}
 	r.log.V(1).Info("Created POD", "pod.Name", pod.Name)
-	if requiresScratch {
```
awels, line 335: Why did this get removed? its important to actually create the scratch space. And I don't see an other place creating it.

pkg/controller/upload-controller.go
```diff
@@ -500,7 +544,19 @@ func getScratchPvcName(name string) string {
...
+		return podName
+	}
+
+	// fallback to legacy naming, in fact the following function is fully compatible with legacy
+	// name concatenation "cdi-upload-{pvc.Name}" if the name length is under the size limits,
+	return naming.GetResourceName("cdi-upload", pvc.Name)
+}
+
+// createUploadResourceName returns the name given to upload resources
+func createUploadResourceName(name string) string {
```
awels, line 559: So I am a little confused at this point, this function is called always with the pvc name, and the result is stored in the annotation. But why are we doing that? This should be a deterministic result based on the pvc name, so instead of storing and reading from the annotation, why not simply call this function whenever we are interested in finding the pod name based on the pvc name?

pkg/controller/import-controller.go
brybacki: it is done anyway in updatePVCFromPod, so here in createPod it is done/tried first time, 
looks like I removed it during the thinking/experimenting process, but yes removal was not needed

pkg/controller/upload-controller.go
brybacki: So you are right, our function is now deterministic, and currently storing in annotation does not help us. So it has no value now, but might help us in the future:
- if we decide to resolve duplicate names/conflicts by trying some other name. 
- if we decide to make name generation more random/unique. 
- if we decide to 'change' the naming method, we might use annotation for backwards compatibility.

Now after you comment, I think, this "store in annotation" commit", and the next one, are nice to have, but really are maybe over engineering for current state. 

I mean - keeping it here, and improving some tests might be beneficial (for the future needs?), but removing it - might simplify the code. So I am open for hearing arguments for one of this directions. Use annotations to store names, or do not (since naming is deterministic)
 
@mhenriks @awels What do you think?

pkg/controller/import-controller.go
brybacki: if r.client.Get does not wok (service error, temporary network error) then we go straight to check if pod is controlled by pvc, and the only error we can see is Pod is not owned by PVC.
```go
	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: pvc.Namespace}, pod); err != nil {
			return nil, errors.Wrapf(err, "error getting import pod %s/%s", pvc.Namespace, podName)
		}
		return nil, nil
	}

	if !metav1.IsControlledBy(pod, pvc) {
		return nil, errors.Errorf("%s pod not controlled by pvc %s", podName, pvc.Name)
	}
```

pkg/controller/import-controller.go
brybacki: see my last comment on storing names in annotations

brybacki: actually not, we create scratch name once, and then it is always stored in a Volume definition on the pod

pkg/controller/import-controller.go
brybacki: I have actually applied the change from previous comment

pkg/controller/import-controller.go
brybacki: so I have removed this annotation, as it is not needed, and change to read from pod spec, if available

pkg/controller/import-controller.go
brybacki: This was started with a plan to change name generation logic, but now as we can see the name generation is backwards compatible, so this does not make sense/does not help, I can simplify

brybacki:
I have rebased, so it is up to date, and added small improvements (last 3 commits), and if we decide how to proceed with annotations (comments on upload-controller.go) then it is finished.

pkg/controller/import-controller.go
awels: So this is here for the case where there is no pod yet, where we know ahead of time we need scratch space, like for upload and import from registry. so if we remove this code, then for upload and registry we have a create pod, determine it need scratch space there, delete pod, create scratch space, create pod. This just removes that cycle.

pkg/apiserver/apiserver.go
```diff
@@ -324,6 +324,7 @@ func (app *cdiAPIApp) uploadHandler(request *restful.Request, response *restful.
 
 	tokenData := &token.Payload{
 		Operation: token.OperationUpload,
+		// this is service or pod but not pvc?
```
mhenriks, line 327: No it is a PVC

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -57,11 +57,10 @@ func validateSourceURL(sourceURL string) string {
 
 func validateDataVolumeName(name string) []metav1.StatusCause {
 	var causes []metav1.StatusCause
-	// name of data volume cannot be more than 55 characters (not including '-scratch')
-	if len(name) > 55 {
+	if len(name) > 253 {
```
mhenriks, line 60: why not use the constant?

pkg/controller/import-controller.go
brybacki: In that case, I'll bring it back, but, when does the pod check for scratch - if it exists?
 
Because between creating the pod and the scratch - there is for sure a non zero time window, so I imagine on some clusters pod might start without having scratch ready. Can we create a scratch before pod (probably not because it is an owner)

awels: So the pod doesn't start unless all the PVCs it requires are available. So the pod will remain pending until the PVC is bound to a PV.

pkg/controller/import-controller.go
```diff
@@ -582,7 +632,7 @@ func createImporterPod(log logr.Logger, client client.Client, image, verbose, pu
 // makeImporterPodSpec creates and return the importer pod spec based on the passed-in endpoint, secret and pvc.
 func makeImporterPodSpec(namespace, image, verbose, pullPolicy string, podEnvVar *importPodEnvVar, pvc *corev1.PersistentVolumeClaim, scratchPvcName *string, podResourceRequirements *corev1.ResourceRequirements) *corev1.Pod {
 	// importer pod name contains the pvc name
-	podName := importPodNameFromPvc(pvc)
+	podName := getImportPodNameFromPvc(pvc)
```
mhenriks, line 635: Will the annotation always be populated at this point?  If so, why not use that value, like above?  I think it would be ideal to only call `getImportPodNameFromPvc()` once per reconcile loop

pkg/controller/import-controller.go
```diff
@@ -316,7 +349,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
 
 	requiresScratch := r.requiresScratchSpace(pvc)
 	if requiresScratch {
-		name := scratchNameFromPvc(pvc)
+		name := createScratchNameFromPvc(pvc)
```
mhenriks, line 352: scratch pvc name not stored in annotation?

pkg/controller/import-controller.go
```diff
@@ -625,9 +675,7 @@ func makeImporterPodSpec(namespace, image, verbose, pullPolicy string, podEnvVar
 			Labels: map[string]string{
 				common.CDILabelKey:       common.CDILabelValue,
 				common.CDIComponentLabel: common.ImporterPodName,
-				// this label is used when searching for a pvc's import pod.
-				LabelImportPvc:         pvc.Name,
```
mhenriks, line 629: can `LabelImportPvc` const be deleted?

pkg/controller/util.go
```diff
@@ -365,3 +364,17 @@ func setBoundConditionFromPVC(anno map[string]string, prefix string, pvc *v1.Per
 		anno[prefix+".reason"] = "Unknown"
 	}
 }
+
+func getScratchNameFromPod(pod *v1.Pod) string {
```
mhenriks, line 368: Maybe just use an annotation?

pkg/controller/clone-controller.go
```diff
@@ -267,7 +267,7 @@ func (r *CloneReconciler) findCloneSourcePod(pvc *corev1.PersistentVolumeClaim)
 
 	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
 		MatchLabels: map[string]string{
-			CloneUniqueID: string(pvc.GetUID()) + "-source-pod",
+			CloneUniqueID: getCloneSourcePodName(pvc),
```
mhenriks, line 270: not stored anywhere?

pkg/apiserver/apiserver.go
brybacki: right it is then used to validate, comment removed

pkg/controller/import-controller.go
brybacki: good point

pkg/controller/import-controller.go
brybacki: all functional tests pass, no usage fund so far,
do you have any information where it might be needed,

awels:
make sure you run the unit tests, seems like we have a bunch of failures there right now

brybacki:
yup, my bad,

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -57,11 +58,10 @@ func validateSourceURL(sourceURL string) string {
 
 func validateDataVolumeName(name string) []metav1.StatusCause {
 	var causes []metav1.StatusCause
-	// name of data volume cannot be more than 55 characters (not including '-scratch')
-	if len(name) > 55 {
+	if len(name) > kvalidation.DNS1123SubdomainMaxLength {
 		causes = append(causes, metav1.StatusCause{
 			Type:    metav1.CauseTypeFieldValueInvalid,
-			Message: fmt.Sprintf("Name of data volume cannot be more than 55 characters"),
+			Message: fmt.Sprintf("Name of data volume cannot be more than 253 characters"),
```
mhenriks, line 64: use constant here as well?

pkg/controller/clone-controller.go
```diff
@@ -264,10 +296,16 @@ func (r *CloneReconciler) findCloneSourcePod(pvc *corev1.PersistentVolumeClaim)
 	if !isCloneRequest {
 		return nil, nil
 	}
+	cloneSourcePodName, exists := pvc.Annotations[AnnCloneSourcePod]
+	if !exists {
+		// fallback to legacy name, to find any pod that still might be running after upgrade
+		cloneSourcePodName = createCloneSourcePodName(pvc)
+	}
 
+	// TODO: I think we might even get pod by name directly not using label selector
```
mhenriks, line 305: yeah no need for the label

mhenriks: I could have sworn this was something that was changed when this controller was refactored

mhenriks: Want to confirm that before this pr, the value of this annotation was equal to the pod name?

pkg/controller/clone-controller.go
```diff
@@ -182,6 +192,28 @@ func (r *CloneReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 	return reconcile.Result{}, nil
 }
 
+func (r *CloneReconciler) initCloneSourcePodName(pvc *corev1.PersistentVolumeClaim, log logr.Logger, sourcePod *corev1.Pod) error {
```
mhenriks, line 195: Something about passing the pod to this function seems a little wonky.  If the pod exists, why call this at all?

pkg/controller/upload-controller.go
```diff
@@ -514,11 +574,18 @@ func UploadPossibleForPVC(pvc *v1.PersistentVolumeClaim) error {
 
 // GetUploadServerURL returns the url the proxy should post to for a particular pvc
 func GetUploadServerURL(namespace, pvc, uploadPath string) string {
-	return fmt.Sprintf("https://%s.%s.svc%s", getUploadResourceName(pvc), namespace, uploadPath)
+	serviceName := createUploadServiceNameFromPvcName(pvc)
+	return fmt.Sprintf("https://%s.%s.svc%s", serviceName, namespace, uploadPath)
+}
+
+// createUploadServiceName returns the name given to upload service shortened if needed
+func createUploadServiceNameFromPvcName(pvc string) string {
```
mhenriks, line 582: This is something that may actually be nice to store in the PVC as well.  Then the upload proxy can use that value.  Understand if you don't want to tackle that right now.

pkg/apiserver/webhooks/datavolume-validate.go
brybacki: done

pkg/controller/import-controller.go

pkg/controller/clone-controller.go
brybacki: Interesting, Before this PR the pod name and CloneUniqueID was already the same:
commit: 17db2920b1582979625e9decf6aed572d436bec0
podName := getCloneSourcePodName(targetPvc)				
CloneUniqueID:          getCloneSourcePodName(targetPvc),

but before 17db2920b1582979625e9decf6aed572d436bec0  it was:
	podName := fmt.Sprintf("%s-%s-", common.ClonerSourcePodName, sourcePvcName)
	id := string(pvc.GetUID())
	CloneUniqueID:          id + "-source-pod",

brybacki: so it looks like some steps to change this were taken (but not finished)

pkg/controller/clone-controller.go
brybacki: I mean, if there is a legacy pod, then we want to store its name on PVC so rest of the logic is ok? but yes, it does not look good/obvious

pkg/controller/import-controller.go
brybacki: Scratch PVC name is created once, and it is here at this line. Then this name is available as a ClaimName on ScratchVolName = "cdi-scratch-vol" on pod, so we can always get the correct name. And we do not need to get scratch name in other places, so I wonder if we need to store it anywhere.

pkg/controller/clone-controller.go
brybacki: ok, analyzed it now, if we already have the pod, then we do not need name from annotation :), if we do not have a pod then we need to create or get name, so everything is cool, and this code is indeed not needed

pkg/controller/upload-controller.go
brybacki: I would create another task for this, small and focused ;)

awels:
looks like you have a linter error, run make test-lint to find it.

pkg/controller/clone-controller.go
brybacki: done

pkg/controller/clone-controller.go
brybacki: done

pkg/apiserver/webhooks/datavolume-validate.go
brybacki: good catch, how could I miss it.

pkg/controller/clone-controller.go
```diff
@@ -182,6 +192,21 @@ func (r *CloneReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 	return reconcile.Result{}, nil
 }
 
+func (r *CloneReconciler) initCloneSourcePodName(pvc *corev1.PersistentVolumeClaim, log logr.Logger) error {
+	currentPvcCopy := pvc.DeepCopyObject()
```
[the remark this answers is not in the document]
brybacki: done, and rebased

PR #1201, opened 2020-05-08
Handle service name generation correctly

brybacki:

Generates service name according to Service naming rules. So the DNS label standard as defined in RFC 1123: (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). Previously it was just an upload pod name. If/When upload pod name is ever changed to be longer then service can not be created.  

Fixes #

We might actually discuss if this is the right approach, or if better is to limit the upload pod name and just keep the old logic that ties service name to upload pod name 1:1.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -152,7 +152,8 @@ func (r *UploadReconciler) reconcilePVC(log logr.Logger, pvc *corev1.PersistentV
 		return reconcile.Result{}, err
 	}
 
-	if _, err = r.getOrCreateUploadService(pvc, resourceName); err != nil {
+	svcName := naming.GetServiceName(resourceName)
+	if _, err = r.getOrCreateUploadService(pvc, svcName); err != nil {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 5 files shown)


maya-r:
/retest
the failures were:
k8s-1.17-hpp: [Fail] [rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:component]Cloner Test Suite [It] [posneg:negative][test_id:3617]Should clone across nodes when multiple local volumes exist, 
os-3.11.0-crio: timeout hit

pkg/controller/upload-controller.go
```diff
@@ -246,7 +248,9 @@ func (r *UploadReconciler) getOrCreateUploadPod(pvc *v1.PersistentVolumeClaim, p
 			return nil, errors.Wrapf(err, "error getting upload pod %s/%s", pvc.Namespace, podName)
 		}
 
-		serverCert, serverKey, err := r.serverCertGenerator.MakeServerCert(pvc.Namespace, podName, uploadServerCertDuration)
+		// TODO: discuss, certificate needs to apply for a service! now implicitly it works, because service name is the same as podname,
```
awels, line 251: @mhenriks Any thoughts on this comment?

brybacki:
Closing this PR and adding changes to generating-service-names (PR https://github.com/kubevirt/containerized-data-importer/pull/1200), as this change does not make sense independent from #1200. Sorry for confusion.

PR #1205, opened 2020-05-15
Change gid check to use ls instead of stat

awels:

Change the gid check in functional tests to use ln -ln instead of stat, as stat might not be available in the container that we use for the verifier pod.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
"shell is bash" sounds like a harder requirement to satisfy than "stat is installed", but I assume you have a real case for wanting this change, and none of our tests seem to use non-bash shells yet, so
/lgtm

awels:
Well out verifier pods are actually using the importer container. So as long as the importer container has bash it will work.

PR #1206, opened 2020-05-15
alpha to beta snapshot API

mhenriks:

The alpha snapshot API is dead.  Moving to beta.

Includes hacked version of rook/ceph that supports beta snapshot api.

Fixes #

**Release note**:

```release-note
Cloning now uses beta snapshot API
```


mhenriks:
/hold

what to get rid of generated snapshot client.  It is only used in functional tests.

cluster-sync/rook-ceph/pool.yaml
```diff
@@ -1,15 +1,22 @@
+apiVersion: ceph.rook.io/v1
+kind: CephBlockPool
+metadata:
+  name: replicapool
+  namespace: rook-ceph
+spec:
+  failureDomain: host
+  replicated:
+    size: 1
```
awels, line 9: We want the size to be the number of nodes in the cluster, so this is the reason we had the CephBlockPool in the .sh file where we could substitute the value based on the KUBEVIRT_NUM_NODES environment variable.

cluster-sync/ephemeral_provider.sh
```diff
@@ -51,37 +51,20 @@ function configure_hpp() {
...
-  set +e
-  retry_counter=0
-  _kubectl get VolumeSnapshotClass
-  while [[ $? -ne "0" ]] && [[ $retry_counter -lt 60 ]]; do
-    retry_counter=$((retry_counter + 1))
-    echo "Sleep 5s, waiting for VolumeSnapshotClass CRD"
-    sleep 5
-    _kubectl get VolumeSnapshotClass
+  # wait for ceph
+  until _kubectl get cephblockpools -n rook-ceph replicapool -o jsonpath='{.status.phase}' | grep Ready; do
```
awels, line 61: Is this enough to verify the snapshot classes are available?

mhenriks: Unlike before, we actually create the VolumeSnapshotClass ourself.  So doesn't really make sense to wait for it.  Waiting for the replicapool to be ready assures PVCs will be able to be bound right after `cluster-sync` returns.  That was not the case before.

mhenriks:
> Looks pretty good, a couple of questions, and I am wondering how hard it would be to also update the ember_lvm lane.

I have not looked at ember yet.  Will attempt in another PR.

cluster-sync/ephemeral_provider.sh
awels: Okay sounds good

cluster-sync/rook-ceph/pool.yaml
awels: True, its not that important.

awels:
Okay ember is optional lane anyway. But it will for sure fail at this point.

PR #1210, opened 2020-05-21
Rename PVC.

tomob:
In order to make DVs and PVCs more user-friendly and to help users
confuse one with the other, we're changing the name of PVC created for
DV.

This PR changes generated PVC spec"s `Name` to `GenerateName` to let k8s to create the actual name. This will make PVC name different from its parent DV.

Fixes #

**Release note**:

```release-note
User's config/code could have made an assumption that the names of PVCs are equal to the names of the DVs. This is no longer true and this might cause some confusion after the update.
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -227,8 +227,8 @@ func (r *DatavolumeReconciler) Reconcile(req reconcile.Request) (reconcile.Resul
 
 	pvcExists := true
 	// Get the pvc with the name specified in DataVolume.spec
-	pvc := &corev1.PersistentVolumeClaim{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: datavolume.Namespace, Name: datavolume.Name}, pvc); err != nil {
+	pvc, err := r.findPVCForDV(datavolume.Namespace, datavolume.Name)
+	if err != nil {
 		// If the resource doesn't exist, we'll create it
 		if k8serrors.IsNotFound(err) {
 			pvcExists = false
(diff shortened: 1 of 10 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -278,6 +278,26 @@ func (r *DatavolumeReconciler) Reconcile(req reconcile.Request) (reconcile.Resul
...
+	pvcList := &corev1.PersistentVolumeClaimList{}
+	err = r.client.List(context.TODO(), pvcList, &client.ListOptions{Namespace: dataVolumeNamespace, LabelSelector: labelSelector})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(pvcList.Items) > 0 {
+		return &pvcList.Items[0], nil
+	}
+	// What if there's more than one? This should not be possible, but stuff happens?
```
awels, line 296: If we find more than one, that is a problem, and we should return an error.

pkg/controller/datavolume-controller.go
```diff
@@ -278,6 +278,26 @@ func (r *DatavolumeReconciler) Reconcile(req reconcile.Request) (reconcile.Resul
 	return r.reconcileDataVolumeStatus(datavolume, pvc)
 }
 
+func (r *DatavolumeReconciler) findPVCForDV(dataVolumeNamespace string, dataVolumeName string) (*corev1.PersistentVolumeClaim, error) {
+	labelSelector, err := labels.Parse(fmt.Sprintf("cdi-controller=%v", dataVolumeName))
```
awels, line 282: Note that bartosz's PR is removing this label, so we need to coordinate this.

tests/datavolume_test.go
```diff
@@ -548,12 +467,6 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 					Type:   cdiv1.DataVolumeReady,
 					Status: v1.ConditionTrue,
 				},
-				boundCondition: &cdiv1.DataVolumeCondition{
```
awels, line 551: Why are we removing these? Because we don't know the exact name of the PVC and its in the comparison string?

pkg/controller/datavolume-controller.go
brybacki: yes, removing, and if we really need this then we need to change the way label is created (directly storing datavolume name will not work for longer names datavolume-253 > label-63), 
maybe look at getPodFromPvc in datavolume-controller, or store uuid-name, or... I do not know

tomob: Oh no!

If the label is removed, I'd have to list all PVCs in the namespace and find the right one using Owner Reference. Or is there a faster way?

tests/datavolume_test.go
tomob: Because the name of the PVC can be only filled in runtime, the `boundCondition` section in each table entry was identical. I moved it to line 161 to remove repetition of the same block.

pkg/controller/datavolume-controller.go
brybacki: Well you can find a way to put there a label that is valid ;), I am not sure if it is a good pattern, but you can for example store string(pvcUID)+"-dv" (like it is done for label CloneUniqueID.

and if not then you should be able to limit pvc to the only cdi managed pvc's? I think we have a label for this.

tomob: I've given it some thought. I think the cost of calling a k8s api dominates over the cost of processing a list of items. We would make the call in any case, so I'd rather go for a simple code than relying another label.

Let me try that.

scuzhanglei:
user can't get the new pvc's name before controller create pvc. 
this may cause problem for some use cases. 
for example, user directly use pvc created by dataVolume instead of dataVolume self.

awels:
@scuzhanglei We will provide a mechanism to locate the PVC associated with the DV. At the very minimum the PVC will have an ownerRef of the DV.

scuzhanglei:
> @scuzhanglei We will provide a mechanism to locate the PVC associated with the DV. At the very minimum the PVC will have an ownerRef of the DV.

@awels  thanks, but in this way. other application want to use cdi have to include DataVolume in its CR definition like kubevirt VirtualMachine.  maybe we can add another field to allow user customize pvc name

awels:
Note that if you want to use CDI and not use the benefits of the DataVolume, you can simply create a PVC with the right annotations and it should all work. That way you have control over the PVC name.

awels:
looks like that one test is failing due to maybe an assumption on the pvc name?

tomob:
> looks like that one test is failing due to maybe an assumption on the pvc name?

You're right!

I run all the cloning tests after the rebase and they passed, but I must have messed something with the cluster setup (maybe sync'd with a wrong branch?)

awels:
I am going to put a hold on this, I think we need to share this with the community and see if people have concerns before proceeding.
/hold

PR #1211, opened 2020-05-22
Enable the use of podman

awels:
Enable the use of podman to start the kubevirtci cluster, and make cluster-sync

Currently kubevirtci uses docker everywhere, and gocli is also tied to docker. For newer fedora (31+) docker no longer works out of the box (it doesn't support cgroupsv2 yet) and we don't know when docker will work. In order for our developers to use a newer fedora we have to support podman which does support cgroupsv2.

Luckily for us, some work has already happened to get a drop in replacement for gocli called [pack8s](https://github.com/kubevirt/kubevirtci/tree/master/pack8s) that can be enabled by setting the KUBEVIRTCI_RUNTIME to podman
```bash
$ export KUBEVIRTCI_RUNTIME=podman
```

This PR enables the rest of our tool chain (bazel, make cluster-sync) to use podman as well. Most of the changes are minor and involve fixing how volumes are mounted (permissions) in podman vs docker.

To enabled podman follow the [instructions](https://github.com/kubevirt/kubevirtci/tree/master/pack8s) for pack8s after disabling and removing docker. Also install the podman-docker package to create an alias between podman and docker.

**known issue**
It only works with a single node, the cluster doesn't come up with multipe nodes

Fixes #

Note when testing this, we don't have a new bazel-builder yet, so you have to manually do a buildah bud in hack/build/docker/builder to build the new image, and update hack/build/config.sh to use the local image instead of the BUILDER_IMAGE in there now. Once we have this merged, we have to do a follow up PR to update the hash. If you don't then make generate will mess up the owner of your entire containerized-data-importer directory. With docker it is required to fix the owner from root to the user, but podman doesn't touch the owner.

**Release note**:

```release-note
Devel: podman now works with kubevirtci cluster-up and cluster-sync
```

### DIFF
--- a/hack/build/bazel-docker.sh
+++ b/hack/build/bazel-docker.sh
@@ -33,16 +33,28 @@ fi
 
 # Create the persistent docker volume
 if [ -z "$(docker volume list | grep ${BUILDER_VOLUME})" ]; then
-    docker volume create --name ${BUILDER_VOLUME}
+    if [ "$KUBEVIRTCI_RUNTIME" = "podman" ]; then
+        docker volume create ${BUILDER_VOLUME}
+    else
+        docker volume create --name ${BUILDER_VOLUME}
+    fi
 fi
 
 # Make sure that the output directory exists
 echo "Making sure output directory exists..."
-docker run -v "${BUILDER_VOLUME}:/root:rw,z" --security-opt label:disable --rm --entrypoint "/entrypoint-bazel.sh" ${BUILDER_IMAGE} mkdir -p /root/go/src/kubevirt.io/containerized-data-importer/_out
+if [ "$KUBEVIRTCI_RUNTIME" = "podman" ]; then
+    docker run -v "${BUILDER_VOLUME}:/root:rw,z" --security-opt label=disable --rm --entrypoint "/entrypoint-bazel.sh" ${BUILDER_IMAGE} mkdir -p /root/go/src/kubevirt.io/containerized-data-importer/_out
+else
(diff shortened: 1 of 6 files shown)


hack/build/bazel-docker.sh
```diff
@@ -93,16 +109,24 @@ _rsync \
...
 
 # Ensure that a bazel server is running
 
 if [ -z "$(docker ps --format '{{.Names}}' | grep ${BAZEL_BUILDER_SERVER})" ]; then
-    docker run --network host -d ${volumes} --security-opt label:disable --name ${BAZEL_BUILDER_SERVER} -w "/root/go/src/kubevirt.io/containerized-data-importer" --rm ${BUILDER_IMAGE} hack/build/bazel-server.sh
+    if [ "$KUBEVIRTCI_RUNTIME" = "podman" ]; then
+        docker run --network host -d ${volumes} --security-opt label=disable --name ${BAZEL_BUILDER_SERVER} -w "/root/go/src/kubevirt.io/containerized-data-importer" --rm ${BUILDER_IMAGE} hack/build/bazel-server.sh
+    else
+        docker run --network host -d ${volumes} --security-opt label:disable --name ${BAZEL_BUILDER_SERVER} -w "/root/go/src/kubevirt.io/containerized-data-importer" --rm ${BUILDER_IMAGE} hack/build/bazel-server.sh
+    fi
```
maya-r, line 129: Are these two commands supposed to be different?

awels: They are different, look at the --security-opt label(:/=)disable

fromanirh:
> **known issue**
> It only works with a single node, the cluster doesn't come up with multipe nodes

Original author of `pack8s` here. I'm very happy if the tool helps you folks. pack8s wasn't indeed thoroughly tested for multi-node scenarios (lack of capacity back in time), but I wonder if this kubevirtci issue is relevant in this case: https://github.com/kubevirt/kubevirtci/issues/264

PR #1216, opened 2020-05-28
Give source clone pvc a gid so it gets copied properly.

awels:

Our cloning source pvc file didn't have a gid, so it wouldn't get copied. This PR gives the source file a GID if it applies, so it gets copied and the check properly identifies the gid.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/pod.go
+++ b/tests/utils/pod.go
@@ -14,6 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"kubevirt.io/containerized-data-importer/pkg/common"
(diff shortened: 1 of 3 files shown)


tests/utils/pod.go
```diff
@@ -67,6 +67,7 @@ func DeletePod(clientSet *kubernetes.Clientset, pod *k8sv1.Pod, namespace string
 
 // NewPodWithPVC creates a new pod that mounts the given PVC
 func NewPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolumeClaim) *k8sv1.Pod {
+	fsGroup := int64(107)
```
brybacki, line 70: what exactly is 107?

arnongilboa:
Looks good to me, except the const mentioned by @brybacki .

tests/utils/pod.go
awels: Good point, should have used the constant we have defined for it.

PR #1225, opened 2020-06-08
Update regex that matches version to accept +hash in version string

awels:

The cdi config test checks the version of kubernetes, in newer versions the version can now have +<hash> in the version string. The existing regex failed to match, this PR fixes the regex.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
I imagine the + character doesn't appear in older versions.
Maybe the pattern \+*\S* does the trick for new and old at the same time?

awels:
That would do the trick, since + is part of \S, however I want to explicitly call out the + that is why I did \+?

PR #1228, opened 2020-06-09
Call finalize in close for imageio source

machacekondra:
By calling Finalize method in Close method of the imageio source
we will make sure we finalise the import immediately, rather then
rely on implicit finalise of imageio, which is based on timeout.

Fixes #

**Release note**:

```release-note
Explicitly finalise image transfer connection instead of let it timeout for imageio datasource.
```

### DIFF
--- a/pkg/importer/imageio-datasource.go
+++ b/pkg/importer/imageio-datasource.go
@@ -47,12 +47,16 @@ type ImageioDataSource struct {
 	url *url.URL
 	// the content length reported by ovirt-imageio.
 	contentLength uint64
+	// imageTransfer is the tranfer object handling the tranfer of oVirt disk
+	imageTransfer *ovirtsdk4.ImageTransfer
+	// connection is connection to the oVirt system
+	connection ConnectionInterface
 }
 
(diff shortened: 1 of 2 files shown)


machacekondra:
cc @pkliczewski

awels:
Can you add a release note? Or NONE if there is no release note, but I think it is worth mentioned an enhancement to finalize the imageio connection instead of letting it time out.

machacekondra:
> Can you add a release note? Or NONE if there is no release note, but I think it is worth mentioned an enhancement to finalize the imageio connection instead of letting it time out.

I've updated the release note.

awels:
looks like you have a linter error, make test-lint will show you

maya-r:
Error was a fedora mirror not resolving
Error downloading [http://download.fedoraproject.org/pub/fedora/linux/releases/31/Everything/x86_64/os/Packages/l/lvm2-2.03.05-2.fc31.x86_64.rpm] to /root/.cache/bazel/_bazel_root/e03bf9038fc5089c0dbc8615812d9838/external/lvm2/file/downloaded: Unknown host: dl.iad2.fedoraproject.org

PR #1237, opened 2020-06-16
Strip status from objects in operator reconcile loop

awels:
This stops the operator controller from trying to modify the CRDs in each reconcile loop.

The operator tries to update the CRDs in each reconcile loop because the desired object status is empty and k8s will have populated some status. Then the operator removes the status because it expects there to be none, which in turn will cause k8s to set the status again. On each reconcile loop this ping pong happened. This PR changes the code to strip the status value from the current object so the comparison will cause not fail due to a minor status change.

Also added default values for deployments that would be filled in by k8s as well as defaults for the webhooks

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/controller/BUILD.bazel
+++ b/pkg/operator/controller/BUILD.bazel
@@ -72,10 +72,12 @@ go_test(
         "certrotation_test.go",
         "controller_suite_test.go",
         "controller_test.go",
+        "util_test.go",
     ],
     embed = [":go_default_library"],
     deps = [
         "//pkg/apis/core/v1beta1:go_default_library",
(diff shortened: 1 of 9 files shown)


mhenriks:
Although this may address the issue, I think it would be better to refactor the diff/merge code to not consider status.

awels:
When you say that, we have runtime.Object, and we need to cast it to something in order to check. Are we worried here about just CRDs and not the other types, because I don't think there is a generic way to detect if a type has a status field that we can ignore.

awels:
@mhenriks This now uses the json conversion to strip out the status part of the object.

PR #1242, opened 2020-06-17
WaitForFirstConsumer handling for DataVolumes

brybacki:

When the underlying status of a PVC that was created by the DataVolume controller is Pending/WaitForFirstConsumer then CDI will not schedule worker pods (import/clone/upload) associated with that PVC. This would allow the workload itself (ie. kubevirt) to handle the initial scheduling which would cause the PVC to change to Bound state.

Without this change PVC is scheduled by the CDI worker pods, which might be scheduled on different nodes then later the VMI. 

Fixes #

Kubevirt needs to be changed first. 

**Release note**:

```release-note
Added special handling for PVCs which have StorageClass with `volumeBindingMode` set to `WaitForFirstConsumer`.
```


brybacki:
/retest
quay had some issues

awels:
make sure you remove the
```go
log = logf.Log.WithName("feature-gates-test")
```

brybacki:
> make sure you remove the
> 
> ```go
> log = logf.Log.WithName("feature-gates-test")
> ```

done

pkg/feature-gates/feature-gates.go
```diff
@@ -20,9 +20,8 @@ type FeatureGates struct {
 }
```
awels, line 20: I probably should have suggested this in #1283 But it might make sense to make an interface with a single function in it:

```go
type FeatureGates interface {
  HonorWaitForFirstConsumerEnabled() (bool, error) 
}
```
And implement the rest as you have, the struct needs a slightly different name. The point of this is so that when we are writing unit tests, we can implement this interface in a test structure with a particular value and we don't need the runtime library client to make it work in the tests.

Then adding new feature gates will be a matter of adding functions on the interface.

pkg/controller/upload-controller.go
```diff
@@ -114,12 +116,17 @@ func (r *UploadReconciler) Reconcile(req reconcile.Request) (reconcile.Result, e
...
 		return reconcile.Result{}, errors.New("PVC has both clone and upload annotations")
 	}
-
+	honorWaitForFirstConsumer, err := r.featureGates.HonorWaitForFirstConsumerEnabled()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 	// force cleanup if PVC pending delete and pod running or the upload/clone annotation was removed
-	if (!isUpload && !isCloneTarget) || podSucceededFromPVC(pvc) || pvc.DeletionTimestamp != nil {
+	if !r.shouldReconcile(isUpload, isCloneTarget, honorWaitForFirstConsumer, pvc, log) || podSucceededFromPVC(pvc) || pvc.DeletionTimestamp != nil {
```
awels, line 124: Same suggestion here about passing in the featuregates instead of looking it up and passing the value.

pkg/controller/upload-controller.go
```diff
@@ -524,6 +536,7 @@ func NewUploadController(mgr manager.Manager, log logr.Logger, uploadImage, pull
 		recorder:            mgr.GetEventRecorderFor("upload-controller"),
 		serverCertGenerator: serverCertGenerator,
 		clientCAFetcher:     clientCAFetcher,
+		featureGates:        featuregates.NewFeatureGates(client),
```
awels, line 539: Same nit

pkg/feature-gates/feature-gates_test.go
```diff
@@ -65,7 +60,19 @@ func createFeatureGates(objects ...runtime.Object) *FeatureGates {
 	objs = append(objs, objects...)
 
 	// Append empty CDIConfig object that normally is created by the reconcile loop
-	cdiConfig := controller.MakeEmptyCDIConfigSpec(common.ConfigName)
+	cdiConfig := &cdiv1.CDIConfig{
```
awels, line 63: Am I missing something here, but is this not identical to what MakeEmptyCDICOnfigSpec returns?

tests/cloner_test.go
```diff
@@ -299,8 +303,10 @@ var _ = Describe("Validate creating multiple clones of same source Data Volume",
 		dvs := []*cdiv1.DataVolume{targetDv1, targetDv2, targetDv3}
 
 		for _, dv := range dvs {
-			_, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
 			Expect(err).ToNot(HaveOccurred())
+			f.ForceBindPvcIfDvIsWaitForFirstConsumer(dataVolume)
```
awels, line 308: Looks like this assumes that DV name = PVC name? If so can you add a comment that this needs to get fixed when we merge the PR that makes this assumption invalid? This happens in numerous places, please put a comment on all of them so we can easily identify them.

tests/datavolume_test.go
```diff
@@ -813,6 +820,142 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 		})
 	})
 
+	Describe("[rfe_id:4223][crit:high] DataVolume - WitForFirstConsumer", func() {
```
awels, line 823:
```
typo DataVolume - WitForFirstConsumer
                   ^
```

tests/datavolume_test.go
```diff
@@ -953,13 +1099,14 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 	Describe("Registry import with missing configmap", func() {
 		const cmName = "cert-registry-cm"
 
-		It("Import POD should remain pending until CM exists", func() {
-			var pvc *corev1.PersistentVolumeClaim
+		It("aaqq Import POD should remain pending until CM exists", func() {
```
awels, line 1102: aaqq??

pkg/controller/datavolume-controller.go
```diff
@@ -162,12 +164,15 @@ func pvcIsPopulated(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) boo
...
+	client := mgr.GetClient()
+	featureGates := featuregates.NewFeatureGates(client)
 	reconciler := &DatavolumeReconciler{
-		client:       mgr.GetClient(),
+		client:       client,
 		scheme:       mgr.GetScheme(),
 		extClientSet: extClientSet,
 		log:          log.WithName("datavolume-controller"),
 		recorder:     mgr.GetEventRecorderFor("datavolume-controller"),
+		featureGates: featureGates,
```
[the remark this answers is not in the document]
brybacki: I was sure I have already did that, maybe in different file.
Updated

pkg/controller/datavolume-controller.go
```diff
@@ -300,6 +305,40 @@ func (r *DatavolumeReconciler) sourceInUse(dv *cdiv1.DataVolume) (bool, error) {
 	return len(pods) > 0, nil
 }
 
+func (r *DatavolumeReconciler) getStorageClassBindingMode(dataVolume *cdiv1.DataVolume) (storagev1.VolumeBindingMode, error) {
+	var targetPvcStorageClass *storagev1.StorageClass
+	targetPvcStorageClassName := dataVolume.Spec.PVC.StorageClassName
+
+	// Handle unspecified storage class name, fallback to default storage class
+	if targetPvcStorageClassName == nil {
```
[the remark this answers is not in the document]
brybacki: sounds good, I'll try that, the only difference would be that the old code in `getSnapshotClassForSmartClone` didn't really try to fetch targetStorageClass if name was set, so if there was any storage class name in dataVolume.Spec.PVC.StorageClassName (even dummy) the check just went forward.

pkg/feature-gates/feature-gates.go
brybacki: very good idea, I would like to create a new PR for that, no here.

pkg/controller/upload-controller.go
brybacki: funny, here it is already good ;)

pkg/feature-gates/feature-gates_test.go
brybacki: yes, but because MakeEmptyCDICOnfigSpec util is in controller, it will introduce cyclic import of packages feature-gates and controller, at least in tests,

tests/cloner_test.go
brybacki: very good catch.

fortunately this is enclosed inside ForceBindPvcIfDvIsWaitForFirstConsumer so only has to be fixed once, I have added comment inside the function

the other function -  ForceBindIfWaitForFirstConsumer just uses pvc, so the code to get pvc was already in the test.

tests/datavolume_test.go
brybacki: It was a temporary tag to focus a few unrelated but failing tests ;) I do it like this (yes I know about `F`), I do not remember why but I had some reason: 
```
make TEST_ARGS="--test-args=-ginkgo.focus=aaqq -ginkgo.regexScansFilePath=false" test-functional\n 
```

pkg/controller/datavolume-controller.go
```diff
@@ -300,6 +304,24 @@ func (r *DatavolumeReconciler) sourceInUse(dv *cdiv1.DataVolume) (bool, error) {
...
 }
 
+func (r *DatavolumeReconciler) getStorageClassBindingMode(dataVolume *cdiv1.DataVolume) (storagev1.VolumeBindingMode, error) {
+	var targetPvcStorageClass *storagev1.StorageClass
+	targetPvcStorageClassName := dataVolume.Spec.PVC.StorageClassName
+
+	// Handle unspecified storage class name, fallback to default storage class
+	targetPvcStorageClass, err := r.getStorageClassByName(targetPvcStorageClassName)
+	if err != nil {
+		return "", err
```
mhenriks, line 314: would returning `*storagev1.VolumeBindingMode`be nicer?

pkg/controller/datavolume-controller.go
```diff
@@ -300,6 +304,24 @@ func (r *DatavolumeReconciler) sourceInUse(dv *cdiv1.DataVolume) (bool, error) {
 	return len(pods) > 0, nil
 }
 
+func (r *DatavolumeReconciler) getStorageClassBindingMode(dataVolume *cdiv1.DataVolume) (storagev1.VolumeBindingMode, error) {
+	var targetPvcStorageClass *storagev1.StorageClass
```
mhenriks, line 308: can be removed

pkg/controller/datavolume-controller.go
brybacki: yes, 
no value == nil

pkg/feature-gates/feature-gates.go
brybacki: actually it might be easy to do and useful right now, in progress...

brybacki:
All the comments addressed, tests are green ;)
@awels @mhenriks what do you think?

pkg/controller/import-controller.go
```diff
@@ -135,8 +140,10 @@ func addImportControllerWatches(mgr manager.Manager, importController controller
 	return nil
 }
 
-func shouldReconcilePVC(pvc *corev1.PersistentVolumeClaim, log logr.Logger) bool {
-	return !isPVCComplete(pvc) && (checkPVC(pvc, AnnEndpoint, log) || checkPVC(pvc, AnnSource, log))
+func shouldReconcilePVC(pvc *corev1.PersistentVolumeClaim, honorWaitForFirstConsumer bool, log logr.Logger) bool {
```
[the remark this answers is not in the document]

pkg/feature-gates/feature-gates.go
brybacki: done

PR #1244, opened 2020-06-18
GetAvailableSpace(block) now returns error

awels:
Modified function that gets the size of a block device/available to return error as well as -1, so we
can distinguish the path not existing from the binary not existing in case the container doesn't have
the required binaries.

GetAvailableSpace(block) did not return an error, it would return -1 in case of a failure, but that left us unable to distinguish between the file/device not existing vs other failures. We now properly return (and report) the error in case of failures.

Fixes #

**Release note**:

```release-note
BugFix: GetAvailableSpace now properly returns error on failures.
```


pkg/importer/data-processor_test.go
```diff
@@ -389,8 +389,8 @@ var _ = Describe("Resize", func() {
 	})
 
 	It("Should return same value as replaced function", func() {
-		replaceAvailableSpaceBlockFunc(func(dataDir string) int64 {
-			return int64(100000)
+		replaceAvailableSpaceBlockFunc(func(dataDir string) (int64, error) {
+			return int64(100000), nil
 		}, func() {
```
aglitke, line 394: We should add a test where the AvailableSpaceBlockFunc returns an error.

awels: So in this particular case we ignore errors from the function and log them. Then continue on base on the result. So adding the test would not add much value.

pkg/importer/data-processor_test.go
```diff
@@ -389,14 +389,26 @@ var _ = Describe("Resize", func() {
...
 			mdp := &MockDataProvider{}
 			dp := NewDataProcessor(mdp, "dest", "dataDir", "scratchDataDir", "")
 			Expect(int64(100000)).To(Equal(dp.calculateTargetSize()))
 		})
 	})
+
+	It("Should fail is calculate size returns failure", func() {
```
aglitke, line 401: Should fail if

PR #1245, opened 2020-06-18
Report better qemu failure messages.

awels:

We were reporting the error generated by the command executions, which normally is something like 'exit status 1' or something similar, which is not really helpful in identifying the problem. Instead this PR will now report the stderr which should have better information on what exactly failed and why.

Fixes #

**Release note**:

```release-note
Enhancement: qemu-img failures are now reported better.
```


pkg/image/qemu.go
```diff
@@ -155,7 +155,7 @@ func (o *qemuOperations) Info(url *url.URL) (*ImgInfo, error) {
 		output, err = qemuExecFunction(qemuInfoLimits, nil, "qemu-img", "info", "--output=json", url.String())
 	}
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error getting info on image %s", url.String())
+		return nil, errors.Errorf("%s, %s", output, err.Error())
```
aglitke, line 158: We now lost the URL in the error but I suppose it's okay because that information is in the spec of the DataVolume.

PR #1247, opened 2020-06-19
Add new dv phase WaitForFirstConsumer

brybacki:

Adds a new phase: WaitForFirstConsumer to the Datavolume API. This phase means that the underlying storage binding mode is WaitForFirstConsumer so the PVC will not be bound until first consumer pod is created. Any CDI action will be paused until the PVC is bound by the consumer pod. 

Fixes #

This is prerequisite for https://github.com/kubevirt/containerized-data-importer/pull/1242

**Release note**:

```release-note
A new DataVolumePhase was added: WaitForFirstConsumer. This phase means that the underlying storage binding mode is WaitForFirstConsumer so the PVC created by CDI will not be bound until first consumer pod is created.
```

### DIFF
--- a/pkg/apis/core/v1alpha1/types.go
+++ b/pkg/apis/core/v1alpha1/types.go
@@ -191,6 +191,9 @@ const (
 	// UploadReady represents a data volume with a current phase of UploadReady
 	UploadReady DataVolumePhase = "UploadReady"
 
+	// WaitForFirstConsumer represents a data volume with a current phase of WaitForFirstConsumer
+	WaitForFirstConsumer DataVolumePhase = "WaitForFirstConsumer"
+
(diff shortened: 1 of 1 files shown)


aglitke:
Do we need to regenerate the api also?

awels:
yes I think a make generate is needed so the api is regenerated.

brybacki:
make generate does nothing here, since this type is a string and the change only adds a new value for it

awels:
The /test doesn't work for travis you have to retrigger from travis UI. I just did it for you because it seems the reporting didn't get back to github.

brybacki:
yes, this is strange, second time I see no report from CI (also had this on previous PR, SUCCESS but no report)

awels:
Well travis has to call an API in GH, and if that fails for whatever reason then it won't report.

PR #1253, opened 2020-06-23
Return stderr from verifier pods

awels:

It was hard to determine why verifier pods failed because they weren't printing stderr. This fixes that and should now print stderr on failure. This should allow us to debug the failures of the verifier pods more easily with more information.

Fixes #

**Release note**:

```release-note
NONE
```


brybacki:
that is helpful
/lgtm

PR #1255, opened 2020-06-23
Block clone across namespace, delete populator/verifier pod

awels:

The populator and verifier pods where still there, which caused multi attach issues in the ceph lane. This PR makes sure the populator and verifier pod are deleted before cloning.

Fixes #

**Release note**:

```release-note
NONE
```


brybacki:
So the source PVC cannot be used by other pod while the clone is in progress?
I wonder if that is only a test problem, or maybe we should make sure in the clone controller logic, that the source pvc is not used by other pods before starting clone?

brybacki:
other than my previous comment
/lgtm

PR #1065, opened 2019-12-26
Fix `GetAvailableSpaceBlock` always return -1 in LXC

clarklee92:
In a normal Linux container, /sys/class/block/ is inherited from the host, and the result presented by lsblk is the metadata of the host, which does not effectively capture the size of the block device transmitted to the container

Signed-off-by: clarklee92 <clarklee1992@hotmail.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1064

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #1258, opened 2020-06-24
Use blockdev instead of lsblk for getting block device size

kincl:

lsblk does not work with raw block devices that are backed by
dm-multipath because the container does not populate the /dev/mapper
symlinks

blockdev comes from util-linux as well and already exists in the
cdi-cloner image so it should be a drop-in replacement

**Release note**:
```release-note
Use blockdev instead of lsblk for getting raw block device size
```


maya-r:
I'm feeling a strong sense of deja-vu :-) (#1065)
/lgtm

kincl:
The failing test doesn't appear to be related to my change but let me know if I need to look into it more

awels:
No you are good, we are having some issues with some flaky tests that we are working on resolving, we will rerun the lane to get it to pass soon.

PR #1261, opened 2020-06-25
[flaky test] verifier pods hanging around keeps cloning from working

mhenriks:

**What this PR does / why we need it**:  Address flakes like this:

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1260/pull-containerized-data-importer-e2e-k8s-1.17-ceph/1275904563907399680

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/pvc.go
+++ b/tests/framework/pvc.go
@@ -204,7 +204,7 @@ func (f *Framework) VerifyPermissions(namespace *k8sv1.Namespace, pvc *k8sv1.Per
 }
 
 // GetDiskGroup returns the group of a disk image.
-func (f *Framework) GetDiskGroup(namespace *k8sv1.Namespace, pvc *k8sv1.PersistentVolumeClaim) (string, error) {
+func (f *Framework) GetDiskGroup(namespace *k8sv1.Namespace, pvc *k8sv1.PersistentVolumeClaim, deletePod bool) (string, error) {
 	var executorPod *k8sv1.Pod
 	var err error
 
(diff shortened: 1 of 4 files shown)


tests/framework/pvc.go
```diff
@@ -225,6 +225,11 @@ func (f *Framework) GetDiskGroup(namespace *k8sv1.Namespace, pvc *k8sv1.Persiste
 	output, err = f.ExecShellInPod(executorPod.Name, namespace.Name, cmd)
 	fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: gid of disk.img: %s\n", string(output))
 
+	if deletePod {
+		err = f.K8sClient.CoreV1().Pods(namespace.Name).Delete(executorPod.Name, nil)
```
awels, line 229: Only question here, do we want to wait to ensure the pod is gone before continuing? I have seen some instances where we assumed after a delete, the pod was gone and then checked for the pod being gone, and it was still in terminating or something and that failed the test.

mhenriks: Yeah that can take awhile though.  And it appeared to me that there is some effort underway to make things (reusing the verifier pods).  Waiting for the pod to be gone may make debugging failed tests easier.  But I don't think it won't keep tests from failing.  And may make things slower.  Your call.

awels: Yeah, I am trying to not delete the pod unless its needed, because it may take a while, but in this case we have explicitly said delete this pod, and the rest of the test might assume its actually gone, while it might still be terminating and cause issues. For this particular PR I don't think that any thing assumes it though.

PR #1263, opened 2020-06-26
Wait for PVCs to be unused before initiating clone operations

mhenriks:

There is no way to 100% safely guard against data corruption when cloning in k8s.  But this improves our situation a bit by at least verifying that source PVCs are not being used in an unsafe way before the clone operation is started.  For smart clone (using snapshots) that means that the PVC cannot be mounted in any Pod.  Things are less strict with dumb cloning (copying over the network).  In that case, read only mounts are okay.

Fixes #

https://bugzilla.redhat.com/show_bug.cgi?id=1845901

**Release note**:

```release-note
Wait for PVCs to be unused before initiating clone operations
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -9,8 +9,6 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/sets"
-
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -18,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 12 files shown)


mhenriks:
There seem to be incompatibilities between the current k8s 1.17 kubevirtci provider and ember.  So I say we ignore ember results for now.

FWIW: I dug into the issue a bit and it appears that `iscsiadm --login` is hanging.

brybacki:
source and target pvc are now checked for usage - that is good, what about import? does it also need to be checked or not?

mhenriks:
> source and target pvc are now checked for usage - that is good, what about import? does it also need to be checked or not?

It should, but I think should be done separately as this is a cloning bug.

pkg/controller/clone-controller.go
```diff
@@ -193,24 +197,46 @@ func (r *CloneReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 	return reconcile.Result{}, nil
 }
 
-func (r *CloneReconciler) reconcileSourcePod(sourcePod *corev1.Pod, pvc *corev1.PersistentVolumeClaim, log logr.Logger) error {
+func (r *CloneReconciler) reconcileSourcePod(sourcePod *corev1.Pod, targetPvc *corev1.PersistentVolumeClaim, log logr.Logger) (bool, error) {
 	if sourcePod == nil {
-		if err := r.validateSourceAndTarget(pvc); err != nil {
-			return err
+		sourcePvc, err := r.getCloneRequestSourcePVC(targetPvc)
```
brybacki, line 202: very good cleanup, I like the explicit naming for source/target pvc's

awels:
I am fine with ignore ember results right now. I am investigating what the issue is, as I have a different PR that should fix most ember issues, but with the change that is also failing.

pkg/controller/datavolume-controller.go
```diff
@@ -84,6 +85,8 @@ const (
 	SnapshotForSmartCloneCreated = "SnapshotForSmartCloneCreated"
 	// SmartClonePVCInProgress provides a const to indicate snapshot creation for smart-clone is in progress
 	SmartClonePVCInProgress = "SmartClonePVCInProgress"
+	// SmartCloneSourceInUse proovides a const to indicate a smart clone is being delayed becasuse the source is in use
```
awels, line 88: Typo in proovides

PR #1272, opened 2020-06-30
New test case "delete file during import"

brybacki:

Adds a new test that checks if the import can continue after failure is resolved. 

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -845,13 +845,10 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 				"rm -f /tmp/shared/images/"+testImageName)
 			Expect(err).To(BeNil())
 
-			Eventually(func() bool {
-				_, err := f.K8sClient.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Get(dataVolume.Name, metav1.GetOptions{})
-				if k8serrors.IsNotFound(err) {
-					return true
-				}
-				return false
-			}, timeout, pollingInterval).Should(BeTrue())
+			By("Verifying pvc was deleted")
+			deleted, err := utils.WaitPVCDeleted(f.K8sClient, dataVolume.Name, dataVolume.Namespace, timeout)
+			Expect(deleted).To(BeTrue())
+			Expect(err).ToNot(HaveOccurred())
 		})
 
 		It("[test_id:1990] CDI Data Volume - file is removed from http server while import is in progress", func() {
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -812,13 +812,14 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 	})
 
 	Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:component][test] CDI Import from HTTP/S3", func() {
-
+		const (
+			originalImageName = "cirros-qcow2.img"
+			testImageName     = "cirros-qcow2-1990.img"
+		)
 		var (
 			dataVolume              *cdiv1.DataVolume
 			err                     error
-			originalImageName       = "cirros-qcow2.img"
-			testImageName           = "cirros-qcow2-1990.img"
-			tinyCoreIsoRateLimitURL = fmt.Sprintf("http://cdi-file-host.%s:82/cirros-qcow2-1990.img", f.CdiInstallNs)
+			tinyCoreIsoRateLimitURL = "http://cdi-file-host." + f.CdiInstallNs + ":82/cirros-qcow2-1990.img"
 		)
 
 		BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


maya-r:
testing new lanes :-)
/test pull-containerized-data-importer-e2e-k8s-1.18-upg
/test pull-containerized-data-importer-e2e-k8s-1.18-hpp
/test pull-containerized-data-importer-e2e-k8s-1.18-nfs

maya-r:
testing for stability...
/test pull-containerized-data-importer-e2e-k8s-1.18-upg
/test pull-containerized-data-importer-e2e-k8s-1.18-hpp
/test pull-containerized-data-importer-e2e-k8s-1.18-nfs

tests/datavolume_test.go
```diff
@@ -811,6 +811,108 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 		})
 	})
 
+	Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:component][test] CDI Import from HTTP/S3", func() {
+
+		var (
+			dataVolume              *cdiv1.DataVolume
+			err                     error
+			originalImageName       = "cirros-qcow2.img"
```
awels, line 819: Are originalImageName and testImageName variables or constants? they look like constants to me.

tests/datavolume_test.go
```diff
@@ -811,6 +811,108 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+
+			By("Cleanup the file")
+			fileHostPod, err := utils.FindPodByPrefix(f.K8sClient, f.CdiInstallNs, utils.FileHostName, "name="+utils.FileHostName)
+			_, _, err = f.ExecCommandInContainerWithFullOutput(fileHostPod.Namespace, fileHostPod.Name, "http",
+				"/bin/sh",
+				"-c",
+				"rm -f /tmp/shared/images/"+testImageName)
+			Expect(err).To(BeNil())
+
+			Eventually(func() bool {
```
awels, line 847: This is fine, and works great, but Tomasz has a PR that adds a WaitForPVCDeleted function to the utils that does exactly this. Depending on which one gets merged first, you might want to use that function.

tests/datavolume_test.go
```diff
@@ -811,6 +811,108 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			dvName := "import-file-removed"
+			By(fmt.Sprintf("Creating new datavolume %s", dvName))
+			dv := utils.NewDataVolumeWithHTTPImport(dvName, "500Mi", tinyCoreIsoRateLimitURL)
+			dataVolume, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
+			Expect(err).ToNot(HaveOccurred())
+
+			phase := cdiv1.ImportInProgress
+			By(fmt.Sprintf("Waiting for datavolume to match phase %s", string(phase)))
+			err = utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, phase, dataVolume.Name)
+			if err != nil {
```
awels, line 866: WaitForDataVolumePhase returns a message like: "DataVolume %s not in phase %s within %v", dataVolumeName, phase, timeout as part of the err. So you can just do Expect(err).ToNot(HaveOccurred()) and it will print what you are doing here.

tests/datavolume_test.go
```diff
@@ -811,6 +811,108 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+				"cp /tmp/shared/images/"+originalImageName+" /tmp/shared/images/"+testImageName)
+			Expect(err).To(BeNil())
+		})
+
+		AfterEach(func() {
+			By("Delete DV")
+			err = utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Cleanup the file")
```
awels, line 839: Do we want to check if the original file is there, and if not restore it. It is possible the test failed after you removed the original file, and before you restored it. Then this after each will remove the backup as well, and we will have lost the file completely.

tests/datavolume_test.go
brybacki: const

tests/datavolume_test.go
brybacki: right, I'll check it

tests/datavolume_test.go
brybacki: nice hint, removed

tests/datavolume_test.go
brybacki: The thing is, I am never removing the original, I am doing test on new `testImageName` file. So cleanup is simpler, and in case of any errors we do not loose original we just leave a little bit of garbage as `testImageName` (but this is handled in afterEach, so should not happen)

tests/datavolume_test.go
brybacki: the util is in PR #1257

awels: https://github.com/kubevirt/containerized-data-importer/blob/master/tests/utils/pvc.go#L224 Tomasz's PR has to deal with the CDI controller creating a new PVC and has to verify the new one is different from the original one. But I think you can just use the util.WaitPVCDeleted which is essentially the same as this.

awels:
Other than using the existing function, looks good.

tests/datavolume_test.go
brybacki: right - updated

tests/datavolume_test.go
awels: Good point, I read it wrong, I thought you removed the original and then put it back, but if you are making a copy and then removing it, that works.

PR #1262, opened 2020-06-25
Correct check in badserver_tests

brybacki:
Signed-off-by: Bartosz Rybacki <brybacki@redhat.com>

**What this PR does / why we need it**:
Adds a missing check at the end of the tests.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #1278, opened 2020-07-01
Correct check in badserver_tests

maya-r:
Adds a missing check at the end of the tests.

It's a duplicate of #1262 which uses random numbers.
It's probably better than switching to always succeeding on GET requests, and I happened to have the code lying around already.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/badserver_test.go
+++ b/tests/badserver_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -14,6 +15,7 @@ import (
 
 var _ = Describe("Problematic server responses", func() {
 	f := framework.NewFrameworkOrDie("badserver-func-test")
+	const dvWaitTimeout = 500 * time.Second
 	var dataVolume *cdiv1.DataVolume
 
 	It("[rfe_id:4109][test_id:4110][crit:low][vendor:cnv-qe@redhat.com][level:component] Should succeed even if HEAD forbidden", func() {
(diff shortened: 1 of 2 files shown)


brybacki:
I am afraid that with this random implementation we might need even more wait time for the pessimistic case or the test itself will be flaky (see pull-containerized-data-importer-e2e-k8s-1.16-nfs)

```
"containerStatuses": [
{
"name": "importer",
"state": {
"waiting": {
"reason": "CrashLoopBackOff",
"message": "back-off 5m0s restarting failed container=importer pod=importer-badserver-dv_cdi-e2e-tests-badserver-func-test-m9cn4(0311f391-47c9-491e-a393-45e0a4d30b2c)"
}
},
"lastState": {
"terminated": {
"exitCode": 1,
"reason": "Error",
"message": "Unable to connect to http data source: expected status code 200, got 503. Status: 503 Service Unavailable",
"startedAt": "2020-07-02T00:41:13Z",
"finishedAt": "2020-07-02T00:41:13Z",
"containerID": "docker://929fc50a083ca24eee919dbc8d56b672a909485cb5044d102977bdceb2174c32"
}
},
"ready": false,
"restartCount": 6,
"image": "registry:5000/cdi-importer:latest",
"imageID": "docker-pullable://registry:5000/cdi-importer@sha256:9b65efd2a77660a886af901b94b472b07c67c45f4bd6d0f7d5b0fe4dbd19ca82",
"containerID": "docker://929fc50a083ca24eee919dbc8d56b672a909485cb5044d102977bdceb2174c32",
"started": false
}
],
```

awels:
I have to agree with Bartosz here, I think we had this discussion on the initial commit as well. We need to ensure that within a reasonable amount of time this succeeds, but we also want to introduce some randomness. Why is the current counter implementation not working?

I am just thinking out loud here. Maybe if on the first attempt to access we take a time stamp, then as long as the timestamp + N seconds > now, use the random to see if we succeed. But once timestamp + N seconds < now, then always succeed and clear the timestamp. That way the next run will take a new timestamp. That might need some synchronization to be sure the next properly sets the timestamp.

maya-r:
I figured that making it more likely to succeed means it's not going to fail (it works fine if HEAD fails, so only GET has to succeed: odds are 1 in 3). I'm somewhat surprised that it is still failing.

awels:
So kubernetes does a retry at increasing intervals. Its like 1,2,5,10,20,30,60s up to about 5 minutes between retries. So if the randomness on the first several attempts manage to elude success, then it is just going to take a much longer time before a retry and a potential success, at which point the time-out on the test has kicked in for failure.

maya-r:
all right

PR #1279, opened 2020-07-02
[test] Attempt to import from registry without the CA config map existing

awels:
Verify that the pod remains pending for a while until we create the config map and then the import succeeds.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -926,7 +926,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 
 			By("Verifying the POD remains pending for 30 seconds")
 			podName := naming.GetResourceName(common.ImporterPodName, pvc.Name)
-			for i := 0; i < 30; i++ {
+			Consistently(func() bool {
 				pod, err := f.K8sClient.CoreV1().Pods(f.Namespace.Name).Get(podName, metav1.GetOptions{})
 				if err == nil {
 					// Found the pod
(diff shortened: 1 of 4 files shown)


tests/datavolume_test.go
```diff
@@ -811,6 +813,50 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+				// Sleep one second and try again.
+				time.Sleep(time.Second)
+			}
+
+			By("Creating the config map")
+			_, err = utils.CopyRegistryCertConfigMapDestName(f.K8sClient, f.Namespace.Name, f.CdiInstallNs, cmName)
+			Expect(err).To(BeNil())
+
+			By(fmt.Sprintf("waiting for datavolume to match phase %s", string(cdiv1.Succeeded)))
+			utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, dataVolume.Name)
```
maya-r, line 857: Missing check of return value?

also: I assume the cleanup is handled by the namespace choice, is that correct?

awels: Yeah part of the framework is deleting the namespace at the end, which will in turn delete the data volume/pvc/pods that are in that namespace. Let me add the check.

tests/datavolume_test.go
```diff
@@ -811,6 +813,50 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			By("verifying pvc was created")
+			Eventually(func() bool {
+				// TODO: fix this to use the mechanism to find the correct PVC once we decouple the DV and PVC names
+				pvc, _ = f.K8sClient.CoreV1().PersistentVolumeClaims(dataVolume.Namespace).Get(dataVolume.Name, metav1.GetOptions{})
+				return pvc != nil && pvc.Name != ""
+			}, timeout, pollingInterval).Should(BeTrue())
+
+			By("Verifying the POD remains pending for 30 seconds")
+			podName := naming.GetResourceName(common.ImporterPodName, pvc.Name)
+			for i := 0; i < 30; i++ {
```
tomob, line 836: Why not use `Consistently` here?

awels: Because I didn't know Consistently existed. I will modify as it much clearer intend.

tomob: That's a fair reason :wink: :+1:

tomob:
LGTM

PR #1280, opened 2020-07-03
Remove k8s-1.16 cluster-sync logic

maya-r:
Now that we have a k8s-1.18 lanes replacing k8s-1.16 lanes in kubevirt/project-infra, we can remove unused logic.
Let's update the documentation while here.

**Release note**:
```release-note
NONE
```


awels:
Can you remove 1.15 as well. But leave the okd 4 lanes, they have some example code that is useful. Or extract that code somewhere else.

awels:
Actually, do that in a separate PR, this PR is fine.

hack/README.md
```diff
@@ -131,12 +131,12 @@ Environment Variables and Supported Values
...
 | Env Variable       | Default       | Additional Values           |
 |--------------------|---------------|-----------------------------|
-|KUBEVIRT_PROVIDER   | k8s-1.16      | k8s-1.17, os-3.11.0-crio,   |
+|KUBEVIRT_PROVIDER   | k8s-1.18      | k8s-1.17, os-3.11.0-crio,   |
 |KUBEVIRT_STORAGE*   | none          | ceph, hpp, nfs, ember_lvm   |
 |KUBEVIRT_PROVIDER_EXTRA_ARGS |      |                             |
 |NUM_NODES           | 1             | 2-5                         |
 
-KUBEVIRT_STORAGE is only supported for k8s-1.16 and OKD 4.3 providers
+KUBEVIRT_STORAGE is only supported for k8s-1.18 and OKD 4.3 providers
```
awels, line 139: This is not true anymore, the 1.17 provider supports the different storages as well. Also the OKD 4.3 provider no longer works, so we should take that part of the message out.

PR #1281, opened 2020-07-03
[test] Import from registry using credentials - tier 1 test

tomob:

This PR adds a tier1 test for importing an image from a repository that requires user credentials.

Fixes #

**Release note**:

```release-note
NONE
```


tomob:
> I would also make both registries readonly as we are not pushing to them in any way.

`registry-populate` container does push images to the registries.

awels:
Hmm I thought we just copied some files somewhere, so I guess we can keep one R/W while the other is read only.

PR #1283, opened 2020-07-06
Feature gates

brybacki:

As a CDI user I need to have the possibility to enable/disable optional or a preview features. 

If we are about to introduce a breaking change, then we need to have a switch that can be turned off by default ( so it is backwards compatible) and can be turned on in the environment where the new behavior/api can be supported.

Fixes #

TODO:
- [x] Discuss the representation - how do we represent the feature gates in the CDIConfig object? An array of key value pairs? We will have to make sure the CDIConfig stays a structural schema.
- [x] We need to double check that the CDIConfig object exists before the controllers are started (if not, and we make it dynamic, then it shouldn't matter because we will read the CDIConfig when it becomes available, we just need to make a sensible default if it is not there yet)
- [x] make the controllers watch the CDIConfig for when the feature gate changes.

**Release note**:

```release-note
Added new section - featureGates to CDIConfig.spec. 
```


brybacki:
for the "make the controllers watch the CDIConfig for when the feature gate changes." I think we are good, there is the CDIConfig conroller that does CDIConfig reconciliation - so it watchers changes, can merge some values, can set defaults.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -310,13 +310,15 @@ type CDIConfigSpec struct {
...
 	PodResourceRequirements  *corev1.ResourceRequirements `json:"podResourceRequirements,omitempty"`
+	FeatureGates             []string                     `json:"featureGates,omitempty"`
 }
 
 //CDIConfigStatus provides
 type CDIConfigStatus struct {
 	UploadProxyURL                 *string                      `json:"uploadProxyURL,omitempty"`
 	ScratchSpaceStorageClass       string                       `json:"scratchSpaceStorageClass,omitempty"`
 	DefaultPodResourceRequirements *corev1.ResourceRequirements `json:"defaultPodResourceRequirements,omitempty"`
+	FeatureGates                   []string                     `json:"featureGates,omitempty"`
```
mhenriks, line 321: When would be this be different from spec?  Are there default feature gates?  Where are they stored?

mhenriks: Similarly are there times when we won't all certain feature gates to be explicitly set?

Just trying to understand both properties

awels: The way I understand this, if the value is set, then the feature gate is on, if it is not set, it is off

brybacki: I did it by analogy to other config variables in CDI, so the user sets this on Spec, and the results are available on Status after it is reconciled and defaults are set. But maybe we do not need any defaults at all. 

Ideally we would do something like this:
1. We create a new feature that has to be off by default (backward compatibility or something), and it is set to ON by using feature gates, the logic checks to find feature gate. 
2. In the future - we decide all backward incompatible clients/users are not longer available so now we make the feature ON by default. To do this we remove the logic that checks feature gate, and remove feature gate. 

Is it ok? If not, then we need to set defaults somewhere, and in my opinion we can handle them in CDIConfig reconciliation  loop (then - the final value is nicely visible to the user i CDIConfig.Status), or maybe we should just handle defaults in FeatureGates?

mhenriks: It's still not clear to me when `spec. featureGates != status.featureGates`.  Can you give an example?  If not, then I say get rid of the status.

awels: That should teach me to look closer, yes I don't think we need the status if the spec is always the same as the spec.

brybacki: good, no need for additional complexity

brybacki:
Just pinging, is this good enough?

pkg/controller/feature-gates.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
+	"kubevirt.io/containerized-data-importer/pkg/common"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sync"
+)
+
+const (
+	// SkipWaitForFirstConsumerVolumes - if enabled will not schedule worker pods on a storage with WaitForFirstConsumer binding mode
+	SkipWaitForFirstConsumerVolumes = "SkipWaitForFirstConsumerVolumes"
```
mhenriks, line 14: Is it just me, or could there be a better name?  Like `HonorWaitForFirstConsumer`?

pkg/controller/feature-gates.go
```diff
@@ -0,0 +1,59 @@
+package controller
```
mhenriks, line 1: Probably have a dedicated package for this.  Will not be used for just controllers.

pkg/controller/feature-gates.go
```diff
@@ -0,0 +1,59 @@
...
+const (
+	// SkipWaitForFirstConsumerVolumes - if enabled will not schedule worker pods on a storage with WaitForFirstConsumer binding mode
+	SkipWaitForFirstConsumerVolumes = "SkipWaitForFirstConsumerVolumes"
+)
+
+// FeatureGates is a util for determining whether an optional feature is enabled or not.
+type FeatureGates struct {
+	client               client.Client
+	lock                 *sync.Mutex
+	lasValidFeatureGates []string
```
mhenriks, line 21: what is the use case for caching feature gates?  I think I would prefer an API that always checks the config object and returns an error if it cannot be obtained

pkg/controller/feature-gates.go
brybacki: Or Maybe even just `WaitForFirstConsumer`, name should be good, because once set it will stay for some time.

pkg/controller/feature-gates.go
brybacki: FeatureGate might be used in a loop in controllers where it is often read, but is by design rarely changed (and only from outside), so I was thinking this is ideal place to cache last valid value in case there is any error on client api/communication with k8s/ or even featureGates format etc... (and I was loosely following https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-config/config-map.go).

But I also understand that it might be good to just return error and fail this reconciliation loop, and retry again. 

Honestly I have no experience which way might behave better in big cluster.

mhenriks: The CDIConfig object is already cached with CR client, so not gaining much with additional cache IMO

pkg/controller/feature-gates.go
brybacki: now as I read this, `HonorWaitForFirstConsumerEnabled` is much better than Skip - done

pkg/controller/feature-gates.go
brybacki: done

pkg/controller/feature-gates.go
brybacki: Ok, so I can can try it. So now the api is (bool, error). I still cannot get used to it and to the `golang` way. I preferred the previous way, but it is not a strong preference.

brybacki:
just curious, can current fail: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1283/pull-containerized-data-importer-e2e-k8s-1.18-hpp/1283338740785221632 

be caused by not handling WFFC - which we are about to add with another PR?

awels:
I don't think so, if you look at the artifacts you will see both PVC and scratch PVC are bound, the upload server is running, but hasn't received a connection yet. The upload proxy is running, but the logs are empty. So my best guess is a transient networking failure. We see those once in a while on here.

pkg/feature-gates/feature-gates_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
+	"kubevirt.io/containerized-data-importer/pkg/common"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var (
+	log = logf.Log.WithName("feature-gates-test")
```
mhenriks, line 36: This does not appear to be used

PR #1292, opened 2020-07-12
Update kubevirtci.

maya-r:
Changes default KUBEVIRT_PROVIDER to k8s-1.18 that works, the previous
CDI-specific change was wrong and got overridden by another
'make generate'.

Removes ocp and okd kubevirtci providers.

```release-note
NONE
```


cluster-sync/okd-4.3/hpp-master-mc.yaml
```diff
@@ -1,27 +0,0 @@
-apiVersion: machineconfiguration.openshift.io/v1
```
awels, line 1: Can you open a PR against maybe the hostpath-provisioner-operator with the contents of these two files (hpp-master-mc.yaml and hpp-worker-mc.yaml). There have been quite a few times I have pointed people to these files with regards to Open Shift.

awels: Maybe add a contrib or something directory.

maya-r: Here: https://github.com/kubevirt/hostpath-provisioner-operator/pull/73

awels: thanks

PR #1296, opened 2020-07-12
Combine datavolume name tests into a single table.

maya-r:
They seem to vary only by the DataVolume name.

**Release note**:
```release-note
NONE
```


maya-r:
cc @brybacki

awels:
looks like a linter failure, make test-lint will tell you

brybacki:
good change 
The only thing lost is the comment on the dv name, but it is not needed
/lgtm

PR #1299, opened 2020-07-14
Run size validation before returning success on upload.

### LINKED ISSUE #1219: Improving upload user experience
/kind enhancement

**What happened**: I had difficulty uploading an image and in general there were too many variables to parse in order to figure out that the size was wrong and how to fix it. Hopefully removing some of these would improve the user experience for others.

I started with `virtctl image-upload --image-path=image.qcow2 --pvc-name=test --pvc-size=30G --uploadproxy-url=https://<service cluster ip> --insecure` . That created the `test` and `test-scratch` volumes, the CLI showed and upload, then it appeared to be hung. It was not, but I didn't catch the subtle cues.

Firstly, I looked for any errors that I could find in `kubectl get dv test`. I didn't really see any, but missed that the upload pod had restarted and the clue was in `kubectl logs cdi-upload-test -p`. The easy path would have been the error at the end of the logs here showing up in the `events` section at the end of `kubectl describe dv test`. 

Next, it's not documented what the different phases of the uploader are. This was a long shot to correcting the problem. I found them in the source, but the comments for the constants are not particularly helpful either. I tried to trace usages of the constants, but for some reason was not finding them with the project loaded into GoLand/IntelliJ, so I couldn't look at the logic to understand how phases progress either. So maybe some kind of block diagram of the phases and what they do could be helpful.

@mhenriks generously spent time with me on this and asked to try `virtctl image-upload dv test --image-path=image.qcow2 --size=30G --uploadproxy-url=https://<service cluster ip> --insecure` instead. I don't think it should be possible to do have `virtctl` accept such subtle differences, TBH. Anyway, after he noticed that there was a restart and we looked at the previous logs, I found that the size I was entering was not the same size as the image. In the virtctl CLI, `30G` was too small and `30Gi` was not parsed. It would be helpful if IEC notation was parsed, but even that wouldn't have been enough as I will discuss.

When the size was *still* too small by copy and pasting the requested size from the logs, I came to find it was still requesting 50102272 more bytes than the size of the image request. Only after adding all that and putting it in the virtctl `--size` argument did everything flow through.

What might have avoided all this is if virtctl stole the CLI UX of LVM2 for sizes. LVM has pretty great handling of options here. One that comes to mind is something like `--size=min` would have just let the uploader choose what size my pvc was going to be. I don't really care what size the image ends up being within 10%, TBH. 

To that end, I also don't care about this extra 50102272 bytes that the uploader seems to need, so IMO it would be better to just add that automatically. Maybe there would be a flag to turn off that behavior, but I bet telemetry would show that nobody uses it. 

I hope this helps. Please feel free to AMA and thanks for a great product!!

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.18.0
- Kubernetes version (use `kubectl version`): v1.17.4
- Cloud provider or hardware configuration: Bare metal
- Install tools: From release YAML per documentation
- Others:
```
[root@centos01 ~]# virtctl version
Client Version: version.Info{GitVersion:"v0.30.0-rc.1", GitCommit:"5f081a10672920f669b5042a7e1c15be47e238d0", GitTreeState:"clean", BuildDate:"2020-05-29T19:59:19Z", GoVersion:"go1.12.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{GitVersion:"{gitVersion}", GitCommit:"{gitCommit}", GitTreeState:"{gitTreeState}", BuildDate:"{buildDate}", GoVersion:"go1.12.8", Compiler:"gc", Platform:"linux/amd64"}
```

awels:
This is to give immediate feedback to the user when uploading.

Virtctl now supports asynchronous uploads, after the data has been transferred the connection is closed and post processing happens (convert/resize/etc) which may take a while. As part of that post processing we do a size validation. If the virtual disk image is > PVC size, the upload is considered failed. But with the async upload the connection is already closed and unless the user looks at the DataVolume, they will have a hard time understanding what happened.

This PR returns an 400 (BadRequest) and a message in the body based on the validation BEFORE we close the connection, so we can give more immediate feedback to the user. Currently virtctl doesn't print the contents of the response body, but this [PR](https://github.com/kubevirt/kubevirt/pull/3825) adds this.

Fixes #1219 (partially, virtctl needs update as well)

**Release note**:

```release-note
Enhancement: Return error code and message on upload when image validation fails on async upload.
```

### DIFF
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -40,7 +40,7 @@ const (
 
 type uploadFunc func(string, string, int) error
 
-type uploadRequestCreator func(string) (*http.Request, error)
+type uploadFileNameRequestCreator func(string, string) (*http.Request, error)
 
 var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:component]Upload tests", func() {
 
@@ -161,6 +161,7 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		Entry("succeed given a valid token (form async)", uploadFormAsync, true, http.StatusOK),
 		Entry("[posneg:negative][test_id:1369]fail given an invalid token", uploadImage, false, http.StatusUnauthorized),
 	)
+
 	It("Verify upload to the same pvc fails", func() {
 		By("Verify PVC annotation says ready")
 		found, err := utils.WaitPVCPodStatusReady(f.K8sClient, pvc)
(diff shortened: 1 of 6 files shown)


tests/upload_test.go
```diff
@@ -279,6 +302,40 @@ func uploadFormAsync(portForwardURL, token string, expectedStatus int) error {
 	return uploadToPath(formRequestFunc, portForwardURL, syncFormPath, token, expectedStatus)
 }
 
+func uploadFileNameToPath(requestFunc uploadFileNameRequestCreator, fileName, portForwardURL, path, token string, expectedStatus int) error {
```
mhenriks, line 305: This is a lot of copy/paste.  Can you get rid of `uploadToPath` or have it call this with the filename?

awels: Done, changed the requestFunc to take the extra file name parameter, and now I can use that everywhere.

PR #1302, opened 2020-07-15
Non-functional change cleanup: remove unused calls to UploadPodName

maya-r:
This function has no side effects and we ignore the return value.
While here, use the same function instead of hard-coding the name.
Might be a bit cleaner.

```release-note
NONE
```


maya-r:
rebased to re-invoke travis

Issue #1298, opened 2020-07-13
Support importing disks from a Vmware source


mansam:
/kind enhancement

In order to support a [VMware provider in the VM Import Operator](https://github.com/kubevirt/vm-import-operator/issues/292), support is needed in CDI for importing disks from a VMware source. This would involve the creation of a new DataVolumeSource. 

Transferring disks from VMware hosts would be done with nbdkit and the VDDK library. Because the VDDK library can't be redistributed, it would need to be provided via some kind of sidecar container with an image that the end user would have to provide.

The DataVolumeSource for VMware will likely need the following fields:

* the ESXi or vCenter URL
* the moRef of the VM this disk is attached to
* the fileName from the virtual disk's VirtualDeviceFileBackingInfo 
* the cert thumbprint of the vCenter or ESXi host
* Reference to a credentials secret containing the username and password

CC @mrnold @fdupont-redhat

fdupont-redhat:
Instead of the ESXi or vCenter URL, I think that `hostname` + `port` would be more generic.

Instead of the `moRef` of the VM, I think that the `UUID` is more valuable, as it is unique across vCenter and ESXi. The `moRef` is different between ESXi and vCenter, which means that passing the `moRef` from the inventory (sync'd from vCenter) to ESXi host would fail. With the `UUID`, you can use the [FindByUuid](https://godoc.org/github.com/vmware/govmomi/object#SearchIndex.FindByUuid) method.

mansam:
`URL` follows the pattern from the other DataVolumeSources, changing to hostname + port would be a break from that convention.

nbdkit wants the moref of the the VM, so we can't pass it the UUID. If in VM Import the import spec is being created with a vCenter URL + moref, then the CDI datasource will get the vCenter URL and moref passed to it. If the spec is created with an ESXi host and moref, that will get passed to the CDI datasource instead.

fdupont-redhat:
Fine to keep `URL` to not break the DataVolumeSources.

About `moRef`, I'll defer to @mrnold. If the importer pod is built with govmomi, it can retrieve the `moRef` from the `UUID`. But does it really need to ship govmomi? @mrnold, besides the fields mentioned in the issue description, do you need to retrieve more properties? If no, then shipping govmomi only to convert `UUID` to `moRef`  seems overkill.

[...]

PR #1306, opened 2020-07-17
Add DataVolumeSourceVDDK type

mansam:
WIP

This PR adds the `DataVolumeSourceVDDK` type and adds support to the
datavolume and import controllers for it. The datasource itself
will need to be done in a follow-up.
See https://github.com/kubevirt/containerized-data-importer/issues/1298

**Release note**:

```release-note
Added DataVolumeSourceVDDK type
```


awels:
Travis is failing on the linter, run make test-lint and make test-unit to make sure both unit tests and linter pass. go fmt -w <file_name> will normally fix any linter issues

pkg/apis/core/v1beta1/types.go
```diff
@@ -63,6 +63,7 @@ type DataVolumeSource struct {
 	Upload   *DataVolumeSourceUpload   `json:"upload,omitempty"`
 	Blank    *DataVolumeBlankImage     `json:"blank,omitempty"`
 	Imageio  *DataVolumeSourceImageIO  `json:"imageio,omitempty"`
+	VDDK     *DataVolumeSourceVDDK     `json:"vddk,omitempty"`
```
mhenriks, line 66: Please update the alpha API as well.  We committed to supporting alpha for a little longer.

awels:
I see the actual call to the data source is commented out, but that is coming in a different PR right?

mansam:
@awels Yes, @mrnold is working on the data source PR.

awels:
Okay, let me start the functional test conversation now, so we can think about it a little. This PR by itself is not functional but combined with the other PR, we would have to have some way of testing it. The usual way we do a functional test for a data source is create some kind of source in the cluster, create a DV that connects to that source, and see if it succeeds. I am not sure how that would work with this particular source. Any thoughts on how we could test this?

mansam:
I've been thinking about that over the last week or so. There is a vCenter server simulator that I've tried to get nbdkit to talk to, but I don't think it fully implements the VDDK apis. On the other hand, the nbdkit VDDK plugin uses a fake VDDK library to stub out the API calls for testing. We might be able to use something like that so that nbdkit on the importer pod can go through the motions, but we'd have to be very careful about making sure that it accurately represented the behavior of the real thing.

awels:
Okay cool, I just wanted to make sure that was on the radar. We want to ensure that once this is finished we have some functional testing in place, so that when we are maintaining it we have confidence it will remain working. 

One other thing I thought about that recently came up that is not represented here. The types.go is where we define what the fields are and you have updated those properly. However the operator creates the CRDs based on code. For the datavolumes the code is [here](https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/operator/resources/cluster/datavolume.go#L77) and as you can see that is out of sync with the reality in types.go. In #1307 I am syncing them up again, and I have a follow up PR ready to verify that they stay in sync. That will mean depending on which one goes in first, there will be some changes that will have to happen in this PR or #1307

mansam:
Thanks for the heads up. It makes sense to me for this PR to go in after #1307 and just make the appropriate changes here.

mrnold:
This one can probably be closed, since the commits were included in #1320.

awels:
Closed by request

PR #1307, opened 2020-07-19
Move CRDS from apiextensions v1beta1 to v1.

awels:
Ensure that our code based schema validation matches the types in the api.

In order to prepare for kubernetes 1.19 in which apiextensions v1beta1 will be removed, this PR moves all our CRDS to use the v1 apiextensions. As a side effect all the CRDS are now 'explainable' by kubectl explain <crd shortname>.
In addition this PR now validates that the generated schema matches the schema we have in code. The generates schemas are based on types.go in the core part of the api package. We generate a yaml from that using controller-gen from the kubebuilder project. At the same time we generate yaml from the code in the operator package and we compare the two ensure they stay in sync.

Fixes #

Please ensure I am not doing funny things to make sure things are in sync. I know I add some technically unneeded fields in the code so I can more easily compare the yamls.

**Release note**:

```release-note
Enhancement: CRDs now use apiextensions v1 instead v1beta, only supported on kubernetes >= 1.16
```


awels:
/hold
I need to add the controller-gen and yq binaries to the builder container so we can run the comparison. I also still need to write script that runs the comparisons.

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -222,6 +222,14 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *v1beta1.A
 	}
 
 	accessModes := spec.PVC.AccessModes
+	if len(accessModes) == 0 {
```
brybacki, line 225: is this part of this change?

pkg/controller/util_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
brybacki, line 16: It would be better to name this apiextensions or k8ext, not extv1, we are not using both versions so there is no need to append v1. And then update 'might' be easier if new API is compatible. 

On the other hand I see it is a rule used in all imports. And if we do not want to change the rule then I'll accept this as is.

pkg/operator/controller/controller_suite_test.go
```diff
@@ -86,7 +86,8 @@ var _ = BeforeSuite(func(done Done) {
 	Expect(err).NotTo(HaveOccurred())
 
 	opts := envtest.CRDInstallOptions{
-		CRDs: []*extv1beta1.CustomResourceDefinition{crd},
+		// TODO: Fix me before merge
```
brybacki, line 89: ?

tools/schema-exporter/schema-exporter.go
```diff
@@ -0,0 +1,52 @@
...
+	"kubevirt.io/containerized-data-importer/pkg/operator/resources/cluster"
+	cdioperator "kubevirt.io/containerized-data-importer/pkg/operator/resources/operator"
+	"kubevirt.io/containerized-data-importer/tools/util"
+)
+
+var (
+	exportPath = flag.String("export-path", "", "")
+)
+
+func main() {
```
brybacki, line 30: Probably it is easy to reason from code, but just a short description of what it does or why it is needed and what format it exports would be useful.

api/openapi-spec/swagger.json
```diff
@@ -3080,8 +3080,7 @@
     "description": "CDI is the CDI Operator CRD",
     "type": "object",
     "required": [
-     "spec",
-     "status"
+     "spec"
```
brybacki, line 3083: Just a question: where from do we get api/openapi-spec/swagger.json, how it compares to 
- pkg/apis/core/v1alpha1/openapi_generated.go and 
- pkg/apis/core/v1beta1/types_swagger_generated.go.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -23,9 +23,15 @@ import (
...
-// DataVolume provides a representation of our data volume
+// DataVolume is an abstraction on top of PersistentVolumeClaims to allow easy population of those PersistentVolumeClaims with relation to VirtualMachines
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=dv;dvs,categories=all
+// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase",description="The phase the data volume is in"
+// +kubebuilder:printcolumn:name="Progress",type="string",JSONPath=".status.progress",description="Transfer progress in percentage if known, N/A otherwise"
+// +kubebuilder:printcolumn:name="Restarts",type="integer",JSONPath=".status.restartCount",description="The number of times the transfer has been restarted."
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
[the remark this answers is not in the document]
awels: So the tool is part of the kubebuilder project, reference can be found here: https://book.kubebuilder.io/reference/markers/crd.html. One of the tools in kubebuilder is called controller-gen, it will generate CRD yamls based on the types.go files. So the thought is, I can now generate proper validation schemas from types.go, as well as generate proper validation  schemas from our code schema as yaml, and I can compare the two to prevent drift from happening (the part that I haven't added to this PR yet)

pkg/apis/core/v1alpha1/types.go
```diff
@@ -67,66 +74,69 @@ type DataVolumeSource struct {
 
 // DataVolumeSourcePVC provides the parameters to create a Data Volume from an existing PVC
 type DataVolumeSourcePVC struct {
-	Namespace string `json:"namespace,omitempty"`
-	Name      string `json:"name,omitempty"`
+	// The namespace of the source PVC
+	Namespace string `json:"namespace"`
```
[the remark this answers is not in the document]
awels: Yes, the name and namespace are required, so omitempty should not have been there. Now we had it correctly in our code schema (that they are required) so technically not a bug. But I want to prevent drift between the two so I have to make sure the generated from types.go and the generated from our code schema are the same.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -67,66 +74,69 @@ type DataVolumeSource struct {
...
 
 // DataVolumeBlankImage provides the parameters to create a new raw blank image for the PVC
 type DataVolumeBlankImage struct{}
 
 // DataVolumeSourceUpload provides the parameters to create a Data Volume by uploading the source
 type DataVolumeSourceUpload struct {
-	//Target string `json:"shouldUpload,omitempty"`
 }
 
 // DataVolumeSourceS3 provides the parameters to create a Data Volume from an S3 source
```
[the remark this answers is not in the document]
awels: AWS is an S3 source, but there are others that implement it.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -67,66 +74,69 @@ type DataVolumeSource struct {
...
 type DataVolumeSourceS3 struct {
 	//URL is the url of the S3 source
-	URL string `json:"url,omitempty"`
+	URL string `json:"url"`
 	//SecretRef provides the secret reference needed to access the S3 source
 	SecretRef string `json:"secretRef,omitempty"`
 }
 
 // DataVolumeSourceRegistry provides the parameters to create a Data Volume from an registry source
```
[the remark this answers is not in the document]
awels: Good point, I guess I should specify docker registry.

pkg/apiserver/webhooks/datavolume-validate.go
awels: Yes it is, so this one is kind of interesting. Before in our code schema we specified that accesModes is required, and this the len cannot be 0. However the PVC.spec doesn't say it is actually required (omitempty like before), and since we take the spec from the kubernetes code we have to follow it. So now it is no longer required. So this means I can make a PVC spec without accessModes. Kubernetes will actually fail that in a validation web hook, and thus so should we. This change is for that reason.

pkg/controller/util_test.go
awels: So if we use a v1 of anything, we usually name it something like <package short name>v1, like extv1, or corev1, or storagev1. We can name it whatever, I picked extv1, but there is no rule we can't name it something else.

pkg/operator/controller/controller_suite_test.go
awels: So envtest package actually uses apiextensions v1beta1 in their code, but since we just moved to v1 this doesn't compile. Now obviously we still want this to work, so we need to do something to fix it. Maybe find a newer version of envtest that uses v1? I didn't want to forget or have it get lost in the shuffle so I put the TODO on here.

tools/schema-exporter/schema-exporter.go
awels: Yes good point, I should put some comments in there about what this does. Its taking the schemas in code, and dumping them out as yaml files. But I should explain that.

api/openapi-spec/swagger.json
awels: Its generated during make generate by openapi-spec-generator that exists in <cdi repo>/bin

awels: And it is compiled by the cdi builder and put there during the rsync that happens when we compile things in the build container.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -219,18 +229,25 @@ const DataVolumeCloneSourceSubresource = "source"
...
+// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase"
 type CDI struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   CDISpec   `json:"spec"`
-	Status CDIStatus `json:"status"`
+	Spec CDISpec `json:"spec"`
+	// +optional
+	Status CDIStatus `json:"status,omitempty"`
```
mhenriks, line 242: not sure `omitempty` makes sense for a non pointer struct member

pkg/operator/controller/controller_suite_test.go
mhenriks: Looks like controller runtime was updated to accept `runtime.Object` instead.  Maybe try updating CR?

mhenriks: But I would not be sad if we get rid of the tests that require the `envtest` package

awels: So it looks like the controller-runtime > 0.4 requires a new version of go than we currently support.

pkg/apis/core/v1alpha1/types.go
awels: Good point, the optional does the trick, same in the betav1

pkg/operator/controller/controller_suite_test.go
mhenriks: Looks like 0.5.0 also may require k8s 1.17 libraries as well.

I think envtest is only needed for one case that is not super important.  How about we disable for now and come back to it when we update go and k8s client libs?

awels: Sounds like a plan. I will put some comments in there explaining what is going on.

awels:
/hold cancel
I will add the schema validation in separate PR.

pkg/apis/core/v1alpha1/types.go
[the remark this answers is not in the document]
brybacki: thanks for info, so it is a standard tool for k8s apis and the hint is in +kubebuilder, so I think it does not need more explanation, I can check the docs

pkg/controller/util_test.go
brybacki: ok

maya-r:
I think the os-3.11.0-crio lane should be removed before this is merged

awels:
Yes, but we want an alternative open shift lane first. I am working on that.

PR #1310, opened 2020-07-22
Update cdi-builder to use go version 1.14.6 (current latest)

awels:

Fix functional test framework to work with go >= 1.13. We were not
using the ginkgo ordering correctly, and our framework would call
flag.Parse() too soon, and mess with the go testing flag parsing.
As a result we had to alter some of the variables that we use to build
urls to be functions so that when called the required information is
available.
Fixed several tests not checking for errors.
Fixed dumping code being initialized several times messing with the
order in which failures were reported.

Fixes #

**Release note**:

```release-note
Enhancement: CDI now uses golang 1.14.6
```

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -42,7 +42,7 @@ RUN pip3 install j2cli && pip3 install operator-courier && \
     ln -s /opt/gradle/gradle-4.3.1/bin/gradle /usr/local/bin/gradle && \
     rm gradle-4.3.1-bin.zip
 
-ENV GIMME_GO_VERSION=1.12.14 GOPATH="/go" KUBEBUILDER_VERSION="2.2.0" ARCH="amd64"
+ENV GIMME_GO_VERSION=1.14.6 GOPATH="/go" KUBEBUILDER_VERSION="2.3.1" ARCH="amd64"
 ENV BAZEL_PYTHON=/usr/bin/python3
 
 RUN mkdir -p /gimme && curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=/gimme bash >> /etc/profile.d/gimme.sh
--- a/hack/build/run-functional-tests.sh
+++ b/hack/build/run-functional-tests.sh
@@ -72,6 +72,5 @@ if [ $retry_counter -eq $MAX_CDI_WAIT_RETRY ]; then
     exit 1
 fi
 
-test_command="${TESTS_OUT_DIR}/tests.test -test.timeout 180m ${test_args}"
-echo "${test_command}"
+test_command="${TESTS_OUT_DIR}/tests.test -test.timeout 360m ${test_args}"
 (cd ${CDI_DIR}/tests; ${test_command})
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -2,10 +2,11 @@ package tests
 
 import (
 	"fmt"
-	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
 	"regexp"
 	"strings"
 	"time"
+
+	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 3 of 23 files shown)


tests/tests_suite_test.go
```diff
@@ -1,17 +1,84 @@
...
 )
 
 func TestTests(t *testing.T) {
 	defer GinkgoRecover()
 	RegisterFailHandler(tests.CDIFailHandler)
+	BuildTestSuite()
 	RunSpecsWithDefaultAndCustomReporters(t, "Tests Suite", reporters.NewReporters())
 }
+
+// To understand the order in which things are run, read http://onsi.github.io/ginkgo/#understanding-ginkgos-lifecycle
```
brybacki, line 35: good to have this info

hack/build/docker/builder/Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN pip3 install j2cli && pip3 install operator-courier && \
     ln -s /opt/gradle/gradle-4.3.1/bin/gradle /usr/local/bin/gradle && \
     rm gradle-4.3.1-bin.zip
 
-ENV GIMME_GO_VERSION=1.12.14 GOPATH="/go" KUBEBUILDER_VERSION="2.2.0" ARCH="amd64"
+ENV GIMME_GO_VERSION=1.14.6 GOPATH="/go" KUBEBUILDER_VERSION="2.3.1" ARCH="amd64"
```
mhenriks, line 45: Should these changes be separate (and committed before) from the rest?

tests/framework/framework.go
```diff
@@ -96,152 +98,34 @@ type Framework struct {
...
-	utils.CacheTestsData(f.K8sClient, f.CdiInstallNs)
-	f.mux.Unlock()
-
-	return f, err
+	return f
 }
 
 // BeforeEach provides a set of operations to run before each test
 func (f *Framework) BeforeEach() {
-	if !f.SkipControllerPodLookup {
```
brybacki, line 236: not needed anymore? I cannot find it

tests/datavolume_test.go
```diff
@@ -763,18 +790,19 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
 					dataVolume := utils.NewDataVolumeWithHTTPImport(dataVolumeName, "1Gi", url)
 
 					By(fmt.Sprintf("creating new datavolume %s", dataVolume.Name))
 					dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
 					Expect(err).ToNot(HaveOccurred())
 
 					By(fmt.Sprintf("waiting for datavolume to match phase %s", cdiv1.Succeeded))
-					utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, dataVolume.Name)
+					err = utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, dataVolume.Name)
+					Expect(err).ToNot(HaveOccurred())
```
brybacki, line 805: good catch, 
I think I know why it happens, some of our test utils do check and fail inside the function logic, and some are just returning an err value so the caller can assert it,

hack/build/run-functional-tests.sh
```diff
@@ -72,6 +72,5 @@ if [ $retry_counter -eq $MAX_CDI_WAIT_RETRY ]; then
     exit 1
 fi
 
-test_command="${TESTS_OUT_DIR}/tests.test -test.timeout 180m ${test_args}"
-echo "${test_command}"
+test_command="${TESTS_OUT_DIR}/tests.test -test.timeout 360m ${test_args}"
```
brybacki, line 75: why 360m?

tests/framework/framework.go
awels: It was never used, so I figured I might as well remove it while I was cleaning house.

tests/datavolume_test.go
awels: Yeah, well I noticed this should have failed while I was testing and the test 'passed' so now it should fail properly if it fails.

hack/build/run-functional-tests.sh
awels: With some failures 180 was not long enough any more. so I increased it

hack/build/docker/builder/Dockerfile
awels: They can, however this PR should work with both go 12.14 and 14.6. I am going to need a follow up PR to change the container hash regardless.

tests/framework/framework.go
awels: So I was wrong, I found a location it is used.

PR #1312, opened 2020-07-23
Remove potential failed state from DV

awels:

Remove last remaining location where we could set a DV to failed during normal operations. We can still get in failed state due to the PVC claim being lost, but that is the only possibility at this point.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -21,13 +21,14 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


maya-r:
/lgtm
for posterity, awels expanded in a private conversation that since we are watching the PVCs associated with the DVs, if the PVC is somehow deleted in the middle of this reconcile loop, then it will trigger another reconcile that will re-create the PVC. So just continuing this particular reconcile without failure in this case is fine.

PR #1319, opened 2020-07-29
Reserve overhead when validating that a Filesystem has enough space

maya-r:
When validating disk space, reserve space for filesystem overhead

The amount of available space in a filesystem is not exactly
the advertise amount. Things like indirect blocks or metadata
may use up some of this space. Reserving it to avoid reaching
full capacity by default.

This value is configurable from the CDIConfig object spec,
both globally and per-storageclass.

The default value is 0.055, or "5.5% of the space is
reserved". This value was chosen because some filesystems
reserve 5% of the space as overhead for the root user and
this space doubles as reservation for the worst case
behaviour for unclear space usage. I've chosen a value
that is slightly higher.

This validation is only necessary because we use sparse
images instead of fallocated ones, which was done to have
reasonable alerts regarding space usage from various
storage providers.

---

Update CDIConfig filesystemOverhead status, validate, and
pass the final value to importer/upload pods.

Only the status values controlled by the config controller
are used, and it's filled out for all available storage
classes in the cluster.

Use this value in Validate calls to ensure that some of the
space is reserved for the filesystem overhead to guard from
accidents.

Caveats:

Doesn't use Default: to define the default of 0.055, instead
it is hard-coded in reconcile. It seems like we can't use a
default value.

Release note:
```release-note
By default, reserve 5.5% of a filesystem for filesystem overhead.
This value is configurable from the CDIConfig object spec, both globally and per-storageclass.

kubectl edit cdiconfig config:
spec:
  filesystemOverhead:
    global: "override-0.055-for-all-values"
    storageClass:
      mystorageclass: "override-per-storageclass"
```


maya-r:
Missing doc, unit tests, functional tests.
Code works, so I figured it's time to put it as a WIP pull request.

pkg/apis/core/v1beta1/types.go
```diff
@@ -305,19 +305,33 @@ type CDIConfig struct {
...
+//Percent is a string that can only be a value between [0,1)
+// XXX: validation doesn't actually seem to do anything.
+// Actually relies on the reconcile not to use bogus values.
+// +kubebuilder:validation:Pattern=`^(0\.[0-9]+|0)$`
+type Percent string
+
+//FilesystemOverhead defines the reserved size for PVCs with VolumeMode: Filesystem
+type FilesystemOverhead struct {
+	Global       Percent            `json:"global,omitempty"`
+	StorageClass map[string]Percent `json:"storageClass,omitempty"`
```
awels, line 317: Add a comment here explaining the purpose of the StorageClass map

pkg/operator/resources/cluster/cdiconfig.go
```diff
@@ -73,6 +73,20 @@ func createCDIConfigCRD() *extv1beta1.CustomResourceDefinition {
...
 								},
+								"filesystemOverhead": {
+									Description: "Override the space reserved for filesystem overhead. A value is between 0 and 1, if not defined it is 0.055 (5.5% overhead)",
+									Type:        "object",
+									Properties: map[string]extv1beta1.JSONSchemaProps{
+										"global": {
+											Description: "How much space of a Filesystem volume should be reserved for safety. This value is the global one to be used unless a per-storageClass value is chosen.",
+											Type:        "string",
+										},
+										"storageClass": {
```
awels, line 84: We need to specify more here about what goes in, in particular we want the pattern on value as well. It should look something like this:
```go
"storageClass": {
  Items: &extv1beta1.JSONSchemaPropsOrArray{
    Schema: &extv1beta1.JSONSchemaProps{
      Description: "Some description explaining what the storage class map does, same as in types.go",
      Type: "array",
      Properties: map[string]extv1beta1.JSONSchemaProps{
      "storageClass": {
        Description: "Name of the storage class",
	Type:        "object",
       },
       "overhead": {
          Description: "The overhead as a decimal",
	  Type:        "object",
          Pattern: "regex for pattern",
	},
    },
```
Not entirely sure this will work since you used a map. But I think the validation is useful

pkg/controller/util.go
```diff
@@ -212,6 +213,68 @@ func CreateScratchPersistentVolumeClaim(client client.Client, pvc *v1.Persistent
 	return scratchPvc, nil
 }
 
+// GetStorageClassByName looks up the storage class based on the name. If no storage class is found returns nil
+func GetStorageClassByName(client client.Client, name *string) (*storagev1.StorageClass, error) {
```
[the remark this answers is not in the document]
maya-r: I haven't modified the function in any way, just moved it to util.

I can offer 'EvaluateStorageClassSpec" as a name, but I'm not sure it's an improvement.

pkg/apis/core/v1beta1/types.go
```diff
@@ -305,19 +305,33 @@ type CDIConfig struct {
 	Status CDIConfigStatus `json:"status,omitempty"`
 }
 
+//Percent is a string that can only be a value between [0,1)
+// XXX: validation doesn't actually seem to do anything.
+// Actually relies on the reconcile not to use bogus values.
+// +kubebuilder:validation:Pattern=`^(0\.[0-9]+|0)$`
+type Percent string
```
mhenriks, line 312: Why is a string better than using a float?

May be nice to have a `NewPercent` function that validates and returns a `Percent`.

pkg/controller/config-controller.go
```diff
@@ -211,6 +216,62 @@ func (r *CDIConfigReconciler) reconcileDefaultPodResourceRequirements(config *cd
...
+	if err := r.client.List(context.TODO(), storageClassList, &client.ListOptions{}); err != nil {
+		return err
+	}
+
+	for _, storageClass := range storageClassList.Items {
+		storageClassName := storageClass.GetName()
+		storageClassNameOverhead, found := perStorageConfig[storageClassName]
+
+		if found {
+			valid, err := validOverhead(storageClassNameOverhead)
```
mhenriks, line 258: can validation be enforced by CRD schema?

pkg/apis/core/v1beta1/types.go
maya-r: An actual float is problematic for portability reasons, and so [the tooling discourages it](https://github.com/kubernetes-sigs/controller-tools/issues/245).

pkg/apis/core/v1alpha1/types.go
```diff
@@ -357,13 +357,28 @@ type CDIConfig struct {
 	Status CDIConfigStatus `json:"status,omitempty"`
 }
 
+//Percent is a string that can only be a value between [0,1)
+// (Note: we actually rely on reconcile to reject invalid values)
+// +kubebuilder:validation:Pattern=`^(0\.[0-9]+|0)$`
```
awels, line 362: I think this pattern will work better:
```go
^(0(?:\.\d{1,3})?|1)$
```

maya-r:
I think I'm going to have to add validation to a lot of other data sources, but that's probably a separate PR

maya-r:
The tests are fine if run independently, I just introduced some flakiness

maya-r:
The NFS tests still fail locally as being flaky. Simply adding a wait between the tests helps, but I don't know what else I could be doing to un-flake them.

pkg/controller/util.go
```diff
@@ -215,6 +216,77 @@ func CreateScratchPersistentVolumeClaim(client client.Client, pvc *v1.Persistent
...
+			return "0", nil
+		}
+
+		return "0", err
+	}
+
+	targetStorageClass, err := GetStorageClassByName(client, pvc.Spec.StorageClassName)
+	if err != nil {
+		klog.V(1).Info("Storage class", pvc.Spec.StorageClassName, "not found, trying default storage class")
+		targetStorageClass, err = GetStorageClassByName(client, nil)
```
awels, line 265: You populate err here, but the rest of the function never uses it. If you don't care about the err use _, or if you do care, check the value and do something with it. If you don't care explain in a comment why not.

tests/framework/nfs-utils.go
```diff
@@ -29,7 +29,11 @@ func createNFSPVs(client *kubernetes.Clientset, cdiNs string) error {
 
 func deleteNFSPVs(client *kubernetes.Clientset, cdiNs string) error {
 	for i := 1; i <= pvCount; i++ {
-		if err := utils.DeletePV(client, nfsPVDef(strconv.Itoa(i), utils.NfsService.Spec.ClusterIP)); err != nil {
+		pv := nfsPVDef(strconv.Itoa(i), utils.NfsService.Spec.ClusterIP)
+		if err := utils.DeletePV(client, pv); err != nil {
+			return err
+		}
+		if err := utils.WaitTimeoutForPVDeleted(client, pv, timeout); err != nil {
```
awels, line 36: Optimization here, right now you are doing the following:
1. Delete PV
2. Wait For PV to actually be gone
3. Go to 1
So you are sequentially waiting for the PV to be deleted before starting a new delete. How about you do
```go
for i := 1; i <= pvCount; i++ {
  pv := nfsPVDef(strconv.Itoa(i), utils.NfsService.Spec.ClusterIP)
  if err := utils.DeletePV(client, pv); err != nil {
    return err
  }
}
// kicked off deletion of all PVs I care about
for i := 1; i <= pvCount; i++ {
  if err := utils.WaitTimeoutForPVDeleted(client, pv, timeout); err != nil {
    return err
  }
}
```
In this order you start the deletion of the PVs before waiting for any of them to really be deleted. That way the deletion is in parallell while the waiting is sequential. And hopefully the wait will be less because when you get to the wait for the later PVs, they will be gone already.

pkg/controller/config-controller.go
awels: There is a regexp on the field now, to enforce 0 <= overhead <= 1

pkg/controller/import-controller.go
```diff
@@ -440,6 +442,11 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 		if err != nil {
 			return nil, err
 		}
+		fsOverhead, err := GetFilesystemOverhead(r.uncachedClient, pvc)
```
awels, line 445: Any particular reason to use the uncached client here? I see the cached client being used in the uploadServer.

awels (commented):
Looking pretty good, a few comments and one want. I would like to see a functional test in cdiconfig_test.go that verifies that if you add an invalid value for the percent field, that the update fails. So things like -0.04 or 1.00 or 0.9999 which don't match the regex.

pkg/controller/config-controller.go
maya-r: btw: I only managed to add it to the global overhead one, not to the map[storageclass]overhead.

pkg/controller/import-controller.go
maya-r: Nope, missed it. thanks

tests/framework/nfs-utils.go
maya-r: Good idea. done

awels: So the way it is now, it looks like you are waiting for just the last PV (since the PV variable is set to the last at the end of the first loop). But there is no guarantee in the order in which the PVs are deleted, it is possible at this point that some of the earlier ones aren't gone yet. I would either just re-create the pv defs in the second loop, or store them in a slice and loop over the slice in the second loop to ensure all the PVs are gone at the end.

awels (commented):
I think there is a small logic error in the nfs deletion logic. Otherwise I am happy.

awels (approved):
/lgtm
/approve

PR #1232, opened 2020-06-10
move api groups to v1beta1

mhenriks:
**What this PR does / why we need it**:

Updates api groups to v1beta1.  v1alpha1 is still supported if resources created with yaml/old clients.  No webhook conversion necessary ATM.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

obsoletes #1196

**Release note**:

```release-note
API Groups updated to v1beta1
```

PR #1321, opened 2020-07-30
enable webhooks for alpha resources

mhenriks:

Shamefully forgot to include webhooks in #1232 

Fixes #

now that alpha go packages added back, using them in functional test.

**Release note**:

```release-note
Enable webhooks for v1alpha1 resources
```

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -60,6 +60,9 @@ type Config struct {
 	// SkipNamespaceCreation sets whether to skip creating a namespace. Use this ONLY for tests that do not require
 	// a namespace at all, like basic sanity or other global tests.
 	SkipNamespaceCreation bool
+
+	// FeatureGates may be overridden for a framework
+	FeatureGates []string
 }
 
 // Clients is the struct containing the client-go kubernetes clients
@@ -111,7 +114,9 @@ type Framework struct {
 // flag parsing happens AFTER ginkgo has constructed the entire testing tree. So anything that uses information from flags
 // cannot work when called during test tree construction.
 func NewFramework(prefix string, config ...Config) *Framework {
-	cfg := Config{}
+	cfg := Config{
+		FeatureGates: []string{featuregates.HonorWaitForFirstConsumer},
+	}
 	if len(config) > 0 {
 		cfg = config[0]
 	}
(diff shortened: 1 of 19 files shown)


tests/framework/framework.go
```diff
@@ -152,9 +157,8 @@ func (f *Framework) BeforeEach() {
 		createNFSPVs(f.K8sClient, f.CdiInstallNs)
 	}
 
-	defaultFeatureGates := []string{featuregates.HonorWaitForFirstConsumer}
-	ginkgo.By(fmt.Sprintf("Configuring default FeatureGates %q", defaultFeatureGates))
-	f.setFeatureGates(defaultFeatureGates)
+	ginkgo.By(fmt.Sprintf("Configuring default FeatureGates %q", f.FeatureGates))
```
awels, line 160: Shouldn't this be f.Config.FeatureGates when assigning?

tests/transport_test.go
```diff
@@ -28,8 +29,11 @@ var _ = Describe("Transport Tests", func() {
 	)
 
 	var (
-		ns  string
-		f   = framework.NewFramework("transport", framework.Config{SkipNamespaceCreation: false})
+		ns string
+		f  = framework.NewFramework("transport", framework.Config{
```
awels, line 33: Isn't that the default config? skip namespace creation false, and honor wait for first consumer set?. I think we can just remove the config parameter.

mhenriks: `Config{}` will set no feature gates

tests/framework/framework.go
mhenriks: `Config` is embedded in `Framework` so `f.FeatureGates` works.  Like `pod.Name` == `pod.ObjectMeta.Name`

tests/transport_test.go
mhenriks: Ah right, that is the same as default, just as before.  Assuming we wanted to be explicit here for a reason?

awels: No I think its a leftover from a long time ago. Feel free to remove it.

tests/framework/framework.go
awels: Right, I missed that. I used that property with the clients to make that work.

PR #1322, opened 2020-07-31
Replace skopeo

tomob:

This commit removes dependence on skopeo (binary) and uses containers API. By doing that we're able to optimize the use of storage (scratch) space, storage I/O and download bandwidth.

Fixes #

Split into two commits to make the review easier.

**Release note**:

```release-note
NONE
```


maya-r:
Being unfamiliar I needed a heads up that "containers/image" is a library skopeo uses.
Looks like UnArchiveLocalTar is unused after this change.
Since we are dropping several tests, the files used by these tests are also orphaned, like docker-image.tar or tests/images/{,in}valid_manifests.

Can you elaborate more on what makes this more efficient? thanks.

tomob:
> Being unfamiliar I needed a heads up that "containers/image" is a library skopeo uses.
> Looks like UnArchiveLocalTar is unused after this change.

Good point.

> Since we are dropping several tests, the files used by these tests are also orphaned, like docker-image.tar or tests/images/{,in}valid_manifests.

Another good point :-)

> Can you elaborate more on what makes this more efficient? thanks.

The old way, with `skopeo` binary, it worked like this:
* the entire image was downloaded to scratch space
* the entire image was untarred to scratch space
* the requested file (`disk/...`) was used, and the rest of untarred files were discarded.

With this PR it works like this:
* the manifest of the image is downloaded
* the image is processed one layer at a time
* if a layer is a not tar or tar+gz, it is skipped (not downloaded)
* tar and tar+gz layers are opened and processed file-by-file
* if the file matches a requested file (`disk/...`), it is saved into the scratch space
* the rest of the image is discarded.

So, the efficiency is improved by:
* downloading only what's needed (saving on net I/O)
* not wasting time untarring the entire image (saving on disk I/O)
* not wasting time removing unnecessary files (saving on disk I/O)

tomob:
/test 
pull-containerized-data-importer-e2e-k8s-1.18-hpp-test-image

pkg/image/transport.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/containers/image/v5/image"
+	"github.com/containers/image/v5/pkg/blobinfocache"
+	"github.com/containers/image/v5/transports/alltransports"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+	"k8s.io/klog"
+)
+
+var (
+	cancelTimeout = 300 * time.Second
```
aglitke, line 39: Should we write constant names in uppercase to make it clearer?

pkg/image/transport.go
```diff
@@ -0,0 +1,266 @@
...
+)
+
+const (
+	ociTarLayerMediaType    = "application/vnd.oci.image.layer.v1.tar"
+	dockerTarLayerMediaType = "application/vnd.docker.image.rootfs.diff.tar"
+	whFilePrefix            = ".wh."
+)
```
aglitke, line 46: Same here.

pkg/image/transport.go
tomob: This will make them public. I think they're only relevant in the context of the file. I've seen lowercase constants in other places (e.g. `pkg/importer/registry-datasource.go`).

pkg/importer/registry-datasource.go
```diff
@@ -82,7 +82,7 @@ func (rd *RegistryDataSource) Transfer(path string) (ProcessingPhase, error) {
 	rd.imageDir = filepath.Join(path, containerDiskImageDir)
 
 	klog.V(1).Infof("Copying registry image to scratch space.")
-	err = image.CopyRegistryImage(rd.endpoint, path, containerDiskImageDir, rd.accessKey, rd.secKey, rd.certDir, rd.insecureTLS)
+	err = image.CopyRegistryImage(rd.endpoint, path, "^"+containerDiskImageDir+"/.+", rd.accessKey, rd.secKey, rd.certDir, rd.insecureTLS)
```
awels, line 85: Why is this now a regex? I am just curious.

tests/datavolume_test.go
```diff
@@ -1092,8 +1095,8 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 		})
 
 		It("Should create a new PVC when PVC is deleted during import", func() {
-			dataVolumeSpec := createRegistryImportDataVolume(dataVolumeName, "1Gi", tinyCoreIsoRegistryURL())
-			By(fmt.Sprintf("creating new datavolume %s", dataVolumeSpec.Name))
+			dataVolumeSpec := utils.NewDataVolumeWithHTTPImport(dataVolumeName, "1Gi", httpTinyCoreRateLimit())
```
awels, line 1098: Wasn't the point of this test to delete during a registry import?

pkg/image/transport.go
tomob: This should be a constant, by the way... :smile:

pkg/importer/registry-datasource.go
tomob: Hm..
I might have overthought it and made it more universal that we need. 

Maybe I should just treat `containerDiskImageDir` as a simple path prefix... which is what the regex does.

tests/datavolume_test.go
tomob: This was a test for https://issues.redhat.com/browse/CNV-5149. I needed to replace registry import with rate-limited http import, because the registry import was too fast and race conditions started to occur, making the test unreliable.

pkg/image/transport.go
awels: Can you change it to a constant then?

tomob: Done.

tests/datavolume_test.go
awels: As I said that test is specifically for registry imports, so if the registry import is too fast now (that is great btw), then we need to figure out some rate limiting for the registry. Maybe we can combine nginx as a proxy for the registry? Some Googling found people doing that.

PR #1326, opened 2020-08-03
Make it possible to import images needing conversion from servers that don't support range requests

### LINKED ISSUE #1073: CDI should fall back if HTTP server doesn't support byte ranges

>
/kind bug
> /kind enhancement

CDI will fail to download via HTTP if the server does not support byte ranges. This is because of the use of qemu-img convert to do the download. qemu-img uses the qemu curl block driver, which fundamentally requires random access, IOW there is no workaround for this limitation in qemu. It's also worth noting that downloading this way will likely not result in a very efficient download.

If qemu-img convert fails because of a limitation of qemu-img, CDI should automatically fallback to download to scratch space and convert from local file.

Here is the relevant part of the qemu curl driver which is going to generate the error:

https://github.com/qemu/qemu/blob/master/block/curl.c#L797-L802

As mentioned, this limitation of fundamental to the architecture of a qemu block driver.

**Environment**:
- Install tools:
- Others:

maya-r:
Some servers don't support range requests, and we have existing fallback for when qemu-img doesn't work, to fetch the file to scratch space and pass the pathname to qemu-img rather than URL. Let's use it.

Fixes #1073

Wrote some tests. Could've probably combined some unit tests in a table, but I'd have to pass a function for every table entry, which seems a bit convoluted.

**Release note**:
```release-note
Support importing from web servers that don't support range requests (used to fail with non-raw images)
```

### DIFF
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -272,6 +272,12 @@ func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certD
 		return nil, uint64(0), true, errors.Errorf("expected status code 200, got %d. Status: %s", resp.StatusCode, resp.Status)
 	}
 
+	acceptRanges, ok := resp.Header["Accept-Ranges"]
+	if !ok || acceptRanges[0] != "bytes" {
+		klog.V(2).Infof("Accept-Ranges isn't bytes, avoiding qemu-img")
+		brokenForQemuImg = true
+	}
+
 	if total == 0 {
 		// The total seems bogus. Let's try the GET Content-Length header
 		total = parseHTTPHeader(resp)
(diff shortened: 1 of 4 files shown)


pkg/importer/http-datasource.go
```diff
@@ -272,6 +272,12 @@ func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certD
 		return nil, uint64(0), true, errors.Errorf("expected status code 200, got %d. Status: %s", resp.StatusCode, resp.Status)
 	}
 
+	acceptRanges, ok := resp.Header["Accept-Ranges"]
+	if !ok || acceptRanges[0] != "bytes" {
```
mhenriks, line 276: Probably not a big deal, but I wonder if this condition should instead be `!= "none"`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

PR #1334, opened 2020-08-07
Fix how we determine base builder image.

awels:
Just hard code instead of trying to determine from previous image.

The way we determine the builder image base would fail if we didn't specify the sha256 in the original. And it would fail if we used anything but kubevirt/kubevirt-cdi-bazel-buider. So just hard code instead of doing matching.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
why is kubevirt-bot reporting failures after merging?

PR #1336, opened 2020-08-10
verify schema is synchronized using travis.

awels:

Over time the schema generated from types.go and the schema encoded in golang started to diverge. Most of the time this meant that types.go was ahead of the schema encoded in golang code. This PR generates yaml from both types.go and the golang code, and compares them with yq to ensure that both remain in sync. This will ensure that if new fields are added or fields are remove, or fields are modified both need to get changed or the build will fail.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -45,3 +45,13 @@ else
     echo "${DIFFROOT} is out of date. Please run hack/update-codegen.sh"
     exit 1
 fi
+
+echo "************** verifying schemas match *****************"
+go build -o ./bin/schema-exporter ./tools/schema-exporter
+./bin/schema-exporter -export-path ./_out/manifests/code_schema
+
+curl https://github.com/mikefarah/yq/releases/download/3.3.2/yq_linux_amd64 -L --output ./bin/yq
+chmod +x ./bin/yq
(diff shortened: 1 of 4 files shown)


hack/verify-codegen.sh
```diff
@@ -45,3 +45,13 @@ else
     echo "${DIFFROOT} is out of date. Please run hack/update-codegen.sh"
     exit 1
 fi
+
+echo "************** verifying schemas match *****************"
+go build -o ./bin/schema-exporter ./tools/schema-exporter
+./bin/schema-exporter -export-path ./_out/manifests/code_schema
+
+curl https://github.com/mikefarah/yq/releases/download/3.3.2/yq_linux_amd64 -L --output ./bin/yq
```
maya-r, line 53: can we do this in the builder docker image? it looks like travis will also run it in docker

awels: We try to put stuff that is relevant for the build in the builder image, this is not relevant for the build.

maya-r: It's just that I've worked on some projects that don't allow built packages to fetch anything from the Internet during the build so I'm conditioned to looking for a way to disable Internet fetches. With docker it's very easy (grab the dependencies, don't use docker), but if random scripts fetch it it's hard.

Probably fine because it is for development only though

awels: Right, all our building happens somewhere else. This is just to help us developers from forgetting to synchronize the two places we define a schema.

PR #1348, opened 2020-08-30
Update CONTRIBUTING.md

maya-r:
Suggest slack which has more people right now, and update maintainers

**Release note**:
```release-note
NONE
```


maya-r:
(This file also appears on the HPP repositories)

PR #1349, opened 2020-09-02
Fix the package download URL for xen-libs and containers-common

duyanyan:
The old download urls are not available now, update them with correct urls.

Fixes #
`WARNING: Download from http://download.fedoraproject.org/pub/fedora/linux/updates/31/Everything/x86_64/Packages/c/containers-common-0.1.41-1.fc31.x86_64.rpm failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from http://download.fedoraproject.org/pub/fedora/linux/updates/31/Everything/x86_64/Packages/x/xen-libs-4.12.2-2.fc31.x86_64.rpm failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found`

**Release note**:

```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -142,10 +142,10 @@ http_file(
 
 http_file(
     name = "xen-libs",
-    sha256 = "06147608a5c32e3678f67dd7ad87abb4cd50b0a234a2bbc3ef643f67eec05e53",
+    sha256 = "6fb0852f73a978d832ba19cb23bfb9c28e14e435842803dc148fbdf8bc5e3fbc",
     urls = [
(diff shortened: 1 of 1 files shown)


awels:
The https://storage.googleapis.com/builddeps/<hash> don't get there automatically. You will have to ask @rmohr to update the storage cache as well. Right now the originals are gone, but since we have the cache everything still works. If we merge this as is, as soon as a new version shows up, the build will start failing.

rmohr:
Oh I missed this ping during my PTO. There is now a periodic job in place. I think it runs twice a day. As soon as it detects merged dependencies which are not mirrored it will upload the dependencies to GCS and open a PR for review with the mirror location attached.

PR #1359, opened 2020-09-07
Use semantic DeepEqual check on DV update

### LINKED ISSUE #1341: Creating a DataVolume using `Mi` unit

/kind bug

**What happened**:
When I create datavolume using `Mi` unit, the status of datavolume is not updated. Only for the value divided by 1024 like 4096, 5120.
I got the following error in cdi-apiserver pod log.
```
E0818 01:49:34.399376       1 datavolume-validate.go:251] Cannot update spec for DataVolume default/dv
```

- create datavolume using `Mi` unit
```yaml
kind: DataVolume
metadata:
  name: dv
spec:
  source:
    http:
      url: "https://download.cirros-cloud.net/contrib/0.3.0/cirros-0.3.0-i386-disk.img"
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 5120Mi
```

**Environment**:
  - Tested on both version:
    - v1.11.0
    - v1.18.0
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```
- Install tools:
- Others:

arnongilboa:
As discussed in #1341 datavolume-controller mistakenly used reflect.DeepEqual() to check DV spec is not updated, however it rejected even a no-change (of size 4096Mi to 4Gi). This fix uses semantic.DeepEqual() instead so it won't be rejected.

Fixes #1341

**Release note**:
```release-note
Fixed a bug that prevented adding DataVolumes of sizes which are a multiple of 1024 (e.g 5120Mi).
```

### DIFF
--- a/hack/build/build-manifests.sh
+++ b/hack/build/build-manifests.sh
@@ -39,3 +39,6 @@ generateResourceManifest $generator $MANIFEST_GENERATED_DIR "operator" "everythi
 tempDir=${MANIFEST_TEMPLATE_DIR}
 processDirTemplates ${tempDir} ${OUT_DIR}/manifests ${OUT_DIR}/manifests/templates ${generator} ${MANIFEST_GENERATED_DIR}
 processDirTemplates ${tempDir}/release ${OUT_DIR}/manifests/release ${OUT_DIR}/manifests/templates/release ${generator} ${MANIFEST_GENERATED_DIR}
+
+testsManifestsDir=${CDI_DIR}/tests/manifests
+processDirTemplates ${testsManifestsDir}/templates ${testsManifestsDir}/out ${testsManifestsDir}/out/templates ${generator} ${MANIFEST_GENERATED_DIR}
--- a/pkg/apiserver/webhooks/datavolume-validate.go
+++ b/pkg/apiserver/webhooks/datavolume-validate.go
@@ -28,6 +28,7 @@ import (
 
 	"k8s.io/api/admission/v1beta1"
 	v1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kvalidation "k8s.io/apimachinery/pkg/util/validation"
@@ -285,7 +286,7 @@ func (wh *dataVolumeValidatingWebhook) Admit(ar v1beta1.AdmissionReview) *v1beta
 			return toAdmissionResponseError(err)
 		}
 
-		if !reflect.DeepEqual(dv.Spec, oldDV.Spec) {
+		if !apiequality.Semantic.DeepEqual(dv.Spec, oldDV.Spec) {
 			klog.Errorf("Cannot update spec for DataVolume %s/%s", dv.GetNamespace(), dv.GetName())
 			var causes []metav1.StatusCause
 			causes = append(causes, metav1.StatusCause{
--- a/pkg/apiserver/webhooks/datavolume-validate_test.go
+++ b/pkg/apiserver/webhooks/datavolume-validate_test.go
@@ -245,6 +245,74 @@ var _ = Describe("Validating Webhook", func() {
 			resp := validateAdmissionReview(ar)
 			Expect(resp.Allowed).To(Equal(true))
 		})
+
+		It("should reject DataVolume spec PVC size update", func() {
+			blankSource := cdiv1.DataVolumeSource{
+				Blank: &cdiv1.DataVolumeBlankImage{},
+			}
+			pvc := newPVCSpec(pvcSizeDefault)
+			newDataVolume := newDataVolume("testDv", blankSource, pvc)
+			newBytes, _ := json.Marshal(&newDataVolume)
+
+			oldDataVolume := newDataVolume.DeepCopy()
+			oldDataVolume.Spec.PVC.Resources.Requests["storage"] =
+				*resource.NewQuantity(pvcSizeDefault+1, resource.BinarySI)
+			oldBytes, _ := json.Marshal(oldDataVolume)
+
+			ar := &v1beta1.AdmissionReview{
+				Request: &v1beta1.AdmissionRequest{
+					Operation: v1beta1.Update,
+					Resource: metav1.GroupVersionResource{
+						Group:    cdiv1.SchemeGroupVersion.Group,
+						Version:  cdiv1.SchemeGroupVersion.Version,
+						Resource: "datavolumes",
+					},
+					Object: runtime.RawExtension{
+						Raw: newBytes,
+					},
+					OldObject: runtime.RawExtension{
+						Raw: oldBytes,
+					},
+				},
+			}
+
+			resp := validateAdmissionReview(ar)
+			Expect(resp.Allowed).To(Equal(false))
+		})
+
+		It("should accept DataVolume spec PVC size format update", func() {
+			blankSource := cdiv1.DataVolumeSource{
+				Blank: &cdiv1.DataVolumeBlankImage{},
+			}
+			pvc := newPVCSpec(pvcSizeDefault)
+			newDataVolume := newDataVolume("testDv", blankSource, pvc)
+			newBytes, _ := json.Marshal(&newDataVolume)
+
+			oldDataVolume := newDataVolume.DeepCopy()
+			oldDataVolume.Spec.PVC.Resources.Requests["storage"] =
+				*resource.NewQuantity(pvcSizeDefault, resource.DecimalSI)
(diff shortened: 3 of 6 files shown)


arnongilboa:
Seems like I pushed too early before running all the tests. Will push a fixed PR soon.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -28,6 +28,7 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:shortName=dv;dvs,categories=all
+// +kubebuilder:subresource:status
```
mhenriks, line 31: What does this do?  If I recall, we do not have subresource status enabled on our CRDs.

pkg/controller/datavolume-controller.go
```diff
@@ -764,10 +764,10 @@ func (r *DatavolumeReconciler) emitFailureConditionEvent(dataVolume *cdiv1.DataV
 }
 
 func (r *DatavolumeReconciler) emitEvent(dataVolume *cdiv1.DataVolume, dataVolumeCopy *cdiv1.DataVolume, curPhase cdiv1.DataVolumePhase, originalCond []cdiv1.DataVolumeCondition, event *DataVolumeEvent) error {
-	// Only update the object if something actually changed in the status.
-	if !reflect.DeepEqual(dataVolume, dataVolumeCopy) {
-		if err := r.client.Update(context.TODO(), dataVolumeCopy); err != nil {
-			r.log.Error(err, "Unable to update datavolume", "name", dataVolumeCopy.Name)
+	// Update datavolume status if it changed
+	if !reflect.DeepEqual(dataVolume.Status, dataVolumeCopy.Status) {
```
mhenriks, line 768: if we do not have subresource status enabled, does this really work as intended?  Spec will get updated when calling `Update`.

Also, this will not detect any changes to metadata.  I can think of at least one place where an annotation is added

pkg/apis/core/v1alpha1/types.go
arnongilboa: Discussed it with @awels.
Part of the patch was enabling status subresource, which is the right way to update status only.

mhenriks: okay I see you added the status subresource

pkg/controller/datavolume-controller.go
arnongilboa: It won't work without subresource status enabled.
Do we want to update the spec in some flows?

mhenriks: Also, I assume `Update` will update the spec so does this work as intended?  Is there an UpdateStatus call on controller runtime client?

mhenriks: We want to update metadata which requires a call to `Update`

mhenriks: https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/controller/datavolume-controller.go#L659-L664

mhenriks: You could substitute any annotation with a status field, though.

I think this issue is perhaps best addressed in the webhook.  I assume that PVCs have the same issue.  I think they just ignore updates to the spec?

arnongilboa: That explains the test failures.
AFAIU, moving annotation to a status field is the right way to do it, but won't it introduce lots of noise?
Looking at the PVC code.

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -271,7 +272,7 @@ func (wh *dataVolumeValidatingWebhook) Admit(ar v1beta1.AdmissionReview) *v1beta
 			return toAdmissionResponseError(err)
 		}
 
-		if !reflect.DeepEqual(dv.Spec, oldDV.Spec) {
+		if !apiequality.Semantic.DeepEqual(dv.Spec, oldDV.Spec) {
```
mhenriks, line 275: may want to add a couple unit tests to confirm this works as expected

arnongilboa: Ok, utests failing when they try to do a real change of the spec?

mhenriks: Yeah succeeding if resource values are the same but in different units.  Failing if values different.

arnongilboa: @mhenriks no way to submit DV update request with different units, so submitted with different format (BinarySI and DecimalSI) which fails before the patch and passes with it.

awels:
For a functional test example look here: https://github.com/kubevirt/containerized-data-importer/blob/dc6bad6473ec7caffdf0f0b7ab2a4a73277f2a54/tests/api_validation_test.go#L248-L293 those are mainly negative tests, but there are some positive tests as well. So you would have to add a new yaml file that uses 1024Mi and you would have to verify it updates progress successfully, so I would use the rate limited http endpoint for source URL.

tests/api_validation_test.go
```diff
@@ -290,6 +291,36 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
...
+			pvc, err := utils.WaitForPVC(f.K8sClient, f.Namespace.Name, dataVolumeName)
+			Expect(err).ToNot(HaveOccurred())
+			f.ForceBindIfWaitForFirstConsumer(pvc)
+
+			//Due to the rate limit, this will take a while, so we can expect the phase to be in progress.
+			By(fmt.Sprintf("Waiting for datavolume to match phase %s", string(cdiv1.ImportInProgress)))
+			err = utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, cdiv1.ImportInProgress, dataVolumeName)
+			if err != nil {
+				PrintControllerLog(f)
+				dv, dverr := f.CdiClient.CdiV1beta1().DataVolumes(f.Namespace.Name).Get(context.TODO(), dataVolumeName, metav1.GetOptions{})
```
awels, line 310: You can write this if dverr != nil as this:
```go
Expect(dverr).ToNot(HaveOccurred(),"datavolume %s phase %s", dv.Name, dv.Status.Phase))
```

awels (commented):
looks good one minor nitpick.

awels (approved):
/lgtm
/approve

awels:
Gah make test-lint failed.

maya-r:
(btw, I restarted the travis job on the travis website, I suspect /retest doesn't work for this job)

hack/build/build-manifests.sh
```diff
@@ -39,3 +39,6 @@ generateResourceManifest $generator $MANIFEST_GENERATED_DIR "operator" "everythi
 tempDir=${MANIFEST_TEMPLATE_DIR}
 processDirTemplates ${tempDir} ${OUT_DIR}/manifests ${OUT_DIR}/manifests/templates ${generator} ${MANIFEST_GENERATED_DIR}
 processDirTemplates ${tempDir}/release ${OUT_DIR}/manifests/release ${OUT_DIR}/manifests/templates/release ${generator} ${MANIFEST_GENERATED_DIR}
+
+testsManifetsDir=${CDI_DIR}/tests/manifests
```
maya-r, line 43: Manifests

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -245,30 +245,98 @@ var _ = Describe("Validating Webhook", func() {
...
+						Raw: oldBytes,
+					},
+				},
+			}
+
+			resp := validateAdmissionReview(ar)
+			Expect(resp.Allowed).To(Equal(false))
+		})
+
+		It("should accept DataVolume spec PVC size format update", func() {
```
maya-r, line 283: not an objection since this only introduces the test, not change the behaviour:
why do we allow changing the format in a way that changes the size, but not the size?

maya-r: oh, I guess I misread what resource.NewQuantity does.

arnongilboa: We don't allow changing the size, just format (binary/decima).

PR #1361, opened 2020-09-08
Make CDI infra deployments as critical addons.

maya-r:
While here, mark cdi-operator as linux-only, it doesn't take
its value from the CDI CR.

To cite Daniel Belenky that posted a pull request for this same feature,

> By toleratin the CriticalAddonsOnly taint, we ensure that our components
> will get a reserved spot for critical addons in case they are being
> evicted. This will reduce of chance of CDI component/s starving for a
> place on a node and permanently unavailable.

Fixes #1152

**Release note**:

```release-note
Add the CriticalAddonsOnly toleration to CDI infrastructure deployments
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"time"
+	"reflect"
 
 	"github.com/kubevirt/controller-lifecycle-operator-sdk/pkg/sdk"
 	sdkapi "github.com/kubevirt/controller-lifecycle-operator-sdk/pkg/sdk/api"
@@ -21,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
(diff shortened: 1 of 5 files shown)


awels:
if you rebase on latest master, the failures should go away.

awels:
/hold
Can you copy the functional test from: https://github.com/kubevirt/containerized-data-importer/pull/1152 and mark that one obsolete?

maya-r:
Not sure I have the ability to mark the PR obsolete

tests/operator_test.go
```diff
@@ -95,6 +97,89 @@ var _ = Describe("Operator tests", func() {
 		Expect(conditionMap[conditions.ConditionProgressing]).To(Equal(corev1.ConditionFalse))
 		Expect(conditionMap[conditions.ConditionDegraded]).To(Equal(corev1.ConditionFalse))
 	})
+
+	FIt("should deploy components that tolerate CriticalAddonsOnly taint", func() {
```
awels, line 101: I think you left in an F here.

maya-r:
Not sure why cdi-apiserver gets killed for not having the critical addon toleration on the Prow run, but not on my local run.

tests/operator_test.go
```diff
@@ -97,6 +99,114 @@ var _ = Describe("Operator tests", func() {
...
+			Key:    "CriticalAddonsOnly",
+			Value:  "",
+			Effect: corev1.TaintEffectNoExecute,
+		}
+
+		nodes, err := f.K8sClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")
+		Expect(err).ToNot(HaveOccurred())
+
+		defer func() {
```
awels, line 139: So this is somewhat nit picky. As the defer will do the same thing. But this will be the only test that cleans up this way, where all the other tests use the AfterEach mechanism provided by Ginkgo. Maybe move this into its own describe or context, and provide the contents of the defer in an AfterEach for consistency with the other tests.

awels:
Looks like the upgrade lane failed on something in the CI system, but the ceph lane failed because CDI is hard coded as a lookup. Either look it up with a list instead of get, and make sure there is only 1. Or somehow figure out how to pass the name of the CDI object to the tests.

maya-r:
Personal take-away from this failure: nodeselectorN are bad names for branches. I was confused why it failed when my own code should've dealt with this problem, but it looks like I pushed something else (an outdated version?)

maya-r:
ping

awels:
Sorry forgot about this PR, if it is taking too long please ping sooner.

PR #1366, opened 2020-09-09
Add import network support for importer pod

### LINKED ISSUE #1364: Add the ability to select a network for data import
/kind enhancement

**What happened**:

When data is imported through Containerized Data Importer, the Importer Pod is created on the default network of the namespace where CDI is deployed. This default network is the pod network and it may not be the most performant path to the data source.

In an environment where multus is configured, the user should be able to specify the network to use for data import. It could be part of the DataVolume spec and used to generate the ImporterPod spec.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 2.4.0
- Kubernetes version (use `kubectl version`): 1.18

fdupont-redhat:
In an environment where multus is configured, the user should be able to specify the network to use for data import. It could be part of the DataVolume spec and used to generate the ImporterPod spec. This pull request adds support for a new annotation on the PVC that will be used to set the network annotation on the pod.

Fixes #1364 

```release-note
Add support to specify import network in multus environment
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -649,6 +649,8 @@ func MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName, sourcePvcNamespace
 		}
 	}
 
+	util.SetPodNetwork(pvc, pod)
+
 	pod.Spec.Containers[0].Env = append(pod.Spec.Containers[0].Env, addVars...)
 
 	return pod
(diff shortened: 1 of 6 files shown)


fdupont-redhat:
@awels, I don't see tests in `import-controller_test.go` that checks the annotations of the pod. Should I add something? Where would you advise I put it?

awels:
So a few things we usually do for anything new, and I would consider being able to specify the network to use 'new'.
- unit test(s)
- functional test(s)
- documentation
In this case you only altered the import, but we also support upload and clone, which should also support this. As for the question, yes for the import, a test in import-controller_test.go should do it.

awels:
For the functional test, you will likely need a new lane in the CI that has multus support. Let me discuss how we can best give you this with the rest of the team.

awels:
Actually now that I think about it more, we only need to test that the annotation ends up on the pod. Once that is there we know kubernetes will work. So we might be able to avoid using a multus lane.

awels:
Note if you checkout your commit that isn't signed, you can do git commit --amend -s and it will sign it.

fdupont-redhat:
@awels thanks. I'll probably squash and sign when it's ready :smiley:

fdupont-redhat:
Closing in favor of https://github.com/kubevirt/containerized-data-importer/pull/1480.

PR #1367, opened 2020-09-10
Update qe-tools dependency to avoid a cryptic Go module issue.

maya-r:
We vendored the v0.1.3 version, but our vendoring code picked a
few commits before it, missing out on code we need.

**Release note**:

```release-note
NONE
```


dollierp:
Once merged, we need to have it backported to `release-1.21` branch.

PR #1374, opened 2020-09-16
Change how we look up the node placement.

awels:
Instead of trying to do it during the reconcile object creation, do it during the reconcile loop where
we already look up the CR anyway, and setting the value is free.

This PR fixes an issue where we tried to look up the CR to find the node placement information during the creation of the reconciler. The CR might not exist at that point in time and it would fail. For example here: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/803 Instead set the node placement information during the reconcile loop since we are
already looking up the CR there anyway.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -117,13 +117,6 @@ func newReconciler(mgr manager.Manager) (*ReconcileCDI, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


maya-r:
I was under the impression I need the arg to be set really early, but I can't quickly tell what I was looking at. But if it passes the one tests/operator_test.go test.

PR #1375, opened 2020-09-16
Replace docker reference parser from alltransport to just docker

awels:

Replace how we parse the docker reference to not include alltransports which pulls in a large number of dependencies we don't want.

Fixes #

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -52,7 +53,6 @@ require (
 	k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6
 	kubevirt.io/qe-tools v0.1.6
 	sigs.k8s.io/controller-runtime v0.6.2
-	sigs.k8s.io/controller-tools v0.4.0
```
mhenriks, line 55: hmmm I wonder why this is gone?  Was it removed in another PR?

pkg/image/transport.go
```diff
@@ -68,7 +68,11 @@ func buildSourceContext(accessKey, secKey, certDir string, insecureRegistry bool
 }
 
 func readImageSource(ctx context.Context, sys *types.SystemContext, img string) (types.ImageSource, error) {
-	ref, err := alltransports.ParseImageName(img)
+	parts := strings.SplitN(img, ":", 2)
```
mhenriks, line 71: I think we should assert that parts[0] == "docker"

go.mod
awels: That is what deps update took out.

pkg/image/transport.go
awels: Yeah the unit tests are failing because we use oci-archive (for tars) there. So we should support that one as well

go.mod
maya-r: it was probably used for generating the CDI CRD, which we are not doing now.

PR #1383, opened 2020-09-21
Ensure we can specify different CDI CR name.

awels:
Fix bug where the CR name was hard coded to CDI and it failed in HCO.

HCO names the CDI CR a different name than CDI, and we had hard coded a lookup of the CDI CR as cdi, and it failed when deployed with HCO. This PR fixes that issue, and adds some code so we can specify a CI lane that has a different cr name so we can guard against this happening again.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/controller/util.go
```diff
@@ -462,11 +463,12 @@ func filterCloneSourcePods(input []v1.Pod) []v1.Pod {
...
-	crKey := client.ObjectKey{Namespace: "", Name: "cdi"}
-	if err := c.Get(context.TODO(), crKey, cr); err != nil {
+	crList := &cdiv1.CDIList{}
+	if err := c.List(context.TODO(), crList, &client.ListOptions{}); err != nil {
 		return nil, err
 	}
-
-	return &cr.Spec.Workloads, nil
+	if len(crList.Items) != 1 {
+		return nil, fmt.Errorf("Number of CDI CRs != 0")
```
aglitke, line 471: != 1 should be the message

awels: Yup fixed.

aglitke (approved):

PR #1385, opened 2020-09-22
Skip the test if CDI CR is called something else

maya-r:
For example if it's deployed by HCO

Follow up to #1383, this is how the operator_tests.go code handles it.

**Release note**:

```release-note
NONE
```


awels (approved):
/lgtm
/approve

PR #1390, opened 2020-09-23
Fixing dependency on Operator Framework.

tomob:
Required through Operator Lifecycle Manager.

A pulled version of OF has a dependency to a repository which no longer exists.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
Ahhh you found how to fix the issue then? Does this mean the other PR can be closed?

tomob:
> Ahhh you found how to fix the issue then? Does this mean the other PR #1386 can be closed?

Yes, I think so. This one is definitely better that #1386.

awels:
So the semver was causing the issue?

tomob:
> So the semver was causing the issue?

No, it needs to be changed because of the change in the dependencies.

awels:
Gotcha. Seems like we just merged something that is causing a conflict.... I see no issues with the PR was just wondering about the changes in the semver

PR #1391, opened 2020-09-23
Only test upgrading from latest release to current if random_cr not set

awels:

Don't test the upgrade from latest release to current PR if RANDOM_CR variable is set. If it is set, then the wait will fail because it is looking for the wrong CR name. If it is set we explicitly want a different CR name.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
/lgtm
oops

PR #1400, opened 2020-09-25
Add library function to determine if a PVC has been populated fully.

awels:

Added a library function that checks if a PVC is fully populated. The logic is as following:
If PVC has no ownerRef, then we assume something else fully populated it and
will return true
If PVC has an ownerRef and its a DataVolume, then look up the DataVolume
If DV.status.Phase == succeeded, return true, return false otherwise.

Fixes #

**Release note**:

```release-note
Enhancement: When cloning a source owned by a DV, the clone now waits for the source to be fully populated before cloning
```

### DIFF
--- a/pkg/apis/core/v1alpha1/utils.go
+++ b/pkg/apis/core/v1alpha1/utils.go
@@ -21,11 +21,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// PopulationComplete indicates if the persistent volume passed in has been fully populated. It follow the following logic
+// IsPopulated indicates if the persistent volume passed in has been fully populated. It follow the following logic
 // 1. If the PVC is not owned by a DataVolume, return true, we assume someone else has properly populated the image
 // 2. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase succeeded return true
 // 3. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase !succeeded return false
-func PopulationComplete(pvc *corev1.PersistentVolumeClaim, getDvFunc func(name, namespace string) (*DataVolume, error)) (bool, error) {
+func IsPopulated(pvc *corev1.PersistentVolumeClaim, getDvFunc func(name, namespace string) (*DataVolume, error)) (bool, error) {
 	pvcOwner := metav1.GetControllerOf(pvc)
 	if pvcOwner != nil && pvcOwner.Kind == "DataVolume" {
 		// Find the data volume:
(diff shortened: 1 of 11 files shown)


pkg/apis/core/v1alpha1/utils.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PopulationComplete indicates if the persistent volume passed in has been fully populated. It follow the following logic
+// 1. If the PVC is not owned by a DataVolume, return true, we assume someone else has properly populated the image
+// 2. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase succeeded return true
+// 3. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase !succeeded return false
+func PopulationComplete(pvc *corev1.PersistentVolumeClaim, getDvFunc func(name, namespace string) (*DataVolume, error)) (bool, error) {
```
mhenriks, line 28: I wonder about the usage of this on KubeVirt side.  Specifically the case where a PVC is owned by a DV but population hasn't completed yet.

I imagine the KubeVirt side would be something like this:

```go
dvOwner, err := getDataVolumeOwner(pvc)
if err != nil{
   ...
}

if dvOwner != nil && dvOwner.Status != Succeeded {
   // wait until succeeded
}

// continue starting vm

```

if the usage above makes sense, then I'm not sure it really makes sense to vendor in a function over there

pkg/controller/clone-controller.go
```diff
@@ -213,6 +213,14 @@ func (r *CloneReconciler) reconcileSourcePod(sourcePod *corev1.Pod, targetPvc *c
 			return false, err
 		}
 
+		sourcePopulated, err := PopulationComplete(sourcePvc, r.client)
+		if err != nil {
+			return false, err
+		}
+		if !sourcePopulated {
```
mhenriks, line 220: Would this be making it suck that clone sources would have to be DataVolumes?  If so, that's not cool.

pkg/apis/core/v1alpha1/utils.go
mhenriks: whoops my bad

pkg/controller/clone-controller.go
mhenriks: again I was confused

awels: No regular PVCs without a DV will already be marked as sourcePopulated = true

PR #1408, opened 2020-09-29
[release-v1.18] Don't consider deleted config maps for CR locking

awels:

There is a race condition in the reconcile if the follow happens:
1. CDI is deployed
2. The CDI CR is deleted
3. The operator will start to delete the appropriate resources including the configmap.
4. A new CDI CR is created but the configmap is still around, because the control plane is slow.
5. Currently in the reconcile loop the configmap associated with the old CR will be considered valid and the phase will be set to deploying.
6. Now the CM is actually removed
7. The reconcile loop will never try to create a new configmap due to way things are being checked.

Fixes #

**Release note**:

```release-note
BugFix: Fix race with CDI CR locking configmap.
```

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -186,6 +186,12 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 		return reconcile.Result{}, err
 	}
 
+	if configMap != nil && configMap.DeletionTimestamp != nil {
+		// this is a left over from a previous CR, wait for it get deleted.
+		reqLogger.Info("Found previous config that is marked for deletion, requeueing")
+		return reconcile.Result{RequeueAfter: time.Second}, nil
+	}
+
 	if configMap == nil {
 		if cr.Status.Phase != "" {
 			reqLogger.Info("Reconciling to error state, no configmap")


pkg/operator/controller/controller.go
```diff
@@ -186,6 +186,12 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 		return reconcile.Result{}, err
 	}
 
+	if configMap != nil && configMap.DeletionTimestamp != nil {
+		// this is a left over from a previous CR, wait for it get deleted.
+		reqLogger.Info("Found previous config that is marked for deletion, requeueing")
+		return reconcile.Result{RequeueAfter: time.Second}, nil
```
brybacki, line 192: why a Second?

awels: Basically to prevent us from looping really fast until the configmap is actually gone. I suspect this might occur on a highly loaded control plane. So instead of adding to the load, just wait a little bit.

pkg/operator/controller/controller.go
```diff
@@ -186,6 +186,12 @@ func (r *ReconcileCDI) Reconcile(request reconcile.Request) (reconcile.Result, e
 		return reconcile.Result{}, err
 	}
 
+	if configMap != nil && configMap.DeletionTimestamp != nil {
```
mhenriks, line 189: I honestly don't think this will be helpful.  I don't think there is any way the configmap will have a deletion timestamp and the CDI CR will not.  And you will not get here if the CDI CR deletion timestamp is not ini

PR #1411, opened 2020-09-29
Expand negative too-big upload tests to xz, sync upload endpoint

maya-r:
Note that this adds an XEntry, because the expected failure doesn't
happen with .raw.xz files.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -194,7 +194,7 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 
 	})
 
-	It("[test_id:4989]Verify validation error message on async upload if virtual size > pvc size", func() {
+	DescribeTable("Verify validation error message on async upload if virtual size > pvc size", func(filename string) {
 		By("Verify PVC annotation says ready")
 		found, err := utils.WaitPVCPodStatusReady(f.K8sClient, pvc)
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


brybacki:
`because the expected failure doesn't happen with .raw.xz files.` - Does it meant we have a bug? Do we have it reported?

awels:
So the thing that is weird, it works on sync upload, but not async upload, that would indicate we have a bug in the async version of the upload.

maya-r:
It doesn't work on sync upload. I'm using "XEntry" for it.
The ceph-only failure on async upload is surprising

awels:
Whoops I reversed, I haven't looked at the failures on the CI lane yet. Just that you skipped the test in one and not the other. Either way it sounds like a bug.

tests/upload_test.go
```diff
@@ -207,9 +207,44 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+		By("Verify size error in logs")
+		matchString := fmt.Sprintf("is larger than available size")
+		Eventually(func() string {
+			log, err := tests.RunKubectlCommand(f, "logs", uploadPod.Name, "-n", uploadPod.Namespace)
+			Expect(err).NotTo(HaveOccurred())
+			By(log)
+			return log
+		}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(ContainSubstring(matchString))
+	},
+		PEntry("fail given a RAW XZ file", utils.UploadFileLargeVirtualDiskXz),
```
awels, line 245: So the way I understand P(or X) is that it is a skip. So we are basically skipping this particular permutation?

maya-r: yeah

maya-r:
OK, I see why the test fails on ceph but not on other storage classes:

1. First we perform the converted write to /data/disk.img, in TransferFile
2. Then we validate the size

For ceph, we have an exactly 1Gi RBD with a filesystem on top of it, so step #1 fails.
For local storage, we have 35Gi (from kubevirtci setup), so step #1 succeeds, but step #2 fails.

tests/upload_test.go
awels: Well my point is, why are we skipping this test. The tests are to validate behavior, if the behavior you are seeing is not the expected result of the test. Then the behavior is wrong and should be fixed. We shouldn't be skipping the test unless we have a good reason to do so.

maya-r:
obsoleted by #1466

PR #1413, opened 2020-10-01
Fix imports for images with no info about MediaType.

tomob:

An image from a registry might have MediaType for its layer empty. This
PR adds code which tries to guess the MediaType basing on so-called
magic numbers aka file signatures.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/image/transport.go
```diff
@@ -129,6 +133,71 @@ func isDir(path string) bool {
...
+		head[0x105] == 0x72
+}
+
+func isFiletypeGzip(head []byte) bool {
+	// Signature from https://en.wikipedia.org/wiki/List_of_file_signatures
+	return len(head) > 1 &&
+		head[0] == 0x1F && head[1] == 0x8B
+}
+
+func guessMediaType(ctx context.Context, src types.ImageSource, layer types.BlobInfo, cache types.BlobInfoCache) (string, error) {
```
awels, line 152: This feels like it is doing the same as https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/importer/format-readers.go#L227 more or less. But I can't come up with a good way of using that here of the top of my head.

tomob: Oh, I didn't know about this code.

I've given it some thought. Maybe we could get rid of the whole `MediaType` checking and just use FormatReaders? If a layer can be read using FR, it's good enough for the import process. The declared edia type wouldn't matter.

What do you think?

awels: Yeah I think that would be good. I just hate re-implementing code we have been using successfully for some time now. Also maybe some unit tests to verify it is working as expected?

awels (approved):
/lgtm
/approve

awels:
looks like there are some failing tests now....

tomob:
> looks like there are some failing tests now....

This is weird... With `FormatReaders` reading is very sloooooow. But I can't see any reason for it.

pkg/importer/transport.go
```diff
@@ -36,15 +34,11 @@ import (
...
-	dockerTarLayerMediaType = "application/vnd.docker.image.rootfs.diff.tar"
-	ociTarLayerMediaType    = "application/vnd.oci.image.layer.v1.tar"
-	whFilePrefix            = ".wh."
+	whFilePrefix = ".wh."
 )
 
 func commandTimeoutContext() (context.Context, context.CancelFunc) {
-	ctx := context.Background()
-	return context.WithTimeout(ctx, cancelTimeout)
+	return context.WithCancel(context.Background())
```
tomob, line 41: @awels I think we cannot say what a "generic" timeout for any situation should be. So I got rid of it; the importer will now patiently download images for as long as it takes.

tools/cdi-func-test-file-host-init/nginx.conf
```diff
@@ -120,7 +120,7 @@ http {
             proxy_set_header X-Forwarded-Proto "https";
             proxy_read_timeout                 900;
 
-            limit_rate 100k;
+            limit_rate 1M;
```
tomob, line 123: @awels It seems 100k (per second) was too harsh. The test was taking ages to run.

awels (commented):
One question, you mentioned with the FormatReaders it was much slower. How did we fix that? Hopefully not by just waiting longer. Or was the timeout causing a restart and that is why it was slower?

tomob:
> One question, you mentioned with the FormatReaders it was much slower. How did we fix that? Hopefully not by just waiting longer. Or was the timeout causing a restart and that is why it was slower?

No, I found the bug in the old code (the one I replaced) which made it skip some image layers (their media type was `...tar`, while it actually was gzipped, so the type should have been `....tar+gz`). After I fixed that both versions took about the same.

So, it's not that FormatReaders are slower – it's that my code that was too fast because it skipped layers it should process.

awels:
Okay so by switching to code we knew was working we fixed a bug before it became a bug. Nice

PR #1417, opened 2020-10-06
Check for expected changes after CDI upgrade

arnongilboa:

Upgrade testing should verify all cdi pod names, uids and image versions changed (CNV-5252).

**Release note**:
```NONE
```

### DIFF
--- a/cluster-sync/sync.sh
+++ b/cluster-sync/sync.sh
@@ -15,6 +15,7 @@ CDI_INSTALL="install-operator"
 CDI_NAMESPACE=${CDI_NAMESPACE:-cdi}
 CDI_INSTALL_TIMEOUT=${CDI_INSTALL_TIMEOUT:-120}
 CDI_AVAILABLE_TIMEOUT=${CDI_AVAILABLE_TIMEOUT:-480}
+CDI_PODS_UPDATE_TIMEOUT=${CDI_PODS_UPDATE_TIMEOUT:-480}
 CDI_UPGRADE_RETRY_COUNT=${CDI_UPGRADE_RETRY_COUNT:-60}
 
 # Set controller verbosity to 3 for functional tests.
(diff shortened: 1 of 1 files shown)


arnongilboa:
It looks like the patch is a bit buggy. Working on a fix.

arnongilboa:
@awels can you please review?

cluster-sync/sync.sh
```diff
@@ -79,6 +80,40 @@ function wait_cdi_available {
   fi
 }
 
+OLD_CDI_VER_PODS="/tmp/old_cdi_ver_pods"
```
mhenriks, line 83: maybe `./_out/tests` is better than `/tmp`?

arnongilboa: sure

cluster-sync/sync.sh
```diff
@@ -79,6 +80,40 @@ function wait_cdi_available {
...
+  if [ -f $NEW_CDI_VER_PODS ] ; then
+    mv $NEW_CDI_VER_PODS $OLD_CDI_VER_PODS
+  fi
+  wait_time=0
+  ret=0
+  while [[ $ret -eq 0 ]] && [[ $wait_time -lt ${CDI_PODS_UPDATE_TIMEOUT} ]]; do
+    wait_time=$((wait_time + 5))
+    _kubectl get pods -n $CDI_NAMESPACE -o=jsonpath='{range .items[*]}{.metadata.name}{"\n"}{.metadata.uid}{"\n"}{.spec.containers[*].image}{"\n"}{end}' > $NEW_CDI_VER_PODS
+    if [ -f $OLD_CDI_VER_PODS ] ; then
+      grep -qFxf $OLD_CDI_VER_PODS $NEW_CDI_VER_PODS || ret=$?
```
mhenriks, line 99: Will this return non zero when there is any difference between old and new?  If so, is that really sufficient to say that all pods have changed?

arnongilboa: @mhenriks no, grep will return non-zero when **none** of the lines (pod name, uid or image) on the old file appears in the new file, which sure means all pods have changed. However it does not check all the new pods are Running etc., which out of scope here.

PR #1424, opened 2020-10-09
Scratch import bug

brybacki:

Fixed the creation of scratch space (after it is being deleted) in some rare scenarios.

Here is the scenario that is fixed by the PR - after requesting the creation of the pod and the scratch-space:
1. The Pod is being created so it shows in the system as Pending (waits for all PVC to be available), this is observed in controller and it tries to create scratch again, but scratch is already in the system (also Pending), so controller sets  condition "Claim Pending" and returns.

2. Some external action - be it a script or a user - removes PVC (if I am not mistaken the finalizer: "kubernetes.io/pvc-protection" does not work if pod is Pending/not yet scheduled), now the only thing the controller can see is a PVC event, but the scratch PVC is not found so controller returns without taking any action.

No further events for the POD (still Pending - no changes). From this point the Pod stays in Pending state indefinitely. 

The **problem is fixed** by changing the event watcher query. When the scratch pvc event is received the PVC that owns the POD that OWNS the scratch pvc is added to the reconcile event queue. So if the ownership looks like this: 

`PVC A -> POD B -> ScratchPVC C 
`
and the `DeleteEvent` for `ScratchPVC C` is received then the `PVC A` is returned. This is the PVC that is the target for import. When this PVC is reconciled the scratch space is correctly re-created if needed.

Also an issue was found in the utility function to create a scratch PVC. It did not return an error.

Additional I have added a check in import controller so it now skips pvc that is in use by other pod. The same way as in upload and clone controller. Without this testing was almost impossible. And sometimes the `consumer pod ` was in "CreatingContainer" state even after the importer pod started.

Fixes #

**Release note**:

```release-note
Import controller now skips pvc that is in use by other pod. 
Fixed the creation of scratch space (after being deleted) in some rare scenario.
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -265,7 +265,7 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 				for _, pod := range podsUsingPVC {
 					r.log.V(1).Info("can't create import pod, pvc in use by other pod",
 						"namespace", pvc.Namespace, "name", pvc.Name, "pod", pod.Name)
-					r.recorder.Eventf(pvc, corev1.EventTypeWarning, UploadTargetInUse,
+					r.recorder.Eventf(pvc, corev1.EventTypeWarning, ImportTargetInUse,
 						"pod %s/%s using PersistentVolumeClaim %s", pod.Namespace, pod.Name, pvc.Name)
 
 				}
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -16,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -84,6 +85,9 @@ const (
 
 	// creatingScratch provides a const to indicate scratch is being created.
 	creatingScratch = "CreatingScratchSpace"
+
+	// ImportTargetInUse is reason for event created when an import pvc is in use
+	ImportTargetInUse = "ImportTargetInUse"
 )
 
 // ImportReconciler members
@@ -246,11 +250,28 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
 	if pod == nil {
 		if isPVCComplete(pvc) {
 			// Don't create the POD if the PVC is completed already
 			log.V(1).Info("PVC is already complete")
 		} else if pvc.DeletionTimestamp == nil {
+			podsUsingPVC, err := getPodsUsingPVCs(r.client, pvc.Namespace, sets.NewString(pvc.Name), false)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			if len(podsUsingPVC) > 0 {
+				for _, pod := range podsUsingPVC {
+					r.log.V(1).Info("can't create import pod, pvc in use by other pod",
+						"namespace", pvc.Namespace, "name", pvc.Name, "pod", pod.Name)
+					r.recorder.Eventf(pvc, corev1.EventTypeWarning, UploadTargetInUse,
+						"pod %s/%s using PersistentVolumeClaim %s", pod.Namespace, pod.Name, pvc.Name)
+
+				}
+				return reconcile.Result{Requeue: true}, nil
+			}
+
 			if _, ok := pvc.Annotations[AnnImportPod]; ok {
 				// Create importer pod, make sure the PVC owns it.
 				if err := r.createImporterPod(pvc); err != nil {
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"reflect"
 	"strconv"
+	"time"
 
 	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
 
@@ -290,13 +291,19 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 			if err := r.client.Delete(context.TODO(), pod); IgnoreNotFound(err) != nil {
 				return reconcile.Result{}, err
 			}
-			return reconcile.Result{}, nil
+		} else {
+			// Pod exists, we need to update the PVC status.
+			if err := r.updatePvcFromPod(pvc, pod, log); err != nil {
+				return reconcile.Result{}, err
+			}
 		}
+	}
 
-		// Pod exists, we need to update the PVC status.
-		if err := r.updatePvcFromPod(pvc, pod, log); err != nil {
-			return reconcile.Result{}, err
-		}
+	if !isPVCComplete(pvc) {
+		// We are not done yet, force a re-reconcile in 2 seconds to get an update.
+		log.V(1).Info("Force Reconcile pvc import not finished", "pvc.Name", pvc.Name)
+
+		return reconcile.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 	return reconcile.Result{}, nil
 }
@@ -599,6 +606,8 @@ func (r *ImportReconciler) createScratchPvcForPod(pvc *corev1.PersistentVolumeCl
 		return err
 	}
 	if k8serrors.IsNotFound(err) {
+		r.log.V(1).Info("Creating scratch space for POD and PVC", "pod.Name", pod.Name, "pvc.Name", pvc.Name)
+
 		storageClassName := GetScratchPvcStorageClass(r.client, pvc)
 		// Scratch PVC doesn't exist yet, create it. Determine which storage class to use.
 		_, err = CreateScratchPersistentVolumeClaim(r.client, pvc, pod, scratchPVCName, storageClassName)
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -211,6 +211,7 @@ func CreateScratchPersistentVolumeClaim(client client.Client, pvc *v1.Persistent
 	scratchPvc := &v1.PersistentVolumeClaim{}
 	if err := client.Get(context.TODO(), types.NamespacedName{Name: scratchPvcSpec.Name, Namespace: pvc.Namespace}, scratchPvc); err != nil {
 		klog.Errorf("Unable to get scratch space pvc, %v\n", err)
+		return nil, err
 	}
 	klog.V(3).Infof("scratch PVC \"%s/%s\" created\n", scratchPvc.Namespace, scratchPvc.Name)
 	return scratchPvc, nil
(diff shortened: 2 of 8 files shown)


pkg/controller/import-controller.go
```diff
@@ -139,15 +145,108 @@ func NewImportController(mgr manager.Manager, log logr.Logger, importerImage, pu
...
+				pvc := evt.Object.(*corev1.PersistentVolumeClaim)
+				owner := metav1.GetControllerOf(pvc)
+				// it has an owner pod, so it has to be a scratch PVC, so we need to fetch the controlling PVC
+				// PVC -> importerPod -> scratchPVC
+				if owner != nil && owner.Kind == "Pod" {
+					log.V(0).Info("---------------->  DeleteEvent 2 for PVC owner", "scratchPvc", pvc.Name, "owner", owner)
+
+					controllingPod := &corev1.Pod{}
+					if err := c.Get(context.TODO(), types.NamespacedName{Name: owner.Name, Namespace: evt.Meta.GetNamespace()}, controllingPod); err != nil {
+						// TODO: assume this is a scratch PVC without owner - so return an error, or schedule event for the pvc?
```
brybacki, line 210: So I wonder if we can leave it like it is right now (so we assume the pvc that has pod owner that has pvc owner must be scratch), or maybe we should mark scratch by Annotation and check the Annotation to be sure!

pkg/controller/import-controller.go
```diff
@@ -139,15 +145,108 @@ func NewImportController(mgr manager.Manager, log logr.Logger, importerImage, pu
...
+				} else {
+					log.Error(nil, "UpdateEvent received with no new metadata", "event", evt)
+				}
+			},
+			DeleteFunc: func(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+				if evt.Meta == nil {
+					log.Error(nil, "DeleteEvent received with no metadata", "event", evt)
+					return
+				}
+				log.V(0).Info("----------------> DeleteEvent 1", "event", evt)
```
brybacki, line 199: I see that I need to remove all this debug logging, or clean it up at least

pkg/controller/util.go
```diff
@@ -211,6 +211,7 @@ func CreateScratchPersistentVolumeClaim(client client.Client, pvc *v1.Persistent
 	scratchPvc := &v1.PersistentVolumeClaim{}
 	if err := client.Get(context.TODO(), types.NamespacedName{Name: scratchPvcSpec.Name, Namespace: pvc.Namespace}, scratchPvc); err != nil {
 		klog.Errorf("Unable to get scratch space pvc, %v\n", err)
+		return nil, err
```
brybacki, line 214: this is interesting, but I did actually catch this a few times while doing manual tests

pkg/controller/import-controller.go
```diff
@@ -246,11 +345,28 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
 	if pod == nil {
 		if isPVCComplete(pvc) {
 			// Don't create the POD if the PVC is completed already
 			log.V(1).Info("PVC is already complete")
 		} else if pvc.DeletionTimestamp == nil {
+			podsUsingPVC, err := getPodsUsingPVCs(r.client, pvc.Namespace, sets.NewString(pvc.Name), false)
```
brybacki, line 354: TODO: ignore completed or failed pvc's

brybacki: Done

pkg/controller/import-controller.go

cluster-sync/sync.sh
```diff
@@ -168,7 +168,7 @@ if [ "${KUBEVIRT_PROVIDER}" != "external" ]; then
   _kubectl apply -f "./_out/manifests/file-host.yaml"
   _kubectl apply -f "./_out/manifests/registry-host.yaml"
   # Imageio test service:
-  _kubectl apply -f "./_out/manifests/imageio.yaml"
+#  _kubectl apply -f "./_out/manifests/imageio.yaml"
```
awels, line 171: Why is this commented, it is needed for the imageio tests.

brybacki: Some experiments in totally unrelated change. 
Restored.

pkg/controller/import-controller.go
awels: For now we can assume that if the PVC has a pod owner, and that pod is owned by a different PVC, that it is a a scratch pvc.

pkg/controller/import-controller.go
```diff
@@ -139,15 +145,96 @@ func NewImportController(mgr manager.Manager, log logr.Logger, importerImage, pu
...
+		client,
+		log,
+	); err != nil {
 		return nil, err
 	}
 	return importController, nil
 }
 
-func addImportControllerWatches(mgr manager.Manager, importController controller.Controller) error {
+func addImportControllerWatches(mgr manager.Manager, importController controller.Controller, c client.Client, log logr.Logger) error {
```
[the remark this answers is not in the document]
brybacki: I do not think I can get anything from Controller interface. But I can get them from reconciler, but passing whole reconciler why I only need client and logger does not look good.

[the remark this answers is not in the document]
awels: Hmmm missed that it returns an interface, which then embeds the Reconciler interface. You are right passed the reconciler when we are only interested in the log and client makes no sense.

awels (commented):
/lgtm

awels:
@mhenriks Can you review here as well, to ensure I didn't suggest something silly with the watches?

pkg/controller/import-controller.go
```diff
@@ -139,15 +146,95 @@ func NewImportController(mgr manager.Manager, log logr.Logger, importerImage, pu
...
+				if evt.MetaNew != nil {
+					q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+						Name:      evt.MetaNew.GetName(),
+						Namespace: evt.MetaNew.GetNamespace(),
+					}})
+				} else {
+					log.Error(nil, "UpdateEvent received with no new metadata", "event", evt)
+				}
+			},
+			DeleteFunc: func(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
```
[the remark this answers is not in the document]
brybacki: I am afraid you are right, this means I need to go back re-implement this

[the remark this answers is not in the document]
brybacki: Ok I've learned a little bit about the CDI controller. So in case of missing the delete callback:

- If the controller restarts it will get the `main` PVC event from the UpdateFunc on first cache synchronization - it can reconcile this correctly. 
- If it just returns error during reconciliation, then I suppose the Reconcile will be retried. 
- If all else fails (event never delivered - is that possible?), then the next re-synchronization will Reconcile the PVC. But the re-synchronization for our cache runs every 10 hours by default.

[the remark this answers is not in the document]
brybacki: Anyway if you still think using a finalizer is more robust then I need to think how to handle this. I can put finalizer while creating the scratch pvc. Then there are a few scenarios possible:

1. Pod is Running, scratchPVC is Bound - the finalizer has to be removed, so it does not block normal deletion. Everything is ok.

2. Scratch was created and someone executed delete on it - before it was bound (but why), it has finalizer so it is marked for termination but still exists!. Pod is Pending but cannot correctly start: `Warning  FailedMount       2m32s (x3 over 5m15s)  kubelet, node02    Unable to attach or mount volumes: unmounted volumes=[cdi-scratch-vol], unattached volumes=[cdi-scratch-vol default-token-4qsdb cdi-data-vol]: error processing PVC cdi-e2e-tests-dv-func-test-dr5kp/scratch-space-delete-scratch: PVC is being deleted`. 

So here the new logic would need to remove the finalizer so PVC can be correctly Deleted, and then recreate the scratchPVC. If create PVC fails (because the old one still exists) it can return error and can be retried.

Is that ok?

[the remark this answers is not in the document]
brybacki: @mhenriks what do you think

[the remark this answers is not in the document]
mhenriks: Okay, if reconciling the target PVC will address the issue, then finalizer is not necessary.  

I would suggest getting rid of these custom handler funcs, though.  Add an annotation to the scratch PVC that has the target PVC name.  When reconciling a PVC, look for that annotation and if it exists, get that PVC and reconcile it.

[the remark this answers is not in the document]
brybacki: I agree, it seems simpler and more explicit than following owner of an owner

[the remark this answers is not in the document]
brybacki: Ok, so I have added a new annotation and simplified the custom handler func. But the func was not removed as it is the only way to access the previous scratch PVC and schedule a PVC for it.  In the reconcile loop on the delete event the scratch PVC is just missing - disappears.

So considering how rare the original problem might be (someone or something trying to delete scratch space just after creation), and that now I cannot recreate the problem in tests. this is in my opinion a good enough solution. 

As mentioned in one of my previous comments, the only problem might be:
"If the DELETE event is never delivered, then the next re-synchronization will Reconcile the PVC, and we need to wait for re-sync."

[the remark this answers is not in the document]
brybacki: changed the main idea, removed the custom handler, now the Reconcile will be rescheduled again and again until pvc is complete

awels (approved):
/lgtm

brybacki:
Is there a possibility there was no nfs Volume available on `pull-containerized-data-importer-e2e-k8s-1.18-nfs`?
```
{
"type": "PodScheduled",
"status": "False",
"lastProbeTime": null,
"lastTransitionTime": "2020-10-29T17:32:55Z",
"reason": "Unschedulable",
"message": "running \"VolumeBinding\" filter plugin for pod \"importer-import-bug\": pod has unbound immediate PersistentVolumeClaims"
}
```

brybacki:
ok, looks like the `pull-containerized-data-importer-e2e-k8s-1.17-hpp` is the new flake, and the `pull-containerized-data-importer-e2e-k8s-1.18-nfs` cannot handle my new scratch-bug test correctly (probably too many pvc's being deleted and created in one test)

brybacki:
/test pull-containerized-data-importer-e2e-k8s-1.17-hpp
the verifier again

brybacki:
looks like it helped,

awels:
Yeah looks like it takes about 20 minutes longer to finish. But I believe I rather have that than constant failures.

pkg/controller/util.go
```diff
@@ -512,6 +513,9 @@ func getPodsUsingPVCs(c client.Client, namespace string, names sets.String, allo
 
 	var pods []v1.Pod
 	for _, pod := range pl.Items {
+		if pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed {
```
awels, line 516: Pod phase should never be Failed because we use a podRestartPolicy of onFailure

tests/utils/pod.go
```diff
@@ -40,8 +40,8 @@ func CreateVerifierPodWithPVC(clientSet *kubernetes.Clientset, namespace string,
 
 // DeleteVerifierPod deletes the verifier pod
 func DeleteVerifierPod(clientSet *kubernetes.Clientset, namespace string) error {
-	zero := int64(0)
-	return DeletePodByName(clientSet, VerifierPodName, namespace, &zero)
+	gracePeriod := int64(30)
+	return DeletePodByName(clientSet, VerifierPodName, namespace, &gracePeriod)
```
awels, line 44: You can just pass nil, and it will use the default grace period.

pkg/controller/util.go
brybacki: it should handle 'other' pods that are not created by cdi

brybacki:
> Yeah looks like it takes about 20 minutes longer to finish. But I believe I rather have that than constant failures.

and I think we can try to improve this in a new PR

PR #1426, opened 2020-10-09
Attempt to schedule clone source/target pods on same node

mhenriks:

This may help clone performance.  Shouldn't make it worse.  Wonder if there's a reason why we haven't done this yet?

Should probably revisit the fact that data is compressed in transit.  Maybe we shouldn't do it at all?  Or only do it when pods are on different nodes?

Fixes #

**Release note**:

```release-note
Try to schedule clone source/target pods on same host
```


awels:
The only reason I can think of, is that maybe we want to influence the target node if we want to move a workload from one node to another. But this change would not stop us from doing that, as it is preferred.

PR #1428, opened 2020-10-12
Fix types.go vs code schema verification to actually fail if they are different

awels:

We checked that the schema matched, however the code that was supposed to generate a failure didn't so even though we checked inconsistencies managed to get in. This PR ensures the job fails so we can't get inconsistent code into the code base.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
Nice, it looks like the reason I have two separate tests for invalid filesystem overhead values is gone

maya-r:
Do you want to mark those as expected fail and I'll fix this in a separate PR?

awels:
Hmmm when I run these locally they all pass....

awels:
Well the thing is the failures are valid, and we should allow blank overhead in the cases that are failing right now. Since it should just pick the global overhead in that case. But they pass for me locally.....

maya-r:
I had assumed that pattern validation doesn't work for the per-storageclass case and checked whether the status gets changed. Your change adds regex validation so the update is rejected

awels:
Now I know why it was passing locally, I was looking at the wrong tests. I fixed the tests.

awels:
@maya-r I fixed the tests, and they are all passing now.

PR #1429, opened 2020-10-12
add finalizer to target PVC before creating clone source pod

mhenriks:

Should address the condition described here:  https://bugzilla.redhat.com/show_bug.cgi?id=1866593#c20

Currently, the finalizer is added to the target PVC after the clone source pod is created.  This means that the source pod may get orphaned if the target PVC is deleted before the finalizer is added

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -190,10 +190,16 @@ func (r *CloneReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 	_, nameExists := pvc.Annotations[AnnCloneSourcePod]
 	if !nameExists && sourcePod == nil {
 		pvc.Annotations[AnnCloneSourcePod] = createCloneSourcePodName(pvc)
+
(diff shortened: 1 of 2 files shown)


pkg/controller/clone-controller.go
```diff
@@ -190,10 +190,16 @@ func (r *CloneReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
...
+		// add finalizer before creating clone source pod
+		pvc = r.addFinalizer(pvc, cloneSourcePodFinalizer)
+
 		if err := r.updatePVC(pvc); err != nil {
 			return reconcile.Result{}, err
 		}
-		return reconcile.Result{Requeue: true}, nil
+
+		// will reconcile again after PVC update notification
+		return reconcile.Result{}, nil
```
mhenriks, line 202: This may keep us from getting some of the "conflict errors seen in the controller log.

PR #1439, opened 2020-10-13
Don't wait for NS to deleted in test before starting next test

awels:

During each test run, we wait until the NS of the current test is fully deleted before moving to the next test. This PR we just delete the NS, and continue to the next test while letting the system clean up the old ns. In theory this should shorten the runtime of a lane.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -13,7 +13,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	v1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
@@ -40,7 +39,6 @@ const (
 )
 
 var _ = Describe("all clone tests", func() {
-
 	var _ = Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:component]Cloner Test Suite", func() {
 		f := framework.NewFramework(namespacePrefix)
 
@@ -105,12 +103,8 @@ var _ = Describe("all clone tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			By("Deleting verifier pod")
-			err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), utils.VerifierPodName, metav1.DeleteOptions{})
+			err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
 			Expect(err).ToNot(HaveOccurred())
-			Eventually(func() bool {
-				_, err := f.K8sClient.CoreV1().Pods(f.Namespace.Name).Get(context.TODO(), utils.VerifierPodName, metav1.GetOptions{})
-				return k8serrors.IsNotFound(err)
-			}, 60, 1).Should(BeTrue())
 			completeClone(f, f.Namespace, targetPvc, diskImagePath, sourceMD5, sourcePvcDiskGroup)
 		})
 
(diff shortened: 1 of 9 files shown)


awels:
/hold
Need to verify at the end that all the NS are gone, but want to see if there is a significant improvement in the time it takes to run a lane.

awels:
/hold cancel
It saves around 1 hour which is significant enough to proceed. I added code to verify all the test namespaces are removed at the end.

maya-r:
> Removed test for 'failed' state that can no longer happen

I don't understand why it can no longer happen. Is this from the current change?

Also: we can probably remove the definition of invalidEndpoint, too.

awels:
Because we have our containers set to restartOnFailure, so the pods will never go into a failed state.

awels:
/hold
adding hold again, some test failures I need to investigate after some of the clone and upload tests have retries.

brybacki:
Interesting, in both tests:
Cloner Test Suite [test_id:4277] Clone datavolume with long name 
Cloner Test Suite [test_id:4278] Clone datavolume with long name including special character '.'

The resulting content has different checksum. I think we already had data integrity problem that was fixed?

awels:
Yeah I think we might have to add some flag to wait for ns deletion on some tests. In particular this seems to occur with NFS storage.

brybacki:
maybe it would be better to create a new `source` namespace, the same way the `target` namespace is created for each test

awels:
So I think the problem is that we have 5 NFS PVs, and they get cleaned up, and recreated at the end of each test. This happens right before the NS delete. And before we would wait for the NS to get deleted, which gave the system some time to fully clean. But here we immediately go to the next test, which uses the same 5 PVs potentially. And it is possible the clean-up wasn't 100% completed yet. Its basically a race on the clean-up. So I am going to try more PVs, and rotate through a range of them on each test. Lets say 15 PVs, and go 1-5 then 6 to 10, then 11 to 15, and then back to 1-5. This will give the clean-up 2 tests worth of time to actually finish.

tests/cloner_test.go
```diff
@@ -105,7 +107,9 @@ var _ = Describe("all clone tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			By("Deleting verifier pod")
-			err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), utils.VerifierPodName, metav1.DeleteOptions{})
+			err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), utils.VerifierPodName, metav1.DeleteOptions{
+				GracePeriodSeconds: &zero,
```
brybacki, line 111: Why not add DeletePodImmediately to utils in tests/utils/pod.go, then It can also be used here:
tests/framework/pvc.go:107 in createConsumerPod and maybe some other places in the future

awels: Yeah I thought about that. At this point I was trying to get NFS stable and happy. If that is true, then I will probably add a function like that, as I think there are quite a few places that could use this.

awels:
@brybacki Updated the code to use a DeletePodNoGrace, and implemented it for all verifier pods.

tests/utils/pod.go
```diff
@@ -77,19 +79,27 @@ func CreatePod(clientSet *kubernetes.Clientset, namespace string, podDef *k8sv1.
 }
 
 // DeletePodByName deletes the pod based on the passed in name from the passed in Namespace
-func DeletePodByName(clientSet *kubernetes.Clientset, podName, namespace string) error {
+func DeletePodByName(clientSet *kubernetes.Clientset, podName, namespace string, gracePeriod *int64) error {
```
arnongilboa, line 82: What's the difference between passing gracePeriod of int64(0) or nil here? nil means some default value?

awels: Yes, nil means use the default value.

PR #1442, opened 2020-10-14
Add library function to determine if a PVC is waiting for first consu…

brybacki:

Add library function to determine if a PVC is waiting for first consumer before binding (and before it can be populated).

IsWaitForFirstConsumerBeforePopulating indicates if the persistent volume passed in is in ClaimPending state and waiting for first consumer. It follows the logic:
1. If the PVC is not owned by a DataVolume, return false, we can not assume it will be populated
2. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase WaitForFirstConsumer return true
3. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase !WaitForFirstConsumer return false

This is needed so the users of CDI (kubevirt) can detect and correctly handle PVC's in this state. See https://github.com/kubevirt/kubevirt/pull/4025

Fixes #

**Release note**:

```release-note
A new library function IsWaitForFirstConsumerBeforePopulating  
```

### DIFF
--- a/pkg/apis/core/v1alpha1/BUILD.bazel
+++ b/pkg/apis/core/v1alpha1/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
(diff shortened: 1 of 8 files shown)


awels:
looks good, but I would like some unit tests. Similar to these: https://github.com/kubevirt/containerized-data-importer/pull/1400/files#diff-b467b5dba3091841c7ee7bdcd589408217fa09bdef49ea0ab9cef01dbb07b197R557-R618

brybacki:
> looks good, but I would like some unit tests. Similar to these: https://github.com/kubevirt/containerized-data-importer/pull/1400/files#diff-b467b5dba3091841c7ee7bdcd589408217fa09bdef49ea0ab9cef01dbb07b197R557-R618

I have added simple unit test, but I wonder if this is the good place. Do we have anything against adding tests to api packages, or is it not there because so far API did not have any logic. I can move this test there to API. 

Anyway now we have some duplication in APIs. It is the same util in just two places so maybe it is not yet worth adding 'commons' but it will make tests duplicated also.

awels:
So the problem is that v1alpha1 DVs != v1beta1 DVs, so you can't write one unit test to exercise both functions. The duplication kills me as well, but I see no alternative.

awels:
Looks like the unit tests are failing?

brybacki:
> Looks like the unit tests are failing?

I did push the wip by mistake, now fixed.

awels (approved):
/lgtm
/approve

PR #1446, opened 2020-10-16
Remove the "Process" data processor phase, simplify state machine.

maya-r:
This phase can mean three things:
HTTP, imageio, s3, upload: Go directly to Convert
VDDK: unreachable code, points to arbitrary other phase.
registry: do minor processing after transfer.

Only registry makes actual use of this phase.

Point all current users directly to Convert, and fold the
work done in the registry Process phase into the previous
phase.

I did some reading of the state diagram for the purpose of adding an unconditional validation, it was pretty complicated and this seems like a possible simplification. Let me know if you disagree. 

**Release note**:

```release-note
NONE
```


awels (approved):
/lgtm
/approve

awels:
Agreed, the processing phase was not adding value, so simpler state machine is better.

Issue #1406, opened 2020-09-29
Leverage checkpoints in VDDK DataSource


fdupont-redhat:
/kind enhancement

When the importer pod is created with checkpoints environment variables, the DataSource will use govmomi to query changed disk areas between the two checkpoints, i.e. snapshots, then `libnbd` to retrieve only these changed disk areas.

PR #1448, opened 2020-10-19
VDDK: replace qemu-img with libnbd

### LINKED ISSUE #1405: Use libnbd instead of qemu-img for NBD-based volume sources.
/kind enhancement

With source volumes that are exported over NBD, it is easier to take advantage of transfer optimizations using libnbd instead of qemu-img.

Right now the only source that works this way is the VDDK source for VMware disks. The initial implementation used qemu-img because of issues building the Go bindings of libnbd, but these issues now have reasonable solutions.

This is likely also required for warm migrations from VMware.

mrnold:

This pull request adds libnbd to the CDI build and uses it to replace qemu-img in the VDDK datasource. This change should not make any immediate functional difference, but it is useful groundwork as libnbd is likely the best way to implement changed-block copies for the incoming warm migration feature (as in #1406). This was the original plan for #1320, but the initial implementation used qemu-img as a temporary measure to get around the difficulties I had including libnbd in CDI.

Fixes #1405

* This requires another new build container image to pick up libnbd-devel.
* This enables CGO for all of CDI. This is required by the Go bindings for libnbd, but I don't know if it will cause trouble in other components. I haven't found another way to do this.

**Release note**:

```release-note
NONE
```


pkg/common/common.go
```diff
@@ -151,4 +151,55 @@ const (
 	VddkConfigMap = "v2v-vmware"
 	// VddkConfigDataKey is the name of the ConfigMap key of the VDDK image reference
 	VddkConfigDataKey = "vddk-init-image"
+
```
mhenriks, line 154: curious why this stuff was moved?

mrnold: This was the easiest way I could find to break a chain of dependencies that was causing libnbd to get linked in to pretty much every other binary. I think the problem was that pkg/uploadserver depends on pkg/importer, and a few other packages depend on pkg/uploadserver for some of the constants I moved. The end result was that any binary that used those constants had to link to libnbd to satisfy the indirect dependency on pkg/importer, even tests.test.

There are probably other ways to break up the dependencies so this doesn't happen, but this was the least disruptive option I found so far.

mhenriks: cool thanks for the info

mrnold:
CI will fail right off the bat without an update to the build container image.

awels:
#1452 will update the builder, you will have to update the hash after that merges.

awels:
@mrnold The sha for the new bulder: sha256:b12b8f7a1b0c0cebe9985ed93dcdc5a04784e88ac7c982b7317d0b96b4f7526f

mrnold:
Thanks!

awels:
@mrnold Do you need a review on this?

mrnold:
@awels Yes, I think this is ready for review now.

pkg/importer/vddk-datasource.go
```diff
@@ -53,18 +60,65 @@ func getVddkPluginPath() string {
...
+	Close()
+}
+
+// VDDKFileSink writes the source disk data to a local file.
+type VDDKFileSink struct {
+	file   *os.File
+	writer *bufio.Writer
+}
+
+func (sink VDDKFileSink) Write(buf []byte) (int, error) {
```
awels, line 82: I believe convention wants a pointer, so
```go
func (sink *VDDKFileSink) Write(buf []byte) (int, error) {
```
I would have to look up what the actual practical differences are

pkg/importer/vddk-datasource.go
```diff
@@ -53,18 +60,65 @@ func getVddkPluginPath() string {
...
+	written, err := sink.writer.Write(buf)
+	if err != nil {
+		return written, err
+	}
+	err = sink.writer.Flush()
+	return written, err
+}
+
+// Close closes the file after a transfer is complete.
+func (sink VDDKFileSink) Close() {
```
awels, line 92: Same sink* VDDKFileSink

pkg/importer/vddk-datasource.go
```diff
@@ -244,10 +338,62 @@ func (vs *VDDKDataSource) Transfer(path string) (ProcessingPhase, error) {
...
-		klog.Infof("Failed to convert disk: %s\n", err)
+		klog.Errorf("Unable to get size from libnbd handle: %v", err)
 		return ProcessingPhaseError, err
 	}
+
+	sink, err := newVddkDataSink(destinationFile, size)
+	if err != nil {
+		return ProcessingPhaseError, err
+	}
+	defer sink.Close()
```
awels, line 351: This is just curiousity on my part, does the sink have like a flush or something at the end to ensure that the data has been written to the actual storage device, and not stuck somewhere in a cache somewhere? I ask because we had a bug where we had RWX pvc, and tried to immediately start the VM on a different node, and not everything was written to the device, and starting the VM failed due to a corrupted disk.

awels (commented):
Looks good a few minor questions.

pkg/importer/vddk-datasource.go
mrnold: Okay, I changed this.

pkg/importer/vddk-datasource.go
mrnold: This one too.

pkg/importer/vddk-datasource.go
mrnold: That's a good idea, I added a final file.Sync in the normal implementation (not the test mock one) of sink.Close.

fdupont-redhat:
@awels are you fine with merging this PR?

awels (approved):
/lgtm
/approve

Issue #1407, opened 2020-09-29
Add support for multi-stage Import DataVolume


fdupont-redhat:
/kind enhancement

In order to support warm import, the DataVolume CRD needs to support new fields:

- a list of checkpoints that represent the intervals of data to copy.
- a boolean that tells whether the current checkpoint is the last.

A new phase of the DataVolume CRD needs to be added to be used between two iterations.

The last and second-to-last checkpoints, as well as the last checkpoint boolean are set as annotations on the PVC and removed once the iteration is finished. They are be passed to the Importer pod that will then leverage them.

PR #1450, opened 2020-10-20
Controller support for Multistage Imports

mansam:

This PR adds controller support that is needed for multistage imports.

* Add Previous, Current, and FinalCheckpoint to DataVolume CRD
* Use checkpoints from the DV to set annotations on the PVC
* If an importer pod succeeds while checkpoint annotations are set,
  then set the DataVolume status to Paused instead of Succeeded.
* Sets the Importer pod environment variables from the checkpoint annotations
* Clear the PVC checkpoint annotations when a stage completes

Fixes https://github.com/kubevirt/containerized-data-importer/issues/1407

**Release note**:

```release-note
NONE
```


awels:
if you rebase most of the flaky failures should go away.

awels:
Also, I am missing something here, I see the new fields and annotations, however I don't really see code using them. Is that a follow up PR?

mansam:
@awels Yes, this will require a follow up PR to enable the importer pod to make use of these annotations to perform warm imports.

mansam:
Rebased on master.

PR #1451, opened 2020-10-21
Add make targets cluster-sync-cdi & cluster-sync-test-infra

arnongilboa:

CDI developer would prefer not to redeploy test containers on each cluster-sync. Currently each time we cluster-sync, it un-installs all of CDI, including test containers, and then re-installs newly compiled CDI containers, and then re-installs the test containers. The test containers don't change often, so it makes sense to not always reinstall them. This is helpful when you are developing and  not interested in the tests. It saves a lot of time.

**Release note**:
```NONE```

### DIFF
--- a/cluster-sync/sync.sh
+++ b/cluster-sync/sync.sh
@@ -111,6 +111,20 @@ function wait_cdi_pods_updated {
   fi
 }
 
+# Start functional test HTTP server.
+# We skip the functional test additions for external provider for now, as they're specific
+if [ "${KUBEVIRT_PROVIDER}" != "external" ] && [ "${CDI_SYNC}" == "test-infra" ]; then
+  _kubectl delete all -n cdi -l cdi.kubevirt.io/testing
+  _kubectl apply -f "./_out/manifests/bad-webserver.yaml"
+  _kubectl apply -f "./_out/manifests/file-host.yaml"
+  _kubectl apply -f "./_out/manifests/registry-host.yaml"
+  # Imageio test service:
+  _kubectl apply -f "./_out/manifests/imageio.yaml"
+  # vCenter (VDDK) test service:
+  _kubectl apply -f "./_out/manifests/vcenter.yaml"
+  exit 0
+fi
+
 mkdir -p ./_out/tests
 rm -f $OLD_CDI_VER_PODS $NEW_CDI_VER_PODS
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by maya-r** [comment]
I'm not sure why there are three make targets now (with rules). I feel like there should be one which just depends on the other two,

cluster-sync: cluster-sync-cdi cluster-sync-test-infra

**Comment by arnongilboa** [comment]
> I'm not sure why there are three make targets now (with rules). I feel like there should be one which just depends on the other two,
> 
> cluster-sync: cluster-sync-cdi cluster-sync-test-infra

**Comment by arnongilboa** [comment]
@maya-r fixed

File: cluster-sync/sync.sh [hunk 1]
```diff
@@ -111,6 +112,20 @@ function wait_cdi_pods_updated {
   fi
 }
 
+# Start functional test HTTP server.
+# We skip the functional test additions for external provider for now, as they're specific
+if [ "${KUBEVIRT_PROVIDER}" != "external" ] && [ "${CDI_SYNC}" == "test-infra" ]; then
```
awels, line 117: Shouldn't we do this when CDI_SYNC is "all" like the default in case CDI_SYNC is somehow not set. Or we should default it to test-infra.

arnongilboa:
"all" was removed thanks to Maya's comment (see the 2nd patch).

awels:
Right, but if no one specifies CDI_SYNC, it currently defaults to "all" and the code as written will then NOT remove all. So it is confusing.

arnongilboa:
Right, it's my mistake/leftover. Default should better be "" and we may
also remove the  CDI_SYNC="cdi" in cluster-sync-cdi rule which is not
needed anymore.
Actually moving the test-infra case to a new sync-test-infra.sh might be
even cleaner as we won't need to pass the env variable  at all. What do you
think?

hack/build/cluster-sync-test.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
maya-r, line 1: I think we would be better off not committing this file. It might discourage anyone who wants to make further changes to the sync code by needing to understand what it does.

arnongilboa:
I don't think that's an issue. As its filename says, it's only a simple
test for timing the different cluster-sync flavors and we better have in
the repo for future 'make' timings.

PR #1461, opened 2020-10-31
Add an API for disabling smart-cloning.

maya-r:
We used to detect the possibility of smart-cloning and always use it
if it's there. This might not be the desirable behaviour if:
- Snapshots cost more money than a host-assisted clone
- Snapshots are broken

The API is:
kubectl edit cdi
cdi.Spec.cloneStrategyOverride = "Host"

We use an enum rather than a boolean for this in an attempt to future
proof if more strategies are available, but only the value that is
currently necessary is used.

If no value is defined, we continue with the existing behaviour of
preferring smart clone if possible.

I've only defined a value in the spec portion of the CDI CR, and not in the status, or in CDIConfig.
Avoiding the config controller feels like it might be the right move, but we have spoken about this enough that I'm not sure.

**Release note**:

```release-note
Add a tunable to disable smart-clone and always prefer host-assisted copy.
CDI.spec.cloneStrategyOverride: "copy"
If this value is not defined to anything, the existing behaviour is unchanged. That is, smart-clone when possible.
```

### DIFF
--- a/pkg/apis/core/v1alpha1/types.go
+++ b/pkg/apis/core/v1alpha1/types.go
@@ -282,9 +282,20 @@ type CDISpec struct {
 	Infra sdkapi.NodePlacement `json:"infra,omitempty"`
 	// Restrict on which nodes CDI workload pods will be scheduled
 	Workloads sdkapi.NodePlacement `json:"workload,omitempty"`
+	// Clone strategy override: should we use a host-assisted copy even if snapshots are available?
+	// +kubebuilder:validation:Enum="copy";"snapshot"
+	CloneStrategyOverride *CDICloneStrategy `json:"cloneStrategyOverride,omitempty"`
 	// CDIConfig at CDI level
 	Config *CDIConfigSpec `json:"config,omitempty"`
 }
+
+// CDICloneStrategy defines the preferred method for performing a CDI clone (override snapshot?)
+type CDICloneStrategy string
+
+const (
+	// CloneStrategyHostAssisted specifies slower, host-assisted copy
+	CloneStrategyHostAssisted = "copy"
+)
 
 // CDIUninstallStrategy defines the state to leave CDI on uninstall
 type CDIUninstallStrategy string
--- a/pkg/apis/core/v1alpha1/types.go
+++ b/pkg/apis/core/v1alpha1/types.go
@@ -295,6 +295,9 @@ type CDICloneStrategy string
 const (
 	// CloneStrategyHostAssisted specifies slower, host-assisted copy
 	CloneStrategyHostAssisted = "copy"
+
+	// CloneStrategySnapshot specifies snapshot-based copying
+	CloneStrategySnapshot = "snapshot"
 )
 
 // CDIUninstallStrategy defines the state to leave CDI on uninstall
--- a/pkg/apis/core/v1beta1/types.go
+++ b/pkg/apis/core/v1beta1/types.go
@@ -284,9 +284,23 @@ type CDISpec struct {
 	Infra sdkapi.NodePlacement `json:"infra,omitempty"`
 	// Restrict on which nodes CDI workload pods will be scheduled
 	Workloads sdkapi.NodePlacement `json:"workload,omitempty"`
+	// Clone strategy override: should we use a host-assisted copy even if snapshots are available?
+	// +kubebuilder:validation:Enum="copy";"snapshot"
+	CloneStrategyOverride *CDICloneStrategy `json:"cloneStrategyOverride,omitempty"`
 	// CDIConfig at CDI level
 	Config *CDIConfigSpec `json:"config,omitempty"`
 }
+
+// CDICloneStrategy defines the preferred method for performing a CDI clone (override snapshot?)
+type CDICloneStrategy string
+
+const (
+	// CloneStrategyHostAssisted specifies slower, host-assisted copy
+	CloneStrategyHostAssisted = "copy"
+
+	// CloneStrategySnapshot specifies snapshot-based copying
+	CloneStrategySnapshot = "snapshot"
+)
 
 // CDIUninstallStrategy defines the state to leave CDI on uninstall
 type CDIUninstallStrategy string
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -277,7 +277,7 @@ func (r *DatavolumeReconciler) Reconcile(req reconcile.Request) (reconcile.Resul
 			return reconcile.Result{}, err
 		}
 		snapshotClassName, err := r.getSnapshotClassForSmartClone(datavolume)
-		if (err == nil) && (cloneStrategy == cdiv1.CloneStrategySnapshot) {
+		if err == nil && cloneStrategy == cdiv1.CloneStrategySnapshot {
 			r.log.V(3).Info("Smart-Clone via Snapshot is available with Volume Snapshot Class", "snapshotClassName", snapshotClassName)
 			if requeue, err := r.sourceInUse(datavolume); requeue || err != nil {
 				return reconcile.Result{Requeue: requeue}, err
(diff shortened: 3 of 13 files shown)


pkg/apis/core/v1alpha1/types.go
```diff
@@ -268,8 +268,19 @@ type CDISpec struct {
 	Infra sdkapi.NodePlacement `json:"infra,omitempty"`
 	// Restrict on which nodes CDI workload pods will be scheduled
 	Workloads sdkapi.NodePlacement `json:"workload,omitempty"`
+	// Clone strategy override: should we should we use a host-assisted copy even if snapshots are available?
```
awels, line 271: double should we in description

pkg/apis/core/v1alpha1/types.go
```diff
@@ -268,8 +268,19 @@ type CDISpec struct {
...
+	// +kubebuilder:validation:Enum="copy"
+	CloneStrategyOverride *CDICloneStrategy `json:"cloneStrategyOverride,omitempty"`
 }
 
+// CDICloneStrategy defines the preferred method for performing a CDI clone (override snapshot?)
+type CDICloneStrategy string
+
+const (
+	// CloneStrategyHostAssisted specifies slower, host-assisted copy
+	CloneStrategyHostAssisted = "copy"
```
awels, line 281: Even though it is the default, I would add the smart clone option as well. I know it will not actually do anything, but having everything listed somewhere will help people understand what we are doing.

pkg/apis/core/v1beta1/types.go
```diff
@@ -270,8 +270,19 @@ type CDISpec struct {
 	Infra sdkapi.NodePlacement `json:"infra,omitempty"`
 	// Restrict on which nodes CDI workload pods will be scheduled
 	Workloads sdkapi.NodePlacement `json:"workload,omitempty"`
+	// Clone strategy override: should we should we use a host-assisted copy even if snapshots are available?
```
awels, line 273: Double should we in description

pkg/apis/core/v1beta1/types.go
```diff
@@ -270,8 +270,19 @@ type CDISpec struct {
...
+	// +kubebuilder:validation:Enum="copy"
+	CloneStrategyOverride *CDICloneStrategy `json:"cloneStrategyOverride,omitempty"`
 }
 
+// CDICloneStrategy defines the preferred method for performing a CDI clone (override snapshot?)
+type CDICloneStrategy string
+
+const (
+	// CloneStrategyHostAssisted specifies slower, host-assisted copy
+	CloneStrategyHostAssisted = "copy"
```
awels, line 283: Even though it is the default, I would add the smart clone option as well. I know it will not actually do anything, but having everything listed somewhere will help people understand what we are doing.

pkg/controller/datavolume-controller.go
```diff
@@ -439,6 +447,20 @@ func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.D
 	return "", errors.New("could not match snapshotter with storage class, falling back to host assisted clone")
 }
 
+func preferHostAssistedCopy(c client.Client) (bool, error) {
```
awels, line 450: I think we should do it a little more future proof. We are expecting to add more options in the future, so it would make sense to have a getPreferredCloneStrategy function that returns the value. Also this is not a generic function used by all controllers, so you can make part of the reconciler so you don't have to pass the client.
Something like
```go
func (r *DatavolumeReconciler) getPreferrredCloneStrategy() (CDICloneStrategy, error) {
...
}
```

pkg/controller/datavolume-controller.go
```diff
@@ -367,6 +367,14 @@ func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.D
 		return "", errors.New("no source PVC provided")
 	}
 
+	preferHostAssisted, err := preferHostAssistedCopy(r.client)
```
awels, line 370: This seems like the wrong spot to check for this. The function is called getSnapshotClassForSmartClone, so we are interested in getting the snapshot class if it exists, regardless of the preference. I would check in whatever function calls getSnapshotClassForSmartClone similar to this:
```go
preferredStrategy, err := c.getPreferrredCloneStrategy()
if err != nil {
  return err
}
if preferredStrategy == "smart" {
   // whatever was calling getSnapshotClassForSmartClone
}
```

pkg/operator/resources/operator/operator.go
```diff
@@ -1512,6 +1512,15 @@ func createCDIListCRD() *extv1.CustomResourceDefinition {
 												},
 											},
 										},
+										"cloneStrategyOverride": {
+											Type:        "string",
+											Description: "Clone strategy override: should we should we use a host-assisted copy even if snapshots are available?",
```
awels, line 1517: Duplicate should we in description

pkg/operator/resources/operator/operator.go
```diff
@@ -2750,6 +2759,15 @@ func createCDIListCRD() *extv1.CustomResourceDefinition {
...
```
awels, line 2764: Duplicate should we in description

maya-r:
this is a very innocent commit - it should only add 1 test for the ceph lane, and otherwise make little difference.  I wonder whether we are affected by the flakiness that Roman mentioned on kubevirt-dev

maya-r:
rebased to fix conflict

pkg/controller/datavolume-controller.go
```diff
@@ -258,8 +258,12 @@ func (r *DatavolumeReconciler) Reconcile(req reconcile.Request) (reconcile.Resul
 	}
 
 	if !pvcExists {
+		cloneStrategy, err := r.getCloneStrategy()
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 		snapshotClassName, err := r.getSnapshotClassForSmartClone(datavolume)
-		if err == nil {
+		if (err == nil) && (cloneStrategy == cdiv1.CloneStrategySnapshot) {
```
awels, line 266: Any reason for the ( .. ) && ( .. ) the equals have higher priority so there is no risk it won't work without the parenthesis.

pkg/controller/datavolume-controller.go
```diff
@@ -439,6 +443,21 @@ func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.D
 	return "", errors.New("could not match snapshotter with storage class, falling back to host assisted clone")
 }
 
+func (r *DatavolumeReconciler) getCloneStrategy() (cdiv1.CDICloneStrategy, error) {
+	crList := &cdiv1.CDIList{}
```
awels, line 447: All of this code is duplicated from pkg/controller/util.go from the GetWorkloadNodePlacement function. Can we make function in util called: GetCDICR? Something like:
```go
func GetCDICR(c client.Client) (*cdiv1.CDI, error) {
	crList := &cdiv1.CDIList{}
	if err := c.List(context.TODO(), crList, &client.ListOptions{}); err != nil {
		return nil, err
	}
	if len(crList.Items) != 1 {
		return nil, fmt.Errorf("Number of CDI CRs != 1")
	}
	return &crList.Items[0], nil
}
```
And then calling that from all places we need to get the CDI CR?

awels (changes requested):

pkg/apis/core/v1alpha1/types.go
awels: Looks like you added it to the betav1 version, just not the alphav1. Other than that it looks good.

awels:
Other than missing the 'snapshot' option in the v1alpha1 API, looks good. Once you get that I will give lgtm.

maya-r:
oops, rebased on top of master for no good reason.

awels (approved):
looks good, I wish there were some unit tests for the getCloneStrategy function in datavolume-controller.

PR #1462, opened 2020-11-01
Simplify the HTTP file host

maya-r:
Eliminate the part of the init container where we copy the images
to /tmp/shared/images, instead have the http container built with
the images into it.

This eliminates several places necessary to modify when adding
an image, an especially painful process until you remember that
the file-host container uses imagePullPolicy: IfNotPresent.

This step also created the tinyCore.qcow2 file using the RAW file,
so we need to add it to the repository. A step like this is useful
to save space, but raw is the fattest format, so we might be better
off reversing it if we want to re-introduce it.
Additionally, it's only used in practice for this one tinyCore.qcow2
image.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -79,6 +79,8 @@ filegroup(
     name = "test-images",
     srcs = [
         "//tests:images/tinyCore.iso",
+        "//tests:images/tinyCore.qcow2",
+        "//tests:images/tinyCore.qcow2.gz",
         "//tests:images/archive.tar",
         "//tests:images/cirros-qcow2.img",
         "//tests:images/cirros.raw",
--- a/hack/build/build-cdi-func-test-file-host.sh
+++ b/hack/build/build-cdi-func-test-file-host.sh
@@ -32,7 +32,5 @@ DOCKER_PREFIX=""
 ${BUILD_DIR}/build-copy-artifacts.sh "${FILE_INIT_PATH}"
 
 cp ${BUILD_DIR}/docker/${FILE_HOST}/* ${OUT_PATH}/${FILE_HOST}/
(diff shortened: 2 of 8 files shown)


maya-r:
I can probably avoid adding tinyCore.qcow2 by switching the tests relying on it to the cirros qcow2, too.

brybacki:
the test failures you see here where fixed recently (Tuesday, 3rd of Nov), if you want a full green run then better rebase

aglitke (commented):
I know this isn't caused by your PR but I wonder why we keep gzipped and xzipped versions of the images in our repo when we could quickly "build them".  Seems like a waste that is bloating our repository.

maya-r:
I am currently removing the change that builds them.
I don't have a problem with them being built, but every time I interacted with adding an image to the file host I had to add the name in 3 different places and remember to `make cluster-down` when testing, because it uses IfNotPresent as an imagePullPolicy.

maya-r:
To be more persuasive: after this change, to add an image to the file host, you only need to modify two BUILD.bazel files (the toplevel one, and the tests one)

awels:
So the question here is really which one do we like better:
1. Build the images on the fly when starting the file host container, and have less images in the repo. But when/if you have to add a new image it is harder to add a new image, and that image will then make the startup of the file host longer because it is now building all the permutations for that image.
2. Add some images to the repo, (in this case about 40M worth of images) and have the file host start a little faster because it doesn't have the build all the permutations. And when/if you have to add an image it will increase the size in the repo more if you want more than one permutation, but its slightly easier to add.

As for the pullPolicy, the number of times you add images to the file host is really small, so to optimize for that case seems strange to me. Otherwise we will be pulling the image every time regardless of if there are any changes.

maya-r:
closing in favour of #1534 which is less invasive.

PR #1466, opened 2020-11-04
Validate image fits in filesystem in a lot more cases. take filesystem overhead into account when resizing.

maya-r:
One of the common paths in the importer didn't validate image size.
Add a call to validation there.
Resizing didn't take filesystem overhead into account.

Adds a lot of missing test coverage for importing/uploading too large images.

https://bugzilla.redhat.com/show_bug.cgi?id=1827793

**Release note**:
```release-note

```


maya-r:
I've been having a hard time doing a full test run on my machine. Tests time out after an hour or so of running.
This will be the first full test run, so there might be some missed failures.

pkg/importer/data-processor.go
```diff
@@ -264,16 +264,32 @@ func (dp *DataProcessor) convert(url *url.URL) (ProcessingPhase, error) {
 }
 
 func (dp *DataProcessor) resize() (ProcessingPhase, error) {
-	// Resize only if we have a resize request, and if the image is on a file system pvc.
 	size, _ := getAvailableSpaceBlockFunc(dp.dataFile)
 	klog.V(3).Infof("Available space in dataFile: %d", size)
+	isBlockDev := size >= int64(0)
```
arnongilboa, line 269: // size will be int64(-1) for non-block dev

pkg/importer/data-processor.go
```diff
@@ -264,16 +264,32 @@ func (dp *DataProcessor) convert(url *url.URL) (ProcessingPhase, error) {
...
-	if dp.requestImageSize != "" && size < int64(0) {
-		var err error
-		klog.V(3).Infoln("Resizing image")
-		shouldPreallocate, err = ResizeImage(dp.dataFile, dp.requestImageSize, dp.availableSpace)
-		if err != nil {
-			return ProcessingPhaseError, errors.Wrap(err, "Resize of image failed")
+	if !isBlockDev {
+		// Unlike block devices, a raw image file on a disk doesn't fill up the
+		// file system space as needed (thus we resize it)
+		// And may be bigger than the available space (sparse).
```
arnongilboa, line 274: doesn't fill up as as needed == sparse?

pkg/importer/data-processor.go
```diff
@@ -264,16 +264,32 @@ func (dp *DataProcessor) convert(url *url.URL) (ProcessingPhase, error) {
...
+		if dp.requestImageSize != "" {
+			var err error
+			klog.V(3).Infoln("Resizing image")
+			shouldPreallocate, err = ResizeImage(dp.dataFile, dp.requestImageSize, dp.availableSpace)
+			if err != nil {
+				return ProcessingPhaseError, errors.Wrap(err, "Resize of image failed")
+			}
+		} else {
+			// A sparse file could still be bigger than the available space.
+			dataFileURL, err := url.Parse(dp.dataFile)
```
arnongilboa, line 286: how is the check related to the comment?

tests/import_test.go
```diff
@@ -48,6 +49,45 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+			log, _ := tests.RunKubectlCommand(f, "logs", importer.Name, "-n", importer.Namespace)
+			if strings.Contains(log, "is larger than available size") {
+				return true
+			}
+			if strings.Contains(log, "no space left on device") {
+				return true
+			}
+			if strings.Contains(log, "qemu-img execution failed") {
+				return true
+			}
```
arnongilboa, line 79: why accept any of the errors? qemu-img execution failure seems irrelevant to your issue.

maya-r: I'm not 100% sure it is the right case, but I think this is the failure mode when trying to import to a block volume. I add no validation for that case, and that's because block devices innately do the validation we expect - you won't be allowed to make an image that appears larger than the actual storage space.

tests/import_test.go
```diff
@@ -48,6 +49,45 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		ns = f.Namespace.Name
 	})
 
+	DescribeTable("[test_id:2329] Should fail to import images that require too much space", func(uploadURL string) {
+		imageURL := fmt.Sprintf(uploadURL, f.CdiInstallNs)
+
+		By(imageURL)
+		dv := utils.NewDataVolumeWithHTTPImport("too-large-import", "500Mi", imageURL)
```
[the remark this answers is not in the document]
maya-r: they are all 2GB :)

BUILD.bazel
```diff
@@ -82,6 +82,10 @@ filegroup(
         "//tests:images/archive.tar",
         "//tests:images/cirros-qcow2.img",
         "//tests:images/cirros.raw",
+	"//tests:images/cirros-large-virtual-size.raw.xz",
+	"//tests:images/cirros-large-virtual-size.qcow2",
+	"//tests:images/cirros-large-physical-size.raw.xz",
+	"//tests:images/cirros-large-physical-size.qcow2",
```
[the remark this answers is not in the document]
maya-r: Not sure why we have to repeat it twice, might take a look at it eventually.

pkg/importer/data-processor.go
maya-r: I am describing what this block does, would it be more clear if I added a space between the code and the comment?

arnongilboa: "A sparse file could still be bigger than the available space" is describing what this block does?

BUILD.bazel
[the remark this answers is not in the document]
arnongilboa: seen the same 4 suffixes x 4 times

pkg/importer/data-processor.go
maya-r: Nope - I mean that QEMU will respect the image file size and not make it bigger as necessary.
[I refactored this bit of code specifically so it requires far fewer comments to explain why things are done](https://github.com/kubevirt/containerized-data-importer/pull/1466/commits/a29e9f9518e83a7ce6bd5920e74e070c713a6667)

pkg/importer/data-processor.go
maya-r: yeah, I left that comment out in favour of using a descriptive name for this condition :-)

pkg/importer/data-processor.go
maya-r: After some looking at this code block, I rewrote it so it requires less comments to explain. Hopefully I didn't uncover some edge case where it fails a test!

pkg/importer/data-processor_test.go
```diff
@@ -311,9 +311,12 @@ var _ = Describe("Resize", func() {
 			url: url,
 		}
 		dp := NewDataProcessor(mdp, "dest", "dataDir", "scratchDataDir", "", 0.055, false)
-		nextPhase, err := dp.resize()
-		Expect(err).ToNot(HaveOccurred())
-		Expect(ProcessingPhaseComplete).To(Equal(nextPhase))
+		qemuOperations := NewFakeQEMUOperations(nil, nil, fakeInfoOpRetVal{&fakeZeroImageInfo, nil}, nil, nil, nil)
```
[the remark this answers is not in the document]
maya-r: I think I needed this change to make the unit tests pass.
if I remember correctly, this unit test was failing because I am indirectly adding a call to a qemu function ( within resize function, so now it needs to be able to call qemu functions and get a valid-ish result.
resize->dp.validate()->qemuOperations.Validate.

pkg/importer/data-processor.go
arnongilboa: so now you parse and validate also for dp.requestImageSize != "" ?

maya-r: yes. we can probably just validate it and it will pass after the resize - easier than explaining why it's not strictly necessary to do it any more

maya-r:
I was having some doubts about my changes like: "it's a little magical to take filesystem overhead when an image size of X is requested, and only give 0.945*X", but this doesn't seem to be a second knob, it's the requested PVC size.
requestImageSize seems to be our way to make storage methods that give us a much larger PVC than asked behave the same (possibly only for testing purposes). So I can introduce this kind of magic.

maya-r:
ping @awels @mhenriks

awels (commented):
So I am trying to understand what this is adding besides the RoundDown. You are calling dp.validate in resize, but the validate has already been called during convert. I guess in the cases we are not converting, the validate will not get called.

maya-r:
Right

maya-r:
The commit message I'm filtering for in the test has changed in the qemu->nbdkit move. I'll adjust.

awels:
Looks like there are some test failures once merge with latest master?

maya-r:
rebased on top of master with no changes to my code and added a single commit to address the test failures.

PR #1477, opened 2020-11-07
Create a Datavolume if a coliding PVC with same name exists but is marked to delete

### LINKED ISSUE #1476: Datavolume creation is rejected if a PVC with same name exists.
/kind enhancement

**What happened**:

As a user of Kubernetes, I would like data volumes to behave in a more Kubernetes fashion. Currently if I try to create a DV with the same name as a PVC that already exists, the creation will be rejected. But it is very possible the PVC is marked for deletion because right before creating the new DV, I deleted the old one but the cluster hasn't finished cleaning up.

Instead I would propose the DV is successfully created, but remains in a Pending state as long as the PVC with the same name still exists. Because the DV is in pending it will be clear that it cannot be used at this time.

As soon as the original PVC is removed, that would allow for the DV to create a new PVC as it does today and continue.

This whole process would more closely follow PVC semantics. If you create a PVC that cannot be bound, it still creates the PVC but it remains in pending.

Create->Delete->Create the same Data Volume

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): latest
- Kubernetes version (use `kubectl version`): v1.18.3
- Install tools: `make cluster-sync`
- Others:

marceloamaral:

Currently if I try to create a DV with the same name as a PVC that already exists, the creation will be rejected. But it is very possible the PVC is marked for deletion because right before creating the new DV, I deleted the old one but the cluster hasn't finished cleaning up.

Instead I would propose the DV is successfully created, but remains in a Pending state as long as the PVC with the same name still exists. Because the DV is in pending it will be clear that it cannot be used at this time.

As soon as the original PVC is removed, that would allow for the DV to create a new PVC as it does today and continue.

Fixes #1476

The simplest solution would be only verifying if the PVC is already marked as deleted using the `DeletionTimestamp` field, but it does not completely solve the problem.

For example, when we delete a DV, it will delete the PVC but there is a tiny delay between these phases (k8s asynchronous behavior...).
Therefore, creating a DV just after deleting it sometimes has the PVC DeletionTimestamp null because of the tiny delay.
To verify that, I executed the DV creation->deletion->creation 10 times, and only 4 times the PVC DeletionTimestamp was not null....

To properly identify if the PVC will be deleted because the "old" DV was deleted, we need to get info from the "old" DV. If the "old" DV does not exist or if its DeletionTimestamp is not null, it implies that the PVC will be deleted.

Therefore, to enable that feature, I needed to use the cdi client in the data volume validation in the apiserver to attempt to retrieve the DV that owns the target PVC, and of courser, update the unit tests accordingly.

**Release note**:

```release-note
N/A
```


pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -311,19 +311,23 @@ func (wh *dataVolumeValidatingWebhook) Admit(ar v1beta1.AdmissionReview) *v1beta
...
-				klog.Errorf("destination PVC %s/%s already exists", dv.GetNamespace(), dv.GetName())
-				var causes []metav1.StatusCause
-				causes = append(causes, metav1.StatusCause{
-					Type:    metav1.CauseTypeFieldValueDuplicate,
-					Message: fmt.Sprintf("Destination PVC already exists"),
-					Field:   k8sfield.NewPath("DataVolume").Child("Name").String(),
-				})
-				return toRejectedAdmissionResponse(causes)
-			}
+			// In the case that the PVC exist but it it is being deleted, we create the DV, otherwise the creation is denied
```
arnongilboa: Don't like `else if` ?
"exist but it it is" -> "exists but it is"

awels:
So unless I am missing something, I think we are making it more complex than it needs to be. If we try to create a DataVolume that already exists, it gets rejected. This is about the case where we create a DV while a PVC with the same name already exists. In the webhook we should only reject the DV if a PVC with the same name exists, and that PVC has no ownerRef. Or that PVC has an ownerRef that is not a DV. Because that means that PVC is not managed by the datavolume controller, and we can't use it.

The in the datavolume controller we will try to create the PVC and fail (because it hasn't been deleted yet completely) and we can use the deletionTimeStamp and ownerRef to determine if we need to re queue.

We also need some functional test that verify the behavior is fixed. And functional tests that test the new functionality.

pkg/apiserver/webhooks/datavolume-validate.go
marceloamaral: thanks @arnongilboa !

marceloamaral:
@awels thanks for the comments! I totally agree with you, so I cleaned the code and updated accordingly to your comments.
Regarding the functional tests, I will create a new issue and PR for that.

awels:
So we have a rule here, if a feature/fix doesn't have tests, it doesn't exist/is not fixed, as we cannot verify it is working as intended. As such we won't merge PRs with features/fixes without unit and functional tests. Sometimes we can do it by massaging existing tests to also test the new thing. In this case I think we will probably need a new functional test. Look in tests/datavolume_test.go for the other datavolume tests. I think the test should look something like:
```go
dataVolumeName := "test-dv"
size := 1Gi
dv := utils.NewDataVolumeForBlankRawImage(dataVolumeName, size)
dataVolume, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
By("verifying pvc was created and is Bound")
pvc, err := utils.WaitForPVC(f.K8sClient, dv.Namespace, dv.Name)
pvcUID := pvc.UID
// Now we know the DV exists and a matching PVC is there. We use the PVC UID to confirm later a new PVC was created
err = utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolume.Name)
// Now create the DV again
dataVolume, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
Eventually(func() bool {
  pvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Get(context.TODO(), dv.Name, metav1.GetOptions{})
  if err != nil {
	return false
  }
  // Return true if the pvc is not being deleted, and the UID is no longer the original one.
  return pvc.deletionTimeStamp == nil && pvc.UID != pvcUID

```

marceloamaral:
I am wondering if this verification should be introduced in the functional test
"Create/Delete same datavolume in a loop" in line 860 in file `tests/datavolume_test.go`. 

This test creates and deletes DV sequentially 5 times, but it does not verify the case of creation->deletion->creation that will have pvc collision.

What do you think, should we extend this test or create a new one?
I am wondering if changing this test could affect some tested behavior that I am not aware of...

awels:
I looked at the test, and was wondering how has this ever worked. until I realized that we are creating each DV in a different namespace. The way our framework is setup, every 'It' gets it own namespace, so I think you can modify that exact test, by changing the Context to an It, and removing the "It" from inside the loop. And it should still work with your PR.

Also if you run make lint and make unit- test it will run the same tests as travis. Travis is failing because of a linter issue.

awels:
Btw make sure to run make test-lint and make test-unit to ensure travis doesn't fail.

marceloamaral:
@awels I ran those tests plus functional test successfully before pushing the commits...
The files that I see that is different are:
```
	modified:   tests/BUILD.bazel
	modified:   tests/utils/BUILD.bazel
```
Do I also need to commit them to fix the lint issues?

awels:
Travis is saying:
AIL: Format errors found in the following files:
pkg/apiserver/webhooks/datavolume-validate.go

run gofmt -w pkg/apiserver/webhooks/datavolume-validate.go to have it fix that issue.

marceloamaral:
\retest

maya-r:
I'd like to see these changes:
- Release notes.
- Would be nice if the title was changed to explain what is the behaviour after this PR is merged, rather than the state before it.

tests/datavolume_test.go
```diff
@@ -857,30 +857,61 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
 
 					By("deleting DataVolume")
-					err = utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolumeName)
+					err = utils.DeleteDataVolume(f.CdiClient, dataVolume.Namespace, dataVolume.Name)
 					Expect(err).ToNot(HaveOccurred())
 
-				})
-			}
+					//Create the DV again to test if a PVC collision with a terminating PVC does not reject the DV creation
+					By("verifying if the DV creation is not rejected if a terminating PVC with same name exists")
```
awels, line 891: Maybe I am missing something, but the whole point of the loop is so we don't have to do anything that is below the delete of the datavolume.
So the following happens in a loop:
1. Create datavolume (it could be it still exists from previous loop)
2. Wait for datavolume to be complete (this is important, it stops us from running through the entire loop immediately)
3. Delete datavolume, but do NOT verify it is gone.
4. Goto 1.
The first iteration of the loop should be like before, but the second iteration 1. should NOT get rejected, but 2 will not happen until the original PVC is actually gone, and a new one is created. So doing it several times in a row will verify the feature is working as expected.

marceloamaral: You are right, we do not need to have the whole verification within one loop, but only verify the collision starting from the second iteration. I updated that.

awels (approved):
/lgtm
/approve

PR #1479, opened 2020-11-08
Fix coverage

mhenriks:

coverage is very low

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/core/v1beta1/utils/utils.go
+++ b/pkg/apis/core/v1beta1/utils/utils.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2020 The CDI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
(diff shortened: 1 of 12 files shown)


pkg/apis/core/v1beta1/utils/utils.go
```diff
@@ -0,0 +1,63 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
+)
+
+// IsPopulated indicates if the persistent volume passed in has been fully populated. It follow the following logic
+// 1. If the PVC is not owned by a DataVolume, return true, we assume someone else has properly populated the image
+// 2. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase succeeded return true
+// 3. If the PVC is owned by a DataVolume, look up the DV and check the phase, if phase !succeeded return false
+func IsPopulated(pvc *corev1.PersistentVolumeClaim, getDvFunc func(name, namespace string) (*cdiv1.DataVolume, error)) (bool, error) {
```
maya-r, line 30: we sure have a lot of copies of this function (4, up from 2 before)

mhenriks: Technically, the alpha and beta versions are different functions because they take different types.

Nevertheless, the code not in a `utils` package will be deleted once KubeVirt is updated to use the code in `utils`.  Didn't want to remove it prematurely and maybe break something

brybacki: Also tests are different because they import and use different types

The good thing is this function was only visible in one version and only known usage so far is in kubevirt, so we can easily update the import when bumping the version

maya-r:
/lgtm
but maybe we would want to make it clear that the duplicate versions can be deleted once kubevirt is updated not to use them.

PR #1498, opened 2020-11-19
Preallocation support

tomob:

This PR delivers preallocation support for DataVolumes

**Release note**:

```release-note
TBD
```


doc/preallocation.md
```diff
@@ -0,0 +1,72 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: sc-with-preallocation
+provisioner: ...
+reclaimPolicy: ...
+volumeBindingMode: ...
+preallocation: true
```
maya-r, line 48: I don't think we control the fields of storageclass

awels: Right we cannot use the storage class for any of this. We should have it in the CDIConfig object at the same level as the fsOverhead. We should probably make it a specific struct for storage class level settings.

tomob: Because the internal object are de/serialized to YAML, that would mean changing the CDIConfig API. Do we want to do it?

pkg/controller/util.go
```diff
@@ -616,3 +625,80 @@ func SetPodPvcAnnotations(pod *v1.Pod, pvc *v1.PersistentVolumeClaim) {
...
+
+// GetStorageClassNameForDV returns storage class to be used for the DV's PVC
+func GetStorageClassNameForDV(c client.Client, dv *cdiv1.DataVolume) string {
+	// If DV has a SC, return it
+	if dv != nil && dv.Spec.PVC != nil && dv.Spec.PVC.StorageClassName != nil && *dv.Spec.PVC.StorageClassName != "" {
+		return *dv.Spec.PVC.StorageClassName
+	}
+
+	// If owning VM has a SC in the teplate, return it
+	for _, or := range dv.OwnerReferences {
```
awels, line 671: Not quite sure why this is here. We want no dependencies on anything kubevirt related.

pkg/controller/util.go
```diff
@@ -616,3 +625,80 @@ func SetPodPvcAnnotations(pod *v1.Pod, pvc *v1.PersistentVolumeClaim) {
...
+	if cdiconfig.Status.Preallocation.Global != nil {
+		return *cdiconfig.Status.Preallocation.Global
+	}
+
+	// Finally, the default false
+	return defaultPreallocation
+}
+
+// GetStorageClassNameForDV returns storage class to be used for the DV's PVC
+func GetStorageClassNameForDV(c client.Client, dv *cdiv1.DataVolume) string {
```
awels, line 664: I see you modified GetStorageClassByName, so you know its there, what is the purpose of this function compared to that one?

awels (commented):
I am not quite sure what the reason is for the dependency on KubeVirt client, other than that the direction looks fine.

pkg/controller/util.go
tomob: Right, this part can be removed.

Correct me if I'm wrong, but if a DV is created from VM's template, the storage class is copied over from the template to the DV. Right?

pkg/controller/util.go
tomob: `GetStorageClassByName` takes a storage class name (string) and returns the storage class (object).
`GetStorageClassNameForDV` takes a DataVolume (object) and returns the storage class name (string) that it uses.

GetStorageClassNameForDV is called to find out what the storage class name is.

pkg/controller/util_test.go
```diff
@@ -255,6 +257,186 @@ var _ = Describe("setConditionFromPod", func() {
...
+	dv.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion: "v1",
+			Kind:       "VirtualMachine",
+			Name:       ownerName,
+		},
+	}
+}
+
+func createVirtualMachineWithStorageClassName(name, ns, dvName, storageClassName string) *kubevirtv1.VirtualMachine {
```
awels, line 360: This doesn't seem to be used anywhere, and if it was, I would be giving some push back on it. Anything from kubevirt/kubevirt, is going to be a problem in cdi, our main user is kubevirt, but we are not dependent on them

tomob: Oh, yes, this was used by the part of code that was removed. Removing this as well.

awels:
Did you run make deps-update. I believe removing the kubevirt code should also remove all the kubevirt vendored code. And that looks like it is still there.

tomob:
> Did you run make deps-update. I believe removing the kubevirt code should also remove all the kubevirt vendored code. And that looks like it is still there.

Right, removing...

pkg/controller/import-controller.go
```diff
@@ -489,6 +490,11 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 		podEnvVar.previousCheckpoint = getValueFromAnnotation(pvc, AnnPreviousCheckpoint)
 		podEnvVar.currentCheckpoint = getValueFromAnnotation(pvc, AnnCurrentCheckpoint)
 		podEnvVar.finalCheckpoint = getValueFromAnnotation(pvc, AnnFinalCheckpoint)
+
+		dv := &cdiv1.DataVolume{}
```
awels, line 494: So all communication between the data volume and a PVC goes through annotations. The established pattern for something like this is to write a 'preallocation' annotation on the PVC by the data volume controller when it creates the PVC. Then in the import/upload/clone controller we read the annotation. This way none of those controller need to know that data volumes even exist.

pkg/controller/upload-controller.go
```diff
@@ -361,6 +363,12 @@ func (r *UploadReconciler) createUploadPodForPvc(pvc *v1.PersistentVolumeClaim,
 		return nil, err
 	}
 
+	dv := &cdiv1.DataVolume{}
+	preallocation := false
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: pvc.Name, Namespace: pvc.Namespace}, dv); err == nil {
```
awels, line 368: Same comment as in import controller, if you have an annotation you can write a utility and re-use it in all the controllers.

pkg/importer/data-processor.go
```diff
@@ -262,34 +274,56 @@ func (dp *DataProcessor) resize() (ProcessingPhase, error) {
...
+			shouldPreallocate = false
 		}
 		if currentImageSizeQuantity.Cmp(minSizeQuantity) == 0 {
 			klog.V(1).Infof("No need to resize image. Requested size: %s, Image size: %d.\n", imageSize, info.VirtualSize)
-			return nil
+			return false, nil
 		}
 		klog.V(1).Infof("Expanding image size to: %s\n", minSizeQuantity.String())
-		return qemuOperations.Resize(dataFile, minSizeQuantity)
+		err := qemuOperations.Resize(dataFile, minSizeQuantity)
```
awels, line 323: Any reason you don't run the preallocation on the resize phase, since the resize phase is called regardless of the file was written by go client or qemu-img?

awels (changes requested):
A few comments, but overal looks pretty good.

pkg/importer/data-processor.go
tomob: I like clear separation of duties. If I do preallocation during that phase, it would make it a "resize-and-maybe-preallocate" phase.

awels (approved):
/lgtm
/approve

PR #1499, opened 2020-11-19
Make sure the DV is the main resource and single source of truth for …

brybacki:
…WaitForFirstConsumer.

Make sure the DV is the main resource and single source of truth for WaitForFirstConsumer logic. 

Removed a shortcut that checked if PVC is Pending, so now it always checks DataVolume. That way we are sure that other tools that sync on DV or use this helper function will see the same results.

Fixes #
The problem was found while working on https://github.com/kubevirt/kubevirt/pull/4025

**Release note**:

```release-note
NONE
```


davidvossel (approved):
I'm not an owner here, but just wanted to throw my support for getting this in :+1: 

I think it's a good fix that addresses some strange consistency issues that can occur with this function now that we allow a PVC owned by a DataVolume to be referenced directly as a Volume on a VMI.

PR #1504, opened 2020-11-23
Downgrade go-libnbd dependency a bit, to v1.3.9-cdi

maya-r:
go-libnbd is very pinned to specific versions of libnbd. This
one works with Fedora 31 (libnbd-1.4.1), but not with an older
libnbd (1.2.2).

This is a quick fix to avoid problems in a second build environment.

**Release note**:

```release-note
NONE
```


maya-r:
/hold
there might be another way

PR #1507, opened 2020-11-27
Add support for proxy in the CDI import pod, reconciling the info from OpenShift cluster wide proxy when available

### LINKED ISSUE #775: Support of http_proxy
**FEATURE REQUEST**:

/kind enhancement

**What happened**:
Not possible to import when cluster is behind proxy

Support for adding http_proxy to the import

In cluster where proxy is enforced to reach outside world, import of image is failing

marceloamaral:
Signed-off-by: Marcelo Amaral <marcelo.amaral1@ibm.com>

Currently, if an OpenShift cluster is configured with the cluster wide proxy object, the proxy configuration does not take effect in the CDI importer pod.

Fixes #775

The Cluster-Wide Proxy automatically injects environment variables (such as HTTP_PROXY, HTTPS_PROXY, and NO_PROXY) in pods of operators deployed with  Operator Lifecycle Manager (OLM). It is worth noting that operators can be optionally installed using OLM. And, if the CDI operator was not deployed using OLM, the importer proxy will not have the proxy env variables even though the proxy obj was configured.
Then, I extended the CDI to do the following if the OpenShift cluster-wide proxy exist:
- Get the Cluster-Wide Proxy object in the importer controller to inject the HTTPS_PROXY, HTTP_PROXY and NO_PROXY env variables into the importer proxy pod
- Extend the importer proxy pod use new env variables

**Release note**:

```release-note
NONE
```


brybacki:
I think this change deserves the `Release note:` might be helpful for someone running CDI on OpenShift to decide whether to upgrade

pkg/controller/import-controller.go
```diff
@@ -699,6 +723,42 @@ func getEndpoint(pvc *corev1.PersistentVolumeClaim) (string, error) {
 	return ep, nil
 }
 
+// returns the OpenShift cluster wide proxy object
+func getClusterWideProxy(log logr.Logger, client client.Client) (*ocpconfigv1.Proxy, error) {
```
brybacki, line 727: does it work? looks strange, where is the return when client.Get returns without error?

Maybe this function deserves a simple unit test

pkg/controller/import-controller.go
```diff
@@ -699,6 +723,42 @@ func getEndpoint(pvc *corev1.PersistentVolumeClaim) (string, error) {
...
+	// Get the Openshift cluster-wide Proxy obj.
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   ClusterWideProxyAPIGroup,
+		Kind:    ClusterWideProxyAPIKind,
+		Version: ClusterWideProxyAPIVersion,
+	})
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: ClusterWideProxyName}, u); err != nil {
```
brybacki, line 735: I do not like that an import-controller has to to know it is running on OpenShift and has to depend on its api. But I do not want to suggest any overengineered solution for a simple task of gathering variables.

Is it possible to make a set of starndard CDI proxy parameters and inject these variables from outside? If not then maybe we can at least hide the `getClusterWideProxy` in some kind of utility and make it return a very simple struct with three fields: `ImporterHTTPSProxy`, `ImporterHTTPProxy` and `ImporterNoProxy`.

pkg/controller/import-controller.go
```diff
@@ -699,6 +723,42 @@ func getEndpoint(pvc *corev1.PersistentVolumeClaim) (string, error) {
...
+			return clusterWideProxy, nil
+		}
+	}
+	return nil, errors.Errorf("The cluster does not have the OpenShift cluster wide proxy object\n")
+}
+
+func getValueFromClusterWideProxy(p *ocpconfigv1.Proxy, field string) string {
+	switch field {
+	case common.ImporterHTTPSProxy:
+		return p.Status.HTTPSProxy
```
brybacki, line 753: I think in current code p can be nil

pkg/controller/import-controller.go
marceloamaral: It makes sense and I moved that to `controller/util.go`

pkg/controller/import-controller.go
marceloamaral: Yes! I fixed it.

pkg/controller/import-controller.go
marceloamaral: Thanks for points this, I fixed the code and created the unit tests.

maya-r:
I'm not sure it's fair to say it fixes #775 if this pull request is OpenShift-specific.Although maybe editing the environment variables of the pod counts (very unusual way of configuring things).

awels (changes requested):
I think you are focusing too much on the Open Shift part. This whole setup needs to work outside of Open Shift in a regular kubernetes cluster as well. I fact I don't care too much at all about the cluster wide thing in Open Shift, we need a proxy configuration for CDI, that can be configured regardless of which k8s cluster we are running. We can use the Open Shift cluster wide proxy for inspiration on which fields we need in our proxy configuration, but it needs to be able to be configured and work without it.

I suggest you add a proxy struct to the CDIConfigSpec and CDIConfigStatus here: https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/apis/core/v1beta1/types.go#L352-L375 something like:
```go
type ImportProxy struct {
  HttpsProxy *string `json:"httpsProxy,omitempty"`
  HttpProxy *string `json:"httpProxy,omitempty"`
  NoProxy *string `json:"noProxy,omitempty"`
}
```
Now the one in spec allows users to manually override what is calculated for the one in the status. Basically the status is the result of what is observed in the cluster or set in the spec. This pattern is used with basically everything else in the CDIConfig already.

Then the import controller can read the STATUS import proxy configuration and use that. Once we have all that, then we can work on figuring out how we can reconcile the Open Shift cluster wide proxy into the CDIConfig status.

marceloamaral:
Thanks, @awels for the comments. In fact, in the previous "off-line" discussions with @dankenigsberg , we had determined to prioritize the OpenShift part in a first PR since it is required by some clients, and then, completely solve the problem in another PR.

In any case, I agree with your proposal, it is very aligned with what we were discussing for the "second PR" that we were planning. But, I will include the complete solution in this PR here then.

dankenigsberg:
> Once we have all that, then we can work on figuring out how we can reconcile the Open Shift cluster wide proxy into the CDIConfig status.

How about hco reading the cluster-wide proxy and copy it into CDIConfig? Please discuss it with the hco team now.

Respecting OpenShift proxy is a requirement, not an afterthought.

awels:
> > Once we have all that, then we can work on figuring out how we can reconcile the Open Shift cluster wide proxy into the CDIConfig status.
> 
> How about hco reading the cluster-wide proxy and copy it into CDIConfig? Please discuss it with the hco team now.
> 
> Respecting OpenShift proxy is a requirement, not an afterthought.

We have established patterns for getting configuration from Open Shift into the CDIConfig object. For instance with reading Routes for the uploadProxyUrl. But we approach this from if we detect the cluster wide proxy object CRD, then we do something. Instead of having it as the default. I can show you the pattern we use, and what I have in mind for this particular part of PR. But we need to establish a CDI import proxy first.

maya-r:
For help with testing, use [this service](https://github.com/maya-r/containerized-data-importer/commit/8ffdfeaa338945546e30444356893b9ac2b98070). I would personally modify cdi badserver to have a URL which only accepts requests if they provide the same header as the proxy, and maybe one that rejects the proxy.

awels:
I am looking at the PR and unless I am missing something, we are not watching the cluster wide proxy if it exists. So when someone modifies the cluster wide proxy, the CDIConfig doesn't automatically update. Adding a watch to a CRD that may not exist is a little tricky, but there is an example in addConfigControllerWatches already, but it is more complicated because it watches namespaced CRDs, and the cluster proxy is a cluster object. The below 'should' work but I haven't tested it:
```go
if err := ocpconfigv1.AddToScheme(mgr.GetScheme()); err != nil {
  return err
}
...
// check if cluster proxy CRD exist
err = mgr.GetClient().List(context.TODO(), &ocpconfigv1.ProxyList{})
if meta.IsNoMatchError(err) {
  return nil
}

if err != nil && !isErrCacheNotStarted(err) {
  return err
}
err = configController.Watch(&source.Kind{Type: &ocpconfigv1.Proxy{}}, &handler.EnqueueRequestsFromMapFunc{
  ToRequests: handler.ToRequestsFunc(func(obj handler.MapObject) []reconcile.Request {
    return []reconcile.Request{{
      NamespacedName: types.NamespacedName{Name: configName},
    }}
  }),
})
if err != nil {
  return err
}
```
What this does is check if the CRD exists, if it does, it adds a watch to changes to the Proxy object. Any changes to the Proxy object will result in the reconcile loop being run. The object in that reconcile loop will be the CDIConfig object, not the Proxy object. So you will have to look up the Proxy object. Because we have registered the types in the first few lines above, you should be able to just do:
```go
proxyList := &ocpconfigv1.ProxyList{}
if err := r.client.List(context.TODO(), proxyList, &client.ListOptions{}); IgnoreIsNoMatchError(err) != nil {
  return err
}
// There should only be one proxy, but since we don't know the name we get a list.
for _, proxy := range proxyList.Items {
  // Copy values from cluster proxy to STATUS (not spec)
  ...
}

```

marceloamaral:
@awels thank you very much for the comments. That is right, that code is not updated, I need to commit the changes. I am going to do that when everything is working well with all the tests.

pkg/operator/resources/operator/operator.go
```diff
@@ -204,6 +204,17 @@ func getNamespacedPolicyRules() []rbacv1.PolicyRule {
...
 		},
+		{
+			APIGroups: []string{
+				"config.openshift.io",
+			},
+			Resources: []string{
+				"proxies",
+			},
+			Verbs: []string{
+				"*",
```
awels, line 215: We don't need * on the cluster wide proxy. get/list/watch is sufficient.

tests/import_proxy_test.go
```diff
@@ -0,0 +1,326 @@
...
+		var auth string
+		port := httpPort
+		if withBasicAuth {
+			port = httpPortWithAuth
+			auth = fmt.Sprintf("%s:%s@", username, password)
+		}
+		return fmt.Sprintf("http://%s%s.%s:%s", auth, proxyServerName, f.CdiInstallNs, port)
+	}
+
+	FDescribe("importing from cdi proxied web server", func() {
```
awels, line 84: If you FDescribe (or FDescribeTable or FIt, or FContext) it will focus just those tests and the overall test run will fail. For testing this is what you want, but in the CI we want to run all the tests, so make sure to remove the F(ocus) from the tests you added before pushing.

tests/import_proxy_test.go
```diff
@@ -0,0 +1,326 @@
...
+			url                func() string
+			proxyURL           func(bool, bool) string
+			isHTTPS            bool
+			proxyWithBasicAuth bool
+			noProxyDomains     string
+			dvFunc             func(string, string, string) *cdiv1.DataVolume
+			expected           func() types.GomegaMatcher
+		}
+
+		FDescribeTable("should", func(args importProxyTestArguments) {
```
awels, line 97: Just an FYI if you FDescribeTable, you don't have to FEntry each entry, you are already focusing on that particular table.

pkg/controller/config-controller.go
```diff
@@ -333,6 +340,81 @@ func (r *CDIConfigReconciler) createCDIConfig() (*cdiv1.CDIConfig, error) {
...
+					if err := r.uncachedClient.Get(context.TODO(), types.NamespacedName{Name: common.ImportProxyConfigMapName, Namespace: r.cdiNamespace}, configMap); err == nil {
+						return nil
+					}
+					return nil
+				}
+			}
+			if configMap != nil {
+				configMap.Data[common.ImportProxyConfigMapKey] = string(certBytes)
+				if err := r.client.Update(context.TODO(), configMap); err != nil {
+					log.V(3).Info(err.Error())
```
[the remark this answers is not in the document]
marceloamaral: Yes, it should! thanks

pkg/controller/config-controller.go
```diff
@@ -333,6 +340,81 @@ func (r *CDIConfigReconciler) createCDIConfig() (*cdiv1.CDIConfig, error) {
...
+					}
+					if err := r.uncachedClient.Get(context.TODO(), types.NamespacedName{Name: common.ImportProxyConfigMapName, Namespace: r.cdiNamespace}, configMap); err == nil {
+						return nil
+					}
+					return nil
+				}
+			}
+			if configMap != nil {
+				configMap.Data[common.ImportProxyConfigMapKey] = string(certBytes)
+				if err := r.client.Update(context.TODO(), configMap); err != nil {
```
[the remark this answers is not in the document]
marceloamaral: Thanks, I missed that!

pkg/controller/config-controller.go
```diff
@@ -333,6 +340,81 @@ func (r *CDIConfigReconciler) createCDIConfig() (*cdiv1.CDIConfig, error) {
...
+						TypeMeta:   metav1.TypeMeta{},
+						ObjectMeta: metav1.ObjectMeta{Name: common.ImportProxyConfigMapName, Namespace: r.cdiNamespace},
+						Immutable:  new(bool),
+						Data:       map[string]string{common.ImportProxyConfigMapKey: string(certBytes)},
+						BinaryData: map[string][]byte{},
+					}
+					if err := r.client.Create(context.TODO(), configMap); err != nil {
+						log.V(3).Info(err.Error())
+					}
+					if err := r.uncachedClient.Get(context.TODO(), types.NamespacedName{Name: common.ImportProxyConfigMapName, Namespace: r.cdiNamespace}, configMap); err == nil {
```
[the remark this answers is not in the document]
marceloamaral: Humm, I used that before for debugging and I forgot to remove it! There is no need for retrieving it after creating it...

pkg/controller/config-controller.go
```diff
@@ -375,7 +457,7 @@ func NewConfigController(mgr manager.Manager, log logr.Logger, uploadProxyServic
 }
 
 // addConfigControllerWatches sets up the watches used by the config controller.
-func addConfigControllerWatches(mgr manager.Manager, configController controller.Controller, cdiNamespace, configName, uploadProxyServiceName string) error {
+func addConfigControllerWatches(mgr manager.Manager, configController controller.Controller, cdiNamespace, configName, uploadProxyServiceName string, log logr.Logger) error {
```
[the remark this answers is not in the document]
marceloamaral: I am using the log there, both the watch router and proxy parts are writing log message when they do not exit, i.e., when the cluster is not OpenShift

pkg/controller/import-controller.go
```diff
@@ -572,7 +583,7 @@ func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim) (boo
 	return false, nil
 }
 
-func (r *ImportReconciler) getCertConfigMap(pvc *corev1.PersistentVolumeClaim) (string, error) {
+func (r *ImportReconciler) getCertConfigMap(pvc *corev1.PersistentVolumeClaim, config *cdiv1.CDIConfig) (string, error) {
```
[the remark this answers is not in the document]
marceloamaral: I used it for debugging and forgot to remove! Thanks for point it!

pkg/controller/util.go
```diff
@@ -678,3 +699,61 @@ func GetStorageClassNameForDV(c client.Client, dv *cdiv1.DataVolume) string {
...
+	clusterWideProxy := &ocpconfigv1.Proxy{}
+	var err error
+	if err = r.Get(context.TODO(), types.NamespacedName{Name: ClusterWideProxyName}, clusterWideProxy); err == nil {
+		return clusterWideProxy, nil
+	}
+	return nil, err
+}
+
+// IsImportProxyConfigured return true if import proxy configuration was defined or reconcillied.
+func IsImportProxyConfigured(config *cdiv1.CDIConfig) (bool, bool) {
```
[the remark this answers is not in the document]
marceloamaral: Actually, I forgot to delete this function...

tests/import_proxy_test.go
marceloamaral: Included it by mistake...

marceloamaral:
thank you very much @awels for the comments, I have updated the code accordingly and it is ready for the second round of review!

awels:
Also looks like a test failure in two lanes. Maybe a timing issue?

pkg/controller/config-controller.go
```diff
@@ -333,6 +339,75 @@ func (r *CDIConfigReconciler) createCDIConfig() (*cdiv1.CDIConfig, error) {
 	return config, nil
 }
 
+func (r *CDIConfigReconciler) reconcileImportProxy(config *cdiv1.CDIConfig, log logr.Logger) error {
```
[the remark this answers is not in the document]
marceloamaral: Right! Thanks!

pkg/controller/util.go
```diff
@@ -678,3 +698,45 @@ func GetStorageClassNameForDV(c client.Client, dv *cdiv1.DataVolume) string {
...
 }
+
+// GetClusterWideProxy returns the OpenShift cluster wide proxy object
+func GetClusterWideProxy(r client.Client) (*ocpconfigv1.Proxy, error) {
+	clusterWideProxy := &ocpconfigv1.Proxy{}
+	var err error
+	if err = r.Get(context.TODO(), types.NamespacedName{Name: ClusterWideProxyName}, clusterWideProxy); err == nil {
+		return clusterWideProxy, nil
+	}
+	return nil, err
```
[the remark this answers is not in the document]
marceloamaral: Sure! Thanks!

pkg/controller/import-controller_test.go
```diff
@@ -629,29 +631,30 @@ var _ = Describe("Import test env", func() {
...
+			insecureTLS:        false,
+			currentCheckpoint:  "",
+			previousCheckpoint: "",
+			finalCheckpoint:    "",
+			preallocation:      false}
 		Expect(reflect.DeepEqual(makeImportEnv(testEnvVar, mockUID), createImportTestEnv(testEnvVar, mockUID))).To(BeTrue())
 	})
 })
 
-var _ = Describe("getSecretName", func() {
```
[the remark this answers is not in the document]
marceloamaral: No, it was a mistake!

tests/import_proxy_test.go
```diff
@@ -0,0 +1,360 @@
...
+	Eventually(func() bool {
+		importerPod := getImporterPod(f)
+		proxyLog := getProxyLog(f)
+		return wasPodProxied(isHTTPS, uri, importerPod, proxyLog)
+	}, time.Second*60, time.Second).Should(expected())
+}
+
+func getImporterPod(f *framework.Framework) *corev1.Pod {
+	fmt.Fprintf(GinkgoWriter, "INFO: Waiting importer pod to become ready\n")
+	err := waitForPodIP(f, common.ImporterPodName, f.Namespace.Name, utils.PodWaitForTime)
```
awels, line 280: It seems we have failures in this particular part of the test in most lanes. I am trying to understand exactly what we are doing here and this is what I think we are doing:
1. Get the IP of the importer pod
2. Get the logs from the proxy pod (we can look up the proxy pod name at the start of the test)
3. Check if IP and URI.host and method are in a line in the proxy logs.

A few things I think we can improve:
1. You can specify a flag like --since or --since-time to kubectl logs to only return recent logs so you are not scanning a huge log.
2. We are only interested in the importer pod IP, once we have it, we don't have to look up the importer pod anymore.
3. The wait.PollImmediate is an old construct, we should be using Eventually instead.

Example of how I would look up the importer pod IP:
```go
podIp := ""
Eventually(func() string {
  pod, err := utils.FindPodByPrefix(f.K8sClient, namespace, podName, common.CDILabelSelector)
  if err != nil {
    return ""
  }
  fmt.Fprintf(GinkgoWriter, "INFO: Checking POD %s phase: %s\n", podName, string(pod.Status.Phase))
  podIp = pod.Status.PodIP != "" {
  return podIp
}, time.Second*60, time.Second).Should(NotBe(Empty()))
// podIp should be something at this point.
```
Now you have the IP address and you can look up the proxy logs.

marceloamaral: Yes, that is the goal of this method, and thanks for the feedback!

I found the problem of the test been failing intermittently.... Sometimes, the importer pod finishes too fast and the method cannot inspect the pod. To solve that, I decrease the import pod log pooling interval and changed the image size from the tiny image to the large one...

Also, I simplified some parts of the code to be cleaner.

awels: So we have that problem in other places as well and we have a special port on the http server that is rate limited, this ensures the import takes longer and thus the pod won't disappear as fast. Look for HTTPRateLimitPort for the port number, and TinyCoreQcow2URLRateLimit for an example of how to use it.

tests/import_proxy_test.go
```diff
@@ -0,0 +1,319 @@
...
+	password         = "bar"
+	httpPort         = "8080"
+	httpPortWithAuth = "8081"
+	proxyServerName  = "cdi-test-proxy"
+	fileHostName     = "cdi-file-host"
+	//we use the a large image to "slowdown" the download process to have time to inspect the importer pod
+	cirrosFileName = "cirros-large-virtual-size.qcow2"
+)
+
+var _ = FDescribe("Import Proxy tests", func() {
```
awels, line 39: You forgot to remove the F again on the describe/describetable and entry

tests/import_proxy_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+		By("Verifying a message was printed to indicate a request was done using proxy")
+		verifyImporterPodInfoInProxyLogs(f, dataVolume, imgURL, args.isHTTPS, args.expected)
+
+		By("Waiting for the img import to be completed")
+		err = utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, dataVolume.Name)
+		Expect(err).ToNot(HaveOccurred(), "Datavolume not in phase succeeded in time")
+
+		By("Deleting DataVolume")
+		err = utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, args.name)
```
awels, line 85: You delete the DV at the end of the test however, if something failed before that, the datavolume is not actually removed because that code is not executed. The normal way you do cleanup is in an:
```go
AfterEach(func() {
// Cleanup code goes here
})
```
Any setup that needs to happen goes in a "BeforeEach"

tests/import_proxy_test.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+func cleanClusterWideProxy(f *framework.Framework) {
+	if utils.IsOpenshift(f.K8sClient) {
+		By("Restoring cluster-wide proxy object to original state")
+		ocpClient, err := configclient.NewForConfig(f.RestConfig)
+		Expect(err).ToNot(HaveOccurred())
+		proxy, err := ocpClient.ConfigV1().Proxies().Get(context.TODO(), controller.ClusterWideProxyName, metav1.GetOptions{})
+		if k8serrors.IsNotFound(err) {
+			Skip("This OpenShift cluster version does not have a Cluster Wide Proxy object")
```
awels, line 301: Skip is a function to tell the ginkgo framework to skip the test. The cleanClusterWideProxy is a cleanup function, which means the test has already completed. So I would just return instead of calling skip.

tests/import_proxy_test.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		proxy.Spec.HTTPProxy = ""
+		proxy.Spec.HTTPSProxy = ""
+		proxy.Spec.NoProxy = ""
+		_, err = ocpClient.ConfigV1().Proxies().Update(context.TODO(), proxy, metav1.UpdateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Wating OpenShift Cluster Wide Proxy reconcile after the cleanup")
```
awels, line 309: Type By("Wating ... should be By("Waiting...

awels:
The pattern we use when modifying cluster wide objects (like the ClusterProxy) is the following:
In the describe we have a variable that stores the 'spec' part of the object:

```go
Describe("Example code", func() {
  var orgSpec *someobject

  BeforeEach(func() {
    currentObject = client.GET("objectname")
    orgSpec = currentObject.Spec.DeepCopy()
  })

  AfterEach(func() {
    Eventually(func() error {
      modifiedObject = client.GET("objectname")
      // Restore original spec
      modifiedObject.Spec = orgSpec
      return client.UPDATE(modifiedObject)
    }, time.Second * 60, time.Second).ShouldNot(HaveOccurred())
  })

  It("the test function", func() {
    ... Test that modifies the object
  })
})
```
This way the original object gets restored to whatever state we found it in.

tests/import_proxy_test.go
marceloamaral: Thanks, I updated it accordingly.

marceloamaral: Thanks, I updated it accordingly.

marceloamaral:
Thanks @awels, I've updated the code accordingly to your comments. 
It is ready for another review round. Now, using the rate limit for the importer pod, all tests are succeeding.

tests/import_proxy_test.go
```diff
@@ -0,0 +1,352 @@
...
+		return true
+	}, timeout, 2*pollingInterval).Should(BeTrue())
+	return err
+}
+
+func cleanClusterWideProxy(ocpClient *configclient.Clientset, clusterWideProxySpec *ocpconfigv1.ProxySpec) {
+	Eventually(func() error {
+		proxy, err := ocpClient.ConfigV1().Proxies().Get(context.TODO(), controller.ClusterWideProxyName, metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		proxy.Status.HTTPProxy = clusterWideProxySpec.HTTPProxy
```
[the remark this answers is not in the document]
marceloamaral: I totally agree with you, it was a mistake. It should update the Spec as it is done in function `updateCDIConfigByUpdatingTheClusterWideProxy`...

My intention of the Eventually here is to wait it reconcile (after updating the Spec) to be sure that the next execution will access the updated object....

tests/import_proxy_test.go
```diff
@@ -0,0 +1,352 @@
...
+			fmt.Fprintf(GinkgoWriter, "INFO: Proxy log: %s\n", line)
+			fmt.Fprintf(GinkgoWriter, "INFO: The import POD requests were proxied\n")
+			return true
+		}
+	}
+	fmt.Fprintf(GinkgoWriter, "INFO: The import POD requests were not proxied\n")
+	return false
+}
+
+func waitForDataVolumePhase(f *framework.Framework, dvName string, phase cdiv1.DataVolumePhase) error {
```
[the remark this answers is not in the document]
marceloamaral: Hum..actually I don't remember, and looking at it now I cannot see in any reason of why not use utils.WaitForDataVolumePhase
I will update it!

awels:
Code all looks good, should be able to merge. Do we have documentation on this feature?

doc/cdi-config.md
```diff
@@ -16,7 +16,7 @@ CDI configuration in specified by administrators in the `spec.config` of the `CD
...
 | storageClass             | nil           | A value of `local: "0.6"` is understood to mean that the overhead for the local storageClass is 0.6.                                                                                                                         |
 | preallocation            | nil           | Preallocation setting to use unless a per-dataVolume value is set                                                                                                                                                            |
-
+| ImportProxy              | nil           | updated when a CDIConfig is updated or when Cluster Wide-Proxy (Openshift) is updated. If `ImportProxy` is set, importer pod URL will be ignored and `ImportProxy.HTTPSProxy` or `ImportProxy.HTTPProxy` will be updated with the user defined URL. HTTPS has priority. If `ImportProxy.NoProxy` is set, the proxy URL will be ignored if it contains any of the listed hostnames and/or CIDRs. If `ImportProxy.HTTPSProxy` is set, the `ImportProxy.TrustedCAProxy` must be provided. |
```
marceloamaral, line 19: @awels I add the documentation here, but I am not sure if we need to update somewhere else.

awels: We need to update the overall documentation that all that stuff is now set in the CDI CR, but that is not your problem. This works for me.

marceloamaral: I see! Thanks!
Just to let you know, I did a last commit updating the description of importerProxy in this documentation.

marceloamaral:
@awels since you approved the PR, can we remove the hold label that you set a while ago?

PR #1486, opened 2020-11-18
Update fedora 33

alicefr:
Update fedora version to 33

**What this PR does / why we need it**:
Needed for new nbdkit filters and plugins not available in fedora 31

**Release note**:
<!--  Write your release note:
Update to fedora 33
```release-note
none
```

PR #1508, opened 2020-11-27
Use nbdkit for direct stream for the http importer 

alicefr:
The goal of this PR is to introduced nbdkit and its filters and plugins to enable to stream the image content directly without the need of additional scratch space. This is used when the source is http and the contentType kubevirt. There are still certain corner cases that nbdkit cannot handle and the scratch space is required. Those are when:
 - the source server doesn't support accept ranges
 - the source server is redirected to S3 because S3 doesn't support head requests.

- This PR adjusts several test cases because they check the use of the scratch space or check the output messages.

**Release note:**
```release-note
Added nbdkit for direct streaming without scratch space for the http importer
```


alicefr:
The [travis job](https://travis-ci.org/github/kubevirt/containerized-data-importer/builds/746257594) is failing because in the current `kubevirt/kubevirt-cdi-bazel-builder` image is based on fedora 31, and it doesn't have all the ndbkit filters and plugin. This requires an update image. For the gzip filter, this needs #1486.

rwmjones:
I don't know much about this topic so my questions are probably quite naive.  Here goes anyway ...

* What are the new dependencies: liburing, libseccomp, device-mapper needed for?
* Why does the "WORKSPACE" file reference particular RPMs from `download.fedoraproject.org`?  I would imagine that there would be a Dockerfile that has a `dnf install nbdkit nbdkit-curl-plugin` (etc) line in it instead.  The problem with referencing particular RPMs is that they will get out of date if we issue updates, especially a problem for CVEs.
* If I understand it correctly,  the approach is to run basically `nbdkit curl ... --run "qemu-img convert ..."`.  The idea being to use nbdkit filter to transparently uncompress while downloading the content?  This saves some copies during download, but I don't think the real benefit can be realised until CDI is modified to use an NBD endpoint for disks (obviously a much more major change).

A couple of comments on the patch itself:

* I guess you don't want to use a hard-coded `/var/run/nbdkit.pid` as that will prevent two copies from running in the same container.  (Maybe this can never happen?  I'm sure it's still a possible security issue to use a hard-coded path though)
* `--exit-with-parent` - good!  However in the case where you're using `--run` it won't do anything additional (although it's not harmful).

alicefr:
> * What are the new dependencies: liburing, libseccomp, device-mapper needed for?

Those deps are needed to uprade to fedora 33 (needed to get the gzip filter). See #1486. Once, that PR is merged I can rebase the changes on top. I included the commits of the other PR to test these changes with the CI. Even, if this is not true because I need to build a new image `kubevirt/kubevirt-cdi-bazel-builder` :/, and the CI doesn't build the image locally. So, I guess, I'll simply remove those commits. Thanks for pointing that.

> * Why does the "WORKSPACE" file reference particular RPMs from `download.fedoraproject.org`?  I would imagine that there would be a Dockerfile that has a `dnf install nbdkit nbdkit-curl-plugin` (etc) line in it instead.  The problem with referencing particular RPMs is that they will get out of date if we issue updates, especially a problem for CVEs.

This is standard bazel to have reproducible builds. If you want to update, you need to reference a new rpm. This is again part of the PR  #1486. I'll remove the commits as it's confusing.

> * If I understand it correctly,  the approach is to run basically `nbdkit curl ... --run "qemu-img convert ..."`.  The idea being to use nbdkit filter to transparently uncompress while downloading the content?  This saves some copies during download, but I don't think the real benefit can be realised until CDI is modified to use an NBD endpoint for disks (obviously a much more major change).
>
I'm not sure that we can change the source from where the image is downloaded. If I understand you correctly
> A couple of comments on the patch itself:
> 
> * I guess you don't want to use a hard-coded `/var/run/nbdkit.pid` as that will prevent two copies from running in the same container.  (Maybe this can never happen?  I'm sure it's still a possible security issue to use a hard-coded path though)
> * `--exit-with-parent` - good!  However in the case where you're using `--run` it won't do anything additional (although it's not harmful).

There will be a single instance of nbdkit in the container. Additionally, the importer container runs just the time for getting the image. I tried to stick with https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/importer/vddk-datasource.go#L45. 
 But I can remove the `--exit-with-parent`  :)

Thanks for the comments.

alicefr:
I leave the commits of the #1486, otherwise, the WORKSPACE is broken with a mix of fedora 31 and 33 rpms. Could somebody suggest to me a nice way how can I overwrite the `kubevirt/kubevirt-cdi-bazel-builder`. Or I can only hack it into the code?

alicefr:
> * If I understand it correctly,  the approach is to run basically `nbdkit curl ... --run "qemu-img convert ..."`.  The idea being to use nbdkit filter to transparently uncompress while downloading the content?  This saves some copies during download, but I don't think the real benefit can be realised until CDI is modified to use an NBD endpoint for disks (obviously a much more major change).

@rwmjones  back to your point. I guess, what will be needed by cdi, it's a new `nbd` importer type. The question for me is if the image provider can expose the image through nbd. But this deserves a separate discussion :)

rwmjones:
> > * If I understand it correctly,  the approach is to run basically `nbdkit curl ... --run "qemu-img convert ..."`.  The idea being to use nbdkit filter to transparently uncompress while downloading the content?  This saves some copies during download, but I don't think the real benefit can be realised until CDI is modified to use an NBD endpoint for disks (obviously a much more major change).
> 
> @rwmjones back to your point. I guess, what will be needed by cdi, it's a new `nbd` importer type. The question for me is if there are providers that expose the image through nbd. But this deserves a separate discussion :)

Right, something like an NBD importer where instead of downloading the file, it creates an NBD endpoint and whatever consumes the CDI content uses that.  Obviously a significant change.

awels:
if you want to try a builder yourself you can build it by running podman (or docker) build in the hack/build/docker/builder directory. That will build the container and save it to your local machine. Then if you run docker image ls --digests and find the builder image you just created, you can set the BUILDER_IMAGE to repository@sha256 and when you run a make bazel-build it will use the new image.

pkg/image/nbdkit.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+func (p NbdkitPlugin) String() string {
+	return string(p)
+}
+func (f NbdkitFilter) String() string {
+	return string(f)
+}
+
+func init() {
```
awels, line 44: Don't need this, since the qemu.go in the same package already calls it.

awels (commented):
I think the main thing bothering me here is the separate nbdkit conversion step in the data processor. On that level we really don't care what is doing the actual conversion. Right now we have this QEMUOperations interface and that is the main interface the data processor uses to call out to conversion functions. nbdkit is basically a wrapper around qemu (for our purposes) so I think it is reasonable for us to simply call the nbdkit ConvertToRawStream from the qemuOperations convertToRaw. This of course needs some setup of the qemuOperations so it can pass the proper arguments when it instantiates the ndkit struct. It might even be interesting to embed the nbdkit struct in the qemu operations struct.

hack/build/docker/builder/Dockerfile
```diff
@@ -29,6 +29,9 @@ RUN sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
 	patch \
 	libnbd-devel \
 	nbdkit-devel \
+        nbdkit-curl-plugin \
```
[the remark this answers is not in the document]
alicefr: @awels because it is used to run the unit tests. I added some tests in the `pkg/image/nbdkit_test.go` without these deps, it cannot find the plugin/filter *.so

pkg/image/nbdkit.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// Nbdkit filters
+const (
+	NbdkitXzFilter   NbdkitFilter = "xz"
+	NbdkitTarFilter  NbdkitFilter = "tar"
+	NbdkitGzipFilter NbdkitFilter = "gzip"
+)
+
+func (p NbdkitPlugin) String() string {
```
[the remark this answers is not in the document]
alicefr: ok, I'll change that thanks!

pkg/image/nbdkit.go
alicefr: ok thanks :)

alicefr:
> I think the main thing bothering me here is the separate nbdkit conversion step in the data processor. On that level we really don't care what is doing the actual conversion. Right now we have this QEMUOperations interface and that is the main interface the data processor uses to call out to conversion functions. nbdkit is basically a wrapper around qemu (for our purposes) so I think it is reasonable for us to simply call the nbdkit ConvertToRawStream from the qemuOperations convertToRaw. This of course needs some setup of the qemuOperations so it can pass the proper arguments when it instantiates the ndkit struct. It might even be interesting to embed the nbdkit struct in the qemu operations struct.

@awels I totally agree with the comment. But this a bigger change and I wanted to have a first opinion from you.  I'm not quite sure how to refactor the code. The nbdkit is not implementing all the [QemuOperations](https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/image/qemu.go#L58), what we really need is ` ConvertToRawStream` and `Info`. Maybe we could separate those in another interface

hack/build/docker/builder/Dockerfile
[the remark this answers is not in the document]
alicefr: This is also the reason why I need to upgrade the `kubevirt-cdi-bazel-builder` image to fedora 33 (#1511)

rwmjones:
`nbdinfo` (https://libguestfs.org/nbdinfo.1.html) could implement the info side.  It also has far fewer dependencies than qemu-img.

Not sure what ConvertToRawStream does, but if (as the name implies) it converts an NBD export to stdout then `nbdcopy` (https://libguestfs.org/nbdcopy.1.html) can do this directly (`qemu-img convert` cannot do this at all, it's a unique feature of nbdcopy).

alicefr:
@rwmjones basically you can translate those functions to:
- Info = `nbdkit -U - curl ... --run "qemu-img info $nbd --output=json"`
- ConvertToRawStream  = `nbdkit -U - curl ... --run "qemu-img convert $nbd -t none -p -O raw /data/disk.img `

Many thanks for the suggestions. I can have a look to nbdinfo and ndbcopy. This can probably substitute the `--run qem-img`. I think the question is if those commands can replace all the options of `qemu-img`

hack/build/docker/builder/Dockerfile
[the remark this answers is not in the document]
awels: So this is the reason we have an interface QEMUOperations for the qemu operations that currently exist. We can mock out the actual calls to qemu by implementing a mock struct that implements all the functions. In those functions we can then verify that we are receiving the correct values and we can return any potential errors without having to worry about if the system running the tests actually has qemu or not. It also makes the test faster as it doesn't have to fork a process to run qemu.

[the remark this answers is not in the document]
alicefr: Oh ok, I see!  I haven't noticed that! I'll change the code accordingly! Many thanks!

awels:
> > I think the main thing bothering me here is the separate nbdkit conversion step in the data processor. On that level we really don't care what is doing the actual conversion. Right now we have this QEMUOperations interface and that is the main interface the data processor uses to call out to conversion functions. nbdkit is basically a wrapper around qemu (for our purposes) so I think it is reasonable for us to simply call the nbdkit ConvertToRawStream from the qemuOperations convertToRaw. This of course needs some setup of the qemuOperations so it can pass the proper arguments when it instantiates the ndkit struct. It might even be interesting to embed the nbdkit struct in the qemu operations struct.
> 
> @awels I totally agree with the comment. But this a bigger change and I wanted to have a first opinion from you. I'm not quite sure how to refactor the code. The nbdkit is not implementing all the [QemuOperations](https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/image/qemu.go#L58), what we really need is ` ConvertToRawStream` and `Info`. Maybe we could separate those in another interface

 I think we named the QEMUOperations a bit unfortunate when we originally implemented this (2+ years ago). Qemu really is an implementation detail, and should not be that prominent in the interface name. It should be completely irrelevant to a caller of the interface what is actually doing the tasks (QEMU/Nbdkit/whatever). What you are doing is replacing part of what qemu-img does right now (ConvertToRawStream), and reading the rest of the comments from Richard, looks like there could be some other things that can be replaced as well. So maybe we should figure out a better name to describe what the interface does, and simply mix and match qemu and nbdkit in the implementation of that interface.

rwmjones:
Just want to clarify a few thoughts about why we're doing this, so this is background info.  It's a bit disorganized ...

The set of qemu developers ∩ libnbd and nbdkit developers is non-empty!  We work on both, because they have slightly different focus.

* libnbd: Small C library (with golang and other bindings) for accessing NBD servers
* nbdkit: Plugin-based NBD server designed for maximum flexibility and ease of use
* qemu: Swiss-army chainsaw that can do everything

nbdcopy and nbdinfo are small client programs built on libnbd so they can only access NBD endpoints.

The main limitation of libnbd/nbdkit is lack of qcow2 support since we don't want any alternate implementation of qcow2 except for the definitive one in qemu.  You shouldn't be sending qcow2 data over NBD really (although people occasionally do it).  NBD should process only raw disks.

You can use nbdcopy with qcow2 like this:
```
stream-in-some-data | nbdcopy -- - [ qemu-nbd -f qcow2 output.qcow2 ]
```
but that's cheating because it runs qemu-nbd as a subprocess.  However this is still more powerful than qemu-img convert (because of the streaming which qemu-img cannot do).

Any time you have a temporary disk image sitting around, consider instead connecting various NBD components together into a pipeline.  These tools are designed to construct pipelines that stream and transform data, both locally and remotely, with format conversion, decompression, etc.

alicefr:
> > > I think the main thing bothering me here is the separate nbdkit conversion step in the data processor. On that level we really don't care what is doing the actual conversion. Right now we have this QEMUOperations interface and that is the main interface the data processor uses to call out to conversion functions. nbdkit is basically a wrapper around qemu (for our purposes) so I think it is reasonable for us to simply call the nbdkit ConvertToRawStream from the qemuOperations convertToRaw. This of course needs some setup of the qemuOperations so it can pass the proper arguments when it instantiates the ndkit struct. It might even be interesting to embed the nbdkit struct in the qemu operations struct.
> > 
> > 
> > @awels I totally agree with the comment. But this a bigger change and I wanted to have a first opinion from you. I'm not quite sure how to refactor the code. The nbdkit is not implementing all the [QemuOperations](https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/image/qemu.go#L58), what we really need is ` ConvertToRawStream` and `Info`. Maybe we could separate those in another interface
> 
> I think we named the QEMUOperations a bit unfortunate when we originally implemented this (2+ years ago). Qemu really is an implementation detail, and should not be that prominent in the interface name. It should be completely irrelevant to a caller of the interface what is actually doing the tasks (QEMU/Nbdkit/whatever). What you are doing is replacing part of what qemu-img does right now (ConvertToRawStream), and reading the rest of the comments from Richard, looks like there could be some other things that can be replaced as well. So maybe we should figure out a better name to describe what the interface does, and simply mix and match qemu and nbdkit in the implementation of that interface.

@awels yes, I agree, I'll try to refactor the code, and push it for review.

alicefr:
@rwmjones thanks for the comment. However, I'm a bit confused about which part should be modified :). Are you suggesting we should use `qemu-nbd` instead of `qemu-img`? Or we should change completely the tools in the flow? The `stream-in-some-data` tool should be able to deal with compressed archives (xz and gz), and set custom certificates.

alicefr:
@rwmjones I see, thanks for the explication! So far I think using `nbdkit` + `qemu img` should be fine as the final image is raw. Once, the code has been refactored, it shouldn't be a problem to change the background tools and do a more performant transfer.

hack/build/docker/builder/Dockerfile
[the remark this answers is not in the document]
alicefr: Removed those deps

alicefr:
I refactored the code in a way that during the `ConvertNbd` phase the `qemuOperations` is overwritten with an object that implements the `QEMUOperations` interface. The nbdkit parts needs to reimplement only `Info` and `ConvertToRawStream` for the other cases it reuses the function from `qemu.go`. I also modified the unit tests for nbdkit. Now, it's using a mock interface and we get rid of the nbdkit filters and plugins dependencies in the cdi-builder container.

alicefr:
Rebased the PR. Now it integrates the new image with fedora 33

pkg/importer/http-datasource.go
```diff
@@ -65,11 +66,13 @@ type HTTPDataSource struct {
 	// url the url to report to the caller of getURL, could be the endpoint, or a file in scratch space.
 	url *url.URL
 	// true if we are using a custom CA (and thus have to use scratch storage)
```
maya-r, line 68: need to adjust comment here

alicefr: oh yes! Thanks! Done :)

alicefr:
Not quite sure what is going wrong in fetching the new packages needed by this PR. However, when I tried locally, it works without problems. In the previous round, some lanes were able to fetch the RPMs correctly. The rpm version hasn't changed, so it's not an issue about new sha or package version. I guess I'll try to retrigger the tests tomorrow.

awels:
@alicefr looks like the fedora mirrors are not working properly. Or the resolution of the mirror.

alicefr:
Now the networking issues have been solved. I rebased the PR, and I think it is ready for another round of review. On my side, I'd like to know if I should remove the tests that use the scratch space. Right now, I simply marked that with P to be skipped (commit edd694179d3ad0bc23)

pkg/importer/data-processor.go
```diff
@@ -49,6 +49,9 @@ const (
 	// ProcessingPhaseConvert is the phase in which the data is taken from the url provided by the source, and it is converted to the target RAW disk image format.
 	// The url can be an http end point or file system end point.
 	ProcessingPhaseConvert ProcessingPhase = "Convert"
+	// ProcessingPhaseConvertNbd is the phase in which the data is taken from the url provided by the source, and it is converted to the target using ndb to directly stream the data into the RAW disk image format.
+	// The url can be an http end point or file system end point.
+	ProcessingPhaseConvertNbd ProcessingPhase = "ConvertNbd"
```
awels, line 54: I am wondering if we need this phase at all. From what I am reading the Nbdkit code is only used in the http data source. And the only place that would use that phase would be the http data source. Would it not make sense to configure the qemuOperations variable in the Info function of that data source? The data source is in the same package so you have access to that variable there.

That way you can skip implementing GetNbdkit() in every data source because you need it in the data processor.

pkg/importer/http-datasource.go
```diff
@@ -115,17 +118,25 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
...
+		hs.n.AddFilter(image.NbdkitGzipFilter)
+		klog.V(2).Infof("Added nbdkit gzip filter")
+	}
+	if hs.readers.ArchiveXz {
+		hs.n.AddFilter(image.NbdkitXzFilter)
+		klog.V(2).Infof("Added nbdkit xz filter")
 	}
-	return ProcessingPhaseTransferScratch, nil
+
+	return ProcessingPhaseConvertNbd, nil
```
awels, line 139: I think you can do something like this here:
```go
qemuOperations = image.NewNbdkitOperations(hs.GetNbdkit())
return ProcessingPhaseConvert, nil
```
And have the same result as you have now without all the extra stuff in data processor.

tests/datavolume_test.go
```diff
@@ -535,7 +542,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 					Message: "Import Complete",
 					Reason:  "Completed",
 				}}),
-			table.Entry("[rfe_id:1115][crit:high][test_id:1379]succeed creating import dv with given valid qcow2 url (https) should require scratchspace", dataVolumeTestArguments{
+			table.PEntry("[rfe_id:1115][crit:high][test_id:1379]succeed creating import dv with given valid qcow2 url (https) should require scratchspace", dataVolumeTestArguments{
```
awels, line 545: I think you can leave this as it, maybe remove the 'should require scratchspace' part of the test label. We never actually check that scratch space is created anyway. We just make sure the import completes successfully.

tests/datavolume_test.go
```diff
@@ -1447,7 +1454,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			}
 		})
 
-		It("Should create a new scratch PVC when PVC is deleted during import", func() {
+		PIt("Should create a new scratch PVC when PVC is deleted during import", func() {
```
awels, line 1457: The goal is to eventually not need scratch space anymore. However there are still some flows that require scratch space (like importing from registry). So we need this test to work. However the flow it uses to test this, no longer uses scratch space. So maybe you can modify this test to use:
```go
dv := utils.NewDataVolumeWithRegistryImport(dvName, "500Mi", tinyCoreIsoRegistryURL())
```
instead? Not sure if that will work, but worth a try.

awels (commented):
Looks pretty good, a few comments.

alicefr:
@awels many thanks for the review! The code got slimmer. I updated the code accordingly. Let's see if the CI is happy.

tests/datavolume_test.go
alicefr: I tried this suggestion but the importer is failing because of the registry certificate:
```
I0108 12:03:09.541298       1 importer.go:52] Starting importer
I0108 12:03:09.541366       1 importer.go:122] begin import process
I0108 12:03:09.541376       1 data-processor.go:350] Calculating available size
I0108 12:03:09.541382       1 data-processor.go:362] Checking out file system volume size.
I0108 12:03:09.541385       1 data-processor.go:370] Request image size not empty.
I0108 12:03:09.541391       1 data-processor.go:375] Target size 500Mi.
I0108 12:03:09.541419       1 data-processor.go:242] New phase: TransferScratch
I0108 12:03:09.541430       1 registry-datasource.go:83] Copying registry image to scratch space.
I0108 12:03:09.541433       1 transport.go:187] Downloading image from 'docker://cdi-docker-registry-host.cdi/tinycoreqcow2', copying file from 'disk' to '/scratch'
E0108 12:03:09.560972       1 transport.go:74] Could not create image reference: error pinging docker registry cdi-docker-registry-host.cdi: Get https://cdi-docker-registry-host.cdi/v2/: x509: certificate signed by unknown authority
E0108 12:03:09.560995       1 data-processor.go:239] Get https://cdi-docker-registry-host.cdi/v2/: x509: certificate signed by unknown authority
error pinging docker registry cdi-docker-registry-host.cdi
```
I need to check what is going wrong

tests/datavolume_test.go
awels: And when I say leave as is, I mean as it was originally by running the test. It should not fail.

tests/datavolume_test.go
awels: Check https://github.com/kubevirt/containerized-data-importer/blob/master/tests/datavolume_test.go#L98-L104 to see an example of how to setup the certificate (or maybe just call the function itself)

alicefr: Wit the right function and certificate, it works

tests/datavolume_test.go
alicefr: the test works and I changed the test description

alicefr:
@awels I adjusted the test cases as you suggested and the test suite is passing. Any other comments? :)

pkg/importer/http-datasource.go
```diff
@@ -115,17 +118,25 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
...
-	// The readers now contain all the information needed to determine if we can stream directly or if we need scratch space to download
-	// the file to, before converting.
-	if !hs.readers.Archived && !hs.customCA && !hs.brokenForQemuImg && hs.readers.Convert {
-		// We can pass straight to conversion from the endpoint. No scratch required.
-		hs.url = hs.endpoint
+	if hs.brokenForQemuImg {
+		return ProcessingPhaseTransferScratch, nil
+	}
+	hs.url = hs.endpoint
+	if !hs.readers.Archived && hs.customCA == "" && hs.readers.Convert {
```
awels, line 125: So this is not really a problem, but I think we can remove this check. With this check in place we will use qemu-img directly instead of nbdkit. If we remove this check we will use nbdkit for this particular case as well.

awels (approved):
/lgtm

PR #1510, opened 2020-12-01
Try updating the node taint in a loop

maya-r:
Sometimes other things will influence the nodes so the update will
fail on an outdated generation number.

Seen [here](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1503/pull-containerized-data-importer-e2e-k8s-1.18-nfs/1333427034675744768)

**Release note**:

```release-note
Stabilize functional test
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -215,14 +215,18 @@ var _ = Describe("Tests needing the restore of nodes", func() {
 		}
 
 		for _, node := range nodes.Items {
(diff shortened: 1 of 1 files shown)


awels:
Looks like the linter failed. if you run make test-lint it will let you know. gofmt -w will auto format the file.

awels:
looks good otherwise.

PR #1512, opened 2020-12-01
Adding support for import of vmdk and vdi file types

visheshtanksale:

Adds support to importer VMDK and VDI type of VM Images

Fixes #

Adding support to download and convert VMDK and VDI images
**Release note**:

```release-note
Adding support to import vmdk and vdi VM Images
```

### DIFF
--- a/pkg/image/filefmt.go
+++ b/pkg/image/filefmt.go
@@ -46,6 +46,18 @@ var knownHeaders = Headers{
 		SizeOff: 0,
 		SizeLen: 0,
 	},
+	"vmdk": Header{
+		Format:      "vmdk",
+		magicNumber: []byte{0x4B, 0x44, 0x4D, 0x56},
+		SizeOff: 0,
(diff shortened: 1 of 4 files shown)


awels:
Hi,
thanks for the PR. A few things to consider here:
- vmdk and vdi files will have drivers installed for their respective source hyper visors, and the conversion will not remove or modify them at all. So once the conversion is done, the VMs will likely be slower due to misconfigured drivers. In particular windows VMs will be like that.
- We documentation stating which combinations we support [here](https://github.com/kubevirt/containerized-data-importer/blob/master/doc/supported_operations.md) that will need to get updated to include the new formats.
- We also want unit tests and functional tests for the new formats. We usually use small images like cirros or tinyCore for the tests. At a minimum I would add some importing tests in tests/datavolume_test.go. There should be ample examples of similar tests there. Its a little tricky adding new formats to the test http server, but I can help with that.

PR #1513, opened 2020-12-02
Fix critical addons operator test to fully restore services before moving to next test

awels:

In the operator test there is a critical addons test that removes and
restarts the pod used in tests. It verified the pods where 'ready' but
not that the service using those pods had working endpoints. Due to the
changes where we start the next test without waiting for the namespace
to finish deleting, those services didn't have enough time to have their
endpoints be available. This ensures the endpoints are available before
moving onto the next tests.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -153,6 +153,23 @@ var _ = Describe("Tests needing the restore of nodes", func() {
 			Expect(err).ToNot(HaveOccurred())
 		}
 
+		Eventually(func() bool {
+			services, err := f.K8sClient.CoreV1().Services(f.CdiInstallNs).List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred(), "failed getting CDI services")
+			for _, service := range services.Items {
+				if service.Name != "cdi-prometheus-metrics" {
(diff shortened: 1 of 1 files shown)


tests/operator_test.go
```diff
@@ -153,6 +153,23 @@ var _ = Describe("Tests needing the restore of nodes", func() {
 			Expect(err).ToNot(HaveOccurred())
 		}
 
+		Eventually(func() bool {
+			services, err := f.K8sClient.CoreV1().Services(f.CdiInstallNs).List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred(), "failed getting CDI services")
+			for _, service := range services.Items {
+				if service.Name != "cdi-prometheus-metrics" {
```
arnongilboa, line 160: why check != cdi-prometheus-metrics?

awels: Because that one never has any endpoints, so checking for endpoints there will always fail. It only gets endpoints for certain specific tests.

awels:
/test pull-containerized-data-importer-e2e-k8s-1.18-upg

retesting because cluster failed during run. node02 lost its cidr and was unable to start the verifier pod because it couldn't get a network.

PR #1514, opened 2020-12-02
Have the verifier pod use the hostnetwork

awels:
as we don't care about it having an ip address as it doesn't use the network at all.

We are seeing random failures in verifier pods, and it appears to be when the cni plugin is unable to give the verifier pod an ip address. Since the verifier pod doesn't use the network at all, giving it the host network should eliminate the cni provider from the equation and hopefully improve stability.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
Apparently "hostNetwork: true" is a privileged operation on OpenShift, this might cause some problems if the tests are run in that environment.

PR #1516, opened 2020-12-03
Designate CDI as CDIConfig authority

mhenriks:

Formally designate CDI as CDIConfig authority by adding annotation `cdi.kubevirt.io/configAuthority` to CDI after copying values from existing CDIConfig.

If the annotation does not exist, the config controller will ignore values in CDI

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/config-controller_test.go
+++ b/pkg/controller/config-controller_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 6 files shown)


pkg/controller/config-controller_test.go
```diff
@@ -636,6 +658,9 @@ func createConfigReconciler(objects ...runtime.Object) (*CDIConfigReconciler, *c
 	cdi := &cdiv1.CDI{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "cdi",
+			Annotations: map[string]string{
+				"cdi.kubevirt.io/configAuthority": "",
```
awels, line 662: You made the const, you could use it here?

awels (approved):
looks good, one super minor nitpick
/lgtm

pkg/controller/config-controller_test.go
mhenriks: Yeah, but I think it is better for tests to have their own copy of these consts that way you catch any changes which may have external impact.

PR #1521, opened 2020-12-09
Clone source pod abort on tar error

mhenriks:

Tar errors were basically ignored as tar output was piped into the clone-source program.  Not the clone-source program executes tar and can therefore abort on error.  This will allow failures to be retried.

Also fixed a couple regressions in clone controller.  Should watch clone source pods.  Also better handle upload pod readiness.

Fixes #

This is one part of a two PR fix for https://bugzilla.redhat.com/show_bug.cgi?id=1893363

It does not address the root cause which is filesystem permission issues.  But it will correctly abort an upload when a issue tarring up a PVC occurs.

**Release note**:

```release-note
Bugfix: Clone source pod will detect tar errors and abort rather than silently ignore
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -131,6 +131,28 @@ func addCloneControllerWatches(mgr manager.Manager, cloneController controller.C
 	}); err != nil {
 		return err
 	}
+	if err := cloneController.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(obj handler.MapObject) []reconcile.Request {
+			target, ok := obj.Meta.GetAnnotations()[AnnOwnerRef]
+			if !ok {
+				return nil
+			}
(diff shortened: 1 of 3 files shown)


pkg/controller/clone-controller.go
```diff
@@ -308,12 +330,7 @@ func (r *CloneReconciler) waitTargetPodRunningOrSucceeded(pvc *corev1.Persistent
 		return false, errors.Wrapf(err, "error parsing %s annotation", AnnPodReady)
 	}
 
-	if !ready {
-		log.V(3).Info("clone target pod not ready")
-		return podSucceededFromPVC(pvc), nil
-	}
-
-	return true, nil
+	return ready || podSucceededFromPVC(pvc), nil
```
maya-r, line 333: this doesn't appear to be a funcional change.

maya-r:
I think that having release notes is desirable, suggesting:
BUGFIX: fix importer eventual consistency in cases where tar receives an error during the import (may happen with some storage classes when making multiple clones and using accessMode: ReadWriteOnly).

PR #1522, opened 2020-12-09
Add missing machine config for external open shift clusters. for hpp storage

awels:

We had a call to deploy the machine config, but never added it to the repo. This PR fixes that.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/external/provider.sh
+++ b/cluster-sync/external/provider.sh
@@ -21,7 +21,7 @@ function up() {
 
 function configure_storage() {
   if [[ $KUBEVIRT_STORAGE == "hpp" ]] ; then
-    _kubectl apply -f ./cluster-sync/external/resources/machineconfig-worker.yaml
+    _kubectl apply -f https://github.com/kubevirt/hostpath-provisioner-operator/blob/master/contrib/machineconfig-selinux-hpp.yaml
     echo "Installing hostpath provisioner storage, please ensure /var/hpvolumes exists and has the right SELinux labeling"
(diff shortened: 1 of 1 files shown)


maya-r:
How come this is different from https://github.com/kubevirt/hostpath-provisioner-operator/blob/master/contrib/machineconfig-selinux-hpp.yaml ?
is the master part not necessary?

awels:
So the master part is for if you want the hpp to also run on masters, by default the expectation is that hpp runs only on workers as that is where the VMs will be running.

mhenriks:
Did not work for me with crc 1.20.0

awels:
Actually now that we thought about it a bit more, it makes more sense to update the script to point to the yaml in the hpp repo, instead of duplicating it here.

cluster-sync/external/provider.sh
```diff
@@ -21,7 +21,7 @@ function up() {
 
 function configure_storage() {
   if [[ $KUBEVIRT_STORAGE == "hpp" ]] ; then
-    _kubectl apply -f ./cluster-sync/external/resources/machineconfig-worker.yaml
+    _kubectl apply -f https://github.com/kubevirt/hostpath-provisioner-operator/blob/master/contrib/machineconfig-selinux-hpp.yaml
```
mhenriks, line 24: Maybe we should check machinconfig crd installed?  SOmething similar to:

https://github.com/kubevirt/containerized-data-importer/blob/master/cluster-sync/ephemeral_provider.sh#L56-L58

PR #521, opened 2018-11-13
change api server name to CdiAPIServer

igoihman:
**What this PR does / why we need it**:
Since API Server isn't upload specific anymore we should change the name into  cdiAPIServer and cdiApiAppm according to @mhenriks suggestion in #500 .

**Release note**:

```release-note
NONE
```

PR #1524, opened 2020-12-10
Read-only mounts in clone source pods

mhenriks:

Read/write mounts will minimize errors when a source PVC is mounted multiple times on the same node.

Fixes #

This is two part of a two PR fix for https://bugzilla.redhat.com/show_bug.cgi?id=1893363

Once this and #521 are merged I will add a new functional test to stress cloning hard

**Release note**:

```release-note
Clone source pods mount PVCs read-only
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -240,10 +240,8 @@ func (r *CloneReconciler) reconcileSourcePod(sourcePod *corev1.Pod, targetPvc *c
 			return false, err
 		}
 
-		filtered := filterCloneSourcePods(pods)
-
-		if len(filtered) > 0 {
-			for _, pod := range filtered {
+		if len(pods) > 0 {
+			for _, pod := range pods {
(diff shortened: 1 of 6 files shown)


maya-r:
Can you elaborate in the commit message and/or split it into: read-only pod, then removing the clone source filtering (not currently explained in the commit message)?

looks good to me otherwise.

maya-r:
I'm not very knowledgeable about selinux. is this too much permissions to grant to a pod? the internet seems to imply it's giving a lot of access to the host.

mhenriks:
> I'm not very knowledgeable about selinux. is this too much permissions to grant to a pod? the internet seems to imply it's giving a lot of access to the host.

Clone source pods don't mount any path on the host, though.  Well, not including the PVC of course.  And that is mounted read only.

awels:
That is the same selinux context as KubeVirt gives the virt-launcher pods.

awels (approved):
/lgtm
/approve

PR #1528, opened 2020-12-13
[release-v1.23] fix travis build

maya-r:
It looks like it would be sufficient, but in case it isn't, there's [this archive](https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/31/Everything/x86_64/os/Packages/)

```release-note
BUGFIX: fix build failing as some Fedora servers have moved RPMs fetched as part of the build process to the archive.
```


maya-r:
that's not what I meant, oops :-)

PR #1532, opened 2020-12-15
Allow passing default multus network annotation to transfer pods

arnongilboa:

Multus adds an extra nic to the pod as passed by annotation `k8s.v1.cni.cncf.io/networks` from the dv, but to set it as default nic of the pod we need to pass `v1.multus-cni.io/default-network` annotation instead.

**Release note**:
```release-note
NONE
```


awels (approved):
/lgtm
/approve

PR #1535, opened 2020-12-16
Preallocation check all paths

tomob:

This PR makes sure all import/upload paths use preallocation.

**Release note**:

```release-note
NONE
```


doc/cdi-config.md
```diff
@@ -16,6 +16,9 @@ Currently it is used only for holding Upload Proxy URL details.
 | filesystemOverhead      |                       | How much of a Filesystem volume's space should be reserved for overhead related to the Filesystem. |
 |   global                | "0.055"               | The amount to reserve for a Filesystem volume unless a per-storageClass value is chosen. |
 |   storageClass          | nil                   | A value of `local: "0.6"` is understood to mean that the overhead for the local storageClass is 0.6. |
+| preallocation           |                       | If set, the space for DataVolumes is preallocated when DVs are created |
+|   global                | false                 | Preallocatin setting to use unless a per-storageClass or per-dataVolume value is set |
```
brybacki: Preallocatin - Preallocation

doc/cdi-config.md
```diff
@@ -25,4 +28,6 @@ Currently it is used only for holding Upload Proxy URL details.
 | filesystemOverhead      |                       | updated when the spec values are updated, to show the per-storageClass calculated result as well as the per-storageClass one. |
 |   global                | "0.055"               | The calculated overhead to be used for all storageClasses unless a specific value is chosen for this storageClass |
 |   storageClass          |                       | The calculated overhead to be used for every storageClass in the system, taking into account both global and per-storageClass values. |
-
+| preallocation           |                       | updated when the spec values are updated, to show the per-storageClass calculated result as well as the per-storageClass one. |
```
brybacki: I do not understand this part "to show the per-storageClass calculated result as well as the per-storageClass one."

doc/cdi-config.md
```diff
@@ -25,4 +28,6 @@ Currently it is used only for holding Upload Proxy URL details.
...
+|   global                | false                 | The calculated preallocation setting for all dataVolumes unless a per-storageClass or per-dataVolume value is set |
+|   storageClass          | nil                   | The calculated preallocation setting to be used for every storageClass in the system, taking into account both global and per-storageClass values. |
```
brybacki: So is this a map or a single value?

doc/preallocation.md
```diff
@@ -0,0 +1,75 @@
...
+  preallocation: true
+```
+
+## Enabling preallocation globally
+
+Preallocation can be also turned on for all DataVolumes with the following CDIConfig entry:
+
+```yaml
+apiVersion: cdi.kubevirt.io/v1beta1
+kind: CDIConfig
```
brybacki, line 39: This needs to updated to reflect the move of cfg to CDI CR

pkg/controller/util.go
```diff
@@ -616,3 +624,62 @@ func SetPodPvcAnnotations(pod *v1.Pod, pvc *v1.PersistentVolumeClaim) {
...
+func GetStorageClassNameForDV(c client.Client, dv *cdiv1.DataVolume) string {
+	// If DV has a SC, return it
+	if dv != nil && dv.Spec.PVC != nil && dv.Spec.PVC.StorageClassName != nil && *dv.Spec.PVC.StorageClassName != "" {
+		return *dv.Spec.PVC.StorageClassName
+	}
+
+	// If DV's PVC has a SC, return it
+	pvc := &v1.PersistentVolumeClaim{}
+	// TODO change when PVC's name is different from DV's name
+	err := c.Get(context.TODO(), types.NamespacedName{Name: dv.Name, Namespace: dv.Namespace}, pvc)
```
brybacki, line 654: I see more comments like that: "// TODO change when PVC's name is different from DV's name", and I am guilty of adding at least one ;)
I suppose this calls for a function GetPvcForDv or something like that. Might be good candidate for a cleanup PR.

pkg/controller/util_test.go
```diff
@@ -394,6 +608,75 @@ func createCDIConfigWithStorageClass(name string, storageClass string) *cdiv1.CD
 	}
 }
 
+func createCDIConfigWithStorageClassPreallocation(storageClass string, preallocation bool) *cdiv1.CDIConfig {
+	return &cdiv1.CDIConfig{
```
brybacki: What do you think about having a default empty CDI, and just adding &cdiv1.Preallocation{} part in the following 3 functions

pkg/controller/upload-controller.go
```diff
@@ -238,6 +241,13 @@ func (r *UploadReconciler) reconcilePVC(log logr.Logger, pvc *corev1.PersistentV
 		if podRestarts > pvcAnnPodRestarts {
 			anno[AnnPodRestarts] = strconv.Itoa(podRestarts)
 		}
+
+		if pod.Status.ContainerStatuses[0].State.Terminated != nil &&
+			pod.Status.ContainerStatuses[0].State.Terminated.ExitCode == 0 {
+			if strings.Contains(pod.Status.ContainerStatuses[0].State.Terminated.Message, "Preallocation applied") {
```
brybacki, line 247: I am not sure we have a const for common messages, but this is a good candidate, it is repeated in a few places

tomob: Good point.

pkg/controller/util.go
tomob: I would qualify such a task a technical debt.

awels (approved):
/lgtm
/approve

PR #1548, opened 2020-12-23
Stop Using Deprecated Packages

nunnatsa:
sigs.k8s.io/controller-runtime/pkg/runtime/* packages are deprecated, and were moved to new paths.

Trying to upgrade sigs.k8s.io/controller-runtime to version v0.7.0 in HCO created a conflict because in v0.7.0 the deprecated packages were removed and cannot be used.

This PR replaces the deprecated packages with their new paths.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
missing a change to go.mod and go.sum? the way we update also involves invoking `make deps-update`

nunnatsa:
> missing a change to go.mod and go.sum? the way we update also involves invoking `make deps-update`

Thanks for the tip. Was naive to just run `go mod tidy & go mod vendor`.

maya-r:
I'm confused about why there are still no changes to go.mod and go.sum

maya-r:
oh, the packages have a way to write code that works on both versions, got it.

nunnatsa:
> oh, the packages have a way to write code that works on both versions, got it.

Yes, The `runtime/* ` packages were are deprecated and were moved to a new locations. in version 0.7.0 they are removed.

awels:
Why would HCO care about internals of CDI? I am not opposed to this PR at all. But I want to understand why HCO had a conflict. The HCO should be using the generated CDI client and not care about the internals of CDI.

nunnatsa:
> Why would HCO care about internals of CDI? I am not opposed to this PR at all. But I want to understand why HCO had a conflict. 

HCO need to build a CDI object and create/update it in K8s. So HCO uses CDI data types.

> The HCO should be using the generated CDI client and not care about the internals of CDI.
Well, it does not

PR #1551, opened 2020-12-28
Change verbosity for preallocation messages, avoid infinite loop

akalenyu:
- D/S the importer pod has verbosity=1 => the preallocation messages are not present in log
- Due to this, we hit an infinite loop in the preallocation tests
  (Import completed without preallocation message => pod doesn't exist anymore but sampling continues)

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -793,6 +793,9 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 
 var _ = Describe("Preallocation", func() {
 	f := framework.NewFramework(namespacePrefix)
+	dvName := "import-dv"
+	importerPodName := "importer-" + dvName
+	preAllocAdded := "Added preallocation"
 
 	var (
 		dataVolume     *cdiv1.DataVolume
@@ -815,7 +818,6 @@ var _ = Describe("Preallocation", func() {
 	})
 
 	It("Importer should add preallocation when requested", func() {
-		dvName := "import-dv"
 		By(fmt.Sprintf("Creating new datavolume %s", dvName))
 		dv := utils.NewDataVolumeWithHTTPImport(dvName, "100Mi", tinyCoreIsoURL())
 		preallocation := true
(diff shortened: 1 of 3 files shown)


maya-r:
/ok-to-test
/lgtm
/approve
/cherrypick release-v1.28
It would be nice to have a look at a different way of solving the log problem, but happy to have a workaround for now.

tests/upload_test.go
```diff
@@ -1027,10 +1028,12 @@ var _ = Describe("Preallocation", func() {
 
 		Eventually(func() string {
 			for {
-				log, err := tests.RunKubectlCommand(f, "logs", "cdi-upload-"+dvName, "-n", f.Namespace.Name)
+				log, err := tests.RunKubectlCommand(f, "logs", uploaderPodName, "-n", f.Namespace.Name)
 				if err != nil {
 					time.Sleep(1 * time.Second)
-					continue
+					if _, err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Get(context.TODO(), uploaderPodName, metav1.GetOptions{}); !k8serrors.IsNotFound(err) {
+						continue
```
maya-r, line 1035: Would it work just as well to return "" on error? it will cause the Eventually timeout to be hit, but won't ever contain the substring resulting in success

maya-r:
/hold
for one suggestion to avoid the infinite loop

tests/import_test.go
```diff
@@ -854,16 +852,14 @@ var _ = Describe("Preallocation", func() {
...
-				if err != nil {
-					time.Sleep(1 * time.Second)
-					continue
-				}
-				if strings.Contains(log, "Import complete") {
-					return log
-				}
+			log, err := tests.RunKubectlCommand(f, "logs", importerPodName, "-n", f.Namespace.Name)
+			if err != nil {
+				return preAllocAdded
```
maya-r, line 857: return "" please :-)
this will cause the test to pass when it shouldn't

akalenyu: this one is about `ShouldNot(ContainSubstring(preAllocAdded))`, we want it to keep looping if error/import not complete
wdyt?

maya-r: Ah, I got it mixed up. so I guess this is two conditions. no string "preallocation added" and yes string "import complete".
I think this might be nicer:

var log string
var err error
Eventually(func() string {
    log, err = tests.RunKubectlCommand(f, "logs", importerPodName, "-n", f.Namespace.Name)
    if err != nil {
        return ""
    }
    return log
}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(ContainSubstring("Import complete")
Expect(log).ToNot(ContainSubstring("preallocation added")

awels:
Any particular reason the verbosity is at 1? We have several other tests that require the verbosity to be at 3 in order for them to work. I am not sure why preallocation would be different.

akalenyu:
> Any particular reason the verbosity is at 1? We have several other tests that require the verbosity to be at 3 in order for them to work. I am not sure why preallocation would be different.

Do you mean why HCO sets --v=1 on our containers such as `cdi-importer` in the first place?

awels:
I know it doesn't but that doesn't mean you can't change the verbosity when running the tests on the HCO level. We have other tests that rely on the verbosity being 3 in order for them to work. I know the testing infra sets it to 3 when running the tier 1 tests, so I am trying to understand why that didn't work for you

PR #1552, opened 2020-12-28
Feat CSI Clone Support

### LINKED ISSUE #1467: Efficient cross namespace PVC cloning
> /kind enhancement

**Abstract**:
Current PVC cloning between namespaces copies the data between the PVCs. This is both slow, and a waste of resources. The current in-namespace cloning depends on Volume Snapshots. The snapshot based clone can be replaced by [CSI Volume Cloning](https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/) which will lead to faster clones requiring less ceremony.

The CSI Volume Cloning can also be leveraged to do instant, zero-copy cross-namespace volume cloning.

**Implementation**:
The cross-namespace cloning can be achieved in a semi hacky way using the following process:

Assume the cloned PVC is `A` and the target is a new PVC named `C` in a different namespacae.

- Create a new PVC, `B` in the source namespace with `dataSource` set to PVC `A`.
- Once created. Patch the newly created PVC `B` with `{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`
- Store the name of the `PersistentVolume` allocated to `B` in memory.
- Delete PVC `B`
- Patch the `PersistentVolume` and remove `spec.claimRef`
- Create the new PVC `C` in the target namespace with `volumeName` set to the name of PV backing the now deleted PVC
- Patch PVC `C` with `{"spec":{"persistentVolumeReclaimPolicy":"Reclaim"}}'`

To ensure security, apply the same permission checks as for the current copy based cross-namespace volume cloning. This method requires `storageClass` to be the same. If they differ, fall back to the copy method.

YitzyD:
This PR implements CSI volume cloning for PVC sourced DataVolumes as described in issue [1467] https://github.com/kubevirt/containerized-data-importer/issues/1467). 
Documentation describing the cloning process is detailed in [csi-cloning](https://github.com/YitzyD/containerized-data-importer/blob/yd/csi-clone/doc/csi-cloning.md) and [efficient-cloning](https://github.com/YitzyD/containerized-data-importer/blob/yd/csi-clone/doc/efficient-cloning.md).
CSI driver backed volume cloning enables incredibly efficient and reliable cloning.    
Closes #1467 

I was unsure of the labeling scheme for cloning resources, so I attempted to match those used in the already implemented cloning processes. These annotations are defined: https://github.com/YitzyD/containerized-data-importer/blob/4bf9d06973d9a398c791e1e077c558195197932d/pkg/controller/csi-clone-controller.go#L31-L39
**Release note**:

```release-note
  [action required]: In order for CSI volume cloning to be enabled for a storage class a cluster admin will have to annotate the storage class with "cdi.kubevirt.io/CSICloneVolumeCapable = true"
```


mhenriks:
Hi @YitzyD thanks for looking into this!  Up until now, our intention has been to leverage a community driven namespace transfer API to facilitate cross-namespace cloning.  See the latest KEP here:  https://github.com/kubernetes/enhancements/pull/2110

IMO leveraging a dedicated namespace transfer API would still be preferable.  Even if it is not an API that has been blessed by the community.

salanki:
I do agree that having this driven by Kubernetes itself would be preferable @mhenriks. However, the path to getting that merged into a Kubernetes release is rather long, and then there is the additional delay of getting production clusters upgraded. @YitzyD is offering you a working implementation that can be replaced by KEP 2110 in the future without any end-user changes. I’d really like to have this merged now, and we can pull it out once KEP 2110 is in a widely adopted Kubernetes release.

mhenriks:
We don't have to wait for the upstream k8s community to have an implementation, but I would prefer that CDI have it's own namespace transfer api/controller.  The datavolume/smart clone controller will then invoke that API as part of the CDI clone operation.  I understand that it is more work now.  But it will be less work if/when there is a namespace transfer implementation upstream.  

I could see this dedicated namespace transfer api/controller having tremendous value not just for CDI while things come together upstream.  What do you think @salanki and @YitzyD?

salanki:
Commenting on behalf of me and my colleague @YitzyD:  As you mentioned, your `StorageTransfer` API is basically what @YitzyD has implemented, but would be an external controller that CDI has to interact with. For the purpose of cross namespace cloning in CDI, we do not see much benefit in breaking it out into an external controller as the implementation is already done inside CDI with this PR. If another contributor wants to do the refactor to break out the code into a separate controller, we welcome that. We'd love to have this PR merged in its current form (after addressing any additional review comments and tests) and focus on other contributions to the project!

mhenriks:
@salanki I assure you that we have a mutual interest in making cross-namespace cloning perform faster.  We also share and appreciate your urgency.  But as maintainers, we also want to reduce churn in the long run.

What I propose (and other maintainers like @awels @aglitke please chime in) is that your `CSICloneReconciler` becomes something like `DataVolumeTransferReconciler`.  Most of the implementation logic is already there.  But it will watch a new (cluster scoped) custom resource in `cdi.kubevirt.io` apiGroup.  The spec has source/target names and the status has the current state of the transfer.  Something like this:

```yaml
apiGroup: cdi.kubevirt.io
kind: DataVolumeTransfer
metadata:
   name: my-transfer
spec:
    source:
        namespace: source-ns
        name: source-name
     target:
        namespace:  target-ns
        name: target-name
```yaml

The existing datavolume controller/smart clone controller will create a temporary clone PVC in the source namespace then create a `DataVolumeTransfer` resource to transfer to the target namespace and wait until completed.  All of this logic basically exists if I read the PR correctly.

What do you think?

I welcome and encourage your contributions.  And in case you are not already aware, we have weekly community meetings on Wednesdays.  We can discuss this and any other issues you may have in real time:  https://github.com/kubevirt/community/

aglitke:
It's really great to see this PR!  One concern I have is that we would like to see namespace transfer work also with smart clone (cloning via snapshots).  Some storage systems (ie. ceph) are able to clone more efficiently if a snapshot is used rather than a CSI clone.  It's important to me that users will have a consistently good experience regardless of the cloning strategy used.  I think @mhenriks suggestion of using a DataVolumeTransfer CR to independently manage the namespace piece fits really well into these goals.

salanki:
In our experience the Snapshot based clones are slower and less reliable than CSI Clone. We have had a lot of issues with Snapshot cloning on both CEPH Luminos and Octopus, which is why we are using the CSI clone for both local namespace and cross namespace clones in this PR.

In what way would the snapshot clones be more efficient?

YitzyD:
Additionally, CSI clone is an opt in feature on a per storage class basis using the `"cdi.kubevirt.io/CSICloneVolumeCapable"` annotation. https://github.com/kubevirt/containerized-data-importer/blob/d778d1f3c14ec4aa51f5facb72c97c9f91138356/pkg/controller/csi-clone-controller.go#L43 
A cluster admin would be responsible to research the underlying storage driver and make that educated decision.

mhenriks:
@salanki direct clone vs snapshot clone performance is largely implementation dependent and kind of beside the point.  CDI wants to support both.  And would like to have a namespace transfer utility that can be leveraged by both.   If you want to keep direct clone support in this PR, that is fine.  But please break out the namespace transfer part.

salanki:
@aglitke: I'd really like to understand your comment about snapshot cloning being more efficient on CEPH than the CSI clone, as we've seen the exact opposite. Can you give me some more detail?

salanki:
Any chance you guys will move this PR forward without a rewrite? We currently do not have the capacity to change this implementation.

salanki:
Pinging again on possibility to get this merged.

awels:
@mhenriks is working hard on the transfer piece, AFAIK he will post a PR soon that can be used by everyone, not just CSI clone. As for the more efficient flow @aglitke was talking about. The idea is that we don't delete the snapshot after a snapshot clone. The most expensive operation is creating the snapshot (according to ceph devs we talked to), afterwards creating a new PVC from the snapshot is relatively cheap. The current snapshot clone implementation deletes the snapshot after it completes the clone, but if we create an API to tell it to keep the snapshot, we could efficiently create many clones from a single snapshot.

Our biggest objection to this PR is the intermingling of clone and the transfer in the same controller, which makes the transfer not usable for the snapshot cloning, we want the transfer to be useful for all cloning methods.

salanki:
If you create a new PVC from a snapshot, will a new RBD image be derived from it? For some reason I thought that the snapshot is the RBD image then, so if you create multiple PVCs from the same snapshot they point to the same image.

awels:
We have no aversion to the clone API, we want to support it. In fact we have an API where you can specify which method you prefer if multiple are available. But we want to support host assisted clone, snapshot clone, csi clone, and the modified snapshot clone I talked about.

We just want the transfer to work for all those methods as well.

mhenriks:
@salanki namespace transfer API here:  https://github.com/kubevirt/containerized-data-importer/pull/1673

If you do not have the time to update this PR to use it, I will fork this branch and do it myself.

mhenriks:
@salanki @YitzyD Without clone support in CI, we cannot go forward with this. So, first step is to update our kubevirtci with a later version of rook/ceph that supports CSI cloning.  Working on that now.

PR #1556, opened 2021-01-08
update cdi config docs

mhenriks:

Update docs to reference setting on `CDI`.  Also added examples.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/core/v1beta1/types.go
+++ b/pkg/apis/core/v1beta1/types.go
@@ -400,7 +400,7 @@ type CDIConfigSpec struct {
 	// FilesystemOverhead describes the space reserved for overhead when using Filesystem volumes. A value is between 0 and 1, if not defined it is 0.055 (5.5% overhead)
 	FilesystemOverhead *FilesystemOverhead `json:"filesystemOverhead,omitempty"`
 	// Preallocation controls whether storage for DataVolumes should be allocated in advance.
-	Preallocation bool `json:"preallocation,omitempty"`
+	Preallocation *bool `json:"preallocation,omitempty"`
 }
 
 //CDIConfigStatus provides the most recently observed status of the CDI Config resource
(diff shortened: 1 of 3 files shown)


pkg/apis/core/v1beta1/types.go
```diff
@@ -377,7 +377,7 @@ type CDIConfigSpec struct {
 	// FilesystemOverhead describes the space reserved for overhead when using Filesystem volumes. A value is between 0 and 1, if not defined it is 0.055 (5.5% overhead)
 	FilesystemOverhead *FilesystemOverhead `json:"filesystemOverhead,omitempty"`
 	// Preallocation controls whether storage for DataVolumes should be allocated in advance.
-	Preallocation bool `json:"preallocation,omitempty"`
+	Preallocation *bool `json:"preallocation,omitempty"`
```
maya-r, line 380: Is this change intentional? it seems unrelated.

mhenriks: Intentional and somewhat related as I was reviewing the config params.  This property is currently unreferenced, though.  But I assume @tomob has plans for it.

tomob: What do you mean "unreferenced"? What about that? https://github.com/kubevirt/containerized-data-importer/blob/d31591de94dc2c210b55cf6755fb55fb11397e64/pkg/controller/util.go#L639

mhenriks: @tomob I updated the spec, not the status

tomob: Oh, crap.

What I actually wanted to use there is the spec. It cased the bug I talked about today: https://github.com/kubevirt/containerized-data-importer/pull/1565

mhenriks: @tomob so do you want to change to `*bool` in your PR? Or keep here?

mhenriks: It should be `*bool` because it is optional.  The status field can remain `bool`

awels: Also make sure to run make generate, this changes the generated deep copy

awels:
I think we should fix the *bool in #1565 especially since that needs to get fixed in some other way.

mhenriks:
@awels I reverted the change to `preallocation`

PR #1557, opened 2021-01-10
Compare logs while ignoring differences in spaces

maya-r:
Some OpenShift versions seem to not add spaces between different
components of the message, become more resilient for this scenario.

**Release note**:
```release-note
Fix some tests failing in newer versions of OpenShift due to whitespace differences
```


awels (approved):
/approve
/lgtm

PR #1559, opened 2021-01-12
core: Preallocate blank block volumes

tomob:

Importer pod needs to be started for blank block volumes and it needs to
handle the case.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -259,3 +259,17 @@ func (o *qemuOperations) CreateBlankImage(dest string, size resource.Quantity, p
 
 	return nil
 }
+
+// PreallocateBlankBlock writes requested amount of zeros to block device mounted at dest
+func PreallocateBlankBlock(dest string, size resource.Quantity) error {
+	klog.V(3).Infof("block volume size is %s", size.String())
+
+	args := []string{"if=/dev/zero", "of=" + dest, "bs=" + convertQuantityToQemuSize(size), "count=1"}
+	_, err := qemuExecFunction(nil, nil, "dd", args...)
+
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Could not preallocate blank block volume at %s with size %s", dest, size.String()))
+	}
+
+	return nil
+}
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -965,5 +965,12 @@ var _ = Describe("Preallocation", func() {
 		Entry("Blank image", true, func() *cdiv1.DataVolume {
 			return utils.NewDataVolumeForBlankRawImage("import-dv", "100Mi")
 		}),
+		Entry("Blank block DataVolume", true, func() *cdiv1.DataVolume {
+			if !f.IsBlockVolumeStorageClassAvailable() {
+				Skip("Storage Class for block volume is not available")
+			}
+
+			return utils.NewDataVolumeForBlankRawImageBlock("import-dv", "100Mi", f.BlockSCName)
+		}),
 	)
 })
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -108,7 +108,14 @@ func main() {
 			}
 		}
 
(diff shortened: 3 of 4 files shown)


maya-r:
Sorry to be asking this question rather late -- why do we want to preallocate block disks?
the code looks good to me.

maya-r:
OK, I can answer my own question: because if someone asks for preallocation it's surprising if it doesn't do that in some scenarios.
/lgtm

pkg/image/qemu.go
```diff
@@ -259,3 +259,17 @@ func (o *qemuOperations) CreateBlankImage(dest string, size resource.Quantity, p
 
 	return nil
 }
+
+// PreallocateBlankBlock writes requested amount of zeros to block device mounted at dest
+func PreallocateBlankBlock(dest string, size resource.Quantity) error {
+	klog.V(3).Infof("block volume size is %s", size.String())
+
+	args := []string{"if=/dev/zero", "of=" + dest, "bs=" + convertQuantityToQemuSize(size), "count=1"}
```
awels, line 267: Any particular reason you are not calling CreateBlankDisk with a preallocate=full?  I noticed we already added a preallocation=falloc, why not just allow full with a different value parameter?

tests/import_test.go
```diff
@@ -965,5 +965,12 @@ var _ = Describe("Preallocation", func() {
 		Entry("Blank image", true, func() *cdiv1.DataVolume {
 			return utils.NewDataVolumeForBlankRawImage("import-dv", "100Mi")
 		}),
+		Entry("Blank block DataVolume", true, func() *cdiv1.DataVolume {
+			if !f.IsBlockVolumeStorageClassAvailable() {
+				Skip("Storage Class for block volume is not available")
+			}
+
+			return utils.NewDataVolumeForBlankRawImageBlock("import-dv", "100Mi", f.BlockSCName)
```
awels, line 973: A full alloc should be pretty slow, so there should be enough time to verify that a pod actually was created.

pkg/image/qemu.go
tomob: What's `CreateBlankDisk`? There doesn't seem to be anything like that in the code.

If you meant `CreateBlankImage` – do we want to use qemu-img at this case? It's just a raw block device. Also, quickly checking, it seems that qemu-img simply does not support it, at least for some storage (local in this case):

```
> cluster-up/ssh.sh node01 "sudo qemu-img create -f raw /dev/loop0 100M -o preallocation=falloc"
2021/01/15 14:41:05 Waiting for host: 192.168.66.101:22
2021/01/15 14:41:05 Connected to tcp://192.168.66.101:22
qemu-img: /dev/loop0: Unsupported preallocation mode 'falloc'
Formatting '/dev/loop0', fmt=raw size=104857600 preallocation=falloc

> cluster-up/ssh.sh node01 "sudo qemu-img create -f raw /dev/loop0 100M -o preallocation=full"
2021/01/15 14:41:08 Waiting for host: 192.168.66.101:22
2021/01/15 14:41:08 Connected to tcp://192.168.66.101:22
Formatting '/dev/loop0', fmt=raw size=104857600 preallocation=full
qemu-img: /dev/loop0: Unsupported preallocation mode 'full'

```

Am I missing something?

tests/import_test.go
tomob: The default 270 seconds should be enough for anybod^W writing 100Mi of zeros.

pkg/image/qemu.go
awels: No you are correct, you can't create a block device, it exists already, so we need to dd.

tests/import_test.go
awels: Well the new part here is that for a block device we DO create a pod if preallocation is set to true. So our test should actually verify the pod is created.

tomob: The test verifies that the preallocation annotation has been created, and it's only created when the pod started AND finished successfully.

awels: Gotcha, didn't realize the annotation was created only if the pod started and finished.

awels (commented):
/lgtm
/approve

PR #1560, opened 2021-01-12
Add a DV/PVC annotation "storage.bind.immediate.requested"

brybacki:

When the annotation is applied the CDI will force bind the PVC (by scheduling worker pods), ignoring the logic to handle WaitForFirstConsumer binding mode.

This is useful when uploading "template" images to the cluster on local storage with WaitForFirstConsumer binding. In this case the image has to be available somewhere on the cluster, the actual placement of image on specific node is not important, so the CDI worker node can be used as a first consumer.

For storage with immediate binging it is effectively a NOOP.

Fixes #

- [x] Upload done, import and clone will follow shortly 
- [x] tests needed

**Release note**:

```release-note
Add an annotation "storage.bind.immediate.requested"
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -179,16 +179,17 @@ func addImportControllerWatches(mgr manager.Manager, importController controller
 }
 
 func shouldReconcilePVC(pvc *corev1.PersistentVolumeClaim,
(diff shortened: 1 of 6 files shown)


awels (approved):
/lgtm
/approve

PR #1561, opened 2021-01-12
core: Preallocate blank image disks as well

tomob:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1914177

This fixes a bug where blank image disks were not preallocated.

This is a cherry-pick from master to v28, including all dependencies
required for functional tests but without new functionality.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/data-processor.go
+++ b/pkg/importer/data-processor.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strconv"
 
 	"github.com/pkg/errors"
 
@@ -120,6 +121,10 @@ type DataProcessor struct {
 	needsDataCleanup bool
 	// preallocation is the flag controlling preallocation setting of qemu-img
 	preallocation bool
+	// preallocationApplied is used to pass information whether preallocation has been performed, skipped or not attempted
+	// "skipped" is used to indicate that preallocation would have been perfomed but there was not enough space, so the
+	// preallocation whould have failed.
+	preallocationApplied string
 }
 
 // NewDataProcessor create a new instance of a data processor using the passed in data provider.
@@ -259,6 +264,7 @@ func (dp *DataProcessor) convert(url *url.URL) (ProcessingPhase, error) {
 	if err != nil {
 		return ProcessingPhaseError, errors.Wrap(err, "Conversion to Raw failed")
 	}
+	dp.preallocationApplied = strconv.FormatBool(dp.preallocation)
 
 	return ProcessingPhaseResize, nil
 }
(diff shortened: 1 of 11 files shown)


pkg/importer/data-processor.go
```diff
@@ -283,14 +289,18 @@ func (dp *DataProcessor) resize() (ProcessingPhase, error) {
 			err = errors.Wrap(err, "Unable to change permissions of target file")
 		}
 	}
-	if dp.preallocation && shouldPreallocate {
-		return ProcessingPhasePreallocate, nil
+	if dp.preallocation {
+		if shouldPreallocate {
+			return ProcessingPhasePreallocate, nil
+		}
+		dp.preallocationApplied = "skipped" // qemu did not preallicate space for a resized file
```
maya-r, line 296: preallicate->preallocate

pkg/uploadserver/uploadserver.go
```diff
@@ -400,14 +407,15 @@ func newAsyncUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string,
 	return processor, processor.ProcessDataWithPause()
 }
 
-func newUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string, filesystemOverhead float64, preallocation bool, contentType string) error {
+func newUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string, filesystemOverhead float64, preallocation bool, contentType string) (string, error) {
 	if contentType == common.FilesystemCloneContentType {
-		return filesystemCloneProcessor(stream, common.ImporterVolumePath)
+		return "false", filesystemCloneProcessor(stream, common.ImporterVolumePath)
```
maya-r, line 412: Is there a strong reason to use strings that are booleans? it's a bummer to lose the benefit of type-checking

tomob: It's "true", "false" or "skipped".

I could do a type redefinition here (like `type PreallocationStatus string`), but it seemed to me like an overkill.

awels: In Go, in general it is very simple to do the redefinition, and it adds some type safety, in particular with typos in the strings. maybe you can do:
```go
type Preallocation string

const PreallocationYes Preallocation="true"
const PreallocationNo Preallocation="false"
const PreallocationSkip Preallocation="skipped"
```
And you can pass the Yes/No instead of the booleans, and return Yes/No/Skip for the applied. I think that would make it easier to read. I had to re-read a few times before I found the place where the value was set to none.

tomob: This is a cherry-pick for 1.28. I don't want to change this here, let me create a new PR for that.

PR #1564, opened 2021-01-13
VDDK: work with block devices better (BZ 1913756).

### LINKED ISSUE #1563: VDDK: block device destinations do not work correctly (BZ 1913756)
/kind bug

**What happened**: VDDK import to a block device destination failed on the first block of all zeroes.

**What you expected to happen**: Successful VDDK import to block device.

**How to reproduce it (as minimally and precisely as possible)**: Create a PV backed by a block device, and create a VDDK DV pointing to it. Most source disks will fail as long as the disk is not completely full.

**Anything else we need to know?**: Fix is in progress.

mrnold:
**What this PR does / why we need it**: Addresses BZ 1913756, where VDDK disk imports fail with block device destinations.

Fixes #1563 

**Release note**:
```release-note
BZ 1913756: fix VDDK import to block device destination volumes.
```

### DIFF
--- a/pkg/importer/vddk-datasource.go
+++ b/pkg/importer/vddk-datasource.go
@@ -37,6 +37,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
+	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/util"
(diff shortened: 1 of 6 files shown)


pkg/importer/vddk-datasource.go
```diff
@@ -668,6 +680,37 @@ func (sink *VDDKFileSink) Write(buf []byte) (int, error) {
...
+func (sink *VDDKFileSink) ZeroRange(offset uint64, length uint32) error {
+	var err error
+	if !sink.isBlock { // For files, see if it's possible to punch a hole
+		err := syscall.Fallocate(int(sink.file.Fd()), unix.FALLOC_FL_ZERO_RANGE, int64(offset), int64(length))
+		if err != nil {
+			klog.Errorf("Unable to zero range on destination, falling back to pwrite: %v", err)
+		}
+	} // For block devices, just use pwrite for now. Replace with BLKZEROOUT later if possible.
+
+	if err != nil { // Fall back to regular pwrite
```
maya-r, line 693: I'm surprised that this code block doesn't write in the block case.
Is that intentional?

fdupont-redhat:
It seems that the failures are not related to VDDK import: 1.17-hpp failed to deploy to ImagePullBackOff, and 1.18-upg failed on a cloning test.

/retest

pkg/importer/vddk-datasource.go
mrnold: It was intentional to avoid the fallocate on block devices, but also wrong - it looks like ZERO_RANGE is actually supported on block devices after all, so I will update it to do that.

The original intention was: for files, try ZERO_RANGE and fall back to pwrite if it fails; for block devices, always skip straight to pwrite. The pwrite fallback is supposed to reset err, so I will also make sure it does that.

mrnold:
/hold

One more commit incoming.

tests/datavolume_test.go
```diff
@@ -855,6 +855,12 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			switch name {
 			case "import-http":
 				dataVolume = utils.NewDataVolumeWithHTTPImportToBlockPV(dataVolumeName, "1G", url(), f.BlockSCName)
+			case "import-vddk":
+				dataVolume = createVddkDataVolume(dataVolumeName, "1Gi", vcenterURL())
+				utils.ModifyDataVolumeWithVDDKImportToBlockPV(dataVolume, f.BlockSCName)
```
awels, line 860: Do we want to test this with a file system import as well? The code looks like it cares what volume mode the target is so we should test file system volumes as well.

mrnold: It should already be getting tested with a filesystem import: https://github.com/kubevirt/containerized-data-importer/blob/7783203b5a027904166c172fb941fd35a72dddb7/tests/datavolume_test.go#L728

mrnold: I'm not sure I understood the original request, so I should clarify: I added cold and warm VDDK import tests to the only existing block device test section I saw in datavolume_test.go, and there were already cold and warm VDDK import tests for filesystem volumes elsewhere in the same file from a previous pull request. So now there are cold and warm VDDK tests for both block device and filesystem volumes.

Does that cover it, or were you looking for a different kind of test entirely?

awels: No that covers it, I just wanted to make sure we had both block and file system tests for both warm and regular vddk imports.

awels (approved):
/lgtm
/approve

PR #1565, opened 2021-01-14
Global preallocation setting is not taken into account correctly.

tomob:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1915704

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/core/v1beta1/types.go
+++ b/pkg/apis/core/v1beta1/types.go
@@ -377,7 +377,7 @@ type CDIConfigSpec struct {
 	// FilesystemOverhead describes the space reserved for overhead when using Filesystem volumes. A value is between 0 and 1, if not defined it is 0.055 (5.5% overhead)
 	FilesystemOverhead *FilesystemOverhead `json:"filesystemOverhead,omitempty"`
 	// Preallocation controls whether storage for DataVolumes should be allocated in advance.
-	Preallocation bool `json:"preallocation,omitempty"`
+	Preallocation *bool `json:"preallocation,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


awels (changes requested):
So this does not actually fix the problem. We should be getting the status.Preallocation, which is calculated from spec.Preallocation in the CDIConfig controller. So the fix would be to properly calculate the value in the controller, and update the CDIConfig object there. Especially since the spec.preallocation is optional, if it is not specified in the spec, the status should return false. If you reference the spec directly you could get nil pointers.

awels:
You can do something like in the main config-controller reconcile loop:
```go
// Keep a copy of the original for comparison later.
currentConfigCopy := config.DeepCopyObject()

config.Status.Preallocation = config.Spec.Preallocation != nil && config.Spec.Preallocation
```
As well as making config.Spec.Preallocation a *bool for the reasons Michael pointed out (its omitempty) That basically sets the Status to false if Spec is nil or false, and true if Spec is not nil and true

tomob:
When is #1556 intended to be merged? The change in config loop would depend on that.

awels:
I asked Michael to take the change out of #1556 that is unrelated to the docs. So if you could change it in this PR.

mhenriks:
@tomob I reverted the change in #1556 so you can make it here

tomob:
> @tomob I reverted the change in #1556 so you can make it here

Thanks!

awels:
Don't forget to run make generate after making the change in the struct.

awels (approved):

PR #1573, opened 2021-01-18
Compare target MD5 to source MD5 in fsGroup test

maya-r:
This also indirectly fixes flakiness in the test:
We use a different name for the pod finding the source MD5.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -103,9 +103,6 @@ var _ = Describe("all clone tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 			fmt.Fprintf(GinkgoWriter, "INFO: MD5SUM for source is: %s\n", sourceMD5[:32])
 
-			By("Deleting verifier pod")
-			err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
-			Expect(err).ToNot(HaveOccurred())
 			completeClone(f, f.Namespace, targetPvc, diskImagePath, sourceMD5[:32], sourcePvcDiskGroup)
 		})
 
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -98,14 +98,15 @@ var _ = Describe("all clone tests", func() {
 			fmt.Fprintf(GinkgoWriter, "INFO: %s\n", sourcePvcDiskGroup)
 			Expect(err).ToNot(HaveOccurred())
 
-			By("verifying pvc content")
-			sourceMD5, err := f.GetMD5(f.Namespace, pvc, diskImagePath, 0)
+			By("Calculating the md5sum of the source data volume")
+			sourceMD5, err := f.RunCommandAndCaptureOutput(utils.PersistentVolumeClaimFromDataVolume(dataVolume), "md5sum "+diskImagePath)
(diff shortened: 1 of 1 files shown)


awels:
Is there a typo in the description here? We check the md5 of the source PVC, and then we delete the verifier pod with utils.DeleteVerifierPod, which waits for the pod to be gone before calling complete clone. So I am unsure what this PR actually does differently from what we have.

maya-r:
> utils.DeleteVerifierPod, which waits for the pod to be gone before calling complete clone

DeletePodByName will succeed in two conditions:
1. Something else already deleted the pod and we get error IsNotFound
2. Delete was successful

So if nothing deleted the pod earlier, it doesn't wait until it's gone.

I could modify it to always wait: try to delete and don't check error, and wait until Get returns IsNotFound.

tests/cloner_test.go
```diff
@@ -98,14 +98,15 @@ var _ = Describe("all clone tests", func() {
 			fmt.Fprintf(GinkgoWriter, "INFO: %s\n", sourcePvcDiskGroup)
 			Expect(err).ToNot(HaveOccurred())
 
-			By("verifying pvc content")
-			sourceMD5, err := f.GetMD5(f.Namespace, pvc, diskImagePath, 0)
+			By("Calculating the md5sum of the source data volume")
+			sourceMD5, err := f.RunCommandAndCaptureOutput(utils.PersistentVolumeClaimFromDataVolume(dataVolume), "md5sum "+diskImagePath)
```
brybacki, line 102: I do not understand what is changed here. The sourcePvc returned by k8s was changed to sourcePvc specification create by helper function. Wich does not change much here. 

But I think the biggest change is the executor pod. Instead of using getMd5 which creates `verifierPod ` now it creates `execute-command` pod.
```
pvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(dataVolume.Namespace).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
```
to this:
utils.PersistentVolumeClaimFromDataVolume(dataVolume)

tests/cloner_test.go
```diff
@@ -98,14 +98,15 @@ var _ = Describe("all clone tests", func() {
...
 			Expect(err).ToNot(HaveOccurred())
+			fmt.Fprintf(GinkgoWriter, "INFO: MD5SUM for source is: %s\n", sourceMD5[:32])
 
 			By("Deleting verifier pod")
 			err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
```
brybacki, line 107: no need to delete `verifier` pod if execute-command Pod is created and already deleted.

maya-r:
I changed the subject/description of this PR for clarity, and made a second PR for the alternative fix - waiting for the verifier pod to be deleted even if it exists. That might be useful for other tests as well.

brybacki:
the #1584 looks good (lets wait for the test runs), so maybe this one is actually not needed?

brybacki:
I think the PR https://github.com/kubevirt/containerized-data-importer/pull/1601  helps to address the real issue.

I think we can wait for a few runs and if test is no more flaky then just close this PR #1573.

PR #1583, opened 2021-01-18
Retry finding the pods for looking up the annotations.

awels:

We sometimes have failures if we look for the pods before they are created (or maybe after they have disappeared). Retry in the case the pods don't exist yet.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -1000,6 +1000,13 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 	})
 
 	Describe("Pass specific datavolume annotations to the transfer pods", func() {
+		verifyAnnotations := func(pod *v1.Pod) {
+			By("verifying passed annotation")
+			Expect(pod.Annotations[controller.AnnPodNetwork]).To(Equal("net1"))
+			By("verifying non-passed annotation")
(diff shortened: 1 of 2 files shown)


tests/datavolume_test.go
```diff
@@ -1013,15 +1013,23 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+				By("find importer pod")
+				var sourcePod *v1.Pod
+				Eventually(func() bool {
+					sourcePod, err = utils.FindPodByPrefix(f.K8sClient, dataVolume.Namespace, common.ImporterPodName, common.CDILabelSelector)
+					return err == nil
+				}, timeout, pollingInterval).Should(BeTrue())
+				By("verifying passed annotation")
+				Expect(sourcePod.Annotations[controller.AnnPodNetwork]).To(Equal("net1"))
+				By("verifying non-passed annotation")
+				Expect(sourcePod.Annotations["annot1"]).ToNot(Equal("value1"))
```
brybacki, line 1029: on one hand the 8 lines here it could be a helper function to check for annotations on a pod (and reused in two other places), on the other hand I like how tests are verbose and expectations are in place right now .

awels: I can make it a function scoped to the describe that uses it. should still be readable.

PR #1584, opened 2021-01-19
Make DeletePodByName always wait for the pod to stop existing.

maya-r:
Previously, if Delete was successful, err was nil,
so we didn't wait for the pod to not be found.

This is an alternative way of fixing the problem pointed out in #1573.

**Release note**:

```release-note
Make functional tests more resilient.
```


brybacki:
This is interesting, so we did not wait correctly, that might explain some problems.
Maybe now we might make the grace period smaller than 30 (but in a different PR, and if this one helps).

PR #1586, opened 2021-01-19
Move smart clone tunable to CDIConfig.

maya-r:
Being the odd one out for CDI tunables is problematic if deployed
through HCO - it currently only allows editing CDI Spec.Config and
so this feature isn't visible.

[BZ #1917514](https://bugzilla.redhat.com/show_bug.cgi?id=1917514)

**Release note**:

```release-note
Move cloneStrategyOverride parameter from CDI.Spec to CDI.Config.Spec
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -655,11 +655,7 @@ func (r *DatavolumeReconciler) getCloneStrategy() (cdiv1.CDICloneStrategy, error
 		return cdiv1.CloneStrategySnapshot, fmt.Errorf("no active CDI")
 	}
 
-	if cr.Spec.Config == nil {
-		return cdiv1.CloneStrategySnapshot, nil
-	}
-
-	if cr.Spec.Config.CloneStrategyOverride == nil {
+	if cr.Spec.Config == nil || cr.Spec.Config.CloneStrategyOverride == nil {
 		return cdiv1.CloneStrategySnapshot, nil
 	}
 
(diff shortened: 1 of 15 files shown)


brybacki:
vfio, fsgroup and connection reset, 3 known problems in the tests here, are the fixes merged?

brybacki:
I do not like the release note. I mean I had to read twice to understand what "disable tunable" means. Maybe use the name of API element (in case one searches for this in changelog):
"Move cloneStrategyOverride parameter from CDI.Spec to CDI.Config.Spec"

maya-r:
I agree, your suggestion is much better, changed :)
/retest

pkg/controller/datavolume-controller.go
```diff
@@ -637,12 +637,16 @@ func (r *DatavolumeReconciler) getCloneStrategy() (cdiv1.CDICloneStrategy, error
 		return cdiv1.CloneStrategySnapshot, fmt.Errorf("no active CDI")
 	}
 
-	if cr.Spec.CloneStrategyOverride == nil {
+	if cr.Spec.Config == nil {
```
awels, line 640: Nitpick: you can do:
```go
if cr.Spec.Config == nil || cr.Spec.Config.CloneStrategyOverride == nil {
  return cdiv1.CloneStrategySnapshot, nil
}
Instead of the two ifs. No functional difference really but groups the two together
```

awels (approved):
/lgtm
let me know if you want to change that one nitpick.

pkg/controller/datavolume-controller.go
maya-r: Done

maya-r:
ping @aglitke @mhenriks @awels -- this PR only needs an approve status now

awels:
There seems to be a merge conflict here.

maya-r:
Looks like a unit-test hit a timeout? (it passes locally)
I feel like I shouldn't be re-testing it to avoid burdening the CI system more

maya-r:
This PR would only have been a reasonable choice if merged right away, and I suspect merging it at this point is damaging.

PR #1589, opened 2021-01-20
[release-v1.23] Skip test 2555 if running on openshift (#1572)

maya-r:
We have a different build which uses a different versioned qemu-img.
This test only fails on newer versions of qemu-img, and it's causing our differential build to always have a single failure.
Skipping on OpenShift to ensure that no regressions enter - we still test kubernetes lanes.
Backport of #1572

**Release note**:

```release-note
NONE
```


maya-r:
oops, merging to wrong branch

PR #1590, opened 2021-01-20
[release-v1.23] Skip test 2555 if running on openshift (#1572)

maya-r:
* Move "isOpenshift" to utils and make public.

No functional change.

Signed-off-by: Maya Rashish <mrashish@redhat.com>

* Skip malformed too large qcow2 on openshift.

Whether this test fails depends on the qemu-img version. The way we
build and test it in this repo is fine, but external builds may fail.

Skipping only on OpenShift means we will continue testing it and
finding regressions.

Signed-off-by: Maya Rashish <mrashish@redhat.com>

Backport of #1572

**Release note**:

```release-note
NONE
```


maya-r:
/retest
(creating a PR for the wrong branch then re-creating it for the right one has some weird side effects)

PR #1597, opened 2021-01-21
Support cloning from Filesystem to Block and vice-versa

arnongilboa:
* validate source and target pvcs has the same content type - for all clones
* if source and target volume modes are different, validate content type is kubevirt, and fallback to host-assisted cloning
* unit & func tests

Fixes #

**Release note**:
```release-note
Add support for cloning kubevirt disk residing on a filesystem mode PV to another PV that is block mode, and vice-versa.
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-validate.go
+++ b/pkg/apiserver/webhooks/datavolume-validate.go
@@ -70,6 +70,18 @@ func validateDataVolumeName(name string) []metav1.StatusCause {
 	return causes
 }
 
+func validateContentTypes(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSpec) (bool, cdiv1.DataVolumeContentType, cdiv1.DataVolumeContentType) {
+	sourceContentType := cdiv1.DataVolumeKubeVirt
+	if contentType, found := sourcePVC.Annotations[controller.AnnContentType]; found {
+		sourceContentType = cdiv1.DataVolumeContentType(contentType)
+	}
+	targetContentType := spec.ContentType
+	if targetContentType == "" {
+		targetContentType = cdiv1.DataVolumeKubeVirt
+	}
+	return sourceContentType == targetContentType, sourceContentType, targetContentType
+}
+
 func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *v1beta1.AdmissionRequest, field *k8sfield.Path, spec *cdiv1.DataVolumeSpec) []metav1.StatusCause {
 	var causes []metav1.StatusCause
 	var url string
@@ -199,7 +211,16 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *v1beta1.A
 					return causes
 				}
 			}
-			err = controller.ValidateCanCloneSourceAndTargetSpec(&sourcePVC.Spec, spec.PVC)
+			valid, sourceContentType, targetContentType := validateContentTypes(sourcePVC, spec)
+			if !valid {
+				causes = append(causes, metav1.StatusCause{
+					Type:    metav1.CauseTypeFieldValueInvalid,
+					Message: fmt.Sprintf("Source contentType (%s) and target contentType (%s) do not match", sourceContentType, targetContentType),
+					Field:   field.Child("PVC").String(),
+				})
+				return causes
+			}
+			err = controller.ValidateCanCloneSourceAndTargetSpec(&sourcePVC.Spec, spec.PVC, targetContentType)
 			if err != nil {
 				causes = append(causes, metav1.StatusCause{
 					Type:    metav1.CauseTypeFieldValueInvalid,
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/token"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/fetcher"
@@ -371,8 +372,11 @@ func (r *CloneReconciler) validateSourceAndTarget(sourcePvc, targetPvc *corev1.P
 	if err := validateCloneToken(r.tokenValidator, sourcePvc, targetPvc); err != nil {
 		return err
 	}
-
-	err := ValidateCanCloneSourceAndTargetSpec(&sourcePvc.Spec, &targetPvc.Spec)
+	contentType, err := ValidateCanCloneSourceAndTargetContentType(sourcePvc, targetPvc)
+	if err != nil {
+		return err
+	}
+	err = ValidateCanCloneSourceAndTargetSpec(&sourcePvc.Spec, &targetPvc.Spec, contentType)
 	if err == nil {
 		// Validation complete, put source PVC bound status in annotation
 		setBoundConditionFromPVC(targetPvc.GetAnnotations(), AnnBoundCondition, sourcePvc)
@@ -484,7 +488,19 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy, clientName str
 		return nil, err
 	}
 
-	pod := MakeCloneSourcePodSpec(image, pullPolicy, sourcePvcName, sourcePvcNamespace, ownerKey, clientKey, clientCert, serverCABundle, pvc, podResourceRequirements, workloadNodePlacement)
+	sourcePvc, err := r.getCloneRequestSourcePVC(pvc)
+	if err != nil {
+		return nil, err
+	}
+
+	var sourceVolumeMode corev1.PersistentVolumeMode
+	if sourcePvc.Spec.VolumeMode != nil {
+		sourceVolumeMode = *sourcePvc.Spec.VolumeMode
+	} else {
+		sourceVolumeMode = corev1.PersistentVolumeFilesystem
+	}
+
+	pod := MakeCloneSourcePodSpec(sourceVolumeMode, image, pullPolicy, sourcePvcName, sourcePvcNamespace, ownerKey, clientKey, clientCert, serverCABundle, pvc, podResourceRequirements, workloadNodePlacement)
 
 	if err := r.client.Create(context.TODO(), pod); err != nil {
 		return nil, errors.Wrap(err, "source pod API create errored")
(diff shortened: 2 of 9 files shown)


pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -199,7 +199,23 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *v1beta1.A
...
+			}
+			if sourceContentType != string(targetContentType) {
+				causes = append(causes, metav1.StatusCause{
+					Type:    metav1.CauseTypeFieldValueInvalid,
+					Message: fmt.Sprintf("Source contentType (%s) and target contentType (%s) do not match", sourceContentType, targetContentType),
+					Field:   field.Child("PVC").String(),
+				})
+				return causes
+			}
+			err = controller.ValidateCanCloneSourceAndTargetSpec(&sourcePVC.Spec, spec.PVC, targetContentType)
```
brybacki, line 218: woow, this validateDataVolumeSpec might be a good candidate for some easy refactoring - dividing into small meaning methods, 
and this new addition would be something like validateMatchingContentType

pkg/controller/clone-controller.go
```diff
@@ -714,27 +747,40 @@ func ParseCloneRequestAnnotation(pvc *corev1.PersistentVolumeClaim) (exists bool
...
-	targetVolumeMode := corev1.PersistentVolumeFilesystem
-	if targetSpec.VolumeMode != nil && *targetSpec.VolumeMode == corev1.PersistentVolumeBlock {
-		targetVolumeMode = corev1.PersistentVolumeBlock
-	}
-	if sourceVolumeMode != targetVolumeMode {
-		return fmt.Errorf("source volumeMode (%s) and target volumeMode (%s) do not match",
-			sourceVolumeMode, targetVolumeMode)
+	// Allow different source and target volume modes only on KubeVirt content type
+	sourceVolumeMode := GetVolumeMode(sourceSpec.VolumeMode)
+	targetVolumeMode := GetVolumeMode(targetSpec.VolumeMode)
```
brybacki, line 770: :+1:

pkg/controller/datavolume-controller.go
```diff
@@ -575,6 +575,14 @@ func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.D
 		return "", errors.New("source PVC not found")
 	}
 
+	sourceVolumeMode := GetVolumeMode(pvc.Spec.VolumeMode)
+	targetVolumeMode := GetVolumeMode(dataVolume.Spec.PVC.VolumeMode)
+	if sourceVolumeMode != targetVolumeMode {
+		r.log.V(3).Info("Source PVC and target PVC has different volume modes, falling back to host assisted clone", "source volume mode",
+			sourceVolumeMode, "target volume mode", targetVolumeMode)
+		return "", errors.New("Source PVC and target PVC has different volume modes, falling back to host assisted clone")
```
brybacki, line 583: I think It should be "Source PVC and target PVC ha[ve]" here and two lines above. Maybe some native English speaker can confirm

pkg/controller/upload-controller_test.go
```diff
@@ -217,8 +217,8 @@ var _ = Describe("Upload controller reconcile loop", func() {
 
 	It("Should return err and not clone if validation error occurs", func() {
```
brybacki, line 218: what about the test for KubeVirt content type

awels (commented):
So I think you are doing a little to much work here. the source shouldn't have to care about what content type the target is and vice versa. I understand you are trying to make sure the source and target can understand each other. But if you leave the source as is, and thus make it possible that the target receives either a tar stream or a raw image stream, we can use a technique similar to what is used in https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/importer/format-readers.go where we use some 'magic number' detection to figure out if we need an archive reader or not. The magic numbers are defined here: https://github.com/kubevirt/containerized-data-importer/blob/master/pkg/image/filefmt.go the code in upload server should look something like this:
```go
func filesystemCloneProcessor(stream io.ReadCloser, destDir string) error {
  buf := make([]byte, image.MaxExpectedHdrSize)
  _, err := io.ReadFull(stream, buf) // read current header
  if err != nil {
    return err
  }
  r = io.MultiReader(bytes.NewReader(buf), stream)
  if image.CopyKnowHrds()["tar"].Match(buf) {
    // tar stream
    if err := util.UnArchiveTar(newSnappyReadCloser(r), destDir); err != nil {
      return errors.Wrapf(err, "error unarchiving to %s", destDir)
    }
  } else {
    // raw stream
    diskImageFullPath := destDir + "/" + common.DiskImageName
    if err := util.StreamDataToFile(newSnappyReadCloser(r), diskImageFullPath); err != nil {
      return errors.Wrapf(err, "error streaming to %s", diskImageFullPath)
    }
  }
}
```
Note above is completely untested, and I might have gotten some parameters wrong. But basically we read buf size bytes, match against the magic number, and then make a multi reader so we can re-read the bytes when actually writing.

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -70,6 +70,19 @@ func validateDataVolumeName(name string) []metav1.StatusCause {
 	return causes
 }
 
+func validateContentTypes(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSpec) (bool, cdiv1.DataVolumeContentType, cdiv1.DataVolumeContentType) {
+	sourceContentType := cdiv1.DataVolumeKubeVirt
+	contentType, found := sourcePVC.Annotations[controller.AnnContentType]
+	if found {
```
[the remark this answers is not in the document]
arnongilboa: Is it really the preferable golang style? looks a bit ugly to me
https://stackoverflow.com/questions/26545883/how-to-do-one-liner-if-else-statement

pkg/controller/clone-controller.go
```diff
@@ -714,27 +747,40 @@ func ParseCloneRequestAnnotation(pvc *corev1.PersistentVolumeClaim) (exists bool
...
+	if sourceVolumeMode != targetVolumeMode && contentType != cdiv1.DataVolumeKubeVirt {
+		return fmt.Errorf("source volumeMode (%s) and target volumeMode (%s) do not match, contentType (%s)",
+			sourceVolumeMode, targetVolumeMode, contentType)
 	}
 	// Can clone.
 	return nil
 }
+
+// GetVolumeMode returns the volume mode if set, otherwise defaults to block mode
```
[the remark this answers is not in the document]
arnongilboa: sure

pkg/apiserver/webhooks/datavolume-validate.go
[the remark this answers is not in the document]
awels: That question is about an if then else type ternary operator. In this case my main goal is to keep sourceContentType variable outside of the scope of the function, since it is only used inside the if statement. But it is more a style preference than a hard requirement.

tests/cloner_test.go
```diff
@@ -250,6 +250,76 @@ var _ = Describe("all clone tests", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			targetDV := utils.NewDataVolumeCloneToBlockPV("target-dv", "1Gi", sourcePvc.Namespace, sourcePvc.Name, f.BlockSCName)
+			tagretDataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+			Expect(err).ToNot(HaveOccurred())
+			targetPvc, err := utils.WaitForPVC(f.K8sClient, tagretDataVolume.Namespace, tagretDataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Wait for target PVC Bound phase")
+			utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, f.Namespace.Name, v1.ClaimBound, targetPvc.Name)
```
maya-r, line 271: I wouldn't wait for ClaimBound as well - you might miss it

tests/cloner_test.go
```diff
@@ -250,6 +250,76 @@ var _ = Describe("all clone tests", func() {
...
+
+			targetDV := utils.NewCloningDataVolume("target-dv", "1Gi", sourcePvc)
+			targetDataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+			Expect(err).ToNot(HaveOccurred())
+			f.ForceBindPvcIfDvIsWaitForFirstConsumer(targetDataVolume)
+			targetPvc, err := utils.WaitForPVC(f.K8sClient, targetDataVolume.Namespace, targetDataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Wait for target PVC Bound phase")
+			utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, f.Namespace.Name, v1.ClaimBound, targetPvc.Name)
```
maya-r, line 307: ditto waiting for Bound and then Success

tests/cloner_test.go
maya-r: never mind -- it's a phase for the pvc and a DV

pkg/apiserver/webhooks/datavolume-validate.go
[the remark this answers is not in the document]
arnongilboa: done

tests/cloner_test.go
```diff
@@ -250,6 +251,90 @@ var _ = Describe("all clone tests", func() {
...
+			By("Wait for target DV Succeeded phase")
+			err = utils.WaitForDataVolumePhaseWithTimeout(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, "target-dv", 3*90*time.Second)
```
awels, line 317: Minor nitpick, you can use cloneCompleteTimeout instead of 3*90 here.

tests/cloner_test.go
```diff
@@ -250,6 +251,90 @@ var _ = Describe("all clone tests", func() {
...
+			targetDV := utils.NewDataVolumeCloneToBlockPV("target-dv", "1Gi", sourcePvc.Namespace, sourcePvc.Name, f.BlockSCName)
+			tagretDataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+			Expect(err).ToNot(HaveOccurred())
+			targetPvc, err := utils.WaitForPVC(f.K8sClient, tagretDataVolume.Namespace, tagretDataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Wait for target PVC Bound phase")
+			utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, f.Namespace.Name, v1.ClaimBound, targetPvc.Name)
+			By("Wait for target DV Succeeded phase")
+			err = utils.WaitForDataVolumePhaseWithTimeout(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, "target-dv", 3*90*time.Second)
```
awels, line 274: Minor nitpick, you can use cloneCompleteTimeout instead of 3*90 here.

awels (commented):
Looks really good. The smallest of nitpicks on the timeout in some tests. But good otherwise

awels (approved):
/lgtm
/approve

PR #1598, opened 2021-01-21
Try to use the CDIConfig proxy URL if it is set, if not use port-forward

awels:

We always use port-forward in our upload tests, but port-forward fails often and does not recover. This PR attempts to use the proxy URL specified in the CDIConfig instead before attempting port-forward. If the cluster being tested on has a more stable proxy URL it will be used instead.

Fixes #

**Release note**:

```release-note
NONE
```


maya-r:
Will we eventually have one configured by `make cluster-sync` so we can retire the `kubectl port-forward` code?

awels:
Well the thought is eventually make cluster-up/sync will create a stable port like for the registry and some other services. And we can use that so we don't have to deal with the instability of kubectl port-forward

brybacki:
And this can already work on systems when the URL is defined - like OCP
/lgtm

PR #1604, opened 2021-01-25
Retry uploading until success

maya-r:
The upload pod might not be ready in time, causing these tests
to become flaky.
**Release note**:

```release-note
Make upload functional tests less flaky
```

### DIFF
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -1042,8 +1042,9 @@ var _ = Describe("Preallocation", func() {
 		Expect(token).ToNot(BeEmpty())
 
 		By("Do upload")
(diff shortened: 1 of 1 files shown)


brybacki:
Actually there is a check for Readiness a few lines before. So if the test is failing it might be because something else is wrong there. Maybe the forwarder/proxy fails.
```
phase := cdiv1.UploadReady
By(fmt.Sprintf("Waiting for datavolume to match phase %s", string(phase)))			
err = utils.WaitForDataVolumePhase(f.CdiClient, f.Namespace.Name, phase, dataVolume.Name)			
```

If the pod is not ready - then I would call it a bug. It has to be ready when the DV has status UploadReady.

PR #1605, opened 2021-01-26
Fix lsblk command parsing

mhenriks:

If a block device has been populated with a disk image, `lsblk` (as we invoke it) will return two rows of data.  The first is the size of the entire device.  The second is the size of the embedded disk image partition.

This was not an issue until we added the `-mount` param to the `clone-source` program.  And it appears to only be a problem with 1.18 (and earlier?).  I assume this is because 1.18 is built with older go version.  I suspect argument parsing changed.  But I'm not sure.  Nevertheless I am adding this in master because it is more correct.

Fixes #

**Release note**:

```release-note
BugFix: fix lsblk output parsing
```

### DIFF
--- a/cmd/cdi-cloner/cloner_startup.sh
+++ b/cmd/cdi-cloner/cloner_startup.sh
@@ -14,7 +14,7 @@
 #See the License for the specific language governing permissions and
 #limitations under the License.
 
-set -euo pipefail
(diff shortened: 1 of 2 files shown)


mhenriks:
not using lsblk here lol

PR #1606, opened 2021-01-26
fix lsblk handling

mhenriks:

If a block device has been populated with a disk image, lsblk (as we invoke it) will return two rows of data. The first is the size of the entire device. The second is the size of the embedded disk image partition.

This was not an issue until we added the -mount param to the clone-source program.

Newer code uses `blockdev` instead of `lsblk` so apparently not a problem.

Fixes #

**Release note**:

```release-note
BugFix: fix lsblk output parsing
```


awels (approved):
/lgtm
/approve

awels:
Oh nevermind this is directly in 1.18 already. The cherrypick will do nothing

PR #1610, opened 2021-01-26
Use nbdkit and not qemu-img in more cases (when scratch space is used)

maya-r:
Avoids QEMU curl bugs like:
https://bugs.launchpad.net/qemu/+bug/1916501
(Which is one of our example YAML files).

With this change nbdkit is used for all the HTTP operations.

I'd like to gradually increase our usage of nbdkit so we can retire the hacky workarounds for using nbdkit.

**Release note**:

```release-note
Use nbdkit for all HTTP imports, including ones that need scratch space.
```


maya-r:
err, hopefully I tested the right subset of things.

awels:
I am trying to understand what this does. It looks like it is supposed to replace the qemu operations with nbdkit for scratch space download. However download to the scratch space always uses the go client because we can't use qemu (otherwise we wouldn't need scratch space). So I am left to wonder what the purpose of this is. Unless I am missing something.

maya-r:
I'd like to simplify our code to only have a single qemuOperations, in legible steps.

alicefr:
@maya-r in certain cases you still need the scratch space and custom logic for downloading the image. Those are:
* the source server doesn't support accept ranges
* the source server is redirected to S3 because S3 doesn't support head requests.

For those cases, you cannot use nbdkit curl plugin. But I understand the need of simplify the code

maya-r:
@alicefr I have a lot of confidence that those two scenarios are handled correctly despite this change because there's a functional test case for them :-) it uses Go code to import them then. (qemu-img had the same limitation).

The functional test failure we get seems to be that we imported some malformed QCOW2 files as functional tests [from this bug report](https://bugs.launchpad.net/ossa/+bug/1449062) and nbdkit seems to handle one of them that qemu-img could not

maya-r:
/hold
make it official with a title - there's a failing test, and it needs more investigation

alicefr:
@maya-r the reason why this is failing is that `nbdkit curl` + `qemu-img` actually don't return an error. Just tried inside the importer to reproduce that manually:
```bash
$  qemu-img info --output=json http://cdi-file-host.cdi/invalid_qcow_images/invalid-qcow-large-json.img
qemu-img: curl: The requested URL returned error: 416 Requested Range Not Satisfiable
qemu-img: Could not open 'http://cdi-file-host.cdi/invalid_qcow_images/invalid-qcow-large-json.img': Failed to read snapshot table: Input/output error
$ nbdkit -U - curl http://cdi-file-host.cdi/invalid_qcow_images/invalid-qcow-large-json.img --run 'qemu-img info $nbd'
                                    0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
[...]
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
                                      0 B 1970-01-01 00:00:00   00:00:00.000
Format specific information:
    compat: 1.1
    compression type: zlib
    lazy refcounts: false
    refcount bits: 16
    corrupt: false

```

rwmjones:
Interesting bug - I'm able to reproduce the problem locally.  Note that `qemu-img info invalid-qcow-large-json.img` works fine, indicating that the file is not actually corrupt, so nbdkit behaviour is the correct one.  I'm looking at what the difference is between qemu's curl driver and nbdkit's curl plugin.

rwmjones:
I enabled debugging in the curl driver in qemu which unfortunately requires recompiling qemu (see https://github.com/qemu/qemu/blob/51db2d7cf26d05a961ec0ee0eb773594b32cc4a1/block/curl.c#L38) and you can see that qemu's block/curl.c is making an invalid request to the web server:

```
> GET /invalid-qcow-large-json.img HTTP/1.1
Host: scratch.home.annexia.org
Range: bytes=197136-459279
Accept: */*

* Mark bundle as not supporting multiuse
* The requested URL returned error: 416 Requested Range Not Satisfiable
* Closing connection 0
```

Notice how the file is 197120 bytes in size, so requesting byte range 197136-459279 (beyond the end of the file) is wrong.

In other words this is a bug in qemu.  nbdkit is correct.

alicefr:
So, we probably have to fix the [test](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/datavolume_test.go#L431-L454) checking the invalid json

maya-r:
I don't see the unit test failure locally, not sure what the difference is.

awels:
So I am trying to understand what this is doing based on the description and what is in the PR, and I am confused. What exactly are we doing here? use nbdkit when reading from scratch space?

aglitke:
@maya-r @awels  Are we still advancing this PR?  @alicefr mentioned that she's blocked on it in order to advance #1690

awels:
@aglitke So I don't understand what this PR does, and I would like someone to explain it to me, because I don't really understand the purpose.

alicefr:
@awels this PR uses nbdkit in the case where the image can be directly streamed. If nbdkit ca be used, then we can complely remove the support of qemu-img-curl that is what this other [PR](https://github.com/kubevirt/containerized-data-importer/pull/1690) is doing.

pkg/importer/http-datasource.go
```diff
@@ -116,6 +116,18 @@ func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType
 func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	var err error
 	hs.readers, err = NewFormatReaders(hs.httpReader, hs.contentLength)
+	hs.url, _ = url.Parse(fmt.Sprintf("nbd:unix:%s", nbdkitSocket))
```
awels, line 119: Can we move
```go
if hs.contentType == cdiv1.DataVolumeArchive {
  return ProcessingPhaseTransferDataDir, nil
}
```
To line 119. I keep thinking the err gets overwritten (it doesn't, but it is hard to read). But if the readers error I don't think there is a need to start nbdkit at all.

awels (commented):
It also needs a rebase.

pkg/importer/http-datasource.go
maya-r: Done, also rebased

PR #1611, opened 2021-01-26
Size checks release v1 28

maya-r:
Cherry pick of #1466 without the last commit (adjust test output to nbdkit)

**Release note**:

```release-note

```


maya-r:
argh wrong branch

PR #1612, opened 2021-01-26
[release-v1.28] Validate image fits in filesystem in a lot more cases. take filesystem overhead into account when resizing.

maya-r:
Cherry pick of #1466 without the last commit (adjust test output to nbdkit)

**Release note**:

```release-note

```


maya-r:
/hold
need to wait a while before we accept such a backport to release-v1.28.

pkg/importer/data-processor_test.go
```diff
@@ -244,7 +244,8 @@ var _ = Describe("Data Processor", func() {
 		}
 		dp := NewDataProcessor(mdp, "dest", "dataDir", tmpDir, "1G", 0.055, false)
 		dp.availableSpace = int64(1500)
-		qemuOperations := NewFakeQEMUOperations(nil, nil, fakeInfoRet, nil, nil, resource.NewScaledQuantity(int64(1500), 0))
+		usableSpace := int64((1 - 0.055) * float64(dp.availableSpace))
```
marceloamaral: Would it be better to use the constant DefaultGlobalOverhead instead of hard code the `0.055` in all tests that are calculating the usable spaces? 

I think it is better to keep consistency. Also, I saw some discussion before about possibly increase the file system overhead parameter, for example, from 0.055 to 0.6.... So, it would be even easier to update it if needed.

pkg/importer/data-processor_test.go
```diff
@@ -311,9 +311,12 @@ var _ = Describe("Resize", func() {
 			url: url,
 		}
 		dp := NewDataProcessor(mdp, "dest", "dataDir", "scratchDataDir", "", 0.055, false)
```
marceloamaral, line 313: For consistency, it might be better to use the constant DefaultGlobalOverhead instead of hard code the `0.055`.

marceloamaral:
Btw, the failing tests seems to be related to merge conflicts...

pkg/importer/data-processor.go
```diff
@@ -393,3 +392,12 @@ func (dp *DataProcessor) calculateTargetSize() int64 {
 func (dp *DataProcessor) PreallocationApplied() string {
 	return dp.preallocationApplied
 }
+
+func (dp *DataProcessor) getUsableSpace() int64 {
+	blockSize := int64(512)
```
marceloamaral, line 397: Shouldn't the block size be a constant variable or dynamically defined?
I think that dynamically is better since It might not be always 512. 
It seems that the function blockSize, err := GetAvailableSpaceBlock(fileName) could be used here

maya-r: My motivation or the 512 hard-coded choice was seeing a lot of code inside qemu/qemu-img.c doing
DIV_ROUND_UP(..., BDRV_SECTOR_SIZE)
where BDRV_SECTOR_SIZE is 512.

awels:
This needs a rebase before we can merge.

PR #1627, opened 2021-02-01
Update DV status to indicate when VDDK configmap is missing.

mrnold:
VMware imports need to have a v2v-vmware ConfigMap defined so the importer can make use of VDDK libraries. It is easy to forget this step, and it is not obvious when the importer is waiting for it. This pull request raises the visibility of this error up to the Bound condition on the DataVolume.

Partial fix to help address BZ#1886566, requested by QE. 

**Release note**:

```release-note
Raise visibility of "missing VDDK image" error up to DV Bound condition.
```


awels (approved):
/lgtm
/approve

PR #1629, opened 2021-02-02
Pvc profiles api

brybacki:

Defines a way to store and present the information about current PVC profiles so it is accessible by CDI (when creating PVC for DV) and users.

The idea is to have a new custom resource that serves the information about recommended properties for PVC - right now the accessMode and volumeMode. This can be used by CDI controllers when creating a PVC from DV. That way the DataVolume can be simplified and if the properties are missing, defaults can be applied from the StorageProfile.

Storage profile maps 1:1 to StorageClass

The `status` contains current defaults and some information about storageclass.

The `spec` is available for end users(who have access to storageprofile) to define/override defaults.

```yaml
kind: StorageProfile
metadata: 
  name: sample-storage-profile
spec: 
  claimPropertySets: 
  - accessModes:
    - ReadWriteMany
    volumeMode: 
      Filesystem
status:
    storageClass: <string>
    provisioner: <string>
    claimPropertySets: 
    - accessModes: array of [ReadWriteMany|ReadWriteOnce]
      volumeMode: [Filesystem|Block]
```

Fixes #

The following files where changed (the rest is auto-generated):
- pkg/apis/core/v1beta1/types.go
- pkg/apis/core/v1beta1/register.go 

- pkg/operator/resources/cluster/storageprofile.go
- pkg/operator/resources/cluster/factory.go 
- tools/schema-exporter/schema-exporter.go

- pkg/operator/controller/controller_test.go 

**Release note**:

```release-note
New CRD StorageProfile available
```


pkg/apis/core/v1alpha1/types.go
```diff
@@ -250,6 +250,45 @@ const (
 // DataVolumeCloneSourceSubresource is the subresource checked for permission to clone
 const DataVolumeCloneSourceSubresource = "source"
 
+//CDIStorageClassCapabilities provides a CDI specific extension for StorageClasses
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=cdistorageclass;cdistorageclasses,scope=Cluster
+type CDIStorageClass struct {
```
awels, line 258: I would name it CDIStorageClassCapabilities. The current name implies it is some kind of special storage class, while in fact it is defining properties of actual storage classes.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -250,6 +250,45 @@ const (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   CDIStorageClassSpec   `json:"spec"`
+	Status CDIStorageClassStatus `json:"status,omitempty"`
+}
+
+//CDIStorageClassStatusSpec defines specification for StorageClasses Capabilities
+type CDIStorageClassSpec struct {
+	// PvcProfiles
+	PvcProfiles []string `json:"pvcProfiles"`
```
awels, line 269: I would call this VolumeProfile (no s) a type that contains some fields something like, (copied these from the PVC struct):
```go
type CDIStorageClassSpec struct {
  // volumeProfiles
  volumeProfiles []VolumeProfile `json:"volumeProfiles,omitempty"`
}

type VolumeProfile struct {
  // AccessModes contains the desired access modes the volume should have.
  // More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#access-modes-1
  // +optional
  AccessModes []corev1.PersistentVolumeAccessMode `json:"accessModes,omitempty" protobuf:"bytes,1,rep,name=accessModes,casttype=PersistentVolumeAccessMode"`
  // volumeMode defines what type of volume is required by the claim.
  // Value of Filesystem is implied when not included in claim spec.
  // +optional
  VolumeMode *corev1.PersistentVolumeMode `json:"volumeMode,omitempty" protobuf:"bytes,6,opt,name=volumeMode,casttype=PersistentVolumeMode"`
}
```
Note the omitempty in the "json:" section. That means the value is optional and can be missing.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -250,6 +250,45 @@ const (
...
+}
+
+//CDIStorageClassStatus provides the most recently observed status of the CDIStorageClass
+type CDIStorageClassStatus struct {
+	// The StorageClass name for which capabilities are defined
+	//StorageClass *string `json:"storageClass,omitempty"`
+	// The Storage class provisioner plugin name
+	//Provisioner *string `json:"provisioner,omitempty"`
+	// PvcProfiles computed from the spec and detected in the system
+	PvcProfiles []string `json:"pvcProfiles"`
```
awels, line 279: Same here, I like volumeProfile better than PVCProfile.

awels:
Also note travis is failing on some comment issues.

pkg/apis/core/v1alpha1/types.go
```diff
@@ -250,6 +250,57 @@ const (
...
+type StorageProfileStatus struct {
+	// The StorageClass name for which capabilities are defined
+	StorageClass *string `json:"storageClass,omitempty"`
+	// The Storage class provisioner plugin name
+	Provisioner *string `json:"provisioner,omitempty"`
+	// ClaimPropertySets computed from the spec and detected in the system
+	ClaimPropertySets []ClaimPropertySet `json:"claimPropertySets,omitempty"`
+}
+
+type ClaimPropertySet struct {
```
marceloamaral, line 283: The travis is failing because the type ClaimPropertySet should have comment
My suggestion is:
// ClaimPropertySet defines the specification of the volume

brybacki: oh yes, thanks for comment, 
it is a copy paste error, as you can see the comment is available in  in v1beta1, and I can see I messed up some other parts in v1beta1, I will fix it, and then I think I can remove [WIP]

pkg/apis/core/v1alpha1/types.go
```diff
@@ -250,6 +250,57 @@ const (
...
+	VolumeMode *corev1.PersistentVolumeMode `json:"volumeMode,omitempty" protobuf:"bytes,6,opt,name=volumeMode,casttype=PersistentVolumeMode"`
+}
+
+//StorageProfileList provides the needed parameters to do request a list of StorageProfile from the system
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type StorageProfileList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+
+	// Items provides a list of CDIStorageClass
```
marceloamaral, line 300: I think the `CDIStorageClass` is the old term that you used before, it should be `StorageProfile` instead

pkg/apis/core/v1alpha1/types.go
brybacki: after discussion settled on the name StorageProfile

pkg/apis/core/v1alpha1/types.go
brybacki: thanks for the sample,
and renamed ClaimPropertySet

pkg/apis/core/v1alpha1/types.go
brybacki: new names, alpha removed

pkg/apis/core/v1alpha1/types.go
brybacki: as I learned now, the alpha1 is no longer needed for something that is a new functionality, so I have now removed the alpha1 CRD, 

also I have corrected the names and comments

pkg/apis/core/v1alpha1/types.go
brybacki: right, fixed (in v1beta1)

awels:
Note make test-lint will run the linter that will catch all the comment issues.

pkg/apis/core/v1beta1/types.go
```diff
@@ -253,6 +253,58 @@ const (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   StorageProfileSpec   `json:"spec"`
+	Status StorageProfileStatus `json:"status,omitempty"`
+}
+
+//StorageProfileSpec defines specification for StorageProfile
+type StorageProfileSpec struct {
+	// ClaimPropertySets is a provided set of properties applicable to PVC
+	ClaimPropertySets []string `json:"claimPropertySets,omitempty"`
```
awels, line 273: I am fairly certain this should be a []ClaimPropertySet?

brybacki: wait... did I forget to commit my local changes?
sorry for that

brybacki:
@awels @mhenriks what do you thing now? Is this a usable API for new PVC Profiles?

marceloamaral:
The code looks good to me. 

Regarding the definition of the parameters, why we can only define `accessModes` and `volumeMode`? What about the `resources` and `selector`?

Regarding the definition of the functionally, can the user create more than one storageProfile? If yes, how the controller will know which one to use?

brybacki:
> The code looks good to me.
> 
> Regarding the definition of the parameters, why we can only define `accessModes` and `volumeMode`? What about the `resources` and `selector`?

Only `accessModes` and `volumeMode` now, we can always extend in the future (adding is backwards compatible)
> 
> Regarding the definition of the functionally, can the user create more than one storageProfile? If yes, how the controller will know which one to use?

Profiles will be created by CDI, the user can _edit the spec_. Right now one StorageProfile maps to one StorageClass - a 1:1 mapping. We match it by name (and reference the storageClass name in the status, should we decide to map profile to class differently).

pkg/apis/core/v1beta1/types.go
```diff
@@ -256,6 +256,58 @@ const (
...
+// ClaimPropertySet is a set of properties applicable to PVC
+type ClaimPropertySet struct {
+	// AccessModes contains the desired access modes the volume should have.
+	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#access-modes-1
+	// +optional
+	AccessModes []corev1.PersistentVolumeAccessMode `json:"accessModes,omitempty" protobuf:"bytes,1,rep,name=accessModes,casttype=PersistentVolumeAccessMode"`
+	// volumeMode defines what type of volume is required by the claim.
+	// Value of Filesystem is implied when not included in claim spec.
+	// +optional
+	VolumeMode *corev1.PersistentVolumeMode `json:"volumeMode,omitempty" protobuf:"bytes,6,opt,name=volumeMode,casttype=PersistentVolumeMode"`
```
mhenriks, line 298: Should snapshot support be included?  If not part of the spec, should it at least be supported in the status?

brybacki: What kind of snapshot support? Right now the main goal is to support recommended parameters that can be used to create a DV/PVC. So to create a DV without specifying accessModes/volumeMode. 

Current scope does not cover snapshot. And as I said before - we can always extend in the future (I think adding things to this Resource is backwards compatible). 

If you think it will be better to extend the scope of the change, then we need to have some discussion.

mhenriks: Perhaps it is just an implementation detail but I think all other things being equal we should prefer a storageclass that supports snapshots over ones that don't.

And I think it may be nice to report snapshot support in the status

awels: We can add the snapshot support at a later time. I do agree it will be nice to prefer storage with snapshot support in the future.

awels (approved):
/lgtm
Seeing if Michael has any objections to just the access mode and volume mode at this point.

PR #1630, opened 2021-02-02
Increase polling speed for upload annotation test

awels:

We are seeing failures because the pods are disappearing too fast. Attempt to fix this
by polling much faster so the odds of us being able to get both pods is increased.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -29,8 +29,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


maya-r:
We can also use the rate-limited HTTP url

awels:
We can't since this is an upload.

PR #1637, opened 2021-02-04
Preallocate even if the size is too small

tomob:
This PR removes "skipped" condition for preallocation. Importer/uploader
will preallocate to the available size.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -67,7 +67,7 @@ func main() {
 	previousCheckpoint, _ := util.ParseEnvVar(common.ImporterPreviousCheckpoint, false)
 	finalCheckpoint, _ := util.ParseEnvVar(common.ImporterFinalCheckpoint, false)
 	preallocation, err := strconv.ParseBool(os.Getenv(common.Preallocation))
-	var preallocationApplied common.PreallocationStatus
+	var preallocationApplied bool
 
 	//Registry import currently support kubevirt content type only
 	if contentType != string(cdiv1.DataVolumeKubeVirt) && (source == controller.SourceRegistry || source == controller.SourceImageio) {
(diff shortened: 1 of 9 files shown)


maya-r:
we have a lot of logic (from before) just to give the preallocation applied termination message and annotation - is that just to make it easier to write tests, or is it useful? perhaps we can check the log for a message?

maya-r:
oops I see @alicefr already finds the preallocation applied annotation possibly useful :-)

tomob:
> oops I see @alicefr already finds the preallocation applied annotation possibly useful :-)

:-)

Also, a test that only looks at the logs from an importer/uploader pod turned out to be very flaky. I wouldn't want to add more problems to the CI.

And also @qwang1 uses the annotations in her test plan.

cmd/cdi-importer/importer.go
```diff
@@ -219,3 +211,12 @@ func main() {
 	}
 	klog.V(1).Infoln(message)
 }
+
+func convertQuantityToInt64(size resource.Quantity) int64 {
+	int64Size, asInt := size.AsInt64()
+	if !asInt {
```
maya-r, line 217: I'm not sure what this case is doing. Is this for 64-bit overflow?

mhenriks:
@tomob Do we need a functional test for this?

cmd/cdi-importer/importer.go
tomob: If I understand the internals of `Quantity`, it can be created as or converted to a Decimal form. My understanding is that in order to safely get `int64` from `Qunatity`, I have to deal with it.

tomob:
> @tomob Do we need a functional test for this?

Well, I found the problem because of a failing functional test, so technically there is a test case for that. However, let me see if I can come up with a way to simulate the problem explicitly.

cmd/cdi-importer/importer.go
```diff
@@ -219,3 +211,12 @@ func main() {
 	}
 	klog.V(1).Infoln(message)
 }
+
+func convertQuantityToInt64(size resource.Quantity) int64 {
```
aglitke, line 215: If this is truly needed it smells like it should be a shared utility function.  This sort of conversion seems like a pretty common operation.  Was this code required for those other cases?  If not, why not?  If so, can we share that logic with this case too?

tomob: Indeed, there is a function for that. I don't know how could I have missed it. But I did. Fixing...

PR #1639, opened 2021-02-04
VDDK: avoid crash when specified disk isn't in VM.

### LINKED ISSUE #1634: VMDK Import crashes when no Snapshot is present in VM object

>
 /kind bug
> /kind enhancement

**What happened**:
When trying to import a VMDK from a vcenter instance the importer container crashes with a SIGSEGV in the vddk-datasource.go file.
Succesful import of the VMDK File
Import a VM from vcenter (using source: vddk in the yaml) which never had a snapshot taken. 
However simply deleting all snapshots the VM still seems to import fine.
I think this might go back to the VM Object not having a snapshot list property when no snapshot was ever taken before. 
**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.28.3
- Kubernetes version (use `kubectl version`): 1.19.4
- Cloud provider or hardware configuration: 4 Node vSphere Cluster
- Install tools:
- Others:

mrnold:
This pull request avoids a VDDK importer crash when the specified disk is not in the source VM's hardware configuration, and the VM has never had any snapshots taken. Now it will return a slightly more helpful error message.

Fixes #1634

**Release note**:
```release-note
Avoid VDDK importer crash when source VM does not contain specified disk.
```

### DIFF
--- a/pkg/importer/vddk-datasource.go
+++ b/pkg/importer/vddk-datasource.go
@@ -394,10 +394,14 @@ func (vmware *VMwareClient) FindDiskFromName(fileName string) (*types.VirtualDis
 		klog.Errorf("Unable to list snapshots: %s\n", err)
 		return nil, err
 	}
-	if disk := vmware.FindDiskInSnapshotTree(snapshot.Snapshot.RootSnapshotList, fileName); disk != nil {
-		return disk, nil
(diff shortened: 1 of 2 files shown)


awels (commented):
/lgtm
/approve

PR #1603, opened 2021-01-25
Decrease grace period to zero to make CI testing faster

tomob:
Signed-off-by: Tomasz Baranski <tbaransk@redhat.com>

**What this PR does / why we need it**:

Setting the grace period of deletion of verifier pod might make CI testing faster.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

Issue #1596, opened 2021-01-21
[flaky CI] ALL Operator tests Operator cert config tests should allow update


awels:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
/triage build-watcher
> /kind enhancement

**What happened**:
Several test runs failed with: ALL Operator tests Operator cert config tests should allow update
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1565/pull-containerized-data-importer-e2e-k8s-1.17-hpp/1351624394878750720
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1565/pull-containerized-data-importer-e2e-k8s-1.17-hpp/1350081798712856576
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

tomob:
Added some diagnostics with PR #1626.

dhiller:
Seen again here: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1631/pull-containerized-data-importer-e2e-k8s-1.18-hpp-destructive/1356712421678387201

maya-r:
flake with diagnostics: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1644/pull-containerized-data-importer-e2e-k8s-1.18-hpp-destructive/1358760418180337664

phoracek:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1644/pull-containerized-data-importer-e2e-k8s-1.18-hpp-destructive/1358760418180337664

PR #1644, opened 2021-02-08
Fix inverted logic - wait for the pod to stop existing.

maya-r:
This is part of #1603 split into its own change.
I would like to have this as its own change so we can backport it right away, whereas #1603 is something we'd like to test to see if it's stable for a while.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/pod.go
+++ b/tests/utils/pod.go
@@ -52,7 +52,7 @@ func DeletePodByName(clientSet *kubernetes.Clientset, podName, namespace string,
 	})
 	return wait.PollImmediate(2*time.Second, podDeleteTime, func() (bool, error) {
 		_, err := clientSet.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


awels (approved):
/lgtm
/approve

maya-r:
flake already reported in #1596
/test pull-containerized-data-importer-e2e-k8s-1.18-hpp-destructive

maya-r:
/test pull-containerized-data-importer-e2e-k8s-1.17-ceph
upload test that failed with:
io: read/write on closed pipe
seems kinda new.

maya-r:
/cherry-pick release-v1.28

the "write on closed pipe" is because if the pipe writer in formRequestFunc encounters an error, it fails the test, so we can't retry.
(after some experimentation...) the only solution I can offer for it is to not report errors, since the goroutine blocks for a while.
I think it's better to pursue "don't rely on kubectl port-forward" than offer this patch...

PR #1654, opened 2021-02-12
Test filesystem overhead on blank images

tomob:
This commit adds a test that verifies, that filesystem overhead is taken
into account when creating a blank image (with preallocation).

Also, some test cleanup.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/pvc.go
+++ b/tests/framework/pvc.go
@@ -240,6 +240,41 @@ func (f *Framework) VerifySparse(namespace *k8sv1.Namespace, pvc *k8sv1.Persiste
 	return info.VirtualSize >= info.ActualSize, nil
 }
 
+// VerifyFSOverhead checks whether virtual size is smaller than actual size. That means FS Overhead has been accounted for.
+// NOTE: this assertion is only valid when preallocation is used.
+func (f *Framework) VerifyFSOverhead(namespace *k8sv1.Namespace, pvc *k8sv1.PersistentVolumeClaim, preallocation bool) (bool, error) {
+	var executorPod *k8sv1.Pod
+	var err error
+
+	if !preallocation {
+		return false, fmt.Errorf("VerifyFSOverhead is only valid when preallocation is used")
+	}
+
+	executorPod, err = f.CreateVerifierPodWithPVC(namespace.Name, pvc)
+	if !apierrs.IsAlreadyExists(err) {
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}
+	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+	cmd := fmt.Sprintf("qemu-img info %s/disk.img --output=json", utils.DefaultPvcMountPath)
+	output, stderr, err := f.ExecShellInPod(executorPod.Name, namespace.Name, cmd)
+
+	if err != nil {
+		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: stderr: [%s]\n", stderr)
+		return false, err
+	}
+	fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: qemu-img info output %s\n", output)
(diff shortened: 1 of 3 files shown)


tests/framework/pvc.go
```diff
@@ -240,6 +240,34 @@ func (f *Framework) VerifySparse(namespace *k8sv1.Namespace, pvc *k8sv1.Persiste
...
+		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: stderr: [%s]\n", stderr)
+		return false, err
+	}
+	fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: qemu-img info output %s\n", output)
+	var info image.ImgInfo
+	err = json.Unmarshal([]byte(output), &info)
+	if err != nil {
+		klog.Errorf("Invalid JSON:\n%s\n", string(output))
+	}
+	return info.VirtualSize < info.ActualSize, nil
```
awels, line 268: I think the actual size can be > virtual size. But not by much,.

maya-r: I don't understand this check. I would have expected to see fs overhead taken into account here, but it's not.

awels:
The one failure in nfs seems relevant to this PR, can you take a look?

tests/framework/pvc.go
```diff
@@ -240,6 +240,35 @@ func (f *Framework) VerifySparse(namespace *k8sv1.Namespace, pvc *k8sv1.Persiste
 	return info.VirtualSize >= info.ActualSize, nil
 }
 
+// VerifyFSOverhead checks whether virtual size is smaller than actual size. That means FS Overhead has been accounted for
```
arnongilboa, line 243: Can't virtual size >= actual size even with FS Overhead?

tests/framework/pvc.go
```diff
@@ -240,6 +240,35 @@ func (f *Framework) VerifySparse(namespace *k8sv1.Namespace, pvc *k8sv1.Persiste
...
+	output, stderr, err := f.ExecShellInPod(executorPod.Name, namespace.Name, cmd)
+
+	if err != nil {
+		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: stderr: [%s]\n", stderr)
+		return false, err
+	}
+	fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: qemu-img info output %s\n", output)
+	var info image.ImgInfo
+	err = json.Unmarshal([]byte(output), &info)
+	if err != nil {
```
arnongilboa, line 265: do we return something meaningful in case of err != nil?
why not return something like false, err ?

tests/import_test.go
```diff
@@ -1104,4 +1124,45 @@ var _ = Describe("Preallocation", func() {
...
+	It("Filesystem overhead is honored with blank volume", func() {
+		defaultSCName := utils.DefaultStorageClass.GetName()
+		testedFilesystemOverhead := &cdiv1.FilesystemOverhead{}
+		globalOverhead := "0.055"
+		testedFilesystemOverhead.Global = cdiv1.Percent(globalOverhead)
+
+		config.Spec.FilesystemOverhead = testedFilesystemOverhead.DeepCopy()
+		By(fmt.Sprintf("Updating CDIConfig filesystem overhead to %v", config.Spec.FilesystemOverhead))
+		err := utils.UpdateCDIConfig(f.CrClient, func(config *cdiv1.CDIConfigSpec) {
+			config.FilesystemOverhead = testedFilesystemOverhead.DeepCopy()
```
arnongilboa, line 1137: Isn't it clearer to use config.Spec.FilesystemOverhead.DeepCopy() here?

tests/framework/pvc.go
tomob: Good point.

tests/framework/pvc.go
tomob: Actually this assertion (and this function) is only valid in context where preallocation is requested. This needs to be made clear (in comment and in code). Good catch.

tests/import_test.go
tomob: Since I used the code from other test, I refactored the test to avoid duplication.

PR #1658, opened 2021-02-14
Implement certConfigMap for s3 datasources, reported missing on #1506

### LINKED ISSUE #1506: Cannot set certConfigmap for s3 source

/kind bug

**What happened**:
[Documentation](https://github.com/kubevirt/containerized-data-importer/blob/master/doc/datavolumes.md#https3registry-source) says the certConfigMap option is available for HTTP/S3/Registry source. However, it is not implemented for S3.

The validation fails when this field is set:
```
error: error validating "dv.yaml": error validating data: ValidationError(DataVolume.spec.source.s3): unknown field "certConfigMap" in io.kubevirt.cdi.v1beta1.DataVolume.spec.source.s3; if you choose to ignore these errors, turn validation off with --validate=false
```

Create a data volume:
```
kind: DataVolume
metadata:
  name: mydv
spec:
  pvc:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 10Gi
    volumeMode: Filesystem
  source:
    s3:
      url: xxxxx
      secretRef: s3
      certConfigMap: import-certs
```

**Environment**:
- CDI version: v1.26.1

rgordill:
Currently, custom CA endpoints are only allowed to be http datasources. This PR implements the same feature for s3 endpoints.

Fixes #1506

The implementation for s3 reuses the features from http datasources. Two alternatives were evaluated, first was using CustomCACerts from aws s3 library, but it overwrites the system certs, so the second one was selected, using a custom httpclient with new certs implemented reusing http datasource code.

A new test case has also been implemented for invalid certs as in http datasource tests.

**Release note**:
```release-note
- New field in s3 datasources for custom CA Config Map like in http datasources
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3817,7 +3817,11 @@
      "url"
     ],
     "properties": {
-     "secretRef": {
+      "certConfigMap": {
+       "description": "CertConfigMap is a configmap reference, containing a Certificate Authority(CA) public key, and a base64 encoded pem certificate",
+       "type": "string"
+      },
+      "secretRef": {
       "description": "SecretRef provides the secret reference needed to access the S3 source",
       "type": "string"
      },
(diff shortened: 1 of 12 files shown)


awels (commented):
looks good, we might want to add a functional test with some kind of S3 server in the future.
/lgtm

awels (commented):
/approve

PR #1659, opened 2021-02-14
Fix default content type validation for cloning

### LINKED ISSUE #1656: Cannot clone DV: Source contentType () and target contentType (kubevirt) do not match

>
/kind bug
> /kind enhancement

**What happened**:
Error "Source contentType () and target contentType (kubevirt) do not match" for field "spec.PVC". when cloning PVC

Success

Create a DV (for example importing containerdisk):

```yaml
kind: DataVolume
metadata:
  name: dv-src
spec:
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 1Gi
  source:
    registry:
      url: docker://quay.io/kubevirt/cirros-container-disk-demo
```

Create a clone DV:
```yaml
kind: DataVolume
metadata:
  name: dv-tgt 
spec:
  pvc:
    accessModes:
      - ReadWriteMany
    resources:
      requests:
        storage: 1Gi
  source:
    pvc:
      name: dv-src
      namespace: default
```
Result is:

```
Error from server: error when creating "dv-clone.yaml": admission webhook "datavolume-validate.cdi.kubevirt.io" denied the request:  Source contentType () and target contentType (kubevirt) do not match
```

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.30.0
- Kubernetes version (use `kubectl version`):v1.18.14"
- Install tools:
- Others:

After a quick investigation I suppose it might be something with handling default values here: `pkg/apiserver/webhooks/datavolume-validate.go:validateContentTypes `

arnongilboa:

Fix content type validation for cloning, where the source content type is an empty string, which defaults to kubevirt.

Fixes #1656

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-validate.go
+++ b/pkg/apiserver/webhooks/datavolume-validate.go
@@ -71,10 +71,7 @@ func validateDataVolumeName(name string) []metav1.StatusCause {
 }
 
 func validateContentTypes(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSpec) (bool, cdiv1.DataVolumeContentType, cdiv1.DataVolumeContentType) {
-	sourceContentType := cdiv1.DataVolumeKubeVirt
-	if contentType, found := sourcePVC.Annotations[controller.AnnContentType]; found {
-		sourceContentType = cdiv1.DataVolumeContentType(contentType)
-	}
+	sourceContentType := cdiv1.DataVolumeContentType(controller.GetContentType(sourcePVC))
 	targetContentType := spec.ContentType
 	if targetContentType == "" {
 		targetContentType = cdiv1.DataVolumeKubeVirt
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -486,7 +486,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim) (*importPodEnvVar, error) {
 	podEnvVar := &importPodEnvVar{}
 	podEnvVar.source = getSource(pvc)
-	podEnvVar.contentType = getContentType(pvc)
+	podEnvVar.contentType = GetContentType(pvc)
 
 	var err error
 	if podEnvVar.source != SourceNone {
@@ -615,7 +615,7 @@ func (r *ImportReconciler) getSecretName(pvc *corev1.PersistentVolumeClaim) stri
 
 func (r *ImportReconciler) requiresScratchSpace(pvc *corev1.PersistentVolumeClaim) bool {
 	scratchRequired := false
-	contentType := getContentType(pvc)
+	contentType := GetContentType(pvc)
 	// All archive requires scratch space.
 	if contentType == "archive" {
 		scratchRequired = true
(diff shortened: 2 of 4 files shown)


brybacki:
https://github.com/kubevirt/containerized-data-importer/issues/1656 affects one of the base functionalities, we should have a tests for it,

the function func validateContentTypes(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSpec) (bool, cdiv1.DataVolumeContentType, cdiv1.DataVolumeContentType) can be covered by a unit tests with some source/target spec combinations

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -71,10 +71,7 @@ func validateDataVolumeName(name string) []metav1.StatusCause {
 }
 
 func validateContentTypes(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSpec) (bool, cdiv1.DataVolumeContentType, cdiv1.DataVolumeContentType) {
-	sourceContentType := cdiv1.DataVolumeKubeVirt
-	if contentType, found := sourcePVC.Annotations[controller.AnnContentType]; found {
-		sourceContentType = cdiv1.DataVolumeContentType(contentType)
-	}
+	sourceContentType := cdiv1.DataVolumeContentType(controller.GetContentType(sourcePVC))
```
brybacki, line 74: Do I understand it correctly thaT here we get DataVolumeKubeVirt, DataVolumeArchive or *empty*, 
and in the next lines for the targetContentType we get DataVolumeKubeVirt, DataVolumeArchive, and empty is not possible?

arnongilboa: @brybacki yep, we make sure targetContentType is not empty:
```
	targetContentType := spec.ContentType
	if targetContentType == "" {
		targetContentType = cdiv1.DataVolumeKubeVirt
	}

```

arnongilboa: > #1656 affects one of the base functionalities, we should have a tests for it,
> 
> the function func validateContentTypes(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSpec) (bool, cdiv1.DataVolumeContentType, cdiv1.DataVolumeContentType) can be covered by a unit tests with some source/target spec combinations

@brybacki good point. I was thinking all clone combinations were covered and all of them use this function, but may add the empty annotation as well.

pkg/controller/import-controller.go
```diff
@@ -705,8 +705,8 @@ func getSource(pvc *corev1.PersistentVolumeClaim) string {
 	return source
 }
 
-// returns the source string which determines the type of source. If no source or invalid source found, default to http
-func getContentType(pvc *corev1.PersistentVolumeClaim) string {
+// GetContentType returns the content type of the source image. If invalid or not set, default to kubevirt
+func GetContentType(pvc *corev1.PersistentVolumeClaim) string {
 	contentType, found := pvc.Annotations[AnnContentType]
 	if !found {
 		contentType = ""
```
awels, line 712: I think we should just return cdiv1.DataVolumeKubeVirt if the annotation does not exist, that is the default value if omitted.

arnongilboa: Sure, that's what the function returns in this case, but I agree an immediate return is better. Fixed.

PR #1660, opened 2021-02-14
Make annotation test less flaky.

### LINKED ISSUE #1640: [flaky ci] [vendor:cnv-qe@redhat.com][level:component]DataVolume tests Pass specific datavolume annotations to the transfer pods [test_id:5366]Cloner pod should have specific datavolume annotations passed but not others

/kind bug
> /kind enhancement

/triage build-watcher

**What happened**:
Test flaked on: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1624/pull-containerized-data-importer-e2e-k8s-1.17-hpp/1355888743168348160

```
/root/go/src/kubevirt.io/containerized-data-importer/tests/datavolume_test.go:1064
Timed out after 278.011s.
Expected
    <bool>: false
to be true
/root/go/src/kubevirt.io/containerized-data-importer/tests/datavolume_test.go:1114
```

**Environment**:
- Install tools:
- Others:

maya-r:
Rather than try to quickly catch the pod before it exits, re-create
the clone until we catch the pod.

Now the check that the DV did not succeed yet is unnecessary since
if we fail to find the pods we'll just try again.

Fixes #1640

**Release note**:

```release-note
NONE
```


tests/datavolume_test.go
```diff
@@ -1082,39 +1083,57 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			Expect(err).ToNot(HaveOccurred())
 			f.ForceBindIfWaitForFirstConsumer(pvc)
 
-			dataVolume := utils.NewCloningDataVolume(dataVolumeName, "1Gi", pvc)
-			Expect(dataVolume).ToNot(BeNil())
+			By("Trying to create cloned DV a couple of times until we catch the elusive pods")
```
arnongilboa, line 1086: Looks ok, but why instead of this rat race after our pods not simply pass them time-to-live-after-completion (default 0) to make such tests a bit cleaner?

maya-r: I think that having it for non-testing purposes will slow us down - any time we spent needlessly holding onto PVC means virt can't pick up the PVC.
As a testing only change, maybe

arnongilboa: Of course only for testing, as we spend a lot of test code lines on flakiness workarounds such as this PR, which has nothing to do with a real flow.

maya-r:
rebased on top of main (in case that helps ci), no changes

tests/datavolume_test.go
brybacki: I do not like it, but is not totally bad, 
Is this test failing often?

maya-r: Looking at flakefinder, it hasn't shown up in a while, but neither has anything else. I suspect something is up with flakefinder (maybe it's confused by the master->main rename)

awels: Let me ask the flakefinder author.

maya-r: we possibly need this change: https://github.com/kubevirt/project-infra/pull/1075

maya-r: it's flaking [fairly often](https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/containerized-data-importer/flakefinder-2021-03-22-672h.html)

brybacki: ok so lets do it

maya-r:
@awels any thoughts?

maya-r:
@awels it doesn't even need rebase and is ready to go :)

awels:
Lets see if this solves the D/S testing.

PR #1661, opened 2021-02-15
Have the healthz server check for the uploadserver first.

maya-r:
Uploadserver and healthz run in separate goroutine, so we need
to synchronize them. This seems to be the easiest way.

Alternative paths considered:
- Unifying them.
  This is troublesome, because the uploadserver requires client certs, and the kubernetes readiness probe doesn't provide them.
- Synchronize using some builtin: somewhat difficult because we either use a blocking function, or a separate goroutine, which means we don't know if it started.
- 
**Release note**:

```release-note
Eliminate a very short window where an upload pod is UploadReady but an upload would fail.
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"strings"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


awels:
looks like some unit test failures

PR #1667, opened 2021-02-18
Add missing fields to v1alpha1 resources

maya-r:
I wanted to use Preallocation from kubevirt, but it currently uses v1alpha1 DataVolumes.
We probably want to keep the features in sync, too, until KubeVirt stops using v1alpha1.

**Release note**:

```release-note
Add preallocation to v1alpha1 datavolume spec, too.
```


awels:
looks like you didn't update the schema in code to match. make generate-verify will show you what you missed.

PR #1674, opened 2021-02-22
Allow passing default annotation value to transfer pods

arnongilboa:
when the annotation is not set explicitly in the DV/PVC

Disable Istio sidecar injection to transfer pods by default, as the pods currently don't support it and may crash during transfer. 
Default can be overridden using `sidecar.istio.io/inject` annotation in the DV/PVC.

Fixes #1914833

**Release note**:
```release-note
Disable Istio sidecar injection to transfer pods by default.
Default can be overridden using `sidecar.istio.io/inject` annotation in the DV/PVC.
```

### DIFF
--- a/pkg/controller/clone-controller_test.go
+++ b/pkg/controller/clone-controller_test.go
@@ -223,6 +223,7 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		Expect(sourcePod.GetLabels()[CloneUniqueID]).To(Equal("default-testPvc1-source-pod"))
 		By("Verifying source pod annotations passed from pvc")
 		Expect(sourcePod.GetAnnotations()[AnnPodNetwork]).To(Equal("net1"))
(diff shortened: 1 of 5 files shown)


brybacki:
Is this test fail correlated to the change? `[vendor:cnv-qe@redhat.com][level:component]DataVolume tests Pass specific datavolume annotations to the transfer pods [test_id:5366]Cloner pod should have specific datavolume annotations passed but not others expand_more`

awels:
Can you investigate the failure on the hpp lane seems annotation related.

awels:
Actually this is because the pods disappear too fast. Feel free to re-run, we see the same failure in other places.

arnongilboa:
@awels @brybacki maybe we should merge @maya-r 's  [PR](https://github.com/kubevirt/containerized-data-importer/pull/1660)  which solved this one, until we have a cleaner approach (rate limit as suggested by Alexander)?

PR #1680, opened 2021-03-01
Nbdkit refactor to unify usage for vddk and http importers

alicefr:
This PR refactors the code to unify the usage of nbdkit. In the current code, we are handling nbdkit differently for the vddk and http importer.

**Release note**:

```release-note
None
```


alicefr:
This PR combined with #1610 simplifies the usage of nbdkit.
cc @maya-r

alicefr:
I probably need an interface for nbdkit operation and create a mock struct and function for the untit tests

awels (commented):
I have no issues with the code changes, however we lost a lot of unit tests in particular the nbdkit_test.go was removed. Any particular reason for that? We want to ensure that we are passing the right arguments around

alicefr:
@awels yes the nbdkit_test was basically testing the same functionalities of qemu-img because both command were run together (nbdkit curl ... --run "qemu-img ..."). With the refactoring, we return to run qemu-img independently also for nbdkit.

alicefr:
With this change qemu-img is reading the information from the nbd socket, the same was before but it was run as a subprocess of nbdkit.

PR #1685, opened 2021-03-02
Get rid of contentType in cloner and uploadserver

arnongilboa:

We removed contentType as a follow-up of PR #1597 to make our code a bit cleaner.
No need to pass it in the header, as uploadserver can learn it from the stream itself.
No need to pass it to the cloner as arg, as it can use volume mode from the env instead.  
Thanks for the tip @awels!

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/cdi-cloner/BUILD.bazel
+++ b/cmd/cdi-cloner/BUILD.bazel
@@ -14,6 +14,7 @@ go_library(
         "//pkg/util/prometheus:go_default_library",
         "//vendor/github.com/golang/snappy:go_default_library",
         "//vendor/github.com/prometheus/client_golang/prometheus:go_default_library",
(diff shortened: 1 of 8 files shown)


awels:
Note all the failing tests on all the lanes seemed similar, likely related to the PR.

arnongilboa:
> Note all the failing tests on all the lanes seemed similar, likely related to the PR.

Right. Thanks @awels !

arnongilboa:
Discussing it with @mhenriks  & @awels  we decided to discard this PR. It's more flexible and "http-ish" keep dealing with explicit content types on a request/response basis, so user has a way to say "here is what I am sending you" rather than introspecting everything. In addition,  the purposed change does not naturally fit the upload flow (content-type  = none/default).

PR #1686, opened 2021-03-03
Remove unit tests that are causing issues and aren't useful because the things they test are tested elsewhere.

awels:
The removed tests don't test anything useful, and seem to cause problems on a regular basis.

We have a test that depends on the manifest being in _out/manifests/generated and if they are missing the test fails. This PR makes sure the manifests are generated before running the unit tests.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
@mhenriks This is needed because one of the operator unit tests uses the manifest. Travis generated the manifests during its run, but I think we should have the make test-unit actually call manifests before running the tests.

PR #1692, opened 2021-03-05
Update release description template (Docker Hub -> Quay.io)

tomob:

We've changed links to CDI containers to Quay, but the description still mentions DOcker Hub.

**Release note**:

```release-note
NONE
```


brybacki:
I think we actually do not need functional-test lanes to pass here.
I wonder if we can do want to skip such tests on "doc" PRs?

PR #1694, opened 2021-03-09
Deduplicate resource definition code

arnongilboa:

Dedup top code duplications noted by sonarcloud.

[Overall Duplications](https://sonarcloud.io/component_measures?id=kubevirt_containerized-data-importer&metric=duplicated_lines_density&view=list) before this PR :
Density 18.9%
Duplicated Lines 6,529
Duplicated Blocks 113
Duplicated Files 27

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/operator/resources/cluster/cdiconfig.go
+++ b/pkg/operator/resources/cluster/cdiconfig.go
@@ -36,13 +36,13 @@ func createCDIConfigCRD() *extv1.CustomResourceDefinition {
 					Name:    "v1alpha1",
 					Served:  true,
 					Storage: false,
-					Schema:  createCDIConfigCRDSchema(),
+					Schema:  createAlphaV1ConfigCRDSchema(),
 				},
 				{
 					Name:    "v1beta1",
 					Served:  true,
 					Storage: true,
-					Schema:  createCDIConfigCRDSchema(),
+					Schema:  createBetaV1ConfigCRDSchema(),
 				},
 			},
 			Conversion: &extv1.CustomResourceConversion{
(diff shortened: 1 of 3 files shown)


pkg/operator/resources/cluster/cdiconfig.go
```diff
@@ -36,497 +36,263 @@ func createCDIConfigCRD() *extv1.CustomResourceDefinition {
 					Name:    "v1alpha1",
 					Served:  true,
 					Storage: false,
-					Schema: &extv1.CustomResourceValidation{
-						OpenAPIV3Schema: &extv1.JSONSchemaProps{
-							Description: "CDIConfig provides a user configuration for CDI",
+					Schema:  createCDIConfigCRDSchema(),
```
awels, line 39: Right here we assume that alphav1 and betav1 always remain the same, which is currently the case, but once we convince KubeVirt to use betav1, we won't update alphav1 anymore. So we would need to diverge the two at that point. Maybe start that by having a createAlphaV1ConfigCRDSchema and createBetaV1ConfigCRDSchema, and for now have alpha call beta, and put in a note that they might start diverging.

awels (commented):
The same comment is true for all 3 files. The function assumes that both alphav1 and betav1 are the same and this is currently the case, but hopefully soon won't be the case.

pkg/operator/resources/cluster/cdiconfig.go
arnongilboa: @awels done.

brybacki:
This is just an opinion: "Deduplicate resource definition code"

I know that the files changed could be easily seen, but having "resources" mentioned in the commit log might help if you plan to continue to deduplicate in some more places. But maybe you think having a series of the same comment is better.
like:
-  "Dedup top code duplications noted by sonarcloud"
-  "Dedup top code duplications noted by sonarcloud"

arnongilboa:
> This is just an opinion: "Deduplicate resource definition code"

@brybacki , much better title, thanks!

PR #1701, opened 2021-03-16
Separate nbdkit plugin arguments.

### LINKED ISSUE #1700: VDDK import not working.
/kind bug

**What happened**:
VDDK import failed because most of its nbdkit plugin arguments got passed in one big string, like this:

`Start nbdkit with: ['--foreground' '--readonly' '--exit-with-parent' '-U' '/var/run/nbd.sock' '--pidfile' '/var/run/nbd.pid' '--filter=retry' '/opt/testing/libvddk-test-plugin.so' 'libdir=opt/vmware-vix-disklib-distrib/lib64 server=vcenter.cdi:8989 user=user password=pass thumbprint=testprint vm=moref=vm-21' 'file=[teststore] testvm/testdisk.vmdk']`

Successful VDDK import.

Any manual VDDK test.

Fix incoming.

mrnold:
This pull request splits nbdkit plugin parameters into separate command line arguments instead of concatenating them into one string, so VDDK imports will work again.

Fixes #1700 

This also adds a quick parameter count check so that functional tests will fail if the same problem happens again, and formats the relevant log message so it's easier to see which parameters are stuck together.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -188,8 +188,20 @@ func (n *Nbdkit) StartNbdkit(source string) error {
 		argsNbdkit = append(argsNbdkit, a)
 	}
 	// append nbdkit plugin arguments
-	argsNbdkit = append(argsNbdkit, string(n.plugin), strings.Join(n.pluginArgs, " "), n.getSourceArg(source))
-	klog.V(3).Infof("Start nbdkit with: %v", argsNbdkit)
+	argsNbdkit = append(argsNbdkit, string(n.plugin))
+	argsNbdkit = append(argsNbdkit, n.pluginArgs...)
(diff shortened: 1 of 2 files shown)


maya-r:
Is it feasible to write a test so this is caught in the future?

mrnold:
> Is it feasible to write a test so this is caught in the future?

Yes, the existing VDDK functional test runs nbdkit with a fake VDDK plugin, so it is possible to catch it there. The reason it wasn't caught is that the fake VDDK plugin did not validate or use the plugin parameters at all. In this pull request I added some really simplistic validation to vddk-test-plugin.c that just counts up the number of plugin parameters, which is able to catch this bug specifically. It's certainly possible to add more elaborate parameter checking, but for this pull request I really just wanted to focus on getting VDDK imports working.

PR #1712, opened 2021-03-22
[release-v1.28] Try different modes of preallocation (#1663)

tomob:
If the falloc mode is not supported, try full first and then zero-sized
sparse blocks.

Bugzilla-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1928730

Signed-off-by: Tomasz Barański <tomob@users.noreply.github.com>

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -78,7 +79,13 @@ var (
 		},
 		[]string{"ownerUID"},
 	)
-	ownerUID string
+	ownerUID             string
+	preallocationMethods = [][]string{
+		{"-o", "preallocation=falloc"},
+		{"-o", "preallocation=full"},
+		{"-S", "0"},
+	}
+	maxPreallocationMethods = len(preallocationMethods)
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


pkg/image/qemu.go
```diff
@@ -273,3 +286,23 @@ func PreallocateBlankBlock(dest string, size resource.Quantity) error {
...
+		var argsToTry []string
+		var output []byte
+		if preallocate {
+			klog.V(1).Info("Added preallocation")
+			argsToTry = append(args, preallocationMethods[preallocationMethod]...)
+		}
+		output, err = fn(argsToTry)
+		if err != nil && strings.Contains(string(output), "Unsupported preallocation mode") {
+			preallocationMethod++
+			klog.V(1).Infof("Unsupported preallocation mode. Retrying with %s", preallocationMethods[preallocationMethod])
```
maya-r, line 303: this code feels un-idiomatic. suggesting:
````
var err error
if !preallocate {
  return nil
}

for _, preallocationMethod := range preallocationMethods {
   klog.V(1).Infof("Added preallocation (method %s)", preallocationMethod)
   argsToTry = append(args, preallocationMethod)
   output, err = fn(argsToTry)
   if err == nil {
      break
   }
   if strings.Contains(string(output), "Unsupported preallocation mode") {
      klog.V(1).Infof("Unsupported preallocation mode %s", preallocationMethod)
   }
}

return err
````

maya-r: I feel like both variations are misbehaving with preallocate being false now that I think about it

maya-r: oh, it's a backport, I suppose we can leave it as-is and discuss improvements on main

tomob: Hmm... You're right but the same code is in main. I'll create a new PR that fixes it upstream (in `main`), and we can later backport it.

What do you think?

maya-r: Sounds good!

tomob: Here: https://github.com/kubevirt/containerized-data-importer/pull/1714

PR #1727, opened 2021-03-29
Hold off AnnCloneOf until the target PVC is correctly updated.

tomob:

When the cloning is complete, the conditions to set `AnnCloneOf` annotation and status annotations (`AnnSourceRunningCondition*`) are different: the former is set when the cloning is done, while the latter depend on clone pod to finish. Sometimes `AnnCloneOf` is set, while the others are not, because even though the cloning is done, the pod is still in the running state. As a result, during the next reconciliation loop the pod is cleared and the status annotations are not set at all.

This PR adds a verification that the pod actually finishes (succeeds or fails) before `AnnCloneOf` is set.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -288,11 +288,6 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 
 	log.V(3).Info("Pod phase for PVC", "PVC phase", pvc.Annotations[AnnPodPhase])
 
-	if podSucceededFromPVC(pvc) && pvc.Annotations[AnnCloneOf] != "true" {
-		log.V(1).Info("Adding CloneOf annotation to PVC")
-		pvc.Annotations[AnnCloneOf] = "true"
-		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, cloneComplete)
-	}
 	if sourcePod != nil && sourcePod.Status.ContainerStatuses != nil {
 		// update pvc annotation tracking pod restarts only if the source pod restart count is greater
 		// see the same in upload-controller
@@ -304,10 +299,20 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 		setConditionFromPodWithPrefix(pvc.Annotations, AnnSourceRunningCondition, sourcePod)
 	}
 
+	var err error = nil
 	if !reflect.DeepEqual(currentPvcCopy, pvc) {
-		return r.updatePVC(pvc)
+		err = r.updatePVC(pvc)
 	}
-	return nil
+
+	if err == nil && podSucceededFromPVC(pvc) && pvc.Annotations[AnnCloneOf] != "true" {
+		// Now PVC has been fully updated and cloning can be completed.
+		log.V(1).Info("Adding CloneOf annotation to PVC")
+		pvc.Annotations[AnnCloneOf] = "true"
+		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, cloneComplete)
+
+		err = r.updatePVC(pvc)
+	}
+	return err
 }
 
 func (r *CloneReconciler) updatePVC(pvc *corev1.PersistentVolumeClaim) error {
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -288,6 +288,12 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 
 	log.V(3).Info("Pod phase for PVC", "PVC phase", pvc.Annotations[AnnPodPhase])
 
+	if podSucceededFromPVC(pvc) && pvc.Annotations[AnnCloneOf] != "true" && sourcePodFinished(sourcePod) {
+		log.V(1).Info("Adding CloneOf annotation to PVC")
+		pvc.Annotations[AnnCloneOf] = "true"
+		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, cloneComplete)
+	}
+
 	if sourcePod != nil && sourcePod.Status.ContainerStatuses != nil {
 		// update pvc annotation tracking pod restarts only if the source pod restart count is greater
 		// see the same in upload-controller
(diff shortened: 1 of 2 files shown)


pkg/controller/clone-controller.go
```diff
@@ -304,10 +299,20 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
...
 	}
-	return nil
+
+	if err == nil && podSucceededFromPVC(pvc) && pvc.Annotations[AnnCloneOf] != "true" {
+		// Now PVC has been fully updated and cloning can be completed.
+		log.V(1).Info("Adding CloneOf annotation to PVC")
+		pvc.Annotations[AnnCloneOf] = "true"
+		r.recorder.Event(pvc, corev1.EventTypeNormal, CloneSucceededPVC, cloneComplete)
+
+		err = r.updatePVC(pvc)
```
arnongilboa, line 313: Sure you need a second updatePVC in this case?

awels: Also I don't think this will solve the problem, we have not refreshed the PVC, so any attempt to update again will fail again for the same reason.
I think changing https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/clone-controller.go#L188-L193 to something like this should solve the problem:
```go
if r.hasFinalizer(pvc, cloneSourcePodFinalizer) && pvc.Annotations[AnnCloneOf] == "true" {
// Clone completed, remove source pod and finalizer.
  if err := r.cleanup(pvc, log); err != nil {
  }
}
```
This way we don't delete the source pod unless we know the annotation is set on the PVC. So if the update fails, it will just re-run the reconcile, and the pod hasn't been deleted.

awels: We probably want to extract the 'isPVCFullyUpdated' check into a function. So we can put all the logic associated with that into that function.

tomob: That wouldn't solve it either. The problem is that we add `AnnCloneOf` when the source pod is still in Running phase, and then we delete it without updating PVC annotations.

pkg/controller/clone-controller_test.go
```diff
@@ -332,6 +332,14 @@ var _ = Describe("Clone controller reconcile loop", func() {
 			AnnCloneRequest: "default/source", AnnPodReady: "true", AnnCloneToken: "foobaz", AnnUploadClientName: "uploadclient", AnnCloneSourcePod: "default-testPvc1-source-pod", AnnPodPhase: string(corev1.PodSucceeded)}
 		err = reconciler.client.Update(context.TODO(), testPvc)
 		Expect(err).ToNot(HaveOccurred())
+
+		sourcePod, err = reconciler.findCloneSourcePod(testPvc)
+		Expect(err).ToNot(HaveOccurred())
+		sourcePod.Status.Phase = corev1.PodSucceeded
```
arnongilboa, line 338: should't it already be corev1.PodSucceeded?

pkg/controller/clone-controller.go
```diff
@@ -310,6 +311,14 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 	return nil
 }
 
+func sourcePodFinished(sourcePod *corev1.Pod) bool {
+	if sourcePod == nil {
+		return true
+	}
+
+	return sourcePod.Status.Phase == corev1.PodSucceeded || sourcePod.Status.Phase == corev1.PodFailed
```
arnongilboa, line 319: Why should also corev1.PodFailed will result pvc.Annotations[AnnCloneOf] = "true"?

tomob: It won't. The first element of the condition, `if podSucceededFromPVC(pvc)...` will allow setting the annotation only if the cloning succeeds.

arnongilboa: So why not drop the PodFailed condition? to support the generic "finished" although there is currently only one caller of this func?

pkg/controller/clone-controller_test.go
tomob: Maybe it should. I've added this line because it wasn't.

pkg/controller/clone-controller.go
tomob: Because it would change the logic. If the cloning pod would fail, the cloning process will never end.

pkg/controller/clone-controller.go
```diff
@@ -288,11 +288,12 @@ func (r *CloneReconciler) updatePvcFromPod(sourcePod *corev1.Pod, pvc *corev1.Pe
 
 	log.V(3).Info("Pod phase for PVC", "PVC phase", pvc.Annotations[AnnPodPhase])
 
-	if podSucceededFromPVC(pvc) && pvc.Annotations[AnnCloneOf] != "true" {
+	if podSucceededFromPVC(pvc) && pvc.Annotations[AnnCloneOf] != "true" && sourcePodFinished(sourcePod) {
```
brybacki, line 291: so, previously the AnnCloneOf was added too early and because of this the next Reconcile of the clone PVC was skipped (line 185) 

now, AnnCloneOf is added at the very end,

brybacki: This would make a good info for a PR or commit message, besides that 
/lgtm

tomob: I've updated the description.

awels:
I am still not convinced this change fixes the idempotent issue. Namely that if an update of the DV fails, that the pod sticks around long enough for the reconcile loop to run again so it can eventually update the DV. With this PR we assume the clone succeeded if the source pod is gone. But as Arnon pointed out the clone could have failed and we want to report that.

We delete the source pod when we remove the finalizer from the PVC, so that is our final chance to update the annotation before the pod disappears. But in the current code we delete the pod, then update the PVC to remove the finalizer. We should probably flip that around, update the PVC (which removes the finalizer, AND sets the annotation) before deleting the pod.

tomob:
You're saying that the cleanup code needs fixing but IMO it's not wrong. 

The problem is that we're setting AnnCloneOf too early, i.e. before we had a change to update the pod's `AnnSourceRunningCondition.*` annotation. I've tried to detect that condition by looking at the state of the pod. But what if I just check what I actually want? How about I add an annotation to signal that it's been properly updated, the cleanup may be performed? This is I think clearer, as it shows the intent and smoothly deals with the "failed to update" issue as well. The extra annotation might be removed when we set AnnCloneOf, in order not to leave any extra annotations.

awels:
Okay so I missed that we are checking in shouldReconcile that the CloneOf annotation is there before going into the code that deletes the pod. That removes my concern.

PR #1732, opened 2021-04-01
[release-v1.28] Revert last minute changes for fs overhead, preallocation

maya-r:
vm-import is trying to pad with overhead, but it's adding too little.
the overzealous checks for filesystem overhead fail vm-import.

Let's revert the last minute changes, and an extra change because we'd like to return to a known-good point.

**Release note**:

```release-note
Revert validating filesystem overhead in more cases, try different modes of preallocation.
```


awels:
I think we can close this one right?

maya-r:
Yes, decided against a last minute revert.

PR #1735, opened 2021-04-04
Restore compatibility in specifying custom CAs by using Go client

maya-r:
This restores support for the following scenarios:
- Now the system certs are considered as valid when a custom CA is used.
- The custom CA will be accepted regardless of the key value used in the configmap.
    
Add a test for the second scenario.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1946100

The scenario of system-wide certs is not tested because it requires using an external server.

**Release note**:
```release-note
Restore compatibility: users are once again able to specify custom CAs for use in import in names other than tls.crt
```

### DIFF
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -138,7 +138,11 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	if hs.brokenForQemuImg {
 		return ProcessingPhaseTransferScratch, nil
 	}
-	if !hs.readers.Archived && hs.customCA == "" && hs.readers.Convert {
+	if hs.customCA != "" {
+		klog.V(1).Infof("Custom CA requested, using scratch space")
+		return ProcessingPhaseTransferScratch, nil
+	}
+	if !hs.readers.Archived && hs.readers.Convert {
 		// We can pass straight to conversion from the endpoint
 		return ProcessingPhaseConvert, nil
 	}
--- a/tests/utils/configmaps.go
+++ b/tests/utils/configmaps.go
@@ -78,3 +78,32 @@ func CopyConfigMap(client kubernetes.Interface, srcNamespace, srcName, destNames
 
 	return destName, nil
 }
+
+// CreateCertConfigMapWeirdFilename copies a configmap with a different key value
+func CreateCertConfigMapWeirdFilename(client kubernetes.Interface, destNamespace, srcNamespace string) (string, error) {
+	var certBytes string
+	srcName := FileHostCertConfigMap
+	srcCm, err := client.CoreV1().ConfigMaps(srcNamespace).Get(context.TODO(), srcName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	for _, value := range srcCm.Data {
+		certBytes = value
+	}
+	destName := srcName + "-" + strings.ToLower(util.RandAlphaNum(8))
+	dst := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: destName,
+		},
+		Data: map[string]string{
+			"weird-filename-should-still-be-accepted.crt": certBytes,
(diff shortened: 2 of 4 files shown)


maya-r:
Needs tests.

awels:
Also needs to fix the unit tests and imageio functional tests that are now broken.

pkg/importer/http-datasource.go
```diff
@@ -204,58 +209,54 @@ func (hs *HTTPDataSource) Close() error {
...
-		// Don't set timeout here, since that will be an absolute timeout, we need a relative to last progress timeout.
-	}
-
-	if certDir == "" {
-		return client, nil
-	}
+// CURL either supports a hashed-directory of certificates or a single file certificate.
+// Create a single file containing all our certificates.
+func createCAFile(certDir string) error {
+	systemCertDir := "/etc/pki/tls/certs/"
```
awels, line 215: Why are we hard coding some path that may or may not exist in various variations of linux. That is why we used x509.SystemCertPool, so we don't have to care about the paths of the system certs.

pkg/importer/http-datasource.go
```diff
@@ -204,58 +209,54 @@ func (hs *HTTPDataSource) Close() error {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "Error getting system certs")
+		klog.Errorf("Failed to create %s, got %v", common.ImporterCertFile, err)
 	}
+	defer certFile.Close()
 
-	// append the user-provided trusted CA certificates bundle when making egress connections using proxy
-	if files, err := ioutil.ReadDir(common.ImporterProxyCertDir); err == nil {
+	for dir := range certDirs {
+		files, err := ioutil.ReadDir(certDir)
```
awels, line 225: Why are looping over the different dirs, and then only reading certDir, I am assuming you want
```go
files, err := ioutils.ReadDir(dir)
```

awels (changes requested):
So why not use x509.SystemCertPool to create a cert pool, and then append the certs from the other directories to it. Then pass this cert pool to the http client somehow instead of writing everything to a file somewhere, and re-reading it into a cert pool.

pkg/importer/http-datasource.go
maya-r: I mainly want to create a file later to be used by nbdkit curl plugin, I didn't find a way to go from a certpool to a file

maya-r:
I went back because I had trouble making the more complex patch work.
We used to use the Go client before the nbdkit changes.

maya-r:
oops, a rebase accident... I wanted to test that the test fails before the fix, and accidentally pushed that version :-)

maya-r:
/test pull-cdi-unit-test
doesn't seem to fail locally. wondering if this is a time-based flake.

tests/utils/configmaps.go
```diff
@@ -81,6 +81,35 @@ func CopyConfigMap(client kubernetes.Interface, srcNamespace, srcName, destNames
...
+	var certBytes string
+	srcName := FileHostCertConfigMap
+	srcCm, err := client.CoreV1().ConfigMaps(srcNamespace).Get(context.TODO(), srcName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	for _, value := range srcCm.Data {
+		certBytes = value
+	}
```
tomob, line 95: Why the iteration? It gets the last item, so let's just to that.

maya-r: I wasn't sure about the idiomatic way to do it, there's usually only going to be one element here, I could add a break for clarity.

maya-r:
ping @awels

awels:
So basically this is reverting some of the usage of nbdkit because we can't get it to properly read the certs?

maya-r:
It's switching the custom cert case to use the Go client for downloading, so we have the same behaviour as we did before.
Nbdkit is still used later on.

Issue #1696, opened 2021-03-14
Functionality for user to provide Priority Class for importer/uploader pod


visheshtanksale:
**Is this a BUG REPORT or FEATURE REQUEST?**:
> /kind enhancement

**What happened**:
User does not have the ability to specify the priority class for the importer or uploader pod created CDI controller.

**What you expected to happen**:
User should be able to specify the priority class name for the Importer or uploader pod on either the DV object or by some other configuration.

**Anything else we need to know?**:
There are two design options 
1) Adding priority class name field to DV object
2) Adding priority class name field to CDI config object.

awels:
Can you provide any use cases for this? Normally the upload/importer pods should be low priority and if they get killed, we will restart them. Is this more a combination of PriorityClass and Quotas?

visheshtanksale:
Yes it is more of a combination of PriorityClass and Quota. 
User has quota available in a higher priority class but has nothing in the lower priority class. Because CDI does not allow priority class selection user's importer/upload pod do not get scheduled.
Also, for some of the critical VMs it would be really nice to have importer pod do its job without being killed.

awels:
Thanks for the use case, I have created an appropriate card and put it on our backlog for prioritization, of course PRs are welcome, but I think we need to do some brain storming to figure out at what level it makes most sense (I am leaning towards the DV level at this point)

visheshtanksale:
For meal as well it makes more sense to put it on the DV level.

visheshtanksale:
@awels I have create PR https://github.com/kubevirt/containerized-data-importer/pull/1740 with proposed API changes. Please let me know what do you think about it.

awels:
This was merged, closing.

PR #1740, opened 2021-04-07
Adding priority class name for importer/upload pod to data volume object

visheshtanksale:

Provides ability to specify the priority class name for the Importer or uploader pod on Data Volume. This PR just includes the API changes. Once the API changes are approved I will add server side changes as a separate PR.

Fixes # https://github.com/kubevirt/containerized-data-importer/issues/1696

Instead of creating a duplicate PodSpec we can just refer to v1 core PodSpec.

**Release note**:

```release-note
Adding ability to specify importer/uploader/cloner priority class on Data Volume Object
```


awels:
Hi, so I have a few questions and comments.
1. you can run make test-lint and it will run the same linter as what is currently failing.
2. Because KubeVirt is still using v1alpha1 CDI CRDs we need to add new fields to both. After adding to the types.go running make generate will generate the appropriate files.
3. CDI is a little complicated in that the DV by themselves don't do anything. DVs just create PVCs with some annotations. Then we have additional controllers (like import-controller, upload-controller, and clone-controller) that read those annotations and they create the actual pods that do the transfers. If I understand correctly we want to influence the PriorityClass on those pods. So we need some way of transferring the information from the DV through the PVCs to the PODs. And the usual mechanism we use is an annotation.
  a. Create a new annotation for priority class, examples of others should be in the code base
  b. Modify the datavolume controller so that if it sees the PriorityClass in the DV, it adds the new annotation from a.
  c. Modify the import/clone/upload controllers to read the annotation, and modify the pod spec with it.
4. We would love some unit and functional tests for a new feature so we can be confident it works properly, and so that we can support it.

visheshtanksale:
@awels Please find my reply in line..
> 1. you can run make test-lint and it will run the same linter as what is currently failing.
Fixed this.
> 2. Because KubeVirt is still using v1alpha1 CDI CRDs we need to add new fields to both. After adding to the types.go running make generate will generate the appropriate files.
Updated v1alpha1 CDI CRDs
> 3. CDI is a little complicated in that the DV by themselves ....
Yes, I am thinking of the similar design. Once we agree on the API updates, I will implemented the internal mechanism to put priority class value on the pod
> 4. We would love some unit and functional tests for a new feature so we can be confident it works properly, and so that we can support it.
I will add the tests when I implement the changes the changes to controller and all.

awels:
Okay so this PR is just establishing the API, but not the underlying code to make it useful. There are two way we can progress with this:
1. Merge the API PR, and have the api available but it not being useful until the controllers are updated.
2. Keep adding commits to this PR, until everything is there and then merge it. That way there won't be a period of time where we just have the API.

We have done both in the past, and both have pluses and minuses. In general we aim to not merge anything without unit and functional tests so we have confidence it works. But with just the API it could make sense in case you want to also integrate with lets say KubeVirt before the entire feature is available in CDI.

Let me know what you prefer, and we can work towards making that happen.

visheshtanksale:
I will keep adding commits to the this PR with controller changes and tests. Just adding the API has no use to me.

awels:
Okay, well if you want a review for some intermittent commits just ping me, and I will review. Also after thinking about for a bit, we don't care about 'pods' in the DV, so I think it makes sense to have PriorityClass on the main DV spec level, instead of a pod spec that is part of the dv spec.

visheshtanksale:
@awels I have added the controller level changes and tests. Can you please help with the review?

pkg/controller/import-controller.go
```diff
@@ -471,9 +471,9 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
 	if err != nil {
 		return err
 	}
-
+	anno := pvc.GetAnnotations()
```
awels, line 474: I would suggest making a GetPriorityClass(pvc) in pkg/controller/util.go that for now just reads the annotation and returns the value. Then you can call that to pass to createImporterPod. This centralizes the lookup of the PriorityClass and make it easier to modify later. You can also just call the same function from all the controllers. We are not very consistent with this, but I think it would improve maintenance effort in the future to be easier.

awels (commented):
Looks pretty good overall, one minor nitpicky comment.

awels:
Oh and you left some debugging message in the bazel push script.

pkg/controller/import-controller.go
visheshtanksale: Added the util methods

visheshtanksale:
@awels Addressed the comments

awels:
/lgtm
Want to write some documentation on what this does and why it is useful? You can do that in another PR. Also are you planning on adding anything else?

visheshtanksale:
@awels I have updated the docs. I am not planning to add anything else.

awels (approved):
/approve

awels:
Oh my, I am sorry, but this silly static analysis tool is upset that we have too many parameters here: https://github.com/kubevirt/containerized-data-importer/pull/1740/files#diff-868af4b730094e9180f78af4cc069ae6cecf7077a46c186553f331dccbd32993R784 and here: https://github.com/kubevirt/containerized-data-importer/pull/1740/files#diff-868af4b730094e9180f78af4cc069ae6cecf7077a46c186553f331dccbd32993R805 but you just added a single parameter and we were already over whatever limit it thinks is too much. The tool was recently added, and apparently is stopping us from merging this PR. Let me see what we can do about this, I might just merge it.

awels:
Okay, merging PR manually, we will work out how to make sonar cloud happy in a follow up PR.

PR #1741, opened 2021-04-07
ADD functionality to dump CDI install strategy in a configmap

visheshtanksale:

- This PR mirrors the install dump strategy feature on kubevirt. The install strategy dump can be used to deploy CDI without using operator.
- The operator will only dump install strategy if an env variable is set on it
Fixes #

**Release note**:

```release-note
Added support for dumping CDI install strategy
```

### DIFF
--- a/pkg/operator/controller/BUILD.bazel
+++ b/pkg/operator/controller/BUILD.bazel
@@ -22,6 +22,7 @@ go_library(
         "//pkg/operator:go_default_library",
         "//pkg/operator/resources/cert:go_default_library",
         "//pkg/operator/resources/cluster:go_default_library",
+        "//pkg/operator/resources/generate/install:go_default_library",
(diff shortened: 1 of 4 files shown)


awels:
make test-lint will show you what files are the problem, if it is formatting issues gofmt -w file will fix it. Also I think it might make sense to put most of this code in https://github.com/kubevirt/controller-lifecycle-operator-sdk so any operators that use it will get the benefit of being able to dump the install strategy.

visheshtanksale:
@awels it might be a good solution to add it to operator SDK. But then it won't be able to dump CRD objects like CDI and CDIConfig. So keeping a separate implementation made more sense.

awels:
Looks fine, need to fix linter, and if sonarcloud is having a problem see if you can fix it also

awels:
Facepalm on the sonarcloud cognitive load of that one function....

PR #1672, opened 2021-02-19
Support DataVolumes without specifying access mode/volume mode

brybacki:
Signed-off-by: Bartosz Rybacki <brybacki@redhat.com>

**What this PR does / why we need it**:

Support a DV without access mode/volume mode specified. CDI selects these parameters automatically based on PVC Profiles.
Since AccessMode is no longer validated in webhook - If value is not specified and the profile does not exist an error event is generated and saved to DataVolume.

 - [x] Use the params from StorageProfile
 - [x] Add Error Event
 - [x] Reconcile StorageProfiles
 
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

Still need to add error event!

**Release note**:

```release-note
Support DataVolumes without specifying access mode/volume mode
```

PR #1753, opened 2021-04-13
Use storage profiles when handling DataVolumes

brybacki:

Use the StorageProfile recommended params when creating the PVC for a DataVolume. To utilize new functionality, a new type `DV.Spec.Storage` has to be used so the existing `DV.Spec.PVC` and its handling remains unchanged (fully backwards compatible).

Fixes #

This the new design implemented after decisions made when working on #1672 . 
**Release note**:

```release-note
Support DataVolumes without specifying access mode/volume mode
```


pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -229,57 +229,80 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *v1beta1.A
...
+			}
+		} else {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: fmt.Sprintf("Storage size is missing"),
+				Field:   field.Child("Storage", "resources", "requests", "size").String(),
+			})
+			return causes
+		}
+		// TODO: check accessMode...
```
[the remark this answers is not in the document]
brybacki: I think this is already done on line 234: `if spec.PVC == nil && spec.Storage == nil{`
https://github.com/kubevirt/containerized-data-importer/pull/1753/files/30c24445ca3b7bd986d4e21177d4c02eacd68330?file-filters%5B%5D=.go#diff-2082e0e19255fc87c811d8b4347d274b06376449f517cca2e34ebbbdcfaf6ceeR234

pkg/controller/datavolume-controller.go
```diff
@@ -531,6 +533,24 @@ func (r *DatavolumeReconciler) getStorageClassBindingMode(dataVolume *cdiv1.Data
...
+		return dataVolume.Spec.Storage.StorageClassName
+	}
+	return nil
+}
+
+func getStorageVolumeMode(dataVolume *cdiv1.DataVolume) *corev1.PersistentVolumeMode {
+	if dataVolume.Spec.PVC != nil {
+		return dataVolume.Spec.PVC.VolumeMode
+	} else if dataVolume.Spec.Storage != nil {
+		return dataVolume.Spec.Storage.VolumeMode
```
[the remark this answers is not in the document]
brybacki: yes, will change this

brybacki:
@awels can you check again?

brybacki:
ok, recreated the problem locally, need to find out if the validator needs to be updated or the test

pkg/apis/core/v1alpha1/types.go
```diff
@@ -56,6 +58,46 @@ type DataVolumeSpec struct {
 	Preallocation *bool `json:"preallocation,omitempty"`
 }
 
+// StorageSpec defines the Storage type specification
+type StorageSpec struct {
```
mhenriks, line 62: If this has exactly the same members as `PersistentVolumeClaimSpec` why not:

`type StorageSpec corev1.PersistentVolumeClaimSpec`

mhenriks:
Looks pretty good to me a first glance

pkg/apis/core/v1alpha1/types.go
brybacki: The story is long, it started like you propose (or similar, because the corev1.PersistentVolumeClaimSpec in that first version was embeded), but then I needed to add some fields, and decided I might need to remove some PVC fields, or change a description.

I think this API needs some better documentation in the struct so I'll leave it as is, and improve the inline descriptions.

brybacki: In the meantime all the fields I have added where removed ;) so now it looks like PVCSpec, but I plan on improving the documentation, and I thin being a little bit independent from PVC is good. 

If think my arguments are too weak, I can change this.

brybacki:
This is ready for a review but right now all the lanes fail on this:
`ERROR: An error occurred during the fetch of repository 'fedora':`

pkg/apiserver/webhooks/datavolume-validate.go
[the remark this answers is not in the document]
brybacki: also added a check and error when both are specified

awels (commented):
only confusing thing left is there are 3 functions to get volumeMode, GetVolumeMode/getVolumeMode, and resolveVolumeMode which is called from getVolumeMode

pkg/apis/core/v1alpha1/types.go
brybacki: Its an array and has omitempty. In fact all fields on PVC are optional in schema ("// +optional. and omitempty") Some are only enforced by Webhooks/Validators - including resources.requests[storage] and accessModes.

awels:
Sonarcloud 'code smells' are all in generated code or imported code from standard K8S structs

PR #1754, opened 2021-04-13
Add insecureRegistries to CDIConfig

arnongilboa:

Keep supporting original ConfigMap for upgrade.

This change will allow HCO to expose a configuration to disable TLS for a container registries

**Release note**:
```release-note
Add insecure registries to CDIConfig, but keep supporting original ConfigMap for upgrade
```

### DIFF
--- a/doc/image-from-registry.md
+++ b/doc/image-from-registry.md
@@ -121,9 +121,8 @@ spec:
 
 To disable TLS security for a registry:
 
-Add the registry to the `cdi-insecure-registries` `ConfigMap` in the `cdi` namespace.
+Add the registry to CDIConfig insecureRegistries in the `cdi` namespace.
 
 ```bash
-kubectl patch configmap cdi-insecure-registries -n cdi \
-  --type merge -p '{"data":{"mykey": "my-private-registry-host:5000"}}'
+kubectl patch cdi cdi --patch '{"spec": {"config": {"insecureRegistries": ["my-private-registry-host:5000"]}}}' --type merge
 ```
--- a/pkg/apis/core/v1beta1/types.go
+++ b/pkg/apis/core/v1beta1/types.go
@@ -463,6 +463,8 @@ type CDIConfigSpec struct {
 	FilesystemOverhead *FilesystemOverhead `json:"filesystemOverhead,omitempty"`
 	// Preallocation controls whether storage for DataVolumes should be allocated in advance.
 	Preallocation *bool `json:"preallocation,omitempty"`
+	// InsecureRegistries is a list of TLS disabled registries
+	InsecureRegistries []string `json:"insecureRegistries,omitempty"`
 }
 
 //CDIConfigStatus provides the most recently observed status of the CDI Config resource
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -509,7 +509,7 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 		if err != nil {
 			return nil, err
 		}
-		podEnvVar.insecureTLS, err = r.isInsecureTLS(pvc)
+		podEnvVar.insecureTLS, err = r.isInsecureTLS(pvc, cdiConfig)
 		if err != nil {
 			return nil, err
 		}
@@ -558,9 +558,7 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 	return podEnvVar, nil
 }
 
-func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim) (bool, error) {
-	var configMapName string
-
+func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim, cdiConfig *cdiv1.CDIConfig) (bool, error) {
 	value, ok := pvc.Annotations[AnnEndpoint]
 	if !ok || value == "" {
 		return false, nil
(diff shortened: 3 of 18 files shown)


tests/utils/common.go
```diff
@@ -199,20 +200,49 @@ func DisableFeatureGate(c client.Client, featureGate string) (*bool, error) {
 	return &previousValue, nil
 }
 
-func removeString(featureGates []string, featureGate string) []string {
+// SetInsecureRegistry sets CDIConfig InsecureRegistries to mark the registry as allowed to be insecure
+func SetInsecureRegistry(c client.Client, registryURL string) error {
```
awels, line 204: I think AddInsecureRegistry is a better name, a set implies you are overwriting the current value which you are not.

tests/utils/common.go
```diff
@@ -199,20 +200,49 @@ func DisableFeatureGate(c client.Client, featureGate string) (*bool, error) {
...
+		if hasString(config.InsecureRegistries, parsedURL.Host) {
+			return
+		}
+		config.InsecureRegistries = append(config.InsecureRegistries, parsedURL.Host)
+	})
+	return err
+}
+
+// UnsetInsecureRegistry removed the registry from CDIConfig InsecureRegistries so it is not allowed to be insecure
+func UnsetInsecureRegistry(c client.Client, registryURL string) error {
```
awels, line 219: I think RemoveInsecureRegistry is a better name, an Unset implies you are restoring a previous value which you are not.

tests/transport_test.go
```diff
@@ -77,9 +77,9 @@ var _ = Describe("Transport Tests", func() {
 		}
 
 		if insecureRegistry {
-			err = utils.SetInsecureRegistry(f.K8sClient, f.CdiInstallNs, ep())
+			err = utils.SetInsecureRegistry(f.CrClient, ep())
```
awels, line 80: it is unlikely, but we might want to ensure the CDIConfig is updated correctly before proceeding so we know the insecure registries contain the values we want.

tests/utils/common.go
arnongilboa: renamed

tests/utils/common.go
arnongilboa: renamed

tests/transport_test.go
arnongilboa: check added

mhenriks:
Where/how is upgrade handled?  Are values in the configmap migrated to the CR?

pkg/operator/resources/cluster/cdiconfig.go
```diff
@@ -54,8 +54,23 @@ func createCDIConfigCRD() *extv1.CustomResourceDefinition {
 }
 
 func createBetaV1ConfigCRDSchema() *extv1.CustomResourceValidation {
-	// BetaV1 might diverge here from AlphaV1
-	return createAlphaV1ConfigCRDSchema()
+	// BetaV1 diverges from AlphaV1
+	config := createAlphaV1ConfigCRDSchema()
+	PatchBetaV1ConfigProps(config.OpenAPIV3Schema.Properties["spec"].Properties)
```
mhenriks, line 59: Why updating in this way and not just changing `createAlphaV1ConfigCRDSchema()`?

arnongilboa: I was assuming adding it only to Beta is smarter than adding to both. I think @awels agreed?

mhenriks: I was curious why using this patch mechanism?

arnongilboa: Afther the dedup we did here (Beta refering the Alpha func), I couldn't think of a cleaner way than a patch function to add the new Beta-only properties, but I'm open for any better suggestion :)

mhenriks: I see no good reason to exclude this from alpha.  I don't think there is any benefit.

awels: We would have to also add it to alpha types.go and generate all associated files or the check to ensure that the generate manifest matches the code manifest will fail.

arnongilboa: sure

arnongilboa: done

pkg/apis/core/v1beta1/types.go
```diff
@@ -463,6 +463,8 @@ type CDIConfigSpec struct {
 	FilesystemOverhead *FilesystemOverhead `json:"filesystemOverhead,omitempty"`
 	// Preallocation controls whether storage for DataVolumes should be allocated in advance.
 	Preallocation *bool `json:"preallocation,omitempty"`
+	// InsecureRegistries is a list of TLS disabled registries
+	InsecureRegistries []string `json:"insecureRegistries,omitempty"`
```
nunnatsa, line 467: I was under the impression that CDIConfig is deprecated. Shouldn't this new field be in the `CDISpec` type instead?

arnongilboa: CDIConfig is not deprecated but now owned by CDISpec

doc/image-from-registry.md
```diff
@@ -121,9 +121,8 @@ spec:
 
 To disable TLS security for a registry:
 
-Add the registry to the `cdi-insecure-registries` `ConfigMap` in the `cdi` namespace.
```
mhenriks, line 124: I wonder if we should put a note about the configmap being deprecated?  Not sure I feel strongly about that though

pkg/controller/import-controller.go
```diff
@@ -558,9 +558,7 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 	return podEnvVar, nil
 }
 
-func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim) (bool, error) {
-	var configMapName string
-
+func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim, cdiConfig *cdiv1.CDIConfig) (bool, error) {
```
nunnatsa, line 561: I think the `cdiConfig` parameter should be of type `*CDIConfigSpec` instead of `CDIConfig` that is not the same as the `CDI` type.

pkg/controller/import-controller.go
```diff
@@ -509,7 +509,7 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 		if err != nil {
 			return nil, err
 		}
-		podEnvVar.insecureTLS, err = r.isInsecureTLS(pvc)
+		podEnvVar.insecureTLS, err = r.isInsecureTLS(pvc, cdiConfig)
```
nunnatsa, line 512: Just to complete the comment from https://github.com/kubevirt/containerized-data-importer/pull/1754/files#r617498355, I think we should change this code to read the `CDI` CR instead of the `CDIConfig` CR.

pkg/controller/import-controller.go
mhenriks: The CDIConfig resource is a read only copy of what is in the CDI CR

pkg/controller/import-controller.go
awels: As Michael said the CDIConfig is a Readonly copy and readable by everyone, vs not everyone would have permission to read the CDI CR. That is why we have the duplication.

PR #1758, opened 2021-04-15
Verify with qemu-img that image has been preallocated

tomob:

This PR adds a step to prellocation verification tests which looks at the output of qemu-img and compares actual and virtual sizes of the image. 

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/framework/pvc.go
+++ b/tests/framework/pvc.go
@@ -161,15 +161,11 @@ func (f *Framework) VerifyTargetPVCContentMD5(namespace *k8sv1.Namespace, pvc *k
 
 // GetMD5 returns the MD5 of a file on a PVC
 func (f *Framework) GetMD5(namespace *k8sv1.Namespace, pvc *k8sv1.PersistentVolumeClaim, fileName string, numBytes int64) (string, error) {
-	var executorPod *k8sv1.Pod
-	var err error
-
-	executorPod, err = f.CreateVerifierPodWithPVC(namespace.Name, pvc)
-	if !apierrs.IsAlreadyExists(err) {
-		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	executorPod, err := f.startVerifierPod(namespace, pvc)
+	if err != nil {
+		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: could not start verifier pod: [%s]\n", err)
+		return "", err
 	}
-	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
 	cmd := "md5sum " + fileName
 	if numBytes > 0 {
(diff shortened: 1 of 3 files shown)


tests/import_test.go
```diff
@@ -1044,6 +1053,13 @@ var _ = Describe("Preallocation", func() {
 			md5, err := f.GetMD5(f.Namespace, pvc, path, md5PrefixSize)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(md5).To(Equal(expectedMD5))
+
+			if !f.IsBlockVolumeStorageClassAvailable() {
```
arnongilboa, line 1057: Sure you need this if only here?

tests/framework/pvc.go
```diff
@@ -161,15 +161,8 @@ func (f *Framework) VerifyTargetPVCContentMD5(namespace *k8sv1.Namespace, pvc *k
...
-	var executorPod *k8sv1.Pod
 	var err error
-
-	executorPod, err = f.CreateVerifierPodWithPVC(namespace.Name, pvc)
-	if !apierrs.IsAlreadyExists(err) {
-		gomega.Expect(err).ToNot(gomega.HaveOccurred())
-	}
-	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	executorPod := f.startVerifierPod(namespace, pvc)
```
arnongilboa, line 165: nice!

tests/framework/pvc.go
```diff
@@ -213,79 +199,43 @@ func (f *Framework) VerifyBlankDisk(namespace *k8sv1.Namespace, pvc *k8sv1.Persi
...
-		gomega.Expect(err).ToNot(gomega.HaveOccurred())
-	}
-	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
-
-	cmd := fmt.Sprintf("qemu-img info %s/disk.img --output=json", utils.DefaultPvcMountPath)
-
-	output, stderr, err := f.ExecShellInPod(executorPod.Name, namespace.Name, cmd)
-
+	info, err := f.GetImageInfo(namespace, pvc)
```
arnongilboa, line 202: very nice!

tests/framework/pvc.go
```diff
@@ -213,79 +199,43 @@ func (f *Framework) VerifyBlankDisk(namespace *k8sv1.Namespace, pvc *k8sv1.Persi
...
-	var info image.ImgInfo
-	err = json.Unmarshal([]byte(output), &info)
-	if err != nil {
-		klog.Errorf("Invalid JSON:\n%s\n", string(output))
-		return false, nil
-	}
-	requestedSize := pvc.Spec.Resources.Requests[k8sv1.ResourceStorage]
-	return info.VirtualSize <= info.ActualSize && info.VirtualSize < requestedSize.Value(), nil
+
+	return float64(info.VirtualSize)/float64(info.ActualSize) > 0.99, nil
```
arnongilboa, line 232: Please explain this check and use a const for the 0.99

tests/framework/pvc.go
```diff
@@ -492,3 +429,41 @@ func addVolumeMounts(pvc *k8sv1.PersistentVolumeClaim, volumeName string) []k8sv
...
+	}
+	fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: qemu-img info output %s\n", output)
+
+	err = json.Unmarshal([]byte(output), &info)
+	if err != nil {
+		klog.Errorf("Invalid JSON:\n%s\n", string(output))
+		return info, err
+	}
+
+	return info, nil
```
arnongilboa, line 454: In the first two err returns info is useless, so I guess passing *image.ImgInfo arg instead would be cleaner

tests/framework/pvc.go
```diff
@@ -492,3 +429,41 @@ func addVolumeMounts(pvc *k8sv1.PersistentVolumeClaim, volumeName string) []k8sv
...
+func (f *Framework) startVerifierPod(namespace *k8sv1.Namespace, pvc *k8sv1.PersistentVolumeClaim) *k8sv1.Pod {
+	var executorPod *k8sv1.Pod
+	var err error
+
+	executorPod, err = f.CreateVerifierPodWithPVC(namespace.Name, pvc)
+	if !apierrs.IsAlreadyExists(err) {
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}
+	err = utils.WaitTimeoutForPodReady(f.K8sClient, executorPod.Name, namespace.Name, utils.PodWaitForTime)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
arnongilboa, line 466: You may return (*k8sv1.Pod, error) here and  Expect() in the callers

tests/import_test.go
tomob: This is the only test where Block volumes are used.

tests/framework/pvc.go
tomob: Hmm... I think it's wrong. Let me rethink it.

tests/framework/pvc.go
tomob: Good idea

tests/framework/pvc.go
tomob: :+1:

tests/framework/pvc.go
tomob: OK, the correct check should simply be:
```golang
info.ActualSize >= info.VirtualSize
```
File should take on disk (ActualSize) at least as much, as it reports (VirtualSize). It's `>=` rather than `==` because of file system overhead, block device allocation etc.

tests/framework/pvc.go
tomob: And this makes SonarCloud unhappy about duplicated code :man_facepalming:

tomob:
Regarding SonarCloud's complaints about duplicated code:

After a small refactoring following Golang's principle of returning error in addition to a proper value and then handling the error at the calling points. Trying to remove the lines marked as duplicates, by either removing error logging or logging in another place, but IMO it would made the code worse, not better.

arnongilboa:
> Regarding SonarCloud's complaints about duplicated code:
> 
> After a small refactoring following Golang's principle of returning error in addition to a proper value and then handling the error at the calling points. Trying to remove the lines marked as duplicates, by either removing error logging or logging in another place, but IMO it would made the code worse, not better.

Many times I don't agree with SonarCloud's complaints. It's strange there's no option to mark them as "lgtm, won't fix".
Anyway, my comment was not a must in that case.

awels (approved):
/approve

PR #1760, opened 2021-04-19
Unlock ImageIO disks after importer failure.

mrnold:
This pull request unlocks ImageIO source disks when the importer returns an error. The importer was not finalizing the transfer through the ImageIO API because the finalize code lives in the Close method, which previously relied on a "defer" that was skipped by os.Exit. In the linked bug, the transfer was successful, but the resize failed, triggering an os.Exit that skipped the deferred Close and therefore the finalize.

This pull request adds an explicit call to Close just before the relevant calls to os.Exit (there are only a few). In the interest of further safety, this pull request also cancels a transfer if the finalize fails.

Fixes [BZ#1945522](https://bugzilla.redhat.com/show_bug.cgi?id=1945522)

**Release note**:
```release-note
Unlock ImageIO disks when transfer returns an error.
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -68,6 +68,7 @@ func main() {
 	finalCheckpoint, _ := util.ParseEnvVar(common.ImporterFinalCheckpoint, false)
 	preallocation, err := strconv.ParseBool(os.Getenv(common.Preallocation))
 	var preallocationApplied bool
+	var dp importer.DataSourceInterface
 
 	//Registry import currently support kubevirt content type only
 	if contentType != string(cdiv1.DataVolumeKubeVirt) && (source == controller.SourceRegistry || source == controller.SourceImageio) {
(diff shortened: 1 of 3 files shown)


cmd/cdi-importer/importer.go
```diff
@@ -190,12 +190,14 @@ func main() {
 		if err != nil {
 			klog.Errorf("%+v", err)
 			if err == importer.ErrRequiresScratchSpace {
+				dp.Close()
```
maya-r, line 193: Why is this needed if we already have `defer dp.Close()` up above?

maya-r: Apparently runtime.Goexit exists to be os.Exit while running deferred calls.

mrnold: Thanks, that would be much cleaner. It looks like it's meant to be called from a goroutine though, I will run some tests and see how well it works here.

mrnold: Goexit does not appear to be a drop-in replacement, it hangs after the transfer if I just use it in place of os.Exit. It does unlock the disk though, so it must be running the deferred Close correctly.

maya-r: thanks for trying

PR #1768, opened 2021-04-29
Fix cluster proxy part of proxy test

awels:
We were setting the cluster proxy noProxy to * to have it ignore
all URLs, but then set it to "" in CDI CR without cleaning up.
This PR leaves it blank.

Fixes PROXY test on Open Shift with a cluster wide proxy.
Also improve some of the error logging to understand why clients cannot be created.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/import_proxy_test.go
+++ b/tests/import_proxy_test.go
@@ -76,10 +76,7 @@ var _ = Describe("Import Proxy tests", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Eventually(func() bool {
 			_, err := f.K8sClient.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Get(context.TODO(), dvName, metav1.GetOptions{})
-			if k8serrors.IsNotFound(err) {
-				return true
-			}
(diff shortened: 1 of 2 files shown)


brybacki:
what exactly was a problem here?

awels:
So the problem is that in Open Shift, we update the cluster proxy to the test values AND also update noProxy: "*" basically disabling proxying for the cluster. This then reconciles in CDIConfig status proxy to match the cluster proxy. So the proxying is not happening. To 'fix' this we also updated CDI CR but this time noProxy: "" and thus enabling proxying. However in the cleanup we never reverted that change, and for the next test run the CDI CR proxy values would be used instead of the cluster proxy values we wanted. This also defeated the purpose of testing if we properly reconcile the cluster proxy.

This PR removes the CDI CR override and just reads the cluster proxy. The reason given for "*" was that it would stop proxying the cluster API calls, however the proxy is never used for that anyway.

brybacki:
ok, great
/lgtm

PR #1773, opened 2021-05-03
Extended importer metrics.

### LINKED ISSUE #1662: Extend metrics for importer pod
/kind enhancement

**What happened**: when calling the metrics endpoint of the importer, we only have a percentage of transferred data.

**What you expected to happen**: it would be better to have raw values, as well as pre-calculated values, such as transfer speed. For example, when migrating with VDDK the importer pod skips zeroed extents to accelerate the transfer. Using the percentage to compute the transfer speed may lead to incorrect values, hence wrong expectations for the user.

**Anything else we need to know?**: since the import mechanisms may not all provide the same data, it would be interesting to have a common set of data, e.g. percentage, and a set of data specific to the import mechanism.

**Environment**:
- CDI version: v2.6.0
- Kubernetes version: v1.19.0

mrnold:
This is an initial pull request for reviewing the implementation of extended transfer metrics, using the existing /metrics endpoint in the importer pod. If this is an acceptable way to do it, then I can use the same pattern to add more metrics.

Fixes #1662 

This has the same kind of problem that #1752 has: the importer pod can get deleted before the DataVolume reconciler picks up the last metrics update, so the data may not be complete. I'm not sure how important this will be in practice, but if there's a better way to do this I would like to hear about it. Maybe the importer can post something to the cdi-deployment pod somehow? Any ideas are welcome.

**Release note**:
```release-note
NONE
```


mrnold:
The feedback so far is that this is an okay way to add metrics, but they should be filled out for all data volume types, not just VDDK. I am looking for ways to make this generic for all nbdkit-based source types.

mrnold:
I am going to close this for now, and open a new one when I get back to working on it.

PR #1779, opened 2021-05-11
Import registry disk into PVC barely big enough

tomob:
Test for importing a registry disk into a PVC that is just barely big enough to hold the disk image.

In previous implementations, the container image was downloaded to scratch space and then extracted to scratch space. This led to two copies of the data needing to be stored in the scratch space so large images (over 50% of the PVC size) would fail to import when the space on the scratch PVC was exhausted.

This shouldn't be a problem in the current implementation because we extract the disk image inline during the transfer. We should still test this case to ensure that large images will work. It is important to test with an image that is not very compressable in order to test the case. The virtual size of the disk image should be close to the size of the PVC being used.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -693,6 +693,30 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 					Message: "Import Complete",
 					Reason:  "Completed",
 				}}),
+			table.Entry("[rfe_id:4334][test_id:6433]succeed creating import dv with given valid registry url and DV barely big enough", dataVolumeTestArguments{
+				name:             "dv-import-registry",
+				size:             "22Mi", // The image has 18M
+				url:              tinyCoreIsoRegistryURL,
+				dvFunc:           createRegistryImportDataVolume,
+				eventReason:      controller.ImportSucceeded,
+				phase:            cdiv1.Succeeded,
+				checkPermissions: true,
+				readyCondition: &cdiv1.DataVolumeCondition{
+					Type:   cdiv1.DataVolumeReady,
+					Status: v1.ConditionTrue,
+				},
+				boundCondition: &cdiv1.DataVolumeCondition{
+					Type:    cdiv1.DataVolumeBound,
+					Status:  v1.ConditionTrue,
+					Message: "PVC dv-import-registry Bound",
+					Reason:  "Bound",
+				},
+				runningCondition: &cdiv1.DataVolumeCondition{
+					Type:    cdiv1.DataVolumeRunning,
+					Status:  v1.ConditionFalse,
+					Message: "Import Complete",
+					Reason:  "Completed",
+				}}),
 			table.Entry("[test_id:5077]succeed creating import dv from VDDK source", dataVolumeTestArguments{
 				name:             "dv-import-vddk",
 				size:             "1Gi",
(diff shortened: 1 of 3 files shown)


maya-r:
/lgtm
Not sure why there's a WIP label, it looks good :)

tomob:
> Not sure why there's a WIP label, it looks good :)

I was not sure the selected size would work with all provisioners. It seems it does. :-)

tests/import_test.go
```diff
@@ -135,6 +135,29 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 			Expect(f.GetDiskGroup(f.Namespace, pvc, false)).To(Equal("107"))
 		}
 	})
+
+	It("Import registry disk into PVC barely big enough", func() {
+		imageURL := fmt.Sprintf(utils.LargePhysicalDiskQcow, f.CdiInstallNs)
```
brybacki, line 140: What is the size of the Image? What does it mean "barely big enough"?

brybacki:
Just two questions about the sizes, and you still have WIP in the commit name.

tests/import_test.go
awels: The image is found in tests/images/cirros-large-physical-size.qcow2, and its virtual size is 2Gi. What barely big enough means, I am not sure either. The requested PVC has 12% overhead so should be fine.

tests/import_test.go
```diff
@@ -135,6 +135,29 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+
+		By(imageURL)
+		dv := utils.NewDataVolumeWithHTTPImport("large-import", "2.25Gi", imageURL)
```
awels, line 143: The description of this PR mentions registry import, while here we are doing an http import.

tests/import_test.go
tomob: The truth is, I don't remember the intention behind the test. Maybe we should reconsider?

tests/datavolume_test.go
```diff
@@ -693,6 +693,31 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
 				}}),
+			table.Entry("[rfe_id:TBD][test_id:TBD]succeed creating import dv with given valid registry url and DV barely big enough", dataVolumeTestArguments{
+				name:             "dv-import-registry",
+				size:             "22Mi", // The image has 18M
+				url:              tinyCoreIsoRegistryURL,
+				dvFunc:           createRegistryImportDataVolume,
+				eventReason:      controller.ImportSucceeded,
+				phase:            cdiv1.Succeeded,
+				checkPermissions: true,
+				repeat:           10,
```
awels, line 704: Any particular reason we are repeating this 10 times? The original reason the 10 times was added for a registry import is that for hostpath WFFC storage we had a bug where there was a race that would only show every once in a while and running the test a bunch of times made it show up enough to debug it. But in this case I don't see a reason to repeat it 10 times.

tomob: No particular reason, I meant "do the same, only with smaller PVC". A can remove the line.

tomob: Done

brybacki:
nice, now the commit message is better
/lgtm

kgoldbla (approved):
The polarion id and rfe id lgtm

awels:
/lgtm
/approve
Just wanted to note the only lane this can possibly fail (if there is an issue) is the ceph lane, as that would be the only provider that would make a scratch space small enough.

PR #1785, opened 2021-05-12
Use a method of deleting the pod that waits for it to no longer exist

maya-r:
Smartclone can begin its clone prior to the pod-using-PVC being gone.
We need this wait to guarantee that there are no pending writes in
the kernel cache.

Intended to fix a flaky test, seen failing here:
https://prow.ci.kubevirt.io/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1752/pull-containerized-data-importer-e2e-k8s-1.17-ceph/1390637610644279296

**Release note**:

```release-note
NONE
```


awels (approved):
/lgtm
/approve

PR #1791, opened 2021-05-18
Fix: Compute fs overhead only for fs volumeMode

brybacki:

Correctly compute fs overhead for an effective VolumeMode. Effective, means one that is computed based on value in storage spec and the storageProfile.

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=1961227

**Release note**:

```release-note
NONE
```


awels:
Looks good, but it is missing a functional import test. I see we have tests that verify this for upload and clone already, so if you could add one for import?

brybacki:
opened a PR with manual cherrypick :
https://github.com/kubevirt/containerized-data-importer/pull/1799

PR #1793, opened 2021-05-18
[WIP] move CDI validation webhook to operator

arnongilboa:


pkg/operator/resources/operator/operator.go
```diff
@@ -1079,3 +1092,113 @@ _The CDI Operator does not support updates yet._
...
+							cdicorev1alpha1.SchemeGroupVersion.Version,
+						},
+						Resources: []string{"cdis"},
+						Scope:     &allScopes,
+					},
+				}},
+				ClientConfig: admissionregistrationv1beta1.WebhookClientConfig{
+					Service: &admissionregistrationv1beta1.ServiceReference{
+						Namespace: namespace,
+						Name:      apiServerServiceName,
```
mhenriks, line 1137: This should not be the same name as the apiserver service.  It should be the name of the service created in this file (which should not be the same as the apiserver service)

pkg/operator/resources/operator/operator.go
```diff
@@ -1079,3 +1092,113 @@ _The CDI Operator does not support updates yet._
...
+						APIGroups: []string{cdicorev1.SchemeGroupVersion.Group},
+						APIVersions: []string{
+							cdicorev1.SchemeGroupVersion.Version,
+							cdicorev1alpha1.SchemeGroupVersion.Version,
+						},
+						Resources: []string{"cdis"},
+						Scope:     &allScopes,
+					},
+				}},
+				ClientConfig: admissionregistrationv1beta1.WebhookClientConfig{
```
mhenriks, line 1134: Turns out that you can't have a webhook without tls.  I'll create a patch that bootstraps that for you

pkg/operator/resources/operator/operator.go
arnongilboa: Sure, but when createAPIService was added registered here with a unique name, for some reason we got 
`{"level":"info","ts":1621352271.742419,"logger":"cdi-operator","msg":"Orphan object exists","Request.Namespace":"","Request.Name":"cdi","obj":{"apiVersion":"apiregistration.k8s.io/v1beta1","kind":"APIService","name":"v1beta1.upload.cdi.kubevirt.io"}}
`

pkg/operator/resources/operator/operator.go
arnongilboa: Even with admissionregistrationv1beta1.Ignore you must have tls?
Patch will be nice:)

PR #1808, opened 2021-05-26
OS image poll and update API

arnongilboa:

The API is based on [this](https://github.com/kubevirt/community/pull/99) design proposal by @davidvossel.

The design provides an approach for automating golden disk image importing and keeping golden disk images up-to-date. 

A `DataSource` object references an import/clone source for a PVC. Functionally, it is similar to the `source` field of a `DataVolume`. The `DataSource`’s spec is mutable, which has the side effect of allowing the `DataSource` to act as a link that will redirect to whatever source the link currently points to. In the future, the `DataSource` will act as a PVC Populator as well as a `DataVolume` source.

A `DataImportCron` object results in its corresponding controller polling for updates to a container image on a recurring interval. When updates to a container image are detected, the `DataImportCron` will automatically import the contents into a new PVC (using the standard CDI import flows) residing in a specific namespace with a unique name. The `DataImportCron` may need to implement a PVC garbage collection policy for removing outdated PVCs after an expiration period has been exceeded.

`DataImportCron` can manage a corresponding `DataSource` to always point to the latest most up-to-date PVC that the cron has created. The `DataImportCron` should also be capable of adopting and taking ownership of an existing `DataSource` in order to satisfy the condition where an admin wishs to opt-in to using auto delivery/updates later on.

The `DataVolume` object will add a new `sourceRef` field capable of importing pvc content using a `DataSource`.

`DataSource` Example:
```yaml
kind: DataSource
metadata:
  name: fedora
spec:
  source:
    pvc:
      namespace: golden-images 
      name: fedora
status:
  conditions:
  - status: "True"
    type: Ready
```
`DataVolume` example with `sourceRef` to `DataSource`:
```yaml
kind: DataVolume
metadata:
   name: my-fedora-vm-image
spec:
  pvc:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 2Gi
  sourceRef:
    kind: DataSource
    name: fedora
    namespace: golden-images 
```
`DataImportCron` Example:
```yaml
kind: DataImportCron
metadata:
  name: fedora-image-import-cron
  namespace: golden-images
spec:
  source:
    registry:
      url: docker://quay.io/kubevirt/fedora
  schedule: "0 */12 * * *"
  garbageCollectOutdated: true
  managedDataSource: fedora
status:
  lastImportedPVC: fedora-v33
  lastExecutionTimestamp: "01:01:01"
  lastImport: "01:01:01"
  conditions:
  - status: "False"
    message: "Unable to reach docker://quay.io/kubevirt/fedora"
    type: ImportSuccessful
```
Fixes #

The following files were changed (the rest are auto-generated):
- pkg/apis/core/v1beta1/types.go
- pkg/apis/core/v1beta1/register.go 
- pkg/operator/resources/cluster/datasource.go
- pkg/operator/resources/cluster/factory.go 
- tools/schema-exporter/schema-exporter.go
- pkg/operator/controller/controller_test.go 

**Release note**:
```release-note
OS image poll and update API added
```


pkg/apis/core/v1beta1/types.go
```diff
@@ -205,6 +207,24 @@ type DataVolumeSourceVDDK struct {
 	SecretRef string `json:"secretRef,omitempty"`
 }
 
+// DataVolumeSourceRef defines an indirect reference to the source of data for the DataVolume
+type DataVolumeSourceRef struct {
+	// The kind of the source reference, currently only DataSource is supported
+	Kind DataVolumeSourceRefKind `json:"kind"`
```
mhenriks, line 213: Not sure we need/want type safety here.  `string` is what similar upstream types use.

pkg/apis/core/v1beta1/types.go
```diff
@@ -205,6 +207,24 @@ type DataVolumeSourceVDDK struct {
...
+	// The namespace of the source reference
+	Namespace string `json:"namespace"`
```
mhenriks, line 215: Maybe make the a pointer and have it optional?  Default is same namespace as the DataVolume

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +382,123 @@ type StorageProfileList struct {
...
+
+// DataSourceSpec defines specification for DataSource
+type DataSourceSpec struct {
+	// Source is the source of the data referenced by the DataSource
+	Source DataSourceSource `json:"source"`
+}
+
+// DataSourceSource represents the source for our DataSource
+type DataSourceSource struct {
+	PVC *DataSourceSourcePVC `json:"pvc"`
```
mhenriks, line 406: omitempty as this will eventually be optional

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +382,123 @@ type StorageProfileList struct {
...
+}
+
+// DataImportCronSpec defines specification for DataImportCron
+type DataImportCronSpec struct {
+	// Source specifies where to poll disk images from
+	Source DataImportCronSource `json:"source"`
+	// Schedule specifies in cron format when and how often to look for new imports
+	Schedule string `json:"schedule"`
+	// GarbageCollectOutdated specifies whether old PVCs should be cleaned up after a new PVC is imported
+	GarbageCollectOutdated bool `json:"garbageCollectOutdated"`
```
mhenriks, line 457: Do we need this to be mandatory?  Maybe make it a pointer, omitempty, and if not set assume false?

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +382,123 @@ type StorageProfileList struct {
...
+	ManagedDataSource string `json:"managedDataSource"`
+}
+
+// DataImportCronSource defines where to poll and import disk images from
+type DataImportCronSource struct {
+	Registry *DataImportCronSourceRegistry `json:"registry"`
+}
+
+// DataImportCronSourceRegistry provides the parameters to refer a registry source
+type DataImportCronSourceRegistry = DataVolumeSourceRegistry
```
mhenriks, line 468: strange use of type alias.  why?

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +382,123 @@ type StorageProfileList struct {
...
+
+// DataImportCronStatus provides the most recently observed status of the DataImportCron
+type DataImportCronStatus struct {
+	// LastImportedPVC is the name of the last imported PVC
+	LastImportedPVC string `json:"lastImportedPVC,omitempty"`
+	// LastExecutionTimestamp is the time of the last polling
+	LastExecutionTimestamp string `json:"lastExecutionTimestamp,omitempty"`
+	// LastImport is the time of the last import
+	LastImport string                    `json:"lastImport,omitempty"`
```
mhenriks, line 477: three fields above should be pointers

mhenriks (commented):
good start, just did a quick first pass

pkg/apis/core/v1beta1/types.go
arnongilboa: we are supporting only specific kinds (currently one...) so having just a `string` seems a bit odd.

pkg/apis/core/v1beta1/types.go
arnongilboa: sure

pkg/apis/core/v1beta1/types.go
arnongilboa: How can it be optional when it is currently the only possible source which is a must?
When we'll have other optional sources here all of them will sure be `optional`.

pkg/apis/core/v1beta1/types.go
arnongilboa: Right

pkg/apis/core/v1beta1/types.go
arnongilboa: Just because it should be the same. Do you mean we should use `DataVolumeSourceRegistry` here, or fork the struct?

pkg/apis/core/v1beta1/types.go
arnongilboa: Sure

pkg/apis/core/v1beta1/types.go
kwiesmueller: I think this is fine, but it doesn't prevent anyone from setting something else than DataSource.
So string would have the same effect, but having a type is just a bit more Go maybe.

mhenriks: I don't see what type safety gets us here.  This value is 99% of the time going to be specified as a string in yaml.  Whether that gets deserialized as a string or some other type in Go doesn't really matter.

Take the dataSource field of PVCs for example.  kind is a string and can only have specific values

pkg/apis/core/v1beta1/types.go
mhenriks: IMO all pointer fields should be optional.  Eventually the fields of this struct will have to be validated in a webhook to enforce a "one of" constraint.  Whether you do that now or later is your choice. If you don't do that now, there should at least be a comment here.

pkg/apis/core/v1beta1/types.go
mhenriks: Not sure you understand what's going on here.

`type B = A` is not the same as `type B A`

The former is an alias.  Used rarely and usually when refactoring to refer to types in different packages.

The later declares a new type.

I think both are unnecessary here.  Just use the type directly

kwiesmueller (changes requested):

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +383,124 @@ type StorageProfileList struct {
...
+	Source DataSourceSource `json:"source"`
+}
+
+// DataSourceSource represents the source for our DataSource
+type DataSourceSource struct {
+	PVC *DataSourceSourcePVC `json:"pvc"`
+}
+
+// DataSourceSourcePVC provides the parameters to refer an existing PVC
+type DataSourceSourcePVC = DataVolumeSourcePVC
```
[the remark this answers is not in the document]
kwiesmueller: Ah, you never stop learning. Didn't know we can do that in Go: https://golang.org/ref/spec#Type_declarations
Reading that I don't think this additional alias is needed and we can use `DataVolumeSourcePVC` above instead of this kind of unergonomic name.

pkg/apis/core/v1beta1/types.go
kwiesmueller: The Kubernetes API recommendation is to avoid booleans.
I'd suggest making this a string like `GarbageCollect` and having the options `Never` and `Outdated`.
That leaves the option to create different functionality in the future without breaking the API or needing a new field.

pkg/apis/core/v1beta1/types.go
arnongilboa: Ok, this will be a `string`.

pkg/apis/core/v1beta1/types.go
arnongilboa: Will fix

pkg/apis/core/v1beta1/types.go
arnongilboa: Sure, my mistake.

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +383,124 @@ type StorageProfileList struct {
...
+type DataImportCronSourceRegistry = DataVolumeSourceRegistry
+
+// DataImportCronStatus provides the most recently observed status of the DataImportCron
+type DataImportCronStatus struct {
+	// LastImportedPVC is the name of the last imported PVC
+	LastImportedPVC *string `json:"lastImportedPVC,omitempty"`
+	// LastExecutionTimestamp is the time of the last polling
+	LastExecutionTimestamp *string `json:"lastExecutionTimestamp,omitempty"`
+	// LastImport is the time of the last import
+	LastImport *string                   `json:"lastImport,omitempty"`
```
davidvossel, line 479: If this is a timestamp, maybe we should call it `LastImportTimestamp` instead of `LastImport`

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +383,124 @@ type StorageProfileList struct {
...
+	Conditions []DataImportCronCondition `json:"conditions,omitempty" optional:"true"`
+}
+
+// DataImportCronCondition represents the state of a data import cron condition
+// FIXME: needs LastTransitionTime/LastHeartbeatTime/Reason?
+type DataImportCronCondition struct {
+	Type    DataImportCronConditionType `json:"type" description:"type of condition ie. Progressing, UpToDate"`
+	Status  corev1.ConditionStatus      `json:"status" description:"status of the condition, one of True, False, Unknown"`
+	Message string                      `json:"message,omitempty" description:"human-readable message indicating details about last transition"`
+}
```
davidvossel, line 489: fyi, there was some further discussion about DataImportCron conditions that can be found [here](https://github.com/kubevirt/community/pull/99#discussion_r639171296)

I think the big takeaway is that we need to be able to express when the DataImportCron is currently executing (Progressing = true|false) and whether it is up-to-date or not (UpToDate = true|false). It's okay if the condition naming changes in the final implementation, we just need a way to represent those two states somehow and have messages that indicate why the condition is set to true|false.

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +383,124 @@ type StorageProfileList struct {
...
+type DataImportCronSpec struct {
+	// Source specifies where to poll disk images from
+	Source DataImportCronSource `json:"source"`
+	// Schedule specifies in cron format when and how often to look for new imports
+	Schedule string `json:"schedule"`
+	// GarbageCollectOutdated specifies whether old PVCs should be cleaned up after a new PVC is imported. Defaults to false.
+	// +optional
+	GarbageCollectOutdated *bool `json:"garbageCollectOutdated,omitempty"`
+	// ManagedDataSource specifies the name of the corresponding DataSource this cron will manage
+	ManagedDataSource string `json:"managedDataSource"`
```
davidvossel, line 461: I added a new header to the original design document labeled `DataSource OwnerReference Behavior`

I just wanted to bring some attention to how this `ManagedDataSource` field needs to work. When the DataSource doesn't exist, the DataImportCron needs to create one matching the name in this field and set an ownerReference on the corresponding object with `controller: true` set.

However in the even that the DataSource does already exist, the DataImportCron should still set an ownerReference, but with `controller: false` set.

This ensures that garbage collection works as expected regardless of what created the DataSource.

davidvossel (commented):
looks great! I made a few minor comments.

pkg/apis/core/v1beta1/types.go
awels: I am thinking Michael is saying why not just use DataVolumeSourceRegistry in the DataImportCronSource type?

pkg/apis/core/v1beta1/types.go
mhenriks: why not `metav1.Time` instead of string?

pkg/apis/core/v1beta1/types.go
arnongilboa: fixed

pkg/apis/core/v1beta1/types.go
arnongilboa: Sounds better, fixed.

pkg/apis/core/v1beta1/types.go
arnongilboa: Fixed

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
arnongilboa: Fixed

pkg/apis/core/v1beta1/types.go
arnongilboa: Note all is mentioned in the field description.

pkg/apis/core/v1beta1/types.go
arnongilboa: Thanks for the important clarification. Afaik it has no effect on this API PR.

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +380,127 @@ type StorageProfileList struct {
...
+	PVC *DataVolumeSourcePVC `json:"pvc,omitempty"`
+}
+
+// DataSourceStatus provides the most recently observed status of the DataSource
+type DataSourceStatus struct {
+	Conditions []DataSourceCondition `json:"conditions,omitempty" optional:"true"`
+}
+
+// DataSourceCondition represents the state of a data source condition
+// FIXME: needs LastTransitionTime/LastHeartbeatTime/Reason/Message?
```
kwiesmueller, line 414: If we want to add those fields later, please keep in mind that they have to be optional then to not break existing clients.

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +380,127 @@ type StorageProfileList struct {
...
+// DataImportCronSpec defines specification for DataImportCron
+type DataImportCronSpec struct {
+	// Source specifies where to poll disk images from
+	Source DataImportCronSource `json:"source"`
+	// Schedule specifies in cron format when and how often to look for new imports
+	Schedule string `json:"schedule"`
+	// GarbageCollect specifies whether old PVCs should be cleaned up after a new PVC is imported
+	// Options are currently "Never" and "Outdated", defaults to "Never"
+	// +optional
+	GarbageCollect string `json:"garbageCollect,omitempty"`
```
kwiesmueller, line 455: if optional it can be `*string` or do we always want it in the spec with the default?

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +380,127 @@ type StorageProfileList struct {
...
+
+// DataImportCronSource defines where to poll and import disk images from
+type DataImportCronSource struct {
+	Registry *DataVolumeSourceRegistry `json:"registry"`
+}
+
+// DataImportCronStatus provides the most recently observed status of the DataImportCron
+type DataImportCronStatus struct {
+	// LastImportedPVC is the name of the last imported PVC
+	LastImportedPVC *string `json:"lastImportedPVC,omitempty"`
```
kwiesmueller, line 475: Sorry if I asked this before, but doesn't this have to be a PVC reference (name,namespace) as it is the same in the DataSource?

pkg/apis/core/v1beta1/types.go
arnongilboa: `*string` seems better. Fixing.

pkg/apis/core/v1beta1/types.go
arnongilboa: Sure, fixing to `*DataVolumeSourcePVC`.

pkg/apis/core/v1beta1/types.go
arnongilboa: Right. Anyone thinks we should add some of fields now?

mhenriks (commented):
whoops forgot to submit this earlier

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +380,127 @@ type StorageProfileList struct {
...
+// DataImportCronSpec defines specification for DataImportCron
+type DataImportCronSpec struct {
+	// Source specifies where to poll disk images from
+	Source DataImportCronSource `json:"source"`
+	// Schedule specifies in cron format when and how often to look for new imports
+	Schedule string `json:"schedule"`
+	// GarbageCollect specifies whether old PVCs should be cleaned up after a new PVC is imported
+	// Options are currently "Never" and "Outdated", defaults to "Never"
+	// +optional
+	GarbageCollect *string `json:"garbageCollect,omitempty"`
```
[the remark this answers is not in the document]
kwiesmueller: I don't think it makes a difference.
Having a dedicated type might make it clearer with the enum style field.
But I'd be fine with either.

pkg/apis/core/v1beta1/types.go
```diff
@@ -362,6 +380,127 @@ type StorageProfileList struct {
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+type DataImportCron struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   DataImportCronSpec   `json:"spec"`
+	Status DataImportCronStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
kwiesmueller: From what I think to remember about other resources, status usually is {} in the yaml if empty.
So it might not be omitempty usually.

pkg/apis/core/v1beta1/types.go
```diff
@@ -44,6 +44,8 @@ type DataVolume struct {
 type DataVolumeSpec struct {
 	//Source is the src of the data for the requested DataVolume
 	Source DataVolumeSource `json:"source"`
```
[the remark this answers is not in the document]
arnongilboa: right

pkg/apis/core/v1beta1/types.go
```diff
@@ -44,6 +44,8 @@ type DataVolume struct {
...
+	//SourceRef is an indirect reference to the source of data for the requested DataVolume
+	SourceRef DataVolumeSourceRef `json:"sourceRef,omitempty"`
```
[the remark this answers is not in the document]
arnongilboa: sure

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
arnongilboa: Status of DataVolume and StorageProfile is `omitempty` and no pointer. What's different here?

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
arnongilboa: @mhenriks a few comments above you've said:
`Not sure we need/want type safety here. string is what similar upstream types use.`
What's different here?

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
kwiesmueller: Yeah I think this is fine. Seen this a few places now and seems normal.

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
mhenriks: Similar but different.  These are consts where all possible values are known.  The other case is supposed to support the name of any arbitrary type.

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
mhenriks: `omitempty` really makes no sense with non-pointer struct members.  Either way it will get serialized as `{}`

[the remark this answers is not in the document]
mhenriks: I understand it is used in a lot of places where it doesn't make sense.  Just being pedantic.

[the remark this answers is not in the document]
arnongilboa: Ok, so just removing the `omitempty`.

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
arnongilboa: Ok, understood. Adding a new type.

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
mhenriks: ah, forgot that that tag is used for schema generation and without "omitempty" it will be required, so I guess add it back.  I did not consider the schema impact.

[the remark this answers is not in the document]
arnongilboa: re-added `omitempty` :)

pkg/apis/core/v1beta1/types.go
```diff
@@ -351,6 +371,130 @@ type StorageProfileList struct {
...
+type DataImportCronSpec struct {
+	// Source specifies where to poll disk images from
+	Source DataImportCronSource `json:"source"`
+	// Schedule specifies in cron format when and how often to look for new imports
+	Schedule string `json:"schedule"`
+	// GarbageCollect specifies whether old PVCs should be cleaned up after a new PVC is imported
+	// Options are currently "Never" and "Outdated", defaults to "Never"
+	// +optional
+	GarbageCollect *DataImportCronGarbageCollect `json:"garbageCollect,omitempty"`
+	// ManagedDataSource specifies the name of the corresponding DataSource this cron will manage
```
kwiesmueller, line 447: Nothing we need to decide right now and it's kind of implicit, but we could mention here that the DataSource has to be in the same namespace.

pkg/apis/core/v1beta1/types.go
```diff
@@ -351,6 +371,130 @@ type StorageProfileList struct {
...
+	LastImportedPVC *DataVolumeSourcePVC `json:"lastImportedPVC,omitempty"`
+	// LastExecutionTimestamp is the time of the last polling
+	LastExecutionTimestamp *metav1.Time `json:"lastExecutionTimestamp,omitempty"`
+	// LastImportTimestamp is the time of the last import
+	LastImportTimestamp *metav1.Time              `json:"lastImportTimestamp,omitempty"`
+	Conditions          []DataImportCronCondition `json:"conditions,omitempty" optional:"true"`
+}
+
+// DataImportCronCondition represents the state of a data import cron condition
+// FIXME: needs LastTransitionTime/LastHeartbeatTime/Reason?
```
kwiesmueller, line 478: What was the reason of not adding them right now?

pkg/apis/core/v1beta1/types.go
kwiesmueller: If we don't want them right now, then I'd also remove the comment.
So either add the fields or remove the comment, I'm fine with it either way for now.

tests/api_validation_test.go
```diff
@@ -283,7 +284,8 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
 			table.Entry("[test_id:1923]fail with missing storage size", "manifests/dvMissingRequestSpec.yaml", true, "PVC size is missing", "spec.pvc.resources in body must be of type object"),
 			table.Entry("[test_id:1915]fail with invalid access modes", "manifests/dvInvalidAccessModes.yaml", true, "supported values: \"ReadOnlyMany\", \"ReadWriteMany\", \"ReadWriteOnce\""),
 			table.Entry("[test_id:3922]fail with multiple access modes", "manifests/dvMultipleAccessModes.yaml", true, "PVC multiple accessModes"),
-			table.Entry("[test_id:1861]fail with missing source (but source key)", "manifests/dvMissingSource2.yaml", true, "Missing Data volume source", "missing required field \"source\" in io.kubevirt.cdi.v1beta1.DataVolume.spec", "invalid: spec.source: Required value"),
+			//FIXME: Data volume source is now optional, so test should be added for missing both source and sourceRef
```
kwiesmueller, line 287: Can we fix those two FIXMEs as well? I'd prefer not checking them in.

kwiesmueller (commented):
Added a few more concerns regarding open todos in here.
If everybody else is happy with leaving them unresolved, I can commit to that as well.

tests/api_validation_test.go
mhenriks: I think I'm okay with FIXMEs in tests assuming they will be addressed in the implementation PR

pkg/apis/core/v1beta1/types.go
mhenriks: Yeah, may as well add them now

arnongilboa: Added

pkg/apis/core/v1beta1/types.go
arnongilboa: Adding

PR #1811, opened 2021-05-31
Remove the logic

ZP-AlwaysWin:
Signed-off-by: ZP-AlwaysWin

This logic has been activated twice

```
// Start processing.
		go func() {
			defer close(app.doneChan)
			if err := processor.ProcessDataResume(); err != nil {
				klog.Errorf("Error during resumed processing: %v", err)
				app.errChan <- err
			}
			app.mutex.Lock()
			defer app.mutex.Unlock()
			app.processing = false
			app.done = true
			app.preallocationApplied = processor.PreallocationApplied()
			klog.Infof("Wrote data to %s", app.destination)
		}()
```

### DIFF
--- a/pkg/uploadserver/uploadserver.go
+++ b/pkg/uploadserver/uploadserver.go
@@ -415,7 +415,7 @@ func newAsyncUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string,
 
 	uds := importer.NewAsyncUploadDataSource(newContentReader(stream, sourceContentType))
 	processor := importer.NewDataProcessor(uds, dest, common.ImporterVolumePath, common.ScratchDataDir, imageSize, filesystemOverhead, preallocation)
-	return processor, processor.ProcessDataWithPause()
+	return processor, nil
 }
 
 func newUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string, filesystemOverhead float64, preallocation bool, sourceContentType string) (bool, error) {


awels:
Can you explain a little bit what this PR does? There are 2 flows available a sync and async version of upload. What is the problem with the async version that you want to disable it?

ZP-AlwaysWin:
> Can you explain a little bit what this PR does? There are 2 flows available a sync and async version of upload. What is the problem with the async version that you want to disable it?

Instead of disabling it, this function is called twice. In fact, it only needs to be called once
```
 processor.ProcessDataResume()
```

awels:
I still don't get it, where is it called twice? Your modification disables the state in the state machine.

ZP-AlwaysWin:
> I still don't get it, where is it called twice? Your modification disables the state in the state machine.

https://github.com/kubevirt/containerized-data-importer/pull/1811#issuecomment-853798803

ZP-AlwaysWin:
> I still don't get it, where is it called twice? Your modification disables the state in the state machine.

//   ++++++++++++++++ First call  uploadProcessorFuncAsync function `processor.ProcessDataWithPause()`++++++++++++++
processor, err := uploadProcessorFuncAsync(readCloser, app.destination, app.imageSize, app.filesystemOverhead, app.preallocation, cdiContentType)

		app.mutex.Lock()

		if err != nil {
			klog.Errorf("Saving stream failed: %s", err)
			if _, ok := err.(importer.ValidationSizeError); ok {
				w.WriteHeader(http.StatusBadRequest)
			} else {
				w.WriteHeader(http.StatusInternalServerError)
			}
			w.Write([]byte(fmt.Sprintf("Saving stream failed: %s", err.Error())))
			app.uploading = false
			app.mutex.Unlock()
			return
		}
		defer app.mutex.Unlock()

		app.uploading = false
		app.processing = true

		// Start processing.
		go func() {
			defer close(app.doneChan)
//   ++++++++++++++++ Second call  ProcessDataResume() function `processor.ProcessDataWithPause()`++++++++++++++
			if err := processor.ProcessDataResume(); err != nil {
				klog.Errorf("Error during resumed processing: %v", err)
				app.errChan <- err
			}
			app.mutex.Lock()
			defer app.mutex.Unlock()
			app.processing = false
			app.done = true
			app.preallocationApplied = processor.PreallocationApplied()
			klog.Infof("Wrote data to %s", app.destination)
		}()

awels:
Okay so it is actually correct that it is called twice. The second call is inside a go routine, so the following steps happen in the async version of upload:
1. The user starts the upload, and the ProcessDataWithPause is called. This manipulates the state machine, until the user is done uploading the bytes to the pod.
2. At this point the pod will return success to the user, so the user client can close the http connection used for uploading.
3. This now triggers ProcessDataResume (note a different function from ProcessDataWithPause) which does the post upload processing (like conversion and resize)

The reason we have this in 2 steps is because the post processing step can take a while, especially on large images, and if we keep the http connection open during this time, it can timeout and cause an error in the user client, even though nothing is actually wrong, it is just taking a while to process.

If you want to keep the http connection open and not use the two step process, you can use the sync version of upload, but you run the risk of timing out on the http connection. In particular the upload proxy will time out due to no activity.

PR #1815, opened 2021-06-02
update deps and bazel

mhenriks:

k8s deps to 20.2 and bazel to 3.7.2

Fixes #

**Release note**:

```release-note
k8s libraries update to 20.2
```


pkg/controller/clone-controller.go
```diff
@@ -168,7 +168,7 @@ func (r *CloneReconciler) shouldReconcile(pvc *corev1.PersistentVolumeClaim, log
 }
 
 // Reconcile the reconcile loop for host assisted clone pvc.
-func (r *CloneReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+func (r *CloneReconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
```
awels, line 171: Random question here, is this the context with back ground or just todo? Would it make sense to use this context for the crud operations in the reconcile loop?

awels (approved):
/approve
looks good to me, one random question

pkg/controller/clone-controller.go
mhenriks: Pretty sure `TODO` and `Background` are the same.  We are not using contexts anywhere but may be wise to

awels: The idea is to use the context passed into the reconcile loop instead of creating context.TODO, I don't think it makes much difference.

awels:
That failure looks kinda weird, the lane failed due to the filler pod not being able to start because the PVC is not ready. And the reason it is not ready seems to be the PVC is lost. Looking at the logs in ceph, it appears something to do with a snapshot. I am very confused by the whole thing but something to investigate maybe?

mhenriks:
@awels seen that issue before.  interesting flake that I've been trying to figure out but I don't think related to anything in this PR.

This event may be a good clue:

```
"Unable to attach or mount volumes: unmounted volumes=[source-pvc], unattached volumes=[source-pvc default-token-mkrxr]: error processing PVC cdi-e2e-tests-cloner-hjnzl/source-pvc: failed to fetch PV pvc-85e37d74-01a9-42af-9fc9-108fbfc3ecd8 from API server: persistentvolumes \"pvc-85e37d74-01a9-42af-9fc9-108fbfc3ecd8\" is forbidden: User \"system:node:node01\" cannot get resource \"persistentvolumes\" in API group \"\" at the cluster scope: no relationship found between node 'node01' and this object"
```

awels:
> @awels seen that issue before. interesting flake that I've been trying to figure out but I don't think related to anything in this PR.
> 
> This event may be a good clue:
> 
> ```
> "Unable to attach or mount volumes: unmounted volumes=[source-pvc], unattached volumes=[source-pvc default-token-mkrxr]: error processing PVC cdi-e2e-tests-cloner-hjnzl/source-pvc: failed to fetch PV pvc-85e37d74-01a9-42af-9fc9-108fbfc3ecd8 from API server: persistentvolumes \"pvc-85e37d74-01a9-42af-9fc9-108fbfc3ecd8\" is forbidden: User \"system:node:node01\" cannot get resource \"persistentvolumes\" in API group \"\" at the cluster scope: no relationship found between node 'node01' and this object"
> ```

Yeah I understand it has nothing to do with this PR, I think I have seen it before, but with rare flakes like that if there is an instance that might have some extra information, its good to take a look at it.

PR #1816, opened 2021-06-02
Add "AwaitingVDDK" back to condition reason.

mrnold:
When waiting for the PVC to be bound, copy the PVC's condition reason to the DataVolume instead of just "Pending".

Fixes [BZ#1965181](https://bugzilla.redhat.com/show_bug.cgi?id=1965181)

**Release note**:
```release-note
Copy AwaitingVDDK condition reason to DV when PVC is waiting for v2v-vmware ConfigMap to be created.
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -174,6 +174,8 @@ const (
 	VddkConfigMap = "v2v-vmware"
 	// VddkConfigDataKey is the name of the ConfigMap key of the VDDK image reference
 	VddkConfigDataKey = "vddk-init-image"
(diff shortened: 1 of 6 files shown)


awels:
Seems like this test is failing in all the lanes: Preallocation All import paths should contain Preallocation step VddkImport

mrnold:
> Seems like this test is failing in all the lanes: Preallocation All import paths should contain Preallocation step VddkImport

Yes, I think it's because I removed the v2v-vmware ConfigMap for the test and forgot to put it back. I'm trying out a fix now.

awels (approved):
/approve

PR #1818, opened 2021-06-03
Add a datavolume condition for the image being too large

maya-r:
- Make sure that the nbdkit messages are in the terminating message, and appear at the exit.
- Avoid newlines in the terminating message, as they don't appear in the `kubectl get pod .. -o yaml` view.
- Avoid CrashLoopBackOff becoming a condition, we prefer to know why the crash had happened.
- If the message is a known long message about the image being too small, pass that stirng.
Tests are adjusted to take advantage of this condition.

This is intended for [BZ #1893526](https://bugzilla.redhat.com/show_bug.cgi?id=1893526).
Makes it easier to tell that the DV failed because it's too small - for humans and programmers alike.

I haven't found out why the termination message doesn't show past newlines.
Perhaps that's just the kubectl get .. -o yaml view.

**Release note**:

```release-note
Avoid "crashloopbackoff" becoming a datavolume condition message.
Make "DataVolume too small to contain image" a possible datavolume condition message, instead of dumping nbdkit logs.
Avoid the terminating message getting cut off, and improve it to include nbdkit logs when they exist.
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rsa"
 	"fmt"
+	"strconv"
 	"strings"
 
 	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
@@ -547,6 +548,11 @@ func podSucceededFromPVC(pvc *v1.PersistentVolumeClaim) bool {
 
 func setAnnotationsFromPodWithPrefix(anno map[string]string, prefix string, pod *v1.Pod) {
 	if pod.Status.ContainerStatuses != nil {
+		annPodRestarts, _ := strconv.Atoi(anno[AnnPodRestarts])
+		podRestarts := int(pod.Status.ContainerStatuses[0].RestartCount)
+		if podRestarts >= annPodRestarts {
+			anno[AnnPodRestarts] = strconv.Itoa(podRestarts)
+		}
 		if pod.Status.ContainerStatuses[0].State.Running != nil {
 			anno[prefix] = "true"
 			anno[prefix+".message"] = ""
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -3,7 +3,9 @@ package controller
 import (
 	"context"
 	"crypto/rsa"
+	"encoding/json"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -27,6 +29,7 @@ import (
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
 	cdiv1utils "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1/utils"
 	"kubevirt.io/containerized-data-importer/pkg/common"
+	"kubevirt.io/containerized-data-importer/pkg/util"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert"
 	"kubevirt.io/containerized-data-importer/pkg/util/naming"
 	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
@@ -144,6 +147,10 @@ const (
 	AnnPodSidecarInjection = "sidecar.istio.io/inject"
 	// AnnPodSidecarInjectionDefault is the default value passed for AnnPodSidecarInjection
 	AnnPodSidecarInjectionDefault = "false"
+)
+
+var (
+	vddkInfoMatch = regexp.MustCompile(`((.*; )|^)VDDK: (?P<info>{.*})`)
 )
 
 func isCrossNamespaceClone(dv *cdiv1.DataVolume) bool {
@@ -555,6 +562,7 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, prefix string, pod 
 	if podRestarts >= annPodRestarts {
 		anno[AnnPodRestarts] = strconv.Itoa(podRestarts)
 	}
+	setVddkAnnotations(anno, pod)
 	if pod.Status.ContainerStatuses[0].State.Running != nil {
 		anno[prefix] = "true"
 		anno[prefix+".message"] = ""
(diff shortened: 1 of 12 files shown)


awels:
if you rebase, you should have better luck getting the CI to pass now.

awels:
if you rebase the leader election failures should disappear. I also reviewed the code, and for some reason it feels like there is some overlap with https://github.com/kubevirt/containerized-data-importer/pull/1752 as that is also watching the nbdkit log for certain strings. Can we refactor some of that to combine the logic?

pkg/controller/util.go
```diff
@@ -546,28 +554,74 @@ func podSucceededFromPVC(pvc *v1.PersistentVolumeClaim) bool {
...
-			if pod.Status.ContainerStatuses[0].State.Waiting != nil {
-				anno[prefix+".message"] = pod.Status.ContainerStatuses[0].State.Waiting.Message
-				anno[prefix+".reason"] = pod.Status.ContainerStatuses[0].State.Waiting.Reason
-			} else if pod.Status.ContainerStatuses[0].State.Terminated != nil {
-				anno[prefix+".message"] = pod.Status.ContainerStatuses[0].State.Terminated.Message
-				anno[prefix+".reason"] = pod.Status.ContainerStatuses[0].State.Terminated.Reason
-				if strings.Contains(pod.Status.ContainerStatuses[0].State.Terminated.Message, common.PreallocationApplied) {
-					anno[AnnPreallocationApplied] = "true"
-				}
+func setAnnotationsFromPodWithPrefix(anno map[string]string, prefix string, pod *v1.Pod) {
```
brybacki, line 557: Maybe a nitpick: With such a name It would be easier to read if the parameter names follow what is mentioned in the name so:
the name states setAnnotationsFromPodWithPrefix(anno, pod, prefix)

pkg/controller/util.go
```diff
@@ -546,28 +554,74 @@ func podSucceededFromPVC(pvc *v1.PersistentVolumeClaim) bool {
...
+	if pod == nil || pod.Status.ContainerStatuses == nil {
+		return
+	}
+	annPodRestarts, _ := strconv.Atoi(anno[AnnPodRestarts])
+	podRestarts := int(pod.Status.ContainerStatuses[0].RestartCount)
+	if podRestarts >= annPodRestarts {
+		anno[AnnPodRestarts] = strconv.Itoa(podRestarts)
+	}
+	setVddkAnnotations(anno, pod)
+	if pod.Status.ContainerStatuses[0].State.Running != nil {
```
brybacki, line 567: extracting pod.Status.ContainerStatuses[0].State as a variable would make the following code easier to read. It is used at least 9 times.

pkg/controller/util.go
```diff
@@ -546,28 +554,74 @@ func podSucceededFromPVC(pvc *v1.PersistentVolumeClaim) bool {
...
-				anno[prefix+".message"] = pod.Status.ContainerStatuses[0].State.Terminated.Message
-				anno[prefix+".reason"] = pod.Status.ContainerStatuses[0].State.Terminated.Reason
-				if strings.Contains(pod.Status.ContainerStatuses[0].State.Terminated.Message, common.PreallocationApplied) {
-					anno[AnnPreallocationApplied] = "true"
-				}
+func setAnnotationsFromPodWithPrefix(anno map[string]string, prefix string, pod *v1.Pod) {
+	if pod == nil || pod.Status.ContainerStatuses == nil {
+		return
+	}
+	annPodRestarts, _ := strconv.Atoi(anno[AnnPodRestarts])
```
brybacki, line 561: Very good, that this "restarts" code is now here instead of being spread in many places in code.

pkg/image/nbdkit.go
```diff
@@ -240,10 +240,18 @@ func (n *Nbdkit) StartNbdkit(source string) error {
 
 // Default nbdkit log watcher, just logs lines as nbdkit prints them.
 func watchNbdLog(output *bufio.Reader) {
+	f, err := os.Create("/tmp/nbdkit.log")
```
brybacki, line 243: Ddescription states "Default nbdkit log watcher, just logs lines as nbdkit prints them."  but now this also creates its own f log file.

Can you tell me why this log is needed? It is stored on the pod where nbd is run, am I right? And the pod is only running for the time it takes operation to complete so there is no problem with managing the log size.

pkg/image/qemu.go
```diff
@@ -121,7 +122,11 @@ func convertToRaw(src, dest string, preallocate bool) error {
 	}
 	if err != nil {
 		os.Remove(dest)
-		return errors.Wrap(err, "could not convert image to raw")
+		errorMsg := "could not convert image to raw"
+		if nbdkitLog, err := ioutil.ReadFile("/tmp/nbdkit.log"); err == nil {
```
brybacki, line 126: Can you create a constant for the log name?

pkg/image/nbdkit.go
maya-r: Yes, there is no problem with managing the log size. It's a few lines at the importer pod, which is cleaned up automatically when the importer completes.
I wanted the messages to be part of the termination message in a failure, but didn't want to stop printing them as they happen, in case some useful message appears at success.
There didn't seem to be a great way to run this process, monitor its output, and receive the updated output when something else terminates unsuccessfully. I also didn't see a way to make nbdkit log to two places.

pkg/image/qemu.go
maya-r: Done

pkg/controller/util.go
maya-r: Done, thanks for the suggestion.

pkg/controller/util.go
maya-r: Done

pkg/image/nbdkit.go
awels: So looking at the code, we are writing to /tmp/nbdkit.log but then never use it for anything as far as I can tell. Am I missing something?

maya-r: [Used here](https://github.com/kubevirt/containerized-data-importer/pull/1818/files#diff-911dad1f9fe8bfd9b383cded047f7472b0b56ad8b201ac8a3f6fad28f00605ab)

pkg/util/util.go
```diff
@@ -232,6 +232,7 @@ func WriteTerminationMessage(message string) error {
 
 // WriteTerminationMessageToFile writes the passed in message to the passed in message file
 func WriteTerminationMessageToFile(file, message string) error {
+	message = strings.Replace(message, "\n", " ", -1)
```
awels, line 235: You can use strings.ReplaceAll here, which is basically the same.

awels (approved):
/lgtm

awels:
Not going to block this PR on the Replace vs ReplaceAll, unless you want to change it.

maya-r:
Switched to ReplaceAll and rebased on latest main (so I can test it -- needed newer fedora checksum)

awels (approved):
/approve

PR #1820, opened 2021-06-04
Adding support to specify externally managed certs

visheshtanksale:

- Add flags to API Server, Controller and upload proxy to specify custom certificate paths. Custom certs path can be used to populate externally managed certs.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -36,7 +36,15 @@ import (
 
 const (
 	readyFile = "/tmp/ready"
+	defaultUploadServerKeyFile = "/var/run/certs/cdi-uploadserver-signer/tls.key"
+	defaultUploadServerCertFile = "/var/run/certs/cdi-uploadserver-signer/tls.crt"
+	defaultUploadClientKeyFile = "/var/run/certs/cdi-uploadserver-client-signer/tls.key"
+	defaultUploadClientCertFile = "/var/run/certs/cdi-uploadserver-client-signer/tls.crt"
+	defaultUploadServerCABundleConfigMap = "cdi-uploadserver-signer-bundle"
+	defaultUploadClientCABundleConfigMap = "cdi-uploadserver-client-signer-bundle"
 )
+
+
 
 var (
 	kubeconfig             string
@@ -49,6 +57,12 @@ var (
 	pullPolicy             string
 	verbose                string
 	log                    = logf.Log.WithName("controller")
+	uploadServerKeyFile   string
+	uploadServerCertFile  string
+	uploadClientKeyFile   string
+	uploadClientCertFile  string
+	uploadServerCABundleConfigMap string
+	uploadClientCABundleConfigMap string
 )
 
 // The importer and cloner images are obtained here along with the supported flags. IMPORTER_IMAGE, CLONER_IMAGE, and UPLOADSERVICE_IMAGE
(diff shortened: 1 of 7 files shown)


awels:
@mhenriks can you review?

mhenriks:
@visheshtanksale thanks for contribution!  These changes look like a good start!  But just want to get a better idea how these parameters could be exposed in the CDI CR and propagated by cdi-operator.  That work does not have to be done in this PR, but we should at least have a rough API for cdi-operator managed installs.


One other question:  a bunch of "signer" certs created for a default install.  In an externally managed world, do you still care about those?

visheshtanksale:
> @visheshtanksale thanks for contribution! These changes look like a good start! But just want to get a better idea how these parameters could be exposed in the CDI CR and propagated by cdi-operator. That work does not have to be done in this PR, but we should at least have a rough API for cdi-operator managed installs.
> 
> One other question: a bunch of "signer" certs created for a default install. In an externally managed world, do you still care about those?

@mhenriks I have tried to put down high level design for supporting externally managed certs on CDI [here](https://github.com/kubevirt/containerized-data-importer/issues/1829). Please let me know what are your thoughts.

mhenriks:
Hi @visheshtanksale thanks for creating the issue!  we can track that separately while keeping it in mind as we review this

cmd/cdi-controller/controller.go
```diff
@@ -58,6 +72,12 @@ var (
 func init() {
 	// flags
 	flag.StringVar(&masterURL, "server", "", "(Optional) URL address of a remote api server.  Do not set for local clusters.")
+	flag.StringVar(&uploadServerKeyFile, "uploadserver-ca-key", defaultUploadServerKeyFile,"(Optional) Client certificate for Upload Server")
+	flag.StringVar(&uploadServerCertFile, "uploadserver-ca-cert", defaultUploadServerCertFile, "(Optional) Private key for the client certificate for Upload Server")
+	flag.StringVar(&uploadClientKeyFile, "uploadserver-client-ca-key", defaultUploadClientKeyFile, "(Optional) Client certificate for Upload Server Client")
+	flag.StringVar(&uploadClientCertFile, "uploadserver-client-ca-cert", defaultUploadClientCertFile, "(Optional) Private key for the client certificate for Upload Server Client")
+	flag.StringVar(&uploadServerCABundleConfigMap, "uploadserver-ca-config-map", defaultUploadServerCABundleConfigMap, "(Optional) Name of Config Map containing CA Bundle for Upload Server")
+	flag.StringVar(&uploadClientCABundleConfigMap, "uploadserver-client-ca-config-map", defaultUploadClientCABundleConfigMap, "(Optional) Name of Config Map containing CA Bundle for Upload Server Client")
```
mhenriks, line 80: The big change I would like to discuss in this PR is moving from command line args to environment variables.  We don't use command line args much.  Clearly, some are there, but I don't think we ever invoke them.

I'm thinking that for each relevant private key/cert there will be mandatory environment variables with the path to the mounted file.  Variables named `UPLOADSERVER_KEY_FILE`, `UPLOADSERVER_CERT_FILE`, etc.  What do you think?

visheshtanksale: Thanks for taking a look. Yes environment variables seem better option than command line args. Updated to take in custom args using env variables.

mhenriks:
Thanks for the contribution @visheshtanksale!

PR #1822, opened 2021-06-07
Add Importer functional tests for Istio sidecar injection annotation

arnongilboa:

Test Import on namespace with Istio sidecar injection, which blocks all external resource access.
Injection is disabled by default on importer pod, so it should successfully import.
However when injection is explicitly enabled by DV, import should fail.

Istio is supported on kubevirtci only from `k8s-1.20`
To deploy Istio, env should have `KUBEVIRT_DEPLOY_ISTIO=true`
Therefore we added a new `k8s-1.20-hpp-istio` (Istio deployed) lane to CDI ci

**Release note**:
```release-note
NONE
```


tests/import_test.go
```diff
@@ -137,6 +141,83 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+
+	It("[test_id:6498] Should fail to import with namespace sidecar injection enabled, and sidecar.istio.io/inject set to true", func() {
+		value := os.Getenv("KUBEVIRT_DEPLOY_ISTIO")
+		if value != "true" {
+			Skip("No Istio enabled, skipping.")
+		}
+
+		dataVolume := utils.NewDataVolumeWithHTTPImport("istio-sidecar-injection-test", "100Mi", tinyCoreIsoExternalURL)
+		By(fmt.Sprintf("Create new datavolume %s", dataVolume.Name))
+		dataVolume.Annotations[controller.AnnImmediateBinding] = "true"
```
awels, line 173: We have a function called
```
ForceBindIfWaitForFirstConsumer
```
Or
```
ForceBindPvcIfDvIsWaitForFirstConsumer
```
If you are using a DV.

tests/import_test.go
```diff
@@ -1203,3 +1284,20 @@ var _ = Describe("Preallocation", func() {
 		Expect(ok).To(BeTrue())
 	})
 })
+
+func generateRegistryOnlySidecar() *unstructured.Unstructured {
```
awels, line 1288: istio doesn't publish a k8s client?

arnongilboa: We only use a single trivial Istio resource (Sidecar). Do we really want to import 10k lines of code worth of dependencies to have the [istio structs and go-client](https://github.com/istio/client-go)?

tests/import_test.go
arnongilboa: In case I use the above, we'll have the consumer pod which will also fail due to the Istio sidecar (already tested). 
I've set the Immediate Binding annotation to true, so afaiu we have no need at all for the consumer pod here.

awels: Okay, can you put in a comment indicating why doing the annotation is superior to using the function (the consumer pod is messed up) so that a future self/reviewer can understand this.

tests/import_test.go
awels: Hmm I figured it would only pull in the things we actually use, but if it is adding 10k lines of code just for something small, I can see why you went this route.

tests/import_test.go
arnongilboa: done

awels (approved):
/approve

PR #1824, opened 2021-06-07
CSI Volume clone

brybacki:

Fixes #
https://github.com/kubevirt/containerized-data-importer/issues/1467

In short this is a rewrite of PR: "Feat CSI Clone Support " https://github.com/kubevirt/containerized-data-importer/pull/1552 to utilize "Cluster scoped DataVolume/PVC namespace transfer API" #1673 

PR is is now open for review but is not yet in the good shape. The main design is here, and I see at least 3 things to be done:
- [x] Correct the clone/expansion coordination - there is a bug somewhere, I am now working on it
- [x] Extract common code  - there is big similarity between CSI Volume Clone and Smart Clone, some code can be made common/simplified
- [x] Documentation

**Release note**:

```release-note
CDI can utilize CSI Volume Cloning
```


brybacki:
PR is is now open for review but is not yet in the good shape. The main design is here, and I see at least 3 things to be done:
- Correct the clone/expansion coordination - there is a bug somewhere, I am now working on it
- Extract common code  - there is big similarity between CSI Volume Clone and Smart Clone, some code can be made common/simplified
- Documentation

awels (commented):
So I am a little confused [according to this](https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/) all we have to do is create a PVC with a datasource. Which would eliminate most if not all of the csi clone controller. In the DV controller if we decide that we are going to csi clone, just make the PVC and set the datasource to be the source defined in the DV. What am I missing here?

brybacki:
> So I am a little confused [according to this](https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/) all we have to do is create a PVC with a datasource. Which would eliminate most if not all of the csi clone controller. In the DV controller if we decide that we are going to csi clone, just make the PVC and set the datasource to be the source defined in the DV. What am I missing here?

You are right, I must have been blinded by fighting some bugs for a few days ;). I have simplified the controllers by removing the CSI - clone controller, and added one check to the DV controller.

pkg/controller/datavolume-controller.go
```diff
@@ -595,24 +553,194 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+		pvcClashControllerRef := metav1.GetControllerOf(pvc)
+
+		if pvc.Name == cloneTargetPvc.Name &&
+			pvc.Namespace == cloneTargetPvc.Namespace &&
+			!reflect.DeepEqual(pvcControllerRef, pvcClashControllerRef) {
+			return reconcile.Result{}, errors.Errorf("Target Pvc Name in use") //r.updateCSICloneDataVolumeStatus(cdiv1.ClonePVCNameInUse, datavolume)
+		}
+
+		if pvc.Status.Phase == corev1.ClaimBound {
+			if v, ok := pvc.Annotations[AnnCloneOf]; !ok || v != "true" {
```
awels, line 616: I see this code several times, maybe we can make it a function, like 'setCloneOfOnPvc'?

pkg/controller/datavolume-controller.go
```diff
@@ -595,24 +553,194 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+		return reconcile.Result{}, err
+	}
+
+	if !done {
+		return reconcile.Result{Requeue: true},
+			r.updateCloneStatusPhase(cdiv1.ExpansionInProgress, datavolume, nil, selectedCloneStrategy)
+	}
+
+	// trigger transfer and next reconcile should have pvcExists == true
+	tmpPVC.Annotations[annReadyForTransfer] = "true"
```
awels, line 695: expandTmpPvcAfterClone and expandPvcAfterClone look very similar, except for some additional annotations and the updateCloneStatusPhase is different. I think maybe we can combine some of that?

pkg/controller/datavolume-controller.go
```diff
@@ -1483,6 +1605,59 @@ func newSnapshot(dataVolume *cdiv1.DataVolume, snapshotName, snapshotClassName s
 	return snapshot
 }
 
+// NewVolumeClonePVC creates a PVC object to be used during CSI volume cloning. csiClonePvcType defines whether the PVC object will be a source cloner PVC or the target PVC
+func newVolumeClonePVC(dv *cdiv1.DataVolume, sourcePvc *corev1.PersistentVolumeClaim, targetPvcSpec *corev1.PersistentVolumeClaimSpec, pvcName string) *corev1.PersistentVolumeClaim {
```
awels, line 1609: Can we just call newPersistentVolumeClaim, and override the pvc.Spec.DataSource before returning the result. That way all the logic that sets annotations and ownerRefs and everything else is contained in that function.

brybacki: I can try, we will see if that is better

brybacki: Done

pkg/controller/datavolume-controller.go
brybacki: Done

pkg/controller/datavolume-controller.go
```diff
@@ -595,24 +546,201 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+	inUse, err := r.sourceInUse(datavolume, CSICloneSourceInUse)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	populated, err := r.isSourcePVCPopulated(datavolume)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if inUse || !populated {
+		return reconcile.Result{Requeue: true},
```
mhenriks, line 582: This is an okay use of requeue because we may not be watching/reconciling these pods

pkg/controller/datavolume-controller.go
```diff
@@ -595,24 +546,201 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+			return reconcile.Result{}, err
+		}
+
+		// done, done
+		return reconcile.Result{}, nil
+	}
+
+	//DO Nothing, not yet ready
+	if pvc.Annotations[AnnCloneOf] != "true" {
+		return reconcile.Result{Requeue: true}, nil
```
mhenriks, line 641: Shouldn't have to requeue

pkg/controller/datavolume-controller.go
```diff
@@ -595,24 +546,201 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
-			tmpPVC.Annotations[AnnPopulatedFor] = datavolume.Name
-			if err := r.client.Update(context.TODO(), tmpPVC); err != nil {
+		if pvc.Status.Phase == corev1.ClaimBound {
+			if err := r.setCloneOfOnPvc(pvc); err != nil {
 				return reconcile.Result{}, err
 			}
+		}
+	}
+
+	return reconcile.Result{Requeue: true}, r.updateCloneStatusPhase(cdiv1.CSICloneInProgress, datavolume, nil, CsiClone)
```
mhenriks, line 618: This requeue necessary?

mhenriks (commented):
Looks pretty good but agree with @awels re: duplicated code.  Do what you can to clean up but would like to get this in.

See comments about requeue.  Usually only be necessary when waiting on resources that are not watched to change

pkg/controller/datavolume-controller.go
```diff
@@ -425,92 +433,30 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
-			return false, errors.New("source PVC not found")
+	if selectedCloneStrategy == SmartClone || selectedCloneStrategy == CsiClone {
+		if selectedCloneStrategy == CsiClone {
+			if pvc.Status.Phase == corev1.ClaimBound {
+				if err := r.setCloneOfOnPvc(pvc); err != nil {
+					return reconcile.Result{}, err
+				}
+			} else if pvc.Status.Phase == corev1.ClaimPending {
+				//not yet ready
+				return reconcile.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
brybacki: right, and the Phase will change

pkg/controller/datavolume-controller.go
brybacki: I was under impression, that when we do Clone, we might want to requeue the same way as in Import to check progress/update status, but no I think that for CSI Clone it does not make any sense, there is nothing to reconcile until PVC is either Bound or Lost (or has some warning event)

Removed

pkg/controller/datavolume-controller.go
brybacki: Removed

pkg/controller/datavolume-controller.go
```diff
@@ -425,92 +433,30 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+			}
+			// TODO: is there any work for ClaimLost,
```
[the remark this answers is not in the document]
brybacki: Do you think it is ok to do:
```
phase = cdiv1.Failed, 
eventType: corev1.EventTypeWarning,
reason:    ErrClaimLost,
```

When you see ClaimLost, then it is probably better to investigate, delete DV (to cleanup any transfer objects etc...), fix what is wrong and retry.

pkg/controller/datavolume-controller.go
brybacki: Looks like it does not work without requeue here, investigating.

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
mhenriks: I've seen cases become unlost so as long as "failed" is not terminal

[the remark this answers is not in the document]
awels: I can in fact give an example of becoming 'unlost'. Create a manual PV and PVC, and ensure they become bound. Then 'export' the yaml for both. Then delete the PV and PVC. Then create both PV and PVC again, the PVC will be lost for a little bit until it gets matched back with the PV.

pkg/controller/datavolume-controller.go
brybacki: Some coordination between Namespace transfer controller and datavolume controller is probably not handled explicitely in this case, when I add Requeue it works, but I think I can be implement in a better way.

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
brybacki: ok, so do you suggest a new phase - like  CloneTargetPVCLost, or leaving the Failed and just testing the scenario to make sure it can eventually continue if PVC changed from ClaimLost to Bound?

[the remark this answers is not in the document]
brybacki: Updated one condition and according to new test the CSI clone logic can continue when PVC is recovered from ClaimLost

pkg/controller/datavolume-controller.go
brybacki: finally got it, not needed, there was a wrong owner was set on transfer/temporary pvc (a DV from different namespace), and it was being deleted. I am not sure if k8s GC was deleting it or our logic in some other branch of code, but now after fixing that problem it works.

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
brybacki: https://github.com/kubevirt/containerized-data-importer/blob/b04a47574e3e58a322d22bfc5629b55a2bfbc314/pkg/controller/datavolume-controller_test.go#L1195

brybacki:
documentation updated

pkg/controller/datavolume-controller.go
brybacki: I did try extracting the expand:
```
func (r *DatavolumeReconciler) expandPvcAfterClone(log logr.Logger,
	datavolume *cdiv1.DataVolume,
	pvc *corev1.PersistentVolumeClaim,
	pvcSpec *corev1.PersistentVolumeClaimSpec,
	selectedCloneStrategy cloneStrategy) (bool, error) {
	done, err := r.expand(log, datavolume, pvc, pvcSpec)
	if err != nil {
		return false, err
	}
	if !done {
		return false, r.updateCloneStatusPhase(cdiv1.ExpansionInProgress, datavolume, pvc, selectedCloneStrategy)
	}

	// done
	return true, nil
}
```

But it does not look much better in my opinion.  Any ideas?

brybacki:
@awels @mhenriks please take a look, it is now ready

tests/utils/storageprofile.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return nil
+}
+
+// ConfigureCloneStrategy updates the storageProfile found by name, with given CDICloneStrategy.
+func ConfigureCloneStrategy(client client.Client,
+	clientSet *cdiclientset.Clientset,
+	storageClassName string,
+	spec *cdiv1.StorageProfileSpec,
+	cloneStrategyCsiClone cdiv1.CDICloneStrategy) error {
```
awels, line 47: this is for configuring any clone strategy right? Why is the variable called cloneStrategyCsiClone?

pkg/controller/datavolume-controller.go
```diff
@@ -600,24 +557,201 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+	populated, err := r.isSourcePVCPopulated(datavolume)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if inUse || !populated {
+		return reconcile.Result{Requeue: true},
+			r.updateCloneStatusPhase(cdiv1.CloneScheduled, datavolume, nil, CsiClone)
+	}
+
+	// here use Temp PVC!!!
```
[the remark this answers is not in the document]
brybacki: It was an old TODO comment to myself, removed now

tests/utils/storageprofile.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// GetStorageProfileSpec attempts to get the StorageProfile for storageClass by Name.
+func GetStorageProfileSpec(clientSet *cdiclientset.Clientset, storageClassName string) (*cdiv1.StorageProfileSpec, error) {
+	storageProfile, err := clientSet.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	originalProfileSpec := storageProfile.Spec.DeepCopy()
```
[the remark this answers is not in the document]
brybacki: No

pkg/controller/datavolume-controller.go
```diff
@@ -600,24 +557,201 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+	log.Info("Creating PVC for datavolume")
+	cloneTargetPvc, err := r.newVolumeClonePVC(datavolume, sourcePvc, pvcSpec, pvcName)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	pvc := &corev1.PersistentVolumeClaim{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: cloneTargetPvc.Namespace, Name: cloneTargetPvc.Name}, pvc); err != nil {
```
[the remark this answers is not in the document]
brybacki: this makes sense, I'll do it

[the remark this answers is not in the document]
brybacki: Now I think I'll leave it like it is now:

Currently it does:
get:
- notExists -> create
- exists -> check for name clash

New way would be like this, and probably harder to cover all possibilities:
create:
- created - ok
- alreadyExists:
 -- get:
    -exists -> check for name clash
    - *not exists (low probability but actually might happen)*

@awels  What do you think

pkg/controller/datavolume-controller.go
```diff
@@ -600,24 +557,201 @@ func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger, datavolum
...
+	return ""
+}
+
+func (r *DatavolumeReconciler) reconcileSmartClonePvc(log logr.Logger,
+	datavolume *cdiv1.DataVolume,
+	pvcSpec *corev1.PersistentVolumeClaimSpec, transferName string, snapshotClassName string) (reconcile.Result, error) {
+
+	pvcName := datavolume.Name
+
+	if isCrossNamespaceClone(datavolume) {
```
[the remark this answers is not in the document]
brybacki: doCrossNamespaceClone done

brybacki:
I have applied almost all the changes suggested in the comments. Only one change request was not applied:

https://github.com/kubevirt/containerized-data-importer/pull/1824/files?file-filters%5B%5D=.bazel&file-filters%5B%5D=.go&file-filters%5B%5D=.md#diff-8f140feb34d5514d8336c8181a67578e3d4938a6bade5129cf24d78d80986f45R587

"So why not simply attempt to create the PVC, and if that fails with an IsAlreadyExists drop into the else part. Something like this:"

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
awels: I was trying to figure out why you would need to get the PVC again, since you have everything you need from the PVC definition you attempted to created, but then I noticed in the else you are checking the pvc.Status.Phase, which you can only get from querying the PVC. For this reason I am okay with leaving it as is.

awels (approved):
/lgtm

PR #1830, opened 2021-06-08
Add Google Cloud Storage (GCS) support in CDI

didovesei:

This PR aims to enable users to import their VM images that are stored on Google Cloud Storage (GCS). Specifically this PR adds a new `GCSDataSource` that fetches the images from the specified GCS bucket.

Users can create a DV config as below to use it:

```
kind: DataVolume
metadata:
  name: centos-boot-disk
spec:
  source:
    gcs:
      url: gs://[bucket]/[object]
  pvc:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 30G
    storageClassName: local-shared
```

Hi Kubevirt CDI community, for now this PR only includes the implementation for the GCS feature and corresponding build files.
While I am finishing the testing and API doc, I would like to collect some initial ideas from you on this feature as well as the code. Thank you in advance!

```release-note
Enhancement: Add new data source of Google Cloud Storage.
```


awels:
To get the dco check to be happy, on each commit do git commit -as the -s will sign the commit.

awels:
/ok-to-test
I did a quick review of the code, and I see nothing jumping out at me. The setup is very similar to an s3 data source. It also has the same draw back the s3 data source has. If the source needs to be converted, we need scratch space to write the data to, before converting. On an http datasource, in most cases we can use nbdkit with appropriate filters and plugins to stream the data directly to the target. I am not aware of a GCS filter/plugin for nbd so we are stuck with this method. In order for an ndbkit filter/plugin to be feasible GCS would need to support byte ranges (which is why in some cases http doesn't work, if the server doesn't support byte ranges)

pkg/importer/gcs-datasource.go
```diff
@@ -0,0 +1,107 @@
...
+	ep *url.URL
+	// Stack of readers
+	readers *FormatReaders
+	// Reader
+	gcsReader *storage.Reader
+	// The image file in scratch space
+	url *url.URL
+}
+
+func NewGCSDataSource(endpoint string) (*GCSDataSource, error) {
```
mhenriks, line 27: I'm not familiar with GCS, but are any auth methods that we may want to support?  Looks like only supporting signed or public access URLs?

didovesei: Hi, this implementation works with Google ADC and expects that the ADC is already set up on the worker node. The Google cloud client library will be responsible for looking for the credentials for Google cloud and authenticates to google cloud services (i.e., GCS in this case). So it does support both public and private GCS resources.

As the next step, we are planning to enable CDI to use a different Google service account other than the default one configured on the node to access GCS. The way to configure a different service account is under discussion.

didovesei:
> /ok-to-test
> I did a quick review of the code, and I see nothing jumping out at me. The setup is very similar to an s3 data source. It also has the same draw back the s3 data source has. If the source needs to be converted, we need scratch space to write the data to, before converting. On an http datasource, in most cases we can use nbdkit with appropriate filters and plugins to stream the data directly to the target. I am not aware of a GCS filter/plugin for nbd so we are stuck with this method. In order for an ndbkit filter/plugin to be feasible GCS would need to support byte ranges (which is why in some cases http doesn't work, if the server doesn't support byte ranges)

Hi, thanks for the note. I am not sure if GCS support byte ranges. I will do some research and get back to you.

mhenriks:
@didovesei `make deps-update` may fix the `generate-verify` failures.  may also want to run `make generate` as well

awels:
To fix the DCO, check out the commit `0a5b5f3 test part 2` and git commit --amend -s to sign it. then go back to the head of the branch and re-push

awels:
also make test-lint will tell you what to do to make the linter happy.

didovesei:
> > /ok-to-test
> > I did a quick review of the code, and I see nothing jumping out at me. The setup is very similar to an s3 data source. It also has the same draw back the s3 data source has. If the source needs to be converted, we need scratch space to write the data to, before converting. On an http datasource, in most cases we can use nbdkit with appropriate filters and plugins to stream the data directly to the target. I am not aware of a GCS filter/plugin for nbd so we are stuck with this method. In order for an ndbkit filter/plugin to be feasible GCS would need to support byte ranges (which is why in some cases http doesn't work, if the server doesn't support byte ranges)
> 
> Hi, thanks for the note. I am not sure if GCS support byte ranges. I will do some research and get back to you.

I took a look at the GCS go sdk, it seems that they allow to resume the transfer but does not provide an API for byte range request. I guess we will still need the scratch space to temporarily store the image.

awels:
That is unfortunate, but not unexpected, it will still work with scratch space, but obviously we would like to not use it if not needed. I will take a look at the PR soon.

didovesei:
Thanks all for the instructions on the test and dco! I have added a feature that allows the user to configure a secret with a Google service account key so that the importer can use that SA key to access GCS.

I have added the unit test and fixed the lint errors. It should be good for another look. Thanks!

didovesei:
> That is unfortunate, but not unexpected, it will still work with scratch space, but obviously we would like to not use it if not needed. I will take a look at the PR soon.

Thanks!

awels:
Hi, I have reviewed the PR, and I don't see anything that looks concerning. In general we have a rule, if there are no functional tests, then a feature doesn't exist. Basically we want to ensure that if we modify code, we don't break stuff, and that requires functional testing. However I have no idea how we could write a functional test for this feature. I would require a google storage account. Let me discuss this with the team, and see what we can do.

didovesei:
> Hi, I have reviewed the PR, and I don't see anything that looks concerning.

Thanks for reviewing!

> In general we have a rule, if there are no functional tests, then a feature doesn't exist. Basically we want to ensure that if we modify code, we don't break stuff, and that requires functional testing.

I see, makes sense. 

> However I have no idea how we could write a functional test for this feature. I would require a google storage account. Let me discuss this with the team, and see what we can do.

Got it, please let me know how the dicussion goes :) Thanks!

mhenriks:
@didovesei any recommendation of GCS emulator we could use for functional testing?  There appear to be a few

didovesei:
> @didovesei any recommendation of GCS emulator we could use for functional testing? There appear to be a few

Hi, I have been testing it with the real GCS. I am not familiar with the emulators unfortunately..

didovesei:
Hi all, I am just checking in to see how the discussion is going. Also by the way, regarding the functional test, is it supposed to be included in this PR?

awels:
Hi, sorry haven't gotten back to you. I don't think you can't go that wrong with an emulator like https://github.com/fsouza/fake-gcs-server It already has releases to docker. So you should be able to configure a service for the test suite to run. We do prefer the PR that introduces a feature to include the functional test.

didovesei:
> Hi, sorry haven't gotten back to you. I don't think you can't go that wrong with an emulator like https://github.com/fsouza/fake-gcs-server It already has releases to docker. So you should be able to configure a service for the test suite to run. We do prefer the PR that introduces a feature to include the functional test.

Cool. Let me take a look at that emulator. Thanks!

PR #1836, opened 2021-06-17
Experiment forwarder restart to fix flakiness

brybacki:
This is a with quick and dirty WORKAROUND for the cases when kubectl
port forward stops working correctly after communication failure.

Using an ingress might be a better long term solution.

Fixes #

This is just an experiment, if it works I can be improved.

**Release note**:

```release-note
NONE
```


brybacki:
I have managed to catch some failures, and see that the re-try probably works. But there are now different failures (probably still connected to the original problem). 

In short, I did not find anything useful.

PR #1837, opened 2021-06-17
URL-encode special characters in VDDK credentials.

mrnold:
For VDDK data sources, this pull request encodes the login credentials so that they can be used in VMware login URLs. Without this, special characters in the username or password can cause immediate importer failures.

This also tries to avoid printing passwords in related error messages.

Fixes [RHBZ#1972895](https://bugzilla.redhat.com/show_bug.cgi?id=1972895)

**Release note**:
```release-note
VDDK: handle special characters in login credentials.
```

### DIFF
--- a/pkg/importer/vddk-datasource.go
+++ b/pkg/importer/vddk-datasource.go
@@ -248,18 +248,24 @@ func createVMwareClient(endpoint string, accessKey string, secKey string, thumbp
 	}
 
 	// Construct VMware SDK URL and get MOref
-	sdkURL := vmwURL.Scheme + "://" + accessKey + ":" + secKey + "@" + vmwURL.Host + "/sdk"
+	sdkURL := vmwURL.Scheme + "://" + url.PathEscape(accessKey) + ":" + url.PathEscape(secKey) + "@" + vmwURL.Host + "/sdk"
+	redacted := vmwURL.Scheme + "://" + url.PathEscape(accessKey) + ":" + "*****" + "@" + vmwURL.Host + "/sdk"
 	vmwURL, err = url.Parse(sdkURL)
 	if err != nil {
-		klog.Errorf("Unable to create VMware URL: %v", err)
-		return nil, err
+		return nil, errors.New("unable to create VMware URL from: " + redacted)
 	}
 
 	// Log in to vCenter
 	ctx, cancel := context.WithCancel(context.Background())
 	conn, err := govmomi.NewClient(ctx, vmwURL, true)
 	if err != nil {
-		klog.Errorf("Unable to connect to vCenter: %v", err)
+		if secKey != "" {
+			wrap := url.PathEscape(accessKey) + ":%s@"
+			target := fmt.Sprintf(wrap, url.PathEscape(secKey))
+			redacted = fmt.Sprintf(wrap, "*****")
+			redacted = strings.ReplaceAll(err.Error(), target, redacted)
+			err = errors.New("unable to connect to vCenter: " + redacted)
+		}
 		cancel()
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/importer/vddk-datasource.go
```diff
@@ -248,18 +248,24 @@ func createVMwareClient(endpoint string, accessKey string, secKey string, thumbp
...
 	// Construct VMware SDK URL and get MOref
-	sdkURL := vmwURL.Scheme + "://" + accessKey + ":" + secKey + "@" + vmwURL.Host + "/sdk"
+	sdkURL := vmwURL.Scheme + "://" + url.PathEscape(accessKey) + ":" + url.PathEscape(secKey) + "@" + vmwURL.Host + "/sdk"
+	redacted := vmwURL.Scheme + "://" + url.PathEscape(accessKey) + ":" + "*****" + "@" + vmwURL.Host + "/sdk"
 	vmwURL, err = url.Parse(sdkURL)
 	if err != nil {
-		klog.Errorf("Unable to create VMware URL: %v", err)
-		return nil, err
+		return nil, errors.New("unable to create VMware URL from: " + redacted)
 	}
```
dankenigsberg, line 256: I don't think that `PathEscape()` is doing what we need here. For example it does not encode `@`. I believe that the right thing to do is to construct a `url.URL` and use its String() function. Encoding is a tricky business that should not be done ad hoc.

I did not try it, but I think that instead of rebuilding a string and reparsing it, we should just modify the field that we wish to.
```suggestion
         vmwURL.User = url.UserPassword(accessKey, secKey)
         vmwURL.Path = "sdk"
```

pkg/importer/vddk-datasource.go
```diff
@@ -248,18 +248,24 @@ func createVMwareClient(endpoint string, accessKey string, secKey string, thumbp
...
 	conn, err := govmomi.NewClient(ctx, vmwURL, true)
 	if err != nil {
-		klog.Errorf("Unable to connect to vCenter: %v", err)
+		if secKey != "" {
+			wrap := url.PathEscape(accessKey) + ":%s@"
+			target := fmt.Sprintf(wrap, url.PathEscape(secKey))
+			redacted = fmt.Sprintf(wrap, "*****")
+			redacted = strings.ReplaceAll(err.Error(), target, redacted)
+			err = errors.New("unable to connect to vCenter: " + redacted)
+		}
```
dankenigsberg, line 268: using `vmwURL.Redacted()` should be safer.

dankenigsberg (changes requested):

pkg/importer/vddk-datasource.go
mrnold: Thank you! This is much better.

pkg/importer/vddk-datasource.go
mrnold: With the previous fix, this whole error branch is no longer necessary. The password isn't part of the bad URL string anymore, so the error from govmomi is now just "incorrect username or password". Also much better.

pkg/importer/vddk-datasource.go
```diff
@@ -247,19 +247,13 @@ func createVMwareClient(endpoint string, accessKey string, secKey string, thumbp
...
-		return nil, err
-	}
+	vmwURL.User = url.UserPassword(accessKey, secKey)
+	vmwURL.Path = "sdk"
 
 	// Log in to vCenter
 	ctx, cancel := context.WithCancel(context.Background())
 	conn, err := govmomi.NewClient(ctx, vmwURL, true)
 	if err != nil {
-		klog.Errorf("Unable to connect to vCenter: %v", err)
```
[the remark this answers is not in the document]
mrnold: I removed it because there are already a few other places where the error is logged, so it piles up in an unhelpfully verbose way without adding much new information. I think there are a few other places in the VDDK importer that should just return an error instead of logging it separately and then returning it.

I left it removed after taking out the previous redaction code, but the change does look out of place without that context. I can put it back if it's tidier that way.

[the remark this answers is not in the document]
dankenigsberg: thanks, if the log is too chatty, better have another PR to fix that.

It does not worth much, but this lgtm.

PR #1841, opened 2021-06-21
Add DataVolume annotation to retain the transfer pods after completion

arnongilboa:

Add DV annotation which if set will cause CDI transfer pods to be retained after a failed/successful completion. This will make debugging and testing easier, as developers an get the pod logs etc.
- Annotation: `cdi.kubevirt.io/storage.pod.retainAfterCompletion=true`
- Pod will be retained after both successful and failed completion.
- User is responsible to delete the pods. They will be deleted anyway when their dv is deleted.

**Release note**:
```release-note
Add optional DataVolume annotation to retain the transfer pods after completion
```

### DIFF
--- a/doc/debug.md
+++ b/doc/debug.md
@@ -1,0 +1,30 @@
+# Debugging CDI
+
+## Introduction
+
+There are various instrumentations available for developers debugging CDI. We will keep expanding this document to make our life easier.
+
+## DataVolume annotation to retain the transfer pods after completion
+
+Adding the annotation `cdi.kubevirt.io/storage.pod.retainAfterCompletion: "true"` will cause CDI transfer pods (importer, uploader, cloner) to be retained after a successful or failed completion. This makes debugging and testing easier, as developers can get the pod state and logs after completion. The pods will be deleted when their dv/pvc is deleted, otherwise the user is responsible for deleting them.
+
+For example:
+
+```yaml
+apiVersion: cdi.kubevirt.io/v1beta1
+kind: DataVolume
+metadata:
+  name: dv-pod-retain
+  annotations:
+      cdi.kubevirt.io/storage.pod.retainAfterCompletion: "true"
+spec:
+  source:
+      http:
+         url: "http://mirrors.nav.ro/fedora/linux/releases/33/Cloud/x86_64/images/Fedora-Cloud-Base-33-1.2.x86_64.qcow2"
+  pvc:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
+```
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -179,14 +179,14 @@ func (r *CloneReconciler) Reconcile(_ context.Context, req reconcile.Request) (r
 	}
 	log := r.log.WithValues("PVC", req.NamespacedName)
 	log.V(1).Info("reconciling Clone PVCs")
-	if pvc.DeletionTimestamp != nil || (!r.shouldReconcile(pvc, log) && !isPodRetainAfterCompletion(pvc)) {
+	if pvc.DeletionTimestamp != nil || !r.shouldReconcile(pvc, log) {
 		log.V(1).Info("Should not reconcile this PVC",
 			"checkPVC(AnnCloneRequest)", checkPVC(pvc, AnnCloneRequest, log),
 			"NOT has annotation(AnnCloneOf)", !metav1.HasAnnotation(pvc.ObjectMeta, AnnCloneOf),
 			"isBound", isBound(pvc, log),
 			"has finalizer?", HasFinalizer(pvc, cloneSourcePodFinalizer))
-		if HasFinalizer(pvc, cloneSourcePodFinalizer) {
-			// Clone completed, remove source pod and finalizer.
+		if HasFinalizer(pvc, cloneSourcePodFinalizer) || pvc.DeletionTimestamp != nil {
+			// Clone completed, remove source pod and/or finalizer
 			if err := r.cleanup(pvc, log); err != nil {
 				return reconcile.Result{}, err
 			}
@@ -420,10 +420,12 @@ func (r *CloneReconciler) cleanup(pvc *corev1.PersistentVolumeClaim, log logr.Lo
 			log.V(3).Info("Clone succeeded, waiting for source pod to stop running", "pod.Namespace", pod.Namespace, "pod.Name", pod.Name)
 			return nil
 		}
-
-		if err = r.client.Delete(context.TODO(), pod); err != nil {
-			if !k8serrors.IsNotFound(err) {
-				return errors.Wrap(err, "error deleting clone source pod")
+		if shouldDeletePod(pvc) {
+			log.V(3).Info("Deleting pod", "pod.Name", pod.Name)
+			if err = r.client.Delete(context.TODO(), pod); err != nil {
+				if !k8serrors.IsNotFound(err) {
+					return errors.Wrap(err, "error deleting clone source pod")
+				}
 			}
 		}
 	}
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -179,7 +179,7 @@ func (r *CloneReconciler) Reconcile(_ context.Context, req reconcile.Request) (r
 	}
 	log := r.log.WithValues("PVC", req.NamespacedName)
 	log.V(1).Info("reconciling Clone PVCs")
-	if pvc.DeletionTimestamp != nil || !r.shouldReconcile(pvc, log) {
+	if pvc.DeletionTimestamp != nil || (!r.shouldReconcile(pvc, log) && !isPodRetainAfterCompletion(pvc)) {
 		log.V(1).Info("Should not reconcile this PVC",
 			"checkPVC(AnnCloneRequest)", checkPVC(pvc, AnnCloneRequest, log),
 			"NOT has annotation(AnnCloneOf)", !metav1.HasAnnotation(pvc.ObjectMeta, AnnCloneOf),
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -192,7 +192,7 @@ func (r *ImportReconciler) shouldReconcilePVC(pvc *corev1.PersistentVolumeClaim,
 	if err != nil {
 		return false, err
 	}
-	return (!isPVCComplete(pvc) || isPodRetainAfterCompletion(pvc)) &&
+	return !isPVCComplete(pvc) &&
 			(checkPVC(pvc, AnnEndpoint, log) || checkPVC(pvc, AnnSource, log)) &&
 			shouldHandlePvc(pvc, waitForFirstConsumerEnabled, log),
 		nil
@@ -428,9 +428,10 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 	if isPVCComplete(pvc) || scratchExitCode {
 		if !scratchExitCode {
 			r.recorder.Event(pvc, corev1.EventTypeNormal, ImportSucceededPVC, "Import Successful")
-			log.V(1).Info("Completed successfully, deleting POD", "pod.Name", pod.Name)
+			log.V(1).Info("Import completed successfully")
 		}
-		if !isPodRetainAfterCompletion(pvc) {
+		if shouldDeletePod(pvc) {
+			log.V(1).Info("Deleting pod", "pod.Name", pod.Name)
 			if err := r.client.Delete(context.TODO(), pod); IgnoreNotFound(err) != nil {
 				return err
 			}
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -192,7 +192,7 @@ func (r *ImportReconciler) shouldReconcilePVC(pvc *corev1.PersistentVolumeClaim,
 	if err != nil {
 		return false, err
 	}
-	return !isPVCComplete(pvc) &&
+	return (!isPVCComplete(pvc) || isPodRetainAfterCompletion(pvc)) &&
 			(checkPVC(pvc, AnnEndpoint, log) || checkPVC(pvc, AnnSource, log)) &&
 			shouldHandlePvc(pvc, waitForFirstConsumerEnabled, log),
 		nil
@@ -430,8 +430,10 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 			r.recorder.Event(pvc, corev1.EventTypeNormal, ImportSucceededPVC, "Import Successful")
 			log.V(1).Info("Completed successfully, deleting POD", "pod.Name", pod.Name)
 		}
-		if err := r.client.Delete(context.TODO(), pod); IgnoreNotFound(err) != nil {
-			return err
+		if !isPodRetainAfterCompletion(pvc) {
+			if err := r.client.Delete(context.TODO(), pod); IgnoreNotFound(err) != nil {
+				return err
+			}
 		}
 	}
 	return nil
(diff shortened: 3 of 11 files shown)


pkg/controller/clone-controller.go
```diff
@@ -179,7 +179,7 @@ func (r *CloneReconciler) Reconcile(_ context.Context, req reconcile.Request) (r
 	}
 	log := r.log.WithValues("PVC", req.NamespacedName)
 	log.V(1).Info("reconciling Clone PVCs")
-	if pvc.DeletionTimestamp != nil || !r.shouldReconcile(pvc, log) {
+	if pvc.DeletionTimestamp != nil || (!r.shouldReconcile(pvc, log) && !isPodRetainAfterCompletion(pvc)) {
```
awels, line 182: We do want to remove the finalizer from the PVC when the clone is done, just not remove the pod if the annotation is set. So I would modify r.cleanup to not delete pod instead of skipping the entire function (which also removes the finalizer)

pkg/controller/import-controller.go
```diff
@@ -192,7 +192,7 @@ func (r *ImportReconciler) shouldReconcilePVC(pvc *corev1.PersistentVolumeClaim,
 	if err != nil {
 		return false, err
 	}
-	return !isPVCComplete(pvc) &&
+	return (!isPVCComplete(pvc) || isPodRetainAfterCompletion(pvc)) &&
```
awels, line 195: I am not quite sure why this || isPodRetainAfterCompletion is here. Am I missing something?

pkg/controller/upload-controller.go
```diff
@@ -126,7 +126,7 @@ func (r *UploadReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 	// force cleanup if PVC pending delete and pod running or the upload/clone annotation was removed
-	if !shouldReconcile || podSucceededFromPVC(pvc) || pvc.DeletionTimestamp != nil {
+	if !shouldReconcile || (podSucceededFromPVC(pvc) && !isPodRetainAfterCompletion(pvc)) || pvc.DeletionTimestamp != nil {
```
awels, line 129: Again, the cleanup does more than just delete the pod, so we should let the cleanup run, but skip the pod deletion in there if the annotation is set.

pkg/controller/util.go
```diff
@@ -762,6 +764,11 @@ func getPriorityClass(pvc *v1.PersistentVolumeClaim) string {
 	return anno[AnnPriorityClassName]
 }
 
+func isPodRetainAfterCompletion(pvc *corev1.PersistentVolumeClaim) bool {
+	retain, exists := pvc.ObjectMeta.Annotations[AnnPodRetainAfterCompletion]
```
awels, line 768: Are we 100% sure pvc.ObjectMeta.Annotations is not nil?

awels (commented):
I am missing some documentation on exactly what the new annotation does, and why we would want it.

pkg/controller/clone-controller.go
arnongilboa: You mean when user manually deletes the pod but the PVC still exists, it will still have the finalizer, which is risky/undesirable?

pkg/controller/import-controller.go
arnongilboa: Otherwise, when we manually delete the pvc/dv, it will prevent Reconcile() to call reconcilePvc() for the importer pod deletion.

pkg/controller/upload-controller.go
arnongilboa: Sure.

pkg/controller/clone-controller.go
awels: Its undesirable because now for some reason unknown to the user, they cannot delete the PVC anymore (due to the finalizer)

pkg/controller/import-controller.go
awels: But isn't the pod owned by the PVC, so if the PVC is deleted, k8s (not our controller) will delete the pod.

pkg/controller/util.go
arnongilboa: Afaik, nowhere in our code we check Annotations for nil, so I assumed it's sure initialized, but I'll check that if needed.
Anyway a nicer line will sure be using `pvc.GetAnnotations()` and not `ObjectMeta`.

pkg/controller/clone-controller.go
arnongilboa: Only if we already deleted the pod, right?

pkg/controller/import-controller.go
arnongilboa: That's not the behavior I've seen, but I'll verify it.

awels:
Might want to also add a release note, this is something more people will care about

pkg/controller/import-controller.go
arnongilboa: Verified. Pod is owned by PVC, but when PVC is deleted, Pod is deleted by our controller and NOT by k8s.
If it smells like a bug, we should fix it anyway in another card.

pkg/controller/clone-controller.go
mhenriks: finalizer on PVC should be removed regardless of this annotation

pkg/controller/import-controller.go
mhenriks: Was pod phase `Succeeded` in your test?  If not, I can understand k8s not terminating a running pod

pkg/controller/util.go
mhenriks: it is okay to get values from a nil map.  Just have to check nil when setting

pkg/controller/upload-controller.go
arnongilboa: so we should still deleteService() even if we keep the pod, right?

pkg/controller/import-controller.go
arnongilboa: You can see that if you comment out the explicit importer pod deletion for the case pvc was deleted (pvc.DeletionTimestamp != nil), the pod won't be deleted.

arnongilboa: Right, fixing it to `|| (isPodRetainAfterCompletion(pvc) && pvc.DeletionTimestamp != nil)`, which is better I guess?

pkg/controller/clone-controller.go
arnongilboa: done

pkg/controller/clone-controller.go
```diff
@@ -185,11 +185,9 @@ func (r *CloneReconciler) Reconcile(_ context.Context, req reconcile.Request) (r
 			"NOT has annotation(AnnCloneOf)", !metav1.HasAnnotation(pvc.ObjectMeta, AnnCloneOf),
 			"isBound", isBound(pvc, log),
 			"has finalizer?", HasFinalizer(pvc, cloneSourcePodFinalizer))
-		if HasFinalizer(pvc, cloneSourcePodFinalizer) {
-			// Clone completed, remove source pod and finalizer.
-			if err := r.cleanup(pvc, log); err != nil {
-				return reconcile.Result{}, err
-			}
+		// Clone completed, remove source pod and/or finalizer
+		if err := r.cleanup(pvc, log); err != nil {
```
mhenriks, line 189: now cleanup will be called very frequently, basically whenever the status of any PVC changes.  And I believe the list call is pretty expensive (not in memory) so I think we should find a way to minimize the frequency of cleanup calls.

tests/utils/pod.go
```diff
@@ -70,45 +70,79 @@ func DeletePod(clientSet *kubernetes.Clientset, pod *k8sv1.Pod, namespace string
...
+		foundPod, err = findPodByCompFuncOnce(clientSet, namespace, prefix, labelSelector, compFunc, &result)
+		return foundPod, err
 	})
 	if !foundPod {
 		return nil, fmt.Errorf("Unable to find pod containing %s", prefix)
 	}
 	return &result, err
 }
 
+func findPodByCompFuncOnce(clientSet *kubernetes.Clientset, namespace, prefix, labelSelector string, compFunc func(string, string) bool, result *k8sv1.Pod) (bool, error) {
```
awels, line 123: Why not have this function return the pod if found, then you can just call it and return it directly from the FindPodBy(Pre|Suf)fix functions. And tine findPodByCompFunc you just check if the pod == nil or not to get the bool value.

pkg/controller/clone-controller.go
```diff
@@ -415,12 +413,13 @@ func (r *CloneReconciler) cleanup(pvc *corev1.PersistentVolumeClaim, log logr.Lo
...
 	}
 
-	if pod != nil && pod.DeletionTimestamp == nil {
+	if pod != nil && pod.DeletionTimestamp == nil && (pvc.DeletionTimestamp != nil || !isPodRetainAfterCompletion(pvc)) {
 		if podSucceededFromPVC(pvc) && pod.Status.Phase == corev1.PodRunning {
 			log.V(3).Info("Clone succeeded, waiting for source pod to stop running", "pod.Namespace", pod.Namespace, "pod.Name", pod.Name)
 			return nil
 		}
 
+		log.V(3).Info("Deleting pod", "pod.Name", pod.Name)
```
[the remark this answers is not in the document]
arnongilboa: Ok, but I guess we must have `pvc.DeletionTimestamp != nil` in the condition as well, otherwise we cannot delete the pod.

pkg/controller/upload-controller.go
```diff
@@ -334,7 +334,7 @@ func (r *UploadReconciler) cleanup(pvc *v1.PersistentVolumeClaim) error {
 		}
 		return err
 	}
-	if pod.DeletionTimestamp == nil {
+	if pod.DeletionTimestamp == nil && (pvc.DeletionTimestamp != nil || !isPodRetainAfterCompletion(pvc)) {
```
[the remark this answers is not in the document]
arnongilboa: No, I guess you must check `pvc.DeletionTimestamp != nil` before you delete the pod.
And yep, `shouldDeletePod()` is better.

pkg/controller/util.go
```diff
@@ -762,6 +764,10 @@ func getPriorityClass(pvc *v1.PersistentVolumeClaim) string {
 	return anno[AnnPriorityClassName]
 }
 
+func isPodRetainAfterCompletion(pvc *corev1.PersistentVolumeClaim) bool {
```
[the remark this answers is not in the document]
arnongilboa: Sure!

tests/cloner_test.go
```diff
@@ -108,6 +108,42 @@ var _ = Describe("all clone tests", func() {
 			completeClone(f, f.Namespace, targetPvc, diskImagePath, sourceMD5, sourcePvcDiskGroup)
 		})
 
+		It("[test_id:6693]Should clone imported data and retain transfer pods after completion", func() {
+			dataVolume := utils.NewDataVolumeWithHTTPImport(dataVolumeName, "1Gi", fmt.Sprintf(utils.TinyCoreIsoURL, f.CdiInstallNs))
+			dataVolume.Annotations[controller.AnnPodRetainAfterCompletion] = "true"
```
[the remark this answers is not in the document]
arnongilboa: Right, no big deal and we already teted it in import_test.

tests/utils/pod.go
arnongilboa: Another great catch, Alexander!

arnongilboa: Another great catch, Alexander!

arnongilboa:
Unfortunatelly added an infra bug in the last `push`. Fixing it.

doc/debug.md
```diff
@@ -0,0 +1,30 @@
+# Debugging CDI
+
+## Introduction
+
+There are various instrumentations available for developers debugging CDI. We will keep expanding this document to make our life easier.
+
+## DataVolume annotation to retain the transfer pods after completion
+
+Adding the annotation `cdi.kubevirt.io/storage.pod.retainAfterCompletion: "true"` will cause CDI transfer pods (importer, uploader, cloner) to be retained after a successful or failed completion. This makes debugging and testing easier, as developers can get the pod state and logs after completion. The pods will be deleted when their dv/pvc is deleted, otherwsie the user is responsible to delete them.
```
awels, line 9: Typo 
```
otherwsie the user is responsible to delete them.
```
Maybe
```
otherwise the user is responsible for deleting them.
```

tests/import_test.go
```diff
@@ -139,6 +139,28 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+		f.ForceBindIfWaitForFirstConsumer(pvc)
+
+		By("Wait for import to be completed")
+		err = utils.WaitForDataVolumePhase(f.CdiClient, dataVolume.Namespace, cdiv1.Succeeded, dataVolume.Name)
+		Expect(err).ToNot(HaveOccurred(), "Datavolume not in phase succeeded in time")
+
+		By("Find importer pod after completion")
+		importer, err := utils.FindPodByPrefixOnce(f.K8sClient, dataVolume.Namespace, common.ImporterPodName, common.CDILabelSelector)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(importer.DeletionTimestamp).To(Equal((*metav1.Time)(nil)))
```
awels, line 162: You can write this as
```go
```

tests/cloner_test.go
```diff
@@ -108,6 +108,48 @@ var _ = Describe("all clone tests", func() {
...
+
+			By("Find cloner source pod after completion")
+			cloner, err := utils.FindPodBySuffixOnce(f.K8sClient, targetDataVolume.Namespace, common.ClonerSourcePodNameSuffix, common.CDILabelSelector)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cloner.DeletionTimestamp).To(Equal((*metav1.Time)(nil)))
+
+			By("Find upload pod after completion")
+			uploader, err := utils.FindPodByPrefixOnce(f.K8sClient, targetDataVolume.Namespace, "cdi-upload-", common.CDILabelSelector)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(uploader.DeletionTimestamp).To(Equal((*metav1.Time)(nil)))
```
awels, line 150: You can write this as
```go
```

tests/cloner_test.go
```diff
@@ -108,6 +108,48 @@ var _ = Describe("all clone tests", func() {
...
+			By("Wait for clone to be completed")
+			_, err = utils.WaitForPVC(f.K8sClient, targetDataVolume.Namespace, targetDataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+			By("Wait for target datavolume phase Succeeded")
+			utils.WaitForDataVolumePhaseWithTimeout(f.CdiClient, targetDataVolume.Namespace, cdiv1.Succeeded, targetDV.Name, cloneCompleteTimeout)
+
+			By("Find cloner source pod after completion")
+			cloner, err := utils.FindPodBySuffixOnce(f.K8sClient, targetDataVolume.Namespace, common.ClonerSourcePodNameSuffix, common.CDILabelSelector)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cloner.DeletionTimestamp).To(Equal((*metav1.Time)(nil)))
```
awels, line 145: You can write this as
```go
```

awels (commented):
Looks pretty good, I am trying to understand why shouldDeletePod is true when the pvc.DeletionTimestamp != nil.

The way I understand the process is:
1. User deletes PVC (or DV and DV controller deletes PVC)
2. k8s sees the PVC controlling the POD, and thus marks the pod for deletion (pod.deletionTimeStamp is !nil)
3. The controller hits HasFinalizer (because the pvc.DeletionTimestap != nil) and removes the finalizer from the pod
4. Now the finalizer is gone, and the pod has a deletionTimestamp, k8s can now clean up the POD.

At least I think that is what is supposed to happen, if that is not the case we should probably investigate why not. Other than that it looks good with some minor comments.

PR #1844, opened 2021-06-24
Fix smartclone sometimes not triggering due to capicity not being reported.

awels:
Updated tests to use a real image instead of data that is filled.

Fixes #

**Release note**:

```release-note
BugFix: Smart clone was not always used if the provisioner was slow to allocate the PV.
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -1209,9 +1209,8 @@ func (r *DatavolumeReconciler) snapshotSmartClonePossible(dataVolume *cdiv1.Data
 	}
 
 	srcStorageClass := &storagev1.StorageClass{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *targetPvcStorageClassName}, srcStorageClass); err != nil {
-		log.Info("Unable to retrieve storage class, falling back to host assisted clone", "storage class", *targetPvcStorageClassName)
-		return false, nil
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *targetPvcStorageClassName}, srcStorageClass); IgnoreNotFound(err) != nil {
+		return false, err
 	}
 	srcCapacity, hasSrcCapacity := pvc.Status.Capacity[corev1.ResourceStorage]
 	targetRequest, hasTargetRequest := targetStorageSpec.Resources.Requests[corev1.ResourceStorage]
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -384,6 +384,9 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 	}
 
 	snapshotClassName, err := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 	doSmartClone := err == nil &&
 		cloneStrategy == cdiv1.CloneStrategySnapshot &&
 		(!isCrossNamespaceClone(datavolume) || *bindingMode == storagev1.VolumeBindingImmediate)
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -387,7 +387,11 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	doSmartClone := err == nil &&
+	snapshotPossible, err := r.snapshotSmartClonePossible(datavolume, pvcSpec)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	doSmartClone := snapshotClassName != "" && snapshotPossible &&
 		cloneStrategy == cdiv1.CloneStrategySnapshot &&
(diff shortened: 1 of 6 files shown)


tests/smartclone_test.go
```diff
@@ -175,15 +164,13 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
...
-	hash := md5.Sum([]byte(expectedData))
-	md5sum := hex.EncodeToString(hash[:])
+func verifyPVC(dataVolume *cdiv1.DataVolume, f *framework.Framework, testPath string, md5sum string) {
 	By("verifying pvc was created")
 	targetPvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(dataVolume.Namespace).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
 	Expect(err).ToNot(HaveOccurred())
 
 	By(fmt.Sprint("Verifying target PVC content"))
-	Expect(f.VerifyTargetPVCContentMD5(f.Namespace, targetPvc, testPath, md5sum, int64(len(expectedData)))).To(BeTrue())
+	Expect(f.VerifyTargetPVCContentMD5(f.Namespace, targetPvc, testPath, md5sum, 18874368)).To(BeTrue())
```
maya-r, line 173: utils.UploadFileSize?

awels: Fixed, I don't like the name of the constant, but we should use it regardless in case we ever rename it.

brybacki:
One smart clone test still fails, is it flake? 

`Tests Suite: [vendor:cnv-qe@redhat.com][level:component]SmartClone tests [rfe_id:1106][test_id:3495][crit:high][vendor:cnv-qe@redhat.com][level:component] Verify DataVolume Smart Cloning - volumeMode block - Positive flow`

/retest

pkg/controller/datavolume-controller.go
```diff
@@ -1122,73 +1171,56 @@ func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.D
 	pvc := &corev1.PersistentVolumeClaim{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: sourcePvcNs, Name: sourcePvcSpec.Name}, pvc); err != nil {
 		if k8serrors.IsNotFound(err) {
-			r.log.V(3).Info("Source PVC is missing", "source namespace", sourcePvcSpec.Namespace, "source name", sourcePvcSpec.Name)
+			log.Info("Source PVC is missing", "source namespace", sourcePvcSpec.Namespace, "source name", sourcePvcSpec.Name)
+			return false, nil
```
brybacki, line 1175: why err is now nil?
previously it stated source PVC not found, and probably the whole operation was retried

pkg/controller/datavolume-controller.go
```diff
@@ -1122,73 +1171,56 @@ func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.D
...
 	}
 
-	// Fetch the source storage class
 	srcStorageClass := &storagev1.StorageClass{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *sourcePvcStorageClassName}, srcStorageClass); err != nil {
-		r.log.V(3).Info("Unable to retrieve storage class, falling back to host assisted clone", "storage class", *sourcePvcStorageClassName)
-		return "", errors.New("unable to retrieve storage class, falling back to host assisted clone")
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *targetPvcStorageClassName}, srcStorageClass); err != nil {
+		log.Info("Unable to retrieve storage class, falling back to host assisted clone", "storage class", *targetPvcStorageClassName)
+		return false, nil
```
brybacki, line 1214: The same question here, shouldn't it return en err, so the operation can be retried again?

brybacki:
I like this change, it decouples 

- the check if cluster is SmartClone capable - probably this logic might be even more explicit, it is now hidden in snapshotClassName, but lets not add things to this change, my change for CSI Volume clone will follow shortly:
```
snapshotClassName, err := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
snapshotClassAvailable := err == nil && snapshotClassName != ""
```

- From the check if we actually can do the SmartClone on current DataVolume -> `snapshotSmartClonePossible`. The `snapshotSmartClonePossible` is almost exactly the same as my new `isCSIClonePossible`, so we might see some more refactoring after this is merged, and I rebase. 

Looks like I need to rebase.

pkg/controller/datavolume-controller.go
awels: So before any error would just have us switch to host assisted clone and not retry at all. When I refactored I tried to decide which failures are 'recoverable' aka a retry might fix the problem, and those returned an error. And which failures are permanent (like no CRDs existing) which should switch to host assisted clone.

In this case if err == nil -> host assisted clone
if err != nil -> retry reconcile.

I think I agree a missing source is a retry not a switch to host assisted clone. But I suspect further down it will generate an error because we can't clone

pkg/controller/datavolume-controller.go
awels: Good point, I think I agree.

pkg/controller/datavolume-controller.go
awels: So I did some digging here and there is actual code that depends on us not returning an error here, and it will error later. So I am leaving this one nil.

PR #1846, opened 2021-06-29
Preferred Clone strategy selection

brybacki:
Adds a new way to select preferred CloneStrategy by extending the StorageProfile with `CloneStrategy` field. 

Fixes #

First commit is the cleanup. Some methods where extracted to make main Reconcile loop easier to read. 
 
**Release note**:

```release-note
Add CloneStrategy selection to StorageProfile. 
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -397,107 +397,11 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 
 	if !pvcExists {
 		if doSmartClone {
-			pvcName := datavolume.Name
-			if isCrossNamespaceClone(datavolume) {
-				pvcName = transferName
-				initialized, err := r.initTransfer(log, datavolume, pvcName)
-				if err != nil {
-					return reconcile.Result{}, err
-				}
-
-				// get reconciled again v soon
-				if !initialized {
-					return reconcile.Result{},
-						r.updateSmartCloneStatusPhase(cdiv1.CloneScheduled, datavolume, nil)
-				}
-
-				tmpPVC := &corev1.PersistentVolumeClaim{}
-				nn := types.NamespacedName{Namespace: datavolume.Spec.Source.PVC.Namespace, Name: pvcName}
-				if err := r.client.Get(context.TODO(), nn, tmpPVC); err != nil {
-					if !k8serrors.IsNotFound(err) {
-						return reconcile.Result{}, err
-					}
-				} else if tmpPVC.Annotations[AnnCloneOf] == "true" {
-					done, err := r.expand(log, datavolume, tmpPVC, pvcSpec)
-					if err != nil {
-						return reconcile.Result{}, err
-					}
-					if !done {
-						return reconcile.Result{},
-							r.updateSmartCloneStatusPhase(cdiv1.ExpansionInProgress, datavolume, nil)
-					}
-
-					// trigger transfer and next reconcile should have pvcExists == true
-					tmpPVC.Annotations[annReadyForTransfer] = "true"
-					tmpPVC.Annotations[AnnPopulatedFor] = datavolume.Name
-					if err := r.client.Update(context.TODO(), tmpPVC); err != nil {
-						return reconcile.Result{}, err
-					}
-
-					return reconcile.Result{},
-						r.updateSmartCloneStatusPhase(cdiv1.NamespaceTransferInProgress, datavolume, nil)
-				} else {
-					return reconcile.Result{}, nil
-				}
-			}
-
-			if datavolume.Status.Phase == cdiv1.NamespaceTransferInProgress {
-				return reconcile.Result{}, nil
-			}
-
-			r.log.V(3).Info("Smart-Clone via Snapshot is available with Volume Snapshot Class",
-				"snapshotClassName", snapshotClassName)
-
-			newSnapshot := newSnapshot(datavolume, pvcName, snapshotClassName)
-			if err := setAnnOwnedByDataVolume(newSnapshot, datavolume); err != nil {
-				return reconcile.Result{}, err
-			}
-
-			nn := client.ObjectKeyFromObject(newSnapshot)
-			if err := r.client.Get(context.TODO(), nn, newSnapshot.DeepCopy()); err != nil {
-				if !k8serrors.IsNotFound(err) {
-					return reconcile.Result{}, err
-				}
-
-				inUse, err := r.sourceInUse(datavolume)
-				if err != nil {
-					return reconcile.Result{}, err
-				}
-				populated, err := r.isSourcePVCPopulated(datavolume)
-				if err != nil {
-					return reconcile.Result{}, err
-				}
-				if inUse || !populated {
-					return reconcile.Result{Requeue: true},
-						r.updateSmartCloneStatusPhase(cdiv1.CloneScheduled, datavolume, nil)
-				}
-
-				if err := r.client.Create(context.TODO(), newSnapshot); err != nil {
-					if k8serrors.IsAlreadyExists(err) {
-						return reconcile.Result{}, nil
-					}
-
-					return reconcile.Result{}, err
-				}
-
-				return reconcile.Result{},
-					r.updateSmartCloneStatusPhase(cdiv1.SnapshotForSmartCloneInProgress, datavolume, nil)
-			}
-
-			return reconcile.Result{}, nil
+			return r.reconcileSmartClonePvc(log, datavolume, pvcSpec, transferName, snapshotClassName)
 		}
-		log.Info("Creating PVC for datavolume")
-		newPvc, err := r.newPersistentVolumeClaim(datavolume, pvcSpec)
+
+		newPvc, err := r.createPvcForDatavolume(log, datavolume, pvcSpec)
 		if err != nil {
-			return reconcile.Result{}, err
-		}
-		checkpoint := r.getNextCheckpoint(datavolume, newPvc)
-		if checkpoint != nil { // Initialize new warm import annotations before creating PVC
-			newPvc.ObjectMeta.Annotations[AnnCurrentCheckpoint] = checkpoint.Current
-			newPvc.ObjectMeta.Annotations[AnnPreviousCheckpoint] = checkpoint.Previous
-			newPvc.ObjectMeta.Annotations[AnnFinalCheckpoint] = strconv.FormatBool(checkpoint.IsFinal)
-		}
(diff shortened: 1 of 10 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -1222,22 +1377,76 @@ func (r *DatavolumeReconciler) snapshotSmartClonePossible(dataVolume *cdiv1.Data
 	return true, nil
 }
 
-func (r *DatavolumeReconciler) getCloneStrategy() (cdiv1.CDICloneStrategy, error) {
+func (r *DatavolumeReconciler) getCloneStrategy(dataVolume *cdiv1.DataVolume) (*cdiv1.CDICloneStrategy, error) {
+	defaultCloneStrategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategySnapshot)
+	sourcePvc, err := r.findSourcePvc(dataVolume)
+	if err != nil {
+		// when k8serrors.IsNotFound(err) -> Source PVC is missing, cannot correctly select strategy
```
brybacki, line 1384: this is a place where new logic differs from the old one, 
- previously when the `source` pvc was not available clone strategy was assumed as HostAssisted and the reconcile loop logic moved forward and reconciled the `target` pvc 
- with new logic it returns err, I wonder if some kind of event might be helpful here?

awels: So the err will result in a new reconcile and we hope that in that time the source appeared, also if there is no source, host assisted clone will also not work, since there is no source. I think an event is probably a good idea, something like source not found unable to clone or something like that.

pkg/controller/datavolume-controller.go
```diff
@@ -373,194 +383,319 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+			if err != nil {
+				return 0, err
 			}
 
-			nn := client.ObjectKeyFromObject(newSnapshot)
-			if err := r.client.Get(context.TODO(), nn, newSnapshot.DeepCopy()); err != nil {
-				if !k8serrors.IsNotFound(err) {
-					return reconcile.Result{}, err
-				}
+			doSmartClone = snapshotClassAvailable && snapshotPossible &&
```
awels, line 547:
```go
if snapshotClassAvailable && snapshotPossible &&
  (!isCrossNamespaceClone(datavolume) || *bindingMode == storagev1.VolumeBindingImmediate) {
  return SmartClone, nil
}
This also means we can't do cross namespace clone on WFFC storage? Why does this not also apply to CSI clone?

File: pkg/controller/datavolume-controller.go [hunk 3]
```diff
@@ -373,194 +383,319 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+			r.log.V(3).Info("CSI-Clone is available")
+		}
+
+		newPvc, err := r.createPvcForDatavolume(log, datavolume, pvcSpec)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		pvc = newPvc
+	} else {
+		//TODO: maybe... when pvc exists, get clone strategy from pvc?
```
awels, line 407: If the PVC exists, we can only do a host assisted clone right, since both CSI and smart clone require the PVC to not exist.

pkg/controller/datavolume-controller.go
```diff
@@ -373,194 +383,319 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+			r.updateSmartCloneStatusPhase(cdiv1.SnapshotForSmartCloneInProgress, datavolume, nil)
 	}
 
-	// Finally, we update the status block of the DataVolume resource to reflect the
-	// current state of the world
-	return r.reconcileDataVolumeStatus(datavolume, pvc)
+	return reconcile.Result{}, nil
+}
+
+func (r *DatavolumeReconciler) reconcileSmartCloneForExistingPvc(log logr.Logger, datavolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim, pvcSpec *corev1.PersistentVolumeClaimSpec, transferName string) (reconcile.Result, error) {
```
awels, line 673: Isn't this whole function just post clone processing (transfer and/or resize) for either CSI clone or smart clone. I think the function name should reflect its purpose. It is not restricted to smart clone. And some of the updateStatus functions should probably also be renamed to reflect this.

awels:
One more thing, need to update documentation for the StorageProfile to include preferred clone strategy now

mhenriks:
So far so good!  Don't have much to add to @awels comments other than to reiterate that there should be a lot of similarity between smart clone and csi clone flows so may want to rethink naming/parameterizing functions sooner rather than later.  But no big deal if you wait

brybacki:
Thanks for good comments, I am on it.

@mhenriks yes I agree, and you will see even more similarity in my next PR, and then when we prove everything is correct, I can (or maybe anyone else who wants to know more about clone) do a followup with refactoring common parts.

pkg/controller/datavolume-controller.go
brybacki: As of this PR the updateStatus you mentioned is only used by smartClone, but I'll for sure rename and reuse in the CSI-clone PR. 

The same is true for `reconcileSmartCloneForExistingPvc`, and I agree it needs a better name.

pkg/controller/datavolume-controller.go
brybacki: Yes, and I forgot what was my original Idea behind this comment

pkg/controller/datavolume-controller.go
brybacki: Yes this logic is unchanged, just extracted to a method `selectCloneStrategy`

And I think you are right, CSI clone has the same. Because with WFFC you would need to schedule a pod to bind a volume, and so far we do not have logic for that. I do not know if we can correctly do it. 

On the other hand I am not sure why exactly this condition
`(!isCrossNamespaceClone(datavolume) || *bindingMode == storagev1.VolumeBindingImmediate) {` 

@mhenriks can you comment on that?

pkg/controller/datavolume-controller.go
```diff
@@ -373,194 +383,319 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+
+		if isClone && *preferredCloneStrategy == cdiv1.CloneStrategyCsiClone {
+			possible, err := r.isCSIClonePossible(datavolume, pvcSpec)
+			if err != nil {
+				return 0, err
 			}
 
-			if datavolume.Status.Phase == cdiv1.NamespaceTransferInProgress {
-				return reconcile.Result{}, nil
+			doCsiClone = possible
```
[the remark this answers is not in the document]
brybacki: Yes, and also new condition, the same as for SmartClone

pkg/controller/datavolume-controller.go
```diff
@@ -373,194 +383,319 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+		} else if isClone && *preferredCloneStrategy == cdiv1.CloneStrategySnapshot {
```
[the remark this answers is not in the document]
brybacki: good catch, must be a leftover after I did it in a few different ways ;)

pkg/controller/datavolume-controller.go
```diff
@@ -373,194 +383,319 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+		isClone = true
+
+		// The source and target PVCs share the same Storage Class
+		// get this the same way as for getSnapshotClassForSmartClone(datavolume, pvcSpec)
+		preferredCloneStrategy, err := r.getCloneStrategy(datavolume)
+		if err != nil {
+			return 0, err
+		}
+
+		if isClone && *preferredCloneStrategy == cdiv1.CloneStrategyCsiClone {
```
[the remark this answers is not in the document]
brybacki: yes

pkg/controller/datavolume-controller.go
brybacki: Yes, I think it is only good  when `k8serrors.IsNotFound(err)` other errors are just an intermittent communication problems or the whole cluster is bad.

brybacki:
documentation will follow

pkg/controller/datavolume-controller.go
awels: Okay, I will hold you to that in the other PR then.

pkg/controller/datavolume-controller.go
```diff
@@ -1222,22 +1371,78 @@ func (r *DatavolumeReconciler) snapshotSmartClonePossible(dataVolume *cdiv1.Data
...
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: sourcePvcNs, Name: sourcePvcSpec.Name}, pvc); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.V(3).Info("Source PVC is missing", "source namespace", sourcePvcSpec.Namespace, "source name", sourcePvcSpec.Name)
+		}
+		return nil, err
+	}
+	return pvc, nil
+}
+
+func getGlobalCloneStrategyOverride(r *DatavolumeReconciler) (*cdiv1.CDICloneStrategy, error) {
```
awels, line 1430: If the only value you are passing into this function is the reconciler, why not make it:
```go
func (r *DatavolumeReconciler) getGlobalCloneStrategyOverride() (*cdiv1.CDICloneStrategy, error) {
...
}
```
Seems more logical to me.

pkg/controller/datavolume-controller.go
```diff
@@ -1965,6 +2170,25 @@ func getName(storageClass *storagev1.StorageClass) string {
 	return ""
 }
 
+func getPreferredCloneStrategyForStorageClass(c client.Client, storageClass *storagev1.StorageClass) (*cdiv1.CDICloneStrategy, error) {
```
awels, line 2173: Same here make it part of the reconciler, then you don't have to pass the client to it.

awels (commented):
looks good overal, some small comments.

pkg/controller/datavolume-controller.go
brybacki: Done

pkg/controller/datavolume-controller.go
brybacki: done

brybacki:
@awels I did address the last two comments

awels:
hmm both failures are on TLS failures in the upload server maybe bad timing on a cert rotation?

PR #1849, opened 2021-07-01
Fix annotation tests to keep pods.

awels:

We had to do a bunch of retries and other things to try and catch completed
pods which was especially hard with a clone of a small disk. Since we can
now retrain completed pods, the tests were updated to do that.

Fixes #

**Release note**:

```release-note
NONE
```


brybacki:
looks simpler ;)

PR #1851, opened 2021-07-05
Add DataSource as an optional sourceRef for DataVolumes

arnongilboa:

A `DataSource` is an abstraction that points to the place we'd like to import/clone from. A `DataVolume` can either have an inlined `dataVolume.Spec.Source` as it exists today, or reference a `DataSource` object via `dataVolume.Spec.SourceRef` to perform the same action. The latter allows us to decouple the inlined source from a DataVolume so we can have a static link that points to a DataSource which is mutable.

For more details, view the [design document](https://github.com/kubevirt/community/blob/master/design-proposals/golden-image-delivery-and-update-pipeline.md).

- currently only DataSource sourceRef kind is supported
- currently only PVC DataSource source is supported

**Release note**:
```release-note
Add DataSource as an optional source reference for DataVolumes
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-mutate.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate.go
@@ -31,13 +31,15 @@ import (
 	"k8s.io/klog/v2"
 
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
+	cdiclient "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"kubevirt.io/containerized-data-importer/pkg/clone"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
 	"kubevirt.io/containerized-data-importer/pkg/token"
 )
 
 type dataVolumeMutatingWebhook struct {
-	client         kubernetes.Interface
+	k8sClient      kubernetes.Interface
+	cdiClient      cdiclient.Interface
 	tokenGenerator token.Generator
 	proxy          clone.SubjectAccessReviewsProxy
 }
@@ -60,6 +62,7 @@ func (p *sarProxy) Create(sar *authv1.SubjectAccessReview) (*authv1.SubjectAcces
 
 func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
 	var dataVolume, oldDataVolume cdiv1.DataVolume
+	var pvcSource *cdiv1.DataVolumeSourcePVC
 
 	klog.V(3).Infof("Got AdmissionReview %+v", ar)
 
@@ -71,7 +74,20 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		return toAdmissionResponseError(err)
 	}
 
-	pvcSource := dataVolume.Spec.Source.PVC
+	if dataVolume.Spec.Source != nil {
+		pvcSource = dataVolume.Spec.Source.PVC
+	} else if dataVolume.Spec.SourceRef != nil && dataVolume.Spec.SourceRef.Kind == cdiv1.DataVolumeDataSource {
+		ns := dataVolume.Namespace
+		if dataVolume.Spec.SourceRef.Namespace != nil && *dataVolume.Spec.SourceRef.Namespace != "" {
+			ns = *dataVolume.Spec.SourceRef.Namespace
+		}
+		dataSource, err := wh.cdiClient.CdiV1beta1().DataSources(ns).Get(context.TODO(), dataVolume.Spec.SourceRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return toAdmissionResponseError(err)
+		}
+		pvcSource = dataSource.Spec.Source.PVC
+	}
+
 	targetNamespace, targetName := dataVolume.Namespace, dataVolume.Name
 	if targetNamespace == "" {
 		targetNamespace = ar.Request.Namespace
--- a/pkg/apiserver/webhooks/datavolume-mutate_test.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate_test.go
@@ -76,6 +76,71 @@ var _ = Describe("Mutating DataVolume Webhook", func() {
 			Expect(resp.Patch).To(BeNil())
 		})
 
+		It("should reject a DataVolume with sourceRef to non-existing DataSource", func() {
+			dataVolume := newDataSourceDataVolume("testDV", nil, "test")
+			Expect(dataVolume.Annotations).To(BeNil())
+			dvBytes, _ := json.Marshal(&dataVolume)
+			ar := &admissionv1.AdmissionReview{
+				Request: &admissionv1.AdmissionRequest{
+					Resource: metav1.GroupVersionResource{
+						Group:    cdicorev1.SchemeGroupVersion.Group,
+						Version:  cdicorev1.SchemeGroupVersion.Version,
+						Resource: "datavolumes",
+					},
+					Object: runtime.RawExtension{
+						Raw: dvBytes,
+					},
+				},
+			}
+			resp := mutateDVs(key, ar, true)
+			Expect(resp.Allowed).To(BeFalse())
+			Expect(resp.Patch).To(BeNil())
+		})
+
+		It("should allow a DataVolume with sourceRef to existing DataSource", func() {
+			dataVolume := newDataSourceDataVolume("testDV", nil, "test")
+			Expect(dataVolume.Annotations).To(BeNil())
+			dvBytes, _ := json.Marshal(&dataVolume)
+
+			ar := &admissionv1.AdmissionReview{
+				Request: &admissionv1.AdmissionRequest{
+					Resource: metav1.GroupVersionResource{
+						Group:    cdicorev1.SchemeGroupVersion.Group,
+						Version:  cdicorev1.SchemeGroupVersion.Version,
+						Resource: "datavolumes",
+					},
+					Object: runtime.RawExtension{
+						Raw: dvBytes,
+					},
+				},
+			}
+
+			dataSource := &cdicorev1.DataSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      dataVolume.Spec.SourceRef.Name,
+					Namespace: "default",
+				},
+				Spec: cdicorev1.DataSourceSpec{
+					Source: cdicorev1.DataSourceSource{
+						PVC: &cdicorev1.DataVolumeSourcePVC{
+							Name: "testPVC",
+						},
+					},
+				},
+			}
+
+			resp := mutateDVsEx(key, ar, true, []runtime.Object{dataSource})
+			Expect(resp.Allowed).To(BeTrue())
+			Expect(resp.Patch).ToNot(BeNil())
+
+			var patchObjs []jsonpatch.Operation
+			err := json.Unmarshal(resp.Patch, &patchObjs)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(patchObjs).Should(HaveLen(1))
+			Expect(patchObjs[0].Operation).Should(Equal("add"))
+			Expect(patchObjs[0].Path).Should(Equal("/metadata/annotations"))
+		})
+
 		It("should allow a DataVolume update with token unchanged", func() {
 			dataVolume := newPVCDataVolume("testDV", "testNamespace", "test")
 			Expect(dataVolume.Annotations).To(BeNil())
@@ -167,6 +232,10 @@ var _ = Describe("Mutating DataVolume Webhook", func() {
 })
 
 func mutateDVs(key *rsa.PrivateKey, ar *admissionv1.AdmissionReview, isAuthorized bool) *admissionv1.AdmissionResponse {
+	return mutateDVsEx(key, ar, isAuthorized, nil)
+}
+
+func mutateDVsEx(key *rsa.PrivateKey, ar *admissionv1.AdmissionReview, isAuthorized bool, cdiObjects []runtime.Object) *admissionv1.AdmissionResponse {
 	client := fakeclient.NewSimpleClientset()
 	client.PrependReactor("create", "subjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
 		if action.GetResource().Resource != "subjectaccessreviews" {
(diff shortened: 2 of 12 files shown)


mhenriks (commented):
I wonder if we can simplify the changes to the DataVolume controller.  Currently, there are calls to `getDataVolumeSourcePVC` sprinkled all over the place.  I wonder if we can simply do the following:

1.  The webhook enforces that all DataVolumes have `spec.source` or `spec.sourceRef` set.  Not both.
2. When DataVolume controller reconciles a DV, it checks if `sourceRef` is set.  If so, it gets the sourceRef and populates `spec.Source`.  All other code stays the same except:
3. When the DataVolume controller updates a DV, have to make sure to nil out `spec.source` when `spec.sourceRef` is set.

I also wonder if we want to support DataSources that do not exist.

arnongilboa:
> I wonder if we can simplify the changes to the DataVolume controller. Currently, there are calls to `getDataVolumeSourcePVC` sprinkled all over the place. I wonder if we can simply do the following:
> 
> 1. The webhook enforces that all DataVolumes have `spec.source` or `spec.sourceRef` set.  Not both.

as currently enforced.

> 2. When DataVolume controller reconciles a DV, it checks if `sourceRef` is set.  If so, it gets the sourceRef and populates `spec.Source` with data from the DataSource.  All other code stays the same except:
> 3. When the DataVolume controller updates a DV, have to make sure to nil out `spec.source` when `spec.sourceRef` is set.
> 
Cool, much smarter and cleaner. although the it makes the solution a bit bit hacky and will require a comment or two explaining the trick.

> I also wonder if we want to support DataSources that do not exist.
We currently don't support sources (e.g PVC) that doesn't exist, so If we do it for DataSources won't it complex a bit your last idea?

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -71,7 +74,20 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
 
-	pvcSource := dataVolume.Spec.Source.PVC
+	if dataVolume.Spec.Source != nil {
+		pvcSource = dataVolume.Spec.Source.PVC
+	} else if dataVolume.Spec.SourceRef != nil && dataVolume.Spec.SourceRef.Kind == cdiv1.DataVolumeDataSource {
+		ns := dataVolume.Namespace
+		if dataVolume.Spec.SourceRef.Namespace != nil && *dataVolume.Spec.SourceRef.Namespace != "" {
+			ns = *dataVolume.Spec.SourceRef.Namespace
+		}
+		dataSource, err := wh.cdiClient.CdiV1beta1().DataSources(ns).Get(context.TODO(), dataVolume.Spec.SourceRef.Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
arnongilboa: I'll sure add it, although not sure it's part of the req/design.

pkg/controller/datavolume-controller.go
```diff
@@ -2061,6 +2072,71 @@ func (r *DatavolumeReconciler) newPersistentVolumeClaim(dataVolume *cdiv1.DataVo
 	return pvc, nil
 }
 
+func (r *DatavolumeReconciler) getDataVolumeSourcePVC(dv *cdiv1.DataVolume) (*cdiv1.DataVolumeSourcePVC, error) {
```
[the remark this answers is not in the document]
arnongilboa: Right

awels:
I would like to add, we currently ensure on a clone that the source PVC exists (mainly to check volume modes and size requirements) during validation. Since the dataSourceRef will be mainly cloning, we will probably want to do the same.

pkg/controller/datavolume-controller.go
```diff
@@ -507,46 +508,48 @@ func (r *DatavolumeReconciler) isCSIClonePossible(dataVolume *cdiv1.DataVolume,
...
+	dvSourcePvc, _ := r.getDataVolumeSourcePVC(datavolume)
+	if dvSourcePvc == nil {
+		return NoClone, nil
+	}
+
+	// The source and target PVCs share the same Storage Class
+	// get this the same way as for getSnapshotClassForSmartClone(datavolume, pvcSpec)
+	preferredCloneStrategy, err := r.getCloneStrategy(datavolume)
+	if err != nil {
+		return 0, err
```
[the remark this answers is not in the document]
arnongilboa: Not my code, just un-tabbed this block... but will sure fix that!

arnongilboa:
> I would like to add, we currently ensure on a clone that the source PVC exists (mainly to check volume modes and size requirements) during validation. Since the dataSourceRef will be mainly cloning, we will probably want to do the same.

@mhenriks - considering that, can we still support dataSources that don't exist?

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
brybacki: yes mine ;) And it should be `NoClone`, but it is ignored anyway because of err

mhenriks:
> @mhenriks - considering that, can we still support dataSources that don't exist?

Okay let's make sure DataSources exist for now.

mhenriks:
> Cool, much smarter and cleaner. although the it makes the solution a bit bit hacky and will require a comment or two explaining the trick.

Kind of a hack, but we really should enable the DataVolume status subresource and take away DataVolume controller permission to update DataVolume specs

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
arnongilboa: Function removed

pkg/apiserver/webhooks/datavolume-mutate.go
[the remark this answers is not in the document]
arnongilboa: Won't add it, follwing our discussion with @awels .

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
arnongilboa: fixed

tests/api_validation_test.go
```diff
@@ -113,6 +113,37 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
...
+					}
+					return false
+				}, timeout, pollingInterval).Should(BeTrue())
+
+			},
+				table.Entry("[test_id:XXXX]fail with empty kind", "", "test", "test-pvc"),
+				table.Entry("[test_id:XXXX]fail with unsupported kind", "no-such-kind", "test", "test-pvc"),
+				table.Entry("[test_id:XXXX]fail with empty sourceRef name", "DataSource", "test", ""),
+				table.Entry("[test_id:XXXX]fail with sourceRef DataSource doesn't exist", "DataSource", "test", "test-pvc"),
+			)
```
duyanyan, line 145: table.Entry("[test_id:6780]fail with empty kind", "", "test", "test-pvc"),
table.Entry("[test_id:6781]fail with unsupported kind", "no-such-kind", "test", "test-pvc"),
table.Entry("[test_id:6782]fail with empty sourceRef name", "DataSource", "test", ""),
table.Entry("[test_id:6783]fail with sourceRef DataSource doesn't exist", "DataSource", "test", "test-pvc"),

tests/cloner_test.go
```diff
@@ -108,6 +108,34 @@ var _ = Describe("all clone tests", func() {
 			completeClone(f, f.Namespace, targetPvc, diskImagePath, sourceMD5, sourcePvcDiskGroup)
 		})
 
+		It("[test_id:XXXX]Should clone imported data from SourceRef PVC DataSource", func() {
```
duyanyan, line 111: It("[test_id:6784]Should clone imported data from SourceRef PVC DataSource", func() {

tests/api_validation_test.go
arnongilboa: Thanks Yan, but you missed:
`[test_id:XXXX]fail with both source and sourceRef`

tests/api_validation_test.go
```diff
@@ -284,13 +314,13 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
...
 			table.Entry("[test_id:1860]fail with missing http url key", "manifests/dvMissingSourceHttp.yaml", true, "Missing Data volume source", "spec.source.http in body must be of type object"),
 			table.Entry("[test_id:1858]fail with missing datavolume spec", "manifests/dvMissingCompleteSpec.yaml", true, "Missing Data volume source", "missing required field \"spec\" in io.kubevirt.cdi.v1beta1.DataVolume", " invalid: spec: Required value"),
 			// k8s < 1.15 return Required value: name or generateName is required, >= 1.15 return error validating data: unknown object type "nil" in DataVolume.metadata
 			table.Entry("[test_id:1857]fail without datavolume name", "manifests/dvNoName.yaml", true, "Required value: name or generateName is required", "error validating data: unknown object type \"nil\" in DataVolume.metadata"),
 			table.Entry("[test_id:1856]fail without meta data", "manifests/dvNoMetaData.yaml", true, "Required value: name or generateName is required"),
+			table.Entry("[test_id:XXXX]fail with both source and sourceRef", "manifests/dvBothSourceAndSourceRef.yaml", true, "Data volume should have either Source or SourceRef"),
```
duyanyan, line 323: table.Entry("[test_id:6786]fail with both source and sourceRef"

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -263,50 +281,90 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *admission
...
+	}
+	ns := namespace
+	if spec.SourceRef.Namespace != nil && *spec.SourceRef.Namespace != "" {
+		ns = spec.SourceRef.Namespace
+	}
+	dataSource, err := wh.cdiClient.CdiV1beta1().DataSources(*ns).Get(context.TODO(), spec.SourceRef.Name, metav1.GetOptions{})
+	if err != nil {
+		return &metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueNotFound,
+			Message: fmt.Sprintf("SourceRef %s/%s/%s doesn't exist, %v", spec.SourceRef.Kind, *ns, spec.SourceRef.Name, err),
```
mhenriks, line 328: Error doesn't necessarily mean the resource does not exist.  Could be a problem communicating with apiserver

pkg/apiserver/webhooks/datavolume-mutate.go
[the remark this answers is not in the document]
awels: Normally I would agree with @mhenriks but we are already checingk if source PVCs exists for clone, and since clone sources are going to be the majority of datasource, we should remain consistent.

pkg/controller/datavolume-controller.go
```diff
@@ -347,6 +347,24 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 		return reconcile.Result{}, nil
 	}
 
+	// If sourceRef is set, populate spec.Source with data from the DataSource
+	if datavolume.Spec.SourceRef != nil {
```
mhenriks, line 351: would it be better to extract this to a function?  That reconcile function is already big may be more readable

mhenriks (commented):
minor comments.  almost there!

pkg/apiserver/webhooks/datavolume-mutate.go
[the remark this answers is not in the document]
kwiesmueller: Is this Get using a watch cache?
Depending on how often we think we'll see these cases it might be better to not cause api requests for this if we can have a watcher instead. Not sure how CDI does that right now.

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -71,7 +74,20 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
+		if err != nil {
+			return toAdmissionResponseError(err)
+		}
+		pvcSource = dataSource.Spec.Source.PVC
```
kwiesmueller, line 88: Isn't it possible, that a DataSource doesn't point to a PVC yet?
Will this cause any problems here?

mhenriks: Yeah should make sure that's not nil

arnongilboa: WRT nil PVC, I don't see any difference from the original one a few lines above:
```
if dataVolume.Spec.Source != nil {
    pvcSource = dataVolume.Spec.Source.PVC
} ...
```
Anyway we have a check a few lines below:
```
if pvcSource == nil {
    klog.V(3).Infof("DataVolume %s/%s not cloning", targetNamespace, targetName)
    return allowedAdmissionResponse()
}
```

pkg/controller/datavolume-controller.go
arnongilboa: sure

tests/manifests/dvMissingSource.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: test-dv
 spec:
   source:
+  sourceRef:
```
awels, line 7: why is this needed? it not being there is the same as leaving it blank no?

awels (commented):
I am missing something, I see the operator give permissions to the api server to get/list datasources, but I don't see anything giving the operator those permissions, but something must give it those permissions otherwise it can't grant them to the api server.

mhenriks:
@awels the operator's cluster role is constructed as the aggregate of all other roles plus other stuff

pkg/apiserver/webhooks/datavolume-mutate.go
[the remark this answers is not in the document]
mhenriks: Not cached but I don't think we expect high volume.  At least initially

pkg/apiserver/webhooks/datavolume-validate.go
mhenriks: Not a huge deal, just thinking that if returning `CauseTypeFieldValueNotFound` then explicitly check for not found and return more generic error if something else fails

awels:
Gotcha, okay, then that was the part I was missing. Ignore that comment then.

tests/manifests/dvMissingSource.yaml
arnongilboa: No real reason. Just because it was there already for `source`.
So may we remove this yaml together with `test_id:1762` and keep only `test_id:1861`. They are simillar except the blank pvc.
```
$ diff tests/manifests/dvMissingSource.yaml tests/manifests/dvMissingSource2.yaml
6,7d5
<   source:
<   sourceRef:
```

pkg/apiserver/webhooks/datavolume-mutate_test.go
```diff
@@ -180,6 +181,7 @@ func mutateDVs(key *rsa.PrivateKey, ar *admissionv1.AdmissionReview, isAuthorize
 		}
 		return true, sar, nil
 	})
-	wh := NewDataVolumeMutatingWebhook(client, key)
+	cdiClient := cdiclientfake.NewSimpleClientset()
+	wh := NewDataVolumeMutatingWebhook(client, cdiClient, key)
```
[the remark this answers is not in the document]
arnongilboa: added

PR #1864, opened 2021-07-20
Add relationship labels

akalenyu:
Users don't want :alien: resources in clusters,
and we should also be able to tell if we're part of a broader installation.

Notes:
- Operator created resources were handled in kubevirt/controller-lifecycle-operator-sdk#18
as these labels will be common to all resources deployed by the HCO.
(thanks mhenriks).
- Now that the controller is guaranteed to have the labels, we can set env vars
that reference the label values (fieldRef) to spare calling GET on the CR in the controllers.

Fixes #

`common.AppKubernetesComponentLabel: "storage",` - not sure about this,
should we maybe go with `containerized-data-importer` instead?

**Release note**:

```release-note
Recommended k8s labels for cdi resources
```


cmd/cdi-controller/leaderelection.go
```diff
@@ -30,10 +33,19 @@ const (
 
 func startLeaderElection(ctx context.Context, config *rest.Config, leaderFunc func()) error {
 	client := kubernetes.NewForConfigOrDie(config)
+	cdiClient := cdiclient.NewForConfigOrDie(config)
 	namespace := util.GetNamespace()
 
 	// create manually so it has CDI component label
-	err := createConfigMap(client, namespace, configMapName)
+	crList, err := cdiClient.CdiV1beta1().CDIs().List(context.TODO(), metav1.ListOptions{})
```
mhenriks, line 40: I'd like to keep our interactions with the CDI CR to a minimum.  Ideally just the operator.  I understand that is not always possible.  But I think it may be better to have the operator pass these label values as env vars to the deployments.  Then the values can just be cached locally.  If the values happen to change, the deployments will get updated.  THink about it

cmd/cdi-controller/leaderelection.go
```diff
@@ -54,14 +66,17 @@ func startLeaderElection(ctx context.Context, config *rest.Config, leaderFunc fu
 	return nil
 }
 
-func createConfigMap(client kubernetes.Interface, namespace, name string) error {
+func createConfigMap(client kubernetes.Interface, cr *cdiv1.CDI, namespace, name string) error {
+	dynamicLabels := util.GetRecommendedLabels(cr, "cdi-controller")
```
mhenriks, line 70: If we go this route, I think it would be better to have a function called `AddDynamicLabels` that takes a `metav1.Object`

cmd/cdi-controller/leaderelection.go
akalenyu: That sounds better, just to be sure -> 
1. This means I'd have to start with `controller-lifecycle-operator-sdk`, make sure that `operator pass these label values as env vars to the deployments` is completed first?
(After this is in, I'll be able to fetch the env vars containing the values in the controllers)
2. Your review is tagging a specific file but I'm guessing the same goes for most occurrences like the ones in https://github.com/kubevirt/containerized-data-importer/pull/1864/files#diff-868af4b730094e9180f78af4cc069ae6cecf7077a46c186553f331dccbd32993 for example?

cmd/cdi-controller/leaderelection.go
akalenyu: I guess if we go with env vars then this comment does not apply (correct me if I'm wrong?)

Just to understand, the suggestion here is to edit the cm after creation?

go.mod
```diff
@@ -47,7 +47,7 @@ require (
 	k8s.io/klog/v2 v2.4.0
 	k8s.io/kube-aggregator v0.20.2
 	k8s.io/kube-openapi v0.0.0-20201113171705-d219536bb9fd
-	kubevirt.io/controller-lifecycle-operator-sdk v0.2.0
+	kubevirt.io/controller-lifecycle-operator-sdk v0.2.1-0.20210723143736-64585ea1d1bd // TODO: update when release is made
```
akalenyu, line 50: A proper release will be made when jakub-dzon from the `controller-lifecycle-operator-sdk` is back
(This is vendoring from the merged PR in main)

PR #1869, opened 2021-07-23
Add DataImportCron controller

arnongilboa:

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -186,6 +186,11 @@ func start(ctx context.Context, cfg *rest.Config) {
 		os.Exit(1)
 	}
 
+	if _, err := controller.NewDataImportCronController(mgr, log); err != nil {
+		klog.Errorf("Unable to setup dataimportcron controller: %v", err)
+		os.Exit(1)
+	}
+
 	klog.V(1).Infoln("created cdi controllers")
 
 	go crdInformerFactory.Start(ctx.Done())
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -68,6 +68,7 @@ func main() {
 	finalCheckpoint, _ := util.ParseEnvVar(common.ImporterFinalCheckpoint, false)
 	preallocation, err := strconv.ParseBool(os.Getenv(common.Preallocation))
 	var preallocationApplied bool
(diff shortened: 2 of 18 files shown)


awels:
So I am quite confused right now. I have not followed everything that closely but from what I understand the dataImportCron controller should be doing the following:
1. Watch dataImportCron objects and create a cron job based on the schedule field in the dataImportCron that triggered the reconcile.
2. The cron job creates a PVC (or DV) with the appropriate annotations so that the CDI controller will start to populate it from the source specified in the dataImportCron object. It also updates the The dataImportCron.Status.LastImportedPVC to the new PVC. (maybe we need to 
3. Once the PVC from 2 is completed. update the DataSource.Spec.Source to the new PVC. For the Datasource that is specified in the ManagedDataSource of the dataImportCron.
Nowhere do we actually update datavolumes so all the code I see that is updating datavolumes is not needed I think. Now the cron job needs to have some k8s knowledge in order to create the PVC (or DV). I haven't looked into Cron to see how hard/easy that is.

arnongilboa:
@awels 
> So I am quite confused right now. I have not followed everything that closely but from what I understand the dataImportCron controller should be doing the following:
> 
> 1. Watch dataImportCron objects and create a cron job based on the schedule field in the dataImportCron that triggered the reconcile.

Right, and also watch the import DVs created by the cronjob Run(). It will also garbage collect old PVCs (FIFO).

> 2. The cron job creates a PVC (or DV) with the appropriate annotations so that the CDI controller will start to populate it from the source specified in the dataImportCron object. It also updates the The dataImportCron.Status.LastImportedPVC to the new PVC. (maybe we need to

Right.

> 3. Once the PVC from 2 is completed. update the DataSource.Spec.Source to the new PVC. For the Datasource that is specified in the ManagedDataSource of the dataImportCron.

Right.

>    Nowhere do we actually update datavolumes so all the code I see that is updating datavolumes is not needed I think. Now the cron job needs to have some k8s knowledge in order to create the PVC (or DV). I haven't looked into Cron to see how hard/easy that is.

We sure have no code updating the DVs...no idea where you see it :)
Cron job Run() uses the contoller-passed client to create the registry import DVs. Is is a bad practice?

awels:
Could have sworn at the end of the reconcile loop I saw an update to a DV. But its not there, let me look again.

PR #1873, opened 2021-07-28
Default disk alignment to 1Mi instead of 512 bytes

awels:

We had hardcoded the block size at 512 bytes, which can cause alignment issues when the hardware is a 4k block device. This will manifest itself as
```
{"component":"virt-launcher","level":"error","msg":"internal error: qemu unexpectedly closed the monitor: 2021-06-15T15:16:06.788431Z qemu-kvm: -device virtio-blk-pci-non-transitional,bus=pci.6,addr=0x0,drive=li
bvirt-1-format,id=ua-disk-0,write-cache=on: Cannot get 'write' permission without 'resize': Image size is not a multiple of request alignment","pos":"qemuProcessReportLogError:2111","subcomponent":"libvirt","thread":"290","timestamp":"2021-06-15T15:16:06.811000Z"}
```
in KubeVirt as reported in https://bugzilla.redhat.com/show_bug.cgi?id=1976730

Fixes #

**Release note**:

```release-note
BugFix: Align disk image size to 1Mi instead of 512 bytes.
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -29,6 +29,8 @@ import (
 
 const (
 	blockdevFileName = "/usr/sbin/blockdev"
+	// DefaultAlignBlockSize is the alignment size we use to align disk images, its a multiple of all known hardware block sizes 512/4k/8k/32k/64k.
+	DefaultAlignBlockSize = 1024 * 1024
 )
 
 // CountingReader is a reader that keeps track of how much has been read
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -2285,11 +2285,10 @@ func getDefaultAccessModes(c client.Client, storageClass *storagev1.StorageClass
 
 // GetRequiredSpace calculates space required taking file system overhead into account
 func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
-	blockSize := int64(512)
 	// count overhead as a percentage of the whole/new size
 	spaceWithOverhead := int64(float64(requestedSpace) / (1 - filesystemOverhead))
 	// qemu-img will round up, making us use more than the usable space.
 	// This later conflicts with image size validation.
-	qemuImgCorrection := util.RoundUp(spaceWithOverhead, blockSize)
+	qemuImgCorrection := util.RoundUp(spaceWithOverhead, util.DefaultAlignBlockSize)
 	return qemuImgCorrection
 }
--- a/pkg/importer/data-processor.go
+++ b/pkg/importer/data-processor.go
@@ -285,7 +285,7 @@ func (dp *DataProcessor) resize() (ProcessingPhase, error) {
 		// Change permissions to 0660
 		err := os.Chmod(dp.dataFile, 0660)
 		if err != nil {
-			err = errors.Wrap(err, "Unable to change permissions of target file")
+			return ProcessingPhaseError, errors.Wrap(err, "Unable to change permissions of target file")
 		}
 	}
 
@@ -361,10 +361,9 @@ func (dp *DataProcessor) getUsableSpace() int64 {
 
 // GetUsableSpace calculates space to use taking file system overhead into account
 func GetUsableSpace(filesystemOverhead float64, availableSpace int64) int64 {
-	blockSize := int64(512)
 	spaceWithOverhead := int64((1 - filesystemOverhead) * float64(availableSpace))
 	// qemu-img will round up, making us use more than the usable space.
 	// This later conflicts with image size validation.
-	qemuImgCorrection := util.RoundDown(spaceWithOverhead, blockSize)
+	qemuImgCorrection := util.RoundDown(spaceWithOverhead, util.DefaultAlignBlockSize)
 	return qemuImgCorrection
 }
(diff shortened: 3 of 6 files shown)


pkg/util/util.go
```diff
@@ -31,6 +32,25 @@ const (
...
+	findMntByPath = func(path string) ([]byte, error) {
+		return exec.Command("/usr/bin/findmnt", "-J", path).CombinedOutput()
+	}
+
+	getHwSectorSize = func(deviceName string) ([]byte, error) {
+		return exec.Command("cat", fmt.Sprintf("/sys/block/%s/queue/hw_sector_size", deviceName)).CombinedOutput()
+	}
+
+	getlsblk = func() ([]byte, error) {
+		return exec.Command("/usr/bin/lsblk", "-J").CombinedOutput()
```
maya-r, line 45: I am remembering #1258  (use blockdev instead of lsblk due to limitations with dev-mapper) and I wonder whether we should use blockdev here for the same reason.

awels: I am not using lsblk to get size, I am using lsblk to find all the block devices, so I can then use the name to find the actual block size.

brybacki:
What about a code path where fsOverhead is applied to requested size of `storage`. Can it also be problematic?

https://github.com/kubevirt/containerized-data-importer/blob/2254cf0c1f1871efe22307c447b2042f87a6779c/pkg/controller/datavolume-controller.go#L2287

awels:
This PR is about calculating the value that is currently hardcoded at 512 in the function you pointed at. We are aligning the size of the disk image to that block size. But if your HW block size is lets say 4k (for a lot of newer devices) then it is possible the calculation won't align if it is hardcoded at 512. So instead we look up the HW block size, and then align to that.

pkg/util/util.go
awels: Also note we can't use blockdev, because it requires a device to be passed. But in certain scenarios (like lets say local storage as in our tests) the actual device is /dev/vda, but that device doesn't exist inside the container (no /dev/vda available to pass to blockdev). Fortunately /sys/block/<device> is available and contains the HW block size.

brybacki:
> This PR is about calculating the value that is currently hardcoded at 512 in the function you pointed at. We are aligning the size of the disk image to that block size. But if your HW block size is lets say 4k (for a lot of newer devices) then it is possible the calculation won't align if it is hardcoded at 512. So instead we look up the HW block size, and then align to that.

I am not sure I understand you correctly, but what I mean is that the PR does not update the GetRequiredSpace in datavolume_controller.go, and my question was if it also should be updated.

Because currently it looks like this:
```
func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
	blockSize := int64(512)
	// count overhead as a percentage of the whole/new size
	spaceWithOverhead := int64(float64(requestedSpace) / (1 - filesystemOverhead))
	// qemu-img will round up, making us use more than the usable space.
	// This later conflicts with image size validation.
	qemuImgCorrection := util.RoundUp(spaceWithOverhead, blockSize)
	return qemuImgCorrection
}
```

awels:
Ahhh okay, that is a good question, and the answer is no. For a few reasons.
1. That calculation is done in the context of the datavolume controller, while the disk aligning part is done in the context of the importer. The datavolume controller doesn't know what hardware is on the node the importer will run on, so it can't actually make that determination.
2. The calculation in the datavolume controller is to determine a value to put in the PVC request. There is no guaranteed that the provisioner will create a PV of that exact size. So IMO it is perfectly fine to not align is the PVC request size, the actual disk.img is the one that needs to align to the HW block size.

brybacki:
> Ahhh okay, that is a good question, and the answer is no. For a few reasons.
> 
> 1. That calculation is done in the context of the datavolume controller, while the disk aligning part is done in the context of the importer. The datavolume controller doesn't know what hardware is on the node the importer will run on, so it can't actually make that determination.
> 2. The calculation in the datavolume controller is to determine a value to put in the PVC request. There is no guaranteed that the provisioner will create a PV of that exact size. So IMO it is perfectly fine to not align is the PVC request size, the actual disk.img is the one that needs to align to the HW block size.

Ok

awels:
Hmmm nfs seems to fail a lot with this PR, and I mean a ton of the tests are failing. Let me try running nfs locally and see if they fail as well

awels:
/hold
Need to think about this a little more. the NFS lane was failing because it was a remote mount and thus we have no possible way of figuring out what the real HW block size is. I am starting to think this entire PR is not really needed. we just need to make the alignment something that will be a duplicate of 512 and 4096, something large enough to be somewhat future proof.

akalenyu:
Just an FYI, saw a similar err msg in kubevirt - https://github.com/kubevirt/kubevirt/issues/5785

awels:
That one has been fixed. It has to do with the alignment of the cloudinit iso that is generated. That is not related to this issue.

awels:
Modified the PR to just align to 1Mi instead of 512 bytes. Plus some test fixes

pkg/util/util.go
```diff
@@ -29,6 +29,8 @@ import (
 
 const (
 	blockdevFileName = "/usr/sbin/blockdev"
+	// DefaultAlignBlockSize is the alignment size we use to align disk images.
```
brybacki, line 32: Since this is our const and not some well defined standard it would be nice to have some very short info on "why 1Mi". Unless you thin it is obvious.

awels: Well since you had to ask it is not obvious, but 1Mi is a multiple of all known HW block sizes 512/4k/8k/32k/64k thus it will align on all of them.

PR #1895, opened 2021-08-16
Correct md5sum verification in tests

brybacki:

A problem was found with the way verification command is run. 

```
head -c [numBytes] [fileName] | md5sum

```
This command fails when the file path is bad, but the empty output is piped to md5 and the result is success with md5 of empty input.

In effect some tests were always passing. This was corrected.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/smartclone_test.go
+++ b/tests/smartclone_test.go
@@ -56,13 +56,13 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests th
 		_, err := f.CdiClient.CdiV1beta1().CDIs().Update(context.TODO(), &cdiCr, metav1.UpdateOptions{})
 		Expect(err).ToNot(HaveOccurred())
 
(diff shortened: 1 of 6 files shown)


tests/smartclone_test.go
```diff
@@ -79,17 +79,18 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
 	f := framework.NewFramework("dv-func-test")
 
 	It("[rfe_id:1106][test_id:3494][crit:high][vendor:cnv-qe@redhat.com][level:component] Verify DataVolume Smart Cloning - volumeMode filesystem - Positive flow", func() {
+		//diskImagePath := filepath.Join(testBaseDir, testFile)
```
awels, line 82: leftover?

awels (commented):
looks good one minor leftover that needs to get removed.

PR #1896, opened 2021-08-16
Allow admin user to set CDI control plane priority class via CDI cr.

awels:

Allow users to specify the priority class of the CDI control
plane using the CDI CR. Verifies that the set value is valid
before passing to the deployments.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1953484

**Release note**:

```release-note
Enhancement: Allow admin user to specify priority class of CDI control plane
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -17,6 +17,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	schedulev1 "k8s.io/api/scheduling/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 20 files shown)


tests/utils/pod.go
```diff
@@ -115,7 +115,7 @@ func findPodByCompFuncOnce(clientSet *kubernetes.Clientset, namespace, prefix, l
 	for _, pod := range podList.Items {
 		if compFunc(pod.Name, prefix) {
 			if result == nil {
-				result = &pod
+				result = pod.DeepCopy()
```
arnongilboa, line 118: why ref is not good enough here?

tests/operator_test.go
```diff
@@ -780,6 +781,119 @@ var _ = Describe("ALL Operator tests", func() {
...
+				// API server
+				verifyPodPriorityClass(cdiApiServerPodPrefix, string(prioClass), "")
+				// Upload server
+				verifyPodPriorityClass(cdiUploadProxyPodPrefix, string(prioClass), "")
+				By("Verifying there is just a single cdi controller pod")
+				Eventually(func() error {
+					_, err := utils.FindPodByPrefix(f.K8sClient, f.CdiInstallNs, cdiDeploymentPodPrefix, common.CDILabelSelector)
+					return err
+				}, 2*time.Minute, 1*time.Second).Should(BeNil())
+				By("Ensuring this pod is the leader")
```
arnongilboa, line 853: why we need this check?

tests/utils/pod.go
awels: It has to do with how go uses the index/value in range loops. If you just use the ref, then next loop the ref will be the new value. But we want to store the old value. You can see it in the output if there is another pod with the same prefix. Both names printed will be the same. So instead I copy the value and store that.

tests/operator_test.go
awels: Because we just changed the cdi deployment pod, and if we just happen to run a leader election test next, it will fail because the leader election hasn't happened yet. So this ensures that the deployment pod is a single pod, AND it is the current active leader. So any following tests won't be messed up.

awels:
I need another lgtm, the CI was down and it didn't label the PR.

tests/utils/pod.go
akalenyu: Just hit this on the 1.38 release branch, this means `result` will hold last pod that we iterated through (can be any pod in the namespace), do we want to backport this small fix?

awels: This particular fix is for debug printing the right pod info. Did you run into a failure and it printed the wrong info? We can backport this but doesn't seem that useful.

akalenyu: This can give you the wrong pod when using `FindPodByPrefix()` which is bad

PR #1886, opened 2021-08-09
Bump kubevirtci for uploadproxy port

akalenyu:
Bumping kubevirtci to get https://github.com/kubevirt/kubevirtci/pull/653

Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
Since https://github.com/kubevirt/kubevirtci/pull/629 (verified)
our `cdi-docker-registry-host` test container breaks:
```bash
[root@localhost containerized-data-importer]# ./cluster-up/kubectl.sh logs -f cdi-docker-registry-host-6bb67d9f4d-xhg2x -n cdi registry-populate
selecting docker as container runtime
health
building image tinycoreiso
Error: error writing "0 0 4294967295\n" to /proc/42/uid_map: write /proc/42/uid_map: operation not permitted
level=error msg="error writing \"0 0 4294967295\\n\" to /proc/42/uid_map: write /proc/42/uid_map: operation not permitted"
level=error msg="(unable to determine exit status)"
Exiting on error
```
This is due to a kernel bump on the node (CentOS VM) between #629 and the commit previous to that (confirmed).
The specific patch that "breaks" us is probably https://lore.kernel.org/patchwork/patch/1415008/.
We were never supposed to be able to run `buildah bud` without the `SETFCAP` capability:
https://www.redhat.com/sysadmin/podman-inside-kubernetes
Building an image using buildah requires a properly installed shadow-utils package.  This requires `SETFCAP`.

**Release note**:

```release-note
NONE
```

PR #1897, opened 2021-08-17
Use NodePort instead of kubectl port-forward for upload tests

akalenyu:

Follow up for #1886, we now have a kubevirtci version that gives us a free port.
We can use that to make the switch from `kubectl port-forward` which has been known to cause many flakes in our CI.

Fixes #

We are only going to fall back to `kubectl port-forward` when override is not set:
https://github.com/kubevirt/containerized-data-importer/blob/main/tests/upload_test.go#L74-L78

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/sync.sh
+++ b/cluster-sync/sync.sh
@@ -80,6 +80,12 @@ function wait_cdi_available {
   fi
 }
 
+function configure_uploadproxy_override {
+  host_port=$(./cluster-up/cli.sh ports uploadproxy | xargs)
+  override="127.0.0.1:$host_port"
+  _kubectl patch cdi ${CR_NAME} --type=merge -p '{"spec": {"config": {"uploadProxyURLOverride": "'"$override"'"}}}'
+}
+
 OLD_CDI_VER_PODS="./_out/tests/old_cdi_ver_pods"
 NEW_CDI_VER_PODS="./_out/tests/new_cdi_ver_pods"
 
@@ -119,6 +125,7 @@ if [ "${KUBEVIRT_PROVIDER}" != "external" ] && [ "${CDI_SYNC}" == "test-infra" ]
   _kubectl apply -f "./_out/manifests/file-host.yaml"
   _kubectl apply -f "./_out/manifests/registry-host.yaml"
   _kubectl apply -f "./_out/manifests/test-proxy.yaml"
+  _kubectl apply -f "./_out/manifests/uploadproxy-nodeport.yaml"
   # Imageio test service:
   _kubectl apply -f "./_out/manifests/imageio.yaml"
   # vCenter (VDDK) test service:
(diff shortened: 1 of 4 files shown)


awels (commented):
So the original thought behind this is that in the cluster-sync before starting tests we would configure the CDI config section with the node port, and then start the test run. The upload tests should have a check to see if the proxyUrl is set with findProxyURLCdiConfig, if it is, then it will use that. So I would just look up the port in bash, and update the CDI config with the proxy URL. If the nodeport service is needed I would create that ahead of time as well.

akalenyu:
> So the original thought behind this is that in the cluster-sync before starting tests we would configure the CDI config section with the node port, and then start the test run. The upload tests should have a check to see if the proxyUrl is set with findProxyURLCdiConfig, if it is, then it will use that. So I would just look up the port in bash, and update the CDI config with the proxy URL. If the nodeport service is needed I would create that ahead of time as well.

If we do this, we are giving up on flexibility using the NodePort for other services, such as Prometheus:
https://github.com/kubevirt/containerized-data-importer/blob/main/tests/import_test.go#L483
I guess we only ever saw a problem with the upload proxy, right?

awels:
If we need node ports for other things, wouldn't we just add new node ports? Also I believe we have one for prometheus already.

akalenyu:
> If we need node ports for other things, wouldn't we just add new node ports? Also I believe we have one for prometheus already.

Cool. made the changes

awels:
Ahhhh now I know why the ceph lane is failing, for that lane the cdi object is not called cdi. See https://github.com/kubevirt/containerized-data-importer/blob/main/automation/test.sh#L33

cluster-sync/sync.sh
```diff
@@ -80,6 +80,12 @@ function wait_cdi_available {
   fi
 }
 
+function configure_uploadproxy_override {
+  host_port=$(./cluster-up/cli.sh ports uploadproxy | xargs)
+  override="127.0.0.1:$host_port"
+  _kubectl patch cdi cdi --type=merge -p '{"spec": {"config": {"uploadProxyURLOverride": "'"$override"'"}}}'
```
awels, line 86: Should be
```
_kubectl patch cdi $CR_NAME --type=merge -p '{"spec": {"config": {"uploadProxyURLOverride": "'"$override"'"}}}'
```

tests/cdiconfig_test.go
```diff
@@ -348,6 +370,15 @@ var _ = Describe("CDI ingress config tests", func() {
 			defaultUrl = *config.Status.UploadProxyURL
 		}
 		By("Making sure no url is set")
+		config, err := f.CdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
```
akalenyu, line 373: Changed setup & teardown around this test suite to address failures in lanes
(We don't take into account that env might have actual override set)

cluster-sync/sync.sh
akalenyu: Thanks!

PR #1899, opened 2021-08-19
Fix cluster-sync PULL_POLICY usage

arnongilboa:

`make cluster-sync` was ignoring `PULL_POLICY`  and used the default `IfNotPresent`
This was found while cluster-sync'ing with `KUBEVIRT_PROVIDER=external` to an existing cluster. 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/build/config.sh
+++ b/hack/build/config.sh
@@ -86,19 +86,3 @@ function parseTestOpts() {
         esac
     done
 }
-
-function getTestPullPolicy() {
-    local pp
-    case "${KUBEVIRT_PROVIDER}" in
-    "k8s-1.13.3")
-        pp=$PULL_POLICY
-        ;;
-    "os-3.11.0")
-        pp=Always
-        ;;
-    "okd-4.1")
-        pp=Always
-        ;;
-    esac
-    echo "$pp"
-}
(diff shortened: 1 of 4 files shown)


hack/build/config.sh
```diff
@@ -88,11 +88,8 @@ function parseTestOpts() {
 }
 
 function getTestPullPolicy() {
-    local pp
+    local pp=$PULL_POLICY
```
awels, line 91: So this function explains a lot, this tells me in Open Shift the pull policy should be always because latest doesn't always pull.. Of course with external you can't tell what it is, so maybe we should default external to always as well.
But I think all of this is too complicated. just return $PULL_POLICY and default it to IfNotPresent if not defined.

arnongilboa: So why not remove this func altogether? btw, won't it break CDI OpenShift ci, as it will use IfNotPresent instead of Always?

awels: There is no cdi openshift ci. This function was when we have okd lanes in our ci, which was a long time ago.

arnongilboa: ok, fixed.

PR #1902, opened 2021-08-23
Clone strategy csi verify

brybacki:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -412,6 +412,21 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 			return r.reconcileSmartClonePvc(log, datavolume, pvcSpec, transferName, snapshotClassName)
 		}
 		if selectedCloneStrategy == CsiClone {
+			csiDriverAvailable, err := r.storageClassCSIDriverExists(pvcSpec.StorageClassName)
+			if err != nil && !k8serrors.IsNotFound(err) {
+				return reconcile.Result{}, err
+			}
+			if !csiDriverAvailable {
+				// err csi clone not possible
+				return reconcile.Result{},
+					r.updateDataVolumeStatusPhaseWithEvent(cdiv1.CloneScheduled, datavolume, pvc, selectedCloneStrategy,
+						DataVolumeEvent{
+							eventType: corev1.EventTypeWarning,
+							reason:    ErrUnableToClone,
+							message:   fmt.Sprintf("CSI Clone configured, but no CSIDriver available for %s", *pvcSpec.StorageClassName),
+						})
+			}
+
 			return r.reconcileCsiClonePvc(log, datavolume, pvcSpec, transferName)
 		}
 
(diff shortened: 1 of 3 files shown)


awels:
Is this still a draft?

brybacki:
@awels  now it is ready for review

pkg/controller/datavolume-controller.go
```diff
@@ -2035,6 +2071,7 @@ func (r *DatavolumeReconciler) emitFailureConditionEvent(dataVolume *cdiv1.DataV
 	}
 }
 
+// TODO:
```
awels, line 2074: Leftover?

pkg/controller/datavolume-controller.go
```diff
@@ -1390,6 +1405,27 @@ func (r *DatavolumeReconciler) reconcileProgressUpdate(datavolume *cdiv1.DataVol
 	return reconcile.Result{RequeueAfter: 2 * time.Second}, nil
 }
 
+func (r *DatavolumeReconciler) getCsiDriverForStorageClass(storageClassName *string) (bool, error) {
```
awels, line 1408: I think this function should be named something like StorageClassCSIDriverExists or something of that nature. We are not actually returning the CSIDriver object, we are just determining if it exists, but the name suggests we are returning it.

tests/csiclone_test.go
```diff
@@ -72,4 +72,35 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component][crit:high] CSI Vol
...
+		if f.IsCSIVolumeCloneStorageClassAvailable() {
+			Skip("Test should only run on non-csi storage")
+		}
+
+		By(fmt.Sprintf("configure storage profile %s", cloneStorageClassName))
+		utils.ConfigureCloneStrategy(f.CrClient, f.CdiClient, cloneStorageClassName, originalProfileSpec, cdiv1.CloneStrategyCsiClone)
+
+		dataVolume, _ := createDataVolumeDontWait("dv-csi-clone-test-1", utils.DefaultImagePath, v1.PersistentVolumeFilesystem, cloneStorageClassName, f)
+		waitForDvPhase(cdiv1.CloneScheduled, dataVolume, f)
+		verifyEvent(controller.ErrUnableToClone, dataVolume.Namespace, f)
```
awels, line 85: I think we would want to revert the clone strategy override after we are done with the test in an AfterEach maybe?

pkg/controller/datavolume-controller.go
brybacki: Yes

tests/csiclone_test.go
brybacki: This is actually what is done here:
https://github.com/kubevirt/containerized-data-importer/blob/80ca2de589616e6046c224471525242eba2d7a21/tests/csiclone_test.go#L35

```
AfterEach(func() {
		if originalProfileSpec != nil {
			By("Restore the profile")
			utils.UpdateStorageProfile(f.CrClient, cloneStorageClassName, *originalProfileSpec)
		}
	})
```

awels: Oh missed that wasn't part of this PR.

awels (approved):
/lgtm
/approve

PR #1889, opened 2021-08-09
WIP: Added go-ovirt-client to simplify ImageIO data source.

janosdebugs:
**What this PR does / why we need it**:

This PR follows up on oVirt/ovirt-engine-sdk-go#199 and adds the new [go-ovirt-client](https://github.com/ovirt/go-ovirt-client) library. This library makes most of the oVirt-specific code unnecessary and also adds the ability to reliably use mocks for image upload and download.

This PR also fixes an issue where the oVirt client would be [configured not to verify server certificates](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/imageio-datasource.go#L639).

**Which issue(s) this PR fixes**:

None.

**Special notes for your reviewer**:

This PR is WIP as the go-ovirt-client changes wait for merge + tagging.

**Release note**:
```release-note
NONE
```

**TODO**:

- [ ] go-ovirt-client needs a tagged version with the required changes.
- [ ] go-ovirt needs a tagged version with the required changes.
- [ ] update vendor.

PR #1903, opened 2021-08-23
Implement multi-stage ImageIO imports.

### LINKED ISSUE #1860: Allow downloading snapshots from ImageIO data sources.

/kind enhancement

**What happened**:

A first pass at MTV warm migration from RHV is planned to work with snapshots, so CDI needs a mechanism to find and download snapshots from ImageIO. Right now the ImageIO data source does not handle snapshots.

I would like the ImageIO data source to download snapshots and apply them to a base disk image for warm migration from RHV.

Add multi-stage migration fields to an ImageIO CR, so that it will loop through the given list of snapshots and keep applying them to the base disk.

I have this most of the way implemented, I should have a pull request ready within the next few weeks.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): latest
- Kubernetes version (use `kubectl version`): 1.20.4

mrnold:
This pull request extends the multi-stage import feature to the ImageIO data source, using disk snapshots as the checkpoint implementation. This is intended to allow warm imports from RHV for MTV.

Fixes #1860

Likely looking for backport to v1.38, if possible. I think I might have missed a feature deadline though?

**Release note**:
```release-note
Implement multi-stage ImageIO imports.
```


pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -455,6 +459,34 @@ func newMultistageDataVolume(name string, final bool, checkpoints []string) *cdi
 		previous = checkpoint
 	}
 
+	var source *cdiv1.DataVolumeSource
+	if sourceType == "vddk" {
```
awels, line 463: Just a bit of food for thought, would it make sense to have a 'sourceFunc' that returns the source data used in the test. Something like:

```go
vddkSource := func() *cdiv1.DataVolumeSource {
  return &cdiv1.DataVolumeSource{
    VDDK: &cdiv1.DataVolumeSourceVDDK{
      BackingFile: "disk.img",
      URL:         "http://example.com/data",
      UUID:        "12345",
      Thumbprint:  "aa:bb:cc",
      SecretRef:   "secret",
    }
}

imageIOSource := func() *cdiv1.DataVolumeSource {
  return &cdiv1.DataVolumeSource{
    Imageio: &cdiv1.DataVolumeSourceImageIO{
      URL:           "http://example.com/data",
      DiskID:        "disk-123",
      SecretRef:     "secret",
      CertConfigMap: "certs",
    },
}

blankSource := func() *cdiv1.DataVolumeSource {
  return &cdiv1.DataVolumeSource{
    Blank: &cdiv1.DataVolumeBlankImage{},
  }
}

func newMultistageDataVolume(name string, final bool, checkpoints []string, sourceFunc func() *cdiv1.DataVolumeSource) * cdiv1.DataVolume {
...
   source := sourceFunc()
...
}
```
And then we just pass the different functions into the table based on what we are testing. That eliminates the if else, else if, else if, else from here

awels (commented):
How is the relationship between this PR and #1889 seems like there would be a lot of conflict there. I did a quick review without looking at the actual data source too closely. I want to figure out how these two PRs relate if at all.

mrnold:
> How is the relationship between this PR and #1889 seems like there would be a lot of conflict there. I did a quick review without looking at the actual data source too closely. I want to figure out how these two PRs relate if at all.

Yes, these will collide messily. I have been talking with Janos and I am planning to help get go-ovirt-client merged cleanly, but it sounds like we can take our time with that. It will take a lot more work to turn #1889 into a drop-in replacement for the current ImageIO importer.

fdupont-redhat:
> How is the relationship between this PR and #1889 seems like there would be a lot of conflict there. I did a quick review without looking at the actual data source too closely. I want to figure out how these two PRs relate if at all.

@awels, there's no direct relationship and indeed there will be conflict.

@janosdebugs, what is the release you're targeting for #1889 ? I really like the idea of using go-ovirt-client, as it will make the ImageIO importer more maintainable. Would you consider splitting #1889 into smaller PRs to allow a more progressive integration and have less conflict ?

pkg/apiserver/webhooks/datavolume-validate_test.go
mrnold: Sure, done.

janosdebugs:
@fdupont-redhat I talked with @awels and we decided that, since my PR has no urgency and I still need to implement snapshot downloading and properly tag the source SDK, this PR has priority. I will rebase as needed.

fdupont-redhat:
Thanks @janosdebugs.

pkg/importer/imageio-datasource.go
```diff
@@ -52,23 +52,29 @@ type ImageioDataSource struct {
...
 	connection ConnectionInterface
+	// currentSnapshot is the UUID of the snapshot to copy, if requested
+	currentSnapshot string
+	// previousSnapshot is the UUID of the parent snapshot, if requested
+	previousSnapshot string
 }
 
 // NewImageioDataSource creates a new instance of the ovirt-imageio data provider.
-func NewImageioDataSource(endpoint string, accessKey string, secKey string, certDir string, diskID string) (*ImageioDataSource, error) {
+func NewImageioDataSource(endpoint string, accessKey string, secKey string, certDir string, diskID string, currentCheckpoint string, previousCheckpoint string, finalCheckpoint string) (*ImageioDataSource, error) {
```
awels, line 62: is finalCheckpoint used anywhere?

pkg/importer/imageio-datasource.go
```diff
@@ -108,24 +110,54 @@ func (is *ImageioDataSource) Info() (ProcessingPhase, error) {
...
+			return ProcessingPhaseError, err
+		}
+
+		backingFile := filepath.Base(imageInfo.BackingFile)
+		if backingFile != is.previousSnapshot {
+			return ProcessingPhaseError, errors.Errorf("snapshot backing file '%s' does not match expected checkpoint '%s', unable to safely rebase snapshot", backingFile, is.previousSnapshot)
+		}
+
+		klog.Info("Successfully copied snapshot data, moving to merge phase.")
+		return ProcessingPhaseMergeDelta, nil
```
awels, line 148: Does it make sense to push this MergeDeltaPhase into the data processor. Are there other sources that would benefit from having the ability to MergeDelta? If not why not just merge the delta in this datasource?

pkg/importer/imageio-datasource.go
```diff
@@ -146,16 +178,7 @@ func (is *ImageioDataSource) Close() error {
...
-			transfersService := is.connection.SystemService().ImageTransfersService()
-			_, err = transfersService.ImageTransferService(itID).Finalize().Send()
-			if err != nil {
-				_, cancelErr := transfersService.ImageTransferService(itID).Cancel().Send()
-				if cancelErr != nil {
-					klog.Errorf("Unable to finalize or cancel transfer! Disk may remain locked until inactivity timeout.\n%v\n%v", err, cancelErr)
-				}
-			}
-		}
+		cleanupTransfer(is.connection, is.imageTransfer)
```
awels, line 181: Should we check the err returned by cleanupTransfer here, like in the defer above?

pkg/importer/imageio-datasource.go
```diff
@@ -196,26 +219,70 @@ func (is *ImageioDataSource) pollProgress(reader *util.CountingReader, idleTime,
...
 	}
+	if it == nil {
+		return nil, uint64(0), nil, conn, errors.New("returned ImageTransfer was nil")
+	}
 
 	// Use the create client from http source.
 	client, err := createHTTPClient(certDir)
 	if err != nil {
-		cancelTransfer(conn, it)
+		cleanupTransfer(conn, it)
```
awels, line 280: Would it make sense to defer the cleanup at the top of this function so we ensure it always get called regardless of errors. Also do we need to clean up the transfer if we are successful? If so, I would also in that defer at least check the err returned during cleanupTransfer, and at a minimum log it

pkg/importer/imageio-datasource.go
```diff
@@ -243,30 +310,183 @@ func createImageioReader(ctx context.Context, ep string, accessKey string, secKe
...
+	}
+
+	for index, storageDomain := range storageDomains.Slice() {
+		storageDomainName, success := storageDomain.Id()
+		if !success {
+			klog.Warningf("Skipping storage domain #%d listed on disk %s", index, diskID)
+			continue
+		}
+
+		storageDomainsService := conn.SystemService().StorageDomainsService()
```
awels, line 459: Can this ever be nil? If so the next line is a potential npe. If not, you can just chain like this:
```go
storageDomainService := conn.SystemService().StorageDomainsService().StorageDomainService(storageDomainName)
```
Same question for storageDomainService, which seems to only be used to find snapshots. And again for diskSnapshotsService, which seems to only be used to list the requests.

pkg/importer/imageio-datasource.go
```diff
@@ -277,30 +497,36 @@ func getTransfer(conn ConnectionInterface, diskID string) (*ovirtsdk4.ImageTrans
...
-		ovirtsdk4.DISKFORMAT_RAW,
+		formatType,
+	).InactivityTimeout(
+		60,
 	).Build()
 	if err != nil {
 		return nil, uint64(0), errors.Wrap(err, "Error preparing transfer object")
 	}
 
+	transfersService := conn.SystemService().ImageTransfersService()
```
awels, line 522: Same question as before, can this be nil? if not why not just chain it? If it can be nil, we should check for it.

pkg/importer/imageio-datasource.go
```diff
@@ -277,30 +497,36 @@ func getTransfer(conn ConnectionInterface, diskID string) (*ovirtsdk4.ImageTrans
...
+	transfer := transfersService.Add()
 	transfer.ImageTransfer(imageTransfer)
 	var it = &ovirtsdk4.ImageTransfer{}
 	for {
 		response, err := transfer.Send()
 		if err != nil {
-			if strings.Contains(err.Error(), "Disk is locked") {
+			if strings.Contains(err.Error(), "Disk is locked") || strings.Contains(err.Error(), "disks are locked") {
```
awels, line 529: Does transfer.Send() return a typed error? I am not a huge fan of checking for strings in an error if the source is sending a typed error.

pkg/importer/imageio-datasource.go
```diff
@@ -320,9 +546,22 @@ func getTransfer(conn ConnectionInterface, diskID string) (*ovirtsdk4.ImageTrans
 		} else if phase == ovirtsdk4.IMAGETRANSFERPHASE_TRANSFERRING {
 			break
 		} else {
+			cleanupTransfer(conn, it)
```
awels, line 549: maybe better as a defer? At a minimum check the returned err

pkg/importer/imageio-datasource.go
janosdebugs: If `conn` is not nil, then the rest will AFAIK never be nil. These are purely Java-esque constructs which we can't change now without breaking compatibility.

pkg/importer/imageio-datasource.go
janosdebugs: Not as far as I know, unless `conn` itself is `nil`.

pkg/importer/imageio-datasource.go
janosdebugs: No, there is no typed error for disk locked unfortunately. This comes directly from the oVirt engine and there is no way to identify it properly. However, you can sidestep the issue by using the correlation_id trick. This involves setting the correlation_id query parameter and then checking if the jobs with that ID are completed. See here: https://github.com/oVirt/go-ovirt-client/blob/5f8e8c7a3b188698e905f54b40a091d64accc88b/client_util.go#L24

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -468,15 +503,7 @@ func newMultistageDataVolume(name string, final bool, checkpoints []string) *cdi
...
-			Source: &cdiv1.DataVolumeSource{
-				VDDK: &cdiv1.DataVolumeSourceVDDK{
-					BackingFile: "disk.img",
-					URL:         "http://example.com/data",
-					UUID:        "12345",
-					Thumbprint:  "aa:bb:cc",
-					SecretRef:   "secret",
-				},
-			},
+			Source:          source,
```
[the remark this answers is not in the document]
mrnold: Done.

pkg/importer/imageio-datasource.go
mrnold: I thought it would be when I started, but the ImageIO data source itself ended up not needing it. I took this out.

pkg/importer/imageio-datasource.go
mrnold: Thanks Janos, I chained most of them.

pkg/importer/imageio-datasource.go
mrnold: I chained this one too.

pkg/importer/imageio-datasource.go
mrnold: This is a good trick, but the disk (or snapshot) is not always locked by our own job. If the disk is already locked, there's not much the importer can do except wait for it to unlock.

pkg/importer/imageio-datasource.go
mrnold: The defer before this does the cleanup immediately after the transfer is complete (or failed) instead of leaving it open through any further phases. It checks the error result and sets is.imageTransfer to nil specifically to skip this cleanupTransfer on line 181, so it doesn't make the same kind of sense to do the same thing here.

pkg/importer/imageio-datasource.go
mrnold: No, at this point the transfer ticket is opened, but no data has been copied. It only calls cleanupTransfer on the error exits to try to make sure the disk does not stay locked after failing. This definitely should not clean up the ticket on success, so the defer would have to have some way of knowing about each failure anyway.

The cleanupTransfer function itself already does a good amount of logging, and I wouldn't want an error from cleanupTransfer to obscure the original error that caused it to get called in the first place.

pkg/importer/imageio-datasource.go
mrnold: This one is actually identical to all the other cleanupTransfers in createImageioReader, but it shouldn't be moved out of getTransfer because it's not certain there's really a transfer to clean up until this point. Either way, I still don't think checking the error will add much because there's already lots of logging inside cleanupTransfer and acting on an error from cleanupTransfer might obscure the original reason for failure.

pkg/importer/imageio-datasource.go
mrnold: I kind of liked it this way better at first because it lined up the qemu-img rebase and commit operations with the existing data processor info/resize/convert code, but maybe the implementation is too specific to ImageIO? I'm not sure if there will be other data sources that would benefit from exactly this sequence.

Here is a version with the MergeDelta work rolled up into the ImageIO data source Transfer:
https://github.com/mrnold/containerized-data-importer/compare/imageio-snapshots...mrnold:imageio-snapshots-mergedelta?expand=1

I can merge that in to this pull request if it's preferable.

pkg/importer/imageio-datasource.go
janosdebugs: Fair enough.

pkg/importer/imageio-datasource.go
janosdebugs: That's fine. The calls that need extra checking usually have a second boolean return parameter that tells you if the object was present in the response.

pkg/importer/imageio-datasource.go
awels: Well my first reaction was hmm this seems very specific to imageio, so I don't think it makes sense to expose it in the data processor. But since it is not part of the datasource interface and thus not all the datasources have to implement it, it doesn't do much harm in the data processor. I just wanted to think about if there would be anything else that would have a similar need that would make better to have it in the data processor. So far I can't think of anything. The vddk importer has all of that inside the nbdkit filter.

pkg/importer/imageio-datasource.go
awels: Okay we should encourage the sdk authors to return some typed errors, so we don't have to search for strings in errors in order to determine stuff.

pkg/importer/imageio-datasource.go
awels: This is kind of like checking a close return value in a defer on a file descriptor. There is not much you can do if an error occurs, but purists will argue you should check all errors. I would argue if we are not interested in the error from the cleanupTransfer, then why are we returning it? If we already log the error inside the function and don't want calling functions to do anything with it. Then there is no sense that it is a return parameter.

tests/datavolume_test.go
```diff
@@ -892,10 +941,16 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 				dataVolume = utils.NewDataVolumeWithHTTPImportToBlockPV(dataVolumeName, "1G", url(), f.BlockSCName)
 			case "import-vddk":
 				dataVolume = createVddkDataVolume(dataVolumeName, "1Gi", vcenterURL())
-				utils.ModifyDataVolumeWithVDDKImportToBlockPV(dataVolume, f.BlockSCName)
+				utils.ModifyDataVolumeWithImportToBlockPV(dataVolume, f.BlockSCName)
```
awels, line 944: Any reason all of these are using the block storage class?

mrnold: I think that's just what this test section is for: `Describe("Verify DataVolume with block mode",`

It seems like a good idea to test multi-stage imports with block and file volumes, I know I had a problem with this in the past.

pkg/importer/imageio-datasource.go
mrnold: I looked closer at all the cleanupTransfer calls, and I think they can be condensed in a way that it makes more sense to always check its return value. I am testing it out to make sure disks still get unlocked when they should be.

pkg/importer/imageio-datasource.go
awels: I am not that invested one way or the other, we can keep it as is for now, and see if there are other sources that will want it the same way (glance maybe?) if not we can always switch it.

tests/datavolume_test.go
awels: Yeah, that was my thought as well, in general block is the more problematic of the two but better that have some testing on both. You can add some FS tests in another PR.

pkg/importer/imageio-datasource.go
mrnold: Glance might actually be a good guess, it appears to allow the use of qcow2 for at least some operations. It may very well be possible to do the same process for warm migration from OpenStack.

mrnold:
I should mention this requires updating the local imageio-deployment image, in case anyone sees ImageIO functional test failures in their existing workspaces.

PR #1906, opened 2021-08-24
Test clone with target pvc name conflict

brybacki:

Adds a new test case to verify correct behavior when the target pvc name is taken.

PVC name conflict is already verified once by validation hook on dv creation. This tests only covers a situation when the target pvc name is taken between the dv creation and the actual clone process starting.  

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -208,6 +208,53 @@ var _ = Describe("all clone tests", func() {
 				doInUseCloneTest(f, pvcDef, targetNs, "target-dv")
 			})
 
+			It("[posneg:negative][test_id:6612]Clone with CSI as PVC source with target name that already exists", func() {
+				if cloneType == "network" {
+					Skip("Cannot simulate target pvc name conflict for host-assisted clone ")
+				}
+				pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
+				sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile+"; chmod 660 "+testBaseDir+testFile)
+				targetNamespaceName := f.Namespace.Name
+
+				// 1. use the srcPvc so the clone cannot be started
+				pod, err := f.CreateExecutorPodWithPVC("temp-pod", f.Namespace.Name, sourcePvc)
+				Expect(err).ToNot(HaveOccurred())
+
+				Eventually(func() bool {
+					pod, err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Get(context.TODO(), pod.Name, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					return pod.Status.Phase == v1.PodRunning
+				}, 90*time.Second, 2*time.Second).Should(BeTrue())
+
+				// 2. Create a clone DataVolume
+				targetDV := utils.NewCloningDataVolume("target-pvc", "1Gi", sourcePvc)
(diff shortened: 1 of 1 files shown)


awels (commented):
So I am missing something here, if I read the test correctly we do this:
1. Make source
2. Have a pod use the source.
3. Start a clone
4. Since 2 is using the source, the clone won't proceed
5. Make a different PVC with the target name in the target namespace
6. Check for events that the resource exists.

The thing I am missing is that we never delete the pod we created in 2, so we never actually attempt to do the clone. So where are the events coming from?

brybacki:
> So I am missing something here, if I read the test correctly we do this:
> 
> 1. Make source
> 2. Have a pod use the source.
> 3. Start a clone
> 4. Since 2 is using the source, the clone won't proceed
> 5. Make a different PVC with the target name in the target namespace
> 6. Check for events that the resource exists.
> 
> The thing I am missing is that we never delete the pod we created in 2, so we never actually attempt to do the clone. So where are the events coming from?

This is the only way I could create a target name conflict in a deterministic way. 

First event is obviously a source in use: `verifyEvent(controller.SmartCloneSourceInUse, targetNamespaceName, f)`. Then the pod is deleted:
https://github.com/kubevirt/containerized-data-importer/blob/db4df4eb348ca2ea0ee876942b24f5abcfba7c19/tests/cloner_test.go#L246
and 253
```					err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
```

And then clone can proceed.

kgoldbla (approved):
CNV-6612 is covered in this PR. Looks good

awels:
Oh I been reviewing too much, I missed the delete completely. Now it makes a lot more sense.

tests/cloner_test.go
```diff
@@ -208,6 +208,56 @@ var _ = Describe("all clone tests", func() {
...
+
+				targetPvc, err := utils.CreatePVCFromDefinition(f.K8sClient, targetNamespaceName,
+					utils.NewPVCDefinition("target-pvc", "1Gi", annotations, nil))
+				Expect(err).ToNot(HaveOccurred())
+				f.ForceBindIfWaitForFirstConsumer(targetPvc)
+
+				actualCloneType := utils.GetCloneType(f.CdiClient, dataVolume)
+				if actualCloneType == "snapshot" {
+					verifyEvent(controller.SmartCloneSourceInUse, targetNamespaceName, f)
+					err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
```
awels, line 246: this line to line 250 is duplicated on line 253-257. In order to keep with the DRY principle can we do
```go
if actualCloneType == "snapshot" {
  verifyEvent(controller.SmartCloneSourceInUse, targetNamespaceName, f)
} else if actualCloneType == "csiclone" {
  verifyEvent(controller.CSICloneSourceInUse, targetNamespaceName, f)
} else {
  Fail(fmt.Sprintf("Unknown clonetype %s", actualCloneType))
}
err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
//verify event
verifyEvent(controller.ErrResourceExists, targetNamespaceName, f)
```
And as a bonus we now catch if for some reason the clone type gets corrupted.

brybacki: thanks, that is good, applied

brybacki:
@awels so wdyt?

PR #1911, opened 2021-08-31
Avoid trying to get metrics from non-running pod.

mrnold:
This pull request avoids calling updateProgressUsingPod when the importer pod exists but is not running. This prevents a 30-second delay in the phase transition to Paused in multi-stage imports. 

No issue or bug, just noticed this while working on #1903.

**Release note**:
```release-note
Avoid trying to get metrics from non-running pod. 
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -1399,7 +1399,7 @@ func (r *DatavolumeReconciler) reconcileProgressUpdate(datavolume *cdiv1.DataVol
 	if err == nil {
 		if pod.Status.Phase != corev1.PodRunning {
 			// Avoid long timeouts and error traces from HTTP get when pod is already gone
-			return reconcile.Result{}, errors.New("Pod is not running, can't get metrics")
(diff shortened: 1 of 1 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -1397,6 +1397,10 @@ func (r *DatavolumeReconciler) reconcileProgressUpdate(datavolume *cdiv1.DataVol
 	}
 	pod, err := r.getPodFromPvc(podNamespace, pvcUID)
 	if err == nil {
+		if pod.Status.Phase != corev1.PodRunning {
+			// Avoid long timeouts and error traces from HTTP get when pod is already gone
+			return reconcile.Result{}, errors.New("Pod is not running, can't get metrics")
```
awels, line 1402: I would just return reconcile.Result{}, nil instead of an error. We don't want to reconcile again because of this, so returning nil will not trigger an immediate reconcile.

mrnold: Done.

PR #1913, opened 2021-09-01
Support registry import using node docker cache

arnongilboa:

The new CRI (container runtime interface) importer pod is created with three containers and a shared emptyDir volume:
- Init container: copies static http server binary to emptyDir
- Server container: container image container configured to run the http binary and serve up the image file in /data
- Client container: uses cdi-import to import from server container, writes "done" file on emptyDir
- Server container sees "done" file and exits

Thanks **mhenriks** for the PoC!

Done:
- added ImportMethod to DataVolumeSourceRegistry (DataVolume.Spec.Source.Registry, DataImportCron.Spec.Source.Registry). Import method can be "skopeo" (default), or "cri" for container runtime interface based import
- added cdi-containerimage-server & import.sh to the cdi-importer container

ToDo:
- [x] utests and func tests
- [x] doc

Fixes #

utests, func tests and doc will be added soon to this PR.

**Release note**:
```release-note
Add support for CRI (container runtime interface) registry import using node docker cache
```


tools/cdi-containerimage-server/import.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
mhenriks, line 1: this file was only included for the proof of concept.  Should be regular http import container/code path

arnongilboa: It uses regular cdi-importer

pkg/controller/import-controller.go
```diff
@@ -792,6 +820,127 @@ func createImporterPod(log logr.Logger, client client.Client, image, verbose, pu
...
+							MountPath: "/shared",
+							Name:      "shared-volume",
+						},
+					},
+					ImagePullPolicy: corev1.PullPolicy(pullPolicy),
+				},
+				{
+					Name:    common.ImporterPodName,
+					Image:   image,
+					Command: []string{"/usr/bin/import.sh", "8100", "/shared/ready", "/shared/done", "disk.img", "/data/disk.img"},
```
mhenriks, line 907: should configure standard http import container here

tools/cdi-containerimage-server/import.sh
mhenriks: it should use the exact same code path as http import not call this script

mhenriks: I stand corrected just missing some env vars

pkg/controller/import-controller.go
mhenriks: may be nicer, your call

pkg/apis/core/v1beta1/types.go
```diff
@@ -152,14 +152,33 @@ type DataVolumeSourceS3 struct {
 
 // DataVolumeSourceRegistry provides the parameters to create a Data Volume from an registry source
 type DataVolumeSourceRegistry struct {
-	//URL is the url of the Docker registry source
+	//URL is the url of the registry source (starting with docker:// for Docker, or is:// for ImageStream)
```
awels, line 155: Actually we also accept oci-archive https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/transport.go#L88-L92

pkg/apis/core/v1beta1/types.go
```diff
@@ -152,14 +152,33 @@ type DataVolumeSourceS3 struct {
...
+	ImportMethod RegistryImportMethod `json:"importMethod,omitempty"`
 	//SecretRef provides the secret reference needed to access the Registry source
 	SecretRef string `json:"secretRef,omitempty"`
 	//CertConfigMap provides a reference to the Registry certs
 	CertConfigMap string `json:"certConfigMap,omitempty"`
 }
 
+const (
+	// RegistryTransportDocker is docker transport prefix
+	RegistryTransportDocker = "docker"
```
awels, line 167:
```go
// RegistrySchemeDocker is docker scheme prefix
RegistrySchemeDocker = "docker"
// RegistrySchemeOci is oci-archive scheme prefix
RegistrySchemeOci = "oci-archive"
// RegistrySchemeImageStream is image stream scheme prefix
RegistrySchemeImageStream = "is"
```
Please also fix https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/transport.go#L88-L92 to use these constants. Could be a follow up PR.

pkg/apis/core/v1beta1/types.go
```diff
@@ -152,14 +152,33 @@ type DataVolumeSourceS3 struct {
...
+	// RegistryTransportImageStream is image streamstransport prefix
+	RegistryTransportImageStream = "is"
+)
+
+// RegistryImportMethod represents the registry import method
+type RegistryImportMethod string
+
+const (
+	// RegistryImportSkopeo is the standard import method
+	RegistryImportSkopeo RegistryImportMethod = "skopeo"
```
awels, line 177: We don't actually use skopeo anymore, we use the containers library which has the same code that skopeo uses to do its thing, but no skopeo itself. The term used in the library is ImageSource, but I don't know if that is any clearer.

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -240,14 +241,45 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *admission
...
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: fmt.Sprintf("Transport is neither %s or %s in url %s", cdiv1.RegistryTransportDocker, cdiv1.RegistryTransportImageStream, url),
+				Field:   field.Child("source", "Registry", "URL").String(),
+			})
+			return causes
+
+		}
+		importMethod := spec.Source.Registry.ImportMethod
+		if importMethod != "" && importMethod != cdiv1.RegistryImportSkopeo && importMethod != cdiv1.RegistryImportCri {
```
awels, line 275: Don't forget oci-archive, we also support that.

pkg/controller/import-controller.go
```diff
@@ -638,7 +641,9 @@ func (r *ImportReconciler) requiresScratchSpace(pvc *corev1.PersistentVolumeClai
 		case SourceGlance:
 			scratchRequired = true
 		case SourceRegistry:
-			scratchRequired = true
+			if pvc.Annotations[AnnRegistryImportMethod] != "cri" {
```
awels, line 644: can you make "cri" a constant, it is used in a lot of places.

pkg/controller/import-controller.go
```diff
@@ -748,6 +753,19 @@ func getEndpoint(pvc *corev1.PersistentVolumeClaim) (string, error) {
 	return ep, nil
 }
 
+// returns the import image part of the endpoint string
+func getImportImage(pvc *corev1.PersistentVolumeClaim) (string, error) {
+	ep, err := getEndpoint(pvc)
+	if err != nil {
+		return "", nil
+	}
+	parts := strings.Split(ep, "://")
```
awels, line 762: url.Parse uses existing libraries to do the parsing.

pkg/controller/import-controller.go
awels: Might be nice to take how we create the container for a standard importer, and put that into a function, and call that function from both places?

cmd/cdi-importer/BUILD.bazel
```diff
@@ -18,6 +18,7 @@ rpm_image(
             "@nbdkit-xz-filter//file",
             "@nbdkit-gzip-filter//file",
             "@qemu-img-aarch64//file",
+            "@qemu-block-curl-aarch64//file",
```
awels, line 21: Why are we adding block curl, but just for arm?

pkg/apis/core/v1beta1/types.go
arnongilboa: Naming is always the hardest thing...
RegistryImportImageSource?
RegistryImportDocker?
RegistryImportCDI?

pkg/controller/import-controller.go
arnongilboa: Fixed. scratchRequired is sure false in this method, right?

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -240,14 +241,45 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *admission
...
+			sourceType = field.Child("contentType").String()
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: fmt.Sprintf("ContentType must be %s when Source is Registry", cdiv1.DataVolumeKubeVirt),
+				Field:   sourceType,
+			})
+			return causes
+		}
+		url := spec.Source.Registry.URL
+		parts := strings.Split(url, "://")
```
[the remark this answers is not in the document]
arnongilboa: Fixed

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -240,14 +241,45 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *admission
...
+				Message: fmt.Sprintf("Illegal registry source URL %s", url),
+				Field:   field.Child("source", "Registry", "URL").String(),
+			})
+			return causes
+		}
+		transport := parts[0]
+		if transport != cdiv1.RegistryTransportDocker && transport != cdiv1.RegistryTransportImageStream {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: fmt.Sprintf("Transport is neither %s or %s in url %s", cdiv1.RegistryTransportDocker, cdiv1.RegistryTransportImageStream, url),
```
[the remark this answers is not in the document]
arnongilboa: Fixed

pkg/apis/core/v1beta1/types.go
arnongilboa: Fixed

pkg/apis/core/v1beta1/types.go
arnongilboa: Fixed

pkg/apiserver/webhooks/datavolume-validate.go
arnongilboa: oci-archive was added to the scheme validation, not importMethod.

pkg/controller/import-controller.go
arnongilboa: Fixed

pkg/controller/import-controller.go
arnongilboa: Fixed

cmd/cdi-importer/BUILD.bazel
arnongilboa: Leftover, removed.

pkg/apiserver/webhooks/datavolume-validate.go
awels: Actually good point, its either use the container runtime (docker cache) or use pulling from registry

pkg/controller/import-controller.go
awels: you can write this as:
```go
  scratchRequired = pvc.Annotations[AnnRegistryImportMethod] != "cri"
```
With the "cri" constant of course.

arnongilboa: Right, fixed.

pkg/controller/import-controller.go
arnongilboa: Done

tools/cdi-containerimage-server/import.sh
arnongilboa: Done

tools/cdi-containerimage-server/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
```
mhenriks, line 1: not sure how to do this in bazel off the top of my head but we should make sure that the binary has no links to c libraries.  In the POC, I set `CGO_ENABLED=0` on `go build` line.

tools/cdi-containerimage-server/main.go
```diff
@@ -0,0 +1,93 @@
...
+	server := &http.Server{
+		Handler: http.FileServer(http.Dir(*directory)),
+	}
+	addr := fmt.Sprintf("localhost:%d", *port)
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		log.Fatalf("Failed listening on %s err: %v", addr, err)
+	}
+
+	f, err := os.OpenFile(*readyFile, os.O_CREATE|os.O_EXCL, 0666)
```
mhenriks, line 67: maybe shouldn't have `O_EXCL` here in case of crash/restart.  I think some thought has to go into this synchronization and crash/restart scenarios.  Perhaps the easiest thing to do would be to not restart containers and have the import delete/recreate the pod on failure?

mhenriks (commented):
food for thought

pkg/apis/core/v1beta1/types.go
```diff
@@ -152,14 +152,35 @@ type DataVolumeSourceS3 struct {
...
 	CertConfigMap string `json:"certConfigMap,omitempty"`
 }
 
+const (
+	// RegistrySchemeDocker is docker scheme prefix
+	RegistrySchemeDocker = "docker"
+	// RegistrySchemeOci is oci-archive scheme prefix
+	RegistrySchemeOci = "oci-archive"
+	// RegistrySchemeImageStream is image stream scheme prefix
+	RegistrySchemeImageStream = "is"
```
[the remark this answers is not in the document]
arnongilboa: I like the longest alternative: `image-stream://`
@aglitke @mhenriks @awels what's your favorite? 
`imagestream:// imgstrm:// istream:// ?`

cmd/cdi-importer/importer.go
```diff
@@ -39,6 +40,32 @@ func init() {
...
+	if readyFile == "" {
+		return
+	}
+	for {
+		if _, err := os.Stat(readyFile); err == nil {
+			break
+		}
+		time.Sleep(time.Second)
+	}
+	os.Remove(readyFile)
```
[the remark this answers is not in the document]
arnongilboa: Right, this is a dumb line. The  server already defers the file removal, so it won't stay there if the server crashed and before it's ready again.

pkg/controller/import-controller.go
```diff
@@ -792,6 +819,113 @@ func createImporterPod(log logr.Logger, client client.Client, image, verbose, pu
 	return pod, nil
 }
 
+// makeCriImporterPodSpec creates and return the container runtime interface importer pod spec based on the passed-in importImage and pvc.
+func makeCriImporterPodSpec(namespace, image, importImage, verbose, pullPolicy string, podEnvVar *importPodEnvVar, pvc *corev1.PersistentVolumeClaim, podResourceRequirements *corev1.ResourceRequirements, workloadNodePlacement *sdkapi.NodePlacement, priorityClassName string) *corev1.Pod {
```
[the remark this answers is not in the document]
arnongilboa: Actually `makeCriImporterPodSpec()` is  similar to the long existing `makeImporterPodSpec()`. So time has come to pass a struct to both of them?

tools/cdi-containerimage-server/BUILD.bazel
arnongilboa: It's sure already done in the previous patch using `pure = "on"` which is equivalent to `CGO_ENABLED=0`.

tools/cdi-containerimage-server/main.go
arnongilboa: Do you really want to move the delete/recreate to pod level by the import-controller, so it adds a specific recovery path to the cri import? I think our server and importer containers are simple enough to do the recovery in the container level. "ready" file removal is deferred by the server, so I see no reason to remove the O_EXCL.

cmd/cdi-importer/importer.go
```diff
@@ -186,7 +213,9 @@ func main() {
 		}
 		defer dp.Close()
 		processor := importer.NewDataProcessor(dp, dest, dataDir, common.ScratchDataDir, imageSize, filesystemOverhead, preallocation)
+		waitForReadyFile()
 		err = processor.ProcessData()
+		touchDoneFile()
```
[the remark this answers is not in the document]
arnongilboa: Mmm, in order to get the importer container restarted and the server container keep serving I guess we shouldn't  `touchDoneFile()` in case of error.

pkg/controller/import-controller.go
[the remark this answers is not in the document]
arnongilboa: Done

tools/cdi-containerimage-server/main.go
mhenriks: I just think we should make sure that we can tolerate container crashes.  If you can make it work without help from the controller, then that's great.  My point was that it may be easier to do that in one place.  

The reason to remove `O_EXCL` is that it is an error if the file already exists and that seems possible after a crash/restart.

mhenriks: Ah now I see the defer removal

arnongilboa: So we can trust the defer removal that when server container crashes the "ready" file is removed, and we can keep the `O_EXCL`. What bothers me more are the loops waiting for the ready/done files. I guess we should error after some time? for the `ready` it's easy (say 10sec?), but for the `done` it depends on the file size.

pkg/apis/core/v1beta1/types.go
aglitke: Ack.  I like the naming suggested here.  PullMethod seems better since you could use either method against the same image source.  node and pod sound good to me.

tests/transport_test.go
```diff
@@ -140,23 +143,26 @@ var _ = Describe("Transport Tests", func() {
 	registryNoAuthEp := func() string { return fmt.Sprintf("docker://%s", utils.RegistryHostName+"."+f.CdiInstallNs) }
 	registryAuthEp := func() string { return fmt.Sprintf("docker://%s.%s:%d", utils.RegistryHostName, f.CdiInstallNs, 1443) }
 	altRegistryNoAuthEp := func() string { return fmt.Sprintf("docker://%s.%s:%d", utils.RegistryHostName, f.CdiInstallNs, 5000) }
+	trustedCriRegistryEp := func() string { return "docker://registry:5000" }
```
mhenriks, line 146: This should be configurable.  Not exactly sure the best way to do it for func tests but `DOCKER_PREFIX` is the env var used for build/sync.

pkg/apis/core/v1beta1/types.go
```diff
@@ -152,14 +152,35 @@ type DataVolumeSourceS3 struct {
 
 // DataVolumeSourceRegistry provides the parameters to create a Data Volume from an registry source
 type DataVolumeSourceRegistry struct {
-	//URL is the url of the Docker registry source
+	//URL is the url of the registry source (starting with the scheme: docker, oci-archive, imagestream)
 	URL string `json:"url"`
+	//PullMethod can be either "pod" (default import), or "node" (node docker cache based import)
+	PullMethod RegistryPullMethod `json:"pullMethod,omitempty"`
```
mhenriks, line 158: This should probably be a pointer if optional

pkg/apis/core/v1beta1/types.go
```diff
@@ -152,14 +152,35 @@ type DataVolumeSourceS3 struct {
...
 	CertConfigMap string `json:"certConfigMap,omitempty"`
 }
 
+const (
+	// RegistrySchemeDocker is docker scheme prefix
+	RegistrySchemeDocker = "docker"
+	// RegistrySchemeOci is oci-archive scheme prefix
+	RegistrySchemeOci = "oci-archive"
+	// RegistrySchemeImageStream is image stream scheme prefix
+	RegistrySchemeImageStream = "imagestream"
```
mhenriks, line 171: I don't see any actual support for this yet in here (not adding annotations to import pod).  What is the plan?

arnongilboa: Yep, it should be moved to the controller PR and somehow got here. Will move.

pkg/apis/core/v1beta1/types.go
arnongilboa: The following 2 fields (SecretRef, CertConfigMap) are also optional but not pointers, so why is it important here?

mhenriks: IMO the others should be pointers as well not sure why they're not.  Fine if you want to keep non pointers for consistency for now.  Can always change to pointers while maintaining API compatibility

pkg/apis/core/v1beta1/types.go
mhenriks: Given recent conversations, should `RegistrySchemeImageStream` be deleted?

arnongilboa: Sure, working on the IS change it this morning.

pkg/apis/core/v1beta1/types.go
arnongilboa: Fixed

tests/transport_test.go
arnongilboa: Done.

pkg/apis/core/v1beta1/types.go
arnongilboa: Done

pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
arnongilboa: Removed imagestream scheme

pkg/controller/import-controller.go
```diff
@@ -771,31 +843,160 @@ func createImportPodNameFromPvc(pvc *corev1.PersistentVolumeClaim) string {
...
+	pod := &corev1.Pod{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Pod",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: args.pvc.Namespace,
+			Annotations: map[string]string{
+				AnnCreatedBy: "yes",
```
mhenriks, line 913: Don't see `alpha.image.policy.openshift.io/resolve-names: '*'` added anywhere if imagestream

pkg/controller/import-controller.go
```diff
@@ -771,31 +843,160 @@ func createImportPodNameFromPvc(pvc *corev1.PersistentVolumeClaim) string {
...
+	}
+
+	/**
+	FIXME: When registry source is ImageStream, if we set importer pod OwnerReference (to its pvc, like all other cases),
+	for some reason (OCP issue?) we get the following error:
+		Failed to pull image "imagestream-name": rpc error: code = Unknown
+		desc = Error reading manifest latest in docker.io/library/imagestream-name: errors:
+		denied: requested access to the resource is denied
+		unauthorized: authentication required
+	When we don't set pod OwnerReferences, all works well.
```
mhenriks, line 967: see where you delete imagestream finalizer but not add it

mhenriks (commented):
Getting there!  Couple comments.

Looking forward to imagestream func tests and docs

awels (commented):
Looks pretty good, just some nits.

pkg/controller/import-controller.go
arnongilboa: This is not needed when you set `Imagestream.spec.lookupPolicy.local: true`. I initially suspected this was the reason for the  OwnerReference issue, but adding it changed nothing.

pkg/controller/import-controller.go
arnongilboa: We add it [here](https://github.com/kubevirt/containerized-data-importer/pull/1913/commits/b942c8d6bc9c7bad4cc27c41769838b55ee13cd7#diff-868af4b730094e9180f78af4cc069ae6cecf7077a46c186553f331dccbd32993R515-R523)

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -68,12 +68,90 @@ var _ = Describe("Validating Webhook", func() {
...
+		It("should accept DataVolume with Registry source ImageStream and node PullMethod on create", func() {
+			imageStream := "istream"
+			pullNode := cdiv1.RegistryPullNode
+			registrySource := cdiv1.DataVolumeSource{
+				Registry: &cdiv1.DataVolumeSourceRegistry{ImageStream: &imageStream, PullMethod: &pullNode},
+			}
+			pvc := newPVCSpec(pvcSizeDefault)
+			dataVolume := newDataVolume("testDV", registrySource, pvc)
+			resp := validateDataVolumeCreate(dataVolume)
+			Expect(resp.Allowed).To(Equal(true))
```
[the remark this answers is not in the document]
arnongilboa: I'd rather do this in the same PR with the above refactoring, as all tests here use Equal(true/false).

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -68,12 +68,90 @@ var _ = Describe("Validating Webhook", func() {
...
 		})
 
-		It("should accept DataVolume with Registry source on create", func() {
+		It("should accept DataVolume with Registry source URL on create", func() {
 			dataVolume := newRegistryDataVolume("testDV", "docker://registry:5000/test")
 			resp := validateDataVolumeCreate(dataVolume)
 			Expect(resp.Allowed).To(Equal(true))
 		})
 
+		It("should accept DataVolume with Registry source ImageStream and node PullMethod on create", func() {
```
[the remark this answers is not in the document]
arnongilboa: I'd rather do this nice refactoring in a different PR.

pkg/controller/import-controller.go
mhenriks: Okay but is it reasonable to expect that to be set?  Will adding the annotations harm anything?

Issue #1414, opened 2020-10-02
Feature Request: allow configuration of qemu-img convert's cache type option (-t)


kenshaw:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind enhancement

**What happened**:

The `cdi-importer` command (from code in `pkg/image/qemu.go`) hard codes `qemu-img import`'s cache option to `-t none`. On a bare metal cluster, using NFSv4.1, this causes `qemu-img` to wait for verification of the written data. For large imported images, this can take an extraordinary (and frustrating!) amount of time, wildly disproportionate to the expected speed of the underlying networks/nodes/storage (10Gbit/EPYC 32core/ZFS raid5, capable 500+ MB/s throughput).

 As per a discussion on the K8s #virtualization channel with Mike Henriksen, the option was hardcoded/added [via this PR](https://github.com/kubevirt/containerized-data-importer/pull/1102). Per Mike, this was in order to prevent issues with the imported images not being fully written prior to the termination of the importer pod.

In my (humble) opinion, this should be configurable via the CDI deployment, as the CDI user should be able to choose the risk profile they want their cluster to operate with.

Using the bare metal cluster my company has, I did a couple of tests to demonstrate the difference in speed between the various options. From a running importer instance, I grabbed the generated `qemu-img` run command, which was the following: 

```sh
qemu-img convert -t none -p -O raw json: {"file.driver": "http", "file.url": "http://[REDACTED]/cloud-images/focal-server-cloudimg-amd64.img", "file.timeout": 3600} /data/disk.img
````

(note that the [REDACTED] URL above is a local HTTP mirror on the same 10Gbit network)

Then, using this pod definition:

```yaml
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: cdi-test
spec:
  storageClassName: nfs-client
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 100Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: cdi-test
spec:
  restartPolicy: Never
  containers:
  - name: cdi-test
    image: kubevirt/cdi-importer:v1.23.5
    imagePullPolicy: IfNotPresent
    command: ["tail", "-f", "/dev/null"]
    volumeMounts:
    - name: cdi-test
      mountPath: /cdi-test
  volumes:
    - name: cdi-test
      persistentVolumeClaim:
        claimName: cdi-test
```

[...]

PR #1919, opened 2021-09-02
Stop qemu-img writing with cache=none

akalenyu:

Since #1102 our `qemu-img convert` writes use the "none" cache mode;
This is because with [default](https://github.com/qemu/qemu/blob/master/qemu-img.c#L151) "unsafe" cache mode its possible that the process can exit before all writes have
reached storage (lazily flushed). Also note "unsafe" [ignores flushes](https://documentation.suse.com/sles/11-SP4/html/SLES-kvm4zseries/cha-qemu-cachemodes.html#id-1.10.6.6.5).
This might be an expensive approach to fix the problem since "none" bypasses the host cache.

We can instead tackle this situation differently by sticking to "writeback" cache mode,
along with ensuring the write was committed (fsync syscall).

Looks like we were already leaning in this direction in the past:
https://github.com/kubevirt/containerized-data-importer/issues/1414#issuecomment-706992460

Fixes #

**Release note**:

```release-note
BugFix: our qemu-img convert cmd writes with cache=none (bypassing cache)
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -113,13 +113,15 @@ func NewQEMUOperations() QEMUOperations {
 }
 
 func convertToRaw(src, dest string, preallocate bool) error {
-	args := []string{"convert", "-t", "none", "-p", "-O", "raw", src, dest}
+	args := []string{"convert", "-t", "writeback", "-p", "-O", "raw", src, dest}
 	var err error
+
 	if preallocate {
 		err = addPreallocation(args, convertPreallocationMethods, func(args []string) ([]byte, error) {
 			return qemuExecFunction(nil, reportProgress, "qemu-img", args...)
 		})
 	} else {
+		klog.V(3).Infof("Running qemu-img convert with args: %v", args)
 		_, err = qemuExecFunction(nil, reportProgress, "qemu-img", args...)
 	}
 	if err != nil {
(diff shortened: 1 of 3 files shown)


akalenyu:
Changed PR to "Draft" as it appears we can't actually pass a specific block size to `qemu-img convert`:
https://bugzilla.redhat.com/show_bug.cgi?id=2000791#c1

pkg/image/qemu.go
```diff
@@ -131,6 +135,18 @@ func convertToRaw(src, dest string, preallocate bool) error {
 		return errors.Wrap(err, errorMsg)
 	}
 
+	// With writeback cache mode it's possible that the process will exit before all writes have been commited to storage.
+	// To guarantee that our write was commited to storage, we make a syncfs syscall and ensure success.
+	file, err := os.Open(dest)
+	if err != nil {
+		return errors.Wrap(err, "could not get file descriptor for syncfs call following qemu-img writing")
+	}
+	if err := unix.Syncfs(int(file.Fd())); err != nil {
```
maya-r, line 144: I read [the syncfs documentation](https://linux.die.net/man/2/syncfs) and I am not sure it does the right thing for a block device. Perhaps this should be fsync instead

akalenyu: I should have probably added this in the PR description, but here is my reasoning for not going with fsync:
https://blogs.gnome.org/cneumair/2006/02/11/ioctl-fsync-how-to-flush-block-device-buffers/#comment-91
And
```
According to the standard specification (e.g., POSIX.1-2001),
       sync() schedules the writes, but may return before the actual
       writing is done.  However Linux waits for I/O completions, and
       thus sync() or syncfs() provide the same guarantees as fsync()
       called on every file in the system or filesystem respectively.
```
From https://man7.org/linux/man-pages/man2/sync.2.html "Notes" section

maya-r: So in the blog post the author had, say, /dev/sda mounted at /mnt.
And fsync(/dev/sda) did not sync the pending filesystem writes to /mnt.
This isn't the case for us, we don't have a filesystem mounted at the same time. We are writing directly to /dev/sda.

Also, just stating the same comment more clearly:
syncfs(/mnt/disk.img) will sync the entire /mnt filesystem - good
syncfs(/dev/sda) will sync the /dev filesystem - not the block device - bad

akalenyu: Got it, thank you
Updated the PR to fsync

aglitke (commented):
looks good to me

PR #1922, opened 2021-09-02
Add long term token to pvcs when host assisted cloning cross namespaces

mhenriks:
See bug:  https://bugzilla.redhat.com/show_bug.cgi?id=1855182

When dumb cloning across namespaces, and a source in use (pod mounting source PVC) then it is possible for the token to time out (5 min).  To address that, the datavolume controller will add a "forever" token to the PVC.  This should be safe because the forever token includes the pvc UID.

Fixes #

**Release note**:

```release-note
BufFix: BZ 1855182 - clone token timing out
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -466,13 +466,14 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 			return reconcile.Result{}, err
 		}
 	case CsiClone:
-		if pvc.Status.Phase == corev1.ClaimBound {
+		switch pvc.Status.Phase {
+		case corev1.ClaimBound:
 			if err := r.setCloneOfOnPvc(pvc); err != nil {
 				return reconcile.Result{}, err
 			}
-		} else if pvc.Status.Phase == corev1.ClaimPending {
+		case corev1.ClaimPending:
 			return reconcile.Result{}, r.updateCloneStatusPhase(cdiv1.CSICloneInProgress, datavolume, pvc, selectedCloneStrategy)
-		} else if pvc.Status.Phase == corev1.ClaimLost {
+		case corev1.ClaimLost:
 			return reconcile.Result{},
 				r.updateDataVolumeStatusPhaseWithEvent(cdiv1.Failed, datavolume, pvc, selectedCloneStrategy,
 					DataVolumeEvent{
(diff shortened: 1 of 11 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -455,25 +459,29 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
-					r.updateDataVolumeStatusPhaseWithEvent(cdiv1.Failed, datavolume, pvc, selectedCloneStrategy,
-						DataVolumeEvent{
-							eventType: corev1.EventTypeWarning,
-							reason:    ErrClaimLost,
-							message:   fmt.Sprintf(MessageErrClaimLost, pvc.Name),
-						})
-			}
+	switch selectedCloneStrategy {
+	case HostAssistedClone:
+		if err := r.ensureExtendedToken(pvc); err != nil {
```
awels, line 464: Does this address the case where we start an import, then immediately start a clone of that import, and the import takes longer than the token is valid, so when we finally get to the clone part, it fails due to an expired token?

mhenriks: Yeah that's basically the whole purpose of this PR.  The once the extended token is added to the target PVC (right after it's created), we've got 10 years to start the clone source pod

awels: So why only one host assisted clone? Doesn't the same issue apply to csi/smart clone?

mhenriks: Object transfer is created at the start of the operation after the short token is validated.

pkg/controller/datavolume-controller.go
```diff
@@ -455,25 +459,29 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+		} else if pvc.Status.Phase == corev1.ClaimLost {
+			return reconcile.Result{},
+				r.updateDataVolumeStatusPhaseWithEvent(cdiv1.Failed, datavolume, pvc, selectedCloneStrategy,
+					DataVolumeEvent{
+						eventType: corev1.EventTypeWarning,
+						reason:    ErrClaimLost,
+						message:   fmt.Sprintf(MessageErrClaimLost, pvc.Name),
+					})
+		}
+		fallthrough
```
brybacki, line 483: explicit fallthrough, nice, 
looks better than previous if else is

mhenriks: maybe should change embedded `if/else if` to `switch pvc.Status.Phase` as well for consistency

awels (approved):
/lgtm

PR #1932, opened 2021-09-13
Update vendor and move to k8s.io/kube-openapi

kwoodson:
This PR updates the go.mod dependent libraries as well as moves from go-openapi to kube-openapi. This is required by other dependent libraries to keep up with the latest library updates.

This PR attempts to update https://github.com/kubevirt/containerized-data-importer/blob/main/go.mod#L29 which has dependencies that include outdated dependencies that do not exist.

Fixes #

This is an attempt to update libraries and move to k8s.io/kube-openapi

**Release note**:
```release-note
Updating libraries and dependencies.
```


kwoodson:
@maya-r Taking a look at the failures, it appears they are failing because of the golang version of the builds:
```
Starting bazel server
go version go1.14.6 linux/amd64
env OPERATOR_DIR=/root/go/src/kubevirt.io/containerized-data-importer go test -v -coverprofile=.coverprofile -test.timeout 180m  ./pkg/... ./cmd/... 
vendor/k8s.io/client-go/plugin/pkg/client/auth/exec/metrics.go:21:2: cannot find package "io/fs" in any of:
	/root/go/src/kubevirt.io/containerized-data-importer/vendor/io/fs (vendor tree)
	/gimme/.gimme/versions/go1.14.6.linux.amd64/src/io/fs (from $GOROOT)
```

mhenriks:
@kwoodson a little more background info would be helpful.  Is this all just for:  https://github.com/kubevirt/containerized-data-importer/blob/main/go.mod#L29?

In what project are you trying to import cdi?

And updating the go version is fine but should ideally be done in a separate pr because it changes the build container.

On more note.  Would *really* like to stay at current version of k8s libraries if possible to stay in sync with kubevirt.  So please have a good reason for updating them.

kwoodson:
@mhenriks Thanks for taking a look. I am relatively new at attempting to update go mod dependencies and this was my best attempt to resolve these conflicts and outdated modules. If this is not the best way forward, please advise and I can take another path.

I am working on updating the https://github.com/openshift/installer. The OpenShift installer has some pretty dated dependencies and as well as the API recently moved to the kube 0.22.1 modules. With these recent updates and in an attempt to resolve some of the existing dependencies on kubevirt's client-go and containerized-data-importer I(https://github.com/openshift/installer/blob/master/go.mod#L118-L119) I arrived with this current PR. I understand these dependencies are required by the terraform-provider-kubevirt.

My next stop was the kubevirt/kubevirt repository to update the kubevirt/client-go to the latest kube libraries. I was hoping this might resolve our dependency issues. 

Again, thanks for your time and please let me know if I'm barking up the wrong tree.

mhenriks:
@kwoodson thanks for the background info!

It is almost impossible to import `kubevirt/client-go`.  Even worse than CDI!  @rmohr and I agreed that it would be wise to create `api` repos in the kubevirt org for KubeVirt and CDI.  Similar to `k8s.io/api`.  That way it is easy to import the api types without bringing in a bunch of crap you don't care about.  From there, you can use the controller runtime client directly on them (I see openshift installer uses controller runtime).

Given the difficult road ahead, I'd advise waiting for the new api repos if possible.  But if you want to go ahead anyway I'd advise starting with updating the go version in the build container.

PR #1937, opened 2021-09-14
Commit write to disk in source-pvc filler pod

akalenyu:

Addresses a flake where the clone target DV is "Succeeded" and empty:
```bash
[WARN] error executing command ["/bin/sh" "-c" "md5sum /pvc/disk.img"], error: command terminated with exit code 1
INFO: stderr: [md5sum: /pvc/disk.img: No such file or directory]
```
What I believe happens here is that the filler pod is not committing writes and directly exiting after the `echo 1245 > disk.img`,
resulting in us cloning an empty pvc and doing a successfull job at it.
Did not see it fail at all with the `&& sync` addition.

Fixes #

**Release note**:

```release-note
NONE
```


mhenriks:
filesystem should be synced when unmounted so should be happening here but obv this isn't going to hurt

akalenyu:
> filesystem should be synced when unmounted so should be happening here but obv this isn't going to hurt

I only saw this on the ceph-rbd lane with Filesystem PVCs btw

PR #1938, opened 2021-09-14
Pause MCPs before proxy tests to prevent unnecessary reboot

akalenyu:

Changes on the cluster wide proxy object in OpenShift trigger a reboot that causes some havoc in our runs:
https://docs.openshift.com/container-platform/4.8/support/troubleshooting/troubleshooting-operator-issues.html#troubleshooting-disabling-autoreboot-mco_troubleshooting-operator-issues
Note that in our much simpler tests we don't really need this reboot to ensure we proxy things - we are not testing the OpenShift proxy.

Thanks @dollierp.

Fixes #

Had trouble importing the machine-config-operator packages, so had to hack it out:
https://github.com/openshift/machine-config-operator/issues/848#issuecomment-609385205
I would be happy to do it cleaner but not sure how to overcome `malformed file path with :` err.

**Release note**:

```release-note

```


tests/import_proxy_test.go
```diff
@@ -359,3 +369,24 @@ func cleanClusterWideProxy(ocpClient *configclient.Clientset, clusterWideProxySp
 		return false
 	}, timeout, pollingInterval).Should(BeTrue())
 }
+
+func manipulateMachineConfigPools(mcfgClient *mcfgclient.Clientset, paused bool) {
+	workerMCP, err := mcfgClient.MachineconfigurationV1().MachineConfigPools().Get(context.TODO(), "worker", metav1.GetOptions{})
+	Expect(err).ToNot(HaveOccurred())
+	masterMCP, err := mcfgClient.MachineconfigurationV1().MachineConfigPools().Get(context.TODO(), "master", metav1.GetOptions{})
```
dollierp, line 376: IIUC there are discussions to rename `master` MCP to `control-plane` due to the concerns with non inclusive language.
Maybe we should loop over all MCP instead of hardcoding their names.

tests/import_proxy_test.go
```diff
@@ -90,6 +97,9 @@ var _ = Describe("Import Proxy tests", func() {
 		if utils.IsOpenshift(f.K8sClient) {
 			By("Reverting the cluster wide proxy spec to the original configuration")
 			cleanClusterWideProxy(ocpClient, clusterWideProxySpec)
+
+			By("Unpausing MCPs")
```
dollierp, line 101: I think the right word is resuming.

awels:
If we don't want to include a bunch of openshift specific schemes, maybe we can do something like this? https://github.com/kubevirt/containerized-data-importer/blob/main/tests/utils.go#L49-L76 which is basically calling kubectl from the command line. We should be able to do the same with oc.

akalenyu:
> If we don't want to include a bunch of openshift specific schemes, maybe we can do something like this? https://github.com/kubevirt/containerized-data-importer/blob/main/tests/utils.go#L49-L76 which is basically calling kubectl from the command line. We should be able to do the same with oc.

Yep definitely better, done

So there is a problem here; when verifying I noticed that after a test, the MCP thinks maintenance is still needed (even though we reverted the changes already):
```
If there are pending changes (where both the Updated and Updating columns are False), it is 
recommended to schedule a maintenance window for a reboot as early as possible.
Use the following steps for unpausing the autoreboot process to apply the changes that were queued since the last reboot.
```
From [docs](https://docs.openshift.com/container-platform/4.8/support/troubleshooting/troubleshooting-operator-issues.html#troubleshooting-disabling-autoreboot-mco_troubleshooting-operator-issues)

So basically this PR still does not solve the problem as the maintenance will still take place during our run.

- Not sure if this should be a bug in OpenShift? no need for maintenance if change reverted?
- If it is a bug, how else can we tackle this meanwhile?

awels:
So the only non hokey thing I can think of is to accept we will have node reboots, and in an AfterEach of the test, wait for the nodes to all finish rebooting, you should be able to tell if everything is done by the status of the MCP. If we have multiple tests that manipulate the cluster wide proxy, we should put those tests in a context, and wait in the after each of that context.

akalenyu:
> So the only non hokey thing I can think of is to accept we will have node reboots, and in an AfterEach of the test, wait for the nodes to all finish rebooting, you should be able to tell if everything is done by the status of the MCP. If we have multiple tests that manipulate the cluster wide proxy, we should put those tests in a context, and wait in the after each of that context.

So I added this check for MCPs to settle and went on to verify;
on BM, this took 45~ minutes, and since we have it in AfterEach its going to be very costly :x

I don't think we should see reboots when MCPs are paused, discussion ongoing in bug:
https://bugzilla.redhat.com/show_bug.cgi?id=2005694

tests/import_proxy_test.go
```diff
@@ -90,6 +94,18 @@ var _ = Describe("Import Proxy tests", func() {
...
+			manipulateMachineConfigPools(f, false)
+
+			By("Finish node reboots")
+			// Wait a bit for the reboot to start so we don't succeed immediately
+			time.Sleep(2 * time.Minute)
+			Eventually(func() error {
+				output, err := RunOcCommand(f, "wait", "mcp", "--all", "--for", "condition=updated", "--timeout", "1s")
+				By(output)
+				return err
+			}, 50*time.Minute, time.Second).ShouldNot(HaveOccurred())
```
akalenyu, line 108: Just making sure this is seen, and keep in mind this will run for every test in the import proxy suite (potentially 5 hours).

awels: Right, but this will only be on openshift clusters.

PR #1949, opened 2021-09-19
Add DataImportCron controller

arnongilboa:
- The new controller polls for updates to a registry source container image, based on a given schedule. When updates to a container image are detected, the controller imports the content into a new uniquely named PVC in a golden image namespace.
- For each DataImportCron, the controller manages a corresponding DataSource to always point to the latest most up-to-date golden image PVC.
- DataImportCron takes ownership of an existing DataSource (with controller: false), allowing an admin to opt-in to using auto delivery/updates later on.
- The controller has PVC garbage collector removing old PVCs. 

**Post-merge ToDos:**
- Complete condition updates
- Replace batch/v1beta1 CronJob (deprecated in v1.21+, unavailable in v1.25+) with batch/v1 CronJob, which is currently missing
- Add ImageStreams functional tests on OpenShift cluster
- Remove external registry url from functional test
- Instead of hardcoded sha values in the tests, consider looking up the with a pod at the beginning of the test
- Support concurrent imports
- Consider garbage collection regardless of import success
- Consider moving annotations to status members

Fixes #

- utests, func tests and doc will be added soon to this PR.

**Release note**:
```release-note
Add DataImportCron controller for polling and updating golden image PVCs
```


pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,532 @@
...
+	}
+	return nil
+}
+
+func newCronJob(cron *cdiv1.DataImportCron) *v1beta1.CronJob {
+	var container corev1.Container
+	if cron.Spec.Source.Registry.URL != nil {
+		container = corev1.Container{
+			Name:  "skopeo",
+			Image: "quay.io/skopeo/stable",
```
awels, line 414: Can't hard code images like this. People might want to use a different image for this.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,532 @@
...
+			Command: []string{"/usr/bin/sh", "-c",
+				"skopeo inspect " + *cron.Spec.Source.Registry.URL +
+					" | awk -F'\"' '/Digest/{print $4}' > /dev/termination-log",
+			},
+			ImagePullPolicy: corev1.PullIfNotPresent,
+		}
+	} else if cron.Spec.Source.Registry.ImageStream != nil {
+		container = corev1.Container{
+			Name:  "busybox",
+			Image: "busybox",
```
awels, line 424: I am guessing this will come from dockerhub, which will start to generate random failures due to the rate limit. Again don't hard code image names.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,532 @@
...
+				return nil, err
+			}
+			return cronJob, nil
+		}
+		return nil, err
+	}
+	return cronJob, nil
+}
+
+func (r *DataImportCronReconciler) getDigest(dataImportCron *cdiv1.DataImportCron, cronJob *v1beta1.CronJob) (string, error) {
```
awels, line 119: So let me see if I understand this code correctly:
1. You check if the job is active.
2. If it is active you put the jobname on the on the cron object
3. You check if the digest is set on the job
4. If it is set, then you return a blank string for the result of getDigest?
5. You look up all the pods owned by the job
6. If there is one pod, then you look up the termination mesage
7. The termination message is either a node or skopeo digest
8. You set the found digest on the cron job
9. You return the digest

How are you reliably going to find the pods. If the digest doesn't take long to obtain the pods might be long gone before you try to read them. Or the job might no longer be active (since it finished) and you won't even try to find the pods.
Wouldn't it make much more sense to have a controller that watches for pods that have an ownerRef of a cron job, and in the reconcile loop of that controller look up the pod (we know it will exist since that triggered the reconcile) and look up the owning cron job. And do the digest update in that controller.

And in this controller, just check for the annotation, and if it is there do the logic for importing if it is different from the current digest. You are doing too many things at once in this controller IMO.

awels (commented):
So I think we need to break up this giant controller into several smaller ones, each with its own purpose.
1. ImageStream controller. Its only purpose is to watch imagestreams, if one is updated, it updates the digest on the dataimport cron object.
2. DataImportCronJob controller. Its only purpose is to create and manage cron jobs, it doesn't care about digests or anything like that, just set up the cron jobs.
3. The 'work' controller, this controller is not always running, the cron job actually creates this controller. This controller checks the digest on the dataimport cron object, if they don't match starts an import. If they do match and are image stream, exists, if it is not an image stream start a skopeo pod to get the digest, and compare, then start import.

Right now a single controller is trying to do this all reliably and I don't that will work.

arnongilboa:
> So I think we need to break up this giant controller into several smaller ones, each with its own purpose.

Giant? 500 LOC means it's the smallest controller in CDI :)
I'll sure go over your suggested split and see if it makes sense.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,532 @@
...
+	"sort"
+	"strings"
+
+	"github.com/go-logr/logr"
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/pkg/errors"
+
+	batchv1 "k8s.io/api/batch/v1"
+	//FIXME: use batchv1 or v2alpha1
+	v1beta1 "k8s.io/api/batch/v1beta1"
```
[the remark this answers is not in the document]
arnongilboa: AFAIR CronJob is supported only in v1beta1, but I'll sure check that (that's reason for the FIXME)

pkg/controller/dataimportcron-controller.go
arnongilboa: Sure

pkg/controller/dataimportcron-controller.go
arnongilboa: Will use const instead. Any better minimal image?

awels:
Well when I say giant, it is trying to do too many things at once. Note the 3rd controller might be simpler not being a controller at all. But just a k8s application:
1. Get dataimport cronjob object
2. Compare digests
3. If they match, and are not image stream, look up with skopeo
4. If they don't match start import
6. Update dataimport cron job if needed (with new digest)
7. exit

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
awels: According to documentation v1 is supported since k8s 1.19. Our client is 1.21 so it should be supported.

pkg/controller/dataimportcron-controller.go
awels: Well if you split this up, we will be making an image that does the actual import. so the passing of the image might not be an issue.

arnongilboa:
> Well when I say giant, it is trying to do too many things at once. 
Ok, seems nice, just note the controller is based on the design doc we reviewed together.

awels:
I missed you had put all the functions in the controller design doc. The functions change during implementation so I wasn't too concerned about the exact function names.

pkg/controller/dataimportcron-controller.go
mhenriks: But I tend to agree with @awels that it would be better to not launch a job pod for imagestreams

mhenriks (commented):
As-is I have concerns about always being able to catch the "active" cron job to get a hash.

I agree with @awels that it would be preferable to break down/isolate imagestream and registry specific code.

Would be great if can rebase on current main

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
arnongilboa: @awels no idea what doc says, but [v1](https://github.com/kubevirt/containerized-data-importer/blob/main/vendor/k8s.io/api/batch/v1/types.go) does not have CronJob. Better use [v2alpha1](https://github.com/kubevirt/containerized-data-importer/blob/main/vendor/k8s.io/api/batch/v2alpha1/types.go), right?

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,532 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// DataImportCronReconciler members
+type DataImportCronReconciler struct {
+	client         client.Client
+	uncachedClient client.Client
```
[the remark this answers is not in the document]
arnongilboa: When `garbageCollectOldImports()` used `client` to list the DVs it returned inconsistent result, but `uncachedClient` always returns the complete list. Does it make sense?

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
awels: Did you try to use v1 and run a make deps-update, v1beta1 and v2alpha1 appear to have CronJob in there, but not v1 which is really strange, I suspect our dependencies are not updated properly.

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
awels: If client is not returning the right results it is likely not configured properly. The uncachedClient should only be used in very rare cases. The difference is that the uncachedClient makes an api call each time, while the regular client uses an informer. In this case I don't see a reason why the informer shouldn't work properly.

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
arnongilboa: `make deps-update` did not help with `v1`. Changing to v2alpha1. Ack?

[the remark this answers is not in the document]
arnongilboa: @awels trying v2alpha1, we get this: ```ERROR	controller-runtime.source	if kind is a CRD, it should be installed before calling Start	{"kind": "CronJob.batch", "error": "no matches for kind \"CronJob\" in version \"batch/v2alpha1\""}```

tools/cdi-source-update-poller/main.go
```diff
@@ -0,0 +1,74 @@
...
+	flag.StringVar(&cronName, "cron", "", "DataImportCron name.")
+	flag.StringVar(&url, "url", "", "registry source url.")
+	flag.Parse()
+	if url == "" || cronNamespace == "" || cronName == "" {
+		log.Fatalf("One or more mandatory parameters are missing")
+	}
+}
+
+func main() {
+	cmd := "skopeo inspect " + url + " | awk -F'\"' '/Digest/{print $4}'"
```
mhenriks, line 39: What about support for these params in registry import?

https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/apis/core/v1beta1/types.go#L166-L169

arnongilboa: How can we use SecretRef and CertConfigMap when we get the image as a ContainerDisk in the server container?

pkg/operator/resources/cluster/controller.go
```diff
@@ -211,6 +211,35 @@ func getControllerClusterPolicyRules() []rbacv1.PolicyRule {
 				"watch",
 			},
 		},
+		{
+			APIGroups: []string{
```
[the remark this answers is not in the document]
awels: This still looks cluster scoped. I see a new namespaced version as well.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,585 @@
...
+
+	if err := r.client.Update(ctx, dataImportCron); err != nil {
+		log.Error(err, "Unable to update DataImportCron", "Name", dataImportCron.Name)
+		return res, err
+	}
+	return res, nil
+}
+
+// Update DataSource and DataImportCron PVC on successful completion
+func (r *DataImportCronReconciler) updateSucceeded(ctx context.Context, dataImportCron *cdiv1.DataImportCron, dataVolume *cdiv1.DataVolume) error {
```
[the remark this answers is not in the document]
awels: Did you address this?

awels (commented):
Looks good, just 2 unaddressed items.

pkg/operator/resources/cluster/controller.go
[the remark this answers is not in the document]
arnongilboa: Verivied again, and in the top commit cronjobs *moved* from cluster scope to namespaced. No idea where do you see them...

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
arnongilboa: Yep. Thanks for the reminder!

pkg/operator/resources/cluster/controller.go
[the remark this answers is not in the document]
awels: Nevermind this is about imagestreams not cronjobs. Which can be in any namespace.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,638 @@
...
+
+	dataImportControllerName      = "dataimportcron-controller"
+	labelDataImportCronName       = "dataimportcron-name"
+	recentImportsToKeepPerCronJob = 3
+	digestPrefix                  = "sha256:"
+	digestDvNameSuffixLength      = 12
+)
+
+// Reconcile loop for DataImportCronReconciler
+func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
```
mhenriks, line 83: This is bad form.  A controller should only reconcile a single resource type.  What if there is a DIC and a ImageStream in the same namespace with the same name?

The ImageStream watch should look up all DICs that refer to theImageStream and queue those

arnongilboa: Nice point, which could be solved much simper if it was 1:1 DIC:IS. So with your suggestion on every DIC reconcile, if it has an IS source, it must Get() it and check the digest.

mhenriks: Yeah, if the DIC is requeued to be in accordance with cron schedule I'm not sure you even have to watch ImageStreams

arnongilboa: Just reminding that not-watching-IS and just getting it when its cron is scheduled was my initial suggestion, and someone opposed...

arnongilboa: Pushed both options (top patch removes ImageStream watch).

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+	dvList := &cdiv1.DataVolumeList{}
+	if err := r.client.List(ctx, dvList, &client.ListOptions{Namespace: dataImportCron.Namespace, LabelSelector: selector}); err != nil {
+		log.Error(err, "error listing dvs")
+		return err
+	}
+	if len(dvList.Items) <= recentImportsToKeepPerCronJob {
+		return nil
+	}
+	sort.Slice(dvList.Items, func(i, j int) bool {
+		return dvList.Items[i].CreationTimestamp.Time.After(dvList.Items[j].CreationTimestamp.Time)
```
mhenriks, line 377: Shouldn't this function return true if element at i < element at j?  Isn't this the opposite?

mhenriks: ah I see now `for _, dv := range dvList.Items[recentImportsToKeepPerCronJob:]`

arnongilboa: yep, you got it :)

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+		dataImportCron.Annotations = make(map[string]string)
+	}
+
+	regSource, err := getCronRegistrySource(dataImportCron)
+	if err != nil {
+		return res, err
+	}
+
+	if regSource.ImageStream != nil {
+		if err := r.updateImageStreamDesiredDigest(ctx, dataImportCron); err != nil {
```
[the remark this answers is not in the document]
arnongilboa: Right. Fixed in `updateImageStreamDesiredDigest()`.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+	} else if regSource.URL != nil {
+		cronJob, err := r.newCronJob(dataImportCron)
+		if err != nil {
+			return res, err
+		}
+		if err := r.client.Create(ctx, cronJob); err != nil {
+			r.log.Error(err, "Unable to create CronJob")
+			return res, err
+		}
+		AddFinalizer(dataImportCron, dataImportCronFinalizer)
```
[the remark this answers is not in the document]
arnongilboa: Right. Fixed.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+						Key: common.KeySecret,
+					},
+				},
+			},
+		}
+	}
+
+	job := &v1beta1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cron.Name,
```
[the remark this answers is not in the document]
arnongilboa: Right. Changed to `cron.Name-cron.UID`.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+		return reconcile.Result{}, err
+	}
+	if nextTime.Before(time.Now()) {
+		desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
+		if desiredDigest != "" && desiredDigest != dataImportCron.Status.CurrentImportDigest {
+			dataImportCron.Annotations[AnnSourceUpdatePending] = "true"
+		}
+		return r.setNextCronTime(dataImportCron)
+	}
+	return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
arnongilboa: Not sure it's a must, as the previously returned `requeueAfter` should cause a reclaim as scheduled (so `nextTime.Before(time.Now())` is true). Adding it just to be on the safe side.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+	// We use the poller returned reconcile.Result for RequeueAfter if needed
+	var err error
+	if dataImportCron.Annotations[AnnNextCronTime] != "" {
+		res, err = r.pollImageStreamDigest(ctx, dataImportCron)
+		if err != nil {
+			return res, err
+		}
+	}
+
+	if dataImportCron.Annotations[AnnSourceUpdatePending] == "true" {
```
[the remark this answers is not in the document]
arnongilboa: After removing the ImageStream watch this sync annotation is not needed anymore, as the DesiredDigest update is now synced anyway with the cron schedule. Removing it.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+		res, err = r.setNextCronTime(dataImportCron)
+		if err != nil {
+			return res, err
+		}
+	} else if regSource.URL != nil {
+		cronJob, err := r.newCronJob(dataImportCron)
+		if err != nil {
+			return res, err
+		}
+		if err := r.client.Create(ctx, cronJob); err != nil {
```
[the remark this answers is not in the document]
arnongilboa: Added

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+		if k8serrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	if dataImportCron.DeletionTimestamp != nil {
+		err := r.cleanup(ctx, dataImportCron)
+		return reconcile.Result{}, err
+	}
+	if dataImportCron.Annotations[AnnCronInitialized] != "true" {
```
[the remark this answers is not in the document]
arnongilboa: Fixed

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
mhenriks: That's not how `requeueAfter` works.  If a resource is updated before the `requeueAfter` time, it will be reconciled again immediately and not again (unless requeueAfter is set again)

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,611 @@
...
+	}
+	return r.update(ctx, dataImportCron)
+}
+
+func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
+	if dataImportCron.Annotations == nil {
+		dataImportCron.Annotations = make(map[string]string)
+	}
+	if len(dataImportCron.Status.CurrentImportDataVolumeNames) == 0 {
+		dataImportCron.Status.CurrentImportDataVolumeNames = make([]string, 1)
```
mhenriks, line 103: my concern with this is that it will create a list with empty string.  Why not just wait until we have a dv name and assign it like this `dataImportCron.Status.CurrentImportDataVolumeName = []string{dvName}`

Then when that dv succeeds set to nil

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
+	log := r.log.WithName("update")
+	res := reconcile.Result{}
+
+	dvName := dataImportCron.Status.CurrentImportDataVolumeName
+	if dvName != "" {
+		// Get the currently imported DataVolume
+		dataVolume := &cdiv1.DataVolume{}
+		if err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: dvName}, dataVolume); err != nil {
+			if k8serrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
arnongilboa: Yep, s#!t happens, so added recreate here. Nice.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,594 @@
...
+		dataImportCron.Status.LastExecutionTimestamp = &now
+
+		if dataVolume.Status.Phase == cdiv1.Succeeded {
+			if err := r.updateDataSourceOnSuccess(ctx, dataImportCron); err != nil {
+				return res, err
+			}
+			if err := r.updateDataImportCronOnSuccess(ctx, dataImportCron); err != nil {
+				return res, err
+			}
+			if err := r.garbageCollectOldImports(ctx, dataImportCron); err != nil {
```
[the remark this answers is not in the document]
arnongilboa: We currently support a single import, so until it succeed there will be no further imports for this DataImportCron. We may improve this behavior after this PR is merged.

awels:
Can we get at least some functional tests so we have some confidence this works?

arnongilboa:
> Can we get at least some functional tests so we have some confidence this works?

@awels I mostly test it with a DIC yaml.
Functional tests needs some more time as they are not trivial at all:
-Which registry source to use (currently only external urls are working for me, such as `docker://quay.io/kubevirt/fedora-cloud-registry-disk-demo:latest`, but local ones such as `docker://cdi-docker-registry-host.cdi/tinycoreqcow2` fail the skopeo inspect app for some reason)
-How to cover ImageStream tests?
-Updating the source image/ImageStream (sha256 digest)
-Causing multiple updates and imports to verify garbage collection
etc.

awels:
At a minimum you should be able to push some test container images to the registry:5000 used for the application containers, the same way we push those. And then demonstrate the functionality works with node import. As for skopeo inspect failing, that is weird, since we are able to import these container images without issue with a regular import. And that uses the same library as skopeo.

You should be able to just add the ImageStream CRD (if it doesn't exist already, like with running the test on OpenShift), and manually create some, maybe even point them to the same registry for the non imagestream tests. 

For updating, that is a good question, we have to consider non kubevirtci environments when thinking about these. We would need a mechanism to push images to the registry from the test, we don't have that capability right now.

arnongilboa:
@awels btw node import functional tests (from registry:5000) are already merged and running in the ci for a long time.

akrejcir:
Could you export somewhere the code that validates `DataImportCronSpec` ? We could use it in the SSP operator. It would be faster than calling a dry-run API.

awels:
> @awels btw node import functional tests (from registry:5000) are already merged and running in the ci for a long time.

Right, but we want to test the other part of the code, and how the actual import happens is less interesting than how the controller behaves. So if there are no other options for the actual import part, that is one way we know works.

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,210 @@
...
+			Eventually(func() bool {
+				dvList, err := f.CdiClient.CdiV1beta1().DataVolumes(f.Namespace.Name).List(context.TODO(), metav1.ListOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return len(dvList.Items) == 2
+			}, dataImportCronTimeout, pollingInterval).Should(BeTrue())
+		}
+	},
+		table.Entry("[test_id:7403] Should successfully import PVC from registry URL as scheduled", scheduleEveryMinute, 1),
+		table.Entry("[test_id:7414] Should successfully import PVC from registry URL on source digest update", scheduleOnceAYear, 2),
+		table.Entry("[test_id:7406] Should successfully grabage collect old PVCs when importing new ones", scheduleOnceAYear, 3),
```
awels, line 143: typo `grabage`

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,210 @@
...
+func NewDataImportCron(name, size, schedule, dataSource string, source cdiv1.DataVolumeSourceRegistry) *cdiv1.DataImportCron {
+	garbageCollect := cdiv1.DataImportCronGarbageCollectOutdated
+	var importsToKeep int32 = 2
+
+	return &cdiv1.DataImportCron{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Annotations: map[string]string{
+				controller.AnnImmediateBinding: "true",
+				//controller.AnnPodRetainAfterCompletion: "true",
```
awels, line 157: leftover?

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,210 @@
...
+				_, err = utils.FindPodByPrefix(f.K8sClient, f.CdiInstallNs, cronjob.Status.Active[0].Name, "")
+				return err == nil
+			}, dataImportCronTimeout, pollingInterval).Should(BeTrue())
+		}
+
+		var lastImportDv, currentImportDv string
+		for i := 0; i < repeat; i++ {
+			if repeat > 1 {
+				// Emulate source update
+				digest := []string{"sha256:0001112223334444", "sha256:0123456789abcdef", "sha256:aaabbbcccdddeeee"}[i%3]
```
awels, line 83: Don't really need i%3 if i is never > 3

awels (commented):
I believe we already have functional tests that ensure the job does what it is supposed to do, and these tests are purely testing the controller (what happens when the annotations on the object changes)

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+var _ = Describe("DataImportCron", func() {
+	const (
+		scheduleEveryMinute = "* * * * *"
+		scheduleOnceAYear   = "0 0 1 1 *"
+	)
+	var (
+		f                = framework.NewFramework(namespacePrefix)
+		registryPullNode = cdiv1.RegistryPullNode
+		//trustedRegistryURL = func() string { return fmt.Sprintf(utils.TrustedRegistryURL, f.DockerPrefix, f.DockerTag) }
```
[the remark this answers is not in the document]
arnongilboa: fixed

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,210 @@
...
+	AfterEach(func() {
+		if cron != nil {
+			By("Delete cron")
+			err = DeleteDataImportCron(f.CdiClient, cron.Namespace, cron.Name)
+			Expect(err).ToNot(HaveOccurred())
+		}
+	})
+
+	table.DescribeTable("should", func(schedule string, repeat int) {
+		//FIXME: skopeo inspect docker://registry:5000/cdi-func-test-tinycore:latest": exit status 1:
```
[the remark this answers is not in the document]
arnongilboa: What's the common way in tests to diff between U/S and D/S?

tests/dataimportcron_test.go
arnongilboa: I tried import loops so it was helpful to be on the safe side. Removed.

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,219 @@
...
+		trustedRegistryURL  = func() string { return fmt.Sprintf(utils.TrustedRegistryURL, f.DockerPrefix, f.DockerTag) }
+		externalRegistryURL = "docker://quay.io/kubevirt/fedora-cloud-registry-disk-demo:latest"
+		cron                *cdiv1.DataImportCron
+		err                 error
+	)
+
+	AfterEach(func() {
+		if cron != nil {
+			By("Delete cron")
+			err = DeleteDataImportCron(f.CdiClient, cron.Namespace, cron.Name)
```
mhenriks, line 47: make sure cluster clean script removes any finalizers as well, just in case.  Look here:

https://github.com/kubevirt/containerized-data-importer/blob/main/cluster-sync/clean.sh#L23-L35

arnongilboa: done

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,219 @@
...
+				_, err = utils.FindPodByPrefix(f.K8sClient, f.CdiInstallNs, cronjob.Status.Active[0].Name, "")
+				return err == nil
+			}, dataImportCronTimeout, pollingInterval).Should(BeTrue())
+		}
+
+		var lastImportDv, currentImportDv string
+		for i := 0; i < repeat; i++ {
+			if repeat > 1 {
+				// Emulate source update
+				digest := []string{"sha256:0001112223334444", "sha256:0123456789abcdef", "sha256:aaabbbcccdddeeee"}[i]
```
mhenriks, line 93: clearly these are fake shas, so I'm wondering why import succeeds.  Shouldn't an import be bound to a specific sha?

arnongilboa: The 1st test case actually uses the cronjob (which uses skopeo inspect to read the source image digest sha, and update the DataImportCron DesiredDigest annotation accordingly)
The 2nd & 3rd test cases simply emulate the cronjob app to test source image updates. So were are just updating the sha which is enough to test all but the cronjob.

awels: I am thinking that Michaels point is that we should be grabbing the exact sha, that is in the desired, this way we can guarantee we are getting the correct one. What if somehow in between us scheduling the job, and the job completing and triggering the controller to create the DV, a new version becomes available in the registry, and the sha is different. We would want the desired sha to actually be imported instead of :latest which may not be the sha we wanted.

arnongilboa: (1) seems reasonable and I may fix it, but
(2) afaiu all we are interested is always importing the latest update, so in the rare case described  that's what happens but we'll  have an entry with the old sha and the new DV name will be based on this old sha. However it will be imported again since the source sha updated.
(3) the testing method suggested here won't work, and we'll need to find a smarter way to do it...
(4) "Perfection is the enemy of progress." -Winston Churchill

awels: No we are interested in the exact sha being in the DV that the datasource points to. There is a reason we keep some of the older ones around, so we can go back to an exact version indicated by the sha. If that is not always the case there is no point in keeping older ones around.
For the test, you are right the current approach won't work, but we could figure out the shas we are interested in ahead of time by basically running the script in the job, and collecting the sha from that. Don't we return it in the exit code? We could do that for some of the older versions of that image in quay, collect the shas from there, and then use that in the test. It should then succeed.

arnongilboa: Convinced. Fixing.

mhenriks:
Just checking in to ask:

What is the plan for testing ImageStreams?

dataimportcron-controller.go coverage currently at 1.5%.  Any plans to add testing there?

arnongilboa:
Found a small issue mistakenly pushed on ImageStreams + sha256 import, holding until it's fixed
@mhenriks - functional tests coverage will be extended in follow-up PR. ImageStreams are planned to be tested only D/S on an OpenShift cluster.

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
awels: Yes please make a note that we probably want to just run the garbage collection regardless of import success.

pkg/controller/dataimportcron-controller.go
```diff
@@ -0,0 +1,657 @@
...
+	}
+
+	return job, nil
+}
+
+func newSourceDataVolume(cron *cdiv1.DataImportCron, dataVolumeName string) *cdiv1.DataVolume {
+	var digestedURL string
+	dv := cron.Spec.Template.DeepCopy()
+	if isURLSource(cron) {
+		digestedURL = *dv.Spec.Source.Registry.URL + "@" + cron.Annotations[AnnSourceDesiredDigest]
```
awels, line 587: I forget does the URL contain the tag element as well? And if that even matter when you specify the sha. If the functional tests pass we would know I guess.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+)
+
+var (
+	cronLog = logf.Log.WithName("data-import-cron-controller-test")
+)
+
+var _ = Describe("All DataImportCron Tests", func() {
```
awels, line 43: More unit tests please, increase our confidence that this all works properly!

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,225 @@
...
+				return err == nil
+			}, dataImportCronTimeout, pollingInterval).Should(BeTrue())
+		}
+
+		var lastImportDv, currentImportDv string
+		for i := 0; i < repeat; i++ {
+			if repeat > 1 {
+				// Emulate source update using digests from https://quay.io/repository/kubevirt/cirros-container-disk-demo?tab=tags
+				digest := []string{
+					"sha256:68b44fc891f3fae6703d4b74bcc9b5f24df8d23f12e642805d1420cbe7a4be70",
```
awels, line 98: Are these shas associated with a tag? I don't think they will get overriden with a different sha in the future, but might want to think about looking up the with a pod at the beginning of the test, and building the digest list from there. I am okay with this for now.

awels (commented):
Few super minor questions/notes. Besides the unit tests.

pkg/controller/dataimportcron-controller.go
awels: Actually just saw you take the tag off. so we are good.

pkg/controller/dataimportcron-controller_test.go
arnongilboa: Added, coverage is 55% now. Will add more utests after merge.

tests/dataimportcron_test.go
arnongilboa: To be fixed post-merge.

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,225 @@
...
+			By("Delete cron")
+			err = DeleteDataImportCron(f.CdiClient, cron.Namespace, cron.Name)
+			Expect(err).ToNot(HaveOccurred())
+		}
+	})
+
+	table.DescribeTable("should", func(schedule string, repeat int, checkGarbageCollection bool) {
+		var url string
+
+		if repeat > 1 || utils.IsOpenshift(f.K8sClient) {
```
mhenriks, line 59: Shouldn't ever go to the 'net for functional tests.  Will this be addressed later?

mhenriks (commented):
Will lgtm once description is updated to include to-do items for subsequent pr

tests/dataimportcron_test.go
arnongilboa: Yes we will. Adding it to the post-merge tasks.

awels: Well the problem here is we don't have a good repo with these D/S. We need to brain storm a little bit on how we can make it work reliable both U/S and D/S without hitting the outside world.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,315 @@
...
+			}
+		})
+
+		It("Should do nothing and return nil when no DataImportCron exists", func() {
+			reconciler = createDataImportCronReconciler()
+			_, err := reconciler.Reconcile(context.TODO(), cronReq)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		DescribeTable("Should", func(setDeletionTimestamp bool) {
```
awels, line 84: I would just make this an "It" should create and delete cronjob if dataimport cron is created and delete. In the second loop you are basically running the first loop again, and then deleting. Its not a huge deal. The tables are mostly useful for if you are running the same function with various inputs and various expectations. Here you are creating a import cron and verify the job is created, and then deleting (why not just delete it instead of setting the timestamp?) and checking that the cron is gone.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+				err = reconciler.client.Get(context.TODO(), cronJobKey(cron), cronjob)
+				Expect(err).To(HaveOccurred())
+			}
+		},
+			Entry("create CronJob and add finalizer on a valid DataImportCron with source registry URL", false),
+			Entry("delete CronJob when DataImportCron with source registry URL has DeletionTimestamp", true),
+		)
+
+		DescribeTable("Should", func(setSucceeded bool) {
```
awels, line 116: Again this is just a sequence of events you are testing. There is no sense in running the same test twice, but the second time something extra. Just run the second one in an it. The coverage will be the same

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,315 @@
...
+				Expect(cron.Status.LastImportedPVC).ToNot(BeNil())
+				Expect(*cron.Status.LastImportedPVC).To(Equal(sourcePVC))
+				Expect(cron.Status.LastImportTimestamp).ToNot(BeNil())
+			}
+		},
+			Entry("create DataVolume on AnnSourceDesiredDigest annotation update", false),
+			Entry("update DataImportCron and DataSource on DataVolume Succeeded", true),
+		)
+
+		DescribeTable("Should", func(updateNextCronTime bool) {
```
awels, line 182: Same thing don't think you need the table. And It with the sequence of events should be sufficient.

tests/dataimportcron_test.go
```diff
@@ -0,0 +1,225 @@
...
+			_, err = utils.WaitForPVC(f.K8sClient, cron.Namespace, currentImportDv)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Wait for import completion")
+			err = utils.WaitForDataVolumePhase(f.CdiClient, cron.Namespace, cdiv1.Succeeded, currentImportDv)
+			Expect(err).ToNot(HaveOccurred(), "Datavolume not in phase succeeded in time")
+
+			By("Verify datasource was updated")
+			Eventually(func() bool {
+				datasource, err := f.CdiClient.CdiV1beta1().DataSources(f.Namespace.Name).Get(context.TODO(), cron.Spec.ManagedDataSource, metav1.GetOptions{})
```
awels, line 137: You have a flaky test noting that the datasource doesn't exist yet sometimes. This is because after the DV is complete the controller then creates the datasource. But we might get here before that actually happens. So I would check if `errors.IsNotFound(err)` is true, and if so return false, so we retry in a few seconds.

arnongilboa: fixed

pkg/controller/dataimportcron-controller_test.go
arnongilboa: fixed

pkg/controller/dataimportcron-controller_test.go
arnongilboa: fixed

pkg/controller/dataimportcron-controller_test.go
arnongilboa: fixed

mhenriks:
/lgtm

@awels has the honor of approval

akalenyu:
/retest
since 1.22 lane switch warm upgrade flake is much more visible

awels (commented):
Some scenarios to think about when writing more unit tests:
- A cronjob with the name you are about to create already exists, but is not owned by the controller (we can't create the cron) what happens, how should the controller handle that.
- A DV with the name you are about to create already exists, but is not owned by the controller, or maybe there is a quota restriction preventing you from creating the PVC associated with the controller.
- What happens when the DV is in running state and you reconcile, how is that reflected in the DIC, what about recounts on the DV, should that be reflected in the DIC?
- What happens when a datasource already exists?
Basically what happens when it is not the happy path.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			_, err = reconciler.Reconcile(context.TODO(), cronReq)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = reconciler.client.Get(context.TODO(), cronKey, cron)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cron.Status.CurrentImports).ToNot(BeNil())
+
+			dvName := cron.Status.CurrentImports[0].DataVolumeName
```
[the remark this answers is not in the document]
arnongilboa: added

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+			Expect(dvName).ToNot(BeEmpty())
+			digest := cron.Status.CurrentImports[0].Digest
+			Expect(digest).ToNot(BeEmpty())
```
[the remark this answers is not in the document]
arnongilboa: added

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+			Expect(cron.Annotations[AnnNextCronTime]).ToNot(Equal(timestamp))
+
+			digest := cron.Annotations[AnnSourceDesiredDigest]
+			Expect(digest).To(Equal(testDigest))
+			dockerRef := cron.Annotations[AnnImageStreamDockerRef]
+			Expect(dockerRef).To(Equal(testDockerRef))
+
+			Expect(cron.Status.CurrentImports).ToNot(BeNil())
+			dvName := cron.Status.CurrentImports[0].DataVolumeName
```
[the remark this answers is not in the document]
arnongilboa: added

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+			dvName := cron.Status.CurrentImports[0].DataVolumeName
+			Expect(dvName).ToNot(BeEmpty())
+			digest := cron.Status.CurrentImports[0].Digest
+			Expect(digest).ToNot(BeEmpty())
+
+			dv := &cdiv1.DataVolume{}
+			err = reconciler.client.Get(context.TODO(), dvKey(dvName), dv)
+			Expect(err).ToNot(HaveOccurred())
+
```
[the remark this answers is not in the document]
arnongilboa: added

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+			Expect(cron.Status.CurrentImports).ToNot(BeNil())
+			dvName := cron.Status.CurrentImports[0].DataVolumeName
+			Expect(dvName).ToNot(BeEmpty())
+			digest = cron.Status.CurrentImports[0].Digest
+			Expect(digest).To(Equal(testDigest))
+
+			dv := &cdiv1.DataVolume{}
+			err = reconciler.client.Get(context.TODO(), dvKey(dvName), dv)
+			Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
arnongilboa: added

PR #1950, opened 2021-09-20
Retry clone kill which is flaky

awels:

[test_id:4000] Create a data volume and then clone it while killing the
container and verify retry count was pretty flaky lately. The likely culprit was the small amount of data being cloned. This causes the pods to be removed before we have a chance to connect issue the kill command.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -1286,9 +1286,12 @@ var _ = Describe("all clone tests", func() {
 
 		It("[test_id:4000] Create a data volume and then clone it while killing the container and verify retry count", func() {
 			By("Prepare source PVC")
-			pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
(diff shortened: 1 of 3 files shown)


awels:
/hold
going to retry the runs to ensure this really fixes the problem.

brybacki (commented):
line 1319:
```			err = utils.WaitTimeoutForPodReady(f.K8sClient, utils.UploadPodName(targetPvc), targetNs.Name, utils.PodWaitForTime)```

This wait has polling interval set to 2 seconds. So, in worst case, it can miss any action that has finished in 2 seconds. I think in good conditions our upload server can handle tinyCoreIso in milliseconds.

brybacki:
hpp destructive has a problem in before each, looks like it cannot find CDI

`s: "Unable to find pod containing cdi-deployment",`
other than that, looks good, trying again

/test all

PR #1955, opened 2021-09-24
build multi-arch builder

zhlhahaha:
Signed-off-by: Howard Zhang <howard.zhang@arm.com>

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/build/bazel-build-builder.sh
+++ b/hack/build/bazel-build-builder.sh
@@ -26,15 +26,35 @@ if ! git diff-index --quiet HEAD~1 hack/build/docker; then
   UNTAGGED_BUILDER_IMAGE=quay.io/kubevirt/kubevirt-cdi-bazel-builder
   BUILDER_TAG=$(date +"%y%m%d%H%M")-$(git rev-parse --short HEAD)
   echo "$DOCKER_PREFIX:$DOCKER_TAG"
+  #Prepare env for build multi-arch image
+  ARCHITECTURES="amd64 arm64"
+  docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
 
-  #Build the encapsulated compile and test container
-  (cd ${BUILDER_SPEC} && docker build --tag ${UNTAGGED_BUILDER_IMAGE}:${BUILDER_TAG} .)
+  #Build and push multi-arch builder
+  for ARCH in ${ARCHITECTURES}; do
+      case ${ARCH} in
+      amd64)
+          bazel_arch="x86_64"
+          ;;
+      *)
+          bazel_arch=${ARCH}
+          ;;
+      esac
 
-  DIGEST=$(docker images --digests | grep ${UNTAGGED_BUILDER_IMAGE} | grep ${BUILDER_TAG} | awk '{ print $4 }')
-  echo "Image: ${UNTAGGED_BUILDER_IMAGE}:${BUILDER_TAG}"
-  echo "Digest: ${DIGEST}"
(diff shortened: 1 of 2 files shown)


zhlhahaha:
cc: @awels

awels:
You will likely have to modify hack/build/config.sh in particular:
```bash
BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder@sha256:45ff791390bcd3902781e0f790083f30bba8553d2568e34e75675228c5d4bfb5
```
That is the builder image we use for building the images, we will have to have an arm version based on the selected arch as well.

awels:
When we update the builder images it is a 2 step process. First we update the dockerfile for the builder image, merge that, that will produce tagged images in quay, then we make another PR that pulls the sha256 from quay. This is used for the additional containers for the tests for instance.

zhlhahaha:
> You will likely have to modify hack/build/config.sh in particular:
> 
> ```shell
> BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder@sha256:45ff791390bcd3902781e0f790083f30bba8553d2568e34e75675228c5d4bfb5
> ```
> 
> That is the builder image we use for building the images, we will have to have an arm version based on the selected arch as well.

Ok, I see, we can only use tag rather than sha256 for multi-arch images. So shall I change it to
```shell
BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder:0.0.13}
```
If yes, do we need to use a higher version like 0.0.14 rather than 0.0.13?

hack/build/config.sh
```diff
@@ -28,8 +28,7 @@ FUNC_TEST_BAD_WEBSERVER="cdi-func-test-bad-webserver"
 FUNC_TEST_PROXY="cdi-func-test-proxy"
 # update this whenever builder Dockerfile is updated
 BUILDER_TAG=${BUILDER_TAG:-0.0.13}
-BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder@sha256:45ff791390bcd3902781e0f790083f30bba8553d2568e34e75675228c5d4bfb5
-}
+BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder:0.0.13}
```
awels, line 31: I think you misunderstood what I was saying, we will need to set the builder image based on the arch, something like
```bash
if [ $ARCH == "amd64" ]; then
  BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder@sha256:45ff791390bcd3902781e0f790083f30bba8553d2568e34e75675228c5d4bfb5
else if [ $ARCH == "arm64" ]; then
  BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder-arm@sha256:<sha>
fi
```
The point I was trying to make we don't know what the sha is, until after we have created the builder image. So that step has to happen after this PR is merged, and a new builder image has been pushed.

zhlhahaha: Ok, I get your idea, but I have one unclear on this. 
As the multi-arch manifest for the builder image has been uploaded, by using a tag to identify the image, docker would automatically detect the server arch and decides which image should be used. Why use sha256 to identify the image? Are there any specific purposes?

PR #1957, opened 2021-09-24
fix cdi-importer wrong rpm package for aarch64

zhlhahaha:
apply correct package for aarch64 build

**Release note**:
```release-note
NONE
```


zhlhahaha:
cc: @awels

PR #1960, opened 2021-09-27
 Use inclusive language, add check for inclusive language

maya-r:
**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/gen-swagger-doc/deploy.sh
+++ b/hack/gen-swagger-doc/deploy.sh
@@ -30,7 +30,7 @@ Content of this repository is generated from OpenAPI specification of
 
 ## KubeVirt Containerized Data Importer API References
 
-* [master](${GITHUB_IO_FQDN}/master/index.html)
+* [main](${GITHUB_IO_FQDN}/main/index.html)
 __EOF__
 find * -type d -regex "^v[0-9.]*" \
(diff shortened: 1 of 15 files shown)


hack/gen-swagger-doc/deploy.sh
```diff
@@ -30,7 +30,7 @@ Content of this repository is generated from OpenAPI specification of
 
 ## KubeVirt Containerized Data Importer API References
 
-* [master](${GITHUB_IO_FQDN}/master/index.html)
+* [main](${GITHUB_IO_FQDN}/main/index.html)
```
awels, line 33: This repo has not been converted yet, so if we make this change it will fail to push the api reference.

hack/gen-swagger-doc/deploy.sh
```diff
@@ -43,7 +43,7 @@ if git status --porcelain | grep -v "index[.]html" | grep --quiet "^ [AM]"; then
     git add -A README.md "${TARGET_DIR}"/*.html
     git commit --message "API Reference update by KubeVirt Prow build ${BUILD_ID}"
 
-    git push origin master >/dev/null 2>&1
+    git push origin main >/dev/null 2>&1
```
awels, line 46: This repo has not been converted yet, so if we make this change it will fail to push the api reference.

awels (commented):
Overal good PR, just one repo has not been converted yet.

PR #1963, opened 2021-09-28
Add go-ovirt-client for ImageIO importer.

mrnold:
This pull request breaks off a small piece of #1889, so that it is easier to review the inclusion of go-ovirt-client in stages. Specifically, this pull request adds go-ovirt-client as a dependency and uses it only to log in to the oVirt API, instead of replacing the whole ImageIO importer up front like in #1889. The longer term plan is to replace the existing go-ovirt SDK with go-ovirt-client as functionality becomes available.

This pull request also adjusts existing tests to accommodate the new library, so they should all pass now.

None, see #1889.

No actual functionality should be impacted by this change. Only the last few commits have real code changes.

**Release note**:
```release-note
NONE
```


awels (commented):
looks good just one question, the secret is normally base64 encoded, did that change for the ovirt datasource?

mrnold:
> looks good just one question, the secret is normally base64 encoded, did that change for the ovirt datasource?

It didn't really change, but go-ovirt-client actually validates the username format, and the functional tests were failing because the importer itself was apparently getting passed the base64-encoded value instead of the plain text. So I guess those base64-encoded credentials have been getting base64-encoded again somewhere else along the test chain, and it hasn't mattered to the go-ovirt SDK or to the fakeovirt/imageiotest containers. I think because it uses stringData instead of data in the call to NewSecretDefinition, the k8s client takes care of the base64-encoding, so it doesn't need to be done in advance.

Nothing changed about the encoding for running a real import, this was purely for the functional test setup.

awels:
Interesting, okay
/lgtm
/approve

mrnold:
It was failing to find the image ID that was removed in #1967, so I rebased. Maybe that will help?

mrnold:
That looks much better. Lost the LGTM though.

PR #1969, opened 2021-10-03
Upload archive

ShellyKa13:

Fixes #

Added support in image-upload api in virtctl in kubevirt repo: PR: 

**Release note**:

```release-note
Support archive upload
```


pkg/uploadserver/uploadserver.go
```diff
@@ -422,6 +459,17 @@ func newUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string, file
 	return processor.PreallocationApplied(), err
 }
 
+func newUploadArchiveStreamProcessor(stream io.ReadCloser) error {
+	destDir := common.ImporterVolumePath
+	if err := importer.CleanDir(destDir); err != nil {
+		return errors.Wrapf(err, "error removing contents of %s", destDir)
+	}
+	if err := util.UnArchiveTar(stream, destDir); err != nil {
```
mhenriks, line 467: @awels should we be going to the data processor api here?  Is there currently a path that will handle this?

doc/upload.md
```diff
@@ -87,6 +87,11 @@ spec:
 ```bash
 kubectl apply -f manifests/example/upload-datavolume.yaml
 ```
+### Create a Data Volume for archive upload
+
+You can also upload an archive. Specifying in the data volume spec: contentType: archive
+will mark the datavolume as archive upload and will untar the recieved source.
```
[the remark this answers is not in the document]
awels: I think the point of archive is to upload arbitrary things that are not images. I would say received content.

pkg/uploadserver/uploadserver.go
awels: That is a good question, had we gone with a single endpoint, it would have gone into an UploadDataSource, and we could have modified the Transfer function to check for content type and call untar in there just like we do in [http-datasource ](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/http-datasource.go#L170-L176) But because we have 2 endpoints, I am not sure if it makes sense to duplicate it all for a new endpoint.

doc/upload.md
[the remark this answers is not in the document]
arnongilboa: right!

pkg/controller/datavolume-controller.go
```diff
@@ -2541,14 +2572,28 @@ func getDefaultVolumeMode(c client.Client, storageClass *storagev1.StorageClass)
...
+		for _, cps := range storageProfile.Status.ClaimPropertySets {
+			for _, accessMode := range cps.AccessModes {
+				for _, pvcAccessMode := range pvcAccessModes {
+					if accessMode == pvcAccessMode {
+						return cps.VolumeMode, nil
+					}
+				}
+			}
+		}
+		// if not found return default volume mode for the storage class
```
brybacki, line 2588: If accessModes are defined but not found - it might mean that a user defined accessMode that is not supported, but I think our current logic is correct. 

The effect would be the pvc could not be bound, and the user can check the PVC events and learn which accessModes are suported on the selected storage.

pkg/controller/datavolume-controller.go
```diff
@@ -2559,10 +2604,22 @@ func getDefaultAccessModes(c client.Client, storageClass *storagev1.StorageClass
...
+	if len(storageProfile.Status.ClaimPropertySets) > 0 {
+		// check for access modes matching with given pvc volume mode
+		defaultAccessModes := []corev1.PersistentVolumeAccessMode{}
+		for _, cps := range storageProfile.Status.ClaimPropertySets {
+			if cps.VolumeMode != nil && *cps.VolumeMode == *pvcVolumeMode {
+				if len(cps.AccessModes) > 0 {
+					return cps.AccessModes, nil
+				}
+			} else if len(cps.AccessModes) > 0 && len(defaultAccessModes) == 0 {
+				defaultAccessModes = cps.AccessModes
```
brybacki, line 2616: so default can be the last accessModes found?
why not take first one as in VolumeModes  `storageProfile.Status.ClaimPropertySets[0].AccessModes`

pkg/controller/storageprofile-controller.go
```diff
@@ -60,22 +61,22 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
...
-		claimPropertySet = &storageProfile.Status.ClaimPropertySets[0]
+	if len(storageProfile.Spec.ClaimPropertySets) > 0 {
+		for _, cps := range storageProfile.Spec.ClaimPropertySets {
+			if len(cps.AccessModes) == 0 && cps.VolumeMode != nil {
+				err = fmt.Errorf("must provide access mode for volume mode: %s", *cps.VolumeMode)
+				log.Error(err, "Unable to update StorageProfile")
+				return reconcile.Result{}, err
+			}
+		}
+		claimPropertySets = storageProfile.Spec.ClaimPropertySets
```
brybacki, line 74: so this is either all set by user or from CDI knowledge, in my opinion this is good and simple to understand

pkg/controller/datavolume-controller.go
ShellyKa13: @brybacki Not sure I understood. You mean that if len(pvcAccessModes) == 0 or we didnt find in the loop the access mode then we should return nil? or if len(pvcAccessModes) == 0 return the storageProfile.Status.ClaimPropertySets[0].VolumeMode and in case we didnt find in the loop we should return error?

pkg/controller/datavolume-controller.go
ShellyKa13: no the default will be the first access modes found. I didnt just take storageProfile.Status.ClaimPropertySets[0].AccessModes cause I thought it is possible that it will be an empty list.

pkg/controller/util_test.go
```diff
@@ -655,25 +655,30 @@ func createStorageClass(name string, annotations map[string]string) *storagev1.S
 func createStorageProfile(name string,
 	accessModes []v1.PersistentVolumeAccessMode,
 	volumeMode v1.PersistentVolumeMode) *cdiv1.StorageProfile {
-	return createStorageProfileWithCloneStrategy(name, accessModes, volumeMode, nil)
+	claimPropertySets := []cdiv1.ClaimPropertySet{{
+		AccessModes: accessModes,
+		VolumeMode:  &volumeMode,
+	}}
+	return createStorageProfileWithCloneStrategy(name, claimPropertySets, nil)
```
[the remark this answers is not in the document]
ShellyKa13: it is by itself calls createStorageProfileWithCloneStrategy so it doesnt shorten the way

pkg/controller/datavolume-controller.go
brybacki: I mean, we did not find in the loop. But maybe all this options are similar. I need to think about this again.

pkg/controller/datavolume-controller.go
```diff
@@ -1413,7 +1413,7 @@ func getStorageVolumeMode(c client.Client, dataVolume *cdiv1.DataVolume, storage
 		if dataVolume.Spec.Storage.VolumeMode != nil {
 			return dataVolume.Spec.Storage.VolumeMode, nil
 		}
-		volumeMode, err := getDefaultVolumeMode(c, storageClass)
+		volumeMode, err := getDefaultVolumeMode(c, storageClass, dataVolume.Spec.Storage.AccessModes)
```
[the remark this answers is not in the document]
ShellyKa13: I didnt do handle the content-type here.. will change it to set the volume mode as FS in case of archive content type

aglitke (approved):
/lgtm

pkg/uploadserver/uploadserver.go
```diff
@@ -83,6 +84,7 @@ type imageReadCloser func(*http.Request) (io.ReadCloser, error)
 // may be overridden in tests
 var uploadProcessorFunc = newUploadStreamProcessor
 var uploadProcessorFuncAsync = newAsyncUploadStreamProcessor
+var uploadArchiveProcessorFunc = newUploadArchiveStreamProcessor
```
awels, line 87: Is this still used?

pkg/uploadserver/uploadserver.go
```diff
@@ -410,18 +429,29 @@ func newAsyncUploadStreamProcessor(stream io.ReadCloser, dest, imageSize string,
...
 
 	// Clone block device to block device or file system
-	uds := importer.NewUploadDataSource(newContentReader(stream, sourceContentType))
+	uds := importer.NewUploadDataSource(newContentReader(stream, sourceContentType), dvContentType)
 	processor := importer.NewDataProcessor(uds, dest, common.ImporterVolumePath, common.ScratchDataDir, imageSize, filesystemOverhead, preallocation)
 	err := processor.ProcessData()
 	return processor.PreallocationApplied(), err
 }
 
+func newUploadArchiveStreamProcessor(stream io.ReadCloser) error {
```
awels, line 444: This still used?

pkg/controller/datavolume-controller.go
```diff
@@ -2558,10 +2613,22 @@ func getDefaultAccessModes(c client.Client, storageClass *storagev1.StorageClass
...
 
-	if len(storageProfile.Status.ClaimPropertySets) > 0 &&
-		len(storageProfile.Status.ClaimPropertySets[0].AccessModes) > 0 {
-		accessModes := storageProfile.Status.ClaimPropertySets[0].AccessModes
-		return accessModes, nil
+	if len(storageProfile.Status.ClaimPropertySets) > 0 {
+		// check for access modes matching with given pvc volume mode
+		defaultAccessModes := []corev1.PersistentVolumeAccessMode{}
+		for _, cps := range storageProfile.Status.ClaimPropertySets {
+			if cps.VolumeMode != nil && *cps.VolumeMode == *pvcVolumeMode {
```
awels, line 2620: can pvcVolumeMode be nil? If so then this is a npe

doc/storageprofile.md
```diff
@@ -38,12 +38,13 @@ status:
...
 - `cloneStrategy` - defines the preferred method for performing a CDI clone
-- `claimPropertySets` contains a list of `claimPropertySet` (currently only one supported) 
+- `claimPropertySets` contains a list of `claimPropertySet`
   - `accessMode` - contains the desired access modes the volume should have
   - `volumeMode` - defines what type of volume is required by the claim
 
 Values for accessModes and volumeMode are exactly the same as for PVC: `accessModes` is a list of `[ReadWriteMany|ReadWriteOnce|ReadOnlyMany]`
-and `volumeMode` is a single value `Filesystem` or `Block`. The value for `cloneStrategy` ca be one of: `copy`,`snapshot`,`csi-clone`. 
+and `volumeMode` is a single value `Filesystem` or `Block`.
+The value for `cloneStrategy` ca be one of: `copy`,`snapshot`,`csi-clone`.
```
awels, line 47: typo `ca be one of`

awels (commented):
So I am kind of confused by the logic that looks up the volumeMode based on access mode. I can understand accessMode based on volumeMode (for ceph-rbd) but not the other way around.

pkg/controller/datavolume-controller.go
```diff
@@ -2527,7 +2546,29 @@ func (r *DatavolumeReconciler) getPreferredCloneStrategyForStorageClass(storageC
 	return storageProfile.Status.CloneStrategy, nil
 }
 
-func getDefaultVolumeMode(c client.Client, storageClass *storagev1.StorageClass) (*corev1.PersistentVolumeMode, error) {
+func getDefaultVolumeAndAccessMode(c client.Client, storageClass *storagev1.StorageClass) ([]corev1.PersistentVolumeAccessMode, *corev1.PersistentVolumeMode, error) {
```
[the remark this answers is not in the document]
ShellyKa13: ok np

pkg/controller/datavolume-controller.go
ShellyKa13: we call this function after making sure: (pvcSpec.VolumeMode == nil || *pvcSpec.VolumeMode == "") is false and pvcVolumeMode == pvcSpec.VolumeMode

ShellyKa13: but I can add a check to be on the safe side is someone will use it in the future wrong..

doc/storageprofile.md
ShellyKa13: just added new line, but will fix :)

ShellyKa13:
@awels `So I am kind of confused by the logic that looks up the volumeMode based on access mode. I can understand accessMode based on volumeMode (for ceph-rbd) but not the other way around.`
up until now getDefaultVolumeMode checked for storageProfile default claim set and took the volume mode or fallback to k8s default. now it is possible to have more then the one option for volume mode, so it seemed like if the user wanted 
specific accessmode it is best to return the first volume mode that supports that accessmode. 
If not what do you think should that function return? same as before? just the first volume mode found?

pkg/controller/datavolume-controller.go
awels: Yeah that is exactly what I was thinking never assume that someone else will not call your function from somewhere that doesn't do the check. Just check in the function to be sure.

awels:
So lets take the ceph-rbd exampled. for accessmode RWX/RWO/ROX it supports block, for RWO/ROX it supports filesystem. So if you pass RWO what do we return? I don't think we specify in the storage profile that RWO should be filesystem right. Its more a preferred so we have RWX, block, what is the second option we added?

doc/upload.md
```diff
@@ -87,6 +87,11 @@ spec:
 ```bash
 kubectl apply -f manifests/example/upload-datavolume.yaml
 ```
+### Create a Data Volume for archive upload
+
+You can also upload an archive. Specifying in the data volume spec: `contentType: archive`
+will mark the datavolume as archive upload and will untar the received content.
```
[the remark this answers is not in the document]
awels: Now that it uses the standard format readers, yes gz and xz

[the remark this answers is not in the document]
ShellyKa13: I will add it to the doc

PR #1974, opened 2021-10-06
Allow increasing DV storage request size

maya-r:
Check with a dryrun attempt to increase the PVC size if it exists.
Removes some unit tests testing that storage size is immutable, in
favour of functional tests that it can be increased.

**Release note**:

```release-note
Propagate datavolume size updates to the PVC
```


maya-r:
Closing as we intend DV spec to remain immutable.

PR #1981, opened 2021-10-11
Follow redirects on import, so mirrors are resolved.

### LINKED ISSUE #1737: CDI should resolve redirections from mirrors before handing it to nbdkit

>
 /kind bug
> /kind enhancement

**What happened**:
When downloading an image from a URL that is backed by a bunch of mirrors, sometimes the importer fails because one of the mirrors is misbehaving or doesn't have the file yet. This is due to nbdkit calling ranges on the URL provided, but if the URL provided is a redirection URL, it will follow it to different mirrors. If all the mirrors are working this is not a problem, however if one is misbehaving this will cause the import to fail.

I expect CDI to resolve the redirects to a real URL before passing it to nbdkit, so it doesn't do a range on many different mirrors, but instead calls the ranges on the resolved URL.

**Environment**:
- Install tools:
- Others:

awels:

Follow redirect before passing to nbdkit, so nbdkit only attempts to connect to a single mirror instead of resolving each fragment. This is especially import for mirrored images. It is possible one mirror is down or broken, and it will cause downloads to fail randomly.

The PR checks in the http redirect to see if a redirect happened, and will switch the URL to the redirected URL before passing it to nbdkit.

Fixes #1737 

**Release note**:

```release-note
BugFix: Switch URL to redirected URL to properly handle mirrorred http sources.
```


pkg/importer/http-datasource.go
```diff
@@ -298,11 +304,11 @@ func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certD
 	klog.V(2).Infof("Attempting to get object %q via http client\n", ep.String())
 	resp, err := client.Do(req)
 	if err != nil {
-		return nil, uint64(0), true, errors.Wrap(err, "HTTP request errored")
+		return nil, uint64(0), true, redirectedURL, errors.Wrap(err, "HTTP request errored")
 	}
 	if resp.StatusCode != 200 {
 		klog.Errorf("http: expected status code 200, got %d", resp.StatusCode)
-		return nil, uint64(0), true, errors.Errorf("expected status code 200, got %d. Status: %s", resp.StatusCode, resp.Status)
+		return nil, uint64(0), true, redirectedURL, errors.Errorf("expected status code 200, got %d. Status: %s", resp.StatusCode, resp.Status)
```
mhenriks, line 311: I don't think `redirectedURL` should be returned in these two cases

awels:
Well nbdkit is doing reasonable behavior from their point of view, it is using the qemu-img-curl plugin to get the data, and that pluging is resolving the url each time. I opened https://bugzilla.redhat.com/show_bug.cgi?id=2013000 to get some eyes on it from their end.

brybacki:
@awels that bugzilla shows some fixes were applied and new nbdkit released. What are your plans.

PR #1982, opened 2021-10-11
Switch builder to using centos:stream9

maya-r:
First PR to change the builder, second PR to change everything else.

**Release note**:

```release-note
NONE
```


maya-r:
Bug resolved but only on centos stream 9, using centos stream 9 blocked on https://bugzilla.redhat.com/show_bug.cgi?id=2028491

PR #1994, opened 2021-10-15
Allow user-specified headers in HTTP data source.

mrnold:
This pull request allows custom HTTP headers to be specified directly in the HTTP data volume spec, or as references to secrets containing headers with information that should not be displayed.

The goal is to get migration from OpenStack working in MTV, which can do the login itself and just pass the token to CDI as a custom header. Hopefully this is easier to digest than the previous attempt in pull request #[392](https://github.com/kubevirt/containerized-data-importer/pull/392/), and it avoids adding a whole new data source.

**Release note**:
```release-note
Allow specification of extra headers in HTTP data source.
```


pkg/apis/core/v1alpha1/types.go
```diff
@@ -165,6 +165,12 @@ type DataVolumeSourceHTTP struct {
 	// CertConfigMap is a configmap reference, containing a Certificate Authority(CA) public key, and a base64 encoded pem certificate
 	// +optional
 	CertConfigMap string `json:"certConfigMap,omitempty"`
+	// ExtraHeaders is a list of strings containing extra headers to include with HTTP transfer requests
+	// +optional
+	ExtraHeaders []string `json:"extraHeaders,omitempty"`
```
mhenriks, line 170: why not a map?

mrnold:
I got as far as adding functional tests faster than I thought I would, so I marked this ready for review. It's still kind of a proposal if there's any better ideas though.

mhenriks:
I think I'm down with this, what do you think @awels?

pkg/importer/http-datasource.go
```diff
@@ -395,3 +420,67 @@ func parseHTTPHeader(resp *http.Response) uint64 {
...
+	secretDir := common.ImporterSecretExtraHeadersDir
+	directories, err := ioutil.ReadDir(secretDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return secretExtraHeaders, nil
+		}
+		return nil, errors.Wrapf(err, "Error listing directories under %s", secretDir)
+	}
+
+	for _, directory := range directories {
```
awels, line 461: would a filepath.Walk make this easier to read?

awels (commented):
Looks good to me, one super minor suggestion.

pkg/importer/http-datasource.go
mrnold: Sure, I tried it out and I think it looks a little cleaner. It is slightly different because it goes through the whole secrets directory recursively instead of only one level deep, but I don't think this should cause any practical trouble.

awels:
/lgtm
@mhenriks I am down, I think you are also?

awels:
@mrnold looks like something changed enough that the verify lane is failing, could you rebase and make generate, that should fix the failure.

PR #2004, opened 2021-10-30
Trigger Host Assisted Cloning for encrypted ocs sc.

pkesavap:
This PR allows storage classes to set the preferred clone method with an annotation on the storage class. This is helpful for known provisioners that want different behavior for certain configurations in the storage class. For instance rook-ceph with encryption enabled, will want to a host assisted clone instead of smart clone.

Signed-off-by: Prajith Kesava Prasad <pkesavap@redhat.com>

Fixes # 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -59,7 +59,7 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
 
 	storageProfile.Status.StorageClass = &sc.Name
 	storageProfile.Status.Provisioner = &sc.Provisioner
-	storageProfile.Status.CloneStrategy = storageProfile.Spec.CloneStrategy
+	storageProfile.Status.CloneStrategy = r.reconcileCloneStrategy(sc, storageProfile.Spec.CloneStrategy)
 
 	var claimPropertySets []cdiv1.ClaimPropertySet
 
@@ -130,6 +130,25 @@ func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageCl
 		}
 	}
 	return claimPropertySets
+}
+
+func (r *StorageProfileReconciler) reconcileCloneStrategy(sc *storagev1.StorageClass, clonestrategy *cdiv1.CDICloneStrategy) *cdiv1.CDICloneStrategy {
+
+	if clonestrategy == nil {
+		if sc.Annotations["cdi.kubevirt.io/clone-strategy"] == "copy" {
+			strategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyHostAssisted)
+			return &strategy
+		} else if sc.Annotations["cdi.kubevirt.io/clone-strategy"] == "snapshot" {
+			strategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategySnapshot)
+			return &strategy
+		} else if sc.Annotations["cdi.kubevirt.io/clone-strategy"] == "csi-clone" {
+			strategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyCsiClone)
+			return &strategy
+		} else {
+			return clonestrategy
+		}
+	}
+	return clonestrategy
 }
 
 func (r *StorageProfileReconciler) createEmptyStorageProfile(sc *storagev1.StorageClass) (*cdiv1.StorageProfile, error) {
--- a/pkg/controller/storageprofile-controller_test.go
+++ b/pkg/controller/storageprofile-controller_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -153,6 +154,30 @@ var _ = Describe("Storage profile controller reconcile loop", func() {
 		Expect(len(updatedSp.Status.ClaimPropertySets)).To(Equal(0))
 		Expect(updatedSp.Spec.ClaimPropertySets).To(Equal(claimPropertySets))
 	})
+
+	table.DescribeTable("should create clone strategy", func(cloneStrategy cdiv1.CDICloneStrategy) {
+		storageClass := createStorageClass(storageClassName, map[string]string{AnnDefaultStorageClass: "true"})
+
+		storageClass.Annotations["cdi.kubevirt.io/clone-strategy"] = string(cloneStrategy)
+		reconciler := createStorageProfileReconciler(storageClass)
+		_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: storageClassName}})
+		Expect(err).ToNot(HaveOccurred())
+
+		storageProfileList := &cdiv1.StorageProfileList{}
+		err = reconciler.client.List(context.TODO(), storageProfileList, &client.ListOptions{})
+
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(storageProfileList.Items)).To(Equal(1))
+
+		sp := storageProfileList.Items[0]
+		Expect(*sp.Status.StorageClass).To(Equal(storageClassName))
+		Expect(*sp.Status.CloneStrategy).To(Equal(cloneStrategy))
+	},
+		table.Entry("None", cdiv1.CDICloneStrategy(cdiv1.CloneStrategyHostAssisted)),
+		table.Entry("Snapshot", cdiv1.CDICloneStrategy(cdiv1.CloneStrategySnapshot)),
+		table.Entry("Clone", cdiv1.CDICloneStrategy(cdiv1.CloneStrategyCsiClone)),
+	)
+
 })
 
 func createStorageProfileReconciler(objects ...runtime.Object) *StorageProfileReconciler {
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -531,6 +531,70 @@ var _ = Describe("all clone tests", func() {
 			})
 			ClonerBehavior(cloneStorageClassName, "csivolumeclone")
 		})
+
+		Context("CloneStrategy on storageclass annotation", func() {
+			cloneType := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyCsiClone)
+
+			BeforeEach(func() {
+				if !f.IsCSIVolumeCloneStorageClassAvailable() {
+					Skip("CSI Volume Clone is not applicable")
+				}
+				cloneStorageClassName = f.CsiCloneSCName
+				By(fmt.Sprintf("Get original storage profile: %s", cloneStorageClassName))
+				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(storageProfile.Spec.CloneStrategy).To(BeNil())
+				Expect(storageProfile.Status.CloneStrategy).To(BeNil())
+
+				storageclass, err := f.K8sClient.StorageV1().StorageClasses().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				if storageclass.GetAnnotations() == nil {
+					storageclass.SetAnnotations(make(map[string]string))
+				}
+
+				storageclass.Annotations["cdi.kubevirt.io/clone-strategy"] = string(cloneType)
+				_, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageclass, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() cdiv1.CDICloneStrategy {
+					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					if storageProfile.Status.CloneStrategy == nil {
+						return ""
+					}
+					return *storageProfile.Status.CloneStrategy
+				}, time.Minute, time.Second).Should(Equal(cloneType))
+
+			})
+			AfterEach(func() {
+				By("[AfterEach] Restore the storage class - remove annotation ")
+				storageclass, err := f.K8sClient.StorageV1().StorageClasses().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				if storageclass.GetAnnotations() != nil {
+					delete(storageclass.Annotations, "cdi.kubevirt.io/clone-strategy")
+				}
+				_, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageclass, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() *cdiv1.CDICloneStrategy {
+					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					return storageProfile.Status.CloneStrategy
+				}, time.Minute, time.Second).Should(BeNil())
+			})
+			It("Should clone  with correct strategy from storageclass annotation ", func() {
+
+				pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
+				pvcDef.Spec.StorageClassName = &cloneStorageClassName
+				pvcDef.Namespace = f.Namespace.Name
+
+				sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile+"; chmod 660 "+testBaseDir+testFile)
+
+				targetDV := utils.NewCloningDataVolume("target-pvc", "1Gi", sourcePvc)
+
+				dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(utils.GetCloneType(f.CdiClient, dataVolume)).To(Equal("csivolumeclone"))
+			})
+		})
 	})
 
 	var _ = Describe("Validate creating multiple clones of same source Data Volume", func() {


pkg/controller/datavolume-controller.go
```diff
@@ -532,6 +532,16 @@ func (r *DatavolumeReconciler) selectCloneStrategy(datavolume *cdiv1.DataVolume,
 		return NoClone, nil
 	}
 
+	storageClass, err := GetStorageClassByName(r.client, pvcSpec.StorageClassName)
+
+	if err != nil {
+		return NoClone, err
+	}
+
+	if strings.HasPrefix(storageClass.Provisioner, "openshift-storage") && storageClass.Parameters["encrypted"] == "true" {
```
aglitke, line 541: I would prefer that the recommended cloning strategy actually appears in the StorageProfile status.  This means that this logic should be part of the StorageProfile controller's reconcile loop.  If we do it there, then there is no need to special case it here.

maya-r:
can you elaborate more on why host-assisted clone is necessary?

pkg/controller/datavolume-controller.go
brybacki: I agree. StorageProfile has a field called cloneStrategy. And it looks like this logic fits nicely to storageProfile. 

That way an end user can see which strategy will be applied by checking the storageprofile status.  Also the cloning in datavolume will remain simple.

pkg/controller/storageprofile-controller.go
```diff
@@ -141,7 +141,13 @@ func (r *StorageProfileReconciler) reconcileAccessModes(sc *storagev1.StorageCla
 }
 
 func (r *StorageProfileReconciler) createEmptyStorageProfile(sc *storagev1.StorageClass) (*cdiv1.StorageProfile, error) {
-	storageProfile := MakeEmptyStorageProfileSpec(sc.Name)
+	var storageProfile *cdiv1.StorageProfile
+	if sc.Annotations["cdi.kubevirt.io/storage-profile"] == "host-assisted" {
```
aglitke, line 145: The annotation should be in a constant.  Also, since there is more to a storage profile than just the clone strategy I think we should use "cdi.kubevirt.io/clonestrategy".  Since the term used in CDI is "copy" we should prefer that to "host-assisted".

aglitke (changes requested):
Looks pretty good.  Just requesting a change to the annotation text.

pkg/controller/storageprofile-controller.go
mhenriks: This code seems to only handle the case when the storage profile does not exist.  It does not handle adding the annotation to the SC when the profile exists.  It also does not handle when the annotation is removed (should set clone strategy back to default)

pkesavap: @mhenriks this change, sets the cloneStrategy as host assisted if a storageClass has an annotation of cdi.kubevirt.io/storage-profile=host-assisted
This way any vendors(even non ODF) who wants to do a host-assisted clone, can annotate their storageClass with the above annotation and CDI would do a host-assisted one, so we do not need to worry about adding the annotation  to the StorageClass, based on the StorageProfile, since storageProfile would only execute Host-assisted, if StorageClass has that annotaion.

And since this logic resides under the reconcile loop, if the annotation is  removed, then it would execute the alternate condition and proceed with Line 149[https://github.com/kubevirt/containerized-data-importer/pull/2004/commits/23c033115a4c88e879bee03169c7bbd27327d60e#diff-3a3e2d2d37a456bef2fad553faed8c75f9fecd62ad31cc3fbe8179ba670fd3eaR149]

This is my understanding. Please correct me if this is wrong. :-)  @aglitke @mhenriks

pkesavap:
> can you elaborate more on why host-assisted clone is necessary?

Hi @maya-r ,
Apologies for the delay in this reply.
We are going with host-assisted because the smart clone (as of now) does not support cross-namespace cloning of encrypted-PV,
 i.e cloning an encrypted pv from one namespace to another is not supported.

pkg/controller/datavolume-controller.go
pkesavap: updated with new changes

awels:
So we do support cross namespace smart/csi clone. What we do in practice is create a clone in the same namespace as the source, then transfer that PVC to the new namespace by removing the PVC from the source (while keeping the PV which is not namespaced) and then creating a new PVC in the target namespace which gets bound to the PV.

awels:
> 
> Apologies I should have added the whole explanation in my previous comment. Yes that is correct, and the problem with that way is that ceph-CSI still identifies the new PVC as the old volume since the underlying PV is essentially the same, and this leads to a problem where ceph-csi 's metadata still points to the old namespace of the PV, and this causes further problems when interacting with other components that require namespace specific permissions, like for eg, fetching passphrase for the encrypted PV from a vault which is namespace specific, or others.

Okay, I was afraid that might be a problem. And the reason you cannot change the spec of the storageProfile because that is on a provisioner level, and you can have multiple storage classes using the same provisioner, where you only want to change the behavior of a single storage class while maintaining the other clone behaviour for the rest of the storage classes with that particular provisioner.

pkg/controller/storageprofile-controller.go
brybacki: it looks like you have added the code in the place which is only invoked at profile init time,
```
r.getStorageProfile(sc) -> 
  	**if k8serrors.IsNotFound(err) {** 
	  			storageProfile, err = r.createEmptyStorageProfile(sc)
	 }
}
...
```
The correct place would be to add new code in a main recocile loop in line 61. You can add `r.reconcileCloneStrategy()` the same way we have `reconcileAccessModes` and `reconcileVolumeMode`. 

Anyway, it would be better to rebase first, as a change to storageProfile reconciliation was merged last week. That PR also added a series of unit tests `storageprofile-controller_test.go` - which might be really helpful in experimenting, and now it will be really easy to add new test.

brybacki: @pkesavap and I totally agree that implementing it with an annotation is a good - non intrusive and vendor agnostic way. I just think that `cdi.kubevirt.io/clonestrategy` is better than `cdi.kubevirt.io/storage-profile`.

brybacki (changes requested):
ok, it looks almost good, I suggest to rebase first, as new code merged in https://github.com/kubevirt/containerized-data-importer/pull/1969 changes a few things in storage profile reconcile loop and makes it really easy to test and experiment (as a whole suite of unit tests for storage-profile was added)

Also, I have one comment regarding the code to reconcile CloneStrategy.

mhenriks:
@pkesavap @brybacki suggestions for functional tests:

If rook-ceph-block StorageClass does not exist, skip.

1. Test adding/removing annotation to rook-ceph-block
- Confirm StorageProfile exists and CloneStrategy not set
- Add annotation
- Confirm CloneStrategy eventually set to "copy"
- Remove annotation
- Confirm CloneStrategy unset

2.  Test creating a new StorageClass with annotation
- Create copy of rook-ceph-block SC with new name and annotation set
- Confirm StorageProfile eventually created with CloneStrategy eventually set to "copy"

pkg/controller/storageprofile-controller.go
```diff
@@ -132,8 +133,22 @@ func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageCl
...
 
+func (r *StorageProfileReconciler) reconcileCloneStrategy(sc *storagev1.StorageClass, clonestrategy *cdiv1.CDICloneStrategy) {
+	if sc.GetAnnotations()["cdi.kubevirt.io/clone-strategy"] != "" {
+		if clonestrategy == nil {
+			clonestrategy = nil // default to nil
+		}
+	}
+}
+
 func (r *StorageProfileReconciler) createEmptyStorageProfile(sc *storagev1.StorageClass) (*cdiv1.StorageProfile, error) {
```
brybacki, line 144: the `reconcileStorageProfile` is written as follows:
1. get the actual profile as storageProfile (or empty if not found) so you should not chnage the `createEmptyStorageProfile` unless you have some sensible and well documented defautls

2. gather facts from environment and apply to storageProfile (currently there is only one such method `reconcilePropertySets`) and you are adding a new one `reconcileCloneStrategy`. And here is the best place to apply all changes. It will work for a new obejct and for a changes during runtime, adding/removing of annotation and so on. 

3. Update the StorageProfile if anything was changed (that is why we keep a cloe of storage progile named `prevStorageProfile` - to detect if there are changes, and kubernets object needs to be updated)

pkg/controller/storageprofile-controller.go
```diff
@@ -132,8 +132,21 @@ func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageCl
...
+		if sc.Annotations["cdi.kubevirt.io/clone-strategy"] == "host-assisted" {
+			strategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyHostAssisted)
+			return &strategy
+		}
+	}
+	return clonestrategy
+}
+
 func (r *StorageProfileReconciler) createEmptyStorageProfile(sc *storagev1.StorageClass) (*cdiv1.StorageProfile, error) {
+
```
pkesavap, line 147: remove space

pkg/controller/storageprofile-controller.go
```diff
@@ -132,8 +132,21 @@ func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageCl
...
 	storageProfile := MakeEmptyStorageProfileSpec(sc.Name)
+
```
pkesavap, line 149: remove space

pkg/controller/storageprofile-controller.go
```diff
@@ -132,6 +132,17 @@ func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageCl
 	return claimPropertySets
 }
 
+func (r *StorageProfileReconciler) reconcileCloneStrategy(sc *storagev1.StorageClass, clonestrategy *cdiv1.CDICloneStrategy) *cdiv1.CDICloneStrategy {
+
+	if clonestrategy == nil {
+		if sc.Annotations["cdi.kubevirt.io/clone-strategy"] == "host-assisted" {
```
brybacki, line 138: This looks very good, but as we are adding the clone-strategy annotation I would like to see all 3 values for cdiv1.CDICloneStrategy covered here. 
So:
```
	// CloneStrategyHostAssisted specifies slower, host-assisted copy
	CloneStrategyHostAssisted = "copy"

	// CloneStrategySnapshot specifies snapshot-based copying
	CloneStrategySnapshot = "snapshot"

	// CloneStrategyCsiClone specifies csi volume clone based cloning
	CloneStrategyCsiClone = "csi-clone"
```

Also please be consistent, with configuration options so use `copy, snapshot and csi-clone`.

It also should be really easy to extend test to (use `DescribeTable` with multiple Entries like here: pkg/apiserver/webhooks/cdi-validate_test.go:86)

pkesavap: @brybacki  i have addressed the changes.
And you mean to follow the DescribeTable structure in implementing the testcases in storageprofile-controller_test.go? Right now i have used It() -> expect model

pkg/controller/storageprofile-controller_test.go
```diff
@@ -153,6 +153,36 @@ var _ = Describe("Storage profile controller reconcile loop", func() {
 		Expect(len(updatedSp.Status.ClaimPropertySets)).To(Equal(0))
 		Expect(updatedSp.Spec.ClaimPropertySets).To(Equal(claimPropertySets))
 	})
+
+	It("Should update storage profile with editted clone strategy", func() {
```
brybacki, line 157:
```suggestion
	table.DescribeTable("Should update storage profile ", func(strategy string) {
```

pkg/controller/storageprofile-controller_test.go
```diff
@@ -153,6 +153,36 @@ var _ = Describe("Storage profile controller reconcile loop", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		storageProfileList = &cdiv1.StorageProfileList{}
+		err = reconciler.client.List(context.TODO(), storageProfileList, &client.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(storageProfileList.Items)).To(Equal(1))
+		updatedSp := storageProfileList.Items[0]
+		Expect(*updatedSp.Status.StorageClass).To(Equal(storageClassName))
+		Expect(len(updatedSp.Status.ClaimPropertySets)).To(Equal(0))
+		Expect(updatedSp.Spec.CloneStrategy).To(Equal(&cloneStrategy))
+	})
```
brybacki, line 184:
```suggestion
		},
		table.Entry("with clone strategy copy", cdiv1.CloneStrategyHostAssisted),
		table.Entry("with clone strategy snaphsot", cdiv1.CloneStrategySnapshot),
		table.Entry("with clone strategy csi-clone", cdiv1.CloneStrategyCsiClone),
	)
```

pkg/controller/storageprofile-controller_test.go
```diff
@@ -153,6 +153,36 @@ var _ = Describe("Storage profile controller reconcile loop", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		storageProfileList := &cdiv1.StorageProfileList{}
+		err = reconciler.client.List(context.TODO(), storageProfileList, &client.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(storageProfileList.Items)).To(Equal(1))
+		sp := storageProfileList.Items[0]
+		Expect(*sp.Status.StorageClass).To(Equal(storageClassName))
+		Expect(len(sp.Status.ClaimPropertySets)).To(Equal(0))
+
+		cloneStrategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyHostAssisted)
```
brybacki, line 169:
```suggestion
		cloneStrategy := cdiv1.CDICloneStrategy(strategy)
```

brybacki (commented):
The code is now ok.

I did add a suggestion to the test. To show you what I mean.

pkesavap:
> The code is now ok.
> 
> I did add a suggestion to the test. To show you what I mean.

I have implemented the tests that way, PTAL.

brybacki:
The code and the unit test are now OK. But I see that @mhenriks suggested a functional test.

brybacki:
The functional tests for clone and storagefile are in cloner_test.go. The new test could land after the line 533 there.

Structure would be something like this (I did copy of a simple clone test and added some TODOs) feel free to message me directly in case of any questions:
```go

		Context("CloneStrategy on storageclass annotation", func() {
			cloneType:= "csi-clone"

			BeforeEach(func() {
				if !f.IsCSIVolumeCloneStorageClassAvailable() {
					Skip("CSI Volume Clone is not applicable")
				}
				cloneStorageClassName = f.CsiCloneSCName
				By(fmt.Sprintf("Get original storage profile: %s", cloneStorageClassName))
				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
				Expect(storageProfile.Spec.CloneStrategy).To(BeNil())
				Expect(storageProfile.Status.CloneStrategy).To(BeNil())

				//TODO:  get storageClass `cloneStorageClassName` and annotate with `cloneType`
			})
			AfterEach(func() {
				By("[AfterEach] Restore the storage class - remove annotation ")
				// TODO: Remove annotation from storageClass `cloneStorageClassName`
				// TODO: verify that the storageProfile `cloneStorageClassName` has empty storageProfile.Status.CloneStrategy
			})
			It("Should clone  with correct strategy from storageclass annotation ", func() {
				//TODO: get storageProfile and verify taht the cloneStrategy is set to `cloneType`

				pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
				pvcDef.Spec.StorageClassName = &cloneStorageClassName
				pvcDef.Namespace = f.Namespace.Name

				sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile+"; chmod 660 "+testBaseDir+testFile)
				doFileBasedCloneTest(f, pvcDef, f.Namespace, "target-dv", "1Gi")

				Expect(utils.GetCloneTypeByDvName(f.CdiClient, f.Namespace.Name,"target-dv")).To(Equal(cloneType))
			})
		})
```

tests/cloner_test.go
```diff
@@ -531,6 +531,65 @@ var _ = Describe("all clone tests", func() {
...
+				}
+				cloneStorageClassName = f.CsiCloneSCName
+				By(fmt.Sprintf("Get original storage profile: %s", cloneStorageClassName))
+				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(storageProfile.Spec.CloneStrategy).To(BeNil())
+				Expect(storageProfile.Status.CloneStrategy).To(BeNil())
+
+				//TODO:  get storageClass `cloneStorageClassName` and annotate with `cloneType`
+				storageProfile.Spec.CloneStrategy = &cloneType
```
brybacki, line 550: we want to test if the Storage Profile logic that you added works, so I you should only add here the annotation
"cdi.kubevirt.io/clone-strategy" = "copy" to storage class with name "cloneStorageClassName"

and the storageProfile Status should be updated by the code you have added

tests/cloner_test.go
```diff
@@ -531,6 +531,65 @@ var _ = Describe("all clone tests", func() {
...
+
+			})
+			AfterEach(func() {
+				By("[AfterEach] Restore the storage class - remove annotation ")
+				// TODO: Remove annotation from storageClass `cloneStorageClassName`
```
brybacki, line 555: and here the same, no need to edit storageProfile, just remove annotation from storageClass and check if storageProfile was updated in an `Eventually()` loop

tests/cloner_test.go
```diff
@@ -531,6 +531,65 @@ var _ = Describe("all clone tests", func() {
...
+
+			})
+			It("Should clone  with correct strategy from storageclass annotation ", func() {
+				//TODO: get storageProfile and verify taht the cloneStrategy is set to `cloneType`
+				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(storageProfile.Spec.CloneStrategy).ToNot(BeNil())
+				Expect(storageProfile.Status.CloneStrategy).ToNot(BeNil())
+				Expect(*storageProfile.Spec.CloneStrategy).To(Equal(cloneType))
+				Expect(*storageProfile.Status.CloneStrategy).To(Equal(cloneType))
```
brybacki, line 577: I think that only `Status` should contain cloneType, `Spec` should be nil

tests/cloner_test.go
```diff
@@ -531,6 +531,65 @@ var _ = Describe("all clone tests", func() {
...
+				pvcDef.Namespace = f.Namespace.Name
+
+				sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile+"; chmod 660 "+testBaseDir+testFile)
+				doFileBasedCloneTest(f, pvcDef, f.Namespace, "target-dv", "1Gi")
+
+				targetDV := utils.NewCloningDataVolume("target-pvc", "1Gi", sourcePvc)
+
+				dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(utils.GetCloneType(f.CdiClient, dataVolume)).To(Equal(cloneType))
```
brybacki, line 590: That should be good

tests/cloner_test.go
```diff
@@ -531,6 +531,73 @@ var _ = Describe("all clone tests", func() {
...
+				}
+				cloneStorageClassName = f.CsiCloneSCName
+				By(fmt.Sprintf("Get original storage profile: %s", cloneStorageClassName))
+				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(storageProfile.Spec.CloneStrategy).To(BeNil())
+				Expect(storageProfile.Status.CloneStrategy).To(BeNil())
+
+				//TODO:  get storageClass `cloneStorageClassName` and annotate with `cloneType`
+				storageProfile, err = f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
```
awels, line 550: Shouldn't we be setting the annotation on the storage class here, instead of the storage profile. Something like this:
```go
  storageclass, err := f.K8sClient.StorageV1().StorageClasses().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
  if storageClass.GetAnnotations() == nil {
    storageClass.SetAnnotations(make(map[string]string))
  }
  storageClass.Annotations["cdi.kubevirt.io/clone-strategy"] = string(cloneType)
  _, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageClass, metav1.UpdateOptions{})
  Eventually(func() string {
    storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
    return storageProfile.Status.CloneStrategy
  }, time.Minute, time.Second).Should(Equal(cloneType))
```
This updates the storage class, then poll the storage profile for a minute every second to check if the clone strategy matches what we set on the storage class.

tests/cloner_test.go
awels: Yeah basically do the reverse of what I showed for the BeforeEach, remove the annotation, and wait until the storage profile no longer has the Status.CloneStrategy. We are not setting the clone strategy in the Spec, so we don't care about that.

tests/cloner_test.go
awels: We just ensured in the before each that the storage profile is what we want, so no need to check again.

tests/cloner_test.go
```diff
@@ -531,6 +531,73 @@ var _ = Describe("all clone tests", func() {
...
+				Expect(storageProfile.Status.CloneStrategy).ToNot(BeNil())
+				Expect(*storageProfile.Spec.CloneStrategy).To(Equal(cloneType))
+				Expect(*storageProfile.Status.CloneStrategy).To(Equal(cloneType))
+
+				pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
+				pvcDef.Spec.StorageClassName = &cloneStorageClassName
+				pvcDef.Namespace = f.Namespace.Name
+
+				sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile+"; chmod 660 "+testBaseDir+testFile)
+				doFileBasedCloneTest(f, pvcDef, f.Namespace, "target-dv", "1Gi")
```
awels, line 592: You are trying to clone before defining the targetDV. I don't think you need this line.

tests/cloner_test.go
pkesavap: yep this makes sense, done

tests/cloner_test.go
pkesavap: done

tests/cloner_test.go
awels: I mean you don't need to look up the storage profile here and do the above expects.

tests/cloner_test.go
```diff
@@ -531,6 +531,78 @@ var _ = Describe("all clone tests", func() {
...
+					Skip("CSI Volume Clone is not applicable")
+				}
+				cloneStorageClassName = f.CsiCloneSCName
+				By(fmt.Sprintf("Get original storage profile: %s", cloneStorageClassName))
+				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(storageProfile.Spec.CloneStrategy).To(BeNil())
+				Expect(storageProfile.Status.CloneStrategy).To(BeNil())
+
+				//TODO:  get storageClass `cloneStorageClassName` and annotate with `cloneType`
```
awels, line 549: I see you have some `//TODO` here and below, but you just did them.

awels (commented):
Looks good, just a couple of cleanup tasks and we should be good to go.

awels (commented):
/lgtm
/approve

brybacki:
That lane failure says:
`            Message: "storageprofiles.cdi.kubevirt.io \"manual\" not found",`

That is strange, I'll try to look into the problem.

tests/cloner_test.go
```diff
@@ -531,6 +531,75 @@ var _ = Describe("all clone tests", func() {
...
+		Context("CloneStrategy on storageclass annotation", func() {
+			cloneType := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyCsiClone)
+
+			BeforeEach(func() {
+				if !f.IsCSIVolumeCloneStorageClassAvailable() {
+					Skip("CSI Volume Clone is not applicable")
+				}
+				cloneStorageClassName = f.CsiCloneSCName
+				By(fmt.Sprintf("Get original storage profile: %s", cloneStorageClassName))
+				storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), storageClassName, metav1.GetOptions{})
```
brybacki, line 544: here it should be `cloneStorageClassName` not `storageClassName`

pkesavap: ah got it, making changes now

brybacki:
looks like a random fail:
/test pull-containerized-data-importer-e2e-k8s-1.21-hpp

brybacki:
/test pull-containerized-data-importer-e2e-k8s-1.21-hpp

What is that? Looks like not related to the change
```
Error from server (InternalError): error when creating "https://raw.githubusercontent.com/kubevirt/hostpath-provisioner-operator/main/deploy/webhook.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": x509: certificate signed by unknown authority
```

awels:
Yeah not related to the PR at all. Might be a timing issue retest is fine.

tests/cloner_test.go
```diff
@@ -531,6 +531,75 @@ var _ = Describe("all clone tests", func() {
...
+				}
+
+				storageclass.Annotations["cdi.kubevirt.io/clone-strategy"] = string(cloneType)
+				_, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageclass, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() string {
+					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+
+					return string(*storageProfile.Status.CloneStrategy)
```
awels, line 562: Looks like it is possible for storageProfile.Status.CloneStrategy to be nil, I guess if it is nil just return a blank string.
```go
if storageProfile.Status.CloneStrategy == nil {
  return ""
}
return string(*storageProfile.Status.CloneStrategy)
```

pkesavap: noted

awels: Its the reason the ceph lane failed btw. so it needs to get fixed before it can pass CI

pkesavap: right, i have sent in the fix

awels: Okay so the test failed again, cloneType is specific type and not string. So it actually found the right value, just in the wrong type, need to cast it to a string.
```
}, time.Minute, time.Second).Should(Equal(string(cloneType)))
```

tests/cloner_test.go
```diff
@@ -531,6 +531,77 @@ var _ = Describe("all clone tests", func() {
...
+				}
+
+				storageclass.Annotations["cdi.kubevirt.io/clone-strategy"] = string(cloneType)
+				_, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageclass, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() *cdiv1.CDICloneStrategy {
+					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					if storageProfile.Status.CloneStrategy == nil {
+						return nil
```
pkesavap, line 562: In that case i have updated this line like this, instead of retuning "", but i dont know if this was needed as its implicitly understood. @awels

brybacki: you compare to CDICloneStrategy in the `				}, time.Minute, time.Second).Should(Equal(cloneType))`

Not it compares address wioth value (which will fail). This should be correct

```
				Eventually(func() cdiv1.CDICloneStrategy {
					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
					if storageProfile.Status.CloneStrategy == nil {
						return ""
					}
					return *storageProfile.Status.CloneStrategy
				}, time.Minute, time.Second).Should(Equal(cloneType))
```

tests/cloner_test.go
```diff
@@ -531,6 +531,77 @@ var _ = Describe("all clone tests", func() {
...
+				storageclass, err := f.K8sClient.StorageV1().StorageClasses().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				if storageclass.GetAnnotations() == nil {
+					storageclass.SetAnnotations(make(map[string]string))
+				}
+
+				storageclass.Annotations["cdi.kubevirt.io/clone-strategy"] = string(cloneType)
+				_, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageclass, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() *cdiv1.CDICloneStrategy {
```
brybacki, line 558:
```suggestion
				Eventually(func() cdiv1.CDICloneStrategy {
					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
					Expect(err).ToNot(HaveOccurred())
					if storageProfile.Status.CloneStrategy == nil {
						return ""
					}
					return *storageProfile.Status.CloneStrategy
				}, time.Minute, time.Second).Should(Equal(cloneType))
```

tests/cloner_test.go
brybacki: actually it should be 
				Expect(utils.GetCloneType(f.CdiClient, dataVolume)).To(Equal("csivolumeclone"))

tests/cloner_test.go
```diff
@@ -531,6 +531,77 @@ var _ = Describe("all clone tests", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				if storageclass.GetAnnotations() != nil {
+					delete(storageclass.Annotations, "cdi.kubevirt.io/clone-strategy")
+				}
+				_, err = f.K8sClient.StorageV1().StorageClasses().Update(context.TODO(), storageclass, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() cdiv1.CDICloneStrategy {
+					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+
```
brybacki, line 580:
```
					if storageProfile.Status.CloneStrategy == nil {
						return ""
					}
```

brybacki: actually this way is correct
```
				Eventually(func() *cdiv1.CDICloneStrategy {
					storageProfile, err := f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), cloneStorageClassName, metav1.GetOptions{})
					return storageProfile.Status.CloneStrategy
				}, time.Minute, time.Second).Should(BeNil())
```

pkesavap: just rechecking does Eventually( expect cdiv1.CDICloneStrategy  with '*' ? or is it the otherway around? because i think it needs to be same with the eventually at BeforeEach() ?

awels: `storageProfile.Status.CloneStrategy` is a pointer, and the Eventually func wants the type. So you have dereference it with *storageProfile.Status.CloneStrategy. But before you dereference you have to make sure it is not nil.

brybacki: but I have also changed this -> now eventually wants `*cdiv1.CDICloneStrategy `

brybacki: but we can do it either way, just need to correct return type, the assertion and the return with nil check

awels:
Btw, if you want quicker feedback on if the test is working. You can run it locally with
```bash
make cluster-up && make cluster-sync
```
This will start a cluster, and sync the current code into the cluster.
Then You can change the `It` part of the functional test to `FIt` which tells the test to focus on that particular test. Then make test-functional to run the functional test. This should run just the one functional test. Once it passed, remove the `F` and push. That way you don't have to wait hours to see if what you did fixed the problem.

pkesavap:
> Btw, if you want quicker feedback on if the test is working. You can run it locally with
> 
> ```shell
> make cluster-up && make cluster-sync
> ```
> 
> This will start a cluster, and sync the current code into the cluster. Then You can change the `It` part of the functional test to `FIt` which tells the test to focus on that particular test. Then make test-functional to run the functional test. This should run just the one functional test. Once it passed, remove the `F` and push. That way you don't have to wait hours to see if what you did fixed the problem.

Yeah, unfortunately my local setup was corrupted in terms of docker and go environment, so i could not run in there, but i was working on running this on vagrant vm, side by side while i pushed the code here.

brybacki:
some timing issue

PR #2009, opened 2021-11-07
Degraded/unusual restartcount alerts

akalenyu:
Add alerts for:
- Unusual restart count on a DV - 3 is the threshold
- CDI degraded - to spare us from yelling during installation/upgrade,
CDI will only be considered degraded (`cdi_cr_ready` == 0) when the CR has both:
    1. stopped progressing (`ConditionProgressing` == `false`)
    2. is not available (`ConditionAvailable` == `false`)

Fixes #

**Release note**:

```release-note
CDI Degraded / DV unusual restartcount alerts
```


tests/import_test.go
```diff
@@ -960,7 +961,36 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 			Expect(err).NotTo(HaveOccurred())
 			restarts := dv.Status.RestartCount
 			return restarts
-		}, timeout, pollingInterval).Should(BeNumerically(">=", 1))
+		}, timeout, pollingInterval).Should(BeNumerically(">", common.UnusualRestartCountThreshold))
+
+		if utils.IsOpenshift(f.K8sClient) {
```
awels, line 966: Are we sure this is openshift only, we could enable prometheus on kubevirtci and get the same information I think.

tests/operator_test.go
```diff
@@ -674,7 +672,7 @@ var _ = Describe("ALL Operator tests", func() {
 		var _ = Describe("Alert tests", func() {
 			f := framework.NewFramework("alert-tests")
 
-			It("CdiOperatorDown alert firing when operator scaled down", func() {
+			It("CDIOperatorDown alert firing when operator scaled down", func() {
 				if !utils.IsOpenshift(f.K8sClient) {
```
awels, line 676: Same not sure if this openshift specific at this point. kubevirtci has a prometheus option now.

tests/operator_test.go
```diff
@@ -748,6 +721,38 @@ var _ = Describe("ALL Operator tests", func() {
 				err = utils.WaitForDeploymentReplicasReady(f.K8sClient, f.CdiInstallNs, deploymentName)
 				Expect(err).ToNot(HaveOccurred())
 			})
+
+			It("CDI ready metric value as expected when ready to use", func() {
+				if !utils.IsOpenshift(f.K8sClient) {
```
awels, line 726: Same not sure if this openshift specific at this point. kubevirtci has a prometheus option now.

tests/utils.go
```diff
@@ -234,3 +237,45 @@ func findConditionByType(conditionType cdiv1.DataVolumeConditionType, conditions
...
 }
+
+// MakePrometheusHTTPRequest makes a request to the prometheus api and returns the response
+func MakePrometheusHTTPRequest(f *framework.Framework, endpoint string) *http.Response {
+	var token, host string
+	var err error
+	var resp *http.Response
+
+	gomega.Eventually(func() bool {
+		host, err = RunOcCommand(f, "-n", "openshift-monitoring", "get", "route", "prometheus-k8s", "--template", "{{.spec.host}}")
```
awels, line 248: If we use kubevirtci prometheus, the namespace will likely be different. Maybe investigate adding a prometheus-namespace to the function test suite, and use that to figure out which namespace to get information from.

pkg/util/util.go
```diff
@@ -358,3 +359,25 @@ func Md5sum(filePath string) (string, error) {
 	hashInBytes := hash.Sum(nil)[:16]
 	return hex.EncodeToString(hashInBytes), nil
 }
+
+// IsCdiAvailable returns whether the CDI installation is available for use
+func IsCdiAvailable(cr *cdiv1.CDI) bool {
```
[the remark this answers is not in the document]
akalenyu: im not sure about firing metrics from the operator SDK, feels a little over its responsibility
also keep in mind the server gets created on our side implicitly https://github.com/kubevirt/containerized-data-importer/blob/main/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go#L174-L177

yeah those funcs definitely belong there

tests/import_test.go
akalenyu: The reason I opted in for the skipNotOpenShift approach is that we didn't have 1.21 lanes,
plus to spare some hacking by grabbing the route. but it will for sure be possible to have these running on k8s as well

tests/utils.go
akalenyu: I already have that ns on the operator's env vars, and we could also check if openshift and adjust accordingly as fallback

but yeah mm this will require some changes to run on both, would prefer to tackle it later on and not in this PR, would that be ok?

tests/import_test.go
brybacki: I would create a funcion `isMonitoringEnabled` or `isPrometheusAvailable`, and if it is hard to correctly check for monitoring/prometheus now, you can just do `utils.IsOpenshift(f.K8sClient)` inside the implementation and leave your description as a comment.
Then you clearly state the intent, and it is easy to find and upgrade the implementation in the future.

pkg/util/util.go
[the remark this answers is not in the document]
brybacki: But it is possible way to just calculate and provide metrics on http endpoint without pubshing them. Then the operators SDK can have the metrics ready on http enpoint (the prometheus registry), and the user of the SDK can decide to publish them to the prometheus system (by instantiating the ServiceMonitor and the PrometheusRules). 

WDYT?

pkg/operator/controller/controller_test.go
```diff
@@ -337,11 +337,12 @@ var _ = Describe("Controller", func() {
 				Expect(err).ToNot(HaveOccurred())
 				rule = obj.(*promv1.PrometheusRule)
 				cdiDownAlert := promv1.Rule{
-					Alert: "CdiOperatorDown",
-					Expr:  intstr.FromString("cdi_num_up_operators == 0"),
+					Alert: "CDIOperatorDown",
+					Expr:  intstr.FromString("kubevirt_cdi_num_up_operators == 0"),
```
brybacki, line 341: Is there a naming scheme for Kubernetes  metrics? Prometheus suggests to have a type of metrics at the end (_total, _seconds_total, _bytes)
this name could be one of? `kubevirt_cdi_up_operators_number`, `kubevirt_cdi_up_operators_count`, `kubevirt_cdi_up_operators`

brybacki: I have found example here: 
https://github.com/kubernetes/kubernetes/blob/master/test/instrumentation/testdata/stable-metrics-list.yaml

tests/datavolume_test.go
```diff
@@ -1154,7 +1154,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			By("find uploader pod")
 			var sourcePod *v1.Pod
 			Eventually(func() bool {
-				sourcePod, err = utils.FindPodByPrefix(f.K8sClient, dataVolume.Namespace, "cdi-upload", common.CDILabelSelector)
+				sourcePod, err = utils.FindPodByPrefix(f.K8sClient, dataVolume.Namespace, common.UploadPodName, common.CDILabelSelector)
```
brybacki, line 1157: :+1:

pkg/util/util.go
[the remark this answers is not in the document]
akalenyu: The thing is, we don't have the HTTP endpoint in the operator SDK, it is controller-runtime that gives us some free Prometheus server stuff, so we will have to create the server there

pkg/operator/controller/controller_test.go
akalenyu: Yes makes sense! thx

tests/import_test.go
akalenyu: Done, thx

pkg/operator/controller/controller_test.go
akalenyu: Done

pkg/util/util.go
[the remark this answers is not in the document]
akalenyu: @awels removed the cdiavailable & progressing funcs, as there is something much simpler available:
https://github.com/kubevirt/containerized-data-importer/pull/2009/files#diff-e492805ba442ae8a4a1872f4650b03454bad344f0aa1b220f9015d1253a5beb3R226-R231

tests/utils.go
awels: Yeah that is fine.

awels (commented):
/lgtm
/approve

PR #2019, opened 2021-11-14
Remove skopeo binary dependency - use containers/image instead

maya-r:
Depending on a Go binary from the package manager is undesirable
as whenever Go notifies us of a vulnerability and we need to
rebuild, we depend on the package to be updated, which may take
a while (especially in the case of, say, centos stream).

With a library, updates are a matter of rebuilding CDI itself,
and that's it.

**Release note**:

```release-note
NONE
```


arnongilboa:
Looks great! please remove `cmdRun()` as well.

PR #2020, opened 2021-11-15
Skip VDI test on OpenShift

maya-r:
For OpenShift we effectively build with centos stream, which builds
qemu-img without VDI support.

Our test previously succeeded because it only checked that the DV
was successful, not that the content is valid.

Skip for now.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -1066,7 +1066,10 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 		)
 	})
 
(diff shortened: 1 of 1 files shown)


brybacki:
this is a temporary solution until newer build with quemu-img is available?

maya-r:
it apparently needs qemu-img from centos 9 too, so a little less temporary

brybacki:
It does not not work on OpenShift, so there is no way to test it

brybacki:
lets see if it works, or needs a rebase
/retest

PR #2021, opened 2021-11-15
Move fsync to be after chmod on disk.img/1.22 flakes

akalenyu:

- We see this test failing quite a lot since we switched to 1.22:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2009/pull-containerized-data-importer-e2e-k8s-1.21-ceph/1458503318815903744
Attempting to fsync after setting the file permissions
- Failing to run command on container saying its not there:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2018/pull-containerized-data-importer-e2e-k8s-1.22-nfs/1459905808480342016
Looping over now which seems to solve it (Ready condition on verifier pod not really an indication to being 100% ready?)

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -84,17 +84,35 @@ func main() {
 	var preallocationApplied bool
 	var ds importer.DataSourceInterface
 
-	//Registry import currently support kubevirt content type only
-	if contentType != string(cdiv1.DataVolumeKubeVirt) && (source == controller.SourceRegistry || source == controller.SourceImageio) {
-		klog.Errorf("Unsupported content type %s when importing from %s", contentType, source)
-		os.Exit(1)
-	}
-
 	volumeMode := v1.PersistentVolumeBlock
 	if _, err := os.Stat(common.WriteBlockPath); os.IsNotExist(err) {
 		volumeMode = v1.PersistentVolumeFilesystem
 	} else {
 		preallocation = true
+	}
+
+	// With writeback cache mode it's possible that the process will exit before all writes have been commited to storage.
+	// To guarantee that our write was commited to storage, we make a fsync syscall and ensure success.
+	// Also might be a good idea to sync any chmod's we might have done.
+	defer func() {
+		dataFile := getImporterDestPath(contentType, volumeMode)
+		file, err := os.Open(dataFile)
+		if err != nil {
+			klog.Errorf("could not get file descriptor for fsync call: %+v", err)
+			os.Exit(1)
+		}
+		if err := file.Sync(); err != nil {
+			klog.Errorf("could not fsync following qemu-img writing: %+v", err)
+			os.Exit(1)
+		}
+		klog.V(3).Infof("Successfully completed fsync(%s) syscall, commited to disk\n", dataFile)
+		file.Close()
+	}()
+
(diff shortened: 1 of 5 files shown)


pkg/image/qemu.go
```diff
@@ -297,9 +297,16 @@ func (o *qemuOperations) CreateBlankImage(dest string, size resource.Quantity, p
 		return errors.Wrap(err, fmt.Sprintf("could not create raw image with size %s in %s", size.String(), dest))
 	}
 	// Change permissions to 0660
-	err = os.Chmod(dest, 0660)
+	file, err := os.Open(dest)
+	if err != nil {
+		return errors.Wrap(err, "could not get file descriptor for chmod call")
+	}
+	if err := file.Sync(); err != nil {
```
brybacki, line 304: so fsync first and then chmod? or the other way around

akalenyu: definitely the other way, thx

cmd/cdi-importer/importer.go
```diff
@@ -135,6 +135,21 @@ func main() {
 		touchDoneFile()
 		preallocationApplied = processor.PreallocationApplied()
 	}
+	// With writeback cache mode it's possible that the process will exit before all writes have been commited to storage.
+	// To guarantee that our write was commited to storage, we make a fsync syscall and ensure success.
+	// Also might be a good idea to sync any chmod's we might have done.
+	dataFile := getImporterDestPath(contentType, volumeMode)
```
brybacki, line 141: I think It only applies for the branch of code which actually creates the file, and before the touchDoneFile() at line 134.

akalenyu: there is a chmod also for the other branch of code:
https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/image/qemu.go#L300

brybacki: you are right!

cmd/cdi-importer/importer.go
```diff
@@ -135,6 +135,21 @@ func main() {
 		touchDoneFile()
 		preallocationApplied = processor.PreallocationApplied()
 	}
+	// With writeback cache mode it's possible that the process will exit before all writes have been commited to storage.
```
awels, line 138: Maybe make this a defer so we are ensured it is called. And put it above `if contentType != string(cdiv1.DataVolumeKubeVirt) && (source == controller.SourceRegistry || source == controller.SourceImageio) `

akalenyu: no problem, just question - why arent we ensured right now?

awels: For instance if there is an error in writing the termination message, we call the exit function, and the defer will get executed before we exit. Right now we won't execute.

awels: Actually just double checked, exit doesn't honor defer either.

akalenyu: Moved to defer above, thx

akalenyu:
/hold
lets get some cycles running so we know we got rid of those flakes?

PR #2027, opened 2021-11-17
Add alert for incomplete storage profiles / delete profile when corresponding SC gone

akalenyu:

It's possible that we are unable to fill in the details of a StorageProfile, which in turn makes it possible for users to fail to create their PVCs - lets alert about that.

Todo:
- [x] Functional test
- [x] Run func tests on k8s as well (stop skipping on not OpenShift)

Fixes #

**Release note**:

```release-note
StorageProfile incomplete alert, delete profile when corresponding SC gone 
```


pkg/controller/storageprofile-controller.go
```diff
@@ -25,6 +26,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+var (
+	// IncompleteProfileGauge is the metric we use to alert about incomplete storage profiles
+	IncompleteProfileGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kubevirt_cdi_incomplete_storageprofiles_total",
```
brybacki, line 33: now as we start seeing the metrics in a few places, we might need a naming util, or at least we can define a kubevirt_cdi as a constant

pkg/controller/storageprofile-controller.go
```diff
@@ -44,8 +54,24 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
...
 		return reconcile.Result{}, err
 	}
+	if _, err := r.reconcileStorageProfile(storageClass); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Alert for incomplete profiles
+	numIncomplete := 0
+	storageProfileList := &cdiv1.StorageProfileList{}
+	if err := r.client.List(context.TODO(), storageProfileList); err != nil {
```
brybacki, line 64: looks like a good way for a generic metric, I suppose `List` is already cached, is it?

akalenyu: Yes, this was my main concern, I am definitely using the cached client (as opposed to unCachedClient), so I think this is not too expensive

akalenyu:
@brybacki Does that `isIncomplete()` func make sense?

brybacki:
yes, the `isIncomplete` looks good.

akalenyu:
> yes, the `isIncomplete` looks good.

Now rethinking about it and wondering what happens if ClaimPropertySets isn't there:
```bash
spec: {}
status:
  provisioner: kubernetes.io/no-provisioner
  storageClass: nfs
```
Maybe should check for that as well in `isIncomplete()`?

brybacki:
> > yes, the `isIncomplete` looks good.
> 
> Now rethinking about it and wondering what happens if ClaimPropertySets isn't there:
> 
> ```shell
> spec: {}
> status:
>   provisioner: kubernetes.io/no-provisioner
>   storageClass: nfs
> ```
> 
> Maybe should check for that as well in `isIncomplete()`?


If ClaimPropertySets is not there, then liength of array `len(ClaimPropertySets)` will be zero. This is how it works in GO If I am not mistaken. 

This is good candidate for a unit test.

akalenyu:
> > > yes, the `isIncomplete` looks good.
> > 
> > 
> > Now rethinking about it and wondering what happens if ClaimPropertySets isn't there:
> > ```shell
> > spec: {}
> > status:
> >   provisioner: kubernetes.io/no-provisioner
> >   storageClass: nfs
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe should check for that as well in `isIncomplete()`?
> 
> If ClaimPropertySets is not there, then the length of array `len(ClaimPropertySets)` will be zero. This is how it works in go If I am not mistaken.
> 
> This is good candidate for a unit test.

Added

pkg/controller/storageprofile-controller.go
```diff
@@ -44,8 +54,24 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
 	if err := r.client.Get(context.TODO(), req.NamespacedName, storageClass); err != nil {
 		return reconcile.Result{}, err
 	}
+	if _, err := r.reconcileStorageProfile(storageClass); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Alert for incomplete profiles
```
ShellyKa13, line 74: I think this code should be in a different function and not in the reconcile function like: updateIncompleteProfiles

pkg/controller/storageprofile-controller.go
```diff
@@ -44,8 +54,24 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
...
+	numIncomplete := 0
+	storageProfileList := &cdiv1.StorageProfileList{}
+	if err := r.client.List(context.TODO(), storageProfileList); err != nil {
+		return reconcile.Result{}, err
+	}
+	for _, profile := range storageProfileList.Items {
```
ShellyKa13, line 80: It seems a bit inefficient for each storage profile update to go over all the list and check incomplete. Can't we change so this alert can be accumulative and then only need to increase incomplete in case the storage profile that was created/updated was incomplete and decrease from it in case there was an update and now it is? WDYT?

tests/operator_test.go
```diff
@@ -728,31 +793,38 @@ var _ = Describe("ALL Operator tests", func() {
...
-							if s == "1" {
-								return true
-							}
-						}
-					}
-					return false
-				}, 1*time.Minute, 1*time.Second).Should(BeTrue())
+				Eventually(func() int {
+					return getMetricValue("kubevirt_cdi_cr_ready")
+				}, 1*time.Minute, 1*time.Second).Should(BeNumerically("==", 1))
```
ShellyKa13, line 798: consider add a 'waitForMetricValue' function since it seems thats what you mostly do with getMetricValue

aglitke (commented):
Regarding deleting storage profiles when the corresponding sc is deleted: lgtm

tests/operator_test.go
```diff
@@ -728,31 +793,38 @@ var _ = Describe("ALL Operator tests", func() {
...
+			})
+
+			It("StorageProfile incomplete metric expected value when creating an incomplete profile", func() {
```
ShellyKa13, line 801: consider adding a test with more then one storageProfile incomplete since you check for a list. Also if you take my suggestion of changing the metric from summing a list to make it accumulative after each update it should probably be a needed test also.

pkg/controller/storageprofile-controller_test.go
```diff
@@ -45,11 +45,11 @@ var (
 
 var _ = Describe("Storage profile controller reconcile loop", func() {
 
-	It("Should return error if storage profile can not be found", func() {
```
ShellyKa13, line 47: not sure if possible or if its not easy if its worth working on it but maybe should check if can add a test which returns error which is different then not found and see that it does reque.

ShellyKa13 (commented):
Over all looks good, some suggestion and comments

pkg/controller/storageprofile-controller.go
akalenyu: Regarding the list, if we want all StorageProfiles they should already be cached
because this controller is watching all of them, so the List call should be cheap

The challenge, if I go accumulative, can be
- Profile 1 is broken
- Metric value increased
- Profile 2 gets created and is completely correct
- Metric decreased, we stop alerting (But there's still a faulty profile)

ShellyKa13: thats why I said only on update, on  create you dont need to decrease and in update you have the previous to compate to

ShellyKa13: or mark a profile as incomplete either on the status or in an internal map and then check for it before decreasing

akalenyu: An internal map sounds good! then I can conditionally increment and decrement
(The same problem in the acc. scenario can occur with updates as well, consider two updates to the same profile)

akalenyu: Thinking about it more - that would mean that we're keeping a map of all storage classes potentially,
which might not be as cheap as the list call

ShellyKa13: but the list call is not the problem the problem is you go over all the list and check for each profile, which is surely less cheap then a map :)

awels: Just remember don't want to keep state in memory in the controller, since the controller could disappear at any time, and have to be recreated. Also keep in mind reconcile loops should be idem potent, so if I run the same thing with the same inputs I should get the same results.

ShellyKa13: right a map is not suitable. its just saving that info in the storage profile itself. so if that seems an over kill for a metric than I guess this is ok

pkg/controller/storageprofile-controller.go
brybacki: or checkIncompleteProfiles

tests/operator_test.go
akalenyu: this sounds like a good idea, I'll have to create a simple storage class here since now just creating a profile that doesn't correspond to any SC wont do anything

akalenyu: Done

pkg/controller/storageprofile-controller_test.go
akalenyu: Yeah that is a challenge, not sure how I could simulate this scenario

pkg/controller/storageprofile-controller.go
akalenyu: Done

pkg/controller/storageprofile-controller.go
```diff
@@ -71,18 +75,9 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
 		return reconcile.Result{}, err
 	}
 
-	// Alert for incomplete profiles
-	numIncomplete := 0
-	storageProfileList := &cdiv1.StorageProfileList{}
-	if err := r.client.List(context.TODO(), storageProfileList); err != nil {
+	if err := r.checkIncompleteProfiles(); err != nil {
```
ShellyKa13, line 78: maybe you can shorten to one line:
return reconcile.Result{}, r.checkIncompleteProfiles()

tests/operator_test.go
```diff
@@ -800,31 +816,42 @@ var _ = Describe("ALL Operator tests", func() {
...
+				for i := 0; i < numAddedStorageClasses; i++ {
+					profile := &cdiv1.StorageProfile{}
+					err = f.CrClient.Get(context.TODO(), types.NamespacedName{Name: fmt.Sprintf("unknown-sc-%d", i)}, profile)
+					Expect(err).ToNot(HaveOccurred())
+					// These might be wrong values, but at least the profile is no longer incomplete
+					profile.Spec.ClaimPropertySets = defaultStorageClassProfile.Status.ClaimPropertySets
+					err = f.CrClient.Update(context.TODO(), profile)
+					Expect(err).ToNot(HaveOccurred())
+				}
 				Eventually(func() int {
```
ShellyKa13, line 847: consider add the check in the loop and check after each fix that the metric decrease by 1

akalenyu: Done

pkg/controller/storageprofile-controller.go
akalenyu: Done

ShellyKa13:
/lgtm
Great job!

pkg/operator/resources/namespaced/controller.go
```diff
@@ -128,6 +129,16 @@ func createControllerDeployment(controllerImage, importerImage, clonerImage, upl
...
+	container.Ports = []corev1.ContainerPort{
+		{
+			Name:          "metrics",
+			ContainerPort: 8080,
+			Protocol:      "TCP",
+		},
+	}
+	labels := util.MergeLabels(deployment.Spec.Template.GetLabels(), map[string]string{common.PrometheusLabelKey: common.PrometheusLabelValue})
+	deployment.SetLabels(labels)
+	deployment.Spec.Template.SetLabels(labels)
```
awels, line 141: Isn't this going to break the deployment if we upgrade? Since the template is immutable.

tests/operator_test.go
```diff
@@ -724,35 +805,55 @@ var _ = Describe("ALL Operator tests", func() {
...
+					// These might be wrong values, but at least the profile is no longer incomplete
+					profile.Spec.ClaimPropertySets = defaultStorageClassProfile.Status.ClaimPropertySets
+					err = f.CrClient.Update(context.TODO(), profile)
+					Expect(err).ToNot(HaveOccurred())
+					expectedIncomplete--
+					Eventually(func() int {
+						return getMetricValue("kubevirt_cdi_incomplete_storageprofiles_total")
+					}, 2*time.Minute, 1*time.Second).Should(BeNumerically("==", expectedIncomplete))
+				}
+				By("Delete unknown storage classes")
```
awels, line 852: You will want to put this in an AfterEach, because if there is test failure between creating and deleting, the deletion section will not get executed. And we want to make sure we clean up. AfterEach is executed even if the test failed.

tests/utils.go
```diff
@@ -245,27 +262,40 @@ func findConditionByType(conditionType cdiv1.DataVolumeConditionType, conditions
...
+			}
+		}
+		secret, err := f.K8sClient.CoreV1().Secrets(monitoringNs).Get(context.TODO(), secretName, metav1.GetOptions{})
 		if err != nil {
 			return false
 		}
+		if _, ok := secret.Data["token"]; !ok {
+			return false
+		}
+		token = string(secret.Data["token"])
```
awels, line 298: Am I missing where token is used?

pkg/controller/storageprofile-controller.go
```diff
@@ -42,10 +52,27 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
 
 	storageClass := &storagev1.StorageClass{}
 	if err := r.client.Get(context.TODO(), req.NamespacedName, storageClass); err != nil {
+		if k8serrors.IsNotFound(err) || storageClass.GetDeletionTimestamp() != nil {
```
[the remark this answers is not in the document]
akalenyu: Yes!

pkg/operator/resources/namespaced/controller.go
akalenyu: `spec.selector` is immutable, spec.template.metadata.labels/deployment.metadata.labels are fine
(otherwise k8s basically says you can't add labels to this resource at all)

tests/utils.go
akalenyu: `req.Header.Add("Authorization", "Bearer "+token)`

pkg/controller/storageprofile-controller.go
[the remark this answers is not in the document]
akalenyu: Done

tests/operator_test.go
akalenyu: Done

tests/operator_test.go
```diff
@@ -671,11 +676,73 @@ var _ = Describe("ALL Operator tests", func() {
...
 
 		var _ = Describe("Alert tests", func() {
+			var numAddedStorageClasses int
 			f := framework.NewFramework("alert-tests")
 
+			AfterEach(func() {
+				By("Delete unknown storage classes")
+				for i := 0; i < numAddedStorageClasses; i++ {
+					name := fmt.Sprintf("unknown-sc-%d", i)
+					_, err := f.K8sClient.StorageV1().StorageClasses().Get(context.TODO(), name, metav1.GetOptions{})
```
awels, line 686: This isn't necessarily wrong, but there really is no reason to get the storage class. Simply delete it, and ignore not found errors. There is no guarantee that in between this get, and the delete, the storage class didn't get deleted by something else.

tests/utils.go
awels: Gotcha, its below in the part that was hidden in the review view.

awels (approved):
/lgtm
let me know if you want to address that one comment or not.

tests/operator_test.go
akalenyu: Yeah I see what you mean, if it's fine with you, we can merge as is

awels: Yeah, its fine with me, we have that pattern all over the place, just noting that it is not really needed.

PR #2030, opened 2021-11-21
[release-v1.38] Use a path under /dev for attaching a PVC

maya-r:
After kubernetes 1.20 it seems like / is mounted nodev, so we can't
have "devicePath: /pvc" any more.

Seems to happen on some updates of OpenShift too.

**Release note**:

```release-note
NONE
```


maya-r:
Doesn't seem to be the case.

PR #2036, opened 2021-11-24
Have the csv generator generate the dataimportcron CRD as well

maya-r:
```release-note
Have the cdi-operator csv generator generate the dataimportcron CRD as well
```

### DIFF
--- a/tools/csv-generator/BUILD.bazel
+++ b/tools/csv-generator/BUILD.bazel
@@ -6,6 +6,7 @@ go_library(
     importpath = "kubevirt.io/containerized-data-importer/tools/csv-generator",
     visibility = ["//visibility:private"],
     deps = [
+        "//pkg/operator/resources/cluster:go_default_library",
         "//pkg/operator/resources/operator:go_default_library",
         "//tools/util:go_default_library",
     ],
--- a/tools/csv-generator/csv-generator.go
+++ b/tools/csv-generator/csv-generator.go
@@ -16,6 +16,7 @@ import (
 	"flag"
 	"os"
 
+	cdicluster "kubevirt.io/containerized-data-importer/pkg/operator/resources/cluster"
 	cdioperator "kubevirt.io/containerized-data-importer/pkg/operator/resources/operator"
 	"kubevirt.io/containerized-data-importer/tools/util"
 )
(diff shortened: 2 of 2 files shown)


arnongilboa:
@maya-r please `gofmt -w tools/csv-generator/csv-generator.go` to get  linter happier

maya-r:
@akalenyu was wondering why this isn't necessary for the datasource crd, and I'm not sure why that is the case.
@nunnatsa is this because SSP is using this CRD and so we can't wait for the cdi-operator to create the CRD?

mhenriks:
DataSource should probably be included as well

mhenriks:
Actually, why is this needed?

mhenriks:
The cdi operator is creating/managing this CRD (in addition to others)

nunnatsa:
> The cdi operator is creating/managing this CRD (in addition to others)

@mhenriks - when installing using OLM, the SSP is trying to read DataImportCrons and failed because of the missing CRD. It then get into crash loop, so OLM can never complete the installation of the csv. Only after creating tje HyperConvered CR, HCO creates the CDI CR and only then, the CRD is created by CDI.

mhenriks:
>. @mhenriks - when installing using OLM, the SSP is trying to read DataImportCrons and failed because of the missing CRD. It then get into crash loop, so OLM can never complete the installation of the csv. Only after creating tje HyperConvered CR, HCO creates the CDI CR and only then, the CRD is created by CDI.

SSP should not be that tightly coupled to CDI, handle the DataImportCron not existing, and discover when they do exist.  CDI does this for multiple resources.  VolumeSnapshots, for example.

awels:
I agree with @mhenriks the CDI operator is the one creating and managing the CRDs, it does this perfectly fine with many other CRDs, including datavolumes. I don't see why DIC is any different.

maya-r:
Closing as I agree with the responses - SSP should be more tolerant of this scenario.

nunnatsa:
@maya-r @mhenriks @awels @arnongilboa - please reconsider this.

Regardless the behavior SSP, what is the benefit to dynamically add/remove the CRD? It causes issues in the installation, and then again during uninstall. It would be much simpler if the CDR will be a static file in the installation.

PR #2037, opened 2021-11-24
Attempt to mitigate flake #1989, 'io: read/write on closed pipe'

### LINKED ISSUE #1989: [flaky] Some tests fail with "io: read/write on closed pipe",

>
> /kind bug
> /kind enhancement

**What happened**:

While the test log looks good (at a first glance) the whole test fails with:
```
/root/go/src/kubevirt.io/containerized-data-importer/tests/upload_test.go:169
Unexpected error:
    <*errors.errorString | 0xc000198140>: {
        s: "io: read/write on closed pipe",
    }
    io: read/write on closed pipe
occurred
/root/go/src/kubevirt.io/containerized-data-importer/tests/upload_test.go:339
```
Sample run: 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1986/pull-containerized-data-importer-e2e-k8s-1.19-ceph/1447744882138419200

**Environment**:
- Install tools:
- Others:

akalenyu:

So the upload func being called in the eventually loop means we're going to be running
this:
https://github.com/kubevirt/containerized-data-importer/blob/main/tests/upload_test.go#L415-L428
a bunch of times, which leads to sometimes failing with
```bash
<*errors.errorString | 0xc000200140>: {
        s: "io: read/write on closed pipe",
    }
    io: read/write on closed pipe
occurred
/root/go/src/kubevirt.io/containerized-data-importer/tests/upload_test.go:424
```
Slightly incrementing the interval in which we call the upload func seems to mitigate this flake.

Fixes #1989

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -153,9 +153,14 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		}
 
 		By("Do upload")
-		Eventually(func() error {
-			return uploader(uploadProxyURL, token, expectedStatus)
-		}, timeout, pollingInterval).Should(BeNil(), "Upload should eventually succeed, even if initially pod is not ready")
+		Eventually(func() bool {
+			err = uploader(uploadProxyURL, token, expectedStatus)
+			if err != nil {
+				fmt.Fprintf(GinkgoWriter, "ERROR: %s\n", err.Error())
+				return false
+			}
+			return true
+		}, timeout, 5*time.Second).Should(BeTrue(), "Upload should eventually succeed, even if initially pod is not ready")
 
 		if validToken {
 			By("Verify PVC status annotation says succeeded")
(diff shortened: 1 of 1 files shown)


akalenyu:
/hold
caught 1 failure with same err, guess its not solved

akalenyu:
> /hold caught 1 failure with same err, guess its not mitigated

/hold cancel
there was another test could fail due to this which I missed initially

brybacki:
`    io: read/write on closed pipe` is happening probably because the timeout kicks in and connection is closed or something like that. There must be some other problem earlier in the Eventually loop. I suppose loging the error from inside the loop could help us find the reason.

akalenyu:
> ` io: read/write on closed pipe` is happening probably because the timeout kicks in and connection is closed or something like that. There must be some other problem earlier in the Eventually loop. I suppose loging the error from inside the loop could help us find the reason.

not sure I got it - if there was an err during the `Eventually` loop, we'd see it since it `.Should(BeNil())`

brybacki:
> > ` io: read/write on closed pipe` is happening probably because the timeout kicks in and connection is closed or something like that. There must be some other problem earlier in the Eventually loop. I suppose loging the error from inside the loop could help us find the reason.
> 
> not sure I got it - if there was an err during the `Eventually` loop, we'd see it since it `.Should(BeNil())`

From the documentation of `Eventually` -  The assertion is tried periodically until it passes or a timeout occurs.

So it can fail many times, as long as the confgured timeout allows.

akalenyu:
> > > ` io: read/write on closed pipe` is happening probably because the timeout kicks in and connection is closed or something like that. There must be some other problem earlier in the Eventually loop. I suppose loging the error from inside the loop could help us find the reason.
> > 
> > 
> > not sure I got it - if there was an err during the `Eventually` loop, we'd see it since it `.Should(BeNil())`
> 
> From the documentation of `Eventually` - The assertion is tried periodically until it passes or a timeout occurs.
> 
> So it can fail many times, as long as the confgured timeout allows.

Yep, now I get what you mean, we're definitely missing the err itself in the logs.
added

brybacki:
So maybe the upload was sometimes too slow?
Anyway this looks good. With that new logger we can at least have some more info available in case of failure.

/lgtm

awels:
If for nothing else but the extra debug information.

akalenyu:
> If for nothing else but the extra debug information.

Keep in mind I also bumped the interval which from local testing has been proven to mitigate the flake

PR #2039, opened 2021-11-28
Add common labels into alert definitions

assafad:
We want to be able to list all kubevirt alerts so we added labels to
differentiate them.

**Release note**:

```release-note
Added common labels into alert definitions
```

### DIFF
--- a/pkg/operator/controller/prometheus.go
+++ b/pkg/operator/controller/prometheus.go
@@ -41,11 +41,16 @@ import (
 )
 
 const (
-	ruleName            = "prometheus-cdi-rules"
-	rbacName            = "cdi-monitoring"
-	monitorName         = "service-monitor-cdi"
-	defaultMonitoringNs = "monitoring"
-	runbookURLBasePath  = "https://kubevirt.io/monitoring/runbooks/"
+	ruleName                 = "prometheus-cdi-rules"
+	rbacName                 = "cdi-monitoring"
+	monitorName              = "service-monitor-cdi"
+	defaultMonitoringNs      = "monitoring"
+	runbookURLBasePath       = "https://kubevirt.io/monitoring/runbooks/"
+	severityAlertLabelKey    = "severity"
+	partOfAlertLabelKey      = "kubernetes_operator_part_of"
+	partOfAlertLabelValue    = "kubevirt"
+	componentAlertLabelKey   = "kubernetes_operator_component"
+	componentAlertLabelValue = common.CDILabelValue
 )
 
 func ensurePrometheusResourcesExist(c client.Client, scheme *runtime.Scheme, owner metav1.Object) error {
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/prometheus.go
```diff
@@ -41,11 +41,13 @@ import (
...
-	rbacName            = "cdi-monitoring"
-	monitorName         = "service-monitor-cdi"
-	defaultMonitoringNs = "monitoring"
-	runbookURLBasePath  = "https://kubevirt.io/monitoring/runbooks/"
+	ruleName                      = "prometheus-cdi-rules"
+	rbacName                      = "cdi-monitoring"
+	monitorName                   = "service-monitor-cdi"
+	defaultMonitoringNs           = "monitoring"
+	runbookURLBasePath            = "https://kubevirt.io/monitoring/runbooks/"
+	kubernetes_operator_part_of   = "kubevirt"
```
sradco, line 49: @assaf-admi Please use Camel case for the constants name

sradco:
@assaf-admi Please check that you didn't miss adding the labels to the tests.

pkg/operator/controller/prometheus.go
```diff
@@ -41,11 +41,13 @@ import (
...
+	kubernetes_operator_component = "containerized-data-importer"
```
akalenyu, line 50: Can use `common.CDILabelValue`

akalenyu:
Btw, we also fetch the installer labels here:
https://github.com/kubevirt/containerized-data-importer/pull/2039/files#diff-d514622f961bb0c251a640e7c31a7f197e167520987748d3205d0a81052be8c7R63

Maybe we can consider naming the constants a little different to distinct between the two?
something like `CDIOperatorComponentAlertLabel`

assaf-admi:
> Btw, we also fetch the installer labels here: https://github.com/kubevirt/containerized-data-importer/pull/2039/files#diff-d514622f961bb0c251a640e7c31a7f197e167520987748d3205d0a81052be8c7R63
> 
> Maybe we can consider naming the constants a little different to distinguish between the two? something like `CDIOperatorComponentAlertLabel`

pkg/operator/controller/prometheus.go
```diff
@@ -46,6 +46,8 @@ const (
 	monitorName         = "service-monitor-cdi"
 	defaultMonitoringNs = "monitoring"
 	runbookURLBasePath  = "https://kubevirt.io/monitoring/runbooks/"
+	partOfLabelValue    = "kubevirt"
```
akalenyu, line 49: I was thinking if we could have `Alert\Metric` as part of the variable name to indicate these are not the standard relationship label values

sradco: @assafad Please update the names 
partOfAlertLabelValue and componentAlertLabelValue

PR #2040, opened 2021-11-29
Add DataImportCron initial import before cron schedule

arnongilboa:

Currently, the DataImportCron polls and imports are handled by the controller according to the cron schedule. We would like to add an initial import on DataImportCron creation.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -111,7 +111,14 @@ func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron 
 			r.log.Error(err, "Unable to create CronJob")
 			return err
 		}
+		if err := r.client.Create(ctx, newInitialJob(cronJob)); err != nil {
+			r.log.Error(err, "Unable to create Job")
+			return err
+		}
+	} else if isImageStreamSource(dataImportCron) && dataImportCron.Annotations[AnnNextCronTime] == "" {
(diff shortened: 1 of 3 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -111,7 +111,14 @@ func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron
 			r.log.Error(err, "Unable to create CronJob")
 			return err
 		}
+		if err := r.client.Create(ctx, newInitialJob(cronJob)); err != nil {
```
mhenriks, line 114: does this resource ever get deleted?

arnongilboa: see the ttl = 0 added

mhenriks: Oh so are jobs currently not getting deleted?

arnongilboa: In `CronJobSpec` we set both `SuccessfulJobsHistoryLimit` and `FailedJobsHistoryLimit` to 0, so Jobs are deleted at the CronJob level.
But now it's an independent job, so we set `JobSpec.TTLSecondsAfterFinished` to 0. Removing the 2 previous ones since the TTL seems to covers them both doesn't seem important to me.

PR #2045, opened 2021-12-02
Add DataImportCron status conditions

arnongilboa:

The `DataImportCron` controller updates the status conditions in a controlled `DataImportCron` and its managed `DataSource`.

DataImportCron:
- UpToDate - indicates if the the most recent import is successful and `DataSource` is up-to-date. Updated to False whenever the source digest (latest sha256) is updated.
- Progressing - indicates whether the cron is currently in the process of importing. Updated to True if there is a current import and its `DataVolume` is `ImportInProgress`, otherwise False.

DataSource:
- Ready - indicates that the corresponding pvc exists and is populated. Update according to `DataImportCron.Status.LastImportedPVC` `DataVolume`'s `DataVolumeReady` condition, if the `DataVolume` exists. Otherwise False. Unlike `DataImportCron` `UpToDate` condition, this one does not care about newer source digest.

**Release note**:
```release-note
NONE
```


pkg/controller/dataimportcron-controller.go
```diff
@@ -218,20 +219,27 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		now := metav1.Now()
 		dataImportCron.Status.LastExecutionTimestamp = &now
 
-		if dataVolume.Status.Phase == cdiv1.Succeeded {
+		switch dataVolume.Status.Phase {
```
mhenriks, line 222: There is an ImportFailed phase.  Not sure if/how it is currently used, may want to check to see if it should be handled here

pkg/controller/dataimportcron-controller.go
```diff
@@ -218,20 +219,27 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
 				return res, err
 			}
+		case cdiv1.ImportScheduled:
+			updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionFalse, "Import is scheduled", scheduled)
+		case cdiv1.ImportInProgress:
+			updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionTrue, "Import is progressing", inProgress)
 		}
 	}
 
+	if err := r.updateDataSource(ctx, dataImportCron); err != nil {
```
mhenriks, line 239: Looks like some new datasource management was added to this PR in addition to the conditions?

pkg/controller/dataimportcron-controller.go
```diff
@@ -460,6 +484,16 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 	}); err != nil {
 		return err
 	}
+	// Watch only for DataSource deletion
+	if err := c.Watch(&source.Kind{Type: &cdiv1.DataSource{}},
```
mhenriks, line 488: It is possible for delete events to get lost.  If it is critical to do something when a resource is deleted, use a finalizer

pkg/controller/dataimportcron-controller.go
```diff
@@ -195,11 +198,9 @@ func getCronRegistrySource(cron *cdiv1.DataImportCron) (*cdiv1.DataVolumeSourceR
...
 }
 
-// FIXME: update all conditions
 func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
 	log := r.log.WithName("update")
 	res := reconcile.Result{}
-	//FIXME: add ImportStatus DataVolumePhase so we'll handle only on transition (to Succeeded etc.)
 	importSucceeded := false
 	imports := dataImportCron.Status.CurrentImports
 	if imports != nil {
```
mhenriks, line 206: I think conditions should be added/updated all the time, not just when `imports != nil`

mhenriks (commented):
Would be helpful if pr description is updated to include conditions/values/transitions.

pkg/controller/dataimportcron-controller.go
arnongilboa: The `DataImportCronUpToDate` condition is already updated all the time. When `imports == nil` the missing condition update is `DataImportCronProgressing` = false. Adding it. Thanks.

pkg/controller/dataimportcron-controller.go
arnongilboa: Not much, just re-creation of DataSource if somehow deleted, and refactoring of updateDataSource so DataSource will be k8s-updated when its conditions are updated.

pkg/controller/dataimportcron-controller.go
arnongilboa: As discussed, we won't use a finalizer for DataSource, but (re-)create it in whenever we see it's missing.

pkg/controller/dataimportcron-controller.go
arnongilboa: Can be added, but I have no idea how to cause the importer pod to fail. When pod is CrashLoopBackOff the DV phase will still be ImportInProgress.

tests/dataimportcron_test.go
```diff
@@ -133,14 +133,20 @@ var _ = Describe("DataImportCron", func() {
...
 				Expect(err).ToNot(HaveOccurred())
-				return cron.Status.LastImportedPVC != nil && cron.Status.LastImportedPVC.Name == currentImportDv
+				progressCond := controller.FindDataImportCronConditionByType(cron, cdiv1.DataImportCronProgressing)
+				upToDateCond := controller.FindDataImportCronConditionByType(cron, cdiv1.DataImportCronUpToDate)
+				return progressCond != nil && progressCond.Status == corev1.ConditionFalse &&
+					upToDateCond != nil && upToDateCond.Status == corev1.ConditionTrue &&
+					cron.Status.LastImportedPVC != nil && cron.Status.LastImportedPVC.Name == currentImportDv
 			}, dataImportCronTimeout, pollingInterval).Should(BeTrue())
 		}
 		if checkGarbageCollection {
```
mhenriks, line 152: looking light on tests?  should test at minimum that retention policies work as expected, datasource/datavolume recreated if deleted

pkg/controller/dataimportcron-controller.go
awels: So we are fine if the delete event gets lost, and it gets recreated on the next reconcile loop?

pkg/controller/dataimportcron-controller.go
```diff
@@ -242,10 +255,20 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	}
 
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
-	if desiredDigest != "" && (imports == nil || (importSucceeded && desiredDigest != imports[0].Digest)) {
-		if err := r.createImportDataVolume(ctx, dataImportCron); err != nil {
-			return res, err
+	digestUpdated := desiredDigest != "" && (len(imports) == 0 || desiredDigest != imports[0].Digest)
+	if digestUpdated {
```
awels, line 259: There are 4 permutations here, in the unit tests below I see only 2 tests. Can we add some unit tests that ensure that the corner cases are doing what they are supposed to do?

pkg/controller/dataimportcron-controller.go
```diff
@@ -101,12 +103,14 @@ func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron
 	if dataImportCron.Annotations == nil {
 		dataImportCron.Annotations = make(map[string]string)
 	}
+	if !HasFinalizer(dataImportCron, dataImportCronFinalizer) {
+		AddFinalizer(dataImportCron, dataImportCronFinalizer)
```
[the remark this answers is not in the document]
arnongilboa: Sure, removed the `HasFinalizer()` check, however we update the CR anyway with `LastExecutionTimestamp` in each reconcile, so I see no reason to check if the finalizes changed.

pkg/controller/dataimportcron-controller.go
arnongilboa: Removing the ImportFailed case, as it's a dead code.

pkg/controller/dataimportcron-controller.go
arnongilboa: Yes, we found that to be the reasonable behavior.

pkg/controller/dataimportcron-controller.go
```diff
@@ -302,22 +322,32 @@ func (r *DataImportCronReconciler) updateDataSourceOnSuccess(ctx context.Context
...
-			Name:       dataImportCron.Name,
-			UID:        dataImportCron.UID,
-		})
-	}
-	sourcePVC := &cdiv1.DataVolumeSourcePVC{
-		Namespace: dataImportCron.Namespace,
-		Name:      dataImportCron.Status.CurrentImports[0].DataVolumeName,
+	dataSourceCopy := dataSource.DeepCopy()
+
+	if dataSource.Labels == nil {
```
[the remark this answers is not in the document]
arnongilboa: @awels For all the CRs created here (CronJob, Job, DataSouce, DVs...)? Also add to new DataImportCron? I don't see you add it to created DVs in its controller.

tests/dataimportcron_test.go
arnongilboa: Tests added. Note that currently `RetentionPolicy` specifies whether the `DataVolumes` and `DataSources` are retained when their `DataImportCron` is deleted. However, when `DataVolume` or `DataSource` are deleted during their `DataImportCron` lifetime, they are recreated regardless of the `RetentionPolicy`.

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
akalenyu: Yes, we have https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/config-controller.go#L118
It would make sense to label DataSources, DataImportCrons, CronJob/Jobs, DVs (that are created for corn purpose)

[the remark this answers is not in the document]
arnongilboa: Added

awels (commented):
/lgtm

pkg/controller/dataimportcron-controller.go
mhenriks: The question is when will that next reconcile happen?

PR #2050, opened 2021-12-06
Improve directory detection on containerdisk imports

rmohr:

Check the tar header metadata if a tar entry is a directory instead of
checking if it has a `/` suffix.

This solves the following issue:

While most applications write a `/` at the end of a directory entry in a
tar header, it is not necessarily there. If it is missing, the directory
will be treated as the target file and CDI tries and fails to import the
`disk` directory itself:

```
iI1203 12:02:26.740362       1 transport.go:149] File 'disk' found in the layer
I1203 12:02:26.740461       1 util.go:172] Writing data...
E1203 12:02:26.742269       1 registry-datasource.go:123] Error reading directory
E1203 12:02:26.742319       1 data-processor.go:229] readdirent: not a directory
image file does not exist in image directory
```

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/transport.go
+++ b/pkg/importer/transport.go
@@ -110,8 +110,8 @@ func isWhiteout(path string) bool {
 	return strings.HasPrefix(filepath.Base(path), whFilePrefix)
 }
 
-func isDir(path string) bool {
-	return strings.HasSuffix(path, "/")
+func isDir(hdr *tar.Header) bool {
+	return hdr.Typeflag == tar.TypeDir
 }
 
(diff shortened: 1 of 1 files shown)


brybacki:
/test pull-containerized-data-importer-e2e-k8s-1.21-hpp

Lane had rare problem with port already used, but I suppose can be mitigated by fixing something in the kubevirtci. 
`The Service "cdi-uploadproxy-nodeport" is invalid: spec.ports[0].nodePort: Invalid value: 31001: provided port is already allocated`

PR #2059, opened 2021-12-13
Whitespace change to builder to trigger postsubmit job

maya-r:
we failed to push an updated builder for #1982, doing a no-op change will trigger the post-submit job.
-->
```release-note
NONE
```


maya-r:
/hold
after https://github.com/kubevirt/project-infra/pull/1796 merges we should merge this

maya-r:
[it checks if there were changes to the builder Dockerfile](https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/bazel-build-builder.sh#L22), and we make such changes really infrequently

awels:
Oh good point, okay
/lgtm
/approve

PR #2060, opened 2021-12-14
add PVC claimName to datavolume status

maya-r:
Currently to go from a datavolume to the underlying PVC you must have the internal knowledge that the PVC claim name is the same as the datavolume name.
Make it possible to discover the underlying PVC without this knowledge by keeping a field in the datavolume status.

Added upgrade testing - the only concern I had was "what happens during upgrade and v1alpha1 DVs", so I added support for testing on artifacts created on an older version.
This namespace is only created on the upgrade lane, and if it doesn't exist, the tests are skipped.

The old version artifacts could be installed regardless of updates, I opted not to do this as it slows down `make cluster-sync`.

**Release note**:

```release-note
Make it more obvious what the underlying PVC name is: add a datavolume field status.claimName which is the PVC name used by the underlying PVC.
```


cluster-sync/sync.sh
```diff
@@ -123,6 +123,25 @@ function wait_cdi_pods_updated {
...
 }
 
+# Setup some datavolumes in older version for testing upgrades
+# Done unconditionally to make it easier to write tests.
+function setup_for_upgrade_testing {
+  if _kubectl get namespace cdi-testing-old-version-artifacts ; then
+    echo "Old version testing environment already setup"
+    return
+  fi
+  echo "Missing old version environment setup, creating"
```
brybacki, line 133: why is that missing? that should be done in preparation of test infrastructure

cluster-sync/sync.sh
```diff
@@ -190,6 +209,7 @@ if [[ ! -z "$UPGRADE_FROM" ]]; then
     fi
     echo "Currently at version: $VERSION"
     wait_cdi_pods_updated
+    setup_for_upgrade_testing
```
brybacki, line 212: is this supposed to be run in a loop?

akalenyu:
Might be late to the party but the 1.21-ceph lane has RANDOM_CR=true,
which means no upgrade will take place:
https://github.com/kubevirt/containerized-data-importer/blob/main/automation/test.sh#L56-L57

akalenyu:
> Might be late to the party but the 1.21-ceph lane has RANDOM_CR=true, which means no upgrade will take place: https://github.com/kubevirt/containerized-data-importer/blob/main/automation/test.sh#L56-L57

Ah okay I see, we're always randomizing the CDI namespace name in CI so can't use `-n cdi`

kgoldbla (approved):
All Polarion ID's and the RFE ID are correct

cluster-sync/sync.sh
maya-r: nope - we want to run it once to setup an environment with DVs from one of the versions, and if we do multiple upgrades before running any tests then not set it up again

cluster-sync/sync.sh
maya-r: It is missing in the first time this function is called, I added the messages for debug/legibility purposes.

brybacki: I still do not get why it can be missing, it is all created in lines 149-152 in steps for setup: `test-infra`

cluster-sync/sync.sh
brybacki: so the function is run inside a loop (lines 176-213) but because of the condition inside it only run on the first time

so maybe it would be better to call it before the loop? or after the loop?

maya-r: I wanted to install DVs in the first version we upgrade from, which is the oldest, so after the first loop run.
before: no CDI version installed
after: fairly recent CDI version

brybacki: in that case, it is good, 
maybe a simple comment for anyone wanting to update/remove it in the future

cluster-sync/sync.sh
```diff
@@ -14,8 +14,8 @@ source ./cluster-sync/${KUBEVIRT_PROVIDER}/provider.sh
 CDI_INSTALL="install-operator"
 CDI_NAMESPACE=${CDI_NAMESPACE:-cdi}
 CDI_INSTALL_TIMEOUT=${CDI_INSTALL_TIMEOUT:-120}
-CDI_AVAILABLE_TIMEOUT=${CDI_AVAILABLE_TIMEOUT:-480}
-CDI_PODS_UPDATE_TIMEOUT=${CDI_PODS_UPDATE_TIMEOUT:-480}
+CDI_AVAILABLE_TIMEOUT=${CDI_AVAILABLE_TIMEOUT:-4800}
```
arnongilboa, line 17: can't it be a bit annoying to timeout only after 80 mins?

cluster-sync/sync.sh
arnongilboa: I agree regarding the first 3 applies but here we also apply the new `upgrade-testing-artifacts.yaml` and wait for Readiness.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -248,6 +248,8 @@ const (
 
 // DataVolumeStatus contains the current status of the DataVolume
 type DataVolumeStatus struct {
+	// ClaimName is the name of the underlying PVC used by the DataVolume.
+	ClaimName string `json:"claimName,omitempty"`
 	//Phase is the current phase of the data volume
```
arnongilboa, line 253: Why put it in status and not in spec? that way one can actually choose a non-default name for the created PVC.
Currently the field is static once set by `reconcileDataVolumeStatus()`, and that's a bit boring for status imho.

cluster-sync/sync.sh
maya-r: ah, I can probably lower it back - I was so confused as to why it was failing on the CI. It was a completely different issue, but I wanted to be sure it's not some too-low timeout.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
maya-r: Things will really start breaking if we allow that, right now a lot of things implicitly assume pvc name == dv name :-)

cluster-sync/sync.sh
maya-r: Yes, it's about upgrade-testing-artifacts.yaml and setting up a quick environment so that I can setup some easy DVs (I didn't want to rely on external URLs)

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
brybacki: and now, in places that assume pvc name == dv name, we can start using the status.claimName, and only in the future we might have the possibility to introduce the spec.pvcName. 

Anyway changing `spec.pvcName` might only be possible before the actual action(upload/import/clone etc...) starts. But that is a future ;)

cluster-sync/sync.sh
brybacki: so if I read the script correctly, previously we have only applied the 3 manifests (cdi-testing-sa, file-host, registry-host) after checking this condition 
```
if [ "${KUBEVIRT_PROVIDER}" != "external" ] && [ "${CDI_SYNC}" == "test-infra" ]; then
```
and now we are going to apply them also for `external` provider if we do `upgrade` tests. Is that ok?

maya-r: I can introduce the external provider limitations

awels:
/override pull-containerized-data-importer-e2e-k8s-1.22-hpp-istio

PR #2062, opened 2021-12-16
Bump timeout in DataImportCron test suite

akalenyu:

Locally see 90 seconds is not enough for ensuring PVCs were cleaned up in `[test_id:7406]`,
bumping the timeout to avoid flake.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/dataimportcron_test.go
+++ b/tests/dataimportcron_test.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	dataImportCronTimeout = 90 * time.Second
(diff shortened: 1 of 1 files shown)


akalenyu:
/test pull-containerized-data-importer-e2e-k8s-1.21-ceph

So #1989 is still here and the extra log tells us that its an intermittent 502 instead of 200 in one of the upload Eventually tries.
This by itself is ok to us, since well just try the HTTP call again, but some of i/o ops there are not playing nicely with 1 try failing

PR #2063, opened 2021-12-16
Add alert for DataImportCron not being up to date

akalenyu:

DataImportCrons now have conditions (particularly UpToDate) that tell us if
things are going as planned. We can utilize those to alert whenever were not UpToDate for a while.

Fixes #

**Release note**:

```release-note
Alert about a DataImportCron not being up to date; possibly not using a PVC that holds latest OS
```

### DIFF
--- a/pkg/controller/dataimportcron-conditions.go
+++ b/pkg/controller/dataimportcron-conditions.go
@@ -17,6 +17,7 @@ See the License for the specific language governing permissions and
 package controller
 
 import (
+	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
@@ -34,6 +35,13 @@ const (
 )
 
 func updateDataImportCronCondition(cron *cdiv1.DataImportCron, conditionType cdiv1.DataImportCronConditionType, status corev1.ConditionStatus, message, reason string) {
+	if conditionType == cdiv1.DataImportCronUpToDate {
+		if status != corev1.ConditionTrue {
+			DataImportCronOutdatedGauge.With(getPrometheusCronLabels(cron)).Set(1)
+		} else {
+			DataImportCronOutdatedGauge.With(getPrometheusCronLabels(cron)).Set(0)
+		}
+	}
 	if condition := FindDataImportCronConditionByType(cron, conditionType); condition != nil {
 		updateConditionState(&condition.ConditionState, status, message, reason)
 	} else {
@@ -88,14 +96,9 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 	}
 }
 
-func updateDataImportCronOutdatedMetric(prevUpToDateCondition *cdiv1.DataImportCronCondition, increment bool) {
-	if increment {
-		if prevUpToDateCondition == nil || prevUpToDateCondition.ConditionState.Status != corev1.ConditionFalse {
-			DataImportCronOutdatedGauge.Inc()
-		}
-	} else {
-		if prevUpToDateCondition.ConditionState.Status != corev1.ConditionTrue {
-			DataImportCronOutdatedGauge.Dec()
-		}
+func getPrometheusCronLabels(cron *cdiv1.DataImportCron) prometheus.Labels {
+	return prometheus.Labels{
+		prometheusNsLabel:       cron.GetNamespace(),
+		prometheusCronNameLabel: cron.GetName(),
 	}
 }
--- a/pkg/controller/dataimportcron-conditions.go
+++ b/pkg/controller/dataimportcron-conditions.go
@@ -88,15 +88,14 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 	}
 }
 
-func updateDataImportCronFailingMetric(cron *cdiv1.DataImportCron, increment bool) {
-	upToDateCondition := FindDataImportCronConditionByType(cron, cdiv1.DataImportCronUpToDate)
+func updateDataImportCronOutdatedMetric(upToDateCondition *cdiv1.DataImportCronCondition, increment bool) {
 	if increment {
 		if upToDateCondition == nil || upToDateCondition.ConditionState.Status != corev1.ConditionFalse {
-			DataImportCronNotUpToDateGauge.Inc()
+			DataImportCronOutdatedGauge.Inc()
 		}
 	} else {
 		if upToDateCondition.ConditionState.Status != corev1.ConditionTrue {
-			DataImportCronNotUpToDateGauge.Dec()
+			DataImportCronOutdatedGauge.Dec()
 		}
 	}
 }
--- a/pkg/controller/dataimportcron-conditions.go
+++ b/pkg/controller/dataimportcron-conditions.go
@@ -87,3 +87,16 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 		condition.Message = message
 	}
 }
+
+func updateDataImportCronFailingMetric(cron *cdiv1.DataImportCron, increment bool) {
+	upToDateCondition := FindDataImportCronConditionByType(cron, cdiv1.DataImportCronUpToDate)
+	if increment {
+		if upToDateCondition == nil || upToDateCondition.ConditionState.Status != corev1.ConditionFalse {
+			DataImportCronNotUpToDateGauge.Inc()
+		}
+	} else {
+		if upToDateCondition.ConditionState.Status != corev1.ConditionTrue {
+			DataImportCronNotUpToDateGauge.Dec()
+		}
+	}
+}
--- a/pkg/controller/dataimportcron-conditions.go
+++ b/pkg/controller/dataimportcron-conditions.go
@@ -88,13 +88,13 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 	}
 }
 
-func updateDataImportCronOutdatedMetric(upToDateCondition *cdiv1.DataImportCronCondition, increment bool) {
+func updateDataImportCronOutdatedMetric(prevUpToDateCondition *cdiv1.DataImportCronCondition, increment bool) {
 	if increment {
-		if upToDateCondition == nil || upToDateCondition.ConditionState.Status != corev1.ConditionFalse {
+		if prevUpToDateCondition == nil || prevUpToDateCondition.ConditionState.Status != corev1.ConditionFalse {
 			DataImportCronOutdatedGauge.Inc()
 		}
 	} else {
-		if upToDateCondition.ConditionState.Status != corev1.ConditionTrue {
+		if prevUpToDateCondition.ConditionState.Status != corev1.ConditionTrue {
 			DataImportCronOutdatedGauge.Dec()
 		}
 	}
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -52,8 +52,8 @@ import (
 )
 
 var (
-	// DataImportCronNotUpToDateGauge is the metric we use to alert about DataImportCrons failing
-	DataImportCronNotUpToDateGauge = prometheus.NewGauge(
+	// DataImportCronOutdatedGauge is the metric we use to alert about DataImportCrons failing
+	DataImportCronOutdatedGauge = prometheus.NewGauge(
 		prometheus.GaugeOpts{
 			Name: "kubevirt_cdi_dataimportcron_not_up_to_date_total",
 			Help: "Total number of DataImportCrons with outdated imports",
@@ -262,8 +262,9 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	digestUpdated := desiredDigest != "" && (len(imports) == 0 || desiredDigest != imports[0].Digest)
+	incrementMetric := true
+	currentUpToDateCondition := FindDataImportCronConditionByType(dataImportCron, cdiv1.DataImportCronUpToDate).DeepCopy()
 	if digestUpdated {
-		updateDataImportCronFailingMetric(dataImportCron, true)
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionFalse, "Source digest updated since last import", outdated)
 		if importSucceeded || len(imports) == 0 {
 			if err := r.createImportDataVolume(ctx, dataImportCron); err != nil {
@@ -271,19 +272,18 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			}
 		}
 	} else if importSucceeded {
-		updateDataImportCronFailingMetric(dataImportCron, false)
+		incrementMetric = false
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionTrue, "Latest import is up to date", upToDate)
 	} else if len(imports) > 0 {
-		updateDataImportCronFailingMetric(dataImportCron, true)
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionFalse, "Import is progressing", inProgress)
 	} else {
-		updateDataImportCronFailingMetric(dataImportCron, true)
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionFalse, "No source digest", noDigest)
 	}
 
 	if err := r.client.Update(ctx, dataImportCron); err != nil {
 		return res, err
 	}
+	updateDataImportCronOutdatedMetric(currentUpToDateCondition, incrementMetric)
 	return res, nil
 }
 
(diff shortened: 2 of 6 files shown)


akalenyu:
@arnongilboa This is a different failure to what I was seeing:
https://github.com/kubevirt/containerized-data-importer/blob/a5b70719588dce64ccbb0408c30eb9242e1cae98/tests/dataimportcron_test.go#L106
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2063/pull-containerized-data-importer-e2e-k8s-1.22-nfs/1471513582083510272

(I was timing out on making sure PVCs cleaned up)

pkg/controller/dataimportcron-controller.go
```diff
@@ -447,7 +457,7 @@ func (r *DataImportCronReconciler) cleanup(ctx context.Context, dataImportCron *
 	if err := r.client.Update(ctx, dataImportCron); err != nil {
 		return err
 	}
-	return nil
+	return r.checkFailingDataImportCrons()
```
arnongilboa, line 460: why do we care about it in cleanup?

pkg/controller/dataimportcron-controller.go
```diff
@@ -87,7 +97,7 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
 	dataImportCron := &cdiv1.DataImportCron{}
 	if err := r.client.Get(ctx, req.NamespacedName, dataImportCron); err != nil {
 		if k8serrors.IsNotFound(err) {
-			return reconcile.Result{}, nil
+			return reconcile.Result{}, r.checkFailingDataImportCrons()
```
arnongilboa, line 100: why do we care about it if DataImportCron not found?

pkg/controller/dataimportcron-controller.go
```diff
@@ -651,6 +661,24 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*v1be
 	return cronJob, nil
 }
 
+func (r *DataImportCronReconciler) checkFailingDataImportCrons() error {
+	numNotUpToDate := 0
+	dataImportCronList := &cdiv1.DataImportCronList{}
+	if err := r.client.List(context.TODO(), dataImportCronList); err != nil {
+		return err
```
arnongilboa, line 668: what about IsNotFound(err)? I think you should return nil in that case

pkg/controller/dataimportcron-controller.go
akalenyu: Cleanup means we need to update the metric since the "failing" cron got deleted
Otherwise we keep firing the alert even though there is not failing cron anymore

pkg/controller/dataimportcron-controller.go
```diff
@@ -271,7 +281,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	if err := r.client.Update(ctx, dataImportCron); err != nil {
 		return res, err
 	}
-	return res, nil
+	return res, r.checkFailingDataImportCrons()
```
arnongilboa, line 284: Isn't it to too much work to scan all the DataImportCrons on each singe DataImportCron update?
Returning checkFailingDataImportCrons() as the error value of the caller function seems a bit strange here.
I would do it right after the conditions update, and consider if we really want to return an error or just warn when it fails.

pkg/controller/dataimportcron-controller.go
akalenyu: If we hit some err in cleanup() and thus requeue the reconcile request, for example
(Otherwise metric value not updated to reflect cleanup)

pkg/controller/dataimportcron-controller.go
akalenyu: Then we will get an empty list and thus set the metric value to 0 which is what we want (back off with nil means we won't update the metric at all)

pkg/controller/dataimportcron-controller.go
akalenyu: Good point
I have a draft where I tried the incremental way (do it right after the conditions update) but it looks less readable especially in the parts where we have to decrement the metric value

Do you think the list call is that bad, even though it's cached?

pkg/operator/controller/prometheus.go
```diff
@@ -214,6 +214,20 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 								componentAlertLabelKey: componentAlertLabelValue,
 							},
 						),
+						generateAlertRule(
+							"CDIDataImportCronNotUpToDate",
+							"kubevirt_cdi_dataimportcron_not_up_to_date_total > 0",
+							"15m",
+							map[string]string{
+								"summary":     "DataImportCron PVCs do not hold the latest OS image, disk might not contain latest OS version",
```
arnongilboa, line 222: DataImportCron PVCs do not hold the latest OS image, disk might not contain latest OS version
Why not simply "DataImportCron latest imports are outdated"?

pkg/controller/dataimportcron-controller.go
arnongilboa: Ok, so why not do a single call to checkFailingDataImportCrons() in the Reconcile() instead of calling it all over?

pkg/controller/dataimportcron-controller.go
arnongilboa: see my comment on doing it once in the Reconcile

pkg/controller/dataimportcron-controller.go
```diff
@@ -50,6 +51,15 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/util"
 )
 
+var (
+	// DataImportCronNotUpToDateGauge is the metric we use to alert about DataImportCrons failing
+	DataImportCronNotUpToDateGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kubevirt_cdi_dataimportcron_not_up_to_date_total",
+			Help: "Total number of DataImportCrons that are failing and as a result not keeping OS images up to date",
```
arnongilboa, line 59: Why not simply "Total number of DataImportCrons with outdated imports"

tests/operator_test.go
```diff
@@ -864,6 +865,67 @@ var _ = Describe("ALL Operator tests", func() {
...
+				if utils.IsOpenshift(f.K8sClient) {
+					url = externalRegistryURL
+				} else {
+					url = trustedRegistryURL
+					err = utils.AddInsecureRegistry(f.CrClient, url)
+					Expect(err).To(BeNil())
+
+					hasInsecReg, err := utils.HasInsecureRegistry(f.CrClient, url)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(hasInsecReg).To(BeTrue())
```
arnongilboa, line 889: not really needed

tests/operator_test.go
```diff
@@ -864,6 +865,67 @@ var _ = Describe("ALL Operator tests", func() {
...
+					var currentImportDv string
+					By("Wait for CurrentImports DataVolumeName update")
+					Eventually(func() bool {
+						cron, err = f.CdiClient.CdiV1beta1().DataImportCrons(f.Namespace.Name).Get(context.TODO(), cron.Name, metav1.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
+						if len(cron.Status.CurrentImports) == 0 {
+							return false
+						}
+						currentImportDv = cron.Status.CurrentImports[0].DataVolumeName
+						return currentImportDv != ""
```
arnongilboa, line 907: why not simply wait for condition UpToDate = true?

pkg/controller/dataimportcron-controller.go
akalenyu: We will call it once per Reconcile() as it is right now if I'm not mistaken:
- IsNotFound case backs off when done (cleanup() wont be called)
- cleanup also backs off when done (and only executed if wasn't IsNotFound)
- update() is only called if we never reached the two above (cleanup, IsNotFound)

tests/operator_test.go
akalenyu: Right, I guess that indirectly tells me DV is succeeded, thanks

pkg/controller/dataimportcron-controller.go
arnongilboa: I meant that it's better that the single call to checkFailingDataImportCrons() will appear only in Reconcile().

akalenyu: We can't do a single call because there are a few code branches that need to call it (the ones listed above)

akalenyu: Done, thanks for the tip, defer used in cleanup()

pkg/controller/dataimportcron-controller.go
```diff
@@ -264,24 +264,28 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	digestUpdated := desiredDigest != "" && (len(imports) == 0 || desiredDigest != imports[0].Digest)
 	if digestUpdated {
+		updateDataImportCronFailingMetric(dataImportCron, true)
```
arnongilboa, line 267: why call it 4 times if you can call it once from updateDataImportCronCondition()?

pkg/controller/dataimportcron-conditions.go
```diff
@@ -87,3 +87,16 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 		condition.Message = message
 	}
 }
+
+func updateDataImportCronFailingMetric(cron *cdiv1.DataImportCron, increment bool) {
+	upToDateCondition := FindDataImportCronConditionByType(cron, cdiv1.DataImportCronUpToDate)
+	if increment {
+		if upToDateCondition == nil || upToDateCondition.ConditionState.Status != corev1.ConditionFalse {
+			DataImportCronNotUpToDateGauge.Inc()
```
arnongilboa, line 95: `DataImportCronOutdatedGauge` might be a better name?

pkg/controller/dataimportcron-conditions.go
```diff
@@ -87,3 +87,16 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 		condition.Message = message
 	}
 }
+
+func updateDataImportCronFailingMetric(cron *cdiv1.DataImportCron, increment bool) {
```
arnongilboa, line 91: `updateDataImportCronOutdatedMetric` might be a better name?

akalenyu: `updateDataImportCronOutdatedMetric()` is the better choice, thx
Prefer to keep the straightforward `increment` word so there is no place for misunderstanding about whats going to happen

pkg/controller/dataimportcron-controller.go
```diff
@@ -253,6 +263,8 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	digestUpdated := desiredDigest != "" && (len(imports) == 0 || desiredDigest != imports[0].Digest)
+	incrementMetric := true
+	currentCronCopy := dataImportCron.DeepCopy()
```
akalenyu, line 267: @arnongilboa To prevent a situation where I increment/decrement but fail r.client.Update() as we discussed
Tagging it here so the change doesn't slip away unreviewed

arnongilboa: all you are interested is in the current `upToDateCondition.ConditionState.Status`, so why DeepCopy is needed?

akalenyu: also need to check `upToDateCondition != nil`

akalenyu: Done

pkg/controller/dataimportcron-controller.go
```diff
@@ -263,28 +263,28 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	digestUpdated := desiredDigest != "" && (len(imports) == 0 || desiredDigest != imports[0].Digest)
+	incrementMetric := true
+	currentUpToDateCondition := FindDataImportCronConditionByType(dataImportCron, cdiv1.DataImportCronUpToDate).DeepCopy()
```
arnongilboa, line 267: again, why use and DeepCopy and pass the Condition if you can use a bool?
```
wasUpToDate := false 
if condition := FindDataImportCronConditionByType(cron, conditionType); condition != nil {
   wasUpToDate = condition.Status == corev1.ConditionFalse
}
```

pkg/controller/dataimportcron-controller.go
```diff
@@ -263,28 +263,28 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	digestUpdated := desiredDigest != "" && (len(imports) == 0 || desiredDigest != imports[0].Digest)
+	incrementMetric := true
```
arnongilboa, line 266: Be positive and use a name like `upToDate` instead here and in the func args

pkg/controller/dataimportcron-controller.go
akalenyu: Tried to fiddle around with it, I prefer to keep this chunk hidden inside `updateDataImportCronCondition()` if that's ok

pkg/controller/dataimportcron-controller.go
akalenyu: Why not straight up say if inc/dec is the desired action?

pkg/controller/dataimportcron-conditions.go
```diff
@@ -88,13 +88,13 @@ func updateConditionState(condition *cdiv1.ConditionState, status corev1.Conditi
 	}
 }
 
-func updateDataImportCronOutdatedMetric(upToDateCondition *cdiv1.DataImportCronCondition, increment bool) {
+func updateDataImportCronOutdatedMetric(prevUpToDateCondition *cdiv1.DataImportCronCondition, increment bool) {
```
akalenyu, line 91: Updated the parameter name here to reflect better that we're looking at the previous condition

awels (commented):
Sorry for being a bit of pain. I don't think the way it is now is going to work. We are storing state in the controller (the current value of the gauge) and that could disappear at any point in time (pod with controller gets killed). I think Alex had the right idea in the first iteration of this PR. At the end of the reconcile loop (right before or right after the update of the DIC) we update the gauge based on the listing of all the DICs and checking which ones are outdated.

akalenyu:
> Sorry for being a bit of pain. I don't think the way it is now is going to work. We are storing state in the controller (the current value of the gauge) and that could disappear at any point in time (pod with controller gets killed). I think Alex had the right idea in the first iteration of this PR. At the end of the reconcile loop (right before or right after the update of the DIC) we update the gauge based on the listing of all the DICs and checking which ones are outdated.

Yes you are right, thanks!
I am now doing things differently - the metric is now displaying info per cron (`kubevirt_cdi_dataimportcron_outdated{ns='default', cron_name='cron-test'}`)

pkg/controller/dataimportcron-conditions.go
```diff
@@ -34,6 +35,14 @@ const (
 )
 
 func updateDataImportCronCondition(cron *cdiv1.DataImportCron, conditionType cdiv1.DataImportCronConditionType, status corev1.ConditionStatus, message, reason string) {
+	if conditionType == cdiv1.DataImportCronUpToDate {
+		outDated := status != corev1.ConditionTrue
+		if outDated {
```
awels, line 40: Minor style change.
```go
if status != corev1.ConditionTrue {
  DataImportCronOutdatedGauge.With(getPrometheusCronLabels(cron)).Set(1)
} else {
  DataImportCronOutdatedGauge.With(getPrometheusCronLabels(cron)).Set(0)
}
```

tests/operator_test.go
```diff
@@ -872,6 +873,68 @@ var _ = Describe("ALL Operator tests", func() {
...
+					cron := NewDataImportCron(fmt.Sprintf("cron-test-%d", i), "5Gi", scheduleOnceAYear, fmt.Sprintf("datasource-test-%d", i), cdiv1.DataVolumeSourceRegistry{URL: &url, PullMethod: &registryPullNode}, retentionPolicy)
+					By(fmt.Sprintf("Create new DataImportCron %s", url))
+					cron, err = f.CdiClient.CdiV1beta1().DataImportCrons(f.Namespace.Name).Create(context.TODO(), cron, metav1.CreateOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					By("Wait for condition UpToDate=true on DataImportCron")
+					Eventually(func() bool {
+						cron, err = f.CdiClient.CdiV1beta1().DataImportCrons(f.Namespace.Name).Get(context.TODO(), cron.Name, metav1.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
+						upToDateCondition := controller.FindDataImportCronConditionByType(cron, cdiv1.DataImportCronUpToDate)
+						if upToDateCondition != nil && upToDateCondition.ConditionState.Status == corev1.ConditionTrue {
```
awels, line 908: can't you do:
```go
  return upToDateCondition != nil && upToDateCondition.ConditionState.Status == corev1.ConditionTrue
```

awels (commented):
looks good some minor comments.

pkg/controller/dataimportcron-conditions.go
akalenyu: Done

tests/operator_test.go
akalenyu: Yep, done

awels (commented):
/lgtm
/approve

PR #2066, opened 2021-12-21
On CDI delete validation ignore DVs labeled with DataImportCron

arnongilboa:

`DataImportCron` `RetentionPolicy` specifies whether the created `DataVolumes` and `DataSources` are retained when their `DataImportCron` is deleted. Default is `RetainAll`. However, when CDI delete validation finds existing DVs, it fails (if `UninstallStrategy` is `cdiv1.CDIUninstallStrategyBlockUninstallIfWorkloadsExist`). Therefore, on CDI delete validation we are now ignoring DVs labeled with `DataImportCron` so it won't fail if they are the only existing DVs.

**Release note**:
```release-note
NONE
```


awels (approved):
Took me a second to realize that in the test you are leaving the DIC labeled DV and deleting the one blocking the delete.

PR #2076, opened 2022-01-03
update builder golang to 1.17 remove kubebuilder

mhenriks:

1.17 needed for k8s 1.23 libs.  kubebuilder binaries removed because they were changed and IMO not needed for unit tests

Fixes #

**Release note**:

```release-note
Update golang in builder image
```

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -1,14 +1,20 @@
-FROM quay.io/centos/centos:stream9
+FROM registry.fedoraproject.org/fedora-minimal:33
 MAINTAINER "The KubeVirt Project" <kubevirt-dev@googlegroups.com>
 
-RUN 	dnf -y install dnf-plugins-core && \
-	dnf config-manager --set-enable crb && dnf update -y && \
-	dnf install -y \
+COPY fedora.repo /tmp/fedora_ci.dnf.repo
+
+RUN sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
+	cat /tmp/fedora_ci.dnf.repo && \
+	mkdir /etc/yum.repos.d/old && \
+	mv /etc/yum.repos.d/*.repo /etc/yum.repos.d/old  && \
+	mv /tmp/fedora_ci.dnf.repo /etc/yum.repos.d/fedora.repo && \
+	microdnf update -y && microdnf install -y \
 	cpio \
 	diffutils \
 	git \
 	python3-pip \
 	python3-devel \
+	mercurial \
 	gcc \
 	gcc-c++ \
 	glibc-devel \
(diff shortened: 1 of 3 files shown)


hack/build/docker/builder/Dockerfile
```diff
@@ -1,14 +1,20 @@
-FROM quay.io/centos/centos:stream9
```
brybacki, line 1: so we have just switched from fedora to centos stream and now back to fedora @maya-r @mhenriks is that ok?

maya-r: the centos builder had some issues, I hope with https://github.com/kubevirt/project-infra/pull/1824 it will be fixed so we don't need to go back.

brybacki: This change introduces `FROM registry.fedoraproject.org/fedora-minimal:33` instead of centos, 
I remember that we have just switched to centos. If I remember you did the change @maya-r. That is why I have asked for your opinion.

So my question is: What was the reason to change from fedora to centos, and why now we go back to fedora? and which one is better?

mhenriks: @maya-r there are additional issues with centos stream.  Specifically the way bazel installs rpms in containers.  So I would like to go back to fedora (although we technically never used the centos builder).  Will make bazel changes in another pr

maya-r: I have been able to update bazel/gazelle/k8s libraries in [this branch](https://github.com/maya-r/containerized-data-importer/tree/wip-k8s-1.23), but I don't mind going back to fedora and re-doing the centos builder change if it is easier for you that way.
(the centos stream CI connectivity issue should be resolved now)

mhenriks: Great, let's get this massive pr out of the way then integrate your builder/bazel changes

hack/build/docker/builder/Dockerfile
```diff
@@ -1,14 +1,20 @@
...
-RUN 	dnf -y install dnf-plugins-core && \
-	dnf config-manager --set-enable crb && dnf update -y && \
-	dnf install -y \
+COPY fedora.repo /tmp/fedora_ci.dnf.repo
+
+RUN sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
+	cat /tmp/fedora_ci.dnf.repo && \
+	mkdir /etc/yum.repos.d/old && \
+	mv /etc/yum.repos.d/*.repo /etc/yum.repos.d/old  && \
+	mv /tmp/fedora_ci.dnf.repo /etc/yum.repos.d/fedora.repo && \
```
awels, line 10: So 
```
sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
	cat /tmp/fedora_ci.dnf.repo && \
	mkdir /etc/yum.repos.d/old && \
	mv /etc/yum.repos.d/*.repo /etc/yum.repos.d/old  && \
	mv /tmp/fedora_ci.dnf.repo /etc/yum.repos.d/fedora.repo &&
```
Was needed a long time ago when we ran in the oVirt CI system. I don't think this is needed anymore.

hack/build/docker/builder/Dockerfile
```diff
@@ -1,14 +1,20 @@
-FROM quay.io/centos/centos:stream9
+FROM registry.fedoraproject.org/fedora-minimal:33
 MAINTAINER "The KubeVirt Project" <kubevirt-dev@googlegroups.com>
 
-RUN 	dnf -y install dnf-plugins-core && \
-	dnf config-manager --set-enable crb && dnf update -y && \
-	dnf install -y \
+COPY fedora.repo /tmp/fedora_ci.dnf.repo
```
awels, line 4: This was for oVirt CI as well, don't need it, as well as the repo file is not needed.

hack/build/docker/builder/Dockerfile
mhenriks: good point but I'd like to ignore for now.  Hopefully moving to centos stream soon after this

PR #2079, opened 2022-01-03
Add test cases for runbook URLs

assafad:
This PR adds new tests cases for alert definitions. We expect to see "runbook_url" annotations in alerts and those urls must be accessible.

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"reflect"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


assafad:
Hi @erkanerol , @sradco , @akalenyu , can you please review this PR?

akalenyu:
/ok-to-test

We have a case that grabs the prometheusrule:
https://github.com/kubevirt/containerized-data-importer/blob/10c85dc73a11e1ec97defdc7a68923d05af6f010/pkg/operator/controller/controller_test.go#L331
If you want to use that 

Otherwise, a little unsure on unit tests making real HTTP call. how horrible is that? does it justify a functest?
(this is just to make sure the runbook page exists i.e. there is a runbook for the alert)

PR #2084, opened 2022-01-06
Set ActiveDeadline for DataImportCron poller job

arnongilboa:

Mostly for the case the job pod is stuck `Pending` (e.g. due to failed MountVolume.SetUp) and not terminated even when deleting its DataImportCron (and verifying no cronjob/job left). 
` Warning  FailedMount  24s (x8 over 87s)  kubelet            MountVolume.SetUp failed for volume "cdi-cert-vol" : configmap "no-certs" not found`

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -640,6 +640,7 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*v1be
 
 	var successfulJobsHistoryLimit int32 = 0
 	var failedJobsHistoryLimit int32 = 0
+	var activeDeadlineSeconds int64 = 120
 	var ttlSecondsAfterFinished int32 = 0
 	var backoffLimit int32 = 2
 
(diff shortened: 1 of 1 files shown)


mhenriks:
not sure about this one.  If the user configured the cert why not wait for it?  Assuming these jobs won't pile up over time

arnongilboa:
> not sure about this one. If the user configured the cert why not wait for it? Assuming these jobs won't pile up over time

@mhenriks as mentioned, the issue was that when the user misconfigured the cert, the pod will be stuck Pending and not terminated even when deleting its DataImportCron (and verifying no cronjob/job left).

pkg/controller/dataimportcron-controller.go
```diff
@@ -640,6 +640,7 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*v1be
 
 	var successfulJobsHistoryLimit int32 = 0
 	var failedJobsHistoryLimit int32 = 0
+	var activeDeadlineSeconds int64 = 120
```
awels, line 643: So the job just updates the digest on the DIC right? And this would put a limit of 2 minutes on it succeeding with that?

PR #2085, opened 2022-01-06
Add DataSource controller to update the Ready condition

arnongilboa:

DataSource is no more dependent on a DataImportCron to update the Ready status condition

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -209,6 +209,10 @@ func start(ctx context.Context, cfg *rest.Config) {
 		klog.Errorf("Unable to setup dataimportcron controller: %v", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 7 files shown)


mhenriks:
New functional tests?  Or point to existing that cover everything?

arnongilboa:
> New functional tests? Or point to existing that cover everything?
@mhenriks DataSource Ready status condition is already basically tested in
https://github.com/kubevirt/containerized-data-importer/blob/main/tests/dataimportcron_test.go#L118-L129
Not sure we need to add other tests there which are covered by the utests

awels:
So the way I read this is that the update of the datasource moved from the DIC controller to its own controller and the functional tests for the DIC controller cover the functionality which really didn't change.

PR #2086, opened 2022-01-09
Remove tag from DataImportCron registry source URL

arnongilboa:

There are cases when DataImportCron registry source URL contains tag referring a specific version. We use this tagged URL when polling for source updates, but on import we would like to remove the tag as we use a specific sha256.
E.g. for `docker://quay.io/containerdisks/centos:8.4` the `:8.4` tag will be removed.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dominikholler (approved):
/lgtm
@arnongilboa Thanks a lot for introducing this improvement!

dominikholler:
/lgtm
@arnongilboa thanks for the improvement

awels:
I was looking at maybe looking some existing library functions to strip the tag.

arnongilboa:
> I was looking at maybe looking some existing library functions to strip the tag.

Sure looked for it, but unfortunately couldn't find. Will be happy to find one.

awels:
So using the [docker reference](https://github.com/kubevirt/containerized-data-importer/tree/main/vendor/github.com/containers/image/v5/docker/reference) code you can parse the URL and it should return you a reference. If the URL contains a tag it should be a `NamedTagged` or `Tagged` reference. Doing a check for either might be a little overkill, but I would at least use the Regex found [here](https://github.com/kubevirt/containerized-data-importer/blob/main/vendor/github.com/containers/image/v5/docker/reference/regexp.go#L68) to capture the right things in a consistent fashion.

PR #2081, opened 2022-01-05
Use directly official libnbd go bindings

alicefr:
**What this PR does / why we need it**:
Currently, CDI relies for the libnbd bindings on [mrnold/go-libnbd](https://github.com/mrnold/go-libnbd). The major issue in using the official go bindings and bazel consists that gazelle doesn't support pkg-conf. Consequently, the link and cflags are not automatically generated together with the BUILD.bazel file. A possible workaround is to add the missing flags using buildozer. 

The support for cgo and pkg-config has been refused upstream (see https://github.com/bazelbuild/bazel-gazelle/pull/556). Hence, I don't think we can hope that it will be added in the future.
   

**Special notes for your reviewer**:
This PR manually adds the flags not generated by gazelle at the BUILD.bazel creation

**Release note**:

```release-note
None
```

PR #1983, opened 2021-10-11
Use centos:stream9 as a base image, cleanup unused code

maya-r:
**What this PR does / why we need it**:
[KubeVirt has done similar with centos stream8](https://github.com/kubevirt/kubevirt/pull/6177).
Fedora has some downsides: frequent releases, and shorter end of life.
CentOS stream will mean fewer base image changes, and it means we use a base image which has vulnerabilities reported to it, [as requested on kubevirt-dev](https://groups.google.com/g/kubevirt-dev/c/Vu89kytp2mQ).

**Special notes for your reviewer**:
Blocked on #1982 which changes the builder image.
Biggest change was switching away from what might be a self-made Fedora RPM for golang-govmomi to downloading the releases and untarring with the entrypoint.sh file.
VDI test now uses a pre-generated tinyCore.vdi as the qemu-img vdi support is read-only.

**Release note**:

```release-note
Use centos:stream9 as a base image for our containers.
```

PR #2087, opened 2022-01-09
Switch to CentOS stream9 builder

maya-r:
Had some trouble building the centos builder on CI.
@mhenriks reverted it to be able to make some changes - let's re-introduce this change.

**Release note**:

```release-note
Switch to CentOS Stream9 for building images
```


maya-r:
/hold
needs https://github.com/kubevirt/project-infra/pull/1829 first.

hack/build/docker/builder/Dockerfile
```diff
@@ -30,35 +24,37 @@ RUN sed -i 's/proxy = None//gI' /tmp/fedora_ci.dnf.repo && \
...
+    source /etc/profile.d/gimme.sh && \
+    eval $(go env) && \
+    go get github.com/onsi/ginkgo/ginkgo && \
+    go get golang.org/x/tools/cmd/goimports && \
+    go get mvdan.cc/sh/cmd/shfmt && \
+    go get github.com/mattn/goveralls && \
+    go get -u golang.org/x/lint/golint && \
+    go get github.com/rmohr/go-swagger-utils/swagger-doc && \
+    go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 && \
+    go get github.com/securego/gosec/v2/cmd/gosec@0ce48a5 && \
```
mhenriks, line 56: Why going back to `go get` Isn't it deprecated?

maya-r: Didn't notice this is a change I should adopt, will change.

maya-r:
/hold cancel
bootstrap image was updated.

awels:
So #1983 is for updating the output containers to use centos stream 9? I am not sure if we can separate the two. The builder image builds the binaries which eventually end up in the target containers. So we would have binaries build on centos 9 being deposited into fedora 33 containers. I am fine with a short period of this being out of sync, but is #1983 ready as well?

alicefr:
If you don't switch at the same time the builder and the base container image, you could end up having a mismatch with the dynamic libraries and symbols. It already happen with a too old fedora package for libnbd (see [this comment](https://github.com/kubevirt/containerized-data-importer/pull/2081#issuecomment-1007242976)). This could happen even more likely if the 2 environments have 2 different OSes.

maya-r:
Yeah, unfortunately our current workflow requires there to be a small window of mismatch between builder and base image.
We only update the builder after the push is done, and cannot use it for the PR we are currently building.

maya-r:
@awels both PRs are as ready as they can be - the second one needs to update BUILDER_IMAGE to the checksum that will be generated once this PR is merged.

PR #2095, opened 2022-01-12
Update DataSource Ready condition on DV update

arnongilboa:

Added the DataSource controller a DV watch, so whenever a DataSource
referred DV/PVC is created/updated/deleted, the DataSource Ready status
condition is updated accordingly.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/datasource-controller.go
+++ b/pkg/controller/datasource-controller.go
@@ -30,8 +30,10 @@ import (
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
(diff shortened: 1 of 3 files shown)


arnongilboa:
/hold

Working on a cheaper DV watch logic, as the current one is quite expensive.

pkg/controller/datasource-controller.go
```diff
@@ -121,24 +126,88 @@ func NewDataSourceController(mgr manager.Manager, log logr.Logger, installerLabe
...
+			CreateFunc: func(e event.CreateEvent) bool { return true },
+			DeleteFunc: func(e event.DeleteEvent) bool { return true },
+			// Only DV status phase update is interesting to reconcile
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				dvOld, okOld := e.ObjectOld.(*cdiv1.DataVolume)
+				dvNew, okNew := e.ObjectNew.(*cdiv1.DataVolume)
+				return okOld && okNew && dvOld.Status.Phase != dvNew.Status.Phase
+			},
+		},
+	); err != nil {
```
mhenriks, line 175: Took me a while to figure out what is going on here.  Looks like the main reason for the map is to associate a PVC to a list of DataSources.  Fine.  Turns out that kubernetes informers have a standard way to build/maintain such indexes.  And controller runtime has their own way to manage them.  Best example I could find is here:  https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/client/example_test.go#L249-L267

You should be able to call `mgr.GetFieldIndexer()` in `NewDataSourceController` function to set everything up.  Then `client.List` can be used.

Alternatively, just list all DataSources in the DataVolume namespace to figure out which one to queue.  They should be cached.  At least that is easy to understand even if performance isn't optimal.

arnongilboa: The second alternative you just suggested is the first commit in this PR, which is quite expansive. Give it a look.
I'll try to switch to informers.

arnongilboa: Fixed

pkg/controller/datasource-controller.go
```diff
@@ -126,19 +128,72 @@ func NewDataSourceController(mgr manager.Manager, log logr.Logger, installerLabe
...
+
+	const dataSourcePvcField = "spec.source.pvc"
+
+	getKey := func(namespace, name string) string {
+		return namespace + "." + name
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &cdiv1.DataSource{}, dataSourcePvcField, func(obj client.Object) []string {
+		if pvc := obj.(*cdiv1.DataSource).Spec.Source.PVC; pvc != nil {
+			return []string{getKey(pvc.Namespace, pvc.Name)}
```
mhenriks, line 160: It looks like you can just give the name and the library will add namespace prefix:  https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/informer_cache.go#L180-L217

arnongilboa: pvc.Namespace can be != DataSource.Namespace, and there can be same pvc names on different ns, so NO way the library can add the pvc ns prefix.

pkg/controller/datasource-controller.go
```diff
@@ -126,19 +128,72 @@ func NewDataSourceController(mgr manager.Manager, log logr.Logger, installerLabe
...
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	mapToDataSource := func(obj client.Object) (reqs []reconcile.Request) {
+		var dataSources cdiv1.DataSourceList
+		matchingFields := client.MatchingFields{dataSourcePvcField: getKey(obj.GetNamespace(), obj.GetName())}
+		mgr.GetClient().List(context.TODO(), &dataSources, matchingFields)
```
mhenriks, line 170: should we log any error here and return?

pkg/controller/datasource-controller.go
arnongilboa: We cannot use the `DataSource.namespace` for the pvc, since we don't know the `DataSource.namespace` when we call `getKey()` for creating `client.MatchingFields` for the `List()`. However, since we don't have  `dataSourceValidatingWebhook`, when `DataSource.spec.source.pvc` is updated and its namespace is emty we can update it to the DataSource namespace. Sounds ok?

mhenriks: I don't understand.  How is the datasource namespace relevant to the list call?  Take the following:

```yaml
apiversion: cdi.kubevirt.io/v1beta1
kind: DataSource
metadata:
  namespace: ns1
  name: ds1
  spec:
    source:
      pvc:
        name: dv1
  ...
---
apiversion: cdi.kubevirt.io/v1beta1
kind: DataVolume:
metadata:
  namespace: ns1
  name: dv1
 ...
```


Given the current code, the index is:

`/dv1` -> [`ns1/ds1`]

and

`mapToDataSource` will lookup `ns1/dv1` and should come up empty, no?

arnongilboa: Right, my mistake. Fixing.

pkg/controller/datasource-controller.go
arnongilboa: added

PR #2096, opened 2022-01-13
BugId: 2038679 -  Clone with volume mode file system using Storage API fails

brybacki:
Fixes a problem with validation webhook failing for a clone that should succeed.

Validation webhook compares real PVC size of source volume with the requested size on spec.storage. 

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=2038679

Still experimenting 
**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -22,6 +22,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"math"
 	"net/http"
 	"reflect"
@@ -1566,7 +1567,7 @@ func (r *DatavolumeReconciler) advancedClonePossible(dataVolume *cdiv1.DataVolum
 		return false, err
 	}
 
-	return r.validateAdvancedCloneSizeCompatible(sourcePvc, targetStorageSpec)
+	return r.validateAdvancedCloneSizeCompatible(dataVolume, sourcePvc, targetStorageSpec)
 }
 
 func (r *DatavolumeReconciler) validateSameStorageClass(
@@ -1609,18 +1610,19 @@ func (r *DatavolumeReconciler) validateSameVolumeMode(
 }
 
 func (r *DatavolumeReconciler) validateAdvancedCloneSizeCompatible(
+	dataVolume *cdiv1.DataVolume,
 	sourcePvc *corev1.PersistentVolumeClaim,
 	targetStorageSpec *corev1.PersistentVolumeClaimSpec) (bool, error) {
-
 	srcStorageClass := &storagev1.StorageClass{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *sourcePvc.Spec.StorageClassName}, srcStorageClass); IgnoreNotFound(err) != nil {
 		return false, err
 	}
 
+	srcRequest, hasSrcRequest := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
 	srcCapacity, hasSrcCapacity := sourcePvc.Status.Capacity[corev1.ResourceStorage]
 	targetRequest, hasTargetRequest := targetStorageSpec.Resources.Requests[corev1.ResourceStorage]
 	allowExpansion := srcStorageClass.AllowVolumeExpansion != nil && *srcStorageClass.AllowVolumeExpansion
-	if !hasSrcCapacity || !hasTargetRequest {
+	if !hasSrcRequest || !hasSrcCapacity || !hasTargetRequest {
 		// return error so we retry the reconcile
 		return false, errors.New("source/target size info missing")
 	}
@@ -1628,8 +1630,36 @@ func (r *DatavolumeReconciler) validateAdvancedCloneSizeCompatible(
 	if srcCapacity.Cmp(targetRequest) < 0 && !allowExpansion {
 		return false, nil
 	}
+	// todo: need to check for pvc - storage, storage-pvc, pvc-pvc and storage-storage combinations.
+	usableSpace, err := r.calculateUsableSpace(srcStorageClass, sourcePvc.Spec.VolumeMode, srcRequest)
+	if err != nil {
+		return false, err
+	}
+	if usableSpace.Cmp(targetRequest) > 0 {
+		message := "target resources requested storage size is smaller than the source requested size"
+		r.recorder.Event(dataVolume, corev1.EventTypeWarning, ErrIncompatiblePVC, message)
+		return false, errors.New(message)
+	}
 
 	return true, nil
+}
+
+func (r *DatavolumeReconciler) calculateUsableSpace(srcStorageClass *storagev1.StorageClass,
+	mode *corev1.PersistentVolumeMode,
+	srcRequest resource.Quantity) (resource.Quantity, error) {
+
+	if resolveVolumeMode(mode) == corev1.PersistentVolumeFilesystem {
+		fsOverhead, err := GetFilesystemOverheadForStorageClass(r.client, &srcStorageClass.Name)
+		if err != nil {
+			return resource.Quantity{}, err
+		}
+		fsOverheadFloat, _ := strconv.ParseFloat(string(fsOverhead), 64)
+		usableSpace := GetUsableSpace(fsOverheadFloat, srcRequest.Value())
+
+		return *resource.NewScaledQuantity(usableSpace, 0), nil
+	}
+
+	return srcRequest, nil
 }
 
 func (r *DatavolumeReconciler) getCloneStrategy(dataVolume *cdiv1.DataVolume) (*cdiv1.CDICloneStrategy, error) {
@@ -2479,4 +2509,13 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
 
 func newLongTermCloneTokenGenerator(key *rsa.PrivateKey) token.Generator {
 	return token.NewGenerator(common.ExtendedCloneTokenIssuer, key, 10*365*24*time.Hour)
+}
+
+// GetUsableSpace calculates space to use taking file system overhead into account
+func GetUsableSpace(filesystemOverhead float64, availableSpace int64) int64 {
+	// +1 always rounds up.
+	spaceWithOverhead := int64(math.Ceil((1 - filesystemOverhead) * float64(availableSpace)))
+	// qemu-img will round up, making us use more than the usable space.
+	// This later conflicts with image size validation.
+	return util.RoundDown(spaceWithOverhead, util.DefaultAlignBlockSize)
 }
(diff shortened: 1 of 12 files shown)


brybacki:
/retest
trying to find a pattern

brybacki:
found it, needed to rebase (sadly github did not see the need ;))
/retest

pkg/controller/datavolume-controller.go
```diff
@@ -2480,3 +2510,12 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
 func newLongTermCloneTokenGenerator(key *rsa.PrivateKey) token.Generator {
 	return token.NewGenerator(common.ExtendedCloneTokenIssuer, key, 10*365*24*time.Hour)
 }
+
+// GetUsableSpace calculates space to use taking file system overhead into account
+func GetUsableSpace(filesystemOverhead float64, availableSpace int64) int64 {
```
awels, line 2515: Would it make sense to push this up into pkg/util/util.go so it can be used here and in data-processor.go which has the same function already.

tests/cloner_test.go
```diff
@@ -466,6 +468,87 @@ var _ = Describe("all clone tests", func() {
...
+				targetMD5, err := f.GetMD5(f.Namespace, targetPvc, diskImagePath, crossVolumeModeCloneMD5NumBytes)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(sourceMD5 == targetMD5).To(BeTrue())
+				By("Deleting verifier pod")
+				err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("[rfe_id:1126][crit:High][vendor:cnv-qe@redhat.com][level:component] Should fail with Event when cloning into a smaller sized data volume", func() {
+				By("Creating a source from a real image") // todo fs....
```
awels, line 520: Did you do this todo?

tests/cloner_test.go
```diff
@@ -801,23 +884,24 @@ var _ = Describe("all clone tests", func() {
 		})
 
 		It("[rfe_id:1126][test_id:1896][crit:High][vendor:cnv-qe@redhat.com][level:component] Should not allow cloning into a smaller sized data volume", func() {
-			By("Creating a source from a real image")
+			By("Creating a source from a real image") // todo fs....
```
awels, line 887: Did you finish this todo?

tests/cloner_test.go
```diff
@@ -508,6 +515,40 @@ var _ = Describe("all clone tests", func() {
...
+				By("Waiting for import to be completed")
+				utils.WaitForDataVolumePhaseWithTimeout(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, sourceDv.Name, 3*90*time.Second)
+
+				By("Cloning from the source DataVolume to under sized target")
+				targetDv := utils.NewDataVolumeForImageCloningAndStorageSpec("target-dv", "100Mi",
+					f.Namespace.Name,
+					sourceDv.Name,
+					sourceDv.Spec.PVC.StorageClassName,
+					sourceDv.Spec.PVC.VolumeMode)
+				fmt.Fprintf(GinkgoWriter, "targetDv %v", targetDv)
```
ShellyKa13, line 539: same here

tests/cloner_test.go
```diff
@@ -508,6 +515,40 @@ var _ = Describe("all clone tests", func() {
...
+
+				targetDv, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDv)
+				Expect(err).ToNot(HaveOccurred())
+				// As of now, csi and smart clone check the values before the pvc is created, and the network clone
+				// checks it in the upload phase, so it needs a PVC, this might be improved
```
ShellyKa13, line 544: so should a check a pvc was created or not should be added here?

tests/cloner_test.go
```diff
@@ -508,6 +515,40 @@ var _ = Describe("all clone tests", func() {
...
+				fmt.Fprintf(GinkgoWriter, "sourceDv %v", sourceDv)
+
+				Expect(err).ToNot(HaveOccurred())
+				f.ForceBindPvcIfDvIsWaitForFirstConsumer(sourceDv)
+
+				By("Waiting for import to be completed")
+				utils.WaitForDataVolumePhaseWithTimeout(f.CdiClient, f.Namespace.Name, cdiv1.Succeeded, sourceDv.Name, 3*90*time.Second)
+
+				By("Cloning from the source DataVolume to under sized target")
+				targetDv := utils.NewDataVolumeForImageCloningAndStorageSpec("target-dv", "100Mi",
```
ShellyKa13, line 534: not sure if it is possible but wasn't the bug found because the requested size was the same size as the src but the fsoverhead wasnt taken into acount? putting 200Mi will work?

tests/cloner_test.go
brybacki: yes

tests/cloner_test.go
brybacki: copy paste of a previous one at line 887

brybacki: removed

pkg/controller/datavolume-controller.go
```diff
@@ -1609,29 +1610,58 @@ func (r *DatavolumeReconciler) validateSameVolumeMode(
...
+	dataVolume *cdiv1.DataVolume,
 	sourcePvc *corev1.PersistentVolumeClaim,
 	targetStorageSpec *corev1.PersistentVolumeClaimSpec) (bool, error) {
-
 	srcStorageClass := &storagev1.StorageClass{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *sourcePvc.Spec.StorageClassName}, srcStorageClass); IgnoreNotFound(err) != nil {
 		return false, err
 	}
 
+	srcRequest, hasSrcRequest := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
```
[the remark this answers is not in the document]
brybacki: very good question, 
and yes, because this is the actual PVC, not the fields from DataVolume, and the PVC is resized by changing the actual request size on the PVC.Spec.Resources

tests/cloner_test.go
```diff
@@ -508,6 +515,40 @@ var _ = Describe("all clone tests", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 			})
+
+			It("[rfe_id:1126][crit:High][vendor:cnv-qe@redhat.com][level:component] Should fail with Event when cloning into a smaller sized data volume", func() {
+				By("Creating a source from a real image") // todo fs....
+				sourceDv := utils.NewDataVolumeWithHTTPImport("source-dv", "200Mi", tinyCoreIsoURL())
+				filesystem := v1.PersistentVolumeFilesystem
+				sourceDv.Spec.PVC.VolumeMode = &filesystem
+				sourceDv, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, sourceDv)
+				fmt.Fprintf(GinkgoWriter, "sourceDv %v", sourceDv)
```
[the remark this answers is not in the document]
brybacki: garbage, removed

tests/cloner_test.go
```diff
@@ -508,6 +515,40 @@ var _ = Describe("all clone tests", func() {
 				err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
 				Expect(err).ToNot(HaveOccurred())
 			})
+
+			It("[rfe_id:1126][crit:High][vendor:cnv-qe@redhat.com][level:component] Should fail with Event when cloning into a smaller sized data volume", func() {
+				By("Creating a source from a real image") // todo fs....
```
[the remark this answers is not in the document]
brybacki: and not needed, removed

tests/cloner_test.go
brybacki: removed

tests/cloner_test.go
brybacki: that is why the next line 2 lines are:  
```
				if cloneType == "network" {
					f.ForceBindPvcIfDvIsWaitForFirstConsumer(targetDv)
				}
```
Because only on network clone the pvc is created so we can try to consume it if WFFC. 
If you think the comment is not clear I should try to change the comment or the code.

tests/cloner_test.go
brybacki: So let me enumerate tests added or changed:
1. A new test for the BUG: cloner_test.go:472
`It("Should clone data from fs to fs while using calculated storage size", func() {`
2. A new test for cloning to smaller data volume with the use of Storage.spec (as it is now bypassing the validation hook) cloner_test.go:519:
			It("[rfe_id:1126][crit:High][vendor:cnv-qe@redhat.com][level:component] Should fail with Event when cloning into a smaller sized data volume", func() {
3. Very small cleanup for the existing test  cloner_test.go:886:
`		It("[rfe_id:1126][test_id:1896][crit:High][vendor:cnv-qe@redhat.com][level:component] Should not allow cloning into a smaller sized data volume", func() {`

pkg/controller/datavolume-controller.go
brybacki: yes, and it would be good to move it together with the reverse function and I think it would be even better to create a filesystem.go file with constants and functions useful for filesystem storage. 

So I wanted to do it quick for the bug to have the bugfix faster. But I can try moving this function now If you think it can be qucik and painless.

brybacki: I think I can easily move it right now

awels (commented):
/lgtm
@ShellyKa13 are you okay with the answers to your comments?

tests/cloner_test.go
ShellyKa13: the function ForceBindPvcIfDvIsWaitForFirstConsumer wasnt clear to me that it also check that a pvc was created. that ok then

tests/cloner_test.go
ShellyKa13: I understand I was wandering about (2) if using the same size 200Mi will work? should it work? or the 200Mi is doesnt include fsoversize so it shouldnt work and if so it should be tested

ShellyKa13:
@brybacki one small question about the test if nothing to do with it then the PR is ok by me too

tests/cloner_test.go
brybacki: yes, it works on a PVC so it really needs a PVC

brybacki:
@ShellyKa13 replied to your questions

tests/cloner_test.go
awels: Yeah I am fine with that, that should be an interesting test but outside of the scope of this PR I think.

awels:
/approve
Please think about the extra tests for different fsOverhead on different storage classes.

pkg/controller/datavolume-controller.go
brybacki: done

PR #2102, opened 2022-01-13
Allow optional per-DataVolume VDDK image.

mrnold:
This pull request adds an optional initImageURL field to the VDDK data source spec, so that a VDDK archive image can be specified per-DataVolume. This allows MTV to migrate VMware VMs without interfering with the HCO-managed v2v-vmware ConfigMap.

Fixes [BZ#2031033](https://bugzilla.redhat.com/show_bug.cgi?id=2031033) partially, enables complete fix in MTV

**Release note**:
```release-note
Allow optional specification of per-DataVolume VDDK archive image.
```

### DIFF
--- a/pkg/controller/import-controller_test.go
+++ b/pkg/controller/import-controller_test.go
@@ -604,7 +604,7 @@ var _ = Describe("Update PVC from POD", func() {
 		err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "testPvc1", Namespace: "default"}, resPvc)
 		Expect(err).ToNot(HaveOccurred())
 		Expect(resPvc.GetAnnotations()[AnnBoundCondition]).To(Equal("false"))
-		Expect(resPvc.GetAnnotations()[AnnBoundConditionMessage]).To(Equal("waiting for v2v-vmware configmap for VDDK image"))
+		Expect(resPvc.GetAnnotations()[AnnBoundConditionMessage]).To(Equal(fmt.Sprintf("waiting for v2v-vmware configmap or %s annotation for VDDK image", AnnVddkInitImageURL)))
 		Expect(resPvc.GetAnnotations()[AnnBoundConditionReason]).To(Equal(common.AwaitingVDDK))
 
 		By("Checking again after creating configmap")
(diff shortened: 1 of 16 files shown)


mrnold:
> @mrnold please see [kubevirt/project-infra#1846](https://github.com/kubevirt/project-infra/pull/1846) which will fix the issue for cdi prow jobs, caused by [kubevirt/project-infra#1848](https://github.com/kubevirt/project-infra/pull/1848). Sorry for the inconvenience!

Thanks! I will try again when that's done.

pkg/controller/import-controller_test.go
```diff
@@ -638,6 +638,17 @@ var _ = Describe("Update PVC from POD", func() {
 		Expect(err).ToNot(HaveOccurred())
 	})
 
+	It("Should not mark PVC as waiting for VDDK configmap, if image URL annotation is present", func() {
+		pvc := createPvcInStorageClass("testPvc1", "default", &testStorageClass, map[string]string{AnnEndpoint: testEndPoint, AnnImportPod: "testpod", AnnSource: SourceVDDK, AnnVddkInitImageURL: "test://image"}, nil, corev1.ClaimBound)
+		reconciler = createImportReconciler(pvc)
+		err := reconciler.createImporterPod(pvc)
+		Expect(err).ToNot(HaveOccurred())
+		resPvc := &corev1.PersistentVolumeClaim{}
```
awels, line 646: Should we be looking for the pod, not the pvc here? We created the PVC above, then call the reconcile createImporterPod, but then check the PVC, should be the pod.

awels (commented):
Looks good one minor question on a test.

pkg/controller/import-controller_test.go
mrnold: In this case, createImporterPod is the thing that adds the AwaitingVDDK annotation to the PVC. The test doesn't really care about the pod, it was only created to test that code path through createImporterPod. The real test is to make sure the PVC was updated afterward, as a side effect of attempting to create the pod. So I think the test is basically correct, but maybe there's a better way to do this PVC update.

Now that you point out this test though, I notice that it doesn't actually check for the absence of the AwaitingVDDK condition reason. I will put that in.

PR #2103, opened 2022-01-13
Refactor the data-processor.go to allow customization

### LINKED ISSUE #2094: Make the `DataProcessor` extendable by adding new `ProcessingPhase`
Hi there,

We have the use case to add customized `ProcessingPhase` in `pkg/importer/data-processor.go`. Right now we just modify the source code but it's not good for maintenance. 

Is it possible to extract the `switch` logic inside `DataProcessor.ProcessDataWithPause()` to be a `map[ProcessingPhase]func()(ProcessingPhase, error)`. Then, the method `ProcessDataWithPause()` will simply lookup the map to process the current phase and provide next phase or error.

The map can be constructed at the object construction time. Also, a method to register new phases and logics can be added.

Is there any objections to this idea? I can send a PR if possible.

zhuchenwang:
Allows user to extend data processor to handle some specific image format or data source.

Fixes #2094

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/data-processor.go
+++ b/pkg/importer/data-processor.go
@@ -123,6 +123,8 @@ type DataProcessor struct {
 	preallocation bool
 	// preallocationApplied is used to pass information whether preallocation has been performed, or not
 	preallocationApplied bool
+	// phaseExecutors is a mapping from the given processing phase to its execution function. The function returns the next processing phase or error.
+	phaseExecutors map[ProcessingPhase]func() (ProcessingPhase, error)
 }
 
 // NewDataProcessor create a new instance of a data processor using the passed in data provider.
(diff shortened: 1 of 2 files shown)


awels:
Hi to fix the DCO issue just `git --amend -s`. The `-s` 'signs' the commit

zhuchenwang:
Looks like some tests are failing. However, it seems not related to my PR. Any suggestions?

awels:
Hi, don't worry too much about the failures right now. Yesterday a new builder images was installed, and something is not right at this point, and the CI will always fail. We are working on figuring out the problem.

pkg/importer/data-processor.go
```diff
@@ -183,66 +195,93 @@ func (dp *DataProcessor) ProcessDataResume() error {
...
-				err = errors.Wrap(err, "Unable to apply delta to base image")
-			}
-		default:
+		if visited[dp.currentPhase] {
+			err := errors.Errorf("loop detected on phase %s", dp.currentPhase)
+			klog.Errorf("%+v", err)
+			return err
+		}
+		executor, ok := dp.phaseExecutors[dp.currentPhase]
+		if !ok {
```
brybacki, line 272: This branch is not covered by test. It would be good to write a test for a missing/unknown phase, so it is protected from future refactorings.

brybacki:
looks good, I would like to see one more test, as I stated in the comment of pkg/importer/data-processor.go

pkg/importer/data-processor.go
zhuchenwang: Test case was added in the last commit.

Thank you for the review!

awels:
@brybacki you happy with the new unit test?

brybacki:
yes, now its good
/lgtm

PR #2111, opened 2022-01-18
Fix error message to be without trace

ShellyKa13:
Pass the function only the err without the wraping
back trace which is too long and causes the message
to overflow and not show the specific error.

Fixes: BZ#2000661

**Release note**:

```release-note
NONE
```


brybacki:
Interestingly the last test failure shows DV as complete, so there must be some kind of flakiness in the test. Will look if it happens again:

`Tests Suite: [vendor:cnv-qe@redhat.com][level:component]SmartClone tests [test_id:4987]Verify DataVolume Smart Cloning - volumeMode filesystem - Waits for source to be available expand_more`

PR #2117, opened 2022-01-19
Add DataImportCron ImageStream tag support

arnongilboa:

Fixes #

**Release note**:
```release-note
Add DataImportCron ImageStream tag support
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"reflect"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/containers/image/v5/docker/reference"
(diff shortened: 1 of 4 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -138,30 +139,61 @@ func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron
...
 }
 
-func (r *DataImportCronReconciler) getImageStream(ctx context.Context, imageStreamName, imageStreamNamespace string) (*imagev1.ImageStream, error) {
+func (r *DataImportCronReconciler) getImageStream(ctx context.Context, imageStreamName, imageStreamNamespace string) (*imagev1.ImageStream, string, error) {
 	if imageStreamName == "" || imageStreamNamespace == "" {
-		return nil, errors.Errorf("Missing imagestream name or namespace")
+		return nil, "", errors.Errorf("Missing imagestream name or namespace")
 	}
 	imageStream := &imagev1.ImageStream{}
+	isName, isTag, err := splitImageStreamName(imageStreamName)
```
awels, line 147: I think the naming is throwing me off here, isName, and isTag I keep thinking why is this not a bool for the 'isSomething' naming convention, but it is actually imageStreamName, and imageStreamTag. maybe not shorten it in this case to make it clearer? Or just name and tag, since we are in the getImageStream function.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -353,7 +354,8 @@ func createDataImportCronReconciler(objects ...runtime.Object) *DataImportCronRe
 
 func newDataImportCronWithImageStream(name string) *cdiv1.DataImportCron {
```
awels, line 355: Can we add a test that tests the permutations of imagestreams, and invalid ones?
imagestream
imagestream:
imagestream:tag
imagestream:tag:label

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -369,6 +371,10 @@ func newImageStream(name string) *imagev1.ImageStream {
 		Status: imagev1.ImageStreamStatus{
 			Tags: []imagev1.NamedTagEventList{
 				{
+					Tag: "nosuch",
```
awels, line 374: Maybe also test the negative case where we have a tag that is not in the status.tags? Also are we still testing the no tag case?

arnongilboa: added

pkg/controller/dataimportcron-controller_test.go
arnongilboa: added

pkg/controller/dataimportcron-controller.go

PR #2119, opened 2022-01-20
Catch changed rejection message

awels:

Every once in a while the rejection message by kubectl is changed and it causes failures in the test. This commit makes the check less strict so it looks for the message output by our webhook, not the extra kubernetes messsage as well.

The message it was failing on `The request is invalid: spec.PVC.resources.requests.size: PVC size is missing` so checking for just PVC size is missing.
Fixes #

**Release note**:

```release-note
NONE
```


tests/api_validation_test.go
```diff
@@ -305,7 +305,7 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
...
 			table.Entry("[test_id:1767]fail with missing PVC spec", "manifests/dvMissingPVCSpec.yaml", true, "Missing Data volume PVC", "missing required field \"pvc\" in io.kubevirt.cdi.v1beta1.DataVolume.spec", "invalid: spec.pvc: Required value"),
 			table.Entry("[test_id:3920]fail with missing PVC accessModes", "manifests/dvMissingPVCAccessModes.yaml", true, "spec.pvc.accessModes in body is required", "spec.pvc.accessModes: Required value", "Required value: at least 1 access mode is required"),
-			table.Entry("[test_id:1768]fail with missing resources spec", "manifests/dvMissingResourceSpec.yaml", true, "spec.pvc.resources in body is required", "spec.pvc.resources: Required value", "admission webhook \"datavolume-validate.cdi.kubevirt.io\" denied the request:  PVC size is missing"),
+			table.Entry("[test_id:1768]fail with missing resources spec", "manifests/dvMissingResourceSpec.yaml", true, "spec.pvc.resources in body is required", "spec.pvc.resources: Required value", "admission webhook \"datavolume-validate.cdi.kubevirt.io\" denied the request:  PVC size is missing", "request is invalid: spec.PVC.resources.requests.size: PVC size is missing"),
```
brybacki, line 308: maybe we should make it even less strict?
the most important part is that it fails and the message conatains `PVC size is missing`

brybacki:
/lgtm 
if you want to merge it quickly it can go ahead, but I think we can not match too strictly, then our tests are too dependent on implementation details

awels:
I have gone back and forth on if I should just check the message and not the whole thing. There is no particular rush, and I think I like the less strict check better.

brybacki:
The failure on ceph lane looks like some kind of flakiness when closing writer in test. All the actual test assertions passed, including last file permissions verification. I suppose either the timeout was reached, or something is flaky when closing the pipewriter and file.

```
Tests Suite: [rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:component]Upload tests should [test_id:5082]succeed given a valid token (form async) expand_less | 2m57s
/root/go/src/kubevirt.io/containerized-data-importer/tests/upload_test.go:193 Unexpected error:     <*errors.errorString \| 0xc000138150>: {    
     s: "io: read/write on closed pipe",
     }
 io: read/write on closed pipe occurred /root/go/src/kubevirt.io/containerized-data-importer/tests/upload_test.go:429
```

PR #2129, opened 2022-01-26
Fix CRDs deletion in operator deletion

arnongilboa:
Also check DataImportCron CRD has no DeletionTimestamp before adding a
finalizer

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -34,6 +34,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	v1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -120,7 +121,19 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
 }
 
 func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
-	AddFinalizer(dataImportCron, dataImportCronFinalizer)
+	log := r.log.WithName("initCron")
+	if !HasFinalizer(dataImportCron, dataImportCronFinalizer) {
+		crd := &extv1.CustomResourceDefinition{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "dataimportcrons.cdi.kubevirt.io"}, crd); err == nil {
+			log.Error(err, "Cannot get CRD")
+			return nil
+		}
+		if crd.DeletionTimestamp != nil {
+			log.Info("CRD has DeletionTimestamp")
+			return nil
+		}
+		AddFinalizer(dataImportCron, dataImportCronFinalizer)
+	}
 	if isURLSource(dataImportCron) && !r.cronJobExists(ctx, dataImportCron) {
 		cronJob, err := r.newCronJob(dataImportCron)
 		if err != nil {
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -121,16 +121,13 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
 }
 
 func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
-	log := r.log.WithName("initCron")
 	if !HasFinalizer(dataImportCron, dataImportCronFinalizer) {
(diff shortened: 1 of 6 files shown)


pkg/operator/controller/controller_test.go
```diff
@@ -1523,12 +1523,14 @@ func createArgs() *args {
 }
 
 func doReconcile(args *args) {
-	result, err := args.reconciler.Reconcile(context.TODO(), reconcileRequest(args.cdi.Name))
-	Expect(err).ToNot(HaveOccurred())
-	Expect(result.Requeue).To(BeFalse())
+	Eventually(func() bool {
+		result, err := args.reconciler.Reconcile(context.TODO(), reconcileRequest(args.cdi.Name))
+		return err == nil && !result.Requeue
+	}, 1*time.Minute, 2*time.Second).Should(BeTrue())
```
mhenriks, line 1529: What's going on here?  Unit tests should be deterministic.

pkg/operator/controller/callbacks.go
```diff
@@ -132,6 +107,28 @@ func reconcileDeleteControllerDeployment(args *callbacks.ReconcileCallbackArgs)
...
 
+func deleteCRDs(args *callbacks.ReconcileCallbackArgs) error {
+	crdNames := []string{
+		"dataimportcrons.cdi.kubevirt.io",
+		"datavolumes.cdi.kubevirt.io",
+		"objecttransfers.cdi.kubevirt.io",
+	}
+	crdsExist := false
+	for _, crdName := range crdNames {
+		crd := &extv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: crdName}}
```
mhenriks, line 118: Maybe changing this to the following to address unit test issues?

Get CRD if not exist continue
If CRD.deletionTimestamp == nil{
  Delete CRD
  Get CRD if not exist continue
}
crdsExist = true

pkg/operator/controller/controller_test.go
mhenriks: I think I get it

pkg/operator/controller/callbacks.go
arnongilboa: So the issue was caused by repeated Delete() of CRD which already had DeletionTimestamp? Late night games...
Fixed

pkg/operator/controller/controller_test.go
arnongilboa: Thanks. Reverted it.

pkg/operator/controller/callbacks.go
```diff
@@ -132,6 +107,44 @@ func reconcileDeleteControllerDeployment(args *callbacks.ReconcileCallbackArgs)
...
+			}
+			if err := args.Client.Get(context.TODO(), types.NamespacedName{Name: crdName}, crd); err != nil {
+				if errors.IsNotFound(err) {
+					continue
+				}
+				return err
+			}
+			args.Logger.Info("CRD is not deleted yet", "crdName", crdName)
+			crdsExist = true
+		}
```
mhenriks, line 140: you have to set `crdsExist = true` here is else block when `crd.DeletionTimestamp != nil`

arnongilboa: Fat fingers. Fixed.

pkg/controller/dataimportcron-controller.go
```diff
@@ -120,7 +121,19 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
 }
 
 func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
-	AddFinalizer(dataImportCron, dataImportCronFinalizer)
+	log := r.log.WithName("initCron")
+	if !HasFinalizer(dataImportCron, dataImportCronFinalizer) {
+		crd := &extv1.CustomResourceDefinition{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "dataimportcrons.cdi.kubevirt.io"}, crd); err == nil {
```
mhenriks, line 127: err == nil
Don’t think that’s correct

pkg/controller/dataimportcron-controller.go
```diff
@@ -120,7 +121,19 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
...
+			log.Error(err, "Cannot get CRD")
+			return nil
```
mhenriks, line 129: Return err?

pkg/controller/dataimportcron-controller.go
```diff
@@ -120,7 +121,19 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
...
+		}
+		if crd.DeletionTimestamp != nil {
+			log.Info("CRD has DeletionTimestamp")
+			return nil
```
mhenriks, line 133: Return an error instead maybe?

pkg/controller/dataimportcron-controller.go

mhenriks:
new func test may be flaky

arnongilboa:
> new func test may be flaky

@mhenriks it was not the new func test, but a missing `extv1.AddToScheme()` in the controller for the CRD. Added.

PR #2140, opened 2022-02-03
use leases AND configmaps for leader election

mhenriks:

Configmap leader election is deprecated.  But we have to continue using them for awhile before fully transitioning to just leases.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -145,7 +145,14 @@ func start(ctx context.Context, cfg *rest.Config) {
 		klog.Fatalf("Error building extClient: %s", err.Error())
 	}
 
-	mgr, err := manager.New(config.GetConfigOrDie(), manager.Options{})
+	opts := manager.Options{
+		LeaderElection:             true,
+		LeaderElectionNamespace:    namespace,
+		LeaderElectionID:           "cdi-controller-leader-election-helper",
+		LeaderElectionResourceLock: "leases",
+	}
+
+	mgr, err := manager.New(config.GetConfigOrDie(), opts)
 	if err != nil {
 		klog.Errorf("Unable to setup controller manager: %v", err)
 		os.Exit(1)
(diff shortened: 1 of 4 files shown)


cmd/cdi-controller/controller.go
```diff
@@ -145,7 +145,14 @@ func start(ctx context.Context, cfg *rest.Config) {
 		klog.Fatalf("Error building extClient: %s", err.Error())
 	}
 
-	mgr, err := manager.New(config.GetConfigOrDie(), manager.Options{})
+	opts := manager.Options{
+		LeaderElection:             true,
+		LeaderElectionNamespace:    namespace,
+		LeaderElectionID:           "cdi-controller-leader-election-helper",
```
awels, line 151: Maybe use the configMapName const from leaderelection.go? Or rename that something more fitting?

cmd/cdi-controller/controller.go
```diff
@@ -145,7 +145,14 @@ func start(ctx context.Context, cfg *rest.Config) {
...
+		LeaderElectionResourceLock: "leases",
```
awels, line 152: Should this ResourceLock be configmapsleases as well? We have a configmap for that right now.

mhenriks: no we still get the old configmap lock first

cmd/cdi-controller/controller.go
mhenriks: I would like to jut delete `leaderelection.go` once we cut over

cmd/cdi-controller/controller.go
awels: Nevermind we get the old configmap lock before calling start which contains this code.

awels (commented):
/lgtm
/approve

cmd/cdi-controller/controller.go
awels: Sure, maybe move the const in here then?

PR #1987, opened 2021-10-12
Simplify importer code

brybacki:
**What this PR does / why we need it**:

The importer `main` is becoming complex. The PR tries to simplify code.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

Complexity warnings:
https://sonarcloud.io/component_measures?id=kubevirt_containerized-data-importer&metric=sqale_index&selected=kubevirt_containerized-data-importer%3Acmd%2Fcdi-importer%2Fimporter.go&view=list

**Special notes for your reviewer**:
As the code was extracted and moved around it is best to view it commit by commit and not the whole PR diff at once. 

**Release note**:

```release-note
NONE
```

PR #2141, opened 2022-02-04
Simplify importer code, cleanup and extract function

brybacki:
During the first attempt to simplify importer (PR #1987) I had to drop one commit
because of some hard to find errors. This is a second attempt, and I have found
that there is an implicit communication through termination message file.

Importer writes the file and data processor appends some data there
after it is closed. This might need some future improvements.

Fixes #

WDYT? Is it better?

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -144,9 +144,7 @@ func handleImport(
 	klog.V(1).Infoln("begin import process")
 
 	ds := newDataSource(source, contentType, volumeMode)
-	defer func() {
-		ds.Close()
-	}()
+	defer ds.Close()
 
 	processor := newDataProcessor(contentType, volumeMode, ds, imageSize, filesystemOverhead, preallocation)
 	waitForReadyFile()
@@ -157,7 +155,7 @@ func handleImport(
 		if err == importer.ErrRequiresScratchSpace {
 			return common.ScratchSpaceNeededExitCode
 		}
-		err = util.WriteTerminationMessage(fmt.Sprintf("Unable to process data: %+v", err))
+		err = util.WriteTerminationMessage(fmt.Sprintf("Unable to process data: %+v", err.Error()))
 		if err != nil {
 			klog.Errorf("%+v", err)
 		}
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -81,8 +81,6 @@ func main() {
 	imageSize, _ := util.ParseEnvVar(common.ImporterImageSize, false)
 	filesystemOverhead, _ := strconv.ParseFloat(os.Getenv(common.FilesystemOverheadVar), 64)
 	preallocation, err := strconv.ParseBool(os.Getenv(common.Preallocation))
-	var preallocationApplied bool
-	var ds importer.DataSourceInterface
 
 	volumeMode := v1.PersistentVolumeBlock
 	if _, err := os.Stat(common.WriteBlockPath); os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


cmd/cdi-importer/importer.go
```diff
@@ -121,51 +106,90 @@ func main() {
...
+	volumeMode v1.PersistentVolumeMode,
+	imageSize string,
+	filesystemOverhead float64,
+	preallocation bool) int {
+	klog.V(1).Infoln("begin import process")
 
-		ds = newDataSource(source, contentType, volumeMode)
-		defer ds.Close()
+	ds := newDataSource(source, contentType, volumeMode)
+	defer func() {
```
ShellyKa13, line 147: Why did you put the close inside a func and not like before defer ds.Close()?

cmd/cdi-importer/importer.go
```diff
@@ -121,51 +106,90 @@ func main() {
...
+	processor := newDataProcessor(contentType, volumeMode, ds, imageSize, filesystemOverhead, preallocation)
+	waitForReadyFile()
+	err := processor.ProcessData()
 
+	if err != nil {
+		klog.Errorf("%+v", err)
+		if err == importer.ErrRequiresScratchSpace {
+			return common.ScratchSpaceNeededExitCode
+		}
+		err = util.WriteTerminationMessage(fmt.Sprintf("Unable to process data: %+v", err))
```
ShellyKa13, line 160: please don't remove the err.Error(), I added it it fixes an issue: https://github.com/kubevirt/containerized-data-importer/pull/2111

cmd/cdi-importer/importer.go
```diff
@@ -121,51 +106,90 @@ func main() {
...
-		touchDoneFile()
-		preallocationApplied = processor.PreallocationApplied()
+
+		return 1
 	}
+	touchDoneFile()
+	// due to the way some data sources can add additional information to termination message
+	// after finished (ds.close() ) termination message has to be written first, before the
+	// the ds is closed
+	// TODO: think about making communication explicit, probably DS interface should be extended
```
ShellyKa13, line 171: I dont completly understand if what you did in this commit fixes this issue or not, cause if so I dont see how this changes the behavior than the previous code

ShellyKa13 (commented):
I think its a good clean. some question and one important comment

cmd/cdi-importer/importer.go
brybacki: Only this part opens dataSource, so it should also close it. By doing it here I am sure that all returns from the handleImport are covered. 

Of course while doing that, I have discovered that util.WriteTerminationMessage MUST happen before ds.close().

brybacki: Oh, now I see, you are not asking why I did the whole thing. But why there is func wrapper?
Just a leftover from tests and debugging. Will remove now.

cmd/cdi-importer/importer.go
brybacki: Sorry, I did play with the format, temporary added some more verbose logging. so this was removed by mistake, reverting that part

cmd/cdi-importer/importer.go
brybacki: Let me start from beginning.

I wanted to make some small cleanup the main() so while reading the code one can see the decisions and branches (as it shows there are only two: emptyDisk, and import). 

Then I have exported methods (example: handleImport) and with that exported method I made all local variables really local. That resulted in problems. And that is when I have discovered that we have to first write Import Complete, and the we can close DataSource. 

I think this is designed in a bad way. And I may think about improving that. Some options:
1. we could add a ds.summary() or ds.terminationMessgae(). So then we just do 
```
ds.close() 
additionalInfo:= ds.terminationMessage()
err = importCompleteTerminationMessage(processor.PreallocationApplied(), additionalInfo)
```

2. made ds.close() return termination message?

Anyway, current code works exactly the same way as old. All I managed to do is extract 2 methods (handleempty and handleimport) and make all local variables really local (so this is basically last commit that I needed to drop on https://github.com/kubevirt/containerized-data-importer/pull/1987). And I have discovered that DS and importer are both trying to write to termnation message file. So this is why I have left the comment. So we can see it clearly and do not forget to improve.

cmd/cdi-importer/importer.go
brybacki: done

cmd/cdi-importer/importer.go
brybacki: done

cmd/cdi-importer/importer.go
ShellyKa13: Thanks for clarifying, getting the termination message after closing sounds good to me.

brybacki:
/retest

failed in aftersuite

PR #2143, opened 2022-02-07
Extend tests to check explicit storage class name

brybacki:

Storage class name is always set on the PVC when that PVC is created
by the CDI. Updated a few tests to check for this behavior. Added two new scenarios.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -1649,7 +1649,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 			}, 30*time.Second, time.Second).Should(BeTrue())
 		})
 
-		It("[test_id:5911]Import succeeds creating a PVC from DV without accessModes", func() {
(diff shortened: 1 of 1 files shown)


tests/datavolume_test.go
```diff
@@ -1681,6 +1681,73 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			By("Restore the profile")
 			updateStorageProfileSpec(f.CrClient, defaultScName, *originalProfileSpec)
 		})
+		It("[test_id:8170]Import succeeds when storage class is not specified, but access mode is", func() {
+			defaultScName := utils.DefaultStorageClass.GetName()
+
+			By(fmt.Sprintf("configure storage profile %s", defaultScName))
```
awels, line 1687: I would put the piece of code that saves the storage profile in a BeforeEach, and the piece that restores it in an AfterEach

brybacki: done

brybacki:
As the imageio was fixed, the test should now pass.

PR #2146, opened 2022-02-09
Fix interaction between multi-stage import and retainAfterCompletion.

mrnold:
This pull request appends a checkpoint ID to the importer name, for multi-stage imports. This way, when the retainAfterCompletion annotation is set, completed pods will not prevent further import stages because they will no longer have the same name.

Fixes [BZ#2016290](https://bugzilla.redhat.com/show_bug.cgi?id=2016290)

**Release note**:
```release-note
Append checkpoint ID to multi-stage importer pod names to fix interaction with retainAfterCompletion annotation.
```


pkg/controller/import-controller_test.go
```diff
@@ -685,6 +685,38 @@ var _ = Describe("Update PVC from POD", func() {
 		Expect(resPvc.GetAnnotations()[AnnVddkVersion]).To(Equal("1.0.0"))
 	})
 
+	It("Should delete pod for scratch space even if retainAfterCompletion is set", func() {
```
awels, line 688: I don't see the retainAfterCompletion annotation set on the PVC

pkg/controller/import-controller_test.go
```diff
@@ -751,6 +783,29 @@ var _ = Describe("Create Importer Pod", func() {
 		table.Entry("should create pod with file system volume mode and scratchspace", createPvc("testPvc1", "default", map[string]string{AnnEndpoint: testEndPoint, AnnPodPhase: string(corev1.PodPending), AnnImportPod: "podName", AnnPriorityClassName: "p0"}, nil), &scratchPvcName),
 		table.Entry("should create pod with block volume mode and scratchspace", createBlockPvc("testBlockPvc1", "default", map[string]string{AnnEndpoint: testEndPoint, AnnPodPhase: string(corev1.PodPending), AnnImportPod: "podName", AnnPriorityClassName: "p0"}, nil), &scratchPvcName),
 	)
+
+	It("should append current checkpoint name to importer pod", func() {
```
awels, line 787: Can you add some tests where the pvc name is long, and the snapshot name is long as well. This should trigger the hashing of the name of the pod.

awels (commented):
looks good, a few questions on some edge case tests

pkg/controller/import-controller_test.go
mrnold: I have added this to the annotations list and made it easier to read.

pkg/controller/import-controller_test.go
mrnold: I reworked this into a table and added tests for long and short PVC names and checkpoint names.

PR #2148, opened 2022-02-09
Fix failing imageio test

awels:

Always enable scratch space for imageio imports. Before imageio was doing the create pod without scratch space, then detect we need it, delete pod, and start new pod with scratch space cycle for every delta in a warm migration. This PR just takes out the cycle and immediately requires scratch space.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -526,7 +526,9 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 		priorityClassName: getPriorityClass(pvc),
 	}
 
-	// Check for old terminating scratch spaces
+	// Check for old terminating scratch spaces, this can happen in warm migration scenarios where the scratch
+	// space for a previous pod has not been deleted yet. In that case check if the scratch PVC is terminating
+	// before creating the new importer pod. This should give the PVC time to get deleted.
 	if requiresScratch && podArgs.scratchPvcName != nil {
 		r.log.V(1).Info("Checking for old scratch pvcs")
 		scratchPvc := &corev1.PersistentVolumeClaim{}
@@ -741,7 +743,9 @@ func (r *ImportReconciler) requiresScratchSpace(pvc *corev1.PersistentVolumeClai
 		case SourceGlance:
 			scratchRequired = true
 		case SourceImageio:
-			scratchRequired = true
+			if val, ok := pvc.Annotations[AnnCurrentCheckpoint]; ok {
+				scratchRequired = val != ""
+			}
 		case SourceRegistry:
 			scratchRequired = pvc.Annotations[AnnRegistryImportMethod] != string(cdiv1.RegistryPullNode)
 		}
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -525,6 +525,21 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 		vddkImageName:     vddkImageName,
 		priorityClassName: getPriorityClass(pvc),
 	}
+
+	// Check for old terminating scratch spaces
+	if requiresScratch && podArgs.scratchPvcName != nil {
+		r.log.V(1).Info("Checking for old scratch pvcs")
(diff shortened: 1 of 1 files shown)


pkg/controller/import-controller.go
```diff
@@ -525,6 +525,21 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
...
+	if requiresScratch && podArgs.scratchPvcName != nil {
+		r.log.V(1).Info("Checking for old scratch pvcs")
+		scratchPvc := &corev1.PersistentVolumeClaim{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: pvc.GetNamespace(), Name: *podArgs.scratchPvcName}, scratchPvc)
+		if IgnoreNotFound(err) != nil {
+			return err
+		}
+		if scratchPvc.DeletionTimestamp != nil {
+			// Old scratch space with same name is still being cleaned up, return error.
+			return errors.New("Waiting for scratch space to be deleted.")
```
arnongilboa, line 539: lint says you should remove the "." at the end.

pkg/controller/import-controller.go
```diff
@@ -525,6 +525,21 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
...
+
+	// Check for old terminating scratch spaces
+	if requiresScratch && podArgs.scratchPvcName != nil {
+		r.log.V(1).Info("Checking for old scratch pvcs")
+		scratchPvc := &corev1.PersistentVolumeClaim{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: pvc.GetNamespace(), Name: *podArgs.scratchPvcName}, scratchPvc)
+		if IgnoreNotFound(err) != nil {
+			return err
+		}
+		if scratchPvc.DeletionTimestamp != nil {
```
ShellyKa13, line 537: Just making sure if the scratchPvc was found but not deleting its ok? a new one will not be created?

ShellyKa13: I looked it up and it seems ok

pkg/controller/import-controller.go
```diff
@@ -525,6 +525,21 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
 		vddkImageName:     vddkImageName,
 		priorityClassName: getPriorityClass(pvc),
 	}
+
```
ShellyKa13, line 528: maybe add an explanation in the commit message why scratch space is now always needed?

brybacki: Yes, a short description of WHY. was it always needed and it worked by chance, or is it something new?

pkg/controller/import-controller.go
awels: So with warm import it is currently possible that the scratch space for a previous pod is not gone yet before the next pod is started and if that next pod uses scratch space as well (with this PR it is always the case) then the scratchSpace PVC will never get deleted (in use by new pod) but the pod will never get started (terminating PVC). This code prevents the creation of the new pod until the scratch space PVC is gone.

pkg/controller/import-controller.go
awels: Done

pkg/controller/import-controller.go
awels: So the proxy PR noticed an unintended interaction between retainPod annotation and scratch space. Basically if you had DV that included the retainPod annotation and got into the create pod, detect you need scratch space, then delete pod, and create new pod with scratch space cycle, it would break on the delete pod part. Since we said retain pod. I fixed that interaction and as part of that fix, after we create the new pod with scratch space I would flip the annotation on the PVC that indicates we need scratch space to false (since it finished its intended purpose). Except in warm migrations where we can have multiple import pods for one DV/PVC. The next pod would see the annotation and not updated it so the scratch space would never get enable and the import would fail.

Since the imageio imports always need scratch space, this PR just enables it by default on imageio DVs

aglitke: Maybe even some explanatory comments in the code?  Also for the question @ShellyKa13 had about the scratch space deletion timestamp.

PR #2149, opened 2022-02-10
Adjust alert summary that is visible in UI to be more informative

akalenyu:

As of now OpenShift UI will not display the runbook URL field for our alerts
(docs and google will though), so let's make sure we provide a better entry point
by being verbose in what is actually visible to the user.

This is how things look like ATM:
image
Runbook index for reference (google will redirect people to this):
http://kubevirt.io/monitoring/runbooks/

Fixes #

**Release note**:

```release-note
NONE
```


pkg/operator/controller/prometheus.go
```diff
@@ -239,7 +239,7 @@ func getAlertRules() []promv1.Rule {
 			"kubevirt_cdi_dataimportcron_outdated_total > 0",
 			"15m",
 			map[string]string{
-				"summary":     "DataImportCron latest imports are outdated",
+				"summary":     "DataImportCron, which is in charge of recurring polling of latest version disk images as PVCs (also known as golden images), is failing to keep images up-to-date",
```
awels, line 242: Maybe something like `DataImportCron, VM templates disk image sources are not being updated on its defined schedule`
I am having a hard time formulating a good message. I don't like the one we have here.

akalenyu: I changed it to a mixture of both the original and your suggestion, let me know if that sounds better

awels:
Much better
/lgtm
/approve

PR #2150, opened 2022-02-10
Ignore error in upload copy test

awels:

Ignore the error, it might be because the other end closed the connection because it was done. We verify the image uploaded
by md5 anyway.

Fixes #

**Release note**:

```release-note
NONE
```


brybacki:
So I suppose that we do not care what kind of error that is (like connection closed or something like that), because it is just a test , and as your comment states everything is verified at the end?

/lgtm

awels:
The idea is that io.Copy can still copy bytes even if there is an error. Like one side closing the connection because it is done sending or receiving. And that error is harmless.

awels:
Seems like maybe I haven't completely solved the terminating scratch space issue with warm imports.

PR #2166, opened 2022-02-21
Use NFS CSI instead of static no provisioner manual PVs

akalenyu:

We are creating (and deleting) NFS PVs on our own before each test,
which seems to trigger a race in the deletion, causing flakes.

There is an NFS CSI driver available that we could use instead for dynamic NFS PV provisioning:
https://github.com/kubernetes-csi/csi-driver-nfs

Fixes #

Builds upon:
https://github.com/kubevirt/project-infra/pull/1943
https://github.com/kubevirt/kubevirtci/pull/753

**Release note**:

```release-note
NONE
```


akalenyu:
/hold

still figuring out if IsStaticNfs is good for all cases, especially the cloner test which will now get skipped

akalenyu:
/hold cancel
/retest

can be reviewed. `6655` is the only test we skip with this going in.

akalenyu:
/hold

nfs csi setup will go to kubevirt/kubevirtci as we need it on both kubevirt & cdi

brybacki:
> /hold cancel /retest
> 
> can be reviewed. `6655` is the only test we skip with this going in.

and why 6655 is skipped?

akalenyu:
> > /hold cancel /retest
> > can be reviewed. `6655` is the only test we skip with this going in.
> 
> and why 6655 is skipped?

Because we now have a CSI driver in place
https://github.com/kubevirt/containerized-data-importer/blob/bf9c51f715280c95990f8ce60e16dd2a226b0a11/tests/csiclone_test.go#L78-L83

brybacki:
> > > /hold cancel /retest
> > > can be reviewed. `6655` is the only test we skip with this going in.
> > 
> > 
> > and why 6655 is skipped?
> 
> Because we now have a CSI driver in place
> 
> https://github.com/kubevirt/containerized-data-importer/blob/bf9c51f715280c95990f8ce60e16dd2a226b0a11/tests/csiclone_test.go#L78-L83

ok, so new HPP can support CSI Clone, and CSI NFS too? We need to evaluate if we want to:
1. have some kind of non CSI lane to test this, 
2. try to convert it to a unit test that check just the part of the logic that selects correct strategy (or maybe there is one already)
3. drop the test

akalenyu:
> > > > /hold cancel /retest
> > > > can be reviewed. `6655` is the only test we skip with this going in.
> > > 
> > > 
> > > and why 6655 is skipped?
> > 
> > 
> > Because we now have a CSI driver in place
> > https://github.com/kubevirt/containerized-data-importer/blob/bf9c51f715280c95990f8ce60e16dd2a226b0a11/tests/csiclone_test.go#L78-L83
> 
> ok, so new HPP can support CSI Clone, and CSI NFS too? We need to evaluate if we want to:
> 
> 1. have some kind of non CSI lane to test this,
> 2. try to convert it to a unit test that check just the part of the logic that selects correct strategy (or maybe there is one already)
> 3. drop the test

Actually no, they don't support clones, but they do of course have a CSI driver each.
We still have a regular NFS OpenShift lane btw

brybacki:
> 
right, our check in test is not too smart (I wrote this ;)), but it would be complex to check if it supports CSI clone. But maybe I can think about this.

akalenyu:
> > 
> 
> right, our check in test is not too smart (I wrote this ;)), but it would be complex to check if it supports CSI clone. But maybe I can think about this.

There are the CSI capabilities, but I am not seeing where one could query those
https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/pkg/nfs/nfs.go#L86-L89

PR #2156, opened 2022-02-11
Create imageio container during CDI build.

awels:
Signed-off-by: Alexander Wels <awels@redhat.com>

**What this PR does / why we need it**:
Instead of using a really old imageio, use bazel to build a new imageio based on 2.5.0. Update the tests to use the new image
and paths in that new image. This requires a new repo in quay for us to push the image to.

Also changed the approach of resolving the warm import potential dead lock (scratch PVC from previous import pod terminating, while the new pod is trying to create itself). Instead of trying to avoid in all scenarios, detect the state, and delete the pod so the dead lock can be resolved.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2145 

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #2170, opened 2022-02-25
Use aarch64 RPMs for aarch64, add lint check for this error

maya-r:
In #2156 we added a 404 URL for aarch64 RPMs.
[This initially caused a post-submit job to fail](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-containerized-data-importer-push-nightly-ARM64/1496319407448657920).

But this error was later concealed by #2164.
We cache files on Google Storage, and the files are fetched by checksum.
Since the checksum was specified the same as the x86_64 files, we started fetching the x86_64 RPMs and not fail to build any more.

This is somewhat hard to spot without the intermediate post-submit job failure, so let's add a test for it, running during `make test-lint`.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ test-functional: build-functest
 test-lint:
 	${DO_BAZ} "./hack/build/run-lint-checks.sh"
 	"./hack/ci/language.sh"
+	"./hack/ci/duplicate-sha256.sh"
 
 docker-registry-cleanup:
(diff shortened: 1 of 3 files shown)


maya-r:
/hold
Would like to get CentOS Stream9 PR in first to avoid conflicting with myself.

maya-r:
included check script in #1983, no need for this PR after it

PR #2174, opened 2022-03-02
Fix running blockdev command when DV is Filesystem volumeMode

nijinashok:

As of now, we are passing "/data/disk.img" to GetAvailableSpaceBlock in Filesystem volumeMode and because of that, we always run blockdev command even in Filesystem volumeMode. With this change, the blockdev command will be only executed if it's a device file.

Fixes #

**Release note**:

```
Fix running blockdev command when DV is Filesystem volumeMode
```


pkg/util/util.go
```diff
@@ -121,15 +121,15 @@ func GetAvailableSpace(path string) (int64, error) {
...
 func GetAvailableSpaceBlock(deviceName string) (int64, error) {
-	// Check if device exists.
-	info, err := os.Stat(deviceName)
+	// Check if the file exists and is a device file.
+	info, err := os.Lstat(deviceName)
 	if os.IsNotExist(err) {
 		return int64(-1), nil
 	}
-	if info.IsDir() {
+	if (info.Mode() & os.ModeDevice) == 0 {
```
brybacki, line 129: A small helper `isDevice(info.Mode()) ` would make it more readable

brybacki: @nijinashok wdyt?

nijinashok: Ack, thanks for the review. Added `isDevice` helper.

pkg/util/util.go
```diff
@@ -121,15 +121,15 @@ func GetAvailableSpace(path string) (int64, error) {
 
 // GetAvailableSpaceBlock gets the amount of available space at the block device path specified.
 func GetAvailableSpaceBlock(deviceName string) (int64, error) {
-	// Check if device exists.
-	info, err := os.Stat(deviceName)
+	// Check if the file exists and is a device file.
+	info, err := os.Lstat(deviceName)
```
awels, line 125: So why change to lstat here? According to the documentation lstat doesn't follow symlinks while stat does. If the file we are looking at is a symlink I think we want to follow it.

awels (commented):
One small question looks good otherwise.

pkg/util/util.go
nijinashok: yes, lstat change is not needed. Reverted.

PR #2176, opened 2022-03-03
check uploadserver using certs in secrets

mhenriks:

Functional test to confirm that certs are referenced in secrets

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -137,11 +137,34 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		Expect(err).ToNot(HaveOccurred())
 	}
 
+	checkUploadCertSecrets := func() {
+		pod, err := f.K8sClient.CoreV1().Pods(pvc.Namespace).Get(context.TODO(), utils.UploadPodName(pvc), metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


tests/upload_test.go
```diff
@@ -137,12 +137,35 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+					ev.ValueFrom != nil &&
+					ev.ValueFrom.SecretKeyRef != nil &&
+					ev.ValueFrom.SecretKeyRef.Name == secret.Name {
+					found = true
+				}
+			}
+			Expect(found).To(BeTrue())
+		}
+	}
+
```
brybacki, line 160:
```suggestion

func HasEnvironmentVariableFromSecret(pod *v1.Pod, name string, secret *v1.Secret) bool {
	for _, ev := range pod.Spec.Containers[0].Env {
		if ev.Name == name &&
			ev.ValueFrom != nil &&
			ev.ValueFrom.SecretKeyRef != nil &&
			ev.ValueFrom.SecretKeyRef.Name == secret.Name {
			return true
		}
	}
	return false
}

```

brybacki (commented):
It is all good, but what do you think about extracting helper and using it twice in `Expect().To()` ?
Isn't it nicer? It is probably a matter of style, but in my opinion main function body is now more readable.

PR #2177, opened 2022-03-04
Bugfix for ceph rbd archive upload

brybacki:

When using contentType: archive during unpacking archives onto storage we always require a Filesystem mode PV since a tar file can only be extracted onto a filesystem. 

For ceph-rbd, RWX/Filesystem is not supported so if we choose to select a Filesystem mode with this storage class it must use RWO. This change adds missing configuration for StorageClass capabilities so the correct option can be selected. 
 
Fixes #

https://bugzilla.redhat.com/show_bug.cgi?id=2054778

**Release note**:

```release-note
BugFix: bz 2054778 -  PVC created with filesystem volume mode in some cases, instead of block volume mode - fails
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -31,7 +31,9 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock},
 		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeFilesystem}},
 	"openshift-storage.rbd.csi.ceph.com": {
-		{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeBlock}},
+		{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeBlock},
+		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock},
+		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeFilesystem}},
 	// ceph-fs
 	"cephfs.csi.ceph.com":                   {{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeFilesystem}},
 	"openshift-storage.cephfs.csi.ceph.com": {{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeFilesystem}},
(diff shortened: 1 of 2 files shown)


mhenriks:
functional test?

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -31,7 +31,9 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock},
 		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeFilesystem}},
 	"openshift-storage.rbd.csi.ceph.com": {
-		{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeBlock}},
+		{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeBlock},
+		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock},
+		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeFilesystem}},
```
arnongilboa, line 36: Caps for all 4 ceph-rbd keys are the same, so to make it more readable amd maintainable, consider using an initializer func, e.g.
```
func getCephRbdCaps() []StorageCapabilities {
	return []StorageCapabilities{
		{AccessMode: v1.ReadWriteMany, VolumeMode: v1.PersistentVolumeBlock},
		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock},
		{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeFilesystem},
	}
}
```
and similarly: `getRWOFilesystemCaps, getRWXFilesystemCaps, getRWOBlockCaps`

brybacki: I would like to get another vote on that. In my opinion it is clearer the way it is. We have an explicit table mapping the key (based on provider string) to capabilities. Almost looks like a spreadsheet.

arnongilboa: Spreadsheets were invented mostly for using functions instead of c&p'ing data across the cells.

brybacki:
> functional test?

I have added a unit test [1] for the dv reconcile, with given profile configuration and contentType = DataVolumeArchive. This covers that scenario. The bug is that the PVC created by cdi had accessMode = RWX, because there were no RWO/FS combination in the profile capabilities.

The fix basically extends the configuration for StorageProfile (well known capabilities) I am not sure we need a functional test for each combination. 

[1] https://github.com/kubevirt/containerized-data-importer/blob/c7c63475ca83c446242da038a90484f49bf1ac24/pkg/controller/datavolume-controller_test.go#L291

PR #2193, opened 2022-03-19
Do not factor fs overhead into available space during validation

aglitke:
Do not factor fs overhead into available space during validation
    
    Fixes bz#2064936.
    
    When validating whether an image will fit into a PV we compare the
    image's virtual size to the filesystem's reported available space to
    guage whether it will fit.  The current calculation reduces the apparent
    available space by the configured filesystem overhead value but the
    overhead is already (mostly) factored into the result of Statfs.  This
    causes the check to fail for PVCs that are just large enough to
    accommodate an image plus overhead (ie. when using the DataVolume
    Storage API with filesystem PVs with capacity constrained by the PVC
    storage request size).
    
    This was not caught in testing because HPP does not have capacity
    constrained PVs and we are typically testing block volumes in the ceph
    lanes.  It can be triggered in our CI by allocating a Filesystem PV on
    ceph-rbd storage because these volumes are capacity constrained and
    subject to filesystem overhead.
    
```release-note
NONE
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -62,7 +62,7 @@ type QEMUOperations interface {
 	ConvertToRawStream(*url.URL, string, bool) error
 	Resize(string, resource.Quantity, bool) error
 	Info(url *url.URL) (*ImgInfo, error)
-	Validate(*url.URL, int64, float64) error
+	Validate(*url.URL, int64) error
 	CreateBlankImage(string, resource.Quantity, bool) error
 	Rebase(backingFile string, delta string) error
 	Commit(image string) error
@@ -215,7 +215,7 @@ func isSupportedFormat(value string) bool {
 	}
 }
 
-func checkIfURLIsValid(info *ImgInfo, availableSize int64, filesystemOverhead float64, image string) error {
+func checkIfURLIsValid(info *ImgInfo, availableSize int64, image string) error {
 	if !isSupportedFormat(info.Format) {
 		return errors.Errorf("Invalid format %s for image %s", info.Format, image)
 	}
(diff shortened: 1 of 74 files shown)


akalenyu:
A little intimidated at first by the amount of changed files but its mostly line breaks/formatting so that should be fine.

My concern about this approach is that the overhead configurable in CDIConfig will no longer be used
(or am I getting it wrong?), are we ok with that?
It does kind of make sense that we took matters to our own hands, making overhead a configurable,
so we might want to consider disregarding the fixed overhead that is represented in `Bavail`, and instead go with `Bfree` in 
https://github.com/kubevirt/containerized-data-importer/blob/654c4cdd11101ad158129c468873581569debdad/pkg/util/util.go#L119

(Thanks mrashish for explaining the statfs bits)

brybacki:
first of all, all the whitespace changes, formatting and so on have to go into another PR. It will be very hard to backport a PR that changes more files. Also it is not good to combine a fix and improvement into one change as it is hard to review. 

I am just starting to review it.

pkg/image/qemu.go
```diff
@@ -226,18 +226,18 @@ func checkIfURLIsValid(info *ImgInfo, availableSize int64, filesystemOverhead fl
 		}
 	}
 
-	if int64(float64(availableSize)*(1-filesystemOverhead)) < info.VirtualSize {
-		return errors.Errorf("Virtual image size %d is larger than available size %d (PVC size %d, reserved overhead %f%%). A larger PVC is required.", info.VirtualSize, int64((1-filesystemOverhead)*float64(availableSize)), info.VirtualSize, filesystemOverhead)
+	if availableSize < info.VirtualSize {
+		return errors.Errorf("Virtual image size %d is larger than the reported available storage %d (PVC size %d). A larger PVC is required.", info.VirtualSize, availableSize, info.VirtualSize)
```
brybacki, line 230: I am not sure it is PVC size that we report here. Isn't it a disk or an image size?

brybacki:
The resulting test failures need some deeper investigation.

brybacki:
looking into the tests now

pkg/image/qemu.go
akalenyu: Yeah, PVC size here is definitely the image's virtual size and can confuse,
we do however print the calculated PVC size in https://github.com/kubevirt/containerized-data-importer/blob/654c4cdd11101ad158129c468873581569debdad/pkg/importer/data-processor.go#L404

brybacki:
cleanup here:
https://github.com/kubevirt/containerized-data-importer/pull/2195

brybacki:
> A little intimidated at first by the amount of changed files but its mostly line breaks/formatting so that should be fine.
> 
> My concern about this approach is that the overhead configurable in CDIConfig will no longer be used (or am I getting it wrong?), are we ok with that? It does kind of make sense that we took matters to our own hands, making overhead a configurable, so we might want to consider disregarding the fixed overhead that is represented in `Bavail`, and instead go with `Bfree` in
> 
> https://github.com/kubevirt/containerized-data-importer/blob/654c4cdd11101ad158129c468873581569debdad/pkg/util/util.go#L119
> 
> (Thanks mrashish for explaining the statfs bits)

We only care about the fact that image fits into the target, and that is why we check avail and do not need to verify overhead here. fsOverhead should only be used to be sure that the volume is big enough to fit image and filesystem metadata (with the idea that the metadata size is the overhead "over data size").

Are there any scenarios we might be missing here?

maya-r:
/close
closing in favour of #2195 which doesn't have unnecessary whitespace changes.

PR #2195, opened 2022-03-21
Do not factor fs overhead into available space during validation

brybacki:

This is the continuation of https://github.com/kubevirt/containerized-data-importer/pull/2193

When validating whether an image will fit into a PV we compare the
image's virtual size to the filesystem's reported available space to
guage whether it will fit.  The current calculation reduces the apparent
available space by the configured filesystem overhead value but the
overhead is already (mostly) factored into the result of Statfs.  This
causes the check to fail for PVCs that are just large enough to
accommodate an image plus overhead (ie. when using the DataVolume
Storage API with filesystem PVs with capacity constrained by the PVC
storage request size).

This was not caught in testing because HPP does not have capacity
constrained PVs and we are typically testing block volumes in the ceph
lanes.  It can be triggered in our CI by allocating a Filesystem PV on
ceph-rbd storage because these volumes are capacity constrained and
subject to filesystem overhead.

Fixes #

bz#2064936.

This is the continuation of https://github.com/kubevirt/containerized-data-importer/pull/2193

Trying to cleanup, remove unwanted changes, verify tests. 

**Release note**:

```release-note
Do not factor fs overhead into available space during validation
```


brybacki:
other tests still need investigation

brybacki:
test 2329 is also about different warning/error message

tests/upload_test.go
```diff
@@ -891,6 +864,71 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+			}
+		}
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Get an upload token")
+		token, err := utils.RequestUploadToken(f.CdiClient, pvc)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(token).ToNot(BeEmpty())
+
+		By("Create empty image with specific size")
```
awels, line 906: You can remove this including the defer.

tests/upload_test.go
```diff
@@ -891,6 +864,71 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+		cmd := exec.Command("qemu-img", args...)
+		cout, err := cmd.CombinedOutput()
+		if err != nil {
+			Fail(fmt.Sprintf("Failed to create test qcow2 image: %s", cout))
+		}
+		//defer os.Remove(uploadFile)
+
+		By("Do upload")
+		Eventually(func() error {
+			return uploadFileNameToPath(binaryRequestFunc, utils.UploadFile, uploadProxyURL, syncUploadPath, token, http.StatusOK)
```
awels, line 917: I think utils.UploadFile should be FsOverheadFile instead no?

brybacki: fixed

tests/upload_test.go
brybacki: yes

PR #2196, opened 2022-03-21
Reconcile DVs waiting for default SC

arnongilboa:

DataVolume controller will not create PVCs without a default storage class, and we'll get error and event:
`DataVolume.storage spec is missing accessMode and no storageClass to choose profile`
However, even after defining default SC, so far DV was not reconciled so it was stuck in this state.

Fixes #

**Release note**:
```release-note
Automatically recover DV from unconfigured default storage class, once default is set
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -340,6 +340,26 @@ func addDatavolumeControllerWatches(mgr manager.Manager, datavolumeController co
 		}
 	}
 
+	// Watch for SC updates and reconcile the DVs waiting for default SC
+	// FIXME: consider labelling DVs with waitingForDefaultSC instead of checking for empty phase
+	if err := datavolumeController.Watch(&source.Kind{Type: &storagev1.StorageClass{}}, handler.EnqueueRequestsFromMapFunc(
+		func(obj client.Object) (reqs []reconcile.Request) {
+			dvList := &cdiv1.DataVolumeList{}
+			if err := mgr.GetClient().List(context.TODO(), dvList, &client.ListOptions{}); err != nil {
+				return
+			}
+			for _, dv := range dvList.Items {
+				if dv.Status.Phase == "" {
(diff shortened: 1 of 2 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -340,6 +340,26 @@ func addDatavolumeControllerWatches(mgr manager.Manager, datavolumeController co
 		}
 	}
 
+	// Watch for SC updates and reconcile the DVs waiting for default SC
```
awels, line 343: Can we use an indexer for this? The idea is very similar to this unit test: https://github.com/kubernetes-sigs/controller-runtime/blob/7e7bf8c1e1df991b957dfaa8b36d975740f309ab/pkg/cache/cache_test.go#L1182-L1210
We get the indexer for the client from mgr, define the indexing function, and then use the index when listing. Should look something like this:
```go
indexer := mgr.GetFieldIndexer()
indexFunc := func(obj client.Object) []string {
  return []string{string(obj.(*cdiv1.DataVolume).Status.Phase)}
}
indexer.IndexField(context.TODO(), &cdiv1.DataVolume{}, "status.phase", indexFunc)

if err := datavolumeController.Watch(&source.Kind{Type: &storagev1.StorageClass{}}, handler.EnqueueRequestsFromMapFunc(
  func(obj client.Object) (reqs []reconcile.Request) {
    dvList := &cdiv1.DataVolumeList{}
    if err := mgr.GetCache().List(context.TODO(), dvList, client.MatchingFields{"status.phase": ""}); err != nil {
      return
    }
    for _, dv := range dvList.Items {
      reqs = append(reqs, reconcile.Request{NamespacedName: types.NamespacedName{Name: dv.Name, Namespace: dv.Namespace}})
    }
    return
  },
),
); err != nil {
  return err
}
```
Might have to configure the indexer before you start the mgr not 100% sure.

PR #2202, opened 2022-03-22
Add unit test to make sure we don't clean up dataimportcrons accidentally

akalenyu:

Saw a flaking test where events were saying that importer can't schedule
because cron PVCs are being deleted:
```bash
            "involvedObject": {
                "kind": "Pod",
                "namespace": "cdi-e2e-tests-importer-x9wfd",
                "name": "importer-datasource-test-90e064fca2f4",
                "uid": "be238296-7adc-41dd-b465-fac45a6a13ab",
                "apiVersion": "v1",
                "resourceVersion": "1609529"
            },
            "reason": "FailedScheduling",
            "message": "0/6 nodes are available: 6 persistentvolumeclaim \"datasource-test-90e064fca2f4\" is being deleted.",
```
But it looks like we are never going to clean up accidentally,
let's test it and keep looking for the actual reason for the failure.
Pretty cool that we'll have a way to intercept the calls in unit test with controller-runtime.

Fixes #

Grabbed the motivation for intercepting fake ctrl runtime from https://github.com/kubernetes/test-infra/blob/12a936787a32c4028dbf0849dd80c3692e50fea3/prow/deck/jobs/jobs_test.go#L510

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/controller/BUILD.bazel
+++ b/pkg/controller/BUILD.bazel
@@ -112,6 +112,7 @@ go_test(
         "//vendor/github.com/openshift/api/config/v1:go_default_library",
         "//vendor/github.com/openshift/api/image/v1:go_default_library",
         "//vendor/github.com/openshift/api/route/v1:go_default_library",
+        "//vendor/github.com/pkg/errors:go_default_library",
         "//vendor/k8s.io/api/batch/v1:go_default_library",
         "//vendor/k8s.io/api/core/v1:go_default_library",
         "//vendor/k8s.io/api/networking/v1:go_default_library",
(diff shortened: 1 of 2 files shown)


brybacki:
I think that in functional tests you can see this message:
`            "message": "0/6 nodes are available: 6 persistentvolumeclaim \"datasource-test-90e064fca2f4\" is being deleted.",`
when the timeout has passed and the test cleanup phase is deleting resources.

akalenyu:
> So I am not understand what we are testing for here. You added a unit test to test something that seems flaky in a functional test?

This is just a unit test that makes sure we won't ever start cleaning up crons when the GET fails on an arbitrary reason.
As it is we do not have this problem, but the coverage shouldn't hurt.

> I think that in functional tests you can see this message: ` "message": "0/6 nodes are available: 6 persistentvolumeclaim \"datasource-test-90e064fca2f4\" is being deleted.",` when the timeout has passed and the test cleanup phase is deleting resources.

possible, what was weird in those logs was that we had no deletion timestamp on the DV

PR #2203, opened 2022-03-23
Log UID of DataImportCron, DataSource & DataVolume

arnongilboa:

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -330,9 +330,10 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 }
 
 func (r *DataImportCronReconciler) deleteErroneousDataVolume(ctx context.Context, cron *cdiv1.DataImportCron, dv *cdiv1.DataVolume) error {
+	log := r.log.WithValues("name", dv.Name).WithValues("uid", dv.UID)
 	if cond := findConditionByType(cdiv1.DataVolumeRunning, dv.Status.Conditions); cond != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -333,7 +332,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 func (r *DataImportCronReconciler) deleteErroneousDataVolume(ctx context.Context, cron *cdiv1.DataImportCron, dv *cdiv1.DataVolume) error {
 	if cond := findConditionByType(cdiv1.DataVolumeRunning, dv.Status.Conditions); cond != nil {
 		if cond.Status == corev1.ConditionFalse && cond.Reason == common.GenericError {
-			r.log.Info("Delete DataVolume and reset DesiredDigest due to error", "message", cond.Message)
+			r.log.Info("Delete DataVolume and reset DesiredDigest due to error", "name", dv.Name, "uid", dv.UID, "message", cond.Message)
```
awels, line 335: I would do
```go
log := r.log.WithValues("name", dv.Name).WithValues("uid", dv.UID)
...
log.Info("Delete DataVolume and reset DesiredDigest due to error")
```
This way the name and UID will always be in the same location in the log. Regardless of what is actually being logged. This should make it easier to parse for both humans and computers. And if you have multiple log statements in a single function, You can define the log at the top, and re-use it. I think most of the added information can be done like this in the PR.

awels (approved):
/approve

PR #2204, opened 2022-03-24
Test flakiness - DataImportCron - retry update at least once

brybacki:

When updating one can get "Operation cannot be fulfilled on NAME the object has been modified;
please apply your changes to the latest version and try again". Doing a get and update
in loop makes it robust.

Fixes #

Tesk sometimes fails on update -> https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2200/pull-containerized-data-importer-e2e-k8s-1.22-hpp/1506737762038976512

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/dataimportcron_test.go
+++ b/tests/dataimportcron_test.go
@@ -3,6 +3,7 @@ package tests_test
 import (
 	"context"
 	"fmt"
+	cdiclientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -45,6 +46,13 @@ var _ = Describe("DataImportCron", func() {
 	BeforeEach(func() {
 		ns = f.Namespace.Name
 	})
+
+	updateDigest := func(digest string) func(cron *cdiv1.DataImportCron) *cdiv1.DataImportCron {
+		return func(cron *cdiv1.DataImportCron) *cdiv1.DataImportCron {
+			cron.Annotations[controller.AnnSourceDesiredDigest] = digest
+			return cron
+		}
+	}
 
 	table.DescribeTable("should", func(garbageCollection, retention, createErrorDv bool, repeat int) {
 		reg, err := getDataVolumeSourceRegistry(f)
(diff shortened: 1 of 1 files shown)


tests/dataimportcron_test.go
```diff
@@ -103,11 +104,13 @@ var _ = Describe("DataImportCron", func() {
...
-					cron.Annotations[controller.AnnSourceDesiredDigest] = "sha256:12345678900987654321"
-					cron, err = f.CdiClient.CdiV1beta1().DataImportCrons(ns).Update(context.TODO(), cron, metav1.UpdateOptions{})
-					Expect(err).ToNot(HaveOccurred())
+
+					//get and update!!!
+					Eventually(updateDataImportCron(f.CdiClient, ns, cronName,
+						func(cron *cdiv1.DataImportCron) *cdiv1.DataImportCron {
+							cron.Annotations[controller.AnnSourceDesiredDigest] = "sha256:12345678900987654321"
+							return cron
+						}), time.Second*30, time.Second).Should(BeNil())
```
arnongilboa, line 113: Not sure if I like it. Why should we fail on get/update if the controller works as expected and it and the test are the only ones updating of the CR? retrying it 30 times seems like sweeping bad controller behavior "under the rug".

awels: I think the idea is to not fail in case something else updated the DataImportCron in between us getting it, and trying to update. This is entirely possible, and since this is a rare failure, it could just be timing, and retrying it would make sense. Now retrying it 30 times seems a little overkill, if one retry is not enough it will likely be an issue in the controller if it is updating the object that much.

brybacki: right, 1 try should be enough ;)

brybacki: @arnongilboa @awels so what do you think, how to write this single retry in the easiest most readable way?
Use `Eventually` with default values, or with small timeout?

brybacki: introduced helper retryOnceOnErr, what do you think is this now good?

maybe we could improve it further.

brybacki:
@awels what do you think?
BTW: only imageio tests are failing here - known problem. all else is green

brybacki:
/test pull-containerized-data-importer-e2e-k8s-1.21-ceph
failed: warm imageio

PR #2205, opened 2022-03-25
Improve handling of prePopulated DV

brybacki:

Previously, when a DataVolume was Reconciled then CDI created PVC
regardless of prePopulated annotation.

There is a possibility that DataVolume and PVC pair is being
restored or moved from other cluster. When DataVolume is marked
as prePopulated, the CDI should not try to populate it.

Fixes #

**Release note**:

```release-note
Handle prepopulated DV
```

### DIFF
--- a/pkg/controller/datavolume-conditions.go
+++ b/pkg/controller/datavolume-conditions.go
@@ -122,7 +122,7 @@ func updateReadyCondition(conditions []cdiv1.DataVolumeCondition, status corev1.
 }
 
 func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.PersistentVolumeClaim, reason string) []cdiv1.DataVolumeCondition {
-	if pvc != nil {
+	if pvc != nil && pvc.Name != "" {
 		pvcCondition := getPVCCondition(pvc.GetAnnotations())
 		switch pvc.Status.Phase {
 		case corev1.ClaimBound:
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -438,88 +438,92 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
-	if !pvcExists {
-		if selectedCloneStrategy == SmartClone {
-			snapshotClassName, _ := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
-			return r.reconcileSmartClonePvc(log, datavolume, pvcSpec, transferName, snapshotClassName)
-		}
-		if selectedCloneStrategy == CsiClone {
-			csiDriverAvailable, err := r.storageClassCSIDriverExists(pvcSpec.StorageClassName)
-			if err != nil && !k8serrors.IsNotFound(err) {
-				return reconcile.Result{}, err
+	_, prePopulated := datavolume.Annotations[AnnPrePopulated]
+
+	if !prePopulated {
+		if !pvcExists {
+			if selectedCloneStrategy == SmartClone {
+				snapshotClassName, _ := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
+				return r.reconcileSmartClonePvc(log, datavolume, pvcSpec, transferName, snapshotClassName)
 			}
-			if !csiDriverAvailable {
-				// err csi clone not possible
-				return reconcile.Result{},
-					r.updateDataVolumeStatusPhaseWithEvent(cdiv1.CloneScheduled, datavolume, pvc, selectedCloneStrategy,
+			if selectedCloneStrategy == CsiClone {
+				csiDriverAvailable, err := r.storageClassCSIDriverExists(pvcSpec.StorageClassName)
+				if err != nil && !k8serrors.IsNotFound(err) {
+					return reconcile.Result{}, err
+				}
+				if !csiDriverAvailable {
+					// err csi clone not possible
+					return reconcile.Result{},
+						r.updateDataVolumeStatusPhaseWithEvent(cdiv1.CloneScheduled, datavolume, pvc, selectedCloneStrategy,
+							DataVolumeEvent{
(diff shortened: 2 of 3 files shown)


brybacki:
Both tests fail on:
```
"reason": "Error",
"message": "Unable to connect to imageio data source: bad status: 403 Forbidden kubevirt.io/containerized-data-importer/pkg/importer.createImageioReader \tpkg/importer/imageio-datasource.go:589 kubevirt.io/containerized-data-importer/pkg/importer.NewImageioDataSource \tpkg/importer/imageio-datasource.go:71 main.newDataSource \tcmd/cdi-importer/importer.go:235 main.handleImport \tcmd/cdi-importer/importer.go:146 main.main \tcmd/cdi-importer/importer.go:116 runtime.main \tGOROOT/src/runtime/proc.go:255 runtime.goexit \tGOROOT/src/runtime/asm_amd64.s:1581"
}
```

mhenriks:
Can you give an example of the bug/issue addressed here?  Is it simply that a "pre populated" DV may exist but the PVC may not yet?  What will be the DV status in that case?

tests/datavolume_test.go
```diff
@@ -1119,6 +1119,24 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			dataVolume.Annotations["cdi.kubevirt.io/storage.prePopulated"] = dataVolumeName
+
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
+			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func() cdiv1.DataVolumePhase {
+				dv, _ := f.CdiClient.CdiV1beta1().DataVolumes(dataVolume.Namespace).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return dv.Status.Phase
+			}, timeout, pollingInterval).Should(Equal(cdiv1.Succeeded))
```
mhenriks, line 1134: If the PVC does not exist, this should not be Succeeded

brybacki: Thanks for review. 

Yes, I have started this PR before the discussions on the behavior, and realizing that it is better to use real statuses like `Pending ` and `Succeeded`. That is why I made it Work In Progress.

Now I can go back to work and make it production ready.

brybacki: Done, now it is in Pending, and transitions to Succeeded when PVC is available

brybacki:
lets try If CI is up
/test pull-containerized-data-importer-e2e-k8s-1.22-hpp

pkg/controller/datavolume-controller.go
```diff
@@ -438,88 +438,92 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
-		if err := r.ensureExtendedToken(pvc); err != nil {
-			return reconcile.Result{}, err
-		}
-	case CsiClone:
-		switch pvc.Status.Phase {
-		case corev1.ClaimBound:
-			if err := r.setCloneOfOnPvc(pvc); err != nil {
+		switch selectedCloneStrategy {
+		case HostAssistedClone:
+			if err := r.ensureExtendedToken(pvc); err != nil {
```
mhenriks, line 503: I think this PR does a good job of handling the case where a DV has "pre populated" annotation and the PVC does not exist.  But it appears we have to better handle the case where the PVC has "populated for".

This bug has more details:  https://bugzilla.redhat.com/show_bug.cgi?id=2070366

Basically, the controller sees "populated for", takes ownership of the PVC, but still gets to this ` ensureExtendedToken` call

brybacki: do you want me to handle the problem with this PR, or create a new PR?

brybacki: updated the logic, now I need to prepare a test so this can be confirmed,

brybacki: looking at the tests

brybacki:
Updated the logic to cover DV and PVC, also added a test that failed on previous logic on host assisted clone and also on csi clone. 

There is only one problem left. If I try to write a test without adding a `source` PVC it fails on :
```
                Message: "admission webhook \"datavolume-validate.cdi.kubevirt.io\" denied the request:  Source PVC cdi-e2e-tests-dv-func-test-rq7w9/sourcepvcempty not found",
```

brybacki:
Ok, that is bad. I know what to do. 
I'll go back to small change, to only handle annotation on DV in this PR, and I'll open a new PR with the changes to handle pvc correctly.

brybacki:
@mhenriks I have reverted change here to only handle DV with prePopulated annotation. I would like this to be reviewed and merged. The rest of the problem will be done in a separate PR. 

WIP PR here https://github.com/kubevirt/containerized-data-importer/pull/2227. I am still trying to analyze the case and discover edge cases so it might be a mess.

pkg/controller/datavolume-conditions.go
```diff
@@ -122,7 +122,7 @@ func updateReadyCondition(conditions []cdiv1.DataVolumeCondition, status corev1.
 }
 
 func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.PersistentVolumeClaim, reason string) []cdiv1.DataVolumeCondition {
-	if pvc != nil {
+	if pvc != nil && pvc.Name != "" {
```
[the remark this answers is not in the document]
brybacki: At the beginning of main reconcile loop we do `pvc := &corev1.PersistentVolumeClaim{}` to get the PVC. 
If it is not found we have an empty PVC with no name. 

I can make it null, but I would need to carefully check if all the code is ready for null. This needs to be improved, and I'll try to clean up on the next PR.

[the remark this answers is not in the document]
mhenriks: Yeah this is pretty ugly, please handle in #2227

aglitke:
Oops.  This is already on release-v1.49

PR #2222, opened 2022-03-31
Make size optional when cloning using the Storage API

alromeros:

When cloning a _Data Volume_, the size of the target can be potentially obtainable via the source _PVC_. This discards the need to explicitly specify it.
    
Considering that, this _PR_ aims to modify the cloning process in order to allow omitting the `resources.request.storage` field when cloning a _PVC_, which will be obtained from the source instead.

**Special notes for your reviewer**: This is a work in progress.

**Release note**:

```release-note
NONE
```


pkg/controller/datavolume-controller.go
```diff
@@ -438,6 +438,20 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
+	// When cloning a PVC, if the target's size is not specified,
+	// said value can be attainable from the source PVC
+	if selectedCloneStrategy != NoClone {
+		targetRequest, hasTargetRequest := pvcSpec.Resources.Requests[corev1.ResourceStorage]
+		if hasTargetRequest && targetRequest.IsZero() {
```
ShellyKa13, line 445: isn't it also ok that hasTargetRequest will be false? like either:
if !hasTargetRequest || targetRequest.IsZero() {
also asking about line 1670: if !hasSrcRequest || !hasSrcCapacity || !hasTargetRequest {
isnt that ok that !hasTargetRequest in case we have storage API?

pkg/controller/datavolume-controller.go
```diff
@@ -438,6 +438,20 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+			sourcePvc, _ := r.findSourcePvc(datavolume)
+			if selectedCloneStrategy == SmartClone || selectedCloneStrategy == CsiClone {
+				pvcSpec.Resources.Requests[corev1.ResourceStorage] = *sourcePvc.Status.Capacity.Storage()
```
ShellyKa13, line 448: Is that already includes the fsoversize?

pkg/controller/datavolume-controller.go
```diff
@@ -1666,7 +1680,7 @@ func (r *DatavolumeReconciler) validateAdvancedCloneSizeCompatible(
 	if err != nil {
 		return false, err
 	}
-	if usableSpace.Cmp(targetRequest) > 0 {
+	if usableSpace.Cmp(targetRequest) > 0 && !targetRequest.IsZero() {
```
ShellyKa13, line 1683: like I asked in a previous commit isn't it ok to also have !hasTargetRequest in case of storage API?

pkg/controller/util.go
```diff
@@ -1087,16 +1087,22 @@ func getDefaultAccessModes(c client.Client, storageClass *storagev1.StorageClass
 	return nil, errors.Errorf("no accessMode defined on StorageProfile for %s StorageClass", storageClass.Name)
 }
 
-func volumeSize(c client.Client, storage *cdiv1.StorageSpec, volumeMode *v1.PersistentVolumeMode) (*resource.Quantity, error) {
+func volumeSize(c client.Client, spec cdiv1.DataVolumeSpec, volumeMode *v1.PersistentVolumeMode) (*resource.Quantity, error) {
```
ShellyKa13, line 1090: just making sure - is this function is only being called in the context of having storage API?

pkg/controller/datavolume-controller.go
```diff
@@ -438,6 +438,20 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
+	// When cloning a PVC, if the target's size is not specified,
+	// said value can be attainable from the source PVC
+	if selectedCloneStrategy != NoClone {
+		targetRequest, hasTargetRequest := pvcSpec.Resources.Requests[corev1.ResourceStorage]
```
ShellyKa13, line 444: if I understand correctly at this point hasTargetRequest can't be false right? maybe return an error if somehow it is?

pkg/controller/datavolume-controller.go
```diff
@@ -438,6 +438,20 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
...
+		if hasTargetRequest && targetRequest.IsZero() {
+			sourcePvc, _ := r.findSourcePvc(datavolume)
+			if selectedCloneStrategy == SmartClone || selectedCloneStrategy == CsiClone {
+				pvcSpec.Resources.Requests[corev1.ResourceStorage] = *sourcePvc.Status.Capacity.Storage()
+			} else {
+				// HostAssistedClone
```
ShellyKa13, line 450: will there be a something here or is it supposed to be like this? what happens in this case?

pkg/controller/datavolume-controller.go
```diff
@@ -438,6 +438,20 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
+	// When cloning a PVC, if the target's size is not specified,
+	// said value can be attainable from the source PVC
+	if selectedCloneStrategy != NoClone {
+		targetRequest, hasTargetRequest := pvcSpec.Resources.Requests[corev1.ResourceStorage]
+		if hasTargetRequest && targetRequest.IsZero() {
+			sourcePvc, _ := r.findSourcePvc(datavolume)
```
ShellyKa13, line 446: currently you only use this value in the if so maybe assign it inside the if?

pkg/controller/datavolume-controller.go
alromeros: Yes, It shouldn't be, but I wanted to make sure. I'll later add proper error handling, thanks!

pkg/controller/datavolume-controller.go
alromeros: Yeah, still in progress. That's probably the hardest case so I'll still need some time before committing the changes.

pkg/controller/util.go
alromeros: Exactly, as far as I know, It's only called when using the storage API.

pkg/controller/datavolume-controller.go
alromeros: It'll probably be also used inside the `else`, but I'm still not 100% sure.

pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -466,7 +474,8 @@ func validateStorageSize(resources v1.ResourceRequirements, field *k8sfield.Path
 			}
 			return &cause, false
 		}
-	} else {
+	} else if spec.Storage == nil || spec.Source.PVC == nil {
```
brybacki, line 477: does it mean we are going to allow empty size on both the `PVC` and `Storage`?

brybacki: nevermind

pkg/controller/datavolume-controller.go
brybacki: I wonder, wouldn't it be cleaner if this code is placed inside the `RenderPvcSpec`? The you would correctly render the size requirements of the target PVC spec at the beginning.

brybacki: The idea of this render function is, that given `Storage` or `PVC` fields or any other things from the DataVolume we can create a single consistent view of a Target PVC for CDI controllers to use.

pkg/controller/util.go
brybacki: yes, I suppose now the name is too generic, previously it was obvious from arguments `cdiv1.StorageSpec`.

pkg/controller/util.go
```diff
@@ -1087,16 +1087,22 @@ func getDefaultAccessModes(c client.Client, storageClass *storagev1.StorageClass
...
 	// resources.requests[storage] - just copy it to pvc,
-	requestedSize, found := storage.Resources.Requests[v1.ResourceStorage]
+	requestedSize, found := spec.Storage.Resources.Requests[v1.ResourceStorage]
+
 	if !found {
-		return nil, errors.Errorf("Datavolume Spec is not valid - missing storage size")
+		// Storage size can be empty when using PVC source
+		if spec.Source.PVC != nil {
+			return &requestedSize, nil
```
brybacki, line 1097: what is the value of `requestedSize` if it is not Found. Isn't it nil?

pkg/controller/util.go
```diff
@@ -1087,16 +1087,22 @@ func getDefaultAccessModes(c client.Client, storageClass *storagev1.StorageClass
...
-func volumeSize(c client.Client, storage *cdiv1.StorageSpec, volumeMode *v1.PersistentVolumeMode) (*resource.Quantity, error) {
+func volumeSize(c client.Client, spec cdiv1.DataVolumeSpec, volumeMode *v1.PersistentVolumeMode) (*resource.Quantity, error) {
 	// resources.requests[storage] - just copy it to pvc,
-	requestedSize, found := storage.Resources.Requests[v1.ResourceStorage]
+	requestedSize, found := spec.Storage.Resources.Requests[v1.ResourceStorage]
+
 	if !found {
-		return nil, errors.Errorf("Datavolume Spec is not valid - missing storage size")
+		// Storage size can be empty when using PVC source
+		if spec.Source.PVC != nil {
```
brybacki, line 1096: what you check here is basicly `isClone`. I think it would be more readable to state:
```
// Storage size can be empty when cloning
```
or even better to do 
```
isClone := spec.Source.PVC != nil
if isClone ...
```

but I think that is just the matter of style.

brybacki:
Very good direction. 
:+1:

pkg/controller/datavolume-controller.go
alromeros: I agree, I'm not very happy including that code inside the reconcile, but we need to know the `selectedCloneStrategy` before assigning the storage size. I don't know if we have a better option.

pkg/controller/util.go
alromeros: In that case `requestedSize` would be 0, so it can be later checked with the `IsZero()` method. Simply returning `nil` would later cause a nil pointer dereference, so I chose to return the empty `requestedSize`. That said, I'm not 100% happy with that, so I can try to make it more readable.

pkg/controller/util.go
alromeros: I agree about that, though maybe is a little bit redundant to create a variable for just that condition?

brybacki: the advocates of clean code might say that this variable is explaining things better than any comment so is not redundant

pkg/controller/util.go
brybacki: maybe a const somewhere that says ZERO, maybe there is such Quantity const defined somewhere?

pkg/controller/datavolume-controller.go
brybacki: do you need other values than NoClone?

because `NoClone` is in practice equivalent to `spec.Source.PVC == nil`

brybacki: ok, I see, you need the `Smart/CsiClone`

pkg/controller/datavolume-controller.go
```diff
@@ -169,10 +170,18 @@ const (
...
 	// MessageNamespaceTransferInProgress is a const for reporting target transfer
 	MessageNamespaceTransferInProgress = "Transferring PersistentVolumeClaim for DataVolume %s/%s"
+	// ImgSizeNotReady is a const to report that it's not yet possible to obtain the size of the image
+	ImgSizeNotReady = "Image size not ready"
 
 	// AnnCSICloneRequest annotation associates object with CSI Clone Request
 	AnnCSICloneRequest = "cdi.kubevirt.io/CSICloneRequest"
 
+	// AnnVirtualImageSize annotation contains the Virtual Image size of a PVC used for host-assisted cloning
+	AnnVirtualImageSize = "cdi.Kubervirt.io/virtualSize"
```
brybacki, line 180: why "SourceCapacity" starts with capital letter and "virtualSize" not? Do we have some rules about that?

alromeros: My fault, I didn't notice that. I'll change "VirtualSize" to lowercase.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+		}
+
+		r.log.V(3).Info("Size-detection pod created\n", "pod.Name", pod.Name, "pod.Namespace", pod.Namespace)
+	}
+
+	return pod, nil
+}
+
+// makeSizeDetectionPodSpec creates and returns the full size-detection pod spec
+func makeSizeDetectionPodSpec(sourcePvc *corev1.PersistentVolumeClaim, client client.Client, pullPolicy string) *corev1.Pod {
```
awels, line 2750: Any reason you didn't make this part of the reconciler? that way you don't have to pass in the client, and if you follow my below suggestion, you don't have to pass in the image name either.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+}
+
+// hasPodSucceeded returns true if a pod is in 'Succeeded' phase, false if not
+func hasPodSucceeded(pod *corev1.Pod) bool {
+	return pod != nil && pod.Status.Phase == corev1.PodSucceeded
+}
+
+// sizeDetectionPodName returns the name of the size-detection pod accoding to the source PVC's UID
+func sizeDetectionPodName(pvc *corev1.PersistentVolumeClaim) string {
+	return "size-detection-" + string(pvc.UID)
```
awels, line 2909: This works and is technically not wrong, but we use `fmt.Sprintf("size-detection-%s", pvc.UID)` in most of these create name type functions.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+	} else {
+		targetSize = *sourceSize
+	}
+
+	pvcSpec.Resources.Requests[corev1.ResourceStorage] = targetSize
+	return true, nil
+}
+
+// getSizeFromAnnotations checks the source PVC's annotations and returns the requested size if it has already been obtained
+func getSizeFromAnnotations(sourcePvc *corev1.PersistentVolumeClaim) (string, bool) {
```
awels, line 2672: When I see a function called getSizeSomething, I am expecting the return value to be a number of some kind, in this case I would expect either a Quantity or an int64 instead of a string.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+	}
+
+	return "0", false
+}
+
+// getSizeFromPod attempts to get the image size from a pod that directly obtains said value from the source PVC
+func (r *DatavolumeReconciler) getSizeFromPod(sourcePvc *corev1.PersistentVolumeClaim) (string, error) {
+	// The pod should not be created until the source PVC has finished the import process
+	if !hasPvcSucceeded(sourcePvc) {
+		return ImgSizeNotReady, nil
```
awels, line 2690: I like that you are differentiating between pod is not ready type errors and other errors. But returning that as part of the actual value is probably not what we want, especially if we decide to return a number. It might make sense to return your own error type so in a higher up function you can decide to do something special based on this special error type. See https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/data-processor.go#L65-L70 for an example of how to make a custom error.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+
+	pod, err := r.createSizeDetectionPod(sourcePvc)
+	if err != nil {
+		return ImgSizeNotReady, err
+	} else if !hasPodSucceeded(pod) {
+		return ImgSizeNotReady, nil
+	}
+	// Parse raw image size from the pod's termination message
+	imgSize := pod.Status.ContainerStatuses[0].State.Terminated.Message
+	err = r.client.Delete(context.TODO(), pod)
```
awels, line 2701: Might want to wait with deleting the pod until you are sure everything you wanted to save the value in has updated. Like the updateClonePVCAnnotations below. If that fails for whatever reason, now your pod is gone and you have to redo the entire pod creation cycle to get the value again. If you reverse the order, then you are sure the update was successful and you can delete confident the value is on the PVC.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+
+	return pod
+}
+
+// makeSizeDetectionObjectMeta creates and returns the object metadata for the size-detection pod
+func makeSizeDetectionObjectMeta(sourcePvc *corev1.PersistentVolumeClaim, blockOwnerDeletion, isController bool) *metav1.ObjectMeta {
+	return &metav1.ObjectMeta{
+		Name:      sizeDetectionPodName(sourcePvc),
+		Namespace: sourcePvc.Namespace,
+		Annotations: map[string]string{
```
awels, line 2794: Don't need these annotations

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+			AnnCreatedBy: "yes",
+		},
+		Labels: map[string]string{
+			common.CDILabelKey:        common.CDILabelValue,
+			common.CDIComponentLabel:  common.ImporterPodName,
+			common.PrometheusLabelKey: common.PrometheusLabelValue,
```
awels, line 2800: We are not exposing any prometheus metrics on this pod, so don't need the prometheus label.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+		},
+		OwnerReferences: []metav1.OwnerReference{
+			{
+				APIVersion:         "v1",
+				Kind:               "PersistentVolumeClaim",
```
awels, line 2805: I would make the owner of this pod the data volume instead of the PVC. This way we can easily watch for pods that are owned by DVs and figure out which DV is associated with the pod. You can also an ownerRef easier like this. The datavolume controller has some examples.
```go
OwnerReferences = []metav1.OwnerReference{
  *metav1.NewControllerRef(dataVolume, schema.GroupVersionKind{
    Group:   cdiv1.SchemeGroupVersion.Group,
    Version: cdiv1.SchemeGroupVersion.Version,
    Kind:    "DataVolume",
  }),
}
```

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+}
+
+// makeSizeDetectionContainerSpec creates and returns the size-detection pod's Container spec
+func makeSizeDetectionContainerSpec(image, pullPolicy string) *corev1.Container {
+	return &corev1.Container{
+		Name:            "size-detection-volume",
+		Image:           image,
+		ImagePullPolicy: corev1.PullPolicy(pullPolicy),
+		Command:         []string{"/usr/bin/cdi-image-size-detection"},
+		Args:            []string{"-url", common.ImporterWritePath, "-path", corev1.TerminationMessagePathDefault},
```
awels, line 2822: I would make -path -termination-message-path so there is no confusion about what this parameter does. At first I thought it was the path to the disk.img file, but that is the -url parameter (which might be better as path)

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+				ClaimName: pvcName,
+				ReadOnly:  false,
+			},
+		},
+	}
+}
+
+// inflateSizeWithOverhead inflates the obtained image size with proper overhead calculations
+func inflateSizeWithOverhead(client client.Client,
+	imgSize string,
```
awels, line 2847: I think this being either a quantity or int64 would be better, that way we parse at the source, and pass it along in the right type format. Then you don't have to parse it in this function.

awels (commented):
Very good progress here, just a bunch of nitpicks and some style issues. But looks pretty good overall.

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -400,6 +400,49 @@ var _ = Describe("Validating Webhook", func() {
...
+				Resources: corev1.ResourceRequirements{
+					Requests: requests,
+				},
+			}
+			dv, pvc := newDataVolumeClone(storage, nil)
+			resp := validateDataVolumeCreate(dv, pvc)
+			Expect(resp.Allowed).To(Equal(true))
+		})
+
+		It("should reject empty Requests when cloning using PVC API", func() {
```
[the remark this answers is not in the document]
alromeros: We don't, I'll add one in the next commit.

pkg/controller/datavolume-controller.go
```diff
@@ -492,6 +504,18 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 	prePopulated bool,
 	selectedCloneStrategy cloneStrategy) (reconcile.Result, error) {
 
+	// When cloning a PVC, if the target's size is not specified,
+	// said value can be attainable from the source PVC
+	targetRequest, hasTargetRequest := pvcSpec.Resources.Requests[corev1.ResourceStorage]
+	if hasTargetRequest && targetRequest.IsZero() {
```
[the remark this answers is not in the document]
alromeros: Yeah, after `RenderPvcSpec` it should be impossible to have a false `hasTargetRequest`. However, I preferred to check the value before accessing `IsZero()` just to be extra sure to avoid possible segfaults. I can add the `!hasTargetRequest || ...` if you prefer.

pkg/controller/datavolume-controller.go
```diff
@@ -492,6 +504,18 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
...
+		done, err := r.detectCloneSize(datavolume, pvcSpec, selectedCloneStrategy)
+		if err != nil {
+			return reconcile.Result{}, err
+		} else if !done {
+			return reconcile.Result{RequeueAfter: 2 * time.Second}, nil
```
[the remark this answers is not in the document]
alromeros: Thank you for this clarification, I guessed that to requeue wasn't the best way to proceed, but I couldn't find other alternatives in the code. That said, wouldn't that affect the [other pod](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume-controller.go#L1400) created in this controller?

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+		if !available {
+			imgSize, err = r.getSizeFromPod(sourcePvc)
+			if err != nil {
+				return false, err
+			} else if imgSize == ImgSizeNotReady {
+				return false, nil
+			}
+		}
+		// We enable the clone-controller to skip the size comparison requirement
+		dv.Annotations[AnnPermissiveClone] = "true"
```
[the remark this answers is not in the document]
alromeros: I agree, I'll change it :+1:

pkg/controller/datavolume-controller.go
alromeros: I chose to return a string so it can be more easily converted to Quantity after the overhead calculation, but `int64` would also work well. I'll change it.

pkg/controller/datavolume-controller.go
alromeros: Yeah, that makes more sense. Thanks!

pkg/controller/datavolume-controller.go
alromeros: I'll change it :+1: Sometimes I've struggled to guess when I should make a function part of the reconciler, or when I should just pass the client as an argument. Are there any guidelines for this?

pkg/controller/datavolume-controller.go
alromeros: Yeah, that makes sense.

pkg/controller/datavolume-controller.go
alromeros: I used [this other function](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume-controller.go#L1315) as a reference, but I can change it if you prefer.

pkg/controller/datavolume-controller.go
```diff
@@ -2599,3 +2623,288 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+// updateClonePVCAnnotations updates the clone-related annotations of the source PVC
+func (r *DatavolumeReconciler) updateClonePVCAnnotations(sourcePvc *corev1.PersistentVolumeClaim, virtualSize string) error {
+	currCapacity := sourcePvc.Status.Capacity
+	sourcePvc.Annotations[AnnVirtualImageSize] = virtualSize
+	sourcePvc.Annotations[AnnSourceCapacity] = currCapacity.Storage().String()
+	return r.client.Update(context.TODO(), sourcePvc)
+}
+
+// hasPvcSucceeded returns true if a PVC is in 'Succeeded' phase, false if not
+func hasPvcSucceeded(pvc *corev1.PersistentVolumeClaim) bool {
```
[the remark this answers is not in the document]
alromeros: Yeah, that sense. I can also move `hasPodSucceeded`.

pkg/controller/clone-controller.go
```diff
@@ -768,14 +768,19 @@ func ValidateCanCloneSourceAndTargetContentType(sourcePvc, targetPvc *corev1.Per
...
 // ValidateCanCloneSourceAndTargetSpec validates the specs passed in are compatible for cloning.
-func ValidateCanCloneSourceAndTargetSpec(sourceSpec, targetSpec *corev1.PersistentVolumeClaimSpec, contentType cdiv1.DataVolumeContentType) error {
-	err := ValidateCloneSize(sourceSpec.Resources, targetSpec.Resources)
-	if err != nil {
+func ValidateCanCloneSourceAndTargetSpec(sourcePvc, targetPvc *corev1.PersistentVolumeClaim, contentType cdiv1.DataVolumeContentType) error {
+	err := ValidateCloneSize(sourcePvc.Spec.Resources, targetPvc.Spec.Resources)
+	_, available := targetPvc.Annotations[AnnPermissiveClone]
+	// Due to filesystem overhead differences, the target's size can sometimes be smaller
+	// than the source's one when obtaining said value with the size-detection pod.
+	// In those cases, we use the 'AnnPermissiveClone' annotation to skip the size validation.
```
brybacki, line 776: What is the actual scenario for this case? Can we handle it in a better way in the future?

pkg/controller/clone-controller.go
```diff
@@ -768,14 +768,19 @@ func ValidateCanCloneSourceAndTargetContentType(sourcePvc, targetPvc *corev1.Per
 }
 
 // ValidateCanCloneSourceAndTargetSpec validates the specs passed in are compatible for cloning.
-func ValidateCanCloneSourceAndTargetSpec(sourceSpec, targetSpec *corev1.PersistentVolumeClaimSpec, contentType cdiv1.DataVolumeContentType) error {
-	err := ValidateCloneSize(sourceSpec.Resources, targetSpec.Resources)
-	if err != nil {
+func ValidateCanCloneSourceAndTargetSpec(sourcePvc, targetPvc *corev1.PersistentVolumeClaim, contentType cdiv1.DataVolumeContentType) error {
```
brybacki, line 771: why the change from ClaimSpec to Claim?

pkg/controller/clone-controller.go
alromeros: When the fs overhead is smaller in the target PVC than in the source PVC (and even when they are the same), the obtained size ends up being smaller than the one specified in the source specs, so I've decided to use this annotation to allow the clone.

I guess this annotation could also be used to fix the already existing bug that happens when cloning from block to filesystem mode.

pkg/controller/clone-controller.go
alromeros: So we can then check the `AnnPermissiveClone` annotation in the size validation-

pkg/controller/datavolume-controller.go
```diff
@@ -2601,3 +2638,261 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+}
+
+// getSizeFromAnnotations checks the source PVC's annotations and returns the requested size if it has already been obtained
+func getSizeFromAnnotations(sourcePvc *corev1.PersistentVolumeClaim) (int64, bool) {
+	virtualImageSize, available := sourcePvc.Annotations[AnnVirtualImageSize]
+	if available == true {
+		sourceCapacity, available := sourcePvc.Annotations[AnnSourceCapacity]
+		currCapacity := sourcePvc.Status.Capacity
+		// Checks if the original PVC's capacity has changed
+		if available && currCapacity.Storage().Cmp(resource.MustParse(sourceCapacity)) == 0 {
```
brybacki, line 2695: :+1:

pkg/controller/datavolume-controller.go
```diff
@@ -2601,3 +2638,261 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+			},
+		},
+	}
+}
+
+// inflateSizeWithOverhead inflates the obtained image size with proper overhead calculations
+func (r *DatavolumeReconciler) inflateSizeWithOverhead(imgSize int64, pvcSpec *corev1.PersistentVolumeClaimSpec) (resource.Quantity, error) {
+	var returnSize resource.Quantity
+
+	if resolveVolumeMode(pvcSpec.VolumeMode) == corev1.PersistentVolumeFilesystem {
```
brybacki, line 2866: I think this can really reuse the function `volumeSize`. It is almost like a copy paste for some non trivial logic. 

```
func volumeSize(c client.Client, storage *cdiv1.StorageSpec, volumeMode *v1.PersistentVolumeMode) (*resource.Quantity, error) {

```

pkg/controller/util.go
brybacki: and I supposer this function can be now divided into two parts, 
1. one called from renderPVC from storage API context, 
2. one that actually does calculate the new size (inside the `	if resolveVolumeMode(volumeMode) == v1.PersistentVolumeFilesystem {` )

And the second one can then be reused inside the function  `inflateSizeWithOverhead` in dv controller

pkg/controller/clone-controller.go
brybacki: ok

pkg/controller/clone-controller.go
brybacki: which bug?

alromeros: This one [Bug 2079781](https://bugzilla.redhat.com/show_bug.cgi?id=2079781) (let me know if is inappropriate to share this information here)

pkg/controller/datavolume-controller_test.go
```diff
@@ -1581,6 +1581,179 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			dv := newCloneDataVolumeWithEmptyStorage("test-dv", "default")
+			cloneStrategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyHostAssisted)
+			storageProfile := createStorageProfileWithCloneStrategy(scName, []cdiv1.ClaimPropertySet{
+				{AccessModes: accessMode, VolumeMode: &blockMode}}, &cloneStrategy)
+
+			reconciler := createDatavolumeReconciler(dv, storageProfile, sc)
+			pvcSpec, err := RenderPvcSpec(reconciler.client, reconciler.recorder, reconciler.log, dv)
+			Expect(err).ToNot(HaveOccurred())
+			err = reconciler.detectCloneSize(dv, pvcSpec, HostAssistedClone)
+			Expect(err).To(HaveOccurred())
```
awels, line 1604: Can you make sure the error is a `IsNotFound` something like
```go
Expect(k8serrors.IsNotFound(err).To(BeTrue())
```

pkg/controller/datavolume-controller_test.go
```diff
@@ -1581,6 +1581,179 @@ var _ = Describe("All DataVolume Tests", func() {
...
+				{AccessModes: accessMode, VolumeMode: &blockMode}}, &cloneStrategy)
+
+			pvc := createPvcInStorageClass("test", metav1.NamespaceDefault, &scName, nil, nil, corev1.ClaimBound)
+			reconciler := createDatavolumeReconciler(dv, pvc, storageProfile, sc)
+
+			pvcSpec, err := RenderPvcSpec(reconciler.client, reconciler.recorder, reconciler.log, dv)
+			Expect(err).ToNot(HaveOccurred())
+			err = reconciler.detectCloneSize(dv, pvcSpec, HostAssistedClone)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(ErrImportPVCNotReady))
```
awels, line 1620: I don't think it should fail here, it should just return a size of 0. You could potentially record an event and check that in the unit test to ensure the right `0` is returned.

pkg/controller/datavolume-controller_test.go
```diff
@@ -1581,6 +1581,179 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			pvc := createPvcInStorageClass("test", metav1.NamespaceDefault, &scName, nil, nil, corev1.ClaimBound)
+			pvc.SetAnnotations(make(map[string]string))
+			pvc.GetAnnotations()[AnnPodPhase] = string(corev1.PodSucceeded)
+			reconciler := createDatavolumeReconciler(dv, pvc, storageProfile, sc)
+
+			pvcSpec, err := RenderPvcSpec(reconciler.client, reconciler.recorder, reconciler.log, dv)
+			Expect(err).ToNot(HaveOccurred())
+			err = reconciler.detectCloneSize(dv, pvcSpec, HostAssistedClone)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(ErrSizeDetectionPodNotReady))
```
awels, line 1638: Same don't think you should fail, just return a `0` size and let the pod phase changes trigger a new reconcile loop.

pkg/controller/datavolume-controller_test.go
```diff
@@ -1581,6 +1581,179 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			pod.Status.Phase = corev1.PodSucceeded
+			err := reconciler.client.Create(context.TODO(), pod)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Checks
+			pvcSpec, err := RenderPvcSpec(reconciler.client, reconciler.recorder, reconciler.log, dv)
+			Expect(err).ToNot(HaveOccurred())
+			err = reconciler.detectCloneSize(dv, pvcSpec, HostAssistedClone)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(ErrNonValidTermMsg))
```
awels, line 1663: See the endless comment above, I would delete the pod (check that we do this in the unit test)

pkg/controller/datavolume-controller_test.go
```diff
@@ -1581,6 +1581,179 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			pvcSpec, err := RenderPvcSpec(reconciler.client, reconciler.recorder, reconciler.log, dv)
+			Expect(err).ToNot(HaveOccurred())
+			expectedSize, err := inflateSizeWithOverhead(reconciler.client, int64(100), pvcSpec)
+
+			// Checks
+			err = reconciler.detectCloneSize(dv, pvcSpec, HostAssistedClone)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(dv.GetAnnotations()[AnnPermissiveClone]).To(Equal("true"))
+			targetSize := pvcSpec.Resources.Requests[corev1.ResourceStorage]
+			Expect(targetSize.Cmp(expectedSize)).To(Equal(0))
```
awels, line 1703: This works, alternative is:
```go
Expect(targetSize.Int64()).To(Equal(int64(100))
```
Not 100% sure if Int64 is a single or double value return function.

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
 						string(dv.Status.Progress) == "N/A"
 				}, timeout, pollingInterval).Should(BeTrue(), "DV Should succeed with storage.prePopulated==pvcName")
 			})
+
+			DescribeTable("Should clone with empty volume size without using size-detection pod",
+				func(sourceVolumeMode v1.PersistentVolumeMode, targetVolumeMode v1.PersistentVolumeMode) {
```
awels, line 580: if you have multiple arguments of the same type you can do arg1, arg2, arg3, etc type. so in this case it would be:
```go
func(sourceVolumeMode, targetVolumeMode v1.PersistentVolumeMode)
```

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
...
+					if sourceVolumeMode != targetVolumeMode && cloneType == "csivolumeclone" {
+						Skip("csivolumeclone only works for the same volumeMode")
+					}
+
+					if (sourceVolumeMode == v1.PersistentVolumeBlock || targetVolumeMode == v1.PersistentVolumeBlock) && !f.IsBlockVolumeStorageClassAvailable() {
+						Skip("Storage Class for block volume is not available")
+					}
+
+					dataVolume := utils.NewDataVolumeWithHTTPImportAndStorageSpec(dataVolumeName, "200Mi", fmt.Sprintf(utils.TinyCoreIsoURL, f.CdiInstallNs))
+					dataVolume.Spec.Storage.VolumeMode = &sourceVolumeMode
```
awels, line 598: You will likely want to specify the source and target storage class. We can't assume the default storage class supports the volume mode we are after. In particular for block storage. The framework has a field called BlockSCName that should contain the name of a storage class that supports block mode. In fact `f.IsBlockVolumeStorageClassAvailable()` uses that to check if the block storage class is available.

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
...
+					Expect(err).ToNot(HaveOccurred())
+					Expect(sourceMD5 == targetMD5).To(BeTrue())
+
+					By("Deleting verifier pod")
+					err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+					Expect(err).ToNot(HaveOccurred())
+				},
+				Entry("Block to block (empty storage size)", v1.PersistentVolumeBlock, v1.PersistentVolumeBlock),
+				Entry("Block to filesystem (empty storage size)", v1.PersistentVolumeBlock, v1.PersistentVolumeFilesystem),
+				Entry("Filesystem to block (empty storage size)", v1.PersistentVolumeFilesystem, v1.PersistentVolumeBlock),
```
awels, line 644: No sense in using this combination in the table. It will always get skipped since the clone type will always be network (fs -> block) is always network.

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
...
+				func(sourceVolumeMode v1.PersistentVolumeMode, targetVolumeMode v1.PersistentVolumeMode) {
+					// When cloning without defining the target's storage size, the source's size can be attainable
+					// by different means depending on the clone type and the volume mode used.
+					// Either if "block" is used as volume mode or smart/csi cloning is used as clone strategy,
+					// the value is simply extracted from the original PVC's spec.
+					if cloneType == "network" && sourceVolumeMode == v1.PersistentVolumeFilesystem {
+						Skip("Clone strategy and volume mode combination requires of size-detection pod")
+					}
+
+					if sourceVolumeMode != targetVolumeMode && cloneType == "csivolumeclone" {
```
awels, line 589: Don't need this skip csi clone/snapshot clone only works if the source and target volume mode are the same, and if they both use the same storage class.

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
...
+					if (sourceVolumeMode == v1.PersistentVolumeBlock || targetVolumeMode == v1.PersistentVolumeBlock) && !f.IsBlockVolumeStorageClassAvailable() {
+						Skip("Storage Class for block volume is not available")
+					}
+
+					dataVolume := utils.NewDataVolumeWithHTTPImportAndStorageSpec(dataVolumeName, "200Mi", fmt.Sprintf(utils.TinyCoreIsoURL, f.CdiInstallNs))
+					dataVolume.Spec.Storage.VolumeMode = &sourceVolumeMode
+					dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
+					Expect(err).ToNot(HaveOccurred())
+					f.ForceBindPvcIfDvIsWaitForFirstConsumer(dataVolume)
+					sourcePvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(dataVolume.Namespace).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
```
awels, line 602: I think we should probably wait for the source DV to be completed before we attempt to clone it without a size. Although it is probably a good test to immediately attempt to clone it and make sure the clone eventually completes once the source is done.

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
...
+
+					// We attempt to create the sizeless DV
+					targetDV := utils.NewDataVolumeForCloningWithEmptySize("target-dv", sourcePvc.Namespace, sourcePvc.Name, nil, &targetVolumeMode)
+					targetDataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, targetDV)
+					Expect(err).ToNot(HaveOccurred())
+					targetPvc, err := utils.WaitForPVC(f.K8sClient, targetDataVolume.Namespace, targetDataVolume.Name)
+					Expect(err).ToNot(HaveOccurred())
+					f.ForceBindPvcIfDvIsWaitForFirstConsumer(targetDataVolume)
+
+					By("Wait for target PVC Bound phase")
```
awels, line 613: Don't need to wait for this phase, waiting for the DV to complete will include that anyway

tests/cloner_test.go
```diff
@@ -574,6 +575,75 @@ var _ = Describe("all clone tests", func() {
...
+					By("Deleting verifier pod")
+					err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+					Expect(err).ToNot(HaveOccurred())
+					_, err = utils.WaitPodDeleted(f.K8sClient, utils.VerifierPodName, f.Namespace.Name, verifyPodDeletedTimeout)
+					Expect(err).ToNot(HaveOccurred())
+
+					By("Target file system pvc md5summing")
+					targetMD5, err := f.GetMD5(f.Namespace, targetPvc, diskImagePath, crossVolumeModeCloneMD5NumBytes)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(sourceMD5 == targetMD5).To(BeTrue())
```
awels, line 636: I think
```go
Expect(sourceMD5).To(Equal(targetMD5))
```
is easier to read. Also I would check after I already delete the verifier pod. If the comparison fails in this order, we won't delete the verifier pod. (Note it is currently not a problem since the namespace will get deleted, but better cleanup)

tools/cdi-image-size-detection/main.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	// Extract the data from the image
+	imgInfo, err := qemuOperations.Info(parsedURL)
+	if err != nil {
+		log.Fatalf("Unable to extract information from '%s': '%s'", imgPath, err.Error())
+	}
+
+	// Write the parsed virtual size to the termination message file
+	strSize := strconv.FormatInt(imgInfo.VirtualSize, decimal)
```
arnongilboa, line 53: why not simply use strconv.Itoa and drop the decimal const?

tools/cdi-image-size-detection/main.go
```diff
@@ -0,0 +1,60 @@
...
+	flag.Parse()
+	if imgPath == "" || termPath == "" {
+		log.Fatalf("One or more mandatory parameters are missing")
+	}
+}
+
+func main() {
+	// Initialize 'qemu-img' handler
+	log.Println("Initializing size-detection pod")
+	var qemuOperations = image.NewQEMUOperations()
```
arnongilboa, line 39: using := is preferable

tools/cdi-image-size-detection/main.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Extract the data from the image
+	imgInfo, err := qemuOperations.Info(parsedURL)
+	if err != nil {
+		log.Fatalf("Unable to extract information from '%s': '%s'", imgPath, err.Error())
+	}
+
+	// Write the parsed virtual size to the termination message file
+	strSize := strconv.FormatInt(imgInfo.VirtualSize, decimal)
+	err = util.WriteTerminationMessageToFile(termPath, strSize)
```
arnongilboa, line 54: why not use WriteTerminationMessage() which already uses common.PodTerminationMessageFile, and remove passing the termination-message-path arg?

pkg/controller/datavolume-controller.go
```diff
@@ -2601,3 +2638,236 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+		return 0, err
+	} else if !isPodComplete(pod) {
+		return 0, ErrSizeDetectionPodNotReady
+	}
+
+	// Parse raw image size from the pod's termination message
+	if pod.Status.ContainerStatuses == nil ||
+		pod.Status.ContainerStatuses[0].State.Terminated == nil ||
+		pod.Status.ContainerStatuses[0].State.Terminated.ExitCode > 0 {
+		return 0, ErrNonValidTermMsg
```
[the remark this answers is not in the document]
alromeros: Thanks for pointing this out! I'll do as proposed.

tools/cdi-image-size-detection/main.go
alromeros: I first tried using itoa but since `imgInfo.VirtualSize` is an int64, it didn't work. Maybe I can change it but I think that using `FormatInt` is a good option.

tools/cdi-image-size-detection/main.go
alromeros: I thought about this, maybe is a better option. However, I think that it's also good to have the possibility of choosing a different termination path, in case the pod uses a different one than the default.

pkg/controller/datavolume-controller.go
```diff
@@ -2601,3 +2638,236 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+			imgSizeInt, _ := strconv.ParseInt(virtualImageSize, 10, 64)
+			return imgSizeInt, true
+		}
+	}
+
+	return 0, false
+}
+
+// getSizeFromPod attempts to get the image size from a pod that directly obtains said value from the source PVC
+func (r *DatavolumeReconciler) getSizeFromPod(sourcePvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) (int64, error) {
```
arnongilboa, line 2706: I think it will be much cleaner if you remove the `getSizeFromPod()` and `datavolumeController.Watch(&source.Kind{Type: &corev1.Pod{}}`, and the containerized app can simply annotate the DV with the virtual size. See [this](https://github.com/kubevirt/containerized-data-importer/blob/main/tools/cdi-source-update-poller/main.go#L75-L76) for example.

tests/cloner_test.go
awels: Actually block -> fs is also always network.

pkg/controller/datavolume-controller.go
awels: The question is always what is better. In the current design the app doesn't know it is inside a k8s cluster and can run independently, but at the cost of having the controller that manages it be more complex (it needs to create/start the pod, wait for it to finish, do some gymnastics to get the output, and then use the value) vs the pod now needs to know it is inside a k8s cluster, it needs to be passed a kubeconfig, you need to build a client, and the app needs to do the update to the datavolume which the controller can then act upon.

In this particular case IMO the extra complexity in the controller is better than teaching the app it is inside a k8s cluster, but it is a totally subjective opinion

pkg/controller/datavolume-controller.go
```diff
@@ -2645,3 +2692,297 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+	exitCode := pod.Status.ContainerStatuses[0].State.Terminated.ExitCode
+
+	// We attempt to delete the pod
+	err := r.client.Delete(context.TODO(), pod)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	switch int(exitCode) {
+	case ErrBadArguments:
```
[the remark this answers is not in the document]
alromeros: Should I set an event with `cdiv1.Failed` in this case then? Or should I just remove this case from the switch?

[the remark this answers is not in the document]
awels: There are only 2 possible types for Events, Normal and Warning (Warnings are for when things go wrong, normal is for events that happen during normal operation). So you can definitely generate a warning event here

pkg/controller/util.go
brybacki: since this is pretty big PR and my comment is just a style nitpick I'll close this comment as resolved, and it can be done later if anyone thinks it can be improved

pkg/apiserver/webhooks/datavolume-validate_test.go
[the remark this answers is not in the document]
brybacki: I can it is already present

brybacki:
unfortunately you are in conflict again, probably with this:
https://github.com/kubevirt/containerized-data-importer/pull/2276
https://github.com/kubevirt/containerized-data-importer/pull/2277

pkg/controller/clone-controller.go
brybacki: Which value is reported by the size detection pod? Is it the:

If it is the real virtual image size then the target should have enough space to fit the image plus size overhead. So we can not skip the validation! 

This is probably the only one thing I would like to discsuss in the PR. 
I think @ShellyKa13 found some interesting problems with clone size, so maybe she can help with some ideas?

pkg/controller/datavolume-controller_test.go
brybacki: It is probably 5th time you need to Rebase, so I think you can leave it as is, because the PR gets bigger and bigger and now it would be good to focus only on important stuff and try to finish it and merge it. 

And keep the details noted somewhere maybe on a new issue, and then we can open follow up PR with some small cleanups. You decide. I prefer to have things fixed the first time. But I also like to have stuff done, and not blocked by such small details.

pkg/controller/datavolume-controller_test.go
```diff
@@ -1563,6 +1563,204 @@ var _ = Describe("All DataVolume Tests", func() {
...
+		It("Should get the size from the source PVC's annotations", func() {
+			dv := newCloneDataVolumeWithEmptyStorage("test-dv", "default")
+			cloneStrategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategyHostAssisted)
+			storageProfile := createStorageProfileWithCloneStrategy(scName, []cdiv1.ClaimPropertySet{
+				{AccessModes: accessMode, VolumeMode: &blockMode}}, &cloneStrategy)
+
+			// Prepare the source PVC with the required annotations
+			pvc := createPvcInStorageClass("test", metav1.NamespaceDefault, &scName, nil, nil, corev1.ClaimBound)
+			pvc.SetAnnotations(make(map[string]string))
+			pvc.GetAnnotations()[AnnVirtualImageSize] = "100" // Mock value
```
brybacki, line 1722: Are we sure the `AnnVirtualImageSize` is always up to date? Is this AnnVirtualImageSize invalidated on updates or maybe compared to some other sizes/versions during read?

brybacki: ok, I have found it, I can see you double check by comparing the original Capacity with Current Capacity

brybacki: Is there a test for that scenario? It is pretty interesting scenario.

alromeros: There isn't, though I agree that it's an interesting case. I add it :+1:

pkg/controller/datavolume-controller.go
```diff
@@ -2722,3 +2777,292 @@ func getDeltaTTL(dv *cdiv1.DataVolume, ttl int32) time.Duration {
...
+		getVolumeMode(sourcePvc) == corev1.PersistentVolumeFilesystem &&
+		GetContentType(sourcePvc) == string(cdiv1.DataVolumeKubeVirt) {
+		// If available, we first try to get the virtual size from previous iterations
+		imgSize, available := getSizeFromAnnotations(sourcePvc)
+		if !available {
+			imgSize, err = r.getSizeFromPod(sourcePvc, dv)
+			if err != nil {
+				return err
+			} else if imgSize == 0 {
+				return nil
```
brybacki, line 2809: Do I understand correctly that when the pod starts, but did not complete its job the size is `zero` and then controller continues with the clone and fails when trying to `createPvcForDatavolume`. After that it will retry.

alromeros: Yeah, that's right. Maybe I could add a return flag again to `detectCloneSize` so, if the pod is not finished, we reconcile without error.

pkg/controller/clone-controller.go
alromeros: The size detection pod gets the virtual size of the image once it's been resized to match the source PVC - overhead. The problem is that, even after overhead calculations are applied, this image size is usually slightly smaller than the one defined in the source PVC's specs due to the round down performed in this function: https://github.com/kubevirt/containerized-data-importer/blob/6366ef397f2fb5824008ae94a9417bca018c4830/pkg/util/util.go#L453-L459

This is hard to explain here so I'll try using an example: If I create a source PVC with "500Mi" as size and with 5'5% overhead, the size-detection pod doesn't obtain **495452160** bytes as one should've expected, but **494927872** due to the round-down performed in the function below. So, even after reversing this calculation back as we do in the datavolume-controller, the obtained size ends up being smaller than the one in the source PVC's spec. This is why we need a mechanism to enable the clone anyway.

I first tried rounding up the value again for the size comparison, but this behavior happens also if the target's overhead ends up being smaller than the one used in the source, so that's why I used that annotation instead.

brybacki: Ok, that is great explanation, now I get it. 

But it means we might need to improve this or maybe the actual clone with fsOverhead logic, so it is simpler and more predictable. I would like a new issue here on github to "investigate" this case and maybe find ways to simplify/refactor  with your example. So we can act on it in near future.

pkg/controller/datavolume-controller.go
brybacki: if that does not make code much more complex I would like an explicit error here.

alromeros: When `imgSize` is zero? I had first an error there, but since it's really an expected behavior, @awels recommended me to just return `nil` instead. Every time that happens, we record an event to the dataVolume in `getSizeFromPod`.

However, if for any reason it's not recommended to attempt the clone with `zero` size as I currently do, I could also return a flag in `detectCloneSize` alongside the error to reconcile in `reconcileClone` until the size has been obtained.

Something like:

```go
func (r *DatavolumeReconciler) reconcileClone(

         ...

        // Done would be false if imgSize is 0
	done, err := r.detectCloneSize(datavolume, pvcSpec, selectedCloneStrategy)
		if err != nil {
		} else if !done {
                }
	}
````

File: tests/cloner_test.go [hunk 51]
```diff
@@ -575,6 +577,102 @@ var _ = Describe("all clone tests", func() {
...
+
+					By("Deleting verifier pod")
+					err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+					Expect(err).ToNot(HaveOccurred())
+
+					By("Checksum comparison")
+					Expect(sourceMD5).To(Equal(targetMD5))
+				},
+				Entry("Block to block (empty storage size)", v1.PersistentVolumeBlock, v1.PersistentVolumeBlock),
+				Entry("Block to filesystem (empty storage size)", v1.PersistentVolumeBlock, v1.PersistentVolumeFilesystem),
```
awels, line 673: Aren't you skipping this combination on line 606?

awels (commented):
Looks good, just a few questions and comments.

tests/cloner_test.go
alromeros: Yeah, maybe that wasn't too clear. I included it as an `Entry` because that should be a perfectly valid case to test, but since we have an unrelated bug that keeps the case from working, I added a provisional `Skip` until we fix it.

pkg/controller/datavolume-controller.go
alromeros: In that case, I think it's more clear to add a new return value to `detectCloneSize`, as shown above. I'll quickly fix that :+1:

tests/cloner_test.go
awels: Oh okay what bug is that? Is that where the sizes don't line up?

tests/cloner_test.go
```diff
@@ -640,6 +738,260 @@ var _ = Describe("all clone tests", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				f.ForceBindPvcIfDvIsWaitForFirstConsumer(dataVolume)
+				sourcePvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(dataVolume.Namespace).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Wait for source DV Succeeded phase")
+				err = utils.WaitForDataVolumePhaseWithTimeout(f, f.Namespace.Name, cdiv1.Succeeded, dataVolumeName, cloneCompleteTimeout)
+				Expect(err).ToNot(HaveOccurred())
+
+				wg.Add(1)
```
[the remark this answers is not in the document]
awels: Right but doing the thread doesn't really much either, because it is entirely possible that the pod is created, finishes and is removed inside the 1 second refresh of the eventually, so there is no guarantee that your thread will catch the pod either. It might make sense to respect the `AnnPodRetainAfterCompletion` on the datavolume. Basically if that annotation is on the DV, then any pods related to it being populated should not get deleted after they complete. Then you have plenty of time to inspect the pod and see if it did what you wanted.

tests/cloner_test.go
alromeros: This one: https://bugzilla.redhat.com/show_bug.cgi?id=2079781 (I think @maya-r is working on it)

tests/cloner_test.go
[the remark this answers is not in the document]
alromeros: You are right, I somehow missed that. I'll try that annotation, thanks!

tests/cloner_test.go
alromeros: Nevermind, I just saw that that bug is for the opposite case, my fault

alromeros: This is the bug I was talking about: https://bugzilla.redhat.com/show_bug.cgi?id=2033191. It says that it happens when using PVC API, but it's also happening to me with Storage. I think @maya-r is working on it.

alromeros: Actually, the description of the bug implies that my case could be different, so I don't know if I'm missing something. In any case, the bug that I'm having can also be reproducible in this previously written [test](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/cloner_test.go#L425), when changing the clone size so it matches the original one:

https://github.com/kubevirt/containerized-data-importer/blob/8db48708fc30a757cb86855bda05bda08eaa8ba0/tests/cloner_test.go#L440

alromeros: I might be flooding here, but after adding `ReadWriteOnce` to the source PVC, the case works. Since I was using Storage API in both source and target I didn't know I needed something like this, so I guess it's my fault.

pkg/controller/datavolume-controller_test.go
brybacki: Is it already added? I cannot find it

brybacki:
good, I cannot find the new test but I think it is ready to go
/lgtm

But it is a big PR so it would be good to have another reviewer do the last check. 
@awels WDYT?

pkg/controller/datavolume-controller_test.go
alromeros: @brybacki yeah, this one: https://github.com/kubevirt/containerized-data-importer/blob/19c4ba477b5260bf9c75b97a4f17ec55e3304215/tests/cloner_test.go#L899

Notice that I've mocked the behavior instead of resizing the source PVC (I hope that's enough)

brybacki: I was looking at unit tests ;)

pkg/controller/datavolume-controller.go
```diff
@@ -2722,3 +2815,294 @@ func getDeltaTTL(dv *cdiv1.DataVolume, ttl int32) time.Duration {
...
+		if sourceSize.CmpInt64(imgSize) == 1 {
+			dv.Annotations[AnnPermissiveClone] = "true"
+		}
+		// If needed, inflate size with filesystem overhead
+		targetSize, err = inflateSizeWithOverhead(r.client, imgSize, pvcSpec)
+		if err != nil {
+			return false, err
+		}
+	} else {
+		targetSize = *sourceSize
```
brybacki, line 2861: here we should also inflate with correct overhead if the target is filesystem!

brybacki (changes requested):
I think that in detectCloneSize, the branch when source is block needs additional logic 
```
		targetSize, err = inflateSizeWithOverhead(r.client, imgSize, pvcSpec)
```

pkg/controller/datavolume-controller.go
```diff
@@ -916,6 +975,40 @@ func cloneStrategyToCloneType(selectedCloneStrategy cloneStrategy) string {
 	return ""
 }
 
+// reconcileSizelessClone handles the reconciling process if the size-detection mechanism is not able to obtain the img size
+func (r *DatavolumeReconciler) reconcileSizelessClone(
```
awels, line 979: When seeing this function this was screaming, use me in all 3 variations.
The csi clone, the snapshot clone, and the no target size clone. I would call it something like `isSourceReadyToClone` that returns a boolean and error. The error is to catch the normal errors like now. When calling this function I would do it something like this for the csi clone and smart clone options.
```go
if readyToClone, err := r.isSourceReadyToClone(dv, cloneStrategy); err != nil {
} else if !readyToClone {
  return reconcile.Result{Requeue: true},
    r.updateCloneStatusPhase(cdiv1.CloneScheduled, datavolume, nil, selectedCloneStrategy)
}
```
For the size less clone option something like this
```go
else if !done {
  if readyToClone, err := r.isSourceReadyToClone(dv, cloneStrategy); err != nil {
  } else if !readyToClone {
    return reconcile.Result{Requeue: true},
      r.updateCloneStatusPhase(cdiv1.CloneScheduled, datavolume, nil, cloneStrategy)
  }
}
```

tests/cloner_test.go
awels: So if you are skipping due to that bug, I would remove the manual skip code above and make this a PEntry (which will skip the entry).

pkg/controller/datavolume-controller.go
alromeros: You are right, I'll add a new commit with the requested changes. Thanks!

awels (approved):

brybacki (approved):

awels (commented):
/approve

PR #2223, opened 2022-04-04
Retry adding ticket on imageio container startup

akalenyu:

Sometimes we see that ticket is not there (thx awels), we might want to
try a few times before giving up.
With `&&` if the add-ticket command fails, the container will keep restarting.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/manifests/templates/imageio.yaml.in
+++ b/manifests/templates/imageio.yaml.in
@@ -40,7 +40,11 @@ spec:
         - name: "certs"
           mountPath: "/tmp"
         command: ["/bin/bash"]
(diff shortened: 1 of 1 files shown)


akalenyu:
/test all

warm import failure again (non-related)

awels:
I have noticed with an increased CI load, there is an increase in the warm migration failure. We had disabled one of them for a while, and I thought I had fixed the dead lock, but there is another race.

akalenyu:
@awels We never backported this, should we?

akalenyu:
> Is there a need? are we getting failures?

Think I am seeing this, yes, it should be possible to hit this in all releases but I'd start with 4.10

PR #2226, opened 2022-04-05
Detect storage capabilities for no-provisioner storage classes

maya-r:
Assume there's a persistent volume that we can look up to infer the
correct values for volume mode and access modes.

Move storagecapabilities code into controller as a path of lower
resistance to using "resolveVolumeMode"

As a side-note, this was aiming to detect the storage configuration of [local storage operator](https://github.com/openshift/local-storage-operator).
Looking up the configmap of this operator felt like it might be a layering violation, so I looked for another way.
Looking up persistent volumes actually works for any no-provisioner storage class, which is even better!

**Release note**:

```release-note
Detect storage capabilities for no-provisioner storage classes by looking up persistent volumes
```

### DIFF
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -141,7 +141,7 @@ func (r *StorageProfileReconciler) getStorageProfile(sc *storagev1.StorageClass)
 
 func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageClass) []cdiv1.ClaimPropertySet {
 	claimPropertySets := []cdiv1.ClaimPropertySet{}
-	capabilities, found := storagecapabilities.Get(sc)
+	capabilities, found := storagecapabilities.Get(r.client, sc)
 	if found {
 		for i := range capabilities {
 			claimPropertySet := cdiv1.ClaimPropertySet{
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -37,7 +37,7 @@ var (
 			Name: monitoring.MetricOptsList[monitoring.IncompleteProfile].Name,
 			Help: monitoring.MetricOptsList[monitoring.IncompleteProfile].Help,
 		})
-	scProvisionerMap = make(map[string]string)
+	globalClient client.Client
 )
 
 // StorageProfileReconciler members
@@ -74,7 +74,6 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
 
 func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.StorageClass) (reconcile.Result, error) {
 	log := r.log.WithValues("StorageProfile", sc.Name)
-	addStorageClass(sc)
 
 	storageProfile, prevStorageProfile, err := r.getStorageProfile(sc)
 	if err != nil {
@@ -190,7 +189,6 @@ func (r *StorageProfileReconciler) createEmptyStorageProfile(sc *storagev1.Stora
 
 func (r *StorageProfileReconciler) deleteStorageProfile(name string, log logr.Logger) error {
 	log.Info("Cleaning up StorageProfile that corresponds to deleted StorageClass", "StorageClass.Name", name)
-	removeStorageClass(name)
 	storageProfileObj := &cdiv1.StorageProfile{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
@@ -204,23 +202,12 @@ func (r *StorageProfileReconciler) deleteStorageProfile(name string, log logr.Lo
 	return r.checkIncompleteProfiles()
 }
 
-func removeStorageClass(name string) {
-	_, found := scProvisionerMap[name]
-	if found {
-		delete(scProvisionerMap, name)
-	}
-}
-
-func addStorageClass(sc *storagev1.StorageClass) {
-	scProvisionerMap[sc.Name] = sc.Provisioner
-}
-
 func isNoProvisioner(name string) bool {
-	provisioner, found := scProvisionerMap[name]
-	if found && provisioner == "kubernetes.io/no-provisioner" {
-		return true
+	storageClass := &storagev1.StorageClass{}
+	if err := globalClient.Get(context.TODO(), types.NamespacedName{Name: name}, storageClass); err != nil {
+		return false
 	}
-	return false
+	return storageClass.Provisioner == "kubernetes.io/no-provisioner"
 }
 
 func (r *StorageProfileReconciler) checkIncompleteProfiles() error {
@@ -303,6 +290,7 @@ func addStorageProfileControllerWatches(mgr manager.Manager, c controller.Contro
 	if err := c.Watch(&source.Kind{Type: &cdiv1.StorageProfile{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return err
 	}
+	globalClient = mgr.GetClient()
 	if err := c.Watch(&source.Kind{Type: &v1.PersistentVolume{}}, handler.EnqueueRequestsFromMapFunc(
 		func(obj client.Object) []reconcile.Request {
 			return []reconcile.Request{{
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -15,8 +15,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
@@ -35,6 +37,7 @@ var (
 			Name: monitoring.MetricOptsList[monitoring.IncompleteProfile].Name,
 			Help: monitoring.MetricOptsList[monitoring.IncompleteProfile].Help,
 		})
+	scProvisionerMap = make(map[string]string)
 )
 
 // StorageProfileReconciler members
@@ -71,6 +74,7 @@ func (r *StorageProfileReconciler) Reconcile(_ context.Context, req reconcile.Re
 
 func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.StorageClass) (reconcile.Result, error) {
 	log := r.log.WithValues("StorageProfile", sc.Name)
+	addStorageClass(sc)
 
 	storageProfile, prevStorageProfile, err := r.getStorageProfile(sc)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/controller/util_test.go
```diff
@@ -525,6 +525,25 @@ func createPvcInStorageClass(name, ns string, storageClassName *string, annotati
 	return pvc
 }
 
+func CreatePv(name string, storageClassName string) *v1.PersistentVolume {
+	volumeMode := v1.PersistentVolumeFilesystem
+	pvc := &v1.PersistentVolume{
```
akalenyu, line 530: you mean pv instead of pvc?

pkg/controller/storageprofile-controller.go
```diff
@@ -140,7 +139,7 @@ func (r *StorageProfileReconciler) getStorageProfile(sc *storagev1.StorageClass)
 
 func (r *StorageProfileReconciler) reconcilePropertySets(sc *storagev1.StorageClass) []cdiv1.ClaimPropertySet {
 	claimPropertySets := []cdiv1.ClaimPropertySet{}
-	capabilities, found := storagecapabilities.Get(sc)
+	capabilities, found := Get(r.client, sc)
```
akalenyu, line 142: should we maybe rename the func to something more specific like GetStorageCapabilities? kind of weird to just say Get() here

pkg/controller/util_test.go
maya-r: Yes, thanks.

pkg/controller/storagecapabilities.go
```diff
@@ -53,13 +56,49 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"csi.trident.netapp.io/ontap-san": {{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock}},
 }
 
-// Get finds and returns a predefined StorageCapabilities for a given StorageClass
-func Get(sc *storagev1.StorageClass) ([]StorageCapabilities, bool) {
+// GetStorageCapabilities finds and returns a predefined StorageCapabilities for a given StorageClass
+func GetStorageCapabilities(cl client.Client, sc *storagev1.StorageClass) ([]StorageCapabilities, bool) {
 	provisionerKey := storageProvisionerKey(sc)
+	if provisionerKey == "kubernetes.io/no-provisioner" {
```
akalenyu, line 62: As part of another task, I am trying to find a place of truth in the k8s libraries for this and `storageclass.kubernetes.io/is-default-class`. So far, no luck.
Doesn't block this PR, just thought you might have an idea

pkg/controller/storagecapabilities.go
```diff
@@ -53,13 +56,49 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
...
+		return []StorageCapabilities{}, false
+	}
+	capabilities := []StorageCapabilities{}
+	for _, pv := range pvs.Items {
+		if pv.Spec.StorageClassName == sc.Name {
+			for _, accessMode := range pv.Spec.AccessModes {
+				capabilities = append(capabilities, StorageCapabilities{
+					AccessMode: accessMode,
+					VolumeMode: resolveVolumeMode(pv.Spec.VolumeMode),
+				})
```
akalenyu, line 82: Nice, we pair each available accessMode with the volumeMode. That is how it looks like for ceph as well:
```bash
  - accessModes:
    - ReadWriteMany
    volumeMode: Block
  - accessModes:
    - ReadWriteOnce
    volumeMode: Block
```

pkg/controller/storagecapabilities.go
maya-r: It looks like it comes from [this package](https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/storage/util)

pkg/controller/storagecapabilities.go
```diff
@@ -53,13 +56,49 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
...
+	if err != nil {
+		return []StorageCapabilities{}, false
+	}
+	capabilities := []StorageCapabilities{}
+	for _, pv := range pvs.Items {
+		if pv.Spec.StorageClassName == sc.Name {
+			for _, accessMode := range pv.Spec.AccessModes {
+				capabilities = append(capabilities, StorageCapabilities{
+					AccessMode: accessMode,
+					VolumeMode: resolveVolumeMode(pv.Spec.VolumeMode),
```
brybacki, line 81: I get what you mean by:  _Move storagecapabilities code into controller as a path of lower resistance to using "resolveVolumeMode"_

I suppose long term solution might be to extract all the tools we use to access volumes (PVC, PV) and put them in separate volume package.

pkg/controller/storagecapabilities.go
```diff
@@ -53,13 +56,49 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
...
 }
 
+func capabilitiesForNoProvisioner(cl client.Client, sc *storagev1.StorageClass) ([]StorageCapabilities, bool) {
+	pvs := &v1.PersistentVolumeList{}
+	err := cl.List(context.TODO(), pvs)
+	if err != nil {
+		return []StorageCapabilities{}, false
+	}
+	capabilities := []StorageCapabilities{}
+	for _, pv := range pvs.Items {
```
brybacki, line 76: We need a pvc, and then we can provide the self-configuring magic. I am not sure what think about this.  But I think there is no other easy way to detect capabilities.

brybacki: Another question. Do we need to scan all PVs, is there any shortcut possible?

pkg/controller/storagecapabilities.go
```diff
@@ -53,13 +56,49 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"csi.trident.netapp.io/ontap-san": {{AccessMode: v1.ReadWriteOnce, VolumeMode: v1.PersistentVolumeBlock}},
 }
 
-// Get finds and returns a predefined StorageCapabilities for a given StorageClass
-func Get(sc *storagev1.StorageClass) ([]StorageCapabilities, bool) {
+// GetStorageCapabilities finds and returns a predefined StorageCapabilities for a given StorageClass
+func GetStorageCapabilities(cl client.Client, sc *storagev1.StorageClass) ([]StorageCapabilities, bool) {
 	provisionerKey := storageProvisionerKey(sc)
+	if provisionerKey == "kubernetes.io/no-provisioner" {
+		return capabilitiesForNoProvisioner(cl, sc)
```
brybacki, line 63: So here we do a full scan of all PVs (filtering by StorageClassName), and we do it each time when GetStorageCapabilities is called, which is on every profile reconcile loop. Can we cache some data? would it be safe to cache it?

maya-r: Given that there's a field called "uncachedClient" and I am using the client I assumed it's cached.
@mhenriks - do you know if this is problematic to query on a larger cluster?

pkg/controller/storagecapabilities.go
maya-r: Worth noting that no-provisioner storage class implies nothing provisions these PVs automatically, you have to create PVs ahead of time to use them, so if they don't exist the PVCs are going to remain in pending state forever (even if correctly defined).

brybacki: Right!

pkg/controller/storagecapabilities.go
brybacki: Right, so maybe that is not problematic at all.

maya-r:
I added a new change where the detection is only done for local storage operator, so we are not extending the scope of things that we now provide storage profiles for.
This is similar to how we list only known storage classes.

pkg/controller/storageprofile-controller.go
```diff
@@ -35,6 +37,7 @@ var (
 			Name: monitoring.MetricOptsList[monitoring.IncompleteProfile].Name,
 			Help: monitoring.MetricOptsList[monitoring.IncompleteProfile].Help,
 		})
+	scProvisionerMap = make(map[string]string)
```
akalenyu, line 40: this map will reconstruct properly if the controller dies right? just making sure since storing these things in pods is usually not what we do

maya-r: It is pretty unusual, I am doing it to filter the PV changes that trigger the reconcile, and I am under the impression it should run very fast.
I am not 100% sure what happens when the controller dies, but as long as it runs once for the correct storage class (and it probably does) it'll build the correct storage capabilities. I don't mind the missed PV updates.

awels (commented):
I was concerned this would be run on every PV create/delete/update, but it is only on no provisioner pv create/delete/update (might even decide to not do this for delete/update). Which will be a manual type operation so won't happen a lot. So I am okay with this, one comment that should make things a little simpler.

pkg/controller/storageprofile-controller.go
```diff
@@ -199,6 +204,25 @@ func (r *StorageProfileReconciler) deleteStorageProfile(name string, log logr.Lo
...
+	}
+}
+
+func addStorageClass(sc *storagev1.StorageClass) {
+	scProvisionerMap[sc.Name] = sc.Provisioner
+}
+
+func isNoProvisioner(name string) bool {
+	provisioner, found := scProvisionerMap[name]
+	if found && provisioner == "kubernetes.io/no-provisioner" {
```
aglitke, line 220: Should we use a constant for the no-provisioner string?

aglitke (commented):
This approach seems pretty simple and reasonable to me.  As long as we are confident about the caching so that the increased watching of PVs will not cause a performance problem.

pkg/controller/storageprofile-controller.go
akalenyu: This ties closely to another discussion on this PR:
https://github.com/kubevirt/containerized-data-importer/pull/2226#discussion_r843077393

How expensive would it be to bring in https://pkg.go.dev/k8s.io/kubernetes/pkg/controller/volume/persistentvolume/util?

maya-r: if I try to do that, I'm going to end up calling it NotSupportedProvisioner which is kinda confusing in our context, because it is supported. Perhaps it's better to leave it as-is.

pkg/controller/storageprofile-controller.go
maya-r: Done

pkg/controller/storageprofile-controller.go
```diff
@@ -279,9 +290,28 @@ func addStorageProfileControllerWatches(mgr manager.Manager, c controller.Contro
...
+	globalClient = mgr.GetClient()
+	if err := c.Watch(&source.Kind{Type: &v1.PersistentVolume{}}, handler.EnqueueRequestsFromMapFunc(
+		func(obj client.Object) []reconcile.Request {
+			return []reconcile.Request{{
+				NamespacedName: types.NamespacedName{Name: scName(obj)},
+			}}
+		},
+	),
+		predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool { return isNoProvisioner(scName(e.Object)) },
```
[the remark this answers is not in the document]
maya-r: Done, thanks!

pkg/controller/storageprofile-controller.go
```diff
@@ -199,6 +202,14 @@ func (r *StorageProfileReconciler) deleteStorageProfile(name string, log logr.Lo
 	return r.checkIncompleteProfiles()
 }
 
+func isNoProvisioner(name string) bool {
```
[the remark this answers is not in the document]
maya-r: Done, thanks!

awels (commented):
/lgtm
/approve

PR #2232, opened 2022-04-10
DataVolume automatic PVC size detection

arnongilboa:

Allow creating a DV with no PVC size, or no Storage/PVC defined at all.

Some folks are bothered by specifying PVC size or even Storage/PVC in
import DV spec, and want CDI to do that for them.

When no size is defined, we first create a zero size PVC, so the import
fails, and based on the error reported virtual image size we recreate
the PVC (todo: resize if possible).

Fixes #

**Release note**:
```release-note
DataVolume automatic PVC size detection
```


pkg/controller/util.go
```diff
@@ -569,6 +569,10 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *v1.Pod, prefix
 			if strings.Contains(containerState.Terminated.Message, common.PreallocationApplied) {
 				anno[AnnPreallocationApplied] = "true"
 			}
+			//FIXME: nicify
```
brybacki, line 572: What do you mean by nicify?
I think it would be better to make message format and parsing more obvious? With some common code used to make an implicit contract between the place where we create the message, and here where we parse the message

pkg/controller/util.go
```diff
@@ -933,54 +941,106 @@ func pvcFromStorage(client client.Client, recorder record.EventRecorder, log log
...
-	pvcSpec.Resources.Requests[v1.ResourceStorage] = *requestedVolumeSize
-
-	return pvcSpec, nil
+	if storageHint := dv.Annotations[annResourceStorageHint]; storageHint != "" {
+		size, err := strconv.Atoi(storageHint)
+		if err != nil {
+			return err
+		}
+		//FIXME: configurable factor/delta for reservation
+		storageResizeFactor := 1.25
```
brybacki, line 1037: I do not understand this, why not just use 1 and leave the size inflation to fsOverhead logic?

pkg/controller/util.go
arnongilboa: Agree, that what I meant. Here it was a quick one for POC.

pkg/controller/util.go
brybacki: So why not just use 1. User wants to import and have usable OS. If I am not mistaken - virtual images already have free space in it. If not, the user can always resize. 

So I think - if you want to have a simple and quick import you just do it with factor = 1. If you want some advanced logic then you do this logic in import/migrate tools, and the tool can set correct Request Size.  

So I do not know, I would like to see a real scenario - a real case why this might be needed.

arnongilboa: Convinced, fixed, thanks!

pkg/controller/util.go
```diff
@@ -902,16 +906,20 @@ func addAnnotation(obj metav1.Object, key, value string) {
...
+		pvcSpec, err = pvcFromStorage(client, recorder, log, dv)
+	} else {
+		pvcSpec, err = pvcDefault(client, recorder, log, dv)
 	}
-
-	if dv.Spec.Storage != nil {
-		return pvcFromStorage(client, recorder, log, dv)
+	if pvcSpec != nil {
+		if _, found := pvcSpec.Resources.Requests[v1.ResourceStorage]; !found {
+			err = setStorageByHint(dv, pvcSpec)
```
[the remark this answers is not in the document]
arnongilboa: fixed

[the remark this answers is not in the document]
arnongilboa: fixed

awels (commented):
So I am a little confused, and I think you are as well. The problem we are solving is determining what the size of the PVC is going to be in the data volume controller when the DV itself has no size information. This all happens before we create the target PVC based on some size information of the SOURCE PVC. At this point the target PVC does NOT exist, and thus the DV cannot be running or anything like that. So we don't have to look up the DV condition. So the steps roughly will be:

1. Check if we have size information in the DV, if not, and the spec.Source.PVC != nil (aka we know the name and namespace of the source) look up the actual source PVC
2. Once we have the source PVC, see if we have storage hint annotation (to get the right size). If so, use that value as the size of the TARGET PVC.
3. If that is not there, create a pod in the source namespace that uses the source PVC and run the qemu-img info command on there.
4. Remember that the pod starting and running and finishing can take a while, so the DV controller will have to watch for that pod completing (I will show you how). And when it completes you can read the exit message.

arnongilboa:
> So I am a little confused, and I think you are as well. The problem we are solving is determining what the size of the PVC is going to be in the data volume controller when the DV itself has no size information. This all happens before we create the target PVC based on some size information of the SOURCE PVC. At this point the target PVC does NOT exist, and thus the DV cannot be running or anything like that. So we don't have to look up the DV condition. So the steps roughly will be:
> 
> 1. Check if we have size information in the DV, if not, and the spec.Source.PVC != nil (aka we know the name and namespace of the source) look up the actual source PVC
> 2. Once we have the source PVC, see if we have storage hint annotation (to get the right size). If so, use that value as the size of the TARGET PVC.
> 3. If that is not there, create a pod in the source namespace that uses the source PVC and run the qemu-img info command on there.
> 4. Remember that the pod starting and running and finishing can take a while, so the DV controller will have to watch for that pod completing (I will show you how). And when it completes you can read the exit message.

@awels I'm not trying to solve the problem you mentioned (clone) which already has wip by @alromeros , but only the import flow.

awels:
Oh wait I confused the PRs, ignore me.

tests/import_test.go
```diff
@@ -182,6 +184,39 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+		dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Verify pvc was created")
+		pvc, err := utils.WaitForPVC(f.K8sClient, dataVolume.Namespace, dataVolume.Name)
+		Expect(err).ToNot(HaveOccurred())
+		f.ForceBindIfWaitForFirstConsumer(pvc)
+
+		By("Wait for import to be completed")
+		err = utils.WaitForDataVolumePhase(f.CdiClient, dataVolume.Namespace, cdiv1.Succeeded, dataVolume.Name)
```
akalenyu, line 207: I sound like a broken record already :rofl: but maybe verify content with `By("Verify content")`

arnongilboa:
@alromeros  @aglitke 
Closing this PR, which will be better rewritten using common code introduced in PR #2222 once it's merged, and using the cdi-image-size-detection containerized app instead of the current approach which is a bit dirty (zero size PVC import fail and resize).

PR #2233, opened 2022-04-10
Garbage Collect Completed DVs

arnongilboa:

See design [here](https://github.com/kubevirt/community/blob/main/design-proposals/garbage-collect-completed-dvs.md).

ToDos:
- [ ] DataImportCron and DataSource controllers adaptation and func tests
- [ ] Add doc for DataVolume, CDIConfig and DataImportCron changes
- [ ] Extend unit tests and functional tests
- [ ] KubeVirt adaptation

Fixes #

**Release note**:
```release-note
Garbage Collect Completed DataVolumes
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-mutate.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate.go
@@ -101,8 +101,11 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 	modifiedDataVolume := dataVolume.DeepCopy()
 	modified := false
 
-	config, _ := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
-	if config != nil && config.Spec.DataVolumeTTLSeconds != nil &&
+	config, err := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
+	if err != nil {
+		return toAdmissionResponseError(err)
+	}
+	if config.Spec.DataVolumeTTLSeconds != nil &&
 		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
 		if modifiedDataVolume.Annotations == nil {
 			modifiedDataVolume.Annotations = make(map[string]string)
--- a/pkg/apiserver/webhooks/datavolume-mutate.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate.go
@@ -33,6 +33,7 @@ import (
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	cdiclient "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"kubevirt.io/containerized-data-importer/pkg/clone"
+	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
 	"kubevirt.io/containerized-data-importer/pkg/token"
 )
@@ -97,8 +98,24 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		targetName = ar.Request.Name
 	}
 
+	modifiedDataVolume := dataVolume.DeepCopy()
+	modified := false
+
+	config, _ := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
+	if config != nil && config.Spec.DataVolumeTTLSeconds != nil &&
+		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
+		if modifiedDataVolume.Annotations == nil {
+			modifiedDataVolume.Annotations = make(map[string]string)
+		}
+		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] = "true"
+		modified = true
+	}
+
 	if pvcSource == nil {
 		klog.V(3).Infof("DataVolume %s/%s not cloning", targetNamespace, targetName)
+		if modified {
+			return toPatchResponse(dataVolume, modifiedDataVolume)
+		}
 		return allowedAdmissionResponse()
 	}
 
@@ -151,11 +168,9 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		return toAdmissionResponseError(err)
 	}
 
-	modifiedDataVolume := dataVolume.DeepCopy()
 	if modifiedDataVolume.Annotations == nil {
 		modifiedDataVolume.Annotations = make(map[string]string)
 	}
-
 	modifiedDataVolume.Annotations[controller.AnnCloneToken] = token
 
 	klog.V(3).Infof("Sending patch response...")
--- a/pkg/apiserver/webhooks/datavolume-mutate_test.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate_test.go
@@ -131,7 +131,7 @@ var _ = Describe("Mutating DataVolume Webhook", func() {
 				},
 			}
 
-			resp := mutateDVsEx(key, ar, true, []runtime.Object{dataSource})
+			resp := mutateDVsEx(key, ar, true, nil, []runtime.Object{dataSource})
 			Expect(resp.Allowed).To(BeTrue())
 			Expect(resp.Patch).ToNot(BeNil())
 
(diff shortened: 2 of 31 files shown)


pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -97,8 +98,24 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		targetName = ar.Request.Name
 	}
 
+	modifiedDataVolume := dataVolume.DeepCopy()
+	modified := false
+
+	config, _ := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
```
akalenyu, line 104: Maybe return an error here like in DataSource's GET above?
https://github.com/kubevirt/containerized-data-importer/blob/9a4061c997220d1ae707006d76a079a7450ff330/pkg/apiserver/webhooks/datavolume-mutate.go#L85-L88

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -97,8 +98,24 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
+	if config != nil && config.Spec.DataVolumeTTLSeconds != nil &&
+		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
```
akalenyu, line 106: `.Annotations` can be nil here/not have that key in it, no?

pkg/controller/datavolume-controller.go
```diff
@@ -2565,3 +2578,81 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
 func newLongTermCloneTokenGenerator(key *rsa.PrivateKey) token.Generator {
 	return token.NewGenerator(common.ExtendedCloneTokenIssuer, key, 10*365*24*time.Hour)
 }
+
+func (r *DatavolumeReconciler) garbageCollect(dataVolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) (res *reconcile.Result, err error) {
+	if dataVolume.Status.Phase != cdiv1.Succeeded {
+		return
+	}
+	cdiConfig := &cdiv1.CDIConfig{}
+	if err = r.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, cdiConfig); err != nil {
```
akalenyu, line 2587: It looks like an innocent and cheap GET so I am not sure, but, maybe this belongs as an env var on the cdi-deployment?

pkg/controller/datavolume-controller.go
```diff
@@ -2565,3 +2578,81 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+	if delta := getDeltaTTL(dataVolume, *dvTTL); delta > 0 {
+		return &reconcile.Result{RequeueAfter: delta}, nil
+	}
+
+	if err = r.detachPvcDeleteDv(pvc, dataVolume); err != nil {
+		return
+	}
+
+	// FIXME: not sure this part is necessary
+	// Lookup for other DVs to garbage collect
```
akalenyu, line 2605: Yeah same here, I am not sure if other DVs won't be queued up for their own loop and just get GC'd there

pkg/controller/datavolume-controller.go
```diff
@@ -2565,3 +2578,81 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+		}
+	}
+	return &reconcile.Result{}, nil
+}
+
+func (r *DatavolumeReconciler) detachPvcDeleteDv(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) error {
+	dvDelete := dv.Annotations[AnnDeleteAfterCompletion]
+	if IsSucceeded(pvc) && dvDelete == "true" {
+		updatePvcOwnerRefs(pvc, dv)
+		delete(pvc.Annotations, AnnPopulatedFor)
```
akalenyu, line 2630: Can there be more annotations we want to get rid of when detaching? `AnnPrePopulated`?

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -742,6 +742,9 @@ type CDIConfigSpec struct {
 	Preallocation *bool `json:"preallocation,omitempty"`
 	// InsecureRegistries is a list of TLS disabled registries
 	InsecureRegistries []string `json:"insecureRegistries,omitempty"`
+	// dataVolumeTTLSeconds is the time in seconds after DataVolume completion it can be garbage collected.
+	// +optional
+	DataVolumeTTLSeconds *int32 `json:"dataVolumeTTLSeconds,omitempty"`
```
akalenyu, line 747: Nice, similar to the TTL field in `Job`

tests/import_test.go
```diff
@@ -184,6 +186,110 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+		Expect(err).ToNot(HaveOccurred())
+		f.ForceBindIfWaitForFirstConsumer(pvc)
+
+		By("Wait for DV to be garbage collected")
+		Eventually(func() bool {
+			_, err = f.CdiClient.CdiV1beta1().DataVolumes(ns).Get(context.TODO(), dv.Name, metav1.GetOptions{})
+			return k8serrors.IsNotFound(err)
+		}, timeout, pollingInterval).Should(BeTrue())
+
+		By("Verify PVC still exists")
```
akalenyu, line 249: WDYT about verifying content in your tests as well similarly to `By("Verify content")`
throughout the importer suite

tests/import_test.go
```diff
@@ -184,6 +186,110 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 
 })
 
+var _ = Describe("DataVolume Garbage Collection", func() {
```
akalenyu, line 189: General questions about functional tests.
How aren't most of the tests failing after we started GCing? I have assumed we will need adjustments for this, but it doesn't look like it?

akalenyu (commented):
Nice!

pkg/apiserver/webhooks/datavolume-mutate.go
arnongilboa: Sure will

pkg/apiserver/webhooks/datavolume-mutate.go
arnongilboa: Reading from a nil map is allowed. In both cases it will get empty string, which is good for us so we can add the annotation with true.

pkg/controller/datavolume-controller.go
arnongilboa: Env var to get rid of the Get? so we need to update it whenever CDIConfig DataVolumeTTLSeconds is updated? not sure it worth it.

pkg/controller/datavolume-controller.go
arnongilboa: Yep, I think this part should be removed.

tests/import_test.go
arnongilboa: It's currently disabled by default, and we set it only for those new tests.
However as noted, we will adapt all DV related tests (quite straightforward) and add a CI lane with GC on.

tests/import_test.go
arnongilboa: Good idea

pkg/apiserver/webhooks/datavolume-mutate.go
akalenyu: right

tests/import_test.go
akalenyu: Right, since for other tests the TTL is not set in the CDIConfig, correct?

pkg/apiserver/webhooks/datavolume-mutate_test.go
```diff
@@ -251,7 +252,11 @@ func mutateDVsEx(key *rsa.PrivateKey, ar *admissionv1.AdmissionReview, isAuthori
 		}
 		return true, sar, nil
 	})
-	cdiClient := cdiclientfake.NewSimpleClientset(cdiObjects...)
+
+	cdiConfig := controller.MakeEmptyCDIConfigSpec(common.ConfigName)
+	objs := []runtime.Object{cdiConfig}
+	objs = append(objs, cdiObjects...)
+	cdiClient := cdiclientfake.NewSimpleClientset(objs...)
```
awels, line 259: Can you add a test that verifies the mutation works if the TTL is set? And one that verifies that if the TTL is not set, the annotation is not added.

pkg/controller/datavolume-controller.go
```diff
@@ -2601,3 +2613,59 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+			return err
+		}
+		if err := r.client.Delete(context.TODO(), dv); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func updatePvcOwnerRefs(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) {
```
awels, line 2654: I would call this function something else like removeDVasOwnerFrom and make the first argument an Object this way we can potentially re-use it to remove the owner from other objects as well.

pkg/controller/smart-clone-controller.go
```diff
@@ -356,6 +356,7 @@ func newPvcFromSnapshot(snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *core
 				AnnRunningConditionMessage: cloneComplete,
 				AnnRunningConditionReason:  "Completed",
 				annSmartCloneSnapshot:      key,
+				AnnPodPhase:                string(corev1.PodSucceeded),
```
awels, line 359: Maybe add a comment here that we are artifically adding this pod phase on a smart clone where no pods are involved. My first reaction when seeing this was why are pods involved in the smart clone controller.

After think I understand that you need this to determine if you can GC the DV, seems odd though.

awels (commented):
Looks good some minor comments.

pkg/apiserver/webhooks/datavolume-mutate_test.go
arnongilboa: will sure add

pkg/controller/datavolume-controller.go
arnongilboa: yep, much better name

tests/import_test.go
```diff
@@ -184,6 +186,104 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+		pvc, err := utils.WaitForPVC(f.K8sClient, ns, dv.Name)
+		Expect(err).ToNot(HaveOccurred())
+		f.ForceBindIfWaitForFirstConsumer(pvc)
+
+		By("Wait for import to be completed")
+		err = utils.WaitForDataVolumePhase(f, ns, cdiv1.Succeeded, dv.Name)
+		Expect(err).ToNot(HaveOccurred(), "Datavolume not in phase succeeded in time")
+
+		By("Verify DV is not garbage collected")
+		time.Sleep(time.Second * time.Duration(ttl+1))
```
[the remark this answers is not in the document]
arnongilboa: Nice point. I'll add log and look for it in the test.

pkg/controller/datavolume-controller.go
```diff
@@ -2601,3 +2613,59 @@ func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
...
+		if r.UID == dv.UID {
+			pvc.OwnerReferences = append(refs[:i], refs[i+1:]...)
+			break
+		}
+	}
+	pvc.OwnerReferences = append(pvc.OwnerReferences, dv.OwnerReferences...)
+}
+
+func getDeltaTTL(dv *cdiv1.DataVolume, ttl int32) time.Duration {
+	delta := time.Second * time.Duration(ttl)
```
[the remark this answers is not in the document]
arnongilboa: Right

pkg/controller/smart-clone-controller.go
arnongilboa: Adding comment. The alternative is an additional PVC annotation for all DV flows, which is even more odd...

pkg/controller/datavolume-controller.go
awels: Right, good point.

tests/import_test.go
```diff
@@ -184,6 +186,108 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+		By(fmt.Sprintf("Create new datavolume %s", dv.Name))
+		dv, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, ns, dv)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Verify pvc was created")
+		pvc, err := utils.WaitForPVC(f.K8sClient, ns, dv.Name)
+		Expect(err).ToNot(HaveOccurred())
+		f.ForceBindIfWaitForFirstConsumer(pvc)
+
+		if annDeleteAfterCompletion != "false" {
```
awels, line 240: Instead of having this if/else in here, it might make sense to pass a function into the table. The common information between the two looks like its the datavolume name and namespace. So if you define a 'verifyGCOccured(name, namespace string)` and a `verifyNoGC(name, namespace string)` then you can have that function passed as an argument to to the table, and in the table.Entry pass which function you would like to call to verify. This eliminates the giant if/else in the table and make stuff more readable.

pkg/controller/datavolume-controller.go
brybacki: `AnnPrePopulated` is on DV, `AnnPopulatedFor` is on PVC

tests/import_test.go
arnongilboa: fixed

tests/import_test.go
arnongilboa: done

tests/import_test.go
```diff
@@ -184,6 +186,110 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+			Expect(err).NotTo(HaveOccurred())
+			return log
+		}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(ContainSubstring(matchString))
+		Expect(err).ToNot(HaveOccurred())
+		dv, err := f.CdiClient.CdiV1beta1().DataVolumes(ns).Get(context.TODO(), dvName, metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(dv.Annotations[controller.AnnDeleteAfterCompletion]).To(Equal("false"))
+	}
+
+	DescribeTable("Should", func(annDeleteAfterCompletion string, dvFunc func(dvName string)) {
```
awels, line 255: So dvFunc doesn't really tell me what this function does, maybe name it verifyGCFunc that tells me we are verifying something about garbage collection.

awels (approved):
/approve

PR #2240, opened 2022-04-15
Tighten DataImportCron & DataSource max name length

arnongilboa:
Tighten DataImportCron & DataSource max name length as it's used for labeling and indexing which are limited to length 63, unlike DataVolume and other CR names which are limited to length 253.

Fixes #

**Release note**:
```release-note
NONE
```


awels:
/lgtm
one alternative to consider is to allow longer names, but everywhere you use them to create labels to use the naming function to reduce the length to < max in a predictable way. We do that in several locations already. It basically calculates a hash of the name and uses that.

arnongilboa:
> /lgtm one alternative to consider is to allow longer names, but everywhere you use them to create labels to use the naming function to reduce the length to < max in a predictable way. We do that in several locations already. It basically calculates a hash of the name and uses that.

@awels DIC controller already uses the naming function to reduce the length for DVs and CronJob:
https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/dataimportcron-controller.go#L876-L886

However, if we use longer names here and label with GetLabelNameFromResourceName(), the Watch mapToCron() mapping function will need to List all DICs and lookup the hash, instead of the current O(1) mapping. wdyt?

awels:
Good point, and we can't do a reverse on the hash easily to get the full length name back.

PR #2247, opened 2022-04-26
Build builder for arm64 and amd64, use Go 1.18.1

### LINKED ISSUE #2211: cdi-bazel-builder: exec user process caused "exec format error"
**What happened**:

When i build in aarch64 environment, i exec 'BUILD_ARCH=aarch64 make bazel-build' command, it occurs error:

+ docker run -v kubevirt-cdi-volume:/root:rw,z --security-opt label:disable --rm --entrypoint /entrypoint-bazel.sh quay.io/kubevirt/kubevirt-cdi-bazel-builder:0.0.13 mkdir -p /root/go/src/kubevirt.io/containerized-data-importer/_out
standard_init_linux.go:207: exec user process caused "exec format error"

I can't get quay.io/kubevirt/kubevirt-cdi-bazel-builder:0.0.13 image for aarch64, in addition,  When i exec './hack/build/bazel-build-builder.sh', it  can generate cdi-bazel-builder image for amd64, but not support aarch64

I can build cdi in aarch64 environment

1. clone code
2. exec  'BUILD_ARCH=aarch64 make bazel-build'

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.42.0
- Kubernetes version (use `kubectl version`): k3s version v1.22.7+k3s1 (8432d7f2)
- DV specification: N/A
- OS (e.g. from /etc/os-release): Kylin 4.0.2
- Kernel (e.g. `uname -a`): Linux 4.4.131-20200704.kylin.server-generic aarch64 
- Install tools: N/A
- Others: N/A

maya-r:
We want to sync up with the upcoming k8s 1.24, and that uses Go 1.18.
We want to build an arm64 builder as well (using buildah and qemu-user-static), so native `make cluster-sync` works on arm64.
Make a check for qemu-user-static on Linux, if it's not installed, the error is "exec format error" and it's hard to tell what is missing.

I have tested arm64 `make cluster-sync` with kind and only non-CDI images are problematic (fakeovirt, minio, quay.io/external_storage/local-volume-provisioner).

Part 1 of 2 to fix #2211 (need to update the builder used before that issue can be closed)

**Release note**:

```release-note
Use Go 1.18.1 for the builder
Allow native builds on arm64
```

### DIFF
--- a/hack/build/bazel-build-builder.sh
+++ b/hack/build/bazel-build-builder.sh
@@ -19,6 +19,19 @@ script_dir="$(cd "$(dirname "$0")" && pwd -P)"
 source "${script_dir}"/common.sh
 source "${script_dir}"/config.sh
 
+if [ "$(uname)" = "Linux" ]; then
+     if [ -e /proc/sys/fs/binfmt_misc/qemu-aarch64 ]; then
+         BUILDAH_PLATFORM_FLAG="--platform linux/amd64,linux/arm64"
+     else
+         echo "Expecting qemu-user-static to be installed on the host machine, please install it"
+         exit 1
+     fi
+else
+     echo "Not a linux host, not sure how to build non-native docker containers."
+     echo "Building container for native architecture only."
+     BUILDAH_PLATFORM_FLAG=""
+fi
+
 if ! git diff-index --quiet HEAD~1 hack/build/docker; then
     #Since this only runs during the post-submit job, the PR will have squashed into a single
     #commit and we can use HEAD~1 to compare.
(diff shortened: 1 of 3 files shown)


mhenriks:
Let's see about not requiring buildah and qemu-user-static on the build machine.  Kubevirt does not have that requirement.  Maybe look here to see what they are ddoing?

https://github.com/kubevirt/kubevirt/blob/main/hack/builder/build.sh

maya-r:
The way it works in kubevirt is that they rely on a container multiarch/qemu-user-static running privileged (so they can run a nested docker) and running the build in it.
I dislike avoiding dependencies with outside docker containers, but I went to list several reasons we might not want to use this particular container:
1. It only supports x86_64 as a host, so having it as a dependency would make it significantly difficult for people trying to build on new architectures.
2. It's on dockerhub, which we've been trying to avoid, but this is a solvable problem.
3. We are not building it in our repository, so anyone who has issues with it has to first look for the recipe, hope that it still works, etc., and it's a big maze. Due to point 1 this might become a problem fairly quickly.

maya-r:
After experimenting a bit and being confused, it turns out SELinux is blocking multiarch/qemu-user-static from working.
The same seems to be true for kubevirt (when using docker, at least), I wonder if this is new to Fedora 35 or people have been building the builder with SELinux disabled.

mhenriks:
> After experimenting a bit and being confused, it turns out SELinux is blocking multiarch/qemu-user-static from working.
The same seems to be true for kubevirt (when using docker, at least), I wonder if this is new to Fedora 35 or people have been building the builder with SELinux disabled.

If the container is being run privileged (as you suggest earlier) I'm surprised that SELinux is an issue

maya-r:
Yeah, SELinux is an issue despite the container being privileged.

```
$ docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
$ getenforce
Enforcing
$ docker run -it --platform linux/arm64 quay.io/centos/centos:stream9 sh
$ sudo setenforce 0
$ docker run -it --platform linux/arm64 quay.io/centos/centos:stream9 sh
sh-5.1# 
```

mhenriks:
> Yeah, SELinux is an issue despite the container being privileged.
> 
> ```
> $ docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
> $ getenforce
> Enforcing
> $ docker run -it --platform linux/arm64 quay.io/centos/centos:stream9 sh
> $ sudo setenforce 0
> $ docker run -it --platform linux/arm64 quay.io/centos/centos:stream9 sh
> sh-5.1# 
> ```

Sorry, I'm not sure what this means.  Running a privileged does "turn off" selinux.  What is being denied by selinux?

maya-r:
I am containing this change in #2279 which also includes more changes to make docker optional.
To avoid additional dependencies, it allows building for one architecture with docker if no buildah is installed.

maya-r:
That one doesn't include the Go change, I guess... I'll do it separately.

PR #2251, opened 2022-04-28
Run update-ca-trust whenever running cdi-source-update-poller

maya-r:
Do this using a shell wrapper that runs it first, then calls the
actual executable with the same argument.

Background:
Our switch to building CentOS Stream images from RPMs has introduced
a small problem that root CAs are not installed.

As a temporary measure, we are running update-ca-trust, a very quick
running binary.

We are working around it using BUILD.bazel to avoid making breaking
changes to non-bazel builds.

Fixes [BZ #2079465](https://bugzilla.redhat.com/show_bug.cgi?id=2079465)

**Release note**:

```release-note
Fix regression since switching to CentOS-stream based images in DataImportCron: run update-ca-trust, so imports using TLS work again.
```

### DIFF
--- a/cmd/cdi-importer/BUILD.bazel
+++ b/cmd/cdi-importer/BUILD.bazel
@@ -47,10 +47,16 @@ container_image(
     ],
     files = [
         ":cdi-importer",
+        ":cdi-source-update-poller-wrapper",
         "//tools/cdi-containerimage-server",
         "//tools/cdi-source-update-poller",
     ],
     visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "cdi-source-update-poller-wrapper",
+    srcs = [":cdi-source-update-poller"],
 )
 
 container_image(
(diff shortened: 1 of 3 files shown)


maya-r:
It turns it into a format openssl/Go seem to understand.
It would be nice to run it at build time - it's personally difficult for me to do these things in Bazel, I figured it'd be exceptionally troublesome because we're supposed to be cross-building too, but maybe I can figure something out.
The script seems to run very fast so it has very little actual impact if any.

akalenyu:
Should we maybe have an external registry func test? That way we would have caught it on the CDI side
I know this is something we don't usually do, maybe we can have this test spawn an importer pod and try to simply reach quay?

akalenyu:
So I have been trying to understand why the update-ca-trust in bazel entrypoint doesn't run for the dataimportcron containers.

I believe the reason is, that for these containers, we override the container's entrypoint with the `Command:` field:
https://github.com/kubevirt/containerized-data-importer/blob/d6bcec367b9173fe06480ae742b3745467bdb712/pkg/controller/dataimportcron-controller.go#L671-L676

maya-r:
@awels is it okay by you to merge this and work separately on generating it at build time?

maya-r:
This was easier than getting update-ca-trust to run!
Closing in favour of #2255 which runs it at build time.

PR #2253, opened 2022-05-01
Address possible nils in dv controller, log CSIDrivers in tests

akalenyu:

- Handle some storageClassName nils
- We'd expect the CSIDriver object to be there, otherwise ceph install might be struggling

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -704,7 +704,10 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 
 	if pvc == nil {
 		if selectedCloneStrategy == SmartClone {
-			snapshotClassName, _ := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
+			snapshotClassName, err := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
 			return r.reconcileSmartClonePvc(log, datavolume, pvcSpec, transferName, snapshotClassName)
 		}
 		if selectedCloneStrategy == CsiClone {
@@ -714,12 +717,20 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 			}
 			if !csiDriverAvailable {
 				// err csi clone not possible
+				storageClass, err := GetStorageClassByName(r.client, pvcSpec.StorageClassName)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+				noCsiDriverMsg := "CSI Clone configured, failed to look for CSIDriver - target storage class could not be found"
+				if storageClass != nil {
+					noCsiDriverMsg = fmt.Sprintf("CSI Clone configured, but no CSIDriver available for %s", storageClass.Name)
+				}
 				return reconcile.Result{},
 					r.updateDataVolumeStatusPhaseWithEvent(cdiv1.CloneScheduled, datavolume, pvc, selectedCloneStrategy,
 						DataVolumeEvent{
 							eventType: corev1.EventTypeWarning,
 							reason:    ErrUnableToClone,
-							message:   fmt.Sprintf("CSI Clone configured, but no CSIDriver available for %s", *pvcSpec.StorageClassName),
+							message:   noCsiDriverMsg,
 						})
 			}
 
@@ -1504,7 +1515,7 @@ func (r *DatavolumeReconciler) cleanupTransfer(log logr.Logger, dv *cdiv1.DataVo
 	}
 
 	RemoveFinalizer(dv, crossNamespaceFinalizer)
-	if err := r.updateDataVolume(dv); dv != nil {
+	if err := r.updateDataVolume(dv); err != nil {
 		return err
 	}
 
@@ -1886,6 +1897,9 @@ func (r *DatavolumeReconciler) validateAdvancedCloneSizeCompatible(
 	sourcePvc *corev1.PersistentVolumeClaim,
 	targetStorageSpec *corev1.PersistentVolumeClaimSpec) (bool, error) {
 	srcStorageClass := &storagev1.StorageClass{}
+	if sourcePvc.Spec.StorageClassName == nil {
+		return false, fmt.Errorf("Source PVC Storage Class name wasn't populated yet by PVC controller")
+	}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *sourcePvc.Spec.StorageClassName}, srcStorageClass); IgnoreNotFound(err) != nil {
 		return false, err
 	}
--- a/pkg/controller/datavolume-controller_test.go
+++ b/pkg/controller/datavolume-controller_test.go
@@ -1403,7 +1403,7 @@ var _ = Describe("All DataVolume Tests", func() {
 			Expect(snapclass).To(BeEmpty())
 		})
 
-		It("Should not return snapshot class, if storage class does not exist", func() {
+		It("Should not return snapshot class, if storage class exists but snapshot class does not exist", func() {
 			dv := newCloneDataVolume("test-dv")
 			scName := "testsc"
 			sc := createStorageClass(scName, map[string]string{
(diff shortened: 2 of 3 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -497,12 +497,19 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 
 	if !prePopulated && !pvcPopulated {
 		if pvc == nil {
+			storageClass, err := GetStorageClassByName(r.client, pvcSpec.StorageClassName)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+			if storageClass == nil {
```
brybacki, line 504: So previously, in the case of host-assisted clone, it would go forward and create a PVC -> line 530 (even without a default storageclass, and without storage class defined), now it will return here with error. 

Is this the way we want to handle this case?

pkg/controller/datavolume-controller.go
```diff
@@ -497,12 +497,19 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
...
+				return reconcile.Result{}, fmt.Errorf("Could not find a default storage class in the cluster")
+			}
 			if selectedCloneStrategy == SmartClone {
-				snapshotClassName, _ := r.getSnapshotClassForSmartClone(datavolume, pvcSpec)
+				snapshotClassName, _ := r.getSnapshotClassForSmartClone(datavolume, storageClass)
 				return r.reconcileSmartClonePvc(log, datavolume, pvcSpec, transferName, snapshotClassName)
 			}
 			if selectedCloneStrategy == CsiClone {
```
brybacki, line 511: In theory we should not be able to get SmartClone is there is no storageClass/defaultStoratgeClass, so it should not get here. 
Anyway, there is an error we ignore here. Maybe we should not?

brybacki: Also, maybe we should look into the `selectCloneStrategy`  and revise what is returned in case there is no StorageClass.

pkg/controller/datavolume-controller.go
akalenyu: Yeah definitely want to avoid logical changes. I think I took this too far, trying to reuse the storage class that gets passed around.

I'll have another go with more direct approach

pkg/controller/datavolume-controller.go
akalenyu: Yeah I think we shouldn't have ignored the error.
I minimized the scope of this commit to handle the nils without altering logic too much

pkg/controller/datavolume-controller.go
```diff
@@ -508,12 +511,20 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 				}
 				if !csiDriverAvailable {
 					// err csi clone not possible
+					storageClass, err := GetStorageClassByName(r.client, pvcSpec.StorageClassName)
+					if err != nil {
+						return reconcile.Result{}, err
+					}
+					if storageClass == nil {
+						log.Info("Could not find a default storage class in the cluster")
```
arnongilboa, line 519: It might be better to remove the word "default" and add the SC name

pkg/controller/datavolume-controller_test.go
```diff
@@ -1518,6 +1518,31 @@ var _ = Describe("All DataVolume Tests", func() {
...
+				[]cdiv1.ClaimPropertySet{{AccessModes: accessMode, VolumeMode: &blockMode}},
+				&strategy)
+
+			reconciler := createDatavolumeReconciler(dv, srcPvc, storageProfile, sc)
+			reconciler.extClientSet = extfake.NewSimpleClientset(createVolumeSnapshotContentCrd(), createVolumeSnapshotClassCrd(), createVolumeSnapshotCrd())
+
+			By("Reconcile")
+			result, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}})
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(result).To(Not(BeNil()))
```
arnongilboa, line 1543: Use ToNot() instead

pkg/controller/datavolume-controller_test.go
```diff
@@ -1518,6 +1518,31 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			accessMode := []corev1.PersistentVolumeAccessMode{corev1.ReadOnlyMany}
+			storageProfile := createStorageProfileWithCloneStrategy(scName,
+				[]cdiv1.ClaimPropertySet{{AccessModes: accessMode, VolumeMode: &blockMode}},
+				&strategy)
+
+			reconciler := createDatavolumeReconciler(dv, srcPvc, storageProfile, sc)
+			reconciler.extClientSet = extfake.NewSimpleClientset(createVolumeSnapshotContentCrd(), createVolumeSnapshotClassCrd(), createVolumeSnapshotCrd())
+
+			By("Reconcile")
+			result, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}})
```
arnongilboa, line 1541: Use dv.Name and dv.Namespace here

pkg/monitoring/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["metrics.go"],
+    importpath = "kubevirt.io/containerized-data-importer/pkg/monitoring",
+    visibility = ["//visibility:public"],
+    deps = ["//pkg/common:go_default_library"],
+)
```
arnongilboa, line 9: Are you sure this is needed?

akalenyu: yeah but it already got added here https://github.com/kubevirt/containerized-data-importer/pull/2226
one of the make targets re-adds this, not sure which one

pkg/controller/datavolume-controller.go
akalenyu: yep, changed it to adjust the event message according to if we have the sc name or not

pkg/monitoring/BUILD.bazel
akalenyu: Rebased and this is now gone

pkg/controller/datavolume-controller.go
akalenyu: @brybacki Should be strictly about avoiding nils now. let me know what you think

awels:
Does this still need to get in or did we solve it some other way?

akalenyu:
> Does this still need to get in or did we solve it some other way?

This NPE is still possible afaik

brybacki:
I think now it looks really good, just one question about source PVC storage class.

pkg/controller/datavolume-controller.go
```diff
@@ -1886,6 +1897,9 @@ func (r *DatavolumeReconciler) validateAdvancedCloneSizeCompatible(
 	sourcePvc *corev1.PersistentVolumeClaim,
 	targetStorageSpec *corev1.PersistentVolumeClaimSpec) (bool, error) {
 	srcStorageClass := &storagev1.StorageClass{}
+	if sourcePvc.Spec.StorageClassName == nil {
```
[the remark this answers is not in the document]
akalenyu: I was just protecting nil access in the following line https://github.com/kubevirt/containerized-data-importer/blob/1ad4cea7712f85cf52e5cb9ae321b1c14c3e7e1f/pkg/controller/datavolume-controller.go#L1903

But this is a good question - I think that a GET from our cache should return PVCs that already have a storage class name populated

akalenyu:
> I think now it looks really good, just one question about source PVC storage class.

ping - I think this is pretty harmless and spares us an NPE

pkg/controller/datavolume-controller.go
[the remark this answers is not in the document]
awels: Storage class is only optional for manually created PVs

PR #2265, opened 2022-05-06
Start smart clone controller from datavolume controller when needed

awels:

It was possible for the CSI snapshot CRD check to fail silently and prevent the smart clone controller from starting during the cdi deployment pod start up. This would prevent smart clone from working properly.

This moves the creation of the smart clone controller into the datavolume controller when someone attempts to do a smart clone. Until that happens the smart clone controller is not started at all. To ensure we don't have multiple smart clone controllers started, we use a channel to synchronize creation, and once created the function will be a no-op.

Fixes #

**Release note**:

```release-note
BugFix: Fix smart clone controller not starting if an error occurred during startup
```


cmd/cdi-controller/controller.go
```diff
@@ -288,7 +289,7 @@ func deleteReadyFile() {
 }
 
 func addCrdInformerEventHandlers(crdInformer cache.SharedIndexInformer, extclient extclientset.Interface, mgr manager.Manager, log logr.Logger) {
-	crdInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+	crdInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
```
mhenriks, line 293: Should handle `UpdateFunc` as well as `AddFunc` as I believe `UpdateFunc` is called on resync

cmd/cdi-controller/controller.go
```diff
@@ -300,11 +301,11 @@ func addCrdInformerEventHandlers(crdInformer cache.SharedIndexInformer, extclien
...
 		},
-	})
+	}, time.Duration(time.Minute))
 }
 
 func startSmartController(extclient extclientset.Interface, mgr manager.Manager, log logr.Logger) {
-	if controller.IsCsiCrdsDeployed(extclient) {
+	if controller.IsCsiCrdsDeployed(extclient, log) {
 		log.Info("CSI CRDs detected, starting smart clone controller")
 		if _, err := controller.NewSmartCloneController(mgr, log, installerLabels); err != nil {
```
mhenriks, line 310: Have to make sure to not do this multiple times I think

awels: Yeah, this probably worked because we never tried to do it more than once.

awels:
It could be a transient network issue for instance, no reason to panic. In essence a panic would cause the control plane of CDI to disappear for some period of time, which could add to whatever causing the transient failure. So why not handle it properly.

pkg/controller/datavolume-controller.go
```diff
@@ -291,6 +294,7 @@ func NewDatavolumeController(
 	if err := addDatavolumeControllerWatches(mgr, datavolumeController); err != nil {
 		return nil, err
 	}
+	go reconciler.startSnapshotCsiCrdDetection()
```
mhenriks, line 297: I think it would be nicer, rather than creating our own goroutine, let the controller runtime manager deal with it.  See here:  https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L61

Basically just pass a Runnable and when the manager is started, it will create/manage the goroutine.

I believe this will clean up all the funniness in the unit test with the dummy goroutine and passing the additional context

mhenriks: We take advantage of this in cdi-operator here:  https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/operator/controller/certrotation.go#L80

pkg/controller/datavolume-controller.go
```diff
@@ -1567,10 +1608,11 @@ func (r *DatavolumeReconciler) storageClassCSIDriverExists(storageClassName *str
 func (r *DatavolumeReconciler) getSnapshotClassForSmartClone(dataVolume *cdiv1.DataVolume, targetStorageSpec *corev1.PersistentVolumeClaimSpec) (string, error) {
 	log := r.log.WithName("getSnapshotClassForSmartClone").V(3)
 	// Check if relevant CRDs are available
-	if !IsCsiCrdsDeployed(r.extClientSet) {
+	if !IsCsiCrdsDeployed(r.client, r.log) {
 		log.Info("Missing CSI snapshotter CRDs, falling back to host assisted clone")
 		return "", nil
 	}
+	r.sccs.StartController()
```
mhenriks, line 1615: This seems a little too hidden in this function, how about here:  https://github.com/kubevirt/containerized-data-importer/blob/f68376fe09b2e72937e9b42712700e0f78404dc0/pkg/controller/datavolume-controller.go#L482

Check if strategy is smart clone and start controller if so.  In this location (main reconcile func) it is more explicit, no weird side effects

awels: I mean if we do manage to attempt to start the smart clone controller, it will fail when we add the watches in the controller because the CRDs don't exist. but you make a good point that we won't pick that strategy without checking if the CRDs exist. I already fixed it btw.

PR #2275, opened 2022-05-14
Add missing file repo.yaml

maya-r:
Forgotten in #1983, needed to run `make rpm-deps`

**Release note**:

```release-note
NONE
```


akalenyu:
We have seen this flake on NFS lanes before, but not as consistently
Somehow the disk.img is twice the size it should be
/retest

PR #2276, opened 2022-05-15
Fix smart clone request size update

ShellyKa13:
In case the pvc got an actual size as the expected
size there wasn't an update of the pvc spec with the
actual user request size, which left the pvc spec with
a smaller size then the user requested(the data size).
This caused a discrepancy when trying to restore such pvc,
which restored a pvc with the small size instead of the user request
size.

Signed-off-by: Shelly Kagan <skagan@redhat.com>

Fixes #BZ2021354

**Release note**:

```release-note
Fix smart clone not updating request size if already got the requested size so expansion is not required
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -1379,10 +1379,11 @@ func (r *DatavolumeReconciler) expand(log logr.Logger,
 	}
 
 	expansionRequired := actualSize.Cmp(requestedSize) < 0
+	updateRequestSizeRequired := actualSize.Cmp(requestedSize) <= 0 && requestedSize.Cmp(currentSize) > 0
 
 	log.V(3).Info("Expand sizes", "req", requestedSize, "cur", currentSize, "act", actualSize, "exp", expansionRequired)
 
-	if expansionRequired && requestedSize.Cmp(currentSize) != 0 {
+	if updateRequestSizeRequired {
 		pvc.Spec.Resources.Requests[corev1.ResourceStorage] = requestedSize
 		if err := r.client.Update(context.TODO(), pvc); err != nil {
 			return false, err
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -1379,7 +1379,7 @@ func (r *DatavolumeReconciler) expand(log logr.Logger,
 	}
 
 	expansionRequired := actualSize.Cmp(requestedSize) < 0
-	updateRequestSizeRequired := actualSize.Cmp(requestedSize) <= 0 && requestedSize.Cmp(currentSize) > 0
+	updateRequestSizeRequired := actualSize.Cmp(requestedSize) <= 0 && currentSize.Cmp(requestedSize) < 0
 
 	log.V(3).Info("Expand sizes", "req", requestedSize, "cur", currentSize, "act", actualSize, "exp", expansionRequired)
 
--- a/pkg/controller/datavolume-controller_test.go
+++ b/pkg/controller/datavolume-controller_test.go
@@ -1454,7 +1454,7 @@ var _ = Describe("All DataVolume Tests", func() {
 		)
 
 		DescribeTable("After smart clone", func(actualSize resource.Quantity, currentSize resource.Quantity, expectedDvPhase cdiv1.DataVolumePhase) {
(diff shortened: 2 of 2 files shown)


akalenyu:
https://github.com/kubevirt/containerized-data-importer/blob/e269d706b3bac1b8063fd3a365bbc3394be6e33e/pkg/controller/smart-clone-controller.go#L378
Doesn't this ensure that target smart clone PVCs always get created with the correct size?

```bash
# ./cluster-up/kubectl.sh explain volumesnapshot.status.restoreSize
KIND:     VolumeSnapshot
VERSION:  snapshot.storage.k8s.io/v1

FIELD:    restoreSize <string>

DESCRIPTION:
     restoreSize represents the minimum size of volume required to create a
     volume from this snapshot. In dynamic snapshot creation case, this field
     will be filled in by the snapshot controller with the "size_bytes" value
     returned from CSI "CreateSnapshot" gRPC call. For a pre-existing snapshot,
     this field will be filled with the "size_bytes" value returned from the CSI
     "ListSnapshots" gRPC call if the driver supports it. When restoring a
     volume from this snapshot, the size of the volume MUST NOT be smaller than
     the restoreSize if it is specified, otherwise the restoration will fail. If
     not specified, it indicates that the size is unknown.
```

ShellyKa13:
@akalenyu the code you sent is the first phase of the smart clone. First a pvc of the size of the restore size is created, then the snapshot is being cloned to that PVC, once that is complete that PVC is being expended to the source PVC size. For example:
source pvc was 20G, 10G was actually written to it. the restore size in that case is 10G. When the smart clone start it creates a PVC of 10G, clone to that the snapshot which is of size 10G, then expand to get the actual source PVC size to 20G.
Specifically in this bug, due to different provisioner, we requested for a 10G size PVC for the first phase, we got a 20G PVC, so when expansion wasn't needed cause the size was as the original PVC was, but we still want it to be represented in the pvc spec request size, this update should still happen - thats the fix.

pkg/controller/datavolume-controller.go
```diff
@@ -1379,10 +1379,11 @@ func (r *DatavolumeReconciler) expand(log logr.Logger,
 	}
 
 	expansionRequired := actualSize.Cmp(requestedSize) < 0
+	updateRequestSizeRequired := actualSize.Cmp(requestedSize) <= 0 && requestedSize.Cmp(currentSize) > 0
```
arnongilboa, line 1382: I guess you haven't used expansionRequired because you want to have also the equals case?
I think the second comparison will be better readable as  currentSize.Cmp(requestedSize) <= 0, similar to the first one.

akalenyu:
So after discussing the possibility of picking `max(restoreSize, userReq)` for the restore PVC from the beginning,
I am convinced it is riskier for two reasons:
- There are many provisioners in the wild and we have experienced (on our flesh) that some don't play nice when `srcSize != targetSize`
(note ceph csi [reverted](https://github.com/ceph/ceph-csi/pull/2716) that starting 3.5.0, @mhenriks FYI)
- Significant change, goes beyond scope of bugfix (whole expansion logic put to question)

/lgtm

pkg/controller/datavolume-controller_test.go
```diff
@@ -1453,6 +1453,59 @@ var _ = Describe("All DataVolume Tests", func() {
 			Entry("snapshot", cdiv1.CDICloneStrategy(cdiv1.CloneStrategySnapshot)),
 		)
 
+		DescribeTable("After smart clone", func(actualSize resource.Quantity, currentSize resource.Quantity, expectedDvPhase cdiv1.DataVolumePhase) {
+			strategy := cdiv1.CDICloneStrategy(cdiv1.CloneStrategySnapshot)
```
awels, line 1457: This line inspired me to write https://github.com/kubevirt/containerized-data-importer/pull/2277 there should be no reason that you need to cast the const to the right type.

pkg/controller/datavolume-controller.go
ShellyKa13: right the expansion required is used later in the code, adding the = to it changes that logic. Ill change as you suggest

ShellyKa13: actually changing to currentSize.Cmp(requestedSize) <= 0 is wrong since you dont need to update if they are equal.. so ill change it to < if its more clear that the requested size is the second parameter in the comparison

PR #2277, opened 2022-05-16
Make CDI clone strategy strings correct types

awels:

Noticed the consts were strings instead of the right types this changes the type to the right type and modifies the usage to not have to cast to the right type all over the place.

Fixes #

**Release note**:

```release-note
BugFix: Modified CDICloneStrategy strings to be correct type
```

### DIFF
--- a/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
+++ b/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
@@ -654,13 +654,13 @@ type CDICloneStrategy string
 
 const (
 	// CloneStrategyHostAssisted specifies slower, host-assisted copy
-	CloneStrategyHostAssisted = "copy"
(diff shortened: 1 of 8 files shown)


staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -654,13 +654,13 @@ type CDICloneStrategy string
 
 const (
 	// CloneStrategyHostAssisted specifies slower, host-assisted copy
-	CloneStrategyHostAssisted = "copy"
+	CloneStrategyHostAssisted CDICloneStrategy = "copy"
```
brybacki, line 657: In yaml it is still a string so no problem that we are changing it here, right?

awels: Right it is actually still a 'string' in memory, it is just properly types as something besides a string in the code. This way we don't have to do as many casts to the data structure type.

PR #2295, opened 2022-05-26
Reconcile DIC only if DataSource is not managed by another DIC

arnongilboa:

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -60,6 +60,11 @@ import (
 )
 
 const (
+	// ErrDataSourceAlreadyManaged provides a const to indicate DataSource already managed error
+	ErrDataSourceAlreadyManaged = "ErrDataSourceAlreadyManaged"
+	// MessageDataSourceAlreadyManaged provides a const to form DataSource already managed error message
+	MessageDataSourceAlreadyManaged = "DataSource %s is already managed by DataImportCron %s"
+
 	prometheusNsLabel       = "ns"
 	prometheusCronNameLabel = "cron_name"
 )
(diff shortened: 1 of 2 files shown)


awels:
So what is the problem this is trying to solve. This seems like a very imperative approach. Why can't two DiCs coexist at the same time, and one of them notices the data source is owned by another DiC, and simply doesn't do anything (emit an event saying this so things observing us can see something is not right in the cluster).

arnongilboa:
> So what is the problem this is trying to solve. This seems like a very imperative approach. Why can't two DiCs coexist at the same time, and one of them notices the data source is owned by another DiC, and simply doesn't do anything (emit an event saying this so things observing us can see something is not right in the cluster).

Why would you allow a second DIC be created and "coexist" (doing nothing) with the active one? it may have totally different source registry url, different storage etc. and add nothing but user confusion which can be eliminated on creation time. If you want them to coexist it complicates life even more as you may expect that if the active one is deleted the other one is activated.

pkg/controller/dataimportcron-controller.go
```diff
@@ -114,12 +119,43 @@ func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.
...
+	otherCron := &cdiv1.DataImportCron{}
+	if err := r.client.Get(ctx, types.NamespacedName{Namespace: cron.Namespace, Name: dataSourceCronLabel}, otherCron); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+	if otherCron.Spec.ManagedDataSource == dataSource.Name {
+		msg := fmt.Sprintf(MessageDataSourceAlreadyManaged, dataSource.Name, otherCron.Name)
+		r.recorder.Event(cron, corev1.EventTypeWarning, ErrDataSourceAlreadyManaged, msg)
```
awels, line 153: besides the event, I would also log this at V(3) in the cdi controller log.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -273,6 +273,42 @@ var _ = Describe("All DataImportCron Tests", func() {
...
+			_, err = reconciler.Reconcile(context.TODO(), cronReq)
+			Expect(err).ToNot(HaveOccurred())
+
+			dataSource = &cdiv1.DataSource{}
+			err = reconciler.client.Get(context.TODO(), dataSourceKey(cron), dataSource)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(dataSource.Labels[common.DataImportCronLabel]).To(Equal(cron.Name))
+
+			cron1 := newDataImportCron(cronName + "1")
+			shouldReconcile, err = reconciler.shouldReconcileCron(context.TODO(), cron1)
```
awels, line 293: Can you check that the event is fired here? The reconciler should have an event slice you can check.

awels:
a few minor comments but looks good otherwise.

pkg/controller/dataimportcron-controller.go
arnongilboa: added

pkg/controller/dataimportcron-controller_test.go
arnongilboa: added

PR #2297, opened 2022-05-26
Small shim so libvddk-test-plugin.so can run on older glibc

maya-r:
vddk-test is special, it creates a dynamic library that is
loaded by the importer for test suite runs.

When running tests on a cluster built with an importer built from a
different base image, e.g. when RedHat builds with a CentOS8
equivalent, we might be attempting to load libvddk-test-plugin.so on
the CentOS8 importer. This breaks, as it doesn't have newer symbols.

There's only one new glibc symbol used, for fstat. Let's use
the older glibc definition of fstat, __fxstat explicitly.

**Release note**:

```release-note
Fix regression where the testsuite cannot be run if the importer image is built with an older base image (like centos 8)
```

### DIFF
--- a/tools/vddk-test/vddk-test-plugin.c
+++ b/tools/vddk-test/vddk-test-plugin.c
@@ -53,6 +53,9 @@ void *fakevddk_open(int readonly) {
     return (void *) &fd;
 }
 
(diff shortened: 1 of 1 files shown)


akalenyu:
Nice! verified against the rhel8 importer
/lgtm

PR #2301, opened 2022-05-29
[WIP] Add some logic to retry host assisted clone if readonly not supported

akalenyu:

We see CSI drivers which won't allow readOnly: true, lets add some best effort logic
to revert back to false if we must.

The indication for this from the storage driver is an event on the created pod, sadly no error:
https://github.com/dell/csi-unity/blob/76cb78decd0c7e0ae3bded9668e7fbe9b39dd2ff/service/node.go#L420-L424

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=2087784

**Release note**:

```release-note
BugFix: Source PVC mounted readonly at all times, discarding storages which do not support this
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -36,6 +36,10 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/generator"
 )
 
+var (
+	retryWithoutReadonlyThreshold = time.Minute
+)
+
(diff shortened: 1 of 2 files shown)


akalenyu:
/close

we dont want to be the compatibility layer, plus it is basically only a guess that the storage doesn't support "readonly" (just assume pending for a long time is bad)

Issue #2250, opened 2022-04-28
Incorrect handling of prepopulated volumes


brybacki:
**What happened**:

The way CDI handles populated PVC is not correct. There are some edge cases that make restore fail. 
1. No source pvc when original DV was cloned - fail.
`Message: "admission webhook \"datavolume-validate.cdi.kubevirt.io\" denied the request:  Source PVC cdi-e2e-tests-dv-func-test-rq7w9/sourcepvcempty not found"`

2.  the controller sees "populated for", takes ownership of the PVC, but still gets to the ensureExtendedToken 
This bug has more details: https://bugzilla.redhat.com/show_bug.cgi?id=2070366

And maybe there are some more problems, in the import/upload. 

The problem was first reported here: https://bugzilla.redhat.com/show_bug.cgi?id=2070366

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
Steps to reproduce the behavior.

**Additional context**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): N/A
- Kubernetes version (use `kubectl version`): N/A
- DV specification: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

brybacki:
The second part is already fixed
"the controller sees "populated for", takes ownership of the PVC, but still gets to the ensureExtendedToken
This bug has more details: https://bugzilla.redhat.com/show_bug.cgi?id=2070366"

Issue #1962, opened 2021-09-28
Accept clones of non existing PVC.


awels:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
/kind enhancement

**What happened**:
CDI Rejects DataVolume clones if the source PVC doesn't exist. This is not very declarative as the order in which you provide yaml matters. 

**What you expected to happen**:
Instead CDI should accept the clone request, and simply not act on the clone until the source is available. It should generate events/condition updates if the source does not exist.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

aglitke:
Fixed by #2306

PR #2306, opened 2022-06-01
Allow creating clones without source PVC

alromeros:
**What this PR does / why we need it**: Following the behavior reported in https://github.com/kubevirt/containerized-data-importer/issues/2250, we should allow the creation of a clone DV even if the source PVC doesn't exist. The DV will be handled by the controller until the source PVC eventually is created, following a more descriptive approach.

Fixes https://github.com/kubevirt/containerized-data-importer/issues/2250

**Special notes for your reviewer**: Work in progress.

**Release note**:

```release-note
Allow creating clones without source PVC
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -190,10 +190,10 @@ const (
 	CloneValidationFailed = "CloneValidationFailed"
 	// MessageCloneValidationFailed reports that a clone wasn't admitted by our validation mechanism (message)
 	MessageCloneValidationFailed = "The clone doesn't meet the validation requirements"
-	// CloneWithoutSource reports that the source PVC of a clone doesn't exists yet (reason)
+	// CloneWithoutSource reports that the source PVC of a clone doesn't exists (reason)
 	CloneWithoutSource = "CloneWithoutSource"
-	// MessageCloneWithoutSource reports that the source PVC of a clone doesn't exists yet (message)
-	MessageCloneWithoutSource = "The source PVC doesn't exist yet"
+	// MessageCloneWithoutSource reports that the source PVC of a clone doesn't exists (message)
+	MessageCloneWithoutSource = "The source PVC doesn't exist"
 
 	// AnnCSICloneRequest annotation associates object with CSI Clone Request
 	AnnCSICloneRequest = "cdi.kubevirt.io/CSICloneRequest"
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -1043,17 +1043,19 @@ var _ = Describe("all clone tests", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}
 
+			getAndHandleMD5 := func(pvc *v1.PersistentVolumeClaim, imgPath string) string {
+				defer deleteAndWaitForVerifierPod()
+				md5, err := f.GetMD5(f.Namespace, pvc, imgPath, crossVolumeModeCloneMD5NumBytes)
+				Expect(err).ToNot(HaveOccurred())
+				return md5
+			}
+
 			compareCloneWithSource := func(sourcePvc, targetPvc *v1.PersistentVolumeClaim, sourceImgPath, targetImgPath string) {
 				By("Source file system pvc md5summing")
-				sourceMD5, err := f.GetMD5(f.Namespace, sourcePvc, sourceImgPath, crossVolumeModeCloneMD5NumBytes)
-				Expect(err).ToNot(HaveOccurred())
-				deleteAndWaitForVerifierPod()
-
+				sourceMD5 := getAndHandleMD5(sourcePvc, sourceImgPath)
 				By("Target file system pvc md5summing")
-				targetMD5, err := f.GetMD5(f.Namespace, targetPvc, targetImgPath, crossVolumeModeCloneMD5NumBytes)
-				Expect(err).ToNot(HaveOccurred())
+				targetMD5 := getAndHandleMD5(targetPvc, targetImgPath)
 				Expect(sourceMD5).To(Equal(targetMD5))
-				deleteAndWaitForVerifierPod()
 			}
 
 			It("Should finish the clone after creating the source PVC", func() {
@@ -1101,6 +1103,7 @@ var _ = Describe("all clone tests", func() {
 				expectEvent(f, f.Namespace.Name).Should(ContainSubstring(controller.CloneWithoutSource))
 
 				By("Create source PVC")
+				// We use a larger size in the source PVC so the validation fails
 				sourceDV := utils.NewDataVolumeWithHTTPImportAndStorageSpec(dataVolumeName, "1Gi", fmt.Sprintf(utils.TinyCoreIsoURL, f.CdiInstallNs))
 				sourceDV.Spec.Storage.VolumeMode = &blockVM
 				sourceDV, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, sourceDV)
--- a/pkg/controller/datavolume-controller_test.go
+++ b/pkg/controller/datavolume-controller_test.go
@@ -1563,7 +1563,7 @@ var _ = Describe("All DataVolume Tests", func() {
 			AnnDefaultStorageClass: "true",
 		}, map[string]string{}, "csi-plugin")
 
-		It("Clone without source PVC", func() {
+		It("Validate clone without source as feasible, but not done", func() {
 			dv := newCloneDataVolume("test-dv")
 			storageProfile := createStorageProfile(scName, nil, filesystemMode)
 			reconciler = createDatavolumeReconciler(dv, storageProfile, sc)
@@ -1573,7 +1573,7 @@ var _ = Describe("All DataVolume Tests", func() {
 			Expect(done).To(BeFalse())
 		})
 
-		It("Create source PVC after clone without source", func() {
+		It("Validate that clone without source completes after PVC is created", func() {
 			dv := newCloneDataVolume("test-dv")
 			storageProfile := createStorageProfile(scName, nil, filesystemMode)
 			reconciler = createDatavolumeReconciler(dv, storageProfile, sc)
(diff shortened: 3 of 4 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -722,10 +725,15 @@ func (r *DatavolumeReconciler) ensureExtendedToken(pvc *corev1.PersistentVolumeC
 }
 
 func (r *DatavolumeReconciler) selectCloneStrategy(datavolume *cdiv1.DataVolume, pvcSpec *corev1.PersistentVolumeClaimSpec) (cloneStrategy, error) {
-	if datavolume.Spec.Source.PVC == nil {
+	isClone := datavolume.Spec.Source.PVC != nil
```
arnongilboa, line 728: Not sure this var adds readability, so either remove it or at least reduce it scope:
`if isClone := datavolume.Spec.Source.PVC != nil; !isClone {...}`

pkg/controller/datavolume-controller.go
```diff
@@ -2779,6 +2787,35 @@ func getDeltaTTL(dv *cdiv1.DataVolume, ttl int32) time.Duration {
...
+// handleCloneWithoutSource does error handling and validation of clones without source PVC
+func (r *DatavolumeReconciler) handleCloneWithoutSource(datavolume *cdiv1.DataVolume) error {
+	sourcePvc, err := r.findSourcePvc(datavolume)
+	if err != nil {
+		// Clone without source
+		if k8serrors.IsNotFound(err) {
+			datavolume.Annotations[AnnNoSourceClone] = "true"
+			r.updateDataVolume(datavolume)
+		}
+		return err
```
arnongilboa, line 2799: Why do you return the error and not nil in case of NotFound?

pkg/controller/datavolume-controller.go
```diff
@@ -2779,6 +2787,35 @@ func getDeltaTTL(dv *cdiv1.DataVolume, ttl int32) time.Duration {
...
+	if available && noSource == "true" {
+		// Since the admission webhook couldn't validate the clone against its source PVC
+		// when creating the DV, we do the validation now
+		err := ValidateClone(sourcePvc, &datavolume.Spec)
+		if err != nil {
+			return err
+		}
+
+		// The clone has a source now
+		datavolume.Annotations[AnnNoSourceClone] = ""
```
arnongilboa, line 2812: So the annotation is removed to do it only once? comment will be nice

pkg/controller/datavolume-controller.go
alromeros: If we just return nil there, the clone will fail anyway when trying to get the clone strategy. I need to think about how to handle this, but I see two main options: 
- The current one: We return an error, so the reconcile loop keeps running again and again until the source PVC is created and the clone can be completed. However, I guess we don't want unhandled errors in our reconciler, so I'm not sure about that. Maybe it'll be cleaner to add a `RequeueAfter`?

- The ideal: If the source PVC is not found, we just end the reconcile loop until the source PVC is created, without constantly rerunning the loop or triggering redundant errors. I tried to do this but I don't see an easy way to trigger the reconcile again when the source PVC is created.

pkg/controller/datavolume-controller.go
alromeros: Exactly, since the webhook validation is skipped when creating the DV, I added that annotation to make sure we still check it once the source PVC is created. However, I don't see the point of keeping the annotation.

pkg/controller/datavolume-controller.go
alromeros: Thanks for sharing that, I think it's exactly what I wanted. I tried to do it by making the clone own the source PVC, but it was getting very complicated.

pkg/controller/datavolume-controller.go
brybacki: In my opinion it really helps, it is not clear from the first look that `PVC != Nil` means it is clone. Of course if you stop and think about it it makes sense. But having it explicitly named makes code easier to quickly read.

pkg/controller/datavolume-controller.go
```diff
@@ -2779,6 +2851,39 @@ func getDeltaTTL(dv *cdiv1.DataVolume, ttl int32) time.Duration {
 	return delta
 }
 
+// handleCloneWithoutSource does error handling and validation of clones without source PVC
+func (r *DatavolumeReconciler) handleCloneWithoutSource(datavolume *cdiv1.DataVolume) (bool, error) {
+	sourcePvc, err := r.findSourcePvc(datavolume)
+	if err != nil {
+		// Clone without source
+		if k8serrors.IsNotFound(err) {
+			datavolume.Annotations[AnnNoSourceClone] = "true"
```
awels, line 2860: Would it not make more sense to instead of setting this annotation, that we update the Running condition to false with `sourcePVC not found`? I am trying to understand the purpose of this annotation, and it is not actually used for anything, we create it and then remove it, but what is its purpose?

awels (commented):
Can someone explain the purpose of the annotation, I don't understand it.

pkg/controller/datavolume-controller.go
alromeros: When we create a clone without source, the clone-specific validation in the admission webhook is skipped, so my idea is to eventually check it once the PVC is created. I use that annotation to specify that a clone has been created without source so, when the source PVC is created, we do the validation. Once the validation succeeds, I remove the annotation because we just want to check the validation once, as we do in the webhook.

Maybe is not ideal to just use an annotation as a flag? Or it's better to rename it as something like `AnnCloneNeedsValidation`? I still see the point of having it unless we want to skip the validation and let the clone-controller handle it.

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -526,7 +526,7 @@ var _ = Describe("Validating Webhook", func() {
 			Expect(resp.Allowed).To(Equal(false))
 		})
 
-		It("should reject DataVolume with SourceRef on create if DataSource exists but PVC does not exist", func() {
+		It("should accept DataVolume with SourceRef on create if DataSource exists but PVC does not exist", func() {
```
aglitke, line 529: I think we should also allow creation of a DV when the DataSource doesn't exist.  These interdependent objects should be allowed to be created in any order.  Would it inflate the scope of this PR too much to apply the same concept to DataSources?

alromeros: Good question, I'm not very well-versed in DataSources so I'm trying to understand if this feature should also work with `SourceRef`. Still have to do some research about that.

pkg/controller/datavolume-controller.go
awels: I am saying why are you not always doing the validation if you see the PVC is there? Regardless of annotation.

awels: Yes, exactly, there is no reason that after we removed the annotation something didn't go wrong, and we need to reconcile, we still need to validate at that point, and with the annotation gone we won't. The validation is not very compute intensive, so why not just do it all the time. That way we will never miss it when we should have run it.

alromeros: hmmm I guess that, In that case, I can remove the annotation and just do the validation on every clone. However, wouldn't it be odd to keep repeating that validation on all clones independently of if they are created with or without source?

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -84,9 +85,14 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		}
 		dataSource, err := wh.cdiClient.CdiV1beta1().DataSources(ns).Get(context.TODO(), dataVolume.Spec.SourceRef.Name, metav1.GetOptions{})
 		if err != nil {
-			return toAdmissionResponseError(err)
+			if !k8serrors.IsNotFound(err) {
+				return toAdmissionResponseError(err)
+			}
+			// Create a mock 'pvcSource' to do the validation even if the dataSource doesn't exist yet
+			pvcSource = &cdiv1.DataVolumeSourcePVC{Name: dataVolume.Spec.SourceRef.Name, Namespace: ns}
```
mhenriks, line 92: There is no way to securely allow creation of a DataVolume with sourceRef that does not exist.  In the webhook, we must validate that the user creating the DataVolume has permission to the source PVC namespace.  But there is no way to know what the source namespace may be.  And this validation cannot be deferred to later (in the controller) because the user context is lost.

mhenriks (changes requested):
I am not aware of any way to securely allow cloning when the resource referred to by `spec.sourceRef` does not exist. 

Regarding the PVC referred to by `spec.source.pvc` it is okay if the PVC does not exist but of course the namespace MUST exist.  I assume `clone.CanUserClonePVC` already handles that case.  But it should probably be explicitly tested.

pkg/apiserver/webhooks/datavolume-validate_test.go
mhenriks: @aglitke there is no way to know what the ultimate source namespace will be if the sourceRef does not exist

pkg/apiserver/webhooks/datavolume-mutate.go
alromeros: You are right, I somehow assumed that the datasource's namespace would be the same as the source PVC's one. However, I'm still trying to figure out if there's a way to do this validation in the controller, or if we could just skip it and let the clone fail if the DataVolume doesn't meet the requirements.

alromeros: I've just seen that `clone.CanUserClonePVC` doesn't check if the source namespace exists, so I'll also modify that.

pkg/apiserver/webhooks/datavolume-mutate_test.go
```diff
@@ -197,6 +198,28 @@ var _ = Describe("Mutating DataVolume Webhook", func() {
 			Expect(resp.Patch).To(BeNil())
 		})
 
+		It("should reject a clone if the source PVC's namespace doesnt' exist", func() {
```
arnongilboa, line 201: doesn't :)

pkg/controller/util.go
```diff
@@ -896,6 +896,50 @@ func validateTokenData(tokenData *token.Payload, srcNamespace, srcName, targetNa
...
+	} else {
+		targetResources = spec.Storage.Resources
+		// The storage size in the target DV can be empty
+		// when cloning using the 'Storage' API
+		if _, ok := targetResources.Requests["storage"]; !ok {
+			isSizelessClone = true
+		}
+	}
+
+	// TODO: Spec.Storage API needs a better more complex check to validate clone size - to account for fsOverhead
```
arnongilboa, line 932: so you don't support Spec.Storage but return nil in that case? why not error?

alromeros: It's actually supported, but the comment suggests that It may need better validation. That code is old (I've only moved it to `util.go` so I can use it in the controller), so maybe the comment should be improved. I think @brybacki is currently improving this mechanism in https://github.com/kubevirt/containerized-data-importer/pull/2299.

tests/cloner_test.go
```diff
@@ -1029,6 +1029,90 @@ var _ = Describe("all clone tests", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				By("Wait for clone DV Succeeded phase")
+				err = utils.WaitForDataVolumePhaseWithTimeout(f, f.Namespace.Name, cdiv1.Succeeded, "clone-dv", cloneCompleteTimeout)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Compare the two clones to see if they have the same hash
+				compareCloneWithSource(sourcePvc, clonePvc, diskImagePath, diskImagePath)
+			})
+
+			It("Should reject the clone after creating the source PVC if the validation fails ", func() {
```
awels, line 1091: Can we put a comment or something indicating the failure will be a size failure? I read this test several times wondering why it was supposed to fail until I realized the tiny size of the target.

tests/cloner_test.go
```diff
@@ -1029,6 +1029,90 @@ var _ = Describe("all clone tests", func() {
...
+
+			deleteAndWaitForVerifierPod := func() {
+				By("Deleting verifier pod")
+				err := utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+				Expect(err).ToNot(HaveOccurred())
+				_, err = utils.WaitPodDeleted(f.K8sClient, utils.VerifierPodName, f.Namespace.Name, verifyPodDeletedTimeout)
+				Expect(err).ToNot(HaveOccurred())
+			}
+
+			compareCloneWithSource := func(sourcePvc, targetPvc *v1.PersistentVolumeClaim, sourceImgPath, targetImgPath string) {
```
awels, line 1046: Can you split this into two functions, my main motivation is that currently it is possible that if `f.GetMD5` errors, the verifier pod is never cleaned up by the test. This is not a huge issue because we delete the namespace at the end of the test, but I want to be diligent about cleaning up things we create without having to rely on the namespace to do it. If you split this, you can `defer` the `deleteAndWaitForVerifierPod` and be sure it is called.

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -127,6 +127,11 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		sourceNamespace = targetNamespace
 	}
 
+	_, err = wh.k8sClient.CoreV1().Namespaces().Get(context.TODO(), sourceNamespace, metav1.GetOptions{})
```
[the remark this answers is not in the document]
alromeros: I also thought that, but I tested several cases where the source namespace didn't exist and `clone.CanUserClonePVC` always succeeded. Mind that, before my PR, all related unit tests also succeeded without having the namespaces, so that's why I had to mock these after adding the code above:

https://github.com/kubevirt/containerized-data-importer/blob/43cc9d3e695a410addb7ce1449123fe8c29f2507/pkg/apiserver/webhooks/datavolume-mutate_test.go#L309-L310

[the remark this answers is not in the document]
awels: So remember in the unit tests we don't have a full fledged API server, so it is unlikely to fail on something like that. Does it not reject in functional tests is the question here. If it does, we need to ensure that we get the same behavior in the unit tests.

awels (approved):
/approve

aglitke:
I believe this fixes #1962

PR #2316, opened 2022-06-07
Update DataImportCron CronJob when needed

arnongilboa:

When CDI is being upgraded, the DataImportCron CronJobs are kept as is, and refer cdi-importer image of the previous version, rather than the current (upgraded) one.

Fixes #

**Release note**:
```release-note
Update DataImportCron CronJob container image if needed due to upgrade
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -158,23 +158,32 @@ func (r *DataImportCronReconciler) shouldReconcileCron(ctx context.Context, cron
 }
 
 func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
-	if isURLSource(dataImportCron) && !r.cronJobExists(ctx, dataImportCron) {
-		cronJob, err := r.newCronJob(dataImportCron)
-		if err != nil {
-			return err
+	if isImageStreamSource(dataImportCron) {
+		if dataImportCron.Annotations[AnnNextCronTime] == "" {
+			AddAnnotation(dataImportCron, AnnNextCronTime, time.Now().Format(time.RFC3339))
 		}
-		if err := r.client.Create(ctx, cronJob); err != nil {
-			return err
-		}
-		job, err := r.newInitialJob(dataImportCron, cronJob)
-		if err != nil {
-			return err
-		}
-		if err := r.client.Create(ctx, job); err != nil {
-			return err
-		}
-	} else if isImageStreamSource(dataImportCron) && dataImportCron.Annotations[AnnNextCronTime] == "" {
-		AddAnnotation(dataImportCron, AnnNextCronTime, time.Now().Format(time.RFC3339))
+		return nil
+	}
+	if !isURLSource(dataImportCron) {
+		return nil
+	}
+	exists, err := r.cronJobExistsAndUpdated(ctx, dataImportCron)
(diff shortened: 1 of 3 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -702,10 +711,19 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 	return nil
 }
 
-func (r *DataImportCronReconciler) cronJobExists(ctx context.Context, cron *cdiv1.DataImportCron) bool {
+func (r *DataImportCronReconciler) cronJobExistsAndUpdated(ctx context.Context, cron *cdiv1.DataImportCron) (bool, error) {
 	var cronJob batchv1.CronJob
 	cronJobNamespacedName := types.NamespacedName{Namespace: r.cdiNamespace, Name: GetCronJobName(cron)}
-	return r.uncachedClient.Get(ctx, cronJobNamespacedName, &cronJob) == nil
```
akalenyu, line 708: Why were we doing uncached client here btw?

pkg/controller/dataimportcron-controller.go
```diff
@@ -702,10 +711,19 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
+	err := r.uncachedClient.Get(ctx, cronJobNamespacedName, &cronJob)
+	podSpec := &cronJob.Spec.JobTemplate.Spec.Template.Spec
+	if err == nil && len(podSpec.Containers) > 0 && podSpec.Containers[0].Image != r.image {
+		podSpec.Containers[0].Image = r.image
```
akalenyu, line 720: Maybe Update() over newCronJob().Spec != foundCron.Spec?
It is possible at some point in the future we decided to change some fields in the cronjob object too, and want to sync existing cronjobs to those values

pkg/controller/dataimportcron-controller.go
arnongilboa: afair r.client.Get() sometime returned NotFound after creating CronJob. However, I'll try again r.client for Get() and Update() here and check if it's consistent.

pkg/controller/dataimportcron-controller.go
arnongilboa: @akalenyu tried this change and it seems that when CronJob is created based on newCronJob(), k8s patches several fields of the spec (dnsPolicy, schedulerName, serviceAccount etc.) so we cannot simply compare (DeepEqual) the current and the desired spec.

pkg/controller/dataimportcron-controller.go
```diff
@@ -702,10 +711,19 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
+		err = r.uncachedClient.Update(ctx, &cronJob)
```
awels, line 721: uncached client here seems wrong as well.

pkg/controller/dataimportcron-controller.go
arnongilboa: Removed. So far no surprises.

pkg/controller/dataimportcron-controller.go
arnongilboa: Use your tip, thanks @akalenyu !

pkg/controller/dataimportcron-controller.go

PR #2326, opened 2022-06-16
Add GC func tests; mutate DV GC annotation only on create

arnongilboa:

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-mutate.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate.go
@@ -101,17 +101,20 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 	modifiedDataVolume := dataVolume.DeepCopy()
 	modified := false
 
-	config, err := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
-	if err != nil {
-		return toAdmissionResponseError(err)
-	}
-	if config.Spec.DataVolumeTTLSeconds != nil &&
-		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
-		if modifiedDataVolume.Annotations == nil {
-			modifiedDataVolume.Annotations = make(map[string]string)
+	if ar.Request.Operation == admissionv1.Create {
+		config, err := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
+		if err != nil {
+			return toAdmissionResponseError(err)
 		}
-		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] = "true"
-		modified = true
+		if config.Spec.DataVolumeTTLSeconds != nil {
+			if modifiedDataVolume.Annotations == nil {
+				modifiedDataVolume.Annotations = make(map[string]string)
+			}
+			if modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
+				modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] = "true"
+				modified = true
+			}
+		}
 	}
 
 	if pvcSource == nil {
@@ -127,7 +130,7 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		sourceNamespace = targetNamespace
 	}
 
-	_, err = wh.k8sClient.CoreV1().Namespaces().Get(context.TODO(), sourceNamespace, metav1.GetOptions{})
+	_, err := wh.k8sClient.CoreV1().Namespaces().Get(context.TODO(), sourceNamespace, metav1.GetOptions{})
 	if err != nil {
 		return toAdmissionResponseError(err)
 	}
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -2790,6 +2790,7 @@ func (r *DatavolumeReconciler) garbageCollect(dataVolume *cdiv1.DataVolume, pvc 
 	}
 	dvTTL := cdiConfig.Spec.DataVolumeTTLSeconds
 	if dvTTL == nil || *dvTTL < 0 {
+		log.Info("Garbage Collection is disabled")
 		return nil, nil
 	}
 	// Current DV still has TTL, so reconcile will return with the needed RequeueAfter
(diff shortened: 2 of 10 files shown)


pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -101,17 +101,19 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
-		if modifiedDataVolume.Annotations == nil {
-			modifiedDataVolume.Annotations = make(map[string]string)
+	if ar.Request.Operation == admissionv1.Create {
+		config, err := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
+		if err != nil {
+			return toAdmissionResponseError(err)
+		}
+		if config.Spec.DataVolumeTTLSeconds != nil &&
+			modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
+			if modifiedDataVolume.Annotations == nil {
```
brybacki, line 111: if `modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion]` is not False, then how can the `Annotations` be nil?

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -101,17 +101,19 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
 
-	config, err := wh.cdiClient.CdiV1beta1().CDIConfigs().Get(context.TODO(), common.ConfigName, metav1.GetOptions{})
-	if err != nil {
-		return toAdmissionResponseError(err)
-	}
-	if config.Spec.DataVolumeTTLSeconds != nil &&
-		modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] != "false" {
-		if modifiedDataVolume.Annotations == nil {
-			modifiedDataVolume.Annotations = make(map[string]string)
+	if ar.Request.Operation == admissionv1.Create {
```
brybacki, line 104: it now only applies on create, does it mean previously it could apply to update or something like that?
Maybe it is worth to state it in commit message, instead of "minor fixes".

pkg/apiserver/webhooks/datavolume-mutate.go
brybacki: oh, it is the go way of handling empty/nil Annotations map

arnongilboa: exactly

pkg/apiserver/webhooks/datavolume-mutate.go
arnongilboa: it's not that dramatic, but I'll add it to the commit message

pkg/apiserver/webhooks/datavolume-mutate.go
ShellyKa13: this is strange to me if modifiedDataVolume.Annotations is nil wouldn't modifiedDataVolume.Annotations[controller.AnnDeleteAfterCompletion] be nil de-referencin and panic?

ShellyKa13: This is an interesting behavior of go, weird :)

arnongilboa: Yep, you can check a key on a nil map in go with no fear :)
However, if both of you didn't like it, I'll re-write these lines.

awels: I am fine with both ways, sometimes it is easier to read with the more explicit version.

tests/cloner_test.go
```diff
@@ -111,6 +113,53 @@ var _ = Describe("all clone tests", func() {
...
+			}
+			verifyDisabledGC := func(dvName string) {
+				VerifyDisabledGC(f, dvName, f.Namespace.Name)
+			}
+			enableGcAndAnnotateLegacyDv := func(dvName string) {
+				EnableGcAndAnnotateLegacyDv(f, dvName, f.Namespace.Name)
+			}
+
+			DescribeTable("Should", func(ttl *int32, verifyGCFunc, additionalTestFunc func(dvName string)) {
+				SetConfigTTL(f, ttl)
```
awels, line 128: I don't see us restoring this setting to the original value in an AfterEach or something like that.

tests/import_test.go
```diff
@@ -216,52 +218,28 @@ var _ = Describe("DataVolume Garbage Collection", func() {
...
 
-	DescribeTable("Should", func(annDeleteAfterCompletion string, verifyGCFunc func(dvName string)) {
-		By("Set DataVolumeTTLSeconds to 0")
-		ttl := int32(0)
-		err = utils.UpdateCDIConfig(f.CrClient, func(config *cdiv1.CDIConfigSpec) {
-			config.DataVolumeTTLSeconds = &ttl
-		})
-		Expect(err).ToNot(HaveOccurred())
+	DescribeTable("Should", func(ttl *int32, annDeleteAfterCompletion string, verifyGCFunc, additionalTestFunc func(dvName string), verifyContent bool) {
+		tests.SetConfigTTL(f, ttl)
```
awels, line 237: Same, don't see us restoring the original value. Maybe I am missing something?

tests/upload_test.go
```diff
@@ -920,6 +920,74 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+		}
+		verifyDisabledGC := func(dvName string) {
+			tests.VerifyDisabledGC(f, dvName, f.Namespace.Name)
+		}
+		enableGcAndAnnotateLegacyDv := func(dvName string) {
+			tests.EnableGcAndAnnotateLegacyDv(f, dvName, f.Namespace.Name)
+		}
+
+		DescribeTable("Should", func(ttl *int32, verifyGCFunc, additionalTestFunc func(dvName string)) {
+			tests.SetConfigTTL(f, ttl)
```
awels, line 935: Same don't see restore.

tests/import_test.go
arnongilboa: AfterEach here is already restoring CDIConfig to original state.

tests/cloner_test.go
arnongilboa: Right, adding.

tests/upload_test.go
arnongilboa: Right, adding.

awels (commented):
looks good just one question

pkg/controller/datavolume-controller.go
```diff
@@ -2803,15 +2804,12 @@ func (r *DatavolumeReconciler) garbageCollect(dataVolume *cdiv1.DataVolume, pvc
 }
 
 func (r *DatavolumeReconciler) detachPvcDeleteDv(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume, log logr.Logger) error {
-	if !IsSucceeded(pvc) {
+	if dv.Status.Phase != cdiv1.Succeeded {
```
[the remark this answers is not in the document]
awels: I am fairly certain the warm migration flows are now broken because they depend on this annotation `pvc.Annotations[AnnCurrentCheckpoint] == ""` to determine if they actually succeeded the final check point or not.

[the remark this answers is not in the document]
arnongilboa: I'll check, but where are the ftest failures?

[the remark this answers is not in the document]
awels: I believe we have most warm migration tests disabled because they are pretty flaky already https://github.com/kubevirt/containerized-data-importer/blob/main/tests/datavolume_test.go#L781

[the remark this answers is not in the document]
arnongilboa: Ok, so I'll re-check and/or fix it in Sunday.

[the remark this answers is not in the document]
arnongilboa: @awels no worries, import dv is succeeded only if pvc pod phase is succeeded and no current checkpoint
https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume-controller.go#L2082-L2097

PR #2331, opened 2022-06-22
Comply with restricted security context

awels:
Ensure the CDI pods comply with the restricted security context.

Fixes #
RHBZ [2088464](https://bugzilla.redhat.com/show_bug.cgi?id=2088464)

**Release note**:

```release-note
BugFix: drop all capabilities and set priviledgeEscalation: false on pods
```

### DIFF
--- a/pkg/controller/BUILD.bazel
+++ b/pkg/controller/BUILD.bazel
@@ -61,6 +61,7 @@ go_library(
         "//vendor/k8s.io/client-go/tools/cache:go_default_library",
         "//vendor/k8s.io/client-go/tools/record:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
+        "//vendor/k8s.io/utils/pointer:go_default_library",
         "//vendor/kubevirt.io/controller-lifecycle-operator-sdk/api:go_default_library",
         "//vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk:go_default_library",
         "//vendor/sigs.k8s.io/controller-runtime/pkg/cache:go_default_library",
--- a/pkg/importer/registry-datasource.go
+++ b/pkg/importer/registry-datasource.go
@@ -57,6 +57,7 @@ type RegistryDataSource struct {
 func NewRegistryDataSource(endpoint, accessKey, secKey, certDir string, insecureTLS bool) *RegistryDataSource {
 	allCertDir, err := createCertificateDir(certDir)
 	if err != nil {
+		klog.Infof("Error creating allCertDir %v", err)
 		if allCertDir != "/" {
 			err = os.RemoveAll(allCertDir)
 			if err != nil {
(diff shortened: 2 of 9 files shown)


akalenyu:
I am not sure, so please correct me, but don't we also have to do this for:
- upload pods
- clone source pods
- job/cronjob in dataimportcron controller
- size detection pod
- smart clone expansion pod
- possibly our deployments, which we chuck over to the lifecycle-operator-sdk to manage for us

BTW I think the release note will be kind of confusing with the bugzilla link,
usually, I see people put the bugzilla following the `Fixes #` bit

awels:
These look like real failures. something wrong with creating a directory that contains all proxy certs `mkdir /all_certs: permission denied` if I don't remove ALL it works, just need to figure out which missing CAP is the preventing me from creating the directory.

pkg/controller/BUILD.bazel
```diff
@@ -61,6 +61,7 @@ go_library(
         "//vendor/k8s.io/client-go/tools/cache:go_default_library",
         "//vendor/k8s.io/client-go/tools/record:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
+        "//vendor/k8s.io/utils/pointer:go_default_library",
```
akalenyu, line 64: I think this needs another `make deps-update` since it is marked "indirect" in go.mod

akalenyu (commented):
Looks awesome!

Where do you see the `mkdir` errs?
A while back we hit a case where we were failing with a PVC mount at `/pvc` but succeeding with `/dev/pvc`

Also, should we also do this for our deployments (cdi-deployment, uploadproxy, etc)?
I understand these run in a "special" namespace but we might as well if they don't require extra caps?

awels:
The destructive lane failures: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2331/pull-containerized-data-importer-e2e-destructive/1542857710301614080 I added some extra logging locally, and the mkdir is failing if I drop ALL, trying to figure out what I need to add back.

akalenyu:
> The destructive lane failures: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2331/pull-containerized-data-importer-e2e-destructive/1542857710301614080 I added some extra logging locally, and the mkdir is failing if I drop ALL, trying to figure out what I need to add back.

Gotcha. maybe instead of adding some capability, we can get around it by creating the folder elsewhere? like not in `/`?

pkg/importer/registry-datasource.go
```diff
@@ -165,14 +166,20 @@ func getImageFileName(dir string) (string, error) {
 }
 
 func createCertificateDir(registryCertDir string) (string, error) {
-	allCerts := "/all_certs"
-	err := os.MkdirAll(allCerts, 0777)
+	allCerts := "/tmp/all_certs"
```
akalenyu, line 169: just double checking that /tmp is the correct place for this.
some distros treat this differently, maybe this gets cleaned up? `/var/tmp` is another option

PR #2335, opened 2022-06-24
Improve error handling when pod creation fails

alromeros:
**What this PR does / why we need it**: When pod creation fails, the error is usually logged without providing additional information to the user. This behavior is especially risky when the user lacks the permits to check the logs, making it unintuitive and almost impossible to find the source of the problem.

This PR aims to improve the error handling of the pod-creation process, so pertinent info about the failure is included in the pod's PVC/DV.

Fixes #

**Release note**:

```release-note
Improve error handling when pod creation fails
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"crypto/rsa"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"reflect"
 	"strconv"
 	"strings"
 	"time"
+
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/api"
 
@@ -284,7 +285,7 @@ func (r *CloneReconciler) reconcileSourcePod(sourcePod *corev1.Pod, targetPvc *c
 
 		sourcePod, err := r.CreateCloneSourcePod(r.image, r.pullPolicy, targetPvc, log)
 		// Check if pod has failed and, in that case, record an event with the error
-		if podErr := handleFailedPod(err, createCloneSourcePodName(targetPvc), sourcePvc, r.recorder, r.client); podErr != nil {
+		if podErr := handleFailedPod(err, createCloneSourcePodName(targetPvc), targetPvc, r.recorder, r.client); podErr != nil {
 			return 0, podErr
 		}
 
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -539,7 +539,8 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *co
 	util.SetRecommendedLabels(pod, r.installerLabels, "cdi-controller")
 
 	if err := r.client.Create(context.TODO(), pod); err != nil {
-		return nil, errors.Wrap(err, "source pod API create errored")
+		err = errors.Wrap(err, "source pod API create errored")
+		return nil, handleFailedPod(err, createCloneSourcePodName(pvc), sourcePvc, r.recorder, r.client)
 	}
 
 	log.V(1).Info("cloning source pod (image) created\n", "pod.Namespace", pod.Namespace, "pod.Name", pod.Name, "image", image)
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -283,9 +283,11 @@ func (r *CloneReconciler) reconcileSourcePod(sourcePod *corev1.Pod, targetPvc *c
 		}
 
 		sourcePod, err := r.CreateCloneSourcePod(r.image, r.pullPolicy, targetPvc, log)
-		if err != nil {
-			return 0, err
+		// Check if pod has failed and, in that case, record an event with the error
+		if podErr := handleFailedPod(err, createCloneSourcePodName(targetPvc), sourcePvc, r.recorder, r.client); podErr != nil {
+			return 0, podErr
 		}
+
 		log.V(3).Info("Created source pod ", "sourcePod.Namespace", sourcePod.Namespace, "sourcePod.Name", sourcePod.Name)
 	}
 	return 0, nil
@@ -539,8 +541,7 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *co
 	util.SetRecommendedLabels(pod, r.installerLabels, "cdi-controller")
 
 	if err := r.client.Create(context.TODO(), pod); err != nil {
-		err = errors.Wrap(err, "source pod API create errored")
-		return nil, handleFailedPod(err, createCloneSourcePodName(pvc), sourcePvc, r.recorder, r.client)
+		return nil, errors.Wrap(err, "source pod API create errored")
 	}
 
 	log.V(1).Info("cloning source pod (image) created\n", "pod.Namespace", pod.Namespace, "pod.Name", pod.Name, "image", image)
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -1556,7 +1556,7 @@ func (r *DatavolumeReconciler) expand(log logr.Logger,
 		var err error
 		pod, err = r.createExpansionPod(pvc, dv, podName)
 		if err != nil {
-			return false, err
+			return false, handleFailedPod(err, podName, pvc, r.recorder, r.client)
 		}
 	}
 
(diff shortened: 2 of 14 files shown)


ShellyKa13 (commented):
looks good, some suggestions inline

pkg/controller/util.go
```diff
@@ -1176,6 +1181,27 @@ func inflateSizeWithOverhead(c client.Client, imgSize int64, pvcSpec *v1.Persist
 	return returnSize, nil
 }
 
+// handleFailedPod handles pod-creation errors and updates the pod's PVC without providing sensitive information
+func handleFailedPod(err error, podName string, pvc *v1.PersistentVolumeClaim, recorder record.EventRecorder, c client.Client) error {
+	// Generic reason and msg to avoid providing sensitive information
```
arnongilboa, line 1186: If you add `if err == nil { return nil }` you can rewrite all the ifs calling this func to:
```
if err := handleFailedPod(...); err != nil {
  return ...err...
}
```
which I think is nicer than calling it in the return call

pkg/controller/util.go
```diff
@@ -1176,6 +1181,27 @@ func inflateSizeWithOverhead(c client.Client, imgSize int64, pvcSpec *v1.Persist
...
+	reason := ErrStartingPod
+	msg := fmt.Sprintf(MessageErrStartingPod, podName)
+
+	// Error handling to fine-tune the event with pertinent info
+	if errQuotaExceeded(err) {
+		reason = ErrExceededQuota
```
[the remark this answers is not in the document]
alromeros: I agree that it's way more readable, but wouldn't that error expose too much information? I also thought about including that message, but after following the conversation between Alexander and Dan in the [Jira issue](https://issues.redhat.com/browse/CNV-19180), I guessed that it's preferred to omit this system-specific information in events. I'm not sure about this anyway, so I ping @awels for help.

pkg/controller/util.go
alromeros: Yeah that's a good one, I'll change it :+1:

tests/utils.go
```diff
@@ -0,0 +1,421 @@
+package tests
```
brybacki, line 1: How did we end up with this utils file?

Either the github diff is "broken" for this case or I can not find the source. But I can see a lot of new functions being added. 

Is it migrated from some other tests/utils?

brybacki: Ok, I Get it. The diff here is shown to the current main. A new PR was just merged
https://github.com/kubevirt/containerized-data-importer/pull/2310

@alromeros I wonder if an automatic merge will work ok in this case. Or maybe you need to rebase to make sure it is correct!

brybacki: Ok I have not seen your previous comment, you replied so fast ;)

brybacki: the rest looks good

pkg/controller/clone-controller.go
```diff
@@ -539,8 +541,7 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *co
 	util.SetRecommendedLabels(pod, r.installerLabels, "cdi-controller")
 
 	if err := r.client.Create(context.TODO(), pod); err != nil {
-		err = errors.Wrap(err, "source pod API create errored")
-		return nil, handleFailedPod(err, createCloneSourcePodName(pvc), sourcePvc, r.recorder, r.client)
+		return nil, errors.Wrap(err, "source pod API create errored")
```
ShellyKa13, line 544: here you dont need to handle failed pod?

pkg/controller/datavolume-controller.go
```diff
@@ -1555,8 +1555,9 @@ func (r *DatavolumeReconciler) expand(log logr.Logger,
 	if !podExists {
 		var err error
 		pod, err = r.createExpansionPod(pvc, dv, podName)
-		if err != nil {
-			return false, handleFailedPod(err, podName, pvc, r.recorder, r.client)
+		// Check if pod has failed and, in that case, record an event with the error
+		if podErr := handleFailedPod(err, podName, pvc, r.recorder, r.client); podErr != nil {
```
ShellyKa13, line 1559: I actually liked it better like the first way you did it but style preferences :)

pkg/controller/datavolume-controller.go
```diff
@@ -2958,8 +2959,9 @@ func (r *DatavolumeReconciler) getSizeFromPod(sourcePvc *corev1.PersistentVolume
 	}
 
 	pod, err := r.getOrCreateSizeDetectionPod(sourcePvc, dv)
-	if err != nil {
-		return 0, handleFailedPod(err, sizeDetectionPodName(sourcePvc), sourcePvc, r.recorder, r.client)
+	// Check if pod has failed and, in that case, record an event with the error
+	if podErr := handleFailedPod(err, sizeDetectionPodName(sourcePvc), sourcePvc, r.recorder, r.client); podErr != nil {
+		return 0, podErr
 	} else if !isPodComplete(pod) {
```
ShellyKa13, line 2965: you dont need else if you return in the if

pkg/controller/clone-controller.go
brybacki: this is already handled by the caller of CreateCloneSourcePod (line 287)

pkg/controller/datavolume-controller.go
alromeros: hmmm I'm torn between the two. I think the current one is easier to read but maybe too verbose.

ShellyKa13: you dont need to change again thats ok

pkg/controller/datavolume-controller.go
alromeros: Sorry, I forgot to comment back. I can change this if it's preferable, but I kinda like how the error handling of the pod is concentrated in a single `if/else` block. Though it's not really necessary, I feel like the `if else` allows concentrating similar behavior in the same statement and makes it more readable.

brybacki: On the other hand, this guard statement is a common idiom in go:
```
if err := xyz(); err != nil {
   return err
}
```
and after spending some time reading `go`, I can say it is easy to read and reason about. It is obvious that in bad case if immediately returns, while in the good case it continues with next statements.

When I see `if err / else ` I expect some kind of complex logic around error handling, or some alternative behavior based on few conditions.

In this case both versions are correct. We might argue which way is more idiomatic.

ShellyKa13 (commented):
/lgtm

pkg/controller/util.go
```diff
@@ -1171,6 +1176,30 @@ func inflateSizeWithOverhead(c client.Client, imgSize int64, pvcSpec *v1.Persist
...
+	} // TODO: Add more error cases
+
+	recorder.Event(pvc, v1.EventTypeWarning, reason, msg)
+
+	AddAnnotation(pvc, AnnPodPhase, string(v1.PodFailed))
```
awels, line 1195: DataVolumes should never be marked failed. That would violate the eventually consistency of datavolumes. We should emit an event (which we do here) and update the datavolume running condition to false with the reason why. It is entirely possible that the namespace this is running in will get more quota and the DV will succeed when that happens. If we mark it failed, that is it, the DV will no longer do anything and you have to delete it, and re-create it.

You can find functions to manipulate the conditions in datavolume-conditions.go

awels (changes requested):
A misunderstanding in what the DV status should be, but looks good otherwise.

pkg/controller/util.go
alromeros: I have some doubts regarding how to handle this. The problem is that the error handling function is mainly called outside the datavolume-controller, so I can't directly access the datavolume to modify its running condition.

I see two options regarding how to handle this in `handleFailedPod`, but both have their downsides:

- To add the `PodFailed` annotation in the PVC and then handle that in the datavolume-controller, so we change the **running condition** to false instead of the **phase** to failed. I see this option more in concordance with what you said in the comment, but I fear it might change some other behavior in the controller. **I've just implemented this one**.

- To change the running condition of the PVC to false, so the DV is updated in the datavolume-controller without modifying a lot of code. This option is way less intrusive, but I see a major problem: When either the clone or the size-detection pod fails, the PVC that should be updated is the source PVC, not the target one. This means that the running condition of the source DV will also change, which is unwanted behavior.

awels: Why would we update the source PVC? the source is completed, we have determined that. So we always update the target PVC. In fact we already have some handling on the bound condition in case the source PVC is not bound (which is also possible). Always update the target PVC.

alromeros: Since the size-detection pod and the source-clone pod both use the source PVC, I thought it made sense to change the `AnnPodFailed` annotation in the source and not the target in these cases. I agree that it makes more sense to change the running condition in the target PVC, but I still find it odd to handle a PVC that has no direct relation with the pod in `handleFailedPod`.

If it's preferred to use the target in all cases, I guess I can just change the running condition and let the datavolume-controller handle everything (second option). However, don't my changes regarding what to do with the `AnnPodFailed` annotation in the datavolume-controller make sense? What case would justify setting the DV to failed after a pod fails?

awels (approved):

PR #2340, opened 2022-06-28
Disable DV garbage collection by default in func tests

arnongilboa:

Since GC if currently disabled by default in OpenShift, we keep it disabled in func tests by default, as we need test coverage when TTL is nil. We already override it for specific GC tests, and we will add a specific lane that enables it. 

Fixes #

**Release note**:
```release-note
NONE
```


maya-r:
Can you write the motivation for doing this change in the commit message? thanks.

PR #2349, opened 2022-06-30
Add Ceph DV GC lane automation script

arnongilboa:

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/automation/ceph-gc.sh
+++ b/automation/ceph-gc.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


akalenyu:
Are we fine with the GC lane being a ceph one? just asking because HPP might be cheaper, I could be missing a downside though

akalenyu:
> > Are we fine with the GC lane being a ceph one? just asking because HPP might be cheaper, I could be missing a downside though
> > @mhenriks commented that since hpp does not support smart clone it would be better to test ttl on/off with rook ceph

gotcha
/lgtm

awels:
Note the lane is not automatically run nor required at this point, which is good because we are still testing it. But please run it a few times manually so we can see it is working well before enabling it full time.

arnongilboa:
> Note the lane is not automatically run nor required at this point, which is good because we are still testing it. But please run it a few times manually so we can see it is working well before enabling it full time.

sure, that's the idea

PR #2363, opened 2022-07-20
Fix race condition in upload-controller to keep PVC annotations after a clone error

alromeros:
## What this PR does / why we need it

When a pod fails after the error handling added in https://github.com/kubevirt/containerized-data-importer/pull/2335, some annotations (running condition, pod phase...) are updated in the affected PVC to let other controllers know.
    
However, due to a lack of synchronization between some controllers, a race condition could happen where, if a pod is successfully created just after a pod fails in a different controller, the annotations set in the error handling would just be updated inappropriately.

This PR updates the upload controller so it checks for clone errors before updating the annotations.

## How to replicate

The described race condition has been only replicable under particular conditions, namely, a cross-namespace clone where the source namespace doesn't have enough quota to create the source-clone pod.

Once the source-clone pod fails due to quota, the upload pod in the target namespace keeps running after being successfully created, updating the annotations set after the previous failure.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -246,11 +246,10 @@ func (r *UploadReconciler) reconcilePVC(log logr.Logger, pvc *corev1.PersistentV
 		return reconcile.Result{}, err
 	}
 
-	podPhase := pod.Status.Phase
-	anno[AnnPodPhase] = string(podPhase)
-	anno[AnnPodReady] = strconv.FormatBool(isPodReady(pod))
-
-	setAnnotationsFromPodWithPrefix(anno, pod, AnnRunningCondition)
+	// Update the annotations in the PVC to reflect the current state of the upload
+	if err := r.updateUploadAnnotations(pvc, anno, pod, isCloneTarget); err != nil {
+		return reconcile.Result{}, err
+	}
 
 	if !reflect.DeepEqual(pvc, pvcCopy) {
 		if err := r.updatePVC(pvcCopy); err != nil {
@@ -480,6 +479,58 @@ func (r *UploadReconciler) deleteService(namespace, serviceName string) error {
 		}
 	}
 
+	return nil
+}
+
+// checkIfCloneFailed inspects clone-related annotations and pods to diagnose if the cloning process has failed
+func (r *UploadReconciler) checkIfCloneFailed(pvc *corev1.PersistentVolumeClaim, isCloneTarget bool) (bool, error) {
+	isCloneRequest, sourceNamespace, _ := ParseCloneRequestAnnotation(pvc)
+	if !isCloneRequest || !isCloneTarget {
+		return false, nil
+	}
+
+	// Clone finished succesfully
+	if pvc.Annotations[AnnCloneOf] == "true" {
+		return false, nil
+	}
+	// Clone source pod not ready
+	sourcePodName, ok := pvc.Annotations[AnnCloneSourcePod]
+	if !ok {
+		return false, nil
+	}
+
+	pod := &corev1.Pod{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: sourcePodName, Namespace: sourceNamespace}, pod); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return true, err
+		}
+		// We assume the clone has failed when the pod doesn't exist, the cloning process hasn't finished and the PVC's running condition is false
+		if pvc.Annotations[AnnRunningCondition] == "false" {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// updateUploadAnnotations updates annotations without overwritting previous conditions or states
+func (r *UploadReconciler) updateUploadAnnotations(pvc *corev1.PersistentVolumeClaim, anno map[string]string, pod *v1.Pod, isCloneTarget bool) error {
+	// First we check if the clone failed to avoid changing the running condition back to true
+	cloneFailed, err := r.checkIfCloneFailed(pvc, isCloneTarget)
+	if err != nil {
+		return err
+	}
+
+	if cloneFailed {
+		return nil
+	}
+
+	// Update PVC with upload annotations
+	podPhase := pod.Status.Phase
+	anno[AnnPodPhase] = string(podPhase)
+	anno[AnnPodReady] = strconv.FormatBool(isPodReady(pod))
+
+	setAnnotationsFromPodWithPrefix(anno, pod, AnnRunningCondition)
 	return nil
 }
 
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -247,9 +247,7 @@ func (r *UploadReconciler) reconcilePVC(log logr.Logger, pvc *corev1.PersistentV
 	}
 
 	// Update the annotations in the PVC to reflect the current state of the upload
-	if err := r.updateUploadAnnotations(pvc, anno, pod, isCloneTarget); err != nil {
-		return reconcile.Result{}, err
-	}
+	updateUploadAnnotations(pvc, anno, pod, isCloneTarget)
 
 	if !reflect.DeepEqual(pvc, pvcCopy) {
 		if err := r.updatePVC(pvcCopy); err != nil {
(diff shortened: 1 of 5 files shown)


mhenriks (commented):
Alternative implementation proposal:  use different annotations for the different controllers.  

If we are worried that other components will rely on the current annotation, it can still be set by the datavolume controller

awels:
Having re-read the problem description. I find it somewhat worrisome that if the source pod failed to create, that the target pod succeeds. The pod should remain running until the source pod connects to it and transfers the data. Am I missing something here?

alromeros:
> Having re-read the problem description. I find it somewhat worrisome that if the source pod failed to create, that the target pod succeeds. The pod should remain running until the source pod connects to it and transfers the data. Am I missing something here?

@awels You are right, I should've specified better what I meant by that. The upload pod then remains in  `Running` phase (no errors at all), so all the annotations in the PVC are overwritten. The upload doesn't fully succeed, but the controller keeps running without errors.

pkg/controller/upload-controller.go
```diff
@@ -483,6 +482,58 @@ func (r *UploadReconciler) deleteService(namespace, serviceName string) error {
...
+		return false, nil
+	}
+	// Clone source pod not ready
+	sourcePodName, ok := pvc.Annotations[AnnCloneSourcePod]
+	if !ok {
+		return false, nil
+	}
+
+	pod := &corev1.Pod{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: sourcePodName, Namespace: sourceNamespace}, pod); err != nil {
```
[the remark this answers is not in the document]
alromeros: Yeah, I can understand why. I'll consider the controller-specific annotations to avoid that.

pkg/controller/util.go
```diff
@@ -1188,11 +1188,17 @@ func handleFailedPod(err error, podName string, pvc *v1.PersistentVolumeClaim, r
...
+	}
 
 	recorder.Event(pvc, v1.EventTypeWarning, reason, msg)
 
+	isCloneSourcePod := createCloneSourcePodName(pvc) == podName
+	if isCloneSourcePod {
+		AddAnnotation(pvc, AnnCloneSourcePodFailed, "true")
+	}
 	AddAnnotation(pvc, AnnRunningCondition, "false")
+	AddAnnotation(pvc, AnnRunningConditionReason, "")
```
awels, line 1200: We should tell the user why we are not running. We have several options for the reason above set in `reason` we should probably use it.

pkg/controller/upload-controller.go
```diff
@@ -483,6 +482,25 @@ func (r *UploadReconciler) deleteService(namespace, serviceName string) error {
 	return nil
 }
 
+// updateUploadAnnotations updates annotations without overwritting previous conditions or states
+func updateUploadAnnotations(pvc *corev1.PersistentVolumeClaim, anno map[string]string, pod *v1.Pod, isCloneTarget bool) error {
+	// First we check if the clone failed to avoid changing the running condition back to true
+	if isCloneTarget {
+		cloneFailed, ok := pvc.Annotations[AnnCloneSourcePodFailed]
+		if ok && cloneFailed == "true" {
+			return nil
```
awels, line 491: I don't think we should return here, we should set the annotations as normal, and in the DV controller in the clone section check for both annotations and update the DV condition accordingly. Something like:
if sourcePod running && targetPod running -> DV running condition = true
if sourcePod !running && targetPod running -> DV running condition = false
if sourcePod running && targetPod !running -> DV running condition = false
if !sourcePod running && targetPod !running -> DV running condition = false
And there we can put a message that is a combination of the message from the source and the message from the target.

pkg/controller/clone-controller.go
```diff
@@ -46,6 +46,8 @@ const (
 	AnnCloneOf = "k8s.io/CloneOf"
 	// AnnCloneToken is the annotation containing the clone token
 	AnnCloneToken = "cdi.kubevirt.io/storage.clone.token"
+	// AnnCloneSourcePodFailed is an annotation used to specify that the source clone pod failed
+	AnnCloneSourcePodFailed = "cdi.kubevirt.io/storage.source.clone.pod.failed"
```
awels, line 50: I would not call it failed, its just not running, so I would use several of the same annotations associated with the target, running/reason/message but just source versions. That way we can take source values and the target values and in the DV controller build a coherent message for the user.

alromeros: hmmm after thinking about it, I don't know if this option is better than using the current annotation: The pod actually fails, in this case, for quota limitations (this annotation is added in the `handleFailedPod` function).

The annotation is used to let the upload controller know that the source-clone pod failed (it did), but it has no direct relation with the source PVC. And, since the source PVC is still running (the condition is changed in the target), I don't think it's necessary to use the `AnnSourceRunningCondition` annotation. Also, I think that handling this in the upload controller with a single annotation is easier and cleaner than adding new logic in the datavolume-controller.

awels: Note that a pod not starting due to quota is not a terminal state. In the future someone could add quota and the pod can start. So we should simply indicate in the datavolume condition that source/target pod was not able to start due to quota (or other unknown reasons). If the condition on the DV indicates not enough quota for instance, that gives the user some indication on what to do to fix it. Just saying failed and not doing anything from that point forward defeats the whole eventually consistent idea.

alromeros: You are right, and It's true that using the `AnnSourceRunningCondition` it's the easiest and shortest way to keep the running condition false in the DV without adding new logic. I'll commit the changes soon, but that said, the upload controller will keep overwriting the target PVC running condition, which I'm sure it's not a big deal, but still could lead to some confusion. Lastly, notice that all the error handling to let users know was already implemented in the previous PR: https://github.com/kubevirt/containerized-data-importer/pull/2335 (the event in the DV, changing the running condition, etc.).

pkg/controller/util.go
```diff
@@ -1188,11 +1188,18 @@ func handleFailedPod(err error, podName string, pvc *v1.PersistentVolumeClaim, r
...
 
 	recorder.Event(pvc, v1.EventTypeWarning, reason, msg)
 
+	if isCloneSourcePod := createCloneSourcePodName(pvc) == podName; isCloneSourcePod {
+		AddAnnotation(pvc, AnnSourceRunningCondition, "false")
+		AddAnnotation(pvc, AnnSourceRunningConditionReason, reason)
+		AddAnnotation(pvc, AnnSourceRunningConditionMessage, msg)
+	}
 	AddAnnotation(pvc, AnnRunningCondition, "false")
+	AddAnnotation(pvc, AnnRunningConditionReason, reason)
```
awels, line 1201: Should this not be in an else.
```
if isCloneSourcePod {
  // AddCloneSourceAnnotation
} else {
  // AddTargetAnnotation
}
```
This way the source pod will not influence the target annotation at all.

awels:
I think we are almost there, with the else the source pod won't influence the target pod annotations anymore.

pkg/controller/util.go
alromeros: Since the target PVC is also affected by the source-clone pod failure (maybe even more than the source), doesn't it also makes sense to add the annotations there? However, I see that the behavior will remain the same and we'll have fewer annotations cluttering up the PVC, so I'll commit the change.

alromeros:
> I think we are almost there, with the else the source pod won't influence the target pod annotations anymore. Note I checked #2335 and I didn't see the logic that takes the source pod annotation and target pod annotation and combines them into the running condition. Can you point me to it?

Sorry, I meant that in that PR we had already added conditions/events to the PVC and DV so the user can have some indications to fix the problem (not just the annotations).

Since the datavolume-controller only uses the PVC `AnnRunningCondition` and `AnnSourceRunningCondition` annotations as a reference to change its own running condition, I thought it was unnecessary to add new logic there, which is already handled in this piece of code:

https://github.com/kubevirt/containerized-data-importer/blob/1290e13246ef1ae06bdab6b0fcad150cd7de64cf/pkg/controller/datavolume-conditions.go#L90-L92

pkg/controller/util.go
awels: Everything is written to the target PVC, (source pod and target pod annotations). We should never update the source PVC in any of this. The datavolume controller only looks at the target PVC for the annotations.

awels:
There it is, yeah that is doing exactly what I described in the comment above with the 4 possible combinations.

PR #2368, opened 2022-07-21
Fix failure to resize when the request image size is specified in milli- scale.

maya-r:
I noticed this when cloning block->file system to a slightly larger (1Gi -> 1.1Gi) DV.
Unclear whether this happens in some more scenarios, I'm surprised this problem wasn't reported by others.

Fixes #

**Release note**:

```release-note
Fix failure to resize when the request image size is specified in milli- scale.
```

### DIFF
--- a/pkg/importer/data-processor.go
+++ b/pkg/importer/data-processor.go
@@ -402,7 +402,7 @@ func (dp *DataProcessor) calculateTargetSize() int64 {
 		targetQuantity = &minQuantity
 	}
 	klog.V(1).Infof("Target size %s.\n", targetQuantity.String())
-	targetSize, _ := targetQuantity.AsInt64()
+	targetSize := targetQuantity.Value()
 	return targetSize
 }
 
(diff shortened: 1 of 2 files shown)


awels:
So basically it was scaling something slightly larger than 1Gi to 2Gi? If so I think I have seen that before, just never connected it to anything.

maya-r:
When I create a PVC with size 1.1Gi that size is stored as 1181116006400m (that's milli-bytes).
When I try to run that through MustParse and AsInt64 [it seems to overflow](https://go.dev/play/p/nUiI_e4KWH8) (not sure why, it's a valid int64?)

It tries to run the clone again but fails for another issue. It might be because we run calculateTargetSize only *after* CleanDir is called so now it's right to think there's not enough space, and perhaps we shouldn't validate in the codepath triggered here at all (resize->validate) because that was done for the filesystem overhead validation which we decided to drop in #2195.

awels:
Is it our code that generates the millibytes value? Or is it kubernetes? Looking at the package, calling .Value() is the same as .ScaledValue(0) (in fact Value calls ScaledValue(0))

maya-r:
Switched to using Value() instead of ScaledValue(0). It is kubernetes that generates this value.

awels:
silly kubernetes creating milliscale values.
/approve

PR #2373, opened 2022-07-25
Queue DataVolumes referred to in "populatedFor" annotations

mhenriks:

This issue came up in testing backup/restore with Kasten k10.  Should queue DataVolumes referred to in "populatedFor" annotations.  Without this change, a "prePopulated" DV created before a "populatedFor" DV will be pending forever if the PVC is not owned by the DV.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -409,10 +409,31 @@ func addDatavolumeControllerWatches(mgr manager.Manager, datavolumeController co
 	if err := datavolumeController.Watch(&source.Kind{Type: &cdiv1.DataVolume{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


brybacki:
I can see that you changed the watcher to also get PVCs with `populatedFor`. 
But what was the actual problem, and what is actually fixed. Can you improve the description of the PR?

mhenriks:
> I can see that you changed the watcher to also get PVCs with `populatedFor`. But what was the actual problem, and what is actually fixed. Can you improve the description of the PR?

Sorry, thought it was pretty obvious if you look at the difference to the functional test.  The reason the test worked before is because the PVC was owned by the DataVolume.  But that will not always be the case (at least initially).  Without this change the PVC (without owner ref) would leave the DV pending forever.

mhenriks:
/hold
while I figure out test failures with GC enabled

PR #2375, opened 2022-07-26
Check for pvc populated before doing any clone validations and actions

ShellyKa13:
In case our pvc is already populated no need to check for source PVC
unknown size etc.

**Missing tests**

Fixes #
This should fix bug: https://bugzilla.redhat.com/show_bug.cgi?id=2104479

**Release note**:

```release-note
Fix clone datavolume with populated target PVC without source 
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -194,7 +194,7 @@ const (
 	// CloneWithoutSource reports that the source PVC of a clone doesn't exists (reason)
 	CloneWithoutSource = "CloneWithoutSource"
 	// MessageCloneWithoutSource reports that the source PVC of a clone doesn't exists (message)
-	MessageCloneWithoutSource = "The source PVC doesn't exist"
+	MessageCloneWithoutSource = "The source PVC %s doesn't exist"
 
 	// AnnCSICloneRequest annotation associates object with CSI Clone Request
 	AnnCSICloneRequest = "cdi.kubevirt.io/CSICloneRequest"
@@ -663,17 +663,21 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 	prePopulated bool,
 	pvcPopulated bool) (reconcile.Result, error) {
 
+	// Get the most appropiate clone strategy
+	selectedCloneStrategy, err := r.selectCloneStrategy(datavolume, pvcSpec)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if pvcPopulated || prePopulated {
+		return r.reconcileDataVolumeStatus(datavolume, pvc, selectedCloneStrategy)
+	}
+
 	// Check if source PVC exists and do proper validation before attempting to clone
 	if done, err := r.validateCloneAndSourcePVC(datavolume); err != nil {
 		return reconcile.Result{}, err
 	} else if !done {
 		return reconcile.Result{}, nil
-	}
-
-	// Get the most appropiate clone strategy
-	selectedCloneStrategy, err := r.selectCloneStrategy(datavolume, pvcSpec)
-	if err != nil {
-		return reconcile.Result{}, err
 	}
 
 	if selectedCloneStrategy == SmartClone {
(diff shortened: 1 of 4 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -808,6 +806,9 @@ func (r *DatavolumeReconciler) ensureExtendedToken(pvc *corev1.PersistentVolumeC
 func (r *DatavolumeReconciler) selectCloneStrategy(datavolume *cdiv1.DataVolume, pvcSpec *corev1.PersistentVolumeClaimSpec) (cloneStrategy, error) {
 	preferredCloneStrategy, err := r.getCloneStrategy(datavolume)
 	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return NoClone, nil
```
mhenriks, line 810: maybe it would be better if `getCloneStrategy()` handles this case?

ShellyKa13: what should be returned in that case? preferredCloneStrategy=nil and err=nil? I would still need to add here a check if preferredCloneStrategy=nil and return NoClone. or create a new CDICloneStrategy-> cdiv1.CloneStrategyNoClone? I would still need to add here a check of if preferredCloneStrategy=cdiv1.CloneStrategyNoClone and return NoClone?
I dont mind doing either but doesnt seem any cleaner or simpler then this

pkg/controller/datavolume-controller.go
```diff
@@ -2836,6 +2836,7 @@ func (r *DatavolumeReconciler) validateCloneAndSourcePVC(datavolume *cdiv1.DataV
 	if err != nil {
 		// Clone without source
 		if k8serrors.IsNotFound(err) {
+			r.recorder.Eventf(datavolume, corev1.EventTypeWarning, ErrUnableToClone, "Source pvc %s not found", datavolume.Spec.Source.PVC.Name)
```
alromeros, line 2839: Overall looks very good, but wouldn't this event be a little bit redundant? We are just recording a very similar event in the line below

alromeros (commented):
Looks good!

pkg/controller/datavolume-controller.go
ShellyKa13: I just moved it from the get clone strategy to here since it didnt seem related there, you added the event so you proabably know best if its redundant, if it is Ill delete :)

alromeros: The first event says `"Source pvc %s not found"` while the second one is `"The source PVC doesn't exist"`, so I think they are too similar to record them at the same time. Maybe it's a good idea to merge the two so the second event also displays the PVC name, but I'd just use one.

pkg/controller/datavolume-controller.go
```diff
@@ -1904,7 +1905,6 @@ func (r *DatavolumeReconciler) getCloneStrategy(dataVolume *cdiv1.DataVolume) (*
 	sourcePvc, err := r.findSourcePvc(dataVolume)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
```
alromeros, line 1907: Just in case, you left an empty condition here.

ShellyKa13: right thanks :)

pkg/controller/datavolume-controller.go
ShellyKa13: done

tests/cloner_test.go
```diff
@@ -1265,6 +1266,29 @@ var _ = Describe("all clone tests", func() {
...
+
+			It("Should not clone when PopulatedFor annotation exists", func() {
+				targetName := "target" + rand.String(12)
+
+				By(fmt.Sprintf("Creating target pvc: %s/%s", f.Namespace.Name, targetName))
+				targetPvc, err := utils.CreatePVCFromDefinition(f.K8sClient, f.Namespace.Name, utils.NewPVCDefinition(
+					targetName,
+					"1Gi",
+					map[string]string{
+						controller.AnnPodPhase:     "Succeeded",
```
mhenriks, line 1278: What happens if this annotation is missing?

ShellyKa13: checked it, the test still passes, should I remove this annotation from the test?

ShellyKa13: removed

pkg/controller/datavolume-controller.go
brybacki: If this behaves correctly then I am ok with the rest.

alromeros: @brybacki I think it behaves as expected: If the source PVC doesn't exist but the target is already populated, we are just updating the DV without triggering any errors. However, if the source PVC doesn't exist and the target is not populated, it'll be handled by `validateCloneAndSourcePVC`, which will exit without errors to retry once the source is created:

https://github.com/kubevirt/containerized-data-importer/blob/083ee010753fea2d2d7e004e8afd48ea3bd90fe4/pkg/controller/datavolume-controller.go#L673-L682

brybacki: ok

ShellyKa13:
/hold
checking if it causes a test flakiness

ShellyKa13:
/unhold
the test flake I saw is not because of my changes:
[test_id:6612]Clone with CSI as PVC source with target name that already exists
failed on timeout to get the clone type annotation on the dv
It will fail in case the target pvc is created before the data volume is being reconciled at least once, 
in that case we will fail on the check of the pvc already exists before we will get to the clone reconcile which there we update the clone type on the datavolume annotation

ShellyKa13:
opened a fix for the test flake: https://github.com/kubevirt/containerized-data-importer/pull/2386

PR #2381, opened 2022-08-04
Update DataSource LastTransitionTime when populated PVC is updated

arnongilboa:

The `DataSourceReady` condition `LastTransitionTime` does not show a timestamp of the last source PVC update. It only shows a timestamp for its initial population, although the PVC name is updated.

Fixes #

**Release note**:
```release-note
Update DataSource LastTransitionTime when populated source PVC is updated
```


awels:
So it looks like there is no good way of detecting that the current PVC has changed, which in turn leads us to doing what you did. Would it make sense to add a `currentSourcePVC` that contains the `DataVolumeSourcePVC` of the current source. Then if the spec changes, we can detect this change because the status doesn't match the spec, we can then clear out the ready condition and everything else takes care of itself.

arnongilboa:
> So it looks like there is no good way of detecting that the current PVC has changed, which in turn leads us to doing what you did. Would it make sense to add a `currentSourcePVC` that contains the `DataVolumeSourcePVC` of the current source. Then if the spec changes, we can detect this change because the status doesn't match the spec, we can then clear out the ready condition and everything else takes care of itself.

so why not add a mutating webhook to reset the Ready condition whenever source PVC is updated?

mhenriks:
I think a new status field is a better solution.  If the resource never transitioned from not ready to ready, the transition time should not change.

arnongilboa:
> I think a new status field is a better solution. If the resource never transitioned from not ready to ready, the transition time should not change.

added

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -443,6 +443,8 @@ type DataSourceSource struct {
 
 // DataSourceStatus provides the most recently observed status of the DataSource
 type DataSourceStatus struct {
+	// Source is the current source of the data referenced by the DataSource
+	Source     DataSourceSource      `json:"source"`
```
akrejcir, line 447: Making this field required in the JSON, breaks API backward compatibility. Can you consider marking it as `omitempty`?

arnongilboa: Sure, sorry I missed it.

arnongilboa: @akrejcir https://github.com/kubevirt/containerized-data-importer/pull/2382

Issue #1317, opened 2020-07-24
Getting "readlink: illegal option -- f" when building on macOS


visheshtanksale:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
Make fail with following error
```
$ make all
./hack/build/in-docker.sh "DOCKER_PREFIX= DOCKER_TAG= VERBOSITY= PULL_POLICY= NAMESPACE= ./hack/build/build-manifests.sh"
readlink: illegal option -- f
usage: readlink [-n] [file ...]
make: *** [manifests] Error 1
```
The `readlink` on macOS does not support `--f` option. 
**What you expected to happen**:
CDI should build cleanly.
**How to reproduce it (as minimally and precisely as possible)**:
Execute on macOS machine `make all` 
**Anything else we need to know?**:
Is macOS a support platform for building CDI. I can create a PR to replace using readlink with more platform independent command.

awels:
None of us use macs for development, so if you can provide a PR that will help, we are more than happy to look at it and accept it that makes it work for you.

PR #2383, opened 2022-08-09
Ability to run tests in D/S

mor120:
This PR will give us the ability to run tests in a D/S environment using the existing scripts.

Fixes #
In D/S we are working to run tests in containers instead of running them on VMs (executors).
We can't run docker in docker in D/S in order to build the tests so we are going to use the builder image directly and then copy the binaries to a container that will run the tests.

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/build-functest.sh
+++ b/hack/build/build-functest.sh
@@ -14,11 +14,11 @@
 
 set -euo pipefail
 
-script_dir="$(cd "$(dirname "$0")" && pwd -P)"
+script_dir=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
 source "${script_dir}"/common.sh
 
 mkdir -p ${TESTS_OUT_DIR}/
 # use vendor
-export GO111MODULE=off
+export ${GO111MODULE:-off}
 ginkgo build ${CDI_DIR}/tests/
 mv ${CDI_DIR}/tests/tests.test ${TESTS_OUT_DIR}/
--- a/hack/build/build-functest.sh
+++ b/hack/build/build-functest.sh
@@ -14,7 +14,7 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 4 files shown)


hack/build/build-functest.sh
```diff
@@ -14,11 +14,11 @@
...
 set -euo pipefail
 
-script_dir="$(cd "$(dirname "$0")" && pwd -P)"
+script_dir=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
 source "${script_dir}"/common.sh
 
 mkdir -p ${TESTS_OUT_DIR}/
 # use vendor
-export GO111MODULE=off
+export ${GO111MODULE:-off}
```
lukas-bednar, line 22: Shouldn't it be like the following?

```
export GO111MODULE=${GO111MODULE:-off}
```

maya-r:
Can you explain why script_dir is changed? we're avoiding readlink -f because of https://github.com/kubevirt/containerized-data-importer/issues/1317

/ok-to-test

hack/build/build-functest.sh
mor120: Totally, thanks!

mor120:
> Can you explain why script_dir is changed? we're avoiding readlink -f because of #1317
> 
> /ok-to-test

I don't really need it, I have reverted these changes, thanks :)

maya-r:
this looks good to me -- the bot is complaining about the lack of a "DCO signoff", the lack of a Signed-off-by: author text in the commit message. You can add it to a commit with `git commit --amend --sign`

akalenyu (commented):
I am probably cheating here because I looked at the downstream PRs, but I think it would be much better to have a make target for your use case that spares you any extras you have to do yourself.

@maya-r This is about containerizing the test runs of CDI without the ability to do docker-in-docker. So all we get is a worker pod in a cluster span up from the builder image

PR #2386, opened 2022-08-10
test fix: Move check of clone type before create target pvc

ShellyKa13:
The test is a negative test that creates the sourcepvc, create a pod to keep it busy, then creates the clone dv which can't do the clone yet since the source is busy. then it creates the targetpvc and expects eventually that the clone will fail since there already exists a pvc with the target name.
now the test after creating the target pvc checks for the clone type annotation on the dv. if the target pvc was created fast enough the dv will get to the check of the target pvc name (and return error since it exists) before getting to the reconcileClone function which there you update the clone type annotation. Thats why if the tests wants to check for the event of "source busy" which requires the clone type annotation on the dv it should be done before creating the target pvc.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -334,6 +334,15 @@ var _ = Describe("all clone tests", func() {
 				dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, targetNamespaceName, targetDV)
 				Expect(err).ToNot(HaveOccurred())
 
+				actualCloneType := utils.GetCloneType(f.CdiClient, dataVolume)
+				if actualCloneType == "snapshot" {
+					f.ExpectEvent(targetNamespaceName).Should(ContainSubstring(controller.SmartCloneSourceInUse))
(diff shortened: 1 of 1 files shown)


brybacki:
I am not sure I understand the PR description, can you elaborate on this?

ShellyKa13:
@brybacki the test is a negative test that creates the sourcepvc, create a pod to keep it busy, then creates the clone dv which can't do the clone yet since the source is busy. then it creates the targetpvc and expects eventually that the clone will fail since there already exists a pvc with the target name.
now the test after creating the target pvc checks for the clone type annotation on the dv. if the target pvc was created fast enough the dv will get to the check of the target pvc name (and return error since it exists) before getting to the reconcileClone function which there you update the clone type annotation. Thats why if the tests wants to check for the event of "source busy" which requires the clone type annotation on the dv it should be done before creating the target pvc.
I didnt want to go to that much details.. and if one will go over the test and read the commit msg I think that enough

PR #2388, opened 2022-08-10
Fix flake where cross ns finalizer can't be removed

akalenyu:

We fail to remove the finalizer because the mutating webhook validates source ns exists,
but source ns might not be around anymore, and that is absolutely fine if DV is marked for deletion.
The "offending" test case is `[test_id:4960]`.

Fixes #

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2363/pull-containerized-data-importer-e2e-ceph-gc/1552213098473984000
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2381/pull-containerized-data-importer-e2e-ceph-gc/1555211218673733632
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2371/pull-containerized-data-importer-e2e-ceph-gc/1553016969739898880

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-mutate.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate.go
@@ -75,6 +75,11 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		return toAdmissionResponseError(err)
 	}
 
+	if dataVolume.GetDeletionTimestamp() != nil {
+		// No point continuing if DV is flagged for deletion
+		return allowedAdmissionResponse()
+	}
+
 	if dataVolume.Spec.Source != nil {
 		pvcSource = dataVolume.Spec.Source.PVC
 	} else if dataVolume.Spec.SourceRef != nil && dataVolume.Spec.SourceRef.Kind == cdiv1.DataVolumeDataSource {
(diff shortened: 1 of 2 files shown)


pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -130,9 +130,12 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
 
-	_, err := wh.k8sClient.CoreV1().Namespaces().Get(context.TODO(), sourceNamespace, metav1.GetOptions{})
-	if err != nil {
-		return toAdmissionResponseError(err)
+	if dataVolume.GetDeletionTimestamp() == nil {
+		// No point validating source ns exists if DV is flagged for deletion
+		_, err := wh.k8sClient.CoreV1().Namespaces().Get(context.TODO(), sourceNamespace, metav1.GetOptions{})
+		if err != nil {
+			return toAdmissionResponseError(err)
+		}
```
mhenriks, line 138: Maybe just return success if deletion timestamp set.  No need to bother with the rest of this function, which I think is just about setting the clone token

akalenyu: Done

awels:
How about a unit test making sure this is working as expected, like `Should always accept mutate when dv is marked for deletion`?

akalenyu:
> How about a unit test making sure this is working as expected, like `Should always accept mutate when dv is marked for deletion`?

good idea, done

awels:
/lgtm
@mhenriks any other thoughts?

PR #2399, opened 2022-08-17
Simplify test PVC creation & force-binding code

maya-r:
**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -118,9 +118,6 @@ var _ = Describe("all clone tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 			f.ForceBindPvcIfDvIsWaitForFirstConsumer(targetDataVolume)
 
-			By("Wait for clone to be completed")
-			_, err = utils.WaitForPVC(f.K8sClient, targetDataVolume.Namespace, targetDataVolume.Name)
-			Expect(err).ToNot(HaveOccurred())
 			By("Wait for target datavolume phase Succeeded")
 			utils.WaitForDataVolumePhaseWithTimeout(f, targetDataVolume.Namespace, cdiv1.Succeeded, targetDV.Name, cloneCompleteTimeout)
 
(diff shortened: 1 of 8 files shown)


tests/framework/pvc.go
```diff
@@ -35,9 +35,13 @@ func (f *Framework) CreatePVCFromDefinition(def *k8sv1.PersistentVolumeClaim) (*
 // on WaitForFirstConsumer storage class by executing f.ForceBindIfWaitForFirstConsumer(pvc)
 func (f *Framework) CreateBoundPVCFromDefinition(def *k8sv1.PersistentVolumeClaim) (*k8sv1.PersistentVolumeClaim, error) {
 	pvc, err := utils.CreatePVCFromDefinition(f.K8sClient, f.Namespace.Name, def)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	if err != nil {
```
awels, line 38: Does it make sense to keep the Expects here, and simply not return an err at all? Less code everywhere else to check the err.

maya-r: Done

tests/cloner_test.go
```diff
@@ -118,9 +118,6 @@ var _ = Describe("all clone tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 			f.ForceBindPvcIfDvIsWaitForFirstConsumer(targetDataVolume)
 
-			By("Wait for clone to be completed")
-			_, err = utils.WaitForPVC(f.K8sClient, targetDataVolume.Namespace, targetDataVolume.Name)
-			Expect(err).ToNot(HaveOccurred())
```
brybacki, line 123: Maybe it is not needed, since we already wait for DV.

But I think it has nothing to do with force binding.So I do not understand your commit message. Force Bind only works for WFFC storage, and when run on Immediate storage the function forceBind works like a noop.

brybacki: anyway, this comments is only about a commit message that is not adequate to change

brybacki:
/lgtm
one small comment about a commit message ;)

maya-r:
Oops, I didn't notice @brybacki lgtm'd it as well and it only failed on a test :)
I changed the commit message as requested, no code changes.

brybacki:
/lgtm
thanks

PR #2366, opened 2022-07-21
[release-v1.49] Manual backport of 'Allow creating clones without source PVC (#2306)' 

alromeros:
**What this PR does / why we need it**: 

Manual backport of #2306 

**Release note**:

```release-note
NONE
```

PR #2406, opened 2022-08-22
[release-v1.49] Manual fix for clone without source pvc but target pvc already populated

ShellyKa13:
This PR takes part of PR: https://github.com/kubevirt/containerized-data-importer/pull/2306
and PR: https://github.com/kubevirt/containerized-data-importer/pull/2375
to fix this issue with minimal changes.
In case the target pvc exists and populated we should allow the clone to complete successfully without doing anything

Fixes # BZ #2109407
https://bugzilla.redhat.com/show_bug.cgi?id=2109407

**Release note**:

```release-note
Fix for clone without source pvc but target pvc already populated
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -480,18 +480,10 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
-	selectedCloneStrategy, err := r.selectCloneStrategy(datavolume, pvcSpec)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-	if selectedCloneStrategy == SmartClone {
-		r.sccs.StartController()
-	}
-
 	_, dvPrePopulated := datavolume.Annotations[AnnPrePopulated]
 
-	if selectedCloneStrategy != NoClone {
-		return r.reconcileClone(log, datavolume, pvc, pvcSpec, transferName, dvPrePopulated, pvcPopulated, selectedCloneStrategy)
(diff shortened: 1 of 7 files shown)


pkg/controller/datavolume-controller.go
```diff
@@ -541,10 +533,30 @@ func (r *DatavolumeReconciler) reconcileClone(log logr.Logger,
 	pvcSpec *corev1.PersistentVolumeClaimSpec,
 	transferName string,
 	prePopulated bool,
-	pvcPopulated bool,
-	selectedCloneStrategy cloneStrategy) (reconcile.Result, error) {
+	pvcPopulated bool) (reconcile.Result, error) {
 
+	_, err := r.findSourcePvc(datavolume)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
brybacki, line 540: I wonder why there is a difference here vs the same lines in https://github.com/kubevirt/containerized-data-importer/pull/2405/commits/5686b4f1318a8bdce2f1f1b4899afa64082eb080

brybacki:
@ShellyKa13 I think you need to coordinnate your change with @alromeros. If https://github.com/kubevirt/containerized-data-importer/pull/2366 is merged, then your change could be smaller and simpler.

pkg/controller/datavolume-controller.go
ShellyKa13: in this version the !pvcPopulated was removed from the if in line 550. that code runs even if pvc populated so I needed to prevent it from running only if source pvc not found.

brybacki: ok

PR #2407, opened 2022-08-24
Remove v1alpha1 as a stored version from our CRDs

maya-r:
It's been two years since v1beta1 became the stored version. Time to retire it.
If we detect it is a stored version (the cluster ever had v1alpha1 with storage: true), copy the old CRD definition for v1alpha1.
Test this scenario in a functional test by re-adding v1alpha1 to the datavolume CRD.

**Release note**:

```release-note
Stop serving v1alpha1 CRDs (except CDI CRD). v1beta1 has been the storage version since CDI v1.20.0.
CDI CRD v1alpha1 version remains available, so updating clusters that had versions of CDI before v1.20.0 by applying the release YAMLs will just work.
```


maya-r:
cc @mhenriks

mhenriks:
@maya-r I don't think a user should have to manually update the stored version.  Why can't cdi-operator do "option 2" as described here:  https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version

FYI:  I'm not sure that step 2 is necessary since we don't have a conversion webhook.  Which gets me to testing.

Pretty sure you will need additional RBAC for step 3 (update crd/status)

Without a dedicated lane, I'm not sure this pr is a good fit for our functional test suite.  But I'd like to see something, even if it is just a stand alone bash script, that installs very old cdi with only alpha api, creates some resources, updates to current build, confirms "old" resources correctly retrieved with new version

maya-r:
The reason that I want the user to do a manual operation is that we have two kinds of CRDs:

1. datavolume/cdiconfig/... and other CRDs created by CDI: they will retain the v1alpha1 version and no manual action needed.
2. CDI CRD. Since it's created by the user and not CDI I don't want to modify this CRD.

But if a user is upgrading by re-applying operator.yaml, it will try to modify the CDI CRD to not have a v1alpha1 version which will fail (in the scenario of updating from a cluster that had CDI<1.20.0 installed at any point). That is the only CRD that requires manual action.

mhenriks:
> The reason that I want the user to do a manual operation is that we have two kinds of CRDs:
> 
> 1. datavolume/cdiconfig/... and other CRDs created by CDI: they will retain the v1alpha1 version and no manual action needed.

Since no longer serving v1alpha1 via k8s apiserver, resources should not be stored as v1alpha1.  We have total control over these, so why not do the "right" thing?

> 2. CDI CRD. Since it's created by the user and not CDI I don't want to modify this CRD.
> 

This is more complicated because have to leave v1alpha1 in the yaml (for now) but once started, cdi-operator can update stored version just like it can do for everything else.  

Then after awhile (six months or so) we can remove v1alpha1 from cdi-operator.yaml (forcing people to uninstall/reinstall)

maya-r:
@mhenriks the new changes do what you asked for - could you have a look?

mhenriks:
@maya-r, finally got around to reviewing/testing this.  I wanted to test upgrading from 1.19 to this version and what a journey it was.  Is that something you tried?  Here's what I did:

1.  `KUBEVIRT_PROVIDER=k8s-1.21` is required because 1.19 still uses v1beta1 CRDs
2. make cluster-up
3. apply 1.19 release manifests
4. Create blank v1alpha1 DV
5. Build/Push this version to registry (with modifications, see below)
6. Build manifests for pushed images
7. Apply `_out/manifests/release/cdi-operator.yaml`
8. See following error in log:

```
{"level":"error","ts":1665085362.1959324,"logger":"cdi-operator","msg":"error invoking callback for","type":"*v1.CustomResourceDefinition","error":"CustomResourceDefinition.apiextensions.k8s.io \"datavolumes.cdi.kubevirt.io\" is invalid: spec.versions: Invalid value: []apiextensions.CustomResourceDefinitionVersion{apiextensions.CustomResourceDefinitionVersion{Name:\"v1alpha1\", Served:true, Storage:false, Deprecated:false, DeprecationWarning:(*string)(nil), Schema:(*apiextensions.CustomResourceValidation)(nil), Subresources:(*apiextensions.CustomResourceSubresources)(nil), AdditionalPrinterColumns:[]apiextensions.CustomResourceColumnDefinition(nil)}}: must have exactly one version marked as storage version","stacktrace":"kubevirt.io/containerized-data-importer/vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/callbacks.(*CallbackDispatcher).InvokeCallbacks\n\tvendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/callbacks/callback-dispatcher.go:157\nkubevirt.io/containerized-data-importer/vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/reconciler.(*Reconciler).InvokeCallbacks\n\tvendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/reconciler/reconciler.go:549\nkubevirt.io/containerized-data-importer/vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/reconciler.(*Reconciler).ReconcileUpdate\n\tvendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/reconciler/reconciler.go:293\nkubevirt.io/containerized-data-importer/vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/reconciler.(*Reconciler).Reconcile\n\tvendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/reconciler/reconciler.go:185\nkubevirt.io/containerized-data-importer/pkg/operator/controller.(*ReconcileCDI).Reconcile\n\tpkg/operator/controller/controller.go:236\nkubevirt.io/containerized-data-importer/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114\nkubevirt.io/containerized-data-importer/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:311\nkubevirt.io/containerized-data-importer/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266\nkubevirt.io/containerized-data-importer/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227"}
```

Error summed up is:  `must have exactly one version marked as storage version`

* In order for current code to run on k8s 1.21, had to remove all calls to set `seccompProfile`.  Should probably (in another pr, make it possible to turn off the restricted context stuff)

maya-r:
@mhenriks with the new commit it's handling the case of updating directly from 1.19, I had to awkwardly re-add the alpha version to be able to kubectl get DVs, but it seems fine

pkg/operator/controller/cruft.go
```diff
@@ -186,6 +192,145 @@ func reconcileServiceAccounts(args *callbacks.ReconcileCallbackArgs) error {
...
+	if oldVersion != nil {
+		oldVersion.Storage = false
+	}
+	if newVersion == nil {
+		_, newVersion = getSpecVersions(newCrd)
+		realCrd.Spec.Versions = append(realCrd.Spec.Versions, *newVersion)
+	}
+	newVersion.Storage = true
+	if !reflect.DeepEqual(realCrdCopy, realCrd) {
+		err := args.Client.Update(context.TODO(), realCrd)
```
mhenriks, line 221: The callback code shouldn't have to update the resource directly since the reconcile loop is going to do that.  Can either update the "desiredObject" on "post read" callback or the "currentObject" on pre update.

pkg/operator/controller/cruft.go
```diff
@@ -186,6 +192,145 @@ func reconcileServiceAccounts(args *callbacks.ReconcileCallbackArgs) error {
...
+	crd.Spec.Versions = newVersions
+	return crd
+}
+
+// Handle upgrade from clusters that had v1alpha1 as a storage version
+// and remove it from all CRDs managed by us
+func reconcileHandleOldVersion(args *callbacks.ReconcileCallbackArgs) error {
+	var err error
+
+	if args.State != callbacks.ReconcileStatePreUpdate {
```
mhenriks, line 302: I would consider making this a "post read" callback that does different things based on the value of `args.currentObject` and `args.desiredObject`.

With a "post read" callback, currentObject is exactly the current object (has not been merged with desired yet)

1.  If there are stored versions not in desiredObject, add them to desired from current making sure they are not "stored" (desired has the current "stored" version)
2. If currentObject is missing any version in desired, exit (they will be added)
3. If stored versions is not a single value that is the most recent "desired" version, rewrite old objects and update CRD status stored version to be latest desired

Once the stored version is the latest version, this function should get called and do none of the above and that will lead to any versions not in desiredObject being deleted

mhenriks (commented):
Great start, suggestions about moving to "post read" and making the update procedure more generic to accommodate eventual move to v1 stored version

maya-r:
I've written the code as a post read reconcile, and removed the hard coding.
Let me know what you think.

alromeros (commented):
I've been trying to review this PR for some time and after the latest changes, I don't have any real objection to it. This is not an easy task and I really like what I'm seeing, kudos to you! That said, I feel like this PR won't benefit that much from my review since it requires a certain expertise in the field, and there are too many things that could've flown over my head. Pinging @mhenriks so we can have some comments after the latest commits.

pkg/operator/controller/cruft.go
```diff
@@ -186,6 +188,132 @@ func reconcileServiceAccounts(args *callbacks.ReconcileCallbackArgs) error {
 	return nil
 }
 
+func getSpecVersion(version string, crd *extv1.CustomResourceDefinition) *extv1.CustomResourceDefinitionVersion {
```
mhenriks, line 191: I think is makes sense to move all these changes to callbacks.go as this is not "cruft" that will eventually get deleted.

mhenriks (commented):
Looks good!  tested upgrading from 1.19 as described here:  https://github.com/kubevirt/containerized-data-importer/pull/2407#issuecomment-1270604527

pkg/operator/controller/cruft.go
maya-r: Done!

pkg/operator/controller/callbacks.go
```diff
@@ -253,3 +256,129 @@ func reconcileSetConfigAuthority(args *callbacks.ReconcileCallbackArgs) error {
...
+			return nil
+		}
+		if err := rewriteOldObjects(args, desiredVersion, currentCrd); err != nil {
+			return err
+		}
+		if err := removeStoredVersion(args, desiredVersion, currentCrd); err != nil {
+			return err
+		}
+	} else {
+		desiredCrd = removeOldVersions(desiredVersion, desiredCrd)
```
mhenriks, line 339: I think this will be problematic when we support v1 and v1beta1.  I think we can simply remove this else block?

maya-r: This code block exists because the generated CRD still includes v1alpha1 versions, for the purpose of producing release YAMLs that contain v1alpha1 (also: it's several objects because CDI has a CDIConfig field).

mhenriks: desiredCrd should not include v1alpha1 in the `else` branch. It is added in the `if` branch.  

Also removeOldVersions is not compatible with serving multiple apiversions.  We will soon want to serve v1 and v1beta1.

maya-r: Any reason that we would want to serve both v1 and v1beta1 and not just migrate to v1 the moment it is introduced using this migration path? it should work just the same.

mhenriks: When new versions are introduced, it is common practice to keep the last version around for a while before removing it.  Want existing user yamls to work

maya-r: Here are different things I can do:

1. Remove v1alpha1 from the generated CRD, so this code can be removed. The downside is that someone upgrading a cluster by applying release YAMLs all the way from 1.20.0 will need to remove v1alpha1 as a stored version from the CDI CRD (only this CRD). I provided the command to do so.
2. Make this logic specific to v1alpha1 (hard-coded), so it is not done for v1beta1 once we introduce v1.

Any preference?

mhenriks: Is there any way the v1alpha1 version can just be added to the CDI CRD (not DataVolume, etc)?  That way this `else` block can be deleted.  cdi-operator does not do anything with the CDI CRD

maya-r:
The rebase is dropping the commit that changes unit tests - this is no longer necessary as we don't generate another event from modifying the CRDs to remove v1alpha1 version.

pkg/operator/controller/callbacks.go
maya-r: Done

maya-r:
/hold
I just realized that kubevirt virtctl image upload uses alpha paths

maya-r:
Rebased commits to fix build issue and also used the opportunity to squash related commits.

pkg/operator/controller/callbacks.go
```diff
@@ -253,3 +256,127 @@ func reconcileSetConfigAuthority(args *callbacks.ReconcileCallbackArgs) error {
...
+	return nil
+}
+
+func removeStoredVersion(args *callbacks.ReconcileCallbackArgs, desiredVersion string, crd *extv1.CustomResourceDefinition) error {
+	args.Logger.Info("Removing stored version")
+	crd.Status.StoredVersions = []string{desiredVersion}
+	return args.Client.Status().Update(context.TODO(), crd)
+}
+
+func removeOldVersions(desiredVersion string, crd *extv1.CustomResourceDefinition) *extv1.CustomResourceDefinition {
```
mhenriks, line 305: This function is unused

mhenriks (commented):
Looks good, one comment

mhenriks:
@maya-r this seems to be missing.  What are your thoughts?

>> CDI CRD. Since it's created by the user and not CDI I don't want to modify this CRD.

>This is more complicated because have to leave v1alpha1 in the yaml (for now) but once started, cdi-operator can update stored version just like it can do for everything else.

>Then after awhile (six months or so) we can remove v1alpha1 from cdi-operator.yaml (forcing people to uninstall/reinstall if they are still on ancient v1alpah1 only version)

pkg/operator/controller/callbacks.go
maya-r: Thanks, removed

maya-r:
> @maya-r this seems to be missing. What are your thoughts?
> 
> > > CDI CRD. Since it's created by the user and not CDI I don't want to modify this CRD.
> 
> > This is more complicated because have to leave v1alpha1 in the yaml (for now) but once started, cdi-operator can update stored version just like it can do for everything else.
> 
> > Then after awhile (six months or so) we can remove v1alpha1 from cdi-operator.yaml (forcing people to uninstall/reinstall if they are still on ancient v1alpah1 only version)

I've attempted to modify the CDI CRD to remove the alpha version, but failed as we don't reconcile on changes to the CDI CRD, and my attempts to watch for changes on it have failed.

It might not be worth doing something considering that this refers to people who install using YAMLs all the way from old versions and expect it to work seamlessly - it won't once the alpha version is removed from the CDI CRD (not done in the latest commit), but we have a documented workaround.

mhenriks:
> I've attempted to modify the CDI CRD to remove the alpha version, but failed as we don't reconcile on changes to the CDI CRD, and my attempts to watch for changes on it have failed.

Did you try adding a watch in this function:  https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/operator/controller/reconciler-hooks.go#L20

A watch on CRDs with a predicate to filter on CRD name would probably do the trick

>It might not be worth doing something considering that this refers to people who install using YAMLs all the way from old versions and expect it to work seamlessly - it won't once the alpha version is removed from the CDI CRD (not done in the latest commit), but we have a documented workaround.

What is the work around?  And where is/will it be documented?  Why not get of v1aplha1 for CDI resources now then?

maya-r:
OK, I'm trying do this.
What do you think about merging this code as-is and doing the additional work in another PR? as it stands, it's not going to break upgrade by applying YAMLs. We can follow up with a second PR for the cleanup of the alpha version on the CDI CRD.

PR #2409, opened 2022-08-25
Always set a valid ContentType for created PVC

arnongilboa:

When DV is garbage collected and PVC ContentType is not correctly set, there is no way to guess the PVC ContetType. In CDI we assume that if there is no content type it's a kubevirt disk, but is other cases (e.g. export) we cannot make that assumption.

Fixes #

**Release note**:
```release-note
NONE
```


awels:
Thanks!
/lgtm
/approve

PR #2414, opened 2022-08-29
Make buildcmd choice a tunable

maya-r:
To be merged after https://github.com/kubevirt/project-infra/pull/2301
This PR is intended so our CI can override the build command chosen and use buildah despite otherwise being docker-based at the moment.
This allows us to build arm64 builder images.

Once we change the default builder, we can call #2211 fixed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/common.sh
+++ b/hack/build/common.sh
@@ -45,7 +45,7 @@ ARCHITECTURE="${BUILD_ARCH:-$(uname -m)}"
 HOST_ARCHITECTURE="$(uname -m)"
 CDI_CRI="$(determine_cri_bin)"
 if [ "${CDI_CRI}" = "docker" ]; then
-   CDI_CONTAINER_BUILDCMD="docker"
+   CDI_CONTAINER_BUILDCMD=${CDI_CONTAINER_BUILDCMD:-docker}
 else
-   CDI_CONTAINER_BUILDCMD="buildah"
+   CDI_CONTAINER_BUILDCMD=${CDI_CONTAINER_BUILDCMD:-buildah}
 fi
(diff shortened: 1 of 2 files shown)


hack/build/common.sh
```diff
@@ -45,7 +45,7 @@ ARCHITECTURE="${BUILD_ARCH:-$(uname -m)}"
 HOST_ARCHITECTURE="$(uname -m)"
 CDI_CRI="$(determine_cri_bin)"
 if [ "${CDI_CRI}" = "docker" ]; then
-   CDI_CONTAINER_BUILDCMD="docker"
+   CDI_CONTAINER_BUILDCMD=${CDI_CONTAINER_BUILDCMD:=docker}
```
brybacki, line 48: I do not think you need to get default and mutate variable `:=`, if you explicitely assignt it

`CDI_CONTAINER_BUILDCMD=${CDI_CONTAINER_BUILDCMD:-docker}` 
is probably the same as 
`${CDI_CONTAINER_BUILDCMD:=docker}` 

And this version does explicit assignment and then special default assignment, so it is redundant and surprising to read.
`CDI_CONTAINER_BUILDCMD=${CDI_CONTAINER_BUILDCMD:=docker}`

brybacki:
Why does it need to wait after https://github.com/kubevirt/project-infra/pull/2301? Do you want to test the lanes in this PR? 

Otherwise it does not look like anything would brake.

maya-r:
I want to build a multi-arch builder after this commit :)

hack/build/common.sh
maya-r: changed it to use :-, if this is what you meant

Issue #1988, opened 2021-10-12
[flaky] cdi-uploadproxy-nodeport" provided port is already allocated


brybacki:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
> /kind enhancement

**What happened**:

In the test run https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/1986/pull-containerized-data-importer-e2e-k8s-1.20-upg/1447744882301997056

Cluster did not start cleanly: The Service "cdi-uploadproxy-nodeport" is invalid: spec.ports[0].nodePort: Invalid value: 31001: provided port is already allocated

**What you expected to happen**:

 A port allocated without conflict.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

brybacki:
Similar problem:
https://github.com/istio/istio/issues/14499#issuecomment-500641742

akalenyu:
I think I see what's happening, some of our manifests let k8s randomize the nodeport, and maybe that puts
us in a situation where a clash is possible:
https://github.com/kubevirt/containerized-data-importer/blob/main/manifests/templates/imageio.yaml.in#L76-L81

(no `nodePort` field)

PR #2416, opened 2022-08-30
Do not add PVC BlockOwnerDeletion on DV GC

arnongilboa:

Eliminate errors like `persistentvolumeclaims SOME_PVC is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers`

Fixes #

**Release note**:
```release-note
Do not add PVC BlockOwnerDeletion on DV GC
```


pkg/controller/datavolume-controller.go
```diff
@@ -2823,14 +2823,27 @@ func (r *DatavolumeReconciler) detachPvcDeleteDv(pvc *corev1.PersistentVolumeCla
...
 	}
-	pvc.OwnerReferences = append(pvc.OwnerReferences, dv.OwnerReferences...)
+	for _, dvOwnerRef := range dv.OwnerReferences {
+		refExists := false
+		for _, pvcOwnerRef := range pvc.OwnerReferences {
+			if dvOwnerRef.UID == pvcOwnerRef.UID {
+				refExists = true
+				break
+			}
+		}
```
brybacki, line 2840: maybe extratc this as 
`findRef(pvc.OwnerReferences, dvOwnerRef.UID)` 
or 
`existsRef(dvOwnerRef.UID, pvc.OwnerReferences)`

pkg/controller/datavolume-controller.go
```diff
@@ -2823,14 +2823,27 @@ func (r *DatavolumeReconciler) detachPvcDeleteDv(pvc *corev1.PersistentVolumeCla
 }
 
 func updatePvcOwnerRefs(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) {
-	refs := pvc.OwnerReferences
-	for i, r := range refs {
-		if r.UID == dv.UID {
-			pvc.OwnerReferences = append(refs[:i], refs[i+1:]...)
+	var refs []metav1.OwnerReference
```
brybacki, line 2826: can it be `foreignRefs` or `otherRefs`,

brybacki:
You say "Eliminate errors like persistentvolumeclaims SOME_PVC is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers".

But I do not understand what does it mean. Why do we have such error?

brybacki:
Ok, so we are trying to copy ownerReferences from DV to PVC, before deleting DV. And we cannot copy `blockOwnerDeletion`. 

So why it was possible to have `blockOwnerDeletion` on DV and not on PVC? Don't we want to have it also on PVC?

pkg/controller/datavolume-controller.go
arnongilboa: It's not foreign or other, but the updated pvc ownerRefs, so I think the name is fine.

pkg/controller/datavolume-controller.go
arnongilboa: It's already a compact function, so extracting this loop to another func seems unnecessary.

arnongilboa:
> You say "Eliminate errors like persistentvolumeclaims SOME_PVC is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers".
> 
> But I do not understand what does it mean. Why do we have such error?

We have this error in OpenShift where VM and DV are created. However, the PVC is created by CDI which is not allowed to set finalizers of the VM, triggered when PVC is updated with its DV ownerRef to the VM with `blockOwnerDeletion: true`. This gets validated with [this](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) admission controller which is enabled in OpenShift, but disabled in our kubevirtci clusters.

arnongilboa:
> Ok, so we are trying to copy ownerReferences from DV to PVC, before deleting DV. And we cannot copy `blockOwnerDeletion`.
> 
> So why it was possible to have `blockOwnerDeletion` on DV and not on PVC? Don't we want to have it also on PVC?

See answer to your other comment. Adding it to the PR description.

pkg/controller/datavolume-controller.go
```diff
@@ -2823,14 +2823,27 @@ func (r *DatavolumeReconciler) detachPvcDeleteDv(pvc *corev1.PersistentVolumeCla
...
+	for _, dvOwnerRef := range dv.OwnerReferences {
+		refExists := false
+		for _, pvcOwnerRef := range pvc.OwnerReferences {
+			if dvOwnerRef.UID == pvcOwnerRef.UID {
+				refExists = true
+				break
+			}
+		}
+		if !refExists {
+			dvOwnerRef.BlockOwnerDeletion = nil
```
akalenyu, line 2842: maybe a comment above this with the admission plugin name and short explanation that we're taking ownerrefs that can be anything in the world?

akalenyu, line 2842: from https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/:
```
Kubernetes automatically sets this field to true if a [controller](https://kubernetes.io/docs/concepts/architecture/controller/) (for example, the Deployment controller) sets the value of the metadata.ownerReferences field.
```
I think we may be better with an explicit false value here, but maybe I got this wrong

pkg/controller/datavolume-controller.go
brybacki: but it will be obvious what it does without reading two levels of for loop and break statement;
```
for _, dvOwnerRef := range dv.OwnerReferences {
		if !refExists(dvOwnerRef.UID, pvc.OwnerReferences) {
           			dvOwnerRef.BlockOwnerDeletion = nil
			        refs = append(refs, dvOwnerRef)
        }
}
```

pkg/controller/datavolume-controller.go
brybacki: ok, maybe you are right, 
but after line 2829 it only contains refs that do not have given DV as an owner, that made me think about name

pkg/controller/datavolume-controller.go
brybacki: btw: this is square complexity, but I suppose we should never see more that a few references so it is ok.

brybacki:
> 

Nice description, thanks. 

Thinking more about the second part of my question:  _Don't we want to have `blockOwnerDeletion` also on PVC?_ 

I see that it is impossible. Does it mean we loose part of functionality - namely deletion protection - by doing DV GC?

arnongilboa:
> > 
> 
> Nice description, thanks.
> 
> Thinking more about the second part of my question: _Don't we want to have `blockOwnerDeletion` also on PVC?_
> 
> I see that it is impossible. Does it mean we loose part of functionality - namely deletion protection - by doing DV GC?

After DV is GC'ed, PVC is still protected from deletion:
```
  finalizers:
```

pkg/controller/datavolume-controller.go
arnongilboa: Sure, fixing.

pkg/controller/datavolume-controller.go
arnongilboa: Right, in the OpenShift case it's actually 1x1 :joy:
Anyway, I'll extract it as I really want to make you happy today, bro.

pkg/controller/datavolume-controller.go
arnongilboa: Added in the PR description. Looks ok?

arnongilboa: Sure, adding.

akalenyu: We should probably try to test this somehow that I am not lying :D
Maybe push the cdi-controller image to quay and we'll try to swap to it in a running cluster?

awels:
So the documentation says this about `blockOwnerDeletion: true`
```
Kubernetes also adds finalizers to an owner resource when you use either [foreground or orphan cascading deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion). In foreground deletion, it adds the foreground finalizer so that the controller must delete dependent resources that also have ownerReferences.blockOwnerDeletion=true before it deletes the owner.
```

For our purposes we use the ownerRefs for cascading deletion of the dependents and we don't really care about the blockOwnerDeletion. So right now if you have a VM with DVTemplates that created the DVs, and then the DVs created the PVCs, if you delete the VM, it also cascade deletes the DVs and PVCs. Now if we do a foreground delete of the VM, I guess it makes sure that the DVs and PVCs are deleted before the VM is deleted. I don't think we care about the order in which the objects are deleted in that scenario.

arnongilboa:
> So the documentation says this about `blockOwnerDeletion: true`
> 
> ```
> Kubernetes also adds finalizers to an owner resource when you use either [foreground or orphan cascading deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion). In foreground deletion, it adds the foreground finalizer so that the controller must delete dependent resources that also have ownerReferences.blockOwnerDeletion=true before it deletes the owner.
> ```
> 
> For our purposes we use the ownerRefs for cascading deletion of the dependents and we don't really care about the blockOwnerDeletion. So right now if you have a VM with DVTemplates that created the DVs, and then the DVs created the PVCs, if you delete the VM, it also cascade deletes the DVs and PVCs. Now if we do a foreground delete of the VM, I guess it makes sure that the DVs and PVCs are deleted before the VM is deleted. I don't think we care about the order in which the objects are deleted in that scenario.

I also think the deletion order is not critical here, so `blockOwnerDeletion=false` doesn't seem that risky.

mhenriks:
What is the minimal rbac that would be required to get this to work (keep block owner deletion) in the KubeVirt universe?  My preference would be go that route.  Then if we encounter a DV that is owned by some other resource (prob very rare in practice) we don't garbage collect it.

pkg/controller/datavolume-controller.go
akalenyu: awesome!

akalenyu:
> What is the minimal rbac that would be required to get this to work (keep block owner deletion) in the KubeVirt universe? My preference would be go that route. Then if we encounter a DV that is owned by some other resource (prob very rare in practice) we don't garbage collect it.

its `"virtualmachines/finalizers"` `update` if I'm not mistaken, but I am not sure I got 
>  Then if we encounter a DV that is owned by some other resource (prob very rare in practice) we don't garbage collect it

Wouldn't we GC it? I thought we would, but would just lose the order mechanism with blockOwnerDeletion set to false
If we go this route people from outside of CDI will need to submit patches for their `"specialoutsideofkubevirtresource/finalizers"` in our RBAC

arnongilboa:
> What is the minimal rbac that would be required to get this to work (keep block owner deletion) in the KubeVirt universe? My preference would be go that route. Then if we encounter a DV that is owned by some other resource (prob very rare in practice) we don't garbage collect it.

Not sure if I like this tightly coupled KubeVirt-CDI behavior and different/no-GC for other envs. Why keeping `blockOwnerDeletion: true` for PVC DV-inherited ownerRefs is that dramatic?

akalenyu:
https://github.com/kubevirt/containerized-data-importer/issues/1988

awels:
Just so I understand the problem properly. When KubeVirt created a VM with some DVTemplates, it creates the VM with an ownerRef of the VM, and it has `blockOwnerDeletion: true` on it. Now when we garbage collect the DV, we attempt to set the ownerRef of the PVC to be whatever is on the DV, and that is failing due to the ownerreferencepermissionenforement controller being enabled by default in 1.24 or 1.25?

There is no issue because when KubeVirt creates the DV it has the permissions to put a finalizer on the VM. But when CDI tries to set the same ownerRef it can't because it has no permissions to put a finalizer on the VM. Is there a possibility for us to push the responsibility of setting `blockOwnerDeletion: true` into KubeVirt (since it has the right RBAC to do so). Maybe have it detect PVCs without that and matching DVTemplates?

pkg/controller/datavolume-controller.go
```diff
@@ -2802,6 +2805,22 @@ func (r *DatavolumeReconciler) garbageCollect(dataVolume *cdiv1.DataVolume, pvc
 	return &reconcile.Result{}, nil
 }
 
+// Since PVC gets its DV ownerRefs which can be any CR, if BlockOwnerDeletion we allow GC only if we know CDI has RBAC
+// to update the owner finalizers (currently it's allowed only for VirtualMachines). BlockOwnerDeletions gets validated
+// with this admission plugin which is enabled in OpenShift, but disabled in our kubevirtci clusters:
+// https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+func isGarbageCollectionAllowed(dv *cdiv1.DataVolume, log logr.Logger) bool {
+	for _, dvOwnerRef := range dv.OwnerReferences {
+		if dvOwnerRef.Kind != "VirtualMachine" &&
```
aglitke, line 2814: I would suggest using the authorization APIs to check if you are able to update the finalizer for the parent object rather than checking for a specific type.  If you do this, then someone can make garbage collection work for any parent type so long as the provide the required rbac rules in the deployment.

akalenyu: > provide the required rbac rules in the deployment

As in, they would provide the RBAC to the cdi controller service account?

arnongilboa: Thanks @aglitke, now using authorization API.

pkg/controller/datavolume-controller.go
```diff
@@ -2824,13 +2865,24 @@ func (r *DatavolumeReconciler) detachPvcDeleteDv(pvc *corev1.PersistentVolumeCla
 
 func updatePvcOwnerRefs(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) {
```
brybacki, line 2866: This requires a unit test, and that will help us in the future if we ever need to change or refactor the code here.

arnongilboa: Sure, adding now.

pkg/controller/datavolume-controller.go
mhenriks: Great idea @aglitke!

PR #2418, opened 2022-09-05
Don't back out until we verify quota applied/removed in tests

akalenyu:

Currently we'll back out on various errors instead of waiting for proper cleanup,
which results in flakiness. Let's try to give any error a decent chance to go away.
Was partly implemented already in https://github.com/kubevirt/containerized-data-importer/pull/2197.

Fixes #

Returning err (`return false, err`) in polling funcs results in backing out of them instead of retrying

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -528,7 +528,8 @@ func (f *Framework) UpdateStorageQuota(numPVCs, requestStorage int64) error {
 	return wait.PollImmediate(5*time.Second, nsDeleteTime, func() (bool, error) {
 		quota, err := f.K8sClient.CoreV1().ResourceQuotas(f.Namespace.GetName()).Get(context.TODO(), "test-storage-quota", metav1.GetOptions{})
 		if err != nil {
-			return false, err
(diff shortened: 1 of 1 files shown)


tests/framework/framework.go
```diff
@@ -528,7 +528,8 @@ func (f *Framework) UpdateStorageQuota(numPVCs, requestStorage int64) error {
 	return wait.PollImmediate(5*time.Second, nsDeleteTime, func() (bool, error) {
 		quota, err := f.K8sClient.CoreV1().ResourceQuotas(f.Namespace.GetName()).Get(context.TODO(), "test-storage-quota", metav1.GetOptions{})
 		if err != nil {
-			return false, err
+			fmt.Fprintf(ginkgo.GinkgoWriter, "ERROR: GET ResourceQuota failed once, retrying: %v\n", err.Error())
```
mhenriks, line 531: I don't see how "failed once" is enforced?  Couldn't this fail multiple times?

akalenyu: It could, and that is what I am after. Right now without this PR we back out on a GET err

PR #2421, opened 2022-09-08
Enable DataVolume garbage collection by default

arnongilboa:

It's time to enable DataVolume garbage collection by default.

Fixes #

On the next kubevirt-bot Bump kubevirtci PR (with bump-cdi), it will fail on all kubevirtci lanes with tests referring DVs, as the tests IsDataVolumeGC() looks at CDIConfig Spec.DataVolumeTTLSeconds and assumes default is disabled. This should be fixed there.

**Release note**:
```release-note
Enable DataVolume garbage collection by default
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -4961,7 +4961,7 @@
     "type": "object",
     "properties": {
      "dataVolumeTTLSeconds": {
-      "description": "DataVolumeTTLSeconds is the time in seconds after DataVolume completion it can be garbage collected.",
+      "description": "DataVolumeTTLSeconds is the time in seconds after DataVolume completion it can be garbage collected. The default is 0 sec. To disable GC use -1.",
       "type": "integer",
       "format": "int32"
      },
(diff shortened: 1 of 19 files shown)


api/openapi-spec/swagger.json
```diff
@@ -5015,6 +5015,11 @@
     "description": "CDIConfigStatus provides the most recently observed status of the CDI Config resource",
     "type": "object",
     "properties": {
+     "dataVolumeTTLSeconds": {
+      "description": "DataVolumeTTLSeconds is the time in seconds after DataVolume completion it can be garbage collected.",
```
brybacki, line 5019: How it works when value is zero? does it mean immediate collection or disabled?

brybacki: ok, I can see in the test it is -1, I wonder where we can describe this fact

akalenyu (commented):
We can figure out what to do with that test case later
/lgtm

api/openapi-spec/swagger.json
```diff
@@ -4961,7 +4961,7 @@
     "type": "object",
     "properties": {
      "dataVolumeTTLSeconds": {
-      "description": "DataVolumeTTLSeconds is the time in seconds after DataVolume completion it can be garbage collected.",
+      "description": "DataVolumeTTLSeconds is the time in seconds after DataVolume completion it can be garbage collected. The default is 0 sec. To disable GC use -1.",
```
brybacki, line 4964: :+1:

brybacki (approved):
I do not see any problems, looks good

brybacki:
Maybe you can update the kubevirt test case proactively, before even new kubevirtci is introduced. 
For example one could add a logic that configures some default ttl (like -1) before the test suite or something like this, and then just change or remove the logic when we are ready.

I suppose the same way it work with HCO?

arnongilboa:
> Maybe you can update the kubevirt test case proactively, before even new kubevirtci is introduced. For example one could add a logic that configures some default ttl (like -1) before the test suite or something like this, and then just change or remove the logic when we are ready.
> 
> I suppose the same way it work with HCO?

Great idea, so simply put 0 instead of false here:
https://github.com/kubevirt/kubevirt/blob/main/hack/cluster-deploy.sh#L24

PR #2434, opened 2022-09-26
Change how we deploy the upgrade testing DVs

awels:

Only deploy the upgrade test datavolumes on multi upgrade before upgrading to latest. Deploy a different set on upgrade from latest released to contents of PR. This set will have an annotation to not garbage collect the DVs. Also dump some extra information if the DVs do not become ready in time.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/sync.sh
+++ b/cluster-sync/sync.sh
@@ -123,6 +123,16 @@ function wait_cdi_pods_updated {
   fi
 }
 
+function dump_upgrade_info {
+  echo "Dumping upgrade object information"
+  _kubectl get all --namespace cdi-testing-old-version-artifacts
+  _kubectl describe pvc olddv-v1alpha1 --namespace cdi-testing-old-version-artifacts
+  _kubectl describe pvc olddv-v1beta1 --namespace cdi-testing-old-version-artifacts
+  _kubectl get dv --namespace cdi-testing-old-version-artifacts -o yaml
+  _kubectl describe dv --namespace cdi-testing-old-version-artifacts
+  exit 1
+}
+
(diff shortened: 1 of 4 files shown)


cluster-sync/sync.sh
```diff
@@ -123,6 +123,16 @@ function wait_cdi_pods_updated {
   fi
 }
 
+function dump_upgrade_info {
+  echo "Dumping upgrade object information"
+  _kubectl get all --namespace cdi-testing-old-version-artifacts
+  _kubectl describe pvc olddv-v1alpha1 --namespace cdi-testing-old-version-artifacts
+  _kubectl describe pvc olddv-v1beta1 --namespace cdi-testing-old-version-artifacts
+  _kubectl get dv --namespace cdi-testing-old-version-artifacts -o yaml
```
alromeros, line 131: Just a nitpick, wouldn't `_kubectl get dv -o yaml` and `_kubectl describe dv` output too much redundant information? Shouldn't `describe` be enough?

alromeros (commented):
Looks very good! Just want to know your opinion on one thing.

cluster-sync/sync.sh
awels: So a describe will output the relevant events, and part of the yaml, the -o yaml outputs the entire yaml. There is definitely some overlapping information, but when debugging its usually better to have too much information instead of not enough. I can take the -o yaml out, its not a huge deal.

alromeros: No problem, I guess in case of error it's just better to have all information possible.

PR #2439, opened 2022-09-29
Add cron-job sa to scc

awels:

Add cronjob-sa to SCC so we can run inside Open Shift without issues.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -214,6 +214,9 @@ const (
 	// ControllerServiceAccountName is the name of the CDI controller service account
 	ControllerServiceAccountName = "cdi-sa"
 
+	// CronJobServiceAccountName is the name of the CDI cron job service account
+	CronJobServiceAccountName = "cdi-cronjob"
+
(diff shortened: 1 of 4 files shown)


mhenriks:
Can someone confirm that this actually works on OpenShift before I approve?  Should be able to simply edit the SCC with current virsion?

akalenyu:
> Can someone confirm that this actually works on OpenShift before I approve? Should be able to simply edit the SCC with current virsion?

Yeah seems to do the trick. I added it and everything kicked off resulting in the DataImportCron's reaching
```
      status: "True"
      type: UpToDate
```

mhenriks:
thanks @akalenyu!

akalenyu:
> We have to handle upgrade case. Currently users are only added to the SCC when created.

Do you mean we have to make the change in https://github.com/kubevirt/containerized-data-importer/blob/3b91645515e8a17c2db2f6051b394a0890133c15/pkg/operator/controller/scc.go#L41 instead?

awels:
Added code to ensure we add the cron user on upgrade as well.

PR #2441, opened 2022-09-29
Adapt DataImportCron and DataSource to DV GC

arnongilboa:

In order to adapt DataImportCron and DataSource controllers to DataVolume garbage collection, we did a few minor changes in behavior.

Before:
1. DIC controller created all DVs with deleteAfterCompletion=false annotation so they are not garbage collected.
2. DIC controller _internally_ garbage collected LRU import DVs (and their PVCs) per DIC based on its spec.importsToKeep (3 by default)
3. DIC & DataSource controllers only regarded a DV with status.phase=Succeeded as successfully completed import.

After:
1. DIC controller creates DVs which are GCed after completion by default.
2. DIC controller internally garbage collects LRU import PVCs (and their DVs if they exist) based on their new lastUseTime annotation.
3. DIC & DataSource controllers regard a PVC with no DV as Succeeded.
4. If the expected digest-named PVC exists, DV is not created.
5. DataSource controller watches indexed PVCs for create/delete. 

Fixes #

**Release note**:
```release-note
Add DataVolume garbage collection support for DataImportCron and DataSource
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -297,46 +297,18 @@ func getCronRegistrySource(cron *cdiv1.DataImportCron) (*cdiv1.DataVolumeSourceR
 }
 
 func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
-	log := r.log.WithName("update")
 	res := reconcile.Result{}
 
-	dataImportCronCopy := dataImportCron.DeepCopy()
-	importSucceeded := false
-	imports := dataImportCron.Status.CurrentImports
-	dataVolume := &cdiv1.DataVolume{}
-	dvFound := false
-	if len(imports) > 0 {
-		dvName := imports[0].DataVolumeName
-		// Get the currently imported DataVolume
-		if err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: dvName}, dataVolume); err != nil {
-			if !k8serrors.IsNotFound(err) {
-				return res, err
-			}
-			pvc := &corev1.PersistentVolumeClaim{}
-			if err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: dvName}, pvc); err != nil {
-				if !k8serrors.IsNotFound(err) {
-					return res, err
-				}
-				log.Info("PVC not found, removing from current imports", "name", dvName)
-				dataImportCron.Status.CurrentImports = imports[1:]
-			} else {
-				// PVC found, and DV was GCed
-				importSucceeded = true
-				pvcCopy := pvc.DeepCopy()
-				r.setDataImportCronResourceLabels(dataImportCron, pvc)
-				if !reflect.DeepEqual(pvc, pvcCopy) {
-					if err := r.client.Update(ctx, pvc); err != nil {
-						return res, err
-					}
-				}
-			}
-		} else {
-			dvFound = true
-		}
+	dv, pvc, err := r.getImportState(ctx, dataImportCron)
+	if err != nil {
+		return res, err
 	}
 
-	if dvFound {
-		switch dataVolume.Status.Phase {
+	dataImportCronCopy := dataImportCron.DeepCopy()
+	imports := dataImportCron.Status.CurrentImports
+	importSucceeded := false
+	if dv != nil {
+		switch dv.Status.Phase {
 		case cdiv1.Succeeded:
 			importSucceeded = true
 		case cdiv1.ImportScheduled:
@@ -344,10 +316,22 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		case cdiv1.ImportInProgress:
 			updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionTrue, "Import is progressing", inProgress)
 		default:
-			dvPhase := string(dataVolume.Status.Phase)
+			dvPhase := string(dv.Status.Phase)
 			updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionFalse, fmt.Sprintf("Import DataVolume phase %s", dvPhase), dvPhase)
 		}
+	} else if pvc != nil {
+		importSucceeded = true
+		pvcCopy := pvc.DeepCopy()
+		r.setDataImportCronResourceLabels(dataImportCron, pvc)
+		if !reflect.DeepEqual(pvc, pvcCopy) {
+			if err := r.client.Update(ctx, pvc); err != nil {
+				return res, err
+			}
+		}
 	} else {
+		if len(imports) > 0 {
+			dataImportCron.Status.CurrentImports = imports[1:]
+		}
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionFalse, "No current import", noImport)
 	}
 
@@ -366,7 +350,6 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	}
 
 	// We use the poller returned reconcile.Result for RequeueAfter if needed
-	var err error
 	if isImageStreamSource(dataImportCron) {
 		res, err = r.pollImageStreamDigest(ctx, dataImportCron)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


tests/datasource_test.go
```diff
@@ -65,6 +64,14 @@ var _ = Describe("DataSource", func() {
 		f.ForceBindIfWaitForFirstConsumer(pvc)
 	}
 
+	// Delete PVC if DV was GCed, otherwise delete both
+	deletePvc := func(pvcName string) {
```
awels, line 68: You are potentially deleting DVs here as well, the name should probably reflect that fact.

tests/operator_test.go
```diff
@@ -466,9 +466,8 @@ var _ = Describe("ALL Operator tests", func() {
 				Expect(pvc).ToNot(BeNil())
 
 				By("Verify dv succeeded")
-				dv, err := f.CdiClient.CdiV1beta1().DataVolumes(pvc.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
+				err = utils.WaitForDataVolumePhase(f, pvc.Namespace, cdiv1.Succeeded, pvc.Name)
```
awels, line 469: Might make more sense to use `WaitForDataVolumeGC` instead. Not sure. This function will call that eventually due to `Succeeded`, but my first question that popped up was, isn't the DV GCed at this point?

pkg/controller/dataimportcron-controller.go
```diff
@@ -506,70 +528,59 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+	if err := r.client.List(ctx, pvcList, &client.ListOptions{Namespace: cron.Namespace, LabelSelector: selector}); err != nil {
+		return err
 	}
-	sort.Slice(dvList.Items, func(i, j int) bool {
-		getDvTimestamp := func(dv cdiv1.DataVolume) time.Time {
-			if cond := findConditionByType(cdiv1.DataVolumeReady, dv.Status.Conditions); cond != nil {
-				return cond.LastHeartbeatTime.Time
+	if len(pvcList.Items) > maxImports {
+		sort.Slice(pvcList.Items, func(i, j int) bool {
+			return pvcList.Items[i].Annotations[AnnLastUseTime] > pvcList.Items[j].Annotations[AnnLastUseTime]
```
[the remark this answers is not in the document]
arnongilboa: Creation time won't always work here. e.g the oldest PVC can be the most recently used (e.g. revert to a previous version), so we won't like to GC it.

pkg/controller/dataimportcron-controller.go
```diff
@@ -304,28 +306,39 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
+		dvName := imports[0].DataVolumeName
 		// Get the currently imported DataVolume
-		if err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: imports[0].DataVolumeName}, dataVolume); err != nil {
+		if err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: dvName}, dataVolume); err != nil {
 			if !k8serrors.IsNotFound(err) {
 				return res, err
 			}
-			log.Info("DataVolume not found, removing from current imports", "name", imports[0].DataVolumeName)
-			dataImportCron.Status.CurrentImports = imports[1:]
+			pvc := &corev1.PersistentVolumeClaim{}
```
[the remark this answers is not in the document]
arnongilboa: Sure, will do.

pkg/controller/dataimportcron-controller.go
```diff
@@ -338,6 +351,16 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionFalse, "No current import", noImport)
 	}
 
+	if importSucceeded {
+		if err := updateDataImportCronOnSuccess(dataImportCron); err != nil {
```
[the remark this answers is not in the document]
arnongilboa: Note importSucceeded is set `true` in two places, so else won't cover both.

tests/datasource_test.go
arnongilboa: deletePvcAndDv? sounds bad

tests/operator_test.go
arnongilboa: WaitForDataVolumePhase is prefarable as it is agostic to wheter GC is enabled or not.

mhenriks:
A little more detail on behavior before vs after this pr would be helpful.  Thanks

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
awels: Okay

awels (commented):
Nothing major, just minor questions. I also don't see the functional tests updated to test the new functionality.

pkg/controller/dataimportcron-controller.go
```diff
@@ -506,70 +535,59 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
-		return getDvTimestamp(dvList.Items[i]).After(getDvTimestamp(dvList.Items[j]))
-	})
-	for _, dv := range dvList.Items[maxDvs:] {
-		logDv := log.WithValues("name", dv.Name).WithValues("uid", dv.UID)
-		if err := r.client.Delete(ctx, &dv); err != nil {
-			if k8serrors.IsNotFound(err) {
-				logDv.Info("DataVolume not found for deletion")
-			} else {
-				logDv.Error(err, "Unable to delete DataVolume")
+			if err := r.client.Delete(ctx, &pvc); err != nil && !k8serrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
arnongilboa: You are right, and that's why we `continue` if DV deletion is successful, 4 lines above.

tests/operator_test.go
arnongilboa: `WaitForDataVolumePhase/WithTimeout(...cdiv1.Succeeded...)` is refered 62 times all around the func tests. I don't think the suggested rename worth it this time.

arnongilboa:
> Not seeing "DIC controller creates DVs which are GCed after completion by default" verified in functest anywhere, but maybe I'm missing it

@mhenriks referring the behavior changes:
```
1. DIC controller creates DVs which are GCed after completion by default.
2. DIC controller internally garbage collects LRU import PVCs (and their DVs if they exist) based on their new lastUseTime annotation.
3. DIC & DataSource controllers regard a PVC with no DV as Succeeded.
4. If the expected digest-named PVC exists, DV is not created.
5. DataSource controller watches indexed PVCs for create/delete.
```
- 1 & 3 are verified in functests by `WaitForDataVolumePhase(...cdiv1.Succeeded...)` which waits for the DV to be GCed (in case GC is enabled, which is the default). See [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/dataimportcron_test.go#L158) and [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/datasource_test.go#L143).
- 2 is verified in functest [here](https://github.com/kubevirt/containerized-data-importer/pull/2441/files#diff-7ae289da57be3122f64c1fd539d7539e9f3b080db79f0a38498caa00615d754bL217-L222).
- 5 is verified in functest [here](https://github.com/kubevirt/containerized-data-importer/pull/2441/files#diff-3f4304f532487d735600b071a64e54db2acd2b13e46eefa78fee54b760f991c5R90), where PVC is deleted after DV was already GCed.
- 2 & 4 are verified by the new [utest](https://github.com/kubevirt/containerized-data-importer/pull/2441/files#diff-3208c786da38345c3e55b1d5b64a531daae6e4c39f69a621665bddb78a798ed6R314).

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
awels: Oh missed that.

tests/operator_test.go
awels: So we have 62 locations that are confusing when GC is enabled? Could we at the very least put a comment on the function indicating that it also accounts for the DV being garbage collected if the passed in phase is succeeded? That way when people look at the function it kind of makes sense.

arnongilboa: Comment added

mhenriks:
@arnongilboa, I hear you, but this is a little sneaky

> 1 & 3 are verified in functests by WaitForDataVolumePhase(...cdiv1.Succeeded...) which waits for the DV to be GCed (in case GC is enabled, which is the default). See [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/dataimportcron_test.go#L158) and [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/datasource_test.go#L143).

buried in  `WaitForDataVolumePhase` is implicit handling of `AnnDeleteAfterCompletion` which used to be set to false, right?

I still think it will be nice to explicitly check that the dv is gone but will not will not require you to make changes.  Let's see what others think

arnongilboa:
> buried in `WaitForDataVolumePhase` is implicit handling of `AnnDeleteAfterCompletion` which used to be set to false, right?
> 
> I still think it will be nice to explicitly check that the dv is gone but will not will not require you to make changes. Let's see what others think

Not sure I agree. Using WaitForDataVolumePhase let us run the same tests also if we disable GC, and it covers the complete chek needed in both cases.

PR #2455, opened 2022-10-30
Add DataImportCron CronJobs Proxy support

arnongilboa:

Use proxy certificates when creating DataImportCron CronJobs, so the pods won't fail on certificate issues when proxy is configured. `cdi-source-update-poller` was also fixed to use the proxy certificates.

Fixes #

**Release note**:
```release-note
Add DataImportCron CronJobs Proxy support
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -771,14 +772,25 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
 		TerminationMessagePolicy: corev1.TerminationMessageReadFile,
 	}
 
+	volumes := []corev1.Volume{}
 	hasCertConfigMap := regSource.CertConfigMap != nil && *regSource.CertConfigMap != ""
 	if hasCertConfigMap {
 		vm := corev1.VolumeMount{
 			Name:      CertVolName,
 			MountPath: common.ImporterCertDir,
 		}
-		container.VolumeMounts = []corev1.VolumeMount{vm}
+		container.VolumeMounts = append(container.VolumeMounts, vm)
 		container.Command = append(container.Command, "-certdir", common.ImporterCertDir)
+		volumes = append(volumes, createConfigMapVolume(CertVolName, *regSource.CertConfigMap))
+	}
+
+	if volName, _ := GetImportProxyConfig(cdiConfig, common.ImportProxyConfigMapName); volName != "" {
+		vm := corev1.VolumeMount{
+			Name:      ProxyCertVolName,
+			MountPath: common.ImporterProxyCertDir,
+		}
+		container.VolumeMounts = append(container.VolumeMounts, vm)
+		volumes = append(volumes, createConfigMapVolume(ProxyCertVolName, volName))
 	}
 
 	if regSource.SecretRef != nil && *regSource.SecretRef != "" {
@@ -809,19 +821,8 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
 	}
 
 	if insecureTLS {
-		container.Env = append(container.Env,
-			corev1.EnvVar{
-				Name:  common.InsecureTLSVar,
-				Value: "true",
-			},
-		)
+		container.Env = append(container.Env, corev1.EnvVar{Name: common.InsecureTLSVar, Value: "true"})
 	}
-
-	successfulJobsHistoryLimit := int32(0)
-	failedJobsHistoryLimit := int32(0)
-	ttlSecondsAfterFinished := int32(0)
-	backoffLimit := int32(2)
-	gracePeriodSeconds := int64(0)
 
 	cronJobName := GetCronJobName(cron)
 	cronJob := &batchv1.CronJob{
@@ -832,37 +833,24 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
 		Spec: batchv1.CronJobSpec{
 			Schedule:                   cron.Spec.Schedule,
 			ConcurrencyPolicy:          batchv1.ForbidConcurrent,
-			SuccessfulJobsHistoryLimit: &successfulJobsHistoryLimit,
-			FailedJobsHistoryLimit:     &failedJobsHistoryLimit,
+			SuccessfulJobsHistoryLimit: pointer.Int32(1),
+			FailedJobsHistoryLimit:     pointer.Int32(0),
 			JobTemplate: batchv1.JobTemplateSpec{
 				Spec: batchv1.JobSpec{
 					Template: corev1.PodTemplateSpec{
 						Spec: corev1.PodSpec{
 							RestartPolicy:                 corev1.RestartPolicyNever,
-							TerminationGracePeriodSeconds: &gracePeriodSeconds,
+							TerminationGracePeriodSeconds: pointer.Int64(0),
 							Containers:                    []corev1.Container{container},
 							ServiceAccountName:            common.CronJobServiceAccountName,
+							Volumes:                       volumes,
 						},
 					},
-					TTLSecondsAfterFinished: &ttlSecondsAfterFinished,
-					BackoffLimit:            &backoffLimit,
+					TTLSecondsAfterFinished: pointer.Int32(3),
+					BackoffLimit:            pointer.Int32(2),
 				},
 			},
 		},
-	}
-
-	if hasCertConfigMap {
-		vol := corev1.Volume{
-			Name: CertVolName,
-			VolumeSource: corev1.VolumeSource{
-				ConfigMap: &corev1.ConfigMapVolumeSource{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: *regSource.CertConfigMap,
-					},
-				},
-			},
-		}
-		cronJob.Spec.JobTemplate.Spec.Template.Spec.Volumes = []corev1.Volume{vol}
 	}
 
 	if err := r.setJobCommon(cron, cronJob); err != nil {
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -820,9 +820,25 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
 		)
 	}
 
-	if insecureTLS {
-		container.Env = append(container.Env, corev1.EnvVar{Name: common.InsecureTLSVar, Value: "true"})
+	addEnvVar := func(varName, value string) {
+		container.Env = append(container.Env, corev1.EnvVar{Name: varName, Value: value})
 	}
+
+	if insecureTLS {
+		addEnvVar(common.InsecureTLSVar, "true")
+	}
+
+	addEnvVarFromImportProxyConfig := func(varName string) {
+		if value, err := GetImportProxyConfig(cdiConfig, varName); err == nil {
+			addEnvVar(varName, value)
+		} else {
+			r.log.Info("Missing", varName, err.Error())
+		}
+	}
+
+	addEnvVarFromImportProxyConfig(common.ImportProxyHTTP)
+	addEnvVarFromImportProxyConfig(common.ImportProxyHTTPS)
+	addEnvVarFromImportProxyConfig(common.ImportProxyNoProxy)
 
 	cronJobName := GetCronJobName(cron)
 	cronJob := &batchv1.CronJob{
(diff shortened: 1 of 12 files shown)


tools/cdi-source-update-poller/main.go
```diff
@@ -58,6 +64,8 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed BuildConfigFromFlags, kubeURL %s configPath %s: %v", kubeURL, configPath, err)
 	}
+	cfg.TLSClientConfig = rest.TLSClientConfig{Insecure: true}
```
akalenyu, line 67: hmm which change in your PR now requires this?
this is just about the cdi client that we use to get stuff (not the polling), why do we have to go insecure?

pkg/controller/import-controller.go
```diff
@@ -1259,7 +1247,7 @@ func makeImporterContainerSpec(image, verbose, pullPolicy string) *corev1.Contai
 	}
 }
 
-func createProxyConfigMapVolume(certVolName, objRef string) corev1.Volume {
+func createConfigMapVolume(certVolName, objRef string) corev1.Volume {
```
akalenyu, line 1250: :pray:

pkg/controller/dataimportcron-controller.go
```diff
@@ -771,14 +772,27 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
...
 	hasCertConfigMap := regSource.CertConfigMap != nil && *regSource.CertConfigMap != ""
 	if hasCertConfigMap {
 		vm := corev1.VolumeMount{
 			Name:      CertVolName,
 			MountPath: common.ImporterCertDir,
 		}
-		container.VolumeMounts = []corev1.VolumeMount{vm}
+		container.VolumeMounts = append(container.VolumeMounts, vm)
 		container.Command = append(container.Command, "-certdir", common.ImporterCertDir)
+		volumes = append(volumes, createConfigMapVolume(CertVolName, *regSource.CertConfigMap))
```
akalenyu, line 784: So prior to this PR we did not support custom CA for polling the dataimportcron registry?
(This is unrelated to the proxy stuff, right?)
I just want to make sure I get this right

pkg/controller/dataimportcron-controller.go
```diff
@@ -808,20 +822,25 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
...
-	gracePeriodSeconds := int64(0)
+	addEnvVarFromImportProxyConfig := func(varName string) {
+		if value, err := GetImportProxyConfig(cdiConfig, varName); err == nil {
+			addEnvVar(varName, value)
+		} else {
+			r.log.Info("Missing", varName, err.Error())
+		}
+	}
+
+	addEnvVarFromImportProxyConfig(common.ImportProxyHTTP)
```
akalenyu, line 841: So adding the env vars is enough to ensure the proxy is respected?
@awels you identified some golang specifics with regards to HTTP env vars before, will this be fine?

akalenyu (commented):
I think we will need to backport this, and I am pretty sure this is backport friendly, so that's awesome

pkg/controller/dataimportcron-controller.go
```diff
@@ -771,14 +772,27 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
...
+	}
+
+	if volName, err := GetImportProxyConfig(cdiConfig, common.ImportProxyConfigMapName); err == nil {
+		vm := corev1.VolumeMount{
+			Name:      ProxyCertVolName,
+			MountPath: common.ImporterProxyCertDir,
+		}
+		container.VolumeMounts = append(container.VolumeMounts, vm)
+		volumes = append(volumes, createConfigMapVolume(ProxyCertVolName, volName))
+	} else {
```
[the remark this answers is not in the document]
arnongilboa: why? nobody said you must have any of the Proxy params set. Similar logic to to the import controller.

pkg/controller/dataimportcron-controller.go
```diff
@@ -808,20 +822,25 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
...
 
-	successfulJobsHistoryLimit := int32(0)
-	failedJobsHistoryLimit := int32(0)
-	ttlSecondsAfterFinished := int32(0)
-	backoffLimit := int32(2)
-	gracePeriodSeconds := int64(0)
+	addEnvVarFromImportProxyConfig := func(varName string) {
+		if value, err := GetImportProxyConfig(cdiConfig, varName); err == nil {
+			addEnvVar(varName, value)
+		} else {
```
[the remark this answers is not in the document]
arnongilboa: same answer as above.

tools/cdi-source-update-poller/main.go
```diff
@@ -48,7 +49,12 @@ func init() {
 }
 
 func main() {
-	digest, err := importer.GetImageDigest(url, accessKey, secretKey, certDir, insecureTLS)
+	allCertDir, err := importer.CreateCertificateDir(certDir)
+	if err != nil {
+		os.Exit(1)
```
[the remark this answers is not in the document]
arnongilboa: Sure, adding.

tools/cdi-source-update-poller/main.go
arnongilboa: The CronJob containerized poller app runs in CDI namespace, and it tries to Get/Update() DataImportCron which is in another namespace, so when proxy is configured, Get/Update() fail on certificate issues unless we configure TLS appropriately. Maybe `Insecure: true` is not the best idea here, but I have no idea how to use the certs dir for setting TLSClientConfig.

pkg/controller/dataimportcron-controller.go
arnongilboa: We used certs but not proxy certs. See it in the refactored code here.

pkg/controller/dataimportcron-controller.go
arnongilboa: No, the CronJob poller app was adapted to use the proxy certs. Maybe these env vars are just leftovers of my debug and not needed at all. Let me check...

pkg/controller/dataimportcron-controller.go
akalenyu: I mean that before, we passed the `regSource.CertConfigMap`, but we did not capitalize on it in the poller
(just polled regularly)

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
akalenyu: :+1:

pkg/controller/dataimportcron-controller.go
arnongilboa: We already used the regular certs in the poller before (see GetImageDigest implementation):
https://github.com/kubevirt/containerized-data-importer/blob/main/tools/cdi-source-update-poller/main.go#L51

tools/cdi-source-update-poller/main.go
akalenyu: AFAIK, under the hood, it should respect proxy env vars so not sure what goes wrong
https://github.com/kubevirt/containerized-data-importer/blob/206c77afdaad91e5be99460a76a1a5d3e5131841/vendor/k8s.io/client-go/rest/config.go#L137-L142

arnongilboa: ok, but what would you append?

akalenyu: hmm if it was erroring on x509 I think
https://github.com/kubernetes/apiserver/blob/25ccbfa75c69642e44ac9be141e3bacc7edb3879/pkg/util/webhook/client.go#L136-L140 could work, or something nicer like placing the cert in the golang "respected" path https://go.dev/src/crypto/x509/root_linux.go

pkg/controller/dataimportcron-controller.go
akalenyu: I am not sure how we will be proxying stuff without the env vars, isn't that the only magic that makes proxying take place? how else would the go/docker client know to proxy it?

arnongilboa: We pass the proxy certs to the pod and the containerized app (poller) use it, 
so it can poll registry url via the configured proxy. We don't initiate any proxying here.

arnongilboa: Without the proxy certs the poller app failed on connecting to the proxy. This was solved.

arnongilboa: I guess it proxied after the first import or something @awels ?

pkg/controller/dataimportcron-controller.go
awels: So I am missing basically the equivalent of https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/import-controller.go#L601-L617 in the job configuration. The environment variables is what enables the proxy to work.

tests/import_proxy_test.go
```diff
@@ -305,20 +314,110 @@ var _ = Describe("Import Proxy tests", func() {
...
+				Expect(dvName).ToNot(BeEmpty())
+				return dvName
+			}, timeout, pollingInterval).ShouldNot(BeEmpty())
+
+			By("Wait for datavolume succeeded")
+			err = utils.WaitForDataVolumePhase(f, ns, cdiv1.Succeeded, dvName)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Checking the importer pod information in the proxy log to verify if the requests were proxied")
+			verifyImporterPodInfoInProxyLogs(f, url, registryUserAgent, now, BeTrue)
```
awels, line 406: You can't use this function the way it is. This functions looks for the importer pod (which we know works already). We are interested in the poller working properly. So you have to either modify this function to pass in the prefix for the pod to look for, or use some other method of finding the poller pod.

pkg/controller/dataimportcron-controller.go
arnongilboa: sure, re-adding, and adding test coverage as well.

tests/import_proxy_test.go
```diff
@@ -305,20 +314,110 @@ var _ = Describe("Import Proxy tests", func() {
...
+				dataSourceName      = "datasource-test"
+				scheduleEveryMinute = "* * * * *"
+			)
+
+			now := time.Now()
+			ns := f.Namespace.Name
+			updateProxy(f, ns, "", createProxyURL(isHTTPS, hasAuth, f.CdiInstallNs), "", ocpClient)
+
+			// Copy user-ca-bundle also to cdi ns for the cronjobs
+			_, err := utils.CopyConfigMap(f.K8sClient, f.CdiInstallNs, cdiProxyCaConfigMapName, f.CdiInstallNs, trustedCaProxyConfigName, "")
```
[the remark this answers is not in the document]
arnongilboa: Yep, we create all cronjobs in the CDI namespace. That was a design decision.

pkg/controller/dataimportcron-controller.go
awels: No the proxying is enabled when the environment variables are set `http_proxy, https_proxy, no_proxy` some applications want to use upper case version of that. The golang client uses the lower case versions though.

The extra certs are basically needed to connect to the proxy if the proxy has self signed CAs.

tests/import_proxy_test.go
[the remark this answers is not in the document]
awels: Okay sounds good just wanted to confirm.

tools/cdi-source-update-poller/main.go
awels: Actually thinking about it a bit more, you don't need to proxy the k8s calls at all. So you don't have to mess with the `cfg` settings at all.

akalenyu: I think they are being proxied because of the env vars, usually, we don't set the env vars on containers that do k8s API interactions

arnongilboa: Fixing that

tests/import_proxy_test.go
arnongilboa: Fixing

awels (commented):
looks good, one question and I would love to see some unit tests for the environment variable adding in the `dataimportcron-controller`

pkg/controller/dataimportcron-controller.go
```diff
@@ -832,39 +849,24 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
 		Spec: batchv1.CronJobSpec{
 			Schedule:                   cron.Spec.Schedule,
 			ConcurrencyPolicy:          batchv1.ForbidConcurrent,
-			SuccessfulJobsHistoryLimit: &successfulJobsHistoryLimit,
-			FailedJobsHistoryLimit:     &failedJobsHistoryLimit,
+			SuccessfulJobsHistoryLimit: pointer.Int32(1),
```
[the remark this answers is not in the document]
arnongilboa: Yep, they are intentional, keeping the last (1 by default) failed job (& pod) history helps debugging, and keeping the last succeeded job is useful for the test, and does not introduce too much noise as well.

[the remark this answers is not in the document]
awels: Okay, sounds good just some unit tests and it is good to go.

tools/cdi-source-update-poller/main.go
akalenyu: Nice and clean
Just making sure here: the reason we don't care about proxying the k8s API calls is that it's somehow taken care of in another internal k8s layer?

arnongilboa: We have no good reason to proxy the 2 k8s get/update calls here. Do we?

awels: No the assumption is that the proxy is on the edge of the cluster (aka stuff that tries to go outside of the cluster). I see no use case where it makes sense to proxy traffic inside the cluster.

awels (commented):
Three things for the unit test:
1. You modified the successful history and failed history counts, can you ensure they are what we expect.
2. You verified the env variables are set when the proxy has a value, can you also verify they are NOT set when the proxy does not have a value?
3. Also please verify that the extra volume is NOT set when the caProxy value is empty. It is possible to have a proxy but not need a custom CA since the CA of the proxy is signed by a known CA

awels (approved):
/approve

PR #2474, opened 2022-11-21
Fix OpenShift cluster-wide proxy config update

arnongilboa:

Fixes #

**Release note**:
```release-note
Fix OpenShift cluster-wide proxy config update
```


akalenyu:
/lgtm

passes on openshift
```
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 453 Skipped
```

awels (commented):
I see the fix for the cron jobs, just don't quite understand the renaming of the config map in the cdi namespace. We have to remember this has to work for both Open Shift and regular k8s environments. This is why we have the common named config map in the cdi namespace. for k8s environments people can create that config map in that namespace and have it work.

pkg/controller/config-controller.go
```diff
@@ -375,54 +376,52 @@ func (r *CDIConfigReconciler) reconcileImportProxy(config *cdiv1.CDIConfig) erro
...
-					return err
-				}
-			}
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: ClusterWideProxyConfigMapNameSpace}, clusterWideProxyConfigMap); err != nil {
+		return err
+	}
+	// Copy the cluster-wide proxy CA certificates to the importer pod proxy CA certificates configMap
+	certBytes, ok := clusterWideProxyConfigMap.Data[ClusterWideProxyConfigMapKey]
+	if !ok {
+		return fmt.Errorf("No cluster-wide proxy CA certifiacte")
```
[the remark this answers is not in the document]
arnongilboa: Fixed the message. I think it's an error if a cluster wide proxy config is there, but missing the certificate data.

pkg/importer/registry-datasource.go
```diff
@@ -191,10 +191,20 @@ func collectCerts(certDir, targetDir, targetPrefix string) error {
...
+		hasSuffix := func(suffix string) bool {
+			return strings.HasSuffix(fileName, suffix)
+		}
+		if !hasSuffix(".crt") && !hasSuffix(".pem") {
 			continue
 		}
-		if err := util.LinkFile(filepath.Join(certDir, obj.Name()), filepath.Join(targetDir, targetPrefix+obj.Name())); err != nil {
+		targetName := ""
+		if hasSuffix(".pem") {
+			targetName = targetPrefix + strings.TrimSuffix(fileName, ".pem") + ".crt"
```
akalenyu, line 203: @awels This has to be done because of containers/image library only allowing .crt's, but is pretty hacky
I'd say if we are documenting this limitation, we could just change https://github.com/kubevirt/containerized-data-importer/blob/97fb3ca9222027032f284aeaa98cedc745921242/pkg/common/common.go#L80
And later decide if we want to be the compatibility layer here

pkg/controller/config-controller.go
```diff
@@ -375,54 +376,52 @@ func (r *CDIConfigReconciler) reconcileImportProxy(config *cdiv1.CDIConfig) erro
...
-			}
-			if configMap != nil {
-				configMap.Data[common.ImportProxyConfigMapKey] = certBytes
-				util.SetRecommendedLabels(configMap, r.installerLabels, "cdi-controller")
-				if err := client.Update(context.TODO(), configMap); err != nil {
-					return err
-				}
-			}
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: ClusterWideProxyConfigMapNameSpace}, clusterWideProxyConfigMap); err != nil {
+		return err
```
awels, line 394: Can we put an event on the cdi cr indicating that the cluster wide proxy doesn't exist in case this error is a NotFound error?

awels (commented):
So if I understand this correctly this is not addressing the issue where we are missing the config map in the workload namespace? Just the synchronization between the cluster wide proxy config map and the config map in CDI?

Also what happens when the cluster wide proxy is removed? Are we removing the copied config map in the CDI namespace? And what happens when the cluster wide proxy config map is named something else? I think this is why I had a known name in the CDI namespace, so that we would not end up with a bunch of config maps we don't know where they came from.

arnongilboa:
> So if I understand this correctly this is not addressing the issue where we are missing the config map in the workload namespace? Just the synchronization between the cluster wide proxy config map and the config map in CDI?

right

> Also what happens when the cluster wide proxy is removed? Are we removing the copied config map in the CDI namespace? And what happens when the cluster wide proxy config map is named something else? I think this is why I had a known name in the CDI namespace, so that we would not end up with a bunch of config maps we don't know where they came from.

we should remove teh copied CM and update when cluster wide proxy is removed. I'll fix that.

pkg/controller/config-controller.go
arnongilboa: Sure, will add.

pkg/controller/import-controller.go
```diff
@@ -358,6 +365,43 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
...
+	cdiConfigMap := &corev1.ConfigMap{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: r.cdiNamespace}, cdiConfigMap); err != nil {
+		return err
+	}
+	importConfigMap := &corev1.ConfigMap{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: pvc.Namespace}, importConfigMap); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+		importConfigMap.Name = cmName
```
awels, line 386: I would make the name unique per importer pod, otherwise you will have name clashes or attempts to delete the cm while multiple imports are running. Also if you make the name unique (use the importer, dv, or pvc name as part of the cm name?) then you don't have to check if it exists, you can just create it (and ignore already existing failures), and you don't need list/watch/update RBAC on config maps.

pkg/operator/resources/cluster/controller.go
```diff
@@ -110,6 +110,10 @@ func getControllerClusterPolicyRules() []rbacv1.PolicyRule {
 			},
 			Verbs: []string{
 				"get",
+				"list",
+				"watch",
```
awels, line 114: See comment about RBAC above, would love to just add create instead of the rest as well.

awels (commented):
looks good a few comments.

pkg/controller/config-controller.go
```diff
@@ -375,54 +378,67 @@ func (r *CDIConfigReconciler) reconcileImportProxy(config *cdiv1.CDIConfig) erro
...
+	configMap.Data[common.ImportProxyConfigMapKey] = certBytes
+	util.SetRecommendedLabels(configMap, r.installerLabels, "cdi-controller")
+	if err := client.Update(context.TODO(), configMap); err != nil {
+		return err
 	}
 	return nil
 }
 
-func (r *CDIConfigReconciler) createProxyConfigMap(certBytes string) *v1.ConfigMap {
+func (r *CDIConfigReconciler) createProxyConfigMap(cmName, certBytes string) *v1.ConfigMap {
```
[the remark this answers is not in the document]
arnongilboa: done

pkg/controller/config-controller_test.go
```diff
@@ -333,7 +333,9 @@ var _ = Describe("Controller ImportProxy reconcile loop", func() {
 	var trustedCAProxy = "user-ca-bundle"
 
 	DescribeTable("Should set ImportProxy correctly if ClusterWideProxy with correct URLs exists", func(proxyHTTPURL string, proxyHTTPSURL string, noProxyDomains string, trustedCAName string, expect string, endpType string) {
-		reconciler, cdiConfig := createConfigReconciler(createClusterWideProxy(proxyHTTPURL, proxyHTTPSURL, noProxyDomains, trustedCAProxy))
+		trustedCAProxyConfigMap := createConfigMap(trustedCAProxy, ClusterWideProxyConfigMapNameSpace)
```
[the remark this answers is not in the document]
arnongilboa: added

pkg/controller/import-controller.go
arnongilboa: Done, and removed the "update" RBAC. However, for getting the cdi namespace ConfigMap, it requires not only “get” but also both “list “ and “watch”.
```
I1207 23:46:17.113513       1 reflector.go:255] Listing and watching *v1.ConfigMap from pkg/cache/internal/informers_map.go:250
W1207 23:46:17.115447       1 reflector.go:324] pkg/cache/internal/informers_map.go:250: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cdi:cdi-sa" cannot list resource "configmaps" in API group "" at the cluster scope
E1207 23:46:17.115491       1 reflector.go:138] pkg/cache/internal/informers_map.go:250: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cdi:cdi-sa" cannot list resource "configmaps" in API group "" at the cluster scope
```
Note that since the config-controller watches CDIConfig, it gets "list" and "watch" for free.

pkg/operator/resources/cluster/controller.go
arnongilboa: see above

awels: So I checked https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/operator/resources/namespaced/controller.go#L66-L76 and we already have `*` on config maps in our namespace. What exactly is not working for you that you need this extra permission? @mhenriks am I being too paranoid with list/watch on config maps?

awels (commented):
Looks good just want to discuss the RBAC a little before approving.

pkg/operator/resources/cluster/controller.go
arnongilboa: Switched to `uncachedClient.Get()` and removed the list/watch rbac.

pkg/controller/import-controller.go
```diff
@@ -358,6 +365,45 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
...
+		Data: cdiConfigMap.Data,
+	}
+	if err := r.client.Create(context.TODO(), importConfigMap); err != nil && !k8serrors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
+}
+
+// GetImportProxyConfigMapName returns the import proxy ConfigMap name
+func GetImportProxyConfigMapName(pvcName string) string {
```
akalenyu, line 403: I think there is a slim chance this will re-introduce some resource name-length problems
so we might have to use `naming.GetResourceName` and friends here

pkg/operator/resources/cluster/controller.go
akalenyu: https://github.com/kubevirt/containerized-data-importer/blob/80ac74bd9cf3a5026a8852752152fb33d1a31c69/cmd/cdi-controller/controller.go#L345
So read once=>cached=>needs list/watch

I also struggle to answer if list/watch are fine here, whether they are cheap or not, etc
But since it's always a GET I think we won't be hitting the API server too hard so uncached is also fine

akalenyu (commented):
Super elegant to have the importer namespace CM stuff being fixed here

arnongilboa:
> So read once=>cached=>needs list/watch
> I also struggle to answer if list/watch are fine here, whether they are cheap or not, etc
> But since it's always a GET I think we won't be hitting the API server too hard so uncached is also fine

@akalenyu  note that the list/watch/update were removed in the following commit

pkg/controller/import-controller.go
arnongilboa: fixed

akalenyu:
> > So read once=>cached=>needs list/watch
> > I also struggle to answer if list/watch are fine here, whether they are cheap or not, etc
> > But since it's always a GET I think we won't be hitting the API server too hard so uncached is also fine
> 
> @akalenyu note that the list/watch/update were removed in the following commit

Yeah I see, just wanted to join Alexander in asking whether they were harmful

awels:
Any time I see changes to cluster wide secret or config map RBAC, I am going to question it.

pkg/controller/import-controller.go
```diff
@@ -181,6 +183,7 @@ func NewImportController(mgr manager.Manager, log logr.Logger, importerImage, pu
 		verbose:         verbose,
 		pullPolicy:      pullPolicy,
 		recorder:        mgr.GetEventRecorderFor("import-controller"),
+		cdiNamespace:    util.GetNamespace(),
```
awels, line 186: I would rename `client` in this function to `c` or something like that (to avoid name collision with the package), and add a namespacedClient to the reconciler.
Then you can do something like this:
```go
reconciler.namespacedClient = client.NewNamespacedClient(c, r.cdiNamespace)
```
And in the function that gets the config map in the CDI namespace, you can now use the namespacedClient and it will work. You will have a cached namespaced client instead of an uncached version.

pkg/controller/import-controller.go
```diff
@@ -358,6 +365,45 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
...
+	cdiConfig := &cdiv1.CDIConfig{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, cdiConfig); err != nil {
+		return err
+	}
+	cmName, err := GetImportProxyConfig(cdiConfig, common.ImportProxyConfigMapName)
+	if err != nil || cmName == "" {
+		return nil
+	}
+	cdiConfigMap := &corev1.ConfigMap{}
+	if err := r.uncachedClient.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: r.cdiNamespace}, cdiConfigMap); err != nil {
```
awels, line 378: See namespaced client information at top of file.

pkg/controller/import-controller.go
arnongilboa: It looks like `namespacedClient.Get()` simply [calls](https://github.com/kubevirt/containerized-data-importer/blob/main/vendor/sigs.k8s.io/controller-runtime/pkg/client/namespaced_client.go#L141-L153) the original `client.Get()`,  so it doesn't do any better than the original code, and we get the same good old errors:
```
I1212 22:35:17.495054       1 reflector.go:255] Listing and watching *v1.ConfigMap from pkg/cache/internal/informers_map.go:250
W1212 22:35:17.501040       1 reflector.go:324] pkg/cache/internal/informers_map.go:250: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cdi:cdi-sa" cannot list resource "configmaps" in API group "" at the cluster scope
E1212 22:35:17.501184       1 reflector.go:138] pkg/cache/internal/informers_map.go:250: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cdi:cdi-sa" cannot list resource "configmaps" in API group "" at the cluster scope
```
I changed to k8s client and it seems to fix the issue.

awels (approved):
/approve

PR #2475, opened 2022-11-21
Add e2e tests for alerts

machadovilaca:
Signed-off-by: João Vilaça <jvilaca@redhat.com>

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
Hey, thanks! I see these specifically test the alerting mechanism but they might overlap a little with https://github.com/kubevirt/containerized-data-importer/blob/a19238ebbdadb8cc02ce91d3ed01c98935ff5475/tests/operator_test.go#L774

machadovilaca:
@akalenyu I saw it, what do you think about moving that `CDIOperatorDown` alert test to this new `monitoring_test.go` file?

akalenyu:
> @akalenyu I saw it, what do you think about moving that `CDIOperatorDown` alert test to this new `monitoring_test.go` file?

We can, sure
Reading the tests, they might overlap quite a bit (plus the 10min timeout),
Are we sure we want to almost dup the restart count test, for example?

machadovilaca:
We can also move the metrics tests here, and in the same tests, verify the metrics is correct and then that the alert is fired. But if the alerts is fired, we can probably assume the metric is also correct, when they are based in each other

tests/monitoring_test.go
```diff
@@ -0,0 +1,128 @@
...
+			cdi.UID = ""
+			cdi.Spec.Infra.NodeSelector = map[string]string{"wrong": "wrong"}
+			_, err = fw.CdiClient.CdiV1beta1().CDIs().Create(context.TODO(), cdi, metav1.CreateOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			By("checking that the CDINotReady alert is firing")
+			checkAlertExists(fw, "CDINotReady")
+
+			By("restoring old CDI config")
+			deleteCDI(fw)
```
akalenyu, line 82: We have a special lane for destructive tests where stuff like this should run (mostly in operator_test.go),
where we take special precautions in the AfterEach of each method to ensure CDI stabilized following delete.
Might make sense to reuse some functions from that suite so a failure in this test does not crash everything that follows

akalenyu (commented):
Awesome!

One concern though, and it might be just me over-evaluating this;
aren't we basically duplicating tests?
I am all up for the split to `monitoring_test.go`, but these types of tests all running twice
can be quite costly.
WDYT about adding the alert check over to existing tests?

tests/monitoring_test.go
```diff
@@ -0,0 +1,128 @@
...
+	Context("Alerts", func() {
+		It("should throw CDIDataVolumeUnusualRestartCount alert when a DataVolume object restarts more than three times", func() {
+			By("creating a DataVolume")
+			dataVolume := utils.NewDataVolumeWithHTTPImport(dataVolumeName, "500Mi", validURL)
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(fw.CdiClient, fw.Namespace.Name, dataVolume)
+			Expect(err).ToNot(HaveOccurred())
+			fw.ForceBindPvcIfDvIsWaitForFirstConsumer(dataVolume)
+
+			By("waiting for the DataVolume to be restarted more than three times")
+			Eventually(func() bool {
```
akalenyu, line 37: Eventually without `.Should(BeSomething())`? What am I missing

tests/monitoring_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	"kubevirt.io/containerized-data-importer/tests/framework"
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+var _ = Describe("Monitoring Tests", func() {
+	fw := framework.NewFramework("monitoring-test")
```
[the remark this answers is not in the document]
machadovilaca: done

tests/monitoring_test.go
machadovilaca: updated, moved those methods to an utils file

tests/monitoring_test.go
machadovilaca: it was missing

tests/monitoring_test.go
```diff
@@ -0,0 +1,135 @@
...
+			checkAlertExists(f, "CDIDataVolumeUnusualRestartCount")
+
+			By("deleting the DataVolume")
+			err = utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should throw CDIStorageProfilesIncomplete alert if there is a incomplete StorageProfile", func() {
+			By("creating a StorageProfile with an incomplete configuration")
+			sp := &cdiv1.StorageProfile{
```
akalenyu, line 78: I think we may be better using storage classes here to close out the flow (profile gets created for any class)

This is one of the reasons I am advocating using the existing tests.
We already have them cleaning up the remaining storageclass/profile after finishing
https://github.com/kubevirt/containerized-data-importer/blob/0de15cc7298921f96311d1bd2e8b079b14a06990/tests/operator_test.go#L807

akalenyu (commented):
> Awesome!
> 
> One concern though, and it might be just me over-evaluating this; aren't we basically duplicating tests? I am all up for the split to `monitoring_test.go`, but these types of tests all running twice can be quite costly. WDYT about adding the alert check over to existing tests?

@machadovilaca WDYT about this? I think running tests like the restart count one
is particularly pricey

machadovilaca:
@akalenyu I updated your singular comments and was still working on moving all monitoring related tests to the new monitoring_test file, but wasn't finished with that yet. But indeed I agree

tests/monitoring_test.go
```diff
@@ -0,0 +1,135 @@
...
+var _ = Describe("Monitoring Tests", func() {
+	f := framework.NewFramework("monitoring-test")
+
+	BeforeEach(func() {
+		if !f.IsPrometheusAvailable() {
+			Skip("This test depends on prometheus infra being available")
+		}
+	})
+
+	Context("Alerts", func() {
```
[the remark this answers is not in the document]
machadovilaca: adding it a the top of the monitoring test file is enough?
like `var _ = Describe("[Destructive] Monitoring Tests", func() {`?

tests/monitoring_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+
+	"kubevirt.io/containerized-data-importer/tests/framework"
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+const (
+	dataVolumeName = "test-dv"
+	validURL       = "http://www.example.com/example.img"
```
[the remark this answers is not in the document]
machadovilaca: i think it is a valid url, just as in `api_validation_test.go`

tests/monitoring_test.go
```diff
@@ -0,0 +1,135 @@
...
+			_, err := f.CdiClient.CdiV1beta1().CDIs().Create(context.TODO(), cdi, metav1.CreateOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			By("checking that the CDINotReady alert is firing")
+			checkAlertExists(f, "CDINotReady")
+		})
+	})
+})
+
+func checkAlertExists(f *framework.Framework, alertName string) {
```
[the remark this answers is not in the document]
machadovilaca: it will disappear when joining all monitoring tests in a single file

tests/monitoring_test.go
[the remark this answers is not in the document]
akalenyu: I think that is fine, yeah, we will double check when the lanes run anyway

akalenyu:
> @akalenyu I updated your singular comments and was still working on moving all monitoring related tests to the new monitoring_test file, but wasn't finished with that yet. But indeed I agree

Looking at the restart count test I think it won't be the end of the world if we leave it 
in it's original module and just add the alert check

tests/monitoring_test.go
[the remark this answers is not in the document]
akalenyu: ahh I see, that test module tests for invalid URL structure. I think in the monitoring module it may make sense to be called
nonExistentImageURL or something like that (the 404 is what triggers the restarts of importer)

tests/monitoring_test.go
```diff
@@ -0,0 +1,534 @@
...
+			setPrometheusRuleAlertDuration(f, originalPromRule, "CDIOperatorDown", "0m")
+			waitForPrometheusAlert(f, "CDIOperatorDown")
+
+			By("Ensuring original value of replicas restored")
+			scaleDeployment(f, deploymentName, originalReplicas)
+			err := utils.WaitForDeploymentReplicasReady(f.K8sClient, f.CdiInstallNs, deploymentName)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should throw CDIDataVolumeUnusualRestartCount alert when a DataVolume object restarts more than three times", func() {
```
akalenyu, line 295: So this test is particularly pricey because we are at the mercy of k8s exponential backoff (waits until 3 restarts)
That is one of the reasons I am advocating adding the alert check for existing tests
I think it's fine if we don't bring this one into monitoring_test.go and just keep it in https://github.com/kubevirt/containerized-data-importer/blob/ca32eba41c93728bb16b607c169054e8c2f60855/tests/import_test.go#L1004

tests/monitoring_test.go
```diff
@@ -0,0 +1,534 @@
...
+			By("checking that the CDIDataVolumeUnusualRestartCount alert is firing")
+			setPrometheusRuleAlertDuration(f, originalPromRule, "CDIDataVolumeUnusualRestartCount", "0m")
+			waitForPrometheusAlert(f, "CDIDataVolumeUnusualRestartCount")
+
+			By("deleting the DataVolume")
+			err = utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should throw CDIStorageProfilesIncomplete alert if there is a incomplete StorageProfile", func() {
```
akalenyu, line 318: Same point about duplicating tests, we can just add the alert check to the existing test,
as these are pretty pricey tests runtime wise

tests/monitoring_test.go
```diff
@@ -0,0 +1,534 @@
...
+			_, err := f.CdiClient.CdiV1beta1().StorageProfiles().Create(context.TODO(), sp, metav1.CreateOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			numAddedStorageClasses = 1
+
+			By("checking that the CDIStorageProfilesIncomplete alert is firing")
+			setPrometheusRuleAlertDuration(f, originalPromRule, "CDIStorageProfilesIncomplete", "0m")
+			waitForPrometheusAlert(f, "CDIStorageProfilesIncomplete")
+		})
+
+		It("should throw CDINotReady if CDI CR is not ready for use", func() {
```
akalenyu, line 334: Same, could add the alert check to the existing test, unless I miss something

tests/monitoring_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+		Eventually(func() int {
+			return getMetricValue(f, "kubevirt_cdi_incomplete_storageprofiles_total")
+		}, 5*time.Minute, 5*time.Second).Should(BeNumerically("==", originalMetricVal))
+	})
+
+	Context("[rfe_id:7101][crit:medium][vendor:cnv-qe@redhat.com][level:component] Metrics tests", func() {
+		var (
+			metricPollingInterval = 5 * time.Second
+			metricPollingTimeout  = 5 * time.Minute
```
akalenyu, line 85: Can move these a level up and use them for https://github.com/kubevirt/containerized-data-importer/blob/ca32eba41c93728bb16b607c169054e8c2f60855/tests/monitoring_test.go#L79

akalenyu (commented):
Looks great, just the main point about duplicating the tests
I checked the prow test runs and the destructive label is set correctly
such that the tests are only run on the destructive lane and skipped elsewhere

tests/monitoring_test.go
```diff
@@ -0,0 +1,472 @@
...
+			return err
+		}
+
+		promRule.ResourceVersion = ""
+		promRule.UID = ""
+		return f.CrClient.Create(context.TODO(), promRule)
+	}, 5*time.Minute, 1*time.Second).Should(BeNil())
+}
+
+func setPrometheusRuleAlertDuration(f *framework.Framework, promRule *promv1.PrometheusRule, alertName, duration string) {
```
akalenyu, line 378: This is not used along with `setPrometheusRule`, IIRC because the operator would just revert the changes

akalenyu (commented):
Looks really good! just the main worry that some recent changes are not included in monitoring_test.go

tests/operator_test.go
```diff
@@ -805,386 +799,6 @@ var _ = Describe("ALL Operator tests", func() {
...
-				Eventually(func() int {
-					return getMetricValue("kubevirt_cdi_incomplete_storageprofiles_total")
-				}, 2*time.Minute, 1*time.Second).ShouldNot(Equal(-1))
-			}
-
-			BeforeEach(func() {
-				cr = getCDI(f)
-				cdiPods = getCDIPods(f)
-
-				waitForIncompleteMetricInitialization()
```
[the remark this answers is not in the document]
machadovilaca: actually, I think I missed it

tests/monitoring_test.go
```diff
@@ -0,0 +1,472 @@
...
+			return false
+		}
+
+		alerts := result["data"].(map[string]interface{})["alerts"].([]interface{})
+		for _, alert := range alerts {
+			name := alert.(map[string]interface{})["labels"].(map[string]interface{})["alertname"].(string)
+			if name == alertName {
+				state := alert.(map[string]interface{})["state"].(string)
+				By(fmt.Sprintf("Alert %s state %s", name, state))
+				return state == "firing"
```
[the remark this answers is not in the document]
machadovilaca: updated

PR #2476, opened 2022-11-21
Suppress alerting for unsupported provisioners

aglitke:
Certain storage provisioners are known not to work with CDI.  The main example is provisioners that implement object storage (ie. buckets).  We don't want to populate the StorageProfile with ClaimPropertySets but there is also no need to raise an "incomplete" alert about such a StorageProfile.  The simplest approach is to simply not count these unsupported provisioners when setting the IncompleteProfileGauge.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -216,7 +216,12 @@ func (r *StorageProfileReconciler) checkIncompleteProfiles() error {
 		return err
 	}
 	for _, profile := range storageProfileList.Items {
-		if isIncomplete(profile.Status.ClaimPropertySets) {
+		if profile.Status.Provisioner == nil {
+			continue
+		}
+		// We don't count explicitly unsupported provisioners as incomplete
+		_, found := storagecapabilities.UnsupportedProvisioners[*profile.Status.Provisioner]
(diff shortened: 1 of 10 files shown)


pkg/controller/storageprofile-controller.go
```diff
@@ -216,7 +216,9 @@ func (r *StorageProfileReconciler) checkIncompleteProfiles() error {
 		return err
 	}
 	for _, profile := range storageProfileList.Items {
-		if isIncomplete(profile.Status.ClaimPropertySets) {
+		// We don't count explicitly unsupported provisioners as incomplete
+		_, found := storagecapabilities.UnsupportedProvisioners[*profile.Status.Provisioner]
```
awels, line 220: Can we ensure that whatever field we are dereferencing here is not nil before doing so. profile.Status is populated only after at least one reconcile happened, and I just want to ensure we don't get nil pointers.

awels (commented):
Looks good just one little comment

pkg/controller/storageprofile-controller.go
aglitke: Makes sense.

PR #2479, opened 2022-11-22
Update to include mount for containers dir in bazel-docker.sh

brianmcarey:

The quay registry credentials are no longer available when trying to push images using the podman based bootstrap[1]. This is due to the bootstrap image now symlinking .docker/config.json to containers/auth.json [2]. bazel-push expects the registry credentials to exist under .docker/config.json but the symlink is broken as the containers directory is not also mounted into the builder container.

This update mounts the container directory into the builder container so that the symlink can work again. This is required to run any prowjobs that publish images with the new bootstrap image. 

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-containerized-data-importer-push-nightly-ARM64/1594685597958541312 
[2] https://github.com/kubevirt/project-infra/pull/2319

Fixes #

/cc @awels @akalenyu 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/build/bazel-docker.sh
+++ b/hack/build/bazel-docker.sh
@@ -109,6 +109,11 @@ else
     volumes="-v ${BUILDER_VOLUME}:/root:rw,z,exec"
 fi
 
+if [ ${CI} == "true" ]; then
+    mkdir -p "$HOME/containers"
+    volumes="$volumes -v ${HOME}/containers:/root/containers:ro,z"
+fi
+
 if [ -n "$DOCKER_CA_CERT_FILE" ]; then
     volumes="$volumes -v ${DOCKER_CA_CERT_FILE}:${DOCKERIZED_CUSTOM_CA_PATH}:ro,z"
 fi


akalenyu:
```
Bazel push expects credentials to be available at ${HOME}/.docker/config.json
```
Shouldn't this be configurable from the bazel side? trying to see where this is implemented

brianmcarey:
> ```
> Bazel push expects credentials to be available at ${HOME}/.docker/config.json
> ```
> 
> Shouldn't this be configurable from the bazel side? trying to see where this is implemented

I think it gets set here - https://github.com/bazelbuild/rules_docker/blob/64827560cbbe9359294b9a14e8ac417224f6fdd2/container/go/cmd/pusher/pusher.go#L192

akalenyu:
> > ```
> > Bazel push expects credentials to be available at ${HOME}/.docker/config.json
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Shouldn't this be configurable from the bazel side? trying to see where this is implemented
> 
> I think it gets set here - https://github.com/bazelbuild/rules_docker/blob/64827560cbbe9359294b9a14e8ac417224f6fdd2/container/go/cmd/pusher/pusher.go#L192

interesting, so bazel is basically tied to docker CRI and that is fine?

hack/build/bazel-docker.sh
```diff
@@ -109,6 +109,11 @@ else
     volumes="-v ${BUILDER_VOLUME}:/root:rw,z,exec"
 fi
 
+if [ ${CI} == "true" ]; then
```
akalenyu, line 112: do we ever set this though?

brianmcarey: Its an environment variable that gets added by prow.
```
						{
							"name": "CI",
							"value": "true"
						},
```
([source](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2474/pull-cdi-unit-test/1595070406727831552/podinfo.json))

brianmcarey: Prow doc - https://github.com/kubernetes-sigs/prow/blob/main/site/content/en/docs/Legacy%20Snapshot/prow/jobs.md#job-environment-variables

PR #2480, opened 2022-11-23
Make cluster-sync more friendly for external provider

akalenyu:
The aim of this change is to make setting up a dev environment on external cluster easier, By doing mainly 2 things:
- Bazel push (CDI containers/e2e testing containers) expects external registry creds to be in the docker CRI default path.
This seems to be half-configurable on the bazel side, except for the [actual file name](https://github.com/bazelbuild/rules_docker/blob/64827560cbbe9359294b9a14e8ac417224f6fdd2/container/go/cmd/pusher/pusher.go#L192) (which differs for podman).
Now we prompt for the credentials and save them on the expected path.
- Deploy the e2e testing manifests on external provider as well
Not sure why not do that today? probably reason dates way back

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/clean.sh
+++ b/cluster-sync/clean.sh
@@ -7,14 +7,14 @@ source ./cluster-up/cluster/${KUBEVIRT_PROVIDER}/provider.sh
 echo "Cleaning up ..."
 
 if [ "${CDI_CLEAN}" == "test-infra" ]; then
-  _kubectl delete all -n cdi -l cdi.kubevirt.io/testing
+  _kubectl delete all -n ${CDI_NAMESPACE} -l cdi.kubevirt.io/testing
   exit 0
 fi
 
 OPERATOR_CR_MANIFEST=./_out/manifests/release/cdi-cr.yaml
 OPERATOR_MANIFEST=./_out/manifests/release/cdi-operator.yaml
 LABELS=("operator.cdi.kubevirt.io" "cdi.kubevirt.io" "prometheus.cdi.kubevirt.io")
-NAMESPACES=(default kube-system cdi)
+NAMESPACES=(default kube-system "${CDI_NAMESPACE}")
 
 _kubectl get ot --all-namespaces -o=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace,FINALIZERS:.metadata.finalizers --no-headers | grep objectTransfer | while read p; do
     arr=($p)
@@ -48,7 +48,7 @@ if [ -f "${OPERATOR_CR_MANIFEST}" ]; then
 	echo "Cleaning CR object ..."
     if _kubectl get crd cdis.cdi.kubevirt.io ; then
         _kubectl delete --ignore-not-found -f "${OPERATOR_CR_MANIFEST}"
-        _kubectl wait cdis.cdi.kubevirt.io/${CR_NAME} --for=delete --timeout=30s | echo "this is fine"
+        _kubectl wait cdis.cdi.kubevirt.io/${CR_NAME} --for=delete --timeout=30s || echo "this is fine"
     fi
 fi
 
--- a/cluster-sync/external/README.md
+++ b/cluster-sync/external/README.md
@@ -1,3 +1,7 @@
+# Introduction
+Sometimes one would want to setup CDI development environment on a totally external cluster (not kubevirtci).  
+We try to support this, with this document aggregating some knowledge around that.
+
 # External Kubernetes Provider
 
 This provider works with an existing Kubernetes cluster.  You'll want to configure your own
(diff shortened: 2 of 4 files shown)


cluster-sync/clean.sh
```diff
@@ -56,7 +56,7 @@ set +e
 _kubectl patch cdi ${CR_NAME} --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
 set -e
 
-if [ "${CDI_CLEAN}" == "all" ] && [ -f "${OPERATOR_MANIFEST}" ]; then
+if [ -f "${OPERATOR_MANIFEST}" ]; then
```
akalenyu, line 59: A little worried about this, maybe there was a good reason to keep the operator around..

cluster-sync/clean.sh
```diff
@@ -7,7 +7,7 @@ source ./cluster-up/cluster/${KUBEVIRT_PROVIDER}/provider.sh
 echo "Cleaning up ..."
 
 if [ "${CDI_CLEAN}" == "test-infra" ]; then
-  _kubectl delete all -n cdi -l cdi.kubevirt.io/testing
+  _kubectl delete all -n ${CDI_NAMESPACE} -l cdi.kubevirt.io/testing
```
awels, line 10: nice catch

cluster-sync/clean.sh
awels: So Arnon did a clean of just cdi or just the test manifests or both. And I think in some of those cases we want to keep the operator around.

cluster-sync/external/README.md
```diff
@@ -7,30 +11,26 @@ such that:
...
-We use the same workflow as the ephemeral dev environment, but skip `make cluster-up`:
+We use the same workflow as the ephemeral dev environment, but skip `make cluster-up`.  
+A suggested flow:
 
 ```bash
-export DOCKER_PREFIX=index.docker.io/barney_rubble
-export DOCKER_TAG=latest # defaults to `latest`
 export KUBEVIRT_PROVIDER=external
+export PULL_POLICY=Always
+export CDI_NAMESPACE=cdi-rand-ns-name
```
awels, line 22: instead of cdi-rand-ns-name, I would say something like namespace cdi is installed into. Defaults to 'cdi'

cluster-sync/clean.sh
akalenyu: Does this effectively mean, that when I am developing CDI operator changes and running `cluster-sync`, I am not getting the latest operator build?

cluster-sync/clean.sh
akalenyu: Yeah, apparently there are some more resources we want to get rid of that are not namespaced..
scc for example, which doesn't show up in "get all"

cluster-sync/external/README.md
akalenyu: I think going with random is a good practice for external clusters, completes the package of testing CDI with ensuring we are name agnostic

awels: This is the documentation on how to work with an external cluster, the namespace will be pretty set in that scenario. We randomize in our CI anyway to catch hardcoding.

cluster-sync/clean.sh
awels: No a cluster-sync will delete everything, but a cluster-sync-test-infra will keep cdi around and remove the test infra, and cluster-sync-cdi will remove cdi (including operator) and keep the test infra.

cluster-sync/external/README.md
akalenyu: I just think its a good idea to go random here instead of default..

akalenyu: The ns will not be set in that scenario, I am working on a clean openshift cluster and dumping CDI into a fresh ns

cluster-sync/clean.sh
akalenyu: `cluster-sync-cdi` will not removed the operator unless Im missing something (note `"${CDI_CLEAN}" == "all"`)

akalenyu: Would really love a second look at this because this is probably the most breaking change in this PR

akalenyu: @awels @arnongilboa

cluster-sync/clean.sh
```diff
@@ -7,14 +7,14 @@ source ./cluster-up/cluster/${KUBEVIRT_PROVIDER}/provider.sh
...
   exit 0
 fi
 
 OPERATOR_CR_MANIFEST=./_out/manifests/release/cdi-cr.yaml
 OPERATOR_MANIFEST=./_out/manifests/release/cdi-operator.yaml
 LABELS=("operator.cdi.kubevirt.io" "cdi.kubevirt.io" "prometheus.cdi.kubevirt.io")
-NAMESPACES=(default kube-system cdi)
```
akalenyu, line 17: actually nvm, this was the reason my operator wasn't being cleared

cluster-sync/clean.sh
akalenyu: actually, we don't need this, the reason was elsewhere https://github.com/kubevirt/containerized-data-importer/pull/2480/files#r1045631832

awels (approved):
/approve

PR #2481, opened 2022-11-23
Bump kubevirtci, bump k8s versions used in CI

maya-r:
**Release note**:

```release-note
NONE
```


awels:
The failures seem legit, can you see why the api test is failing (maybe the failure message changed again?)

akalenyu:
/cherrypick release-v1.55

We need the test change commit for testing against openshift,
we could also use the newer k8s lanes on this release though

PR #2487, opened 2022-11-27
[Draft] Annotate pvc for completion

maya-r:
This was inspired by two things:
- the desire to simplify backup/restore to not require custom code
- a [race seen in kubevirt's CI](https://prow.ci.kubevirt.io/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/8796/pull-kubevirt-e2e-k8s-1.24-sig-storage-root/1592810005592018944), which we can simplify ignore.

The race is explained as follows:
We try to see if the PVC is ready to be consumed. We look at the PVC, and it references a DV so we need to inspect its status.
In the meantime, garbage collection has happened, so when we look at the DV, it no longer exists.

However, it's a lot of work for the small window before a data volume is garbage collected - assuming this DV will be garbage collected. Perhaps we don't want to do this at all, and just assume from now on that datavolumes are garbage collected, or all of them will be real soon now.

**Release note**:

```release-note
Populate datavolume AnnPrePopulated, PVC AnnPopulatedFor when a DV is complete, to simplify writing backup/restore code.
Add a new annotation allowing to eliminate a race condition that resulted in IsPopulated in the CDI API sometimes returning an error
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -573,6 +573,9 @@ func (r *DatavolumeReconciler) Reconcile(_ context.Context, req reconcile.Reques
 			return reconcile.Result{}, err
 		}
 	} else {
+		if err := r.annotateForCompletion(datavolume, pvc); err != nil {
+			return reconcile.Result{}, err
+		}
 		res, err := r.garbageCollect(datavolume, pvc, log)
 		if err != nil {
 			return reconcile.Result{}, err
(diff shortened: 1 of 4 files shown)


mhenriks:
I think it is better to just not generate a warning in this case.  GC of a DataVolume involves updating two resources.  That cannot be done atomically.  Clients should expect/handle that in the regular kubernetes way of exiting the reconcile loop and retrying

maya-r:
Closing as we expect consumers to handle this small window of inconsistency.

PR #2494, opened 2022-11-30
[release-1.43] Queue DataVolumes referred to in "populatedFor" PVC annotations

mhenriks:

Fixes #

https://bugzilla.redhat.com/show_bug.cgi?id=2149671

Manual backport of https://github.com/kubevirt/containerized-data-importer/pull/2373

**Release note**:

```release-note
Fix DataVolume/PVC restore - don't require PVC initially owned by DataVolume
```

### DIFF
--- a/pkg/controller/datavolume-controller.go
+++ b/pkg/controller/datavolume-controller.go
@@ -349,10 +349,31 @@ func addDatavolumeControllerWatches(mgr manager.Manager, datavolumeController co
 	if err := datavolumeController.Watch(&source.Kind{Type: &cdiv1.DataVolume{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return err
 	}
-	if err := datavolumeController.Watch(&source.Kind{Type: &corev1.PersistentVolumeClaim{}}, &handler.EnqueueRequestForOwner{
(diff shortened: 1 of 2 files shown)


awels:
Hmm so we should not have DV GC enabled here...

mhenriks:
> Hmm so we should not have DV GC enabled here...

Yes that is how these backport PRs differ from the original.  Do you see an issue?

awels:
So I am guessing that this lane is failing due to us not backporting the fix into this branch and stuff getting garbage collected, and the DVs disappearing. We can override the lane or backport the fix.

mhenriks:
> So I am guessing that this lane is failing due to us not backporting the fix into this branch and stuff getting garbage collected, and the DVs disappearing. We can override the lane or backport the fix.

What fix are you talking about?

awels:
So the problem is that we are installing the 'latest released' and then 'upgrading to the contents of the PR' but in this case we are actually downgrading. So what happens is we install 1.55, then upgrade to 1.43, during the 1.55 the GC will delete the DV used for the failing tests, and the test will fail. This was fixed with https://github.com/kubevirt/containerized-data-importer/pull/2434 on main. We should probably check the branch, we are on and not install a version newer than what we are on when running tests.

PR #2512, opened 2022-12-20
Add some logging for better debugging of host assisted clone

akalenyu:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -323,6 +323,24 @@ var (
 	apiServerKey     *rsa.PrivateKey
 )
 
+// SizeMismatchError is an error returned if source size is larger than target
+type SizeMismatchError struct {
+	targetQuantity resource.Quantity
+	sourceQuantity resource.Quantity
+}
+
+func (e *SizeMismatchError) Error() string {
+	return fmt.Sprintf("target resources requests storage size %d is smaller than the source %d", e.targetQuantity.Value(), e.sourceQuantity.Value())
+}
+
+// NewSizeMismatchError returns a new SizeMismatchError
+func NewSizeMismatchError(target, source resource.Quantity) *SizeMismatchError {
+	return &SizeMismatchError{
+		targetQuantity: target,
+		sourceQuantity: source,
+	}
+}
+
 // FakeValidator is a fake token validator
 type FakeValidator struct {
 	Match     string
@@ -965,7 +983,7 @@ func ValidateRequestedCloneSize(sourceResources corev1.ResourceRequirements, tar
 
 	// Verify that the target PVC size is equal or larger than the source.
 	if sourceRequest.Value() > targetRequest.Value() {
-		return errors.Errorf("target resources requests storage size is smaller than the source %d < %d", targetRequest.Value(), sourceRequest.Value())
+		return NewSizeMismatchError(targetRequest, sourceRequest)
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


pkg/controller/common/util.go
```diff
@@ -765,7 +783,7 @@ func ValidateRequestedCloneSize(sourceResources corev1.ResourceRequirements, tar
 	targetRequest := targetResources.Requests[corev1.ResourceStorage]
 	// Verify that the target PVC size is equal or larger than the source.
 	if sourceRequest.Value() > targetRequest.Value() {
```
ShellyKa13, line 785: maybe if you are already changing stuff here you can use: sourceUsableSpace.Cmp(targetUsableSpace) > 0 like done in the other check instead? I think thats the better way? not a must

pkg/controller/util_test.go
```diff
@@ -285,7 +285,7 @@ var _ = Describe("ValidateClone", func() {
 
 		err := ValidateClone(sourcePvc, dvSpec)
 		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(ContainSubstring("target resources requests storage size is smaller than the source"))
+		Expect(err.Error()).To(ContainSubstring("is smaller than the source"))
```
ShellyKa13, line 288: if possible maybe you should check the full err including the the values that you added to the log? it will give a fuller cover to the tests

ShellyKa13 (commented):
Nice! a couple of suggestions

pkg/controller/common/util.go
akalenyu: Prefer to keep this PR on the safe side

pkg/controller/common/util.go
```diff
@@ -279,6 +279,24 @@ var (
...
+	targetQuantity resource.Quantity
+	sourceQuantity resource.Quantity
+}
+
+func (e *SizeMismatchError) Error() string {
+	return fmt.Sprintf("target resources requests storage size %d is smaller than the source %d", e.targetQuantity.Value(), e.sourceQuantity.Value())
+}
+
+// NewSizeMismatchError returns a new SizeMismatchError
+func NewSizeMismatchError(target, source resource.Quantity) *SizeMismatchError {
```
alromeros, line 293: Any reasons to add the new error type? I'm not opposed to it, but at first glance seemed overcomplicated for a simple string.

pkg/controller/datavolume/controller-base.go
```diff
@@ -620,7 +620,7 @@ func (r *ReconcilerBase) reconcileProgressUpdate(datavolume *cdiv1.DataVolume, p
 			return err
 		}
 	}
-	// We are not done yet, force a re-reconcile in 2 seconds to get an update.
+	r.log.V(3).Info("We are not done yet, force a re-reconcile in 2 seconds to get an update.")
```
alromeros, line 623: Nitpick: I think the `We are not done yet` part sounds too informal to have in the logs, maybe I would use `Datavolume not ready yet` or `Datavolume not in succeeded phase`. wdyt?

alromeros (commented):
Looks good to me, but I want to know the motivation behind the new error type. Also, giving a review to bring attention and hopefully merge it soon.

pkg/controller/datavolume/controller-base.go
akalenyu: This is a V(3) log so hidden unless verbosity explicitly set this high

pkg/controller/common/util.go
akalenyu: Yeah - that string was used in different pkgs, and needed to parameterize it with values,
So I thought it is a little nicer than a func with the sizes as params

pkg/controller/datavolume/controller-base.go
akalenyu: done

PR #2520, opened 2022-12-22
Fix bug in size-detection mechanism when cloning across namespaces

alromeros:

This PR changes the owner of the size-detection pod (from the target DataVolume to the source PVC) to avoid errors when cloning across namespaces.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2152537

**Release note**:

```release-note
[bug-fix] Fix bug in size-detection mechanism when cloning across namespaces
```

### DIFF
--- a/pkg/controller/datavolume/clone-controller.go
+++ b/pkg/controller/datavolume/clone-controller.go
@@ -1920,6 +1920,21 @@ func (r *CloneReconciler) makeSizeDetectionPodSpec(
 		},
 	}
 
+	if sourcePvc.Namespace == dv.Namespace {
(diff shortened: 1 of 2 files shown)


akalenyu:
> (from the target DataVolume to the source PVC)

Is this correct? looks like you're still setting the DV as owner just via an annotation if ns differs

alromeros:
> > (from the target DataVolume to the source PVC)
> 
> Is this correct? looks like you're still setting the DV as owner just via an annotation if ns differs

Thanks for noticing Alex, the description was referencing a previous commit that I then changed, so it wasn't accurate.

PR #2521, opened 2022-12-27
Add `operator_health_impact` label to CDI alerts

assafad:

This PR adds to each of CDI alerts a label named `operator_health_impact`, which indicates how the alerts impact the operator health. This will enable us filtering alerts based on their impact on the operator health, and to later have an operator health metric, which will tell the user what is the aggregated operator health.

This label gets one of the following values:
- `critical` indicates that the alert fires due to an issue that impacts the operator's functionality badly, and an action must be taken immediately in order to solve it.
-  `warning` indicates that the alert fires due to an issue that soon might impact the operator's functionality badly, and the user should be warned in order to solve the issue.
- `none` indicates that the alert fires due to an issue that doesn't affect the operator health, and in most cases is related to the workload rather than the operator itself.

https://issues.redhat.com/browse/CNV-18923

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/operator/controller/prometheus.go
+++ b/pkg/operator/controller/prometheus.go
@@ -43,16 +43,17 @@ import (
 )
 
 const (
-	ruleName                 = "prometheus-cdi-rules"
-	rbacName                 = "cdi-monitoring"
-	monitorName              = "service-monitor-cdi"
-	defaultMonitoringNs      = "monitoring"
-	runbookURLBasePath       = "https://kubevirt.io/monitoring/runbooks/"
-	severityAlertLabelKey    = "severity"
-	partOfAlertLabelKey      = "kubernetes_operator_part_of"
-	partOfAlertLabelValue    = "kubevirt"
-	componentAlertLabelKey   = "kubernetes_operator_component"
-	componentAlertLabelValue = common.CDILabelValue
+	ruleName                  = "prometheus-cdi-rules"
+	rbacName                  = "cdi-monitoring"
+	monitorName               = "service-monitor-cdi"
+	defaultMonitoringNs       = "monitoring"
+	runbookURLBasePath        = "https://kubevirt.io/monitoring/runbooks/"
+	severityAlertLabelKey     = "severity"
+	healthImpactAlertLabelKey = "operator_health_impact"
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/prometheus.go
```diff
@@ -192,9 +196,10 @@ func getAlertRules() []promv1.Rule {
 				"runbook_url": runbookURLBasePath + "CDIStorageProfilesIncomplete",
 			},
 			map[string]string{
-				severityAlertLabelKey:  "info",
-				partOfAlertLabelKey:    partOfAlertLabelValue,
-				componentAlertLabelKey: componentAlertLabelValue,
+				severityAlertLabelKey:     "info",
+				healthImpactAlertLabelKey: "warning",
```
sradco, line 200: @assafad  We should consider this as even a critical impact on operator health, since the  CDI cannot create a VM disk on the PVC. Its an important functionality. But I think someone from the CDI team should advise here.

pkg/operator/controller/prometheus.go
```diff
@@ -206,9 +211,10 @@ func getAlertRules() []promv1.Rule {
 				"runbook_url": runbookURLBasePath + "CDIDataImportCronOutdated",
 			},
 			map[string]string{
-				severityAlertLabelKey:  "info",
-				partOfAlertLabelKey:    partOfAlertLabelValue,
-				componentAlertLabelKey: componentAlertLabelValue,
+				severityAlertLabelKey:     "info",
+				healthImpactAlertLabelKey: "warning",
```
sradco, line 215: @assafad Same as above, There is a loss of functionality and I think we should consider making this critical impact.

pkg/operator/controller/prometheus.go
akalenyu: I can see arguments for both, but I am in favor of `warning` since it's still possible to specify the missing properties manually on the DV object

pkg/operator/controller/prometheus.go
akalenyu: I think critical is too aggressive here, I would save it for the whole install being messed up and things of that kind

assafad:
@mhenriks anything we're missing in order to merge this one?

PR #2355, opened 2022-07-07
[WIP] PoC: Smart Clone from existing snapshot

akalenyu:
Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

**What this PR does / why we need it**:
Today's smart cloning methods use a 1:1 cloning method which creates a temporary snapshot of the source PVC for each intended clone.
Some storage systems (possibly ceph) are designed to scale better with a 1:N method where a single snapshot of the source PVC is used to create multiple clones.

Once we have this PoC we can compare it at scale to the current approach to see if it makes sense to productize it.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #2522, opened 2022-12-29
Clone from VolumeSnapshot source

akalenyu:
Following a successful [PoC](https://github.com/kubevirt/containerized-data-importer/pull/2355), introducing
cloning from a volumesnapshot source with the main idea being cloning a bunch of times from a single snapshot
("golden snapshot").

Follow ups needed:
- [x] Clone token mechanism extend for VolumeSnapshots
- [ ] Take into account new volumesnapshot API `sourceVolumeMode` and converting volume mode - [prerequisite](https://github.com/kubevirt/kubevirtci/pull/939) (follow-up)
- [ ] `SourceRef` support (follow-up)

Fixes #

Considering getting rid of expansion logic for clones from snapshot, discussed in the past at
https://github.com/kubevirt/containerized-data-importer/pull/2276#issuecomment-1127742285

**Release note**:

```release-note
Clone from VolumeSnapshot source
```


pkg/controller/datavolume/clone-controller.go
```diff
@@ -65,6 +65,8 @@ const (
 	CloneInProgress = "CloneInProgress"
 	// SnapshotForSmartCloneInProgress provides a const to indicate snapshot creation for smart-clone is in progress
 	SnapshotForSmartCloneInProgress = "SnapshotForSmartCloneInProgress"
+	// CloneFromSnapshotSourceInProgress provides a const to indicate clone from snapshot source is in progress
```
mhenriks, line 68: I won't get to into the nuts an bolts of this implementation until rebased on #2525 but I'm wondering if PVC source and snapshot source should be two different controllers?  May be more readable than having everything in this file?

mhenriks (commented):
REally great start, some high leve comments for now

pkg/controller/datavolume/clone-controller.go
akalenyu: Yeah I was holding back the rebase push until the review so I don't throw anyone off, my thinking is:

- Lots of shared functionality with clone => need to make a clone controller template to reuse => not sure it's worth the overcomplication when snapshots should be the last "clone" source we add
(we go all PVCs soon, people will be doing restores themselves, otherwise, they'll use host assisted populator?)
- Path distinction between PVC & Snapshot source happens very early on so not too horrible?

pkg/clone/auth.go
```diff
@@ -75,13 +83,57 @@ func CanServiceAccountClonePVC(client SubjectAccessReviewsProxy, pvcNamespace, p
...
+		User:   userInfo.Username,
+		Groups: userInfo.Groups,
+		Extra:  newExtra,
+	}
+
+	return sendSubjectAccessReviewsSnapshot(client, sourceNamespace, pvcName, sarSpec)
+}
+
+// CanServiceAccountCloneSnapshot checks if a ServiceAccount has "appropriate" permission to clone from the given snapshot
+func CanServiceAccountCloneSnapshot(client SubjectAccessReviewsProxy, pvcNamespace, pvcName, saNamespace, saName string) (bool, string, error) {
```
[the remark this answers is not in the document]
akalenyu: Yeah I'd like that, but I had doubts if we can make these accept a client param
(remember we have DataSources so we can't just pass a DV.Spec.Source and infer from that)

Another concern with this kind of revamp is whether it fits in this PR,
I feel like it'll add a layer of complexity

pkg/controller/datavolume/clone-controller_test.go
```diff
@@ -43,6 +43,7 @@ import (
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	. "kubevirt.io/containerized-data-importer/pkg/controller/common"
+	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
```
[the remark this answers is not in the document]
akalenyu: thanks! fixed

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -193,3 +225,13 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 
 	return toPatchResponse(dataVolume, modifiedDataVolume)
 }
+
+func getTokenResource(pvcSource *cdiv1.DataVolumeSourcePVC, snapshotSource *cdiv1.DataVolumeSourceSnapshot) (metav1.GroupVersionResource, error) {
+	if pvcSource != nil {
+		return tokenResourcePvc, nil
+	} else if snapshotSource != nil {
+		return tokenResourceSnapshot, nil
+	}
```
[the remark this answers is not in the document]
akalenyu: yep definitely, thanks
hopefully this is better now

pkg/clone/auth.go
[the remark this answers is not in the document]
mhenriks: Yeah, would probably want to do something similar to [1] to get the DataSource.  Kind of ugly.  Your call.  I'm not going to be a stickler about this.

[1] https://github.com/kubevirt/kubevirt/blob/2baf5303dddfe71c98a2510357c464a8f7bdd3ca/pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go#L61-L67

pkg/controller/datavolume/clone-controller.go
mhenriks: Totally understand that there are many shared operations between pvc and snapshot source.  My gut tells me that snapshot should be a dedicated controller.  What do you think @arnongilboa?

pkg/clone/auth.go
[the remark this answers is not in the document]
akalenyu: yeah I would prefer to push this out of this PR
but I think that regardless of this PR, having kubevirt "know less" about types of clones is a good thing
should I open an issue?

pkg/controller/datavolume/clone-controller.go
arnongilboa: I tend to think snapshot should have a dedicated controller, and both should have a common base (like the current DV controllers). Since it introduces lots of new code, maintenance will be much easier this way. We should learn the lesson from the huge DV controller spagetti dish we had before the split/refactoring.

akalenyu: Yep totally makes sense!
Let's hop down the bunny trail over to https://github.com/kubevirt/containerized-data-importer/pull/2544

akalenyu: @mhenriks @arnongilboa implementation with separate controller ready for review

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,486 @@
...
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	return reconcile.Result{}, r.syncCloneStatusPhase(syncRes, cdiv1.CloneFromSnapshotSourceInProgress, nil)
+}
+
+func (r *SnapshotCloneReconciler) createTempHostAssistedSourcePvc(dv *cdiv1.DataVolume, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec, syncRes *dataVolumeCloneSyncResult) error {
+	tempHostAssistedSourcePvc, err := r.makePvcFromSnapshot(snapshot.Name, dv, snapshot, targetPvcSpec)
```
mhenriks, line 345: it `snapshot.Name` is not a good name for the temp pvc.  what about concurrent clones?  The create will fail for the second.

It would of course be nice to have a single shared temp PVC but I don't think that was your intention.  And implementing that scheme would add complexity.  Maybe a subsequent pr?

tests/cloner_test.go
```diff
@@ -2537,6 +2538,106 @@ var _ = Describe("all clone tests", func() {
...
+				path = utils.DefaultPvcMountPath
+			}
+			same, err := f.VerifyTargetPVCContentMD5(f.Namespace, pvc, path, utils.UploadFileMD5, utils.UploadFileSize)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(same).To(BeTrue())
+		},
+			Entry("with filesystem single clone", v1.PersistentVolumeMode(v1.PersistentVolumeFilesystem), 1),
+			Entry("with filesystem multiple clones", v1.PersistentVolumeMode(v1.PersistentVolumeFilesystem), 5),
+			Entry("with block single clone", v1.PersistentVolumeMode(v1.PersistentVolumeBlock), 1),
+			Entry("with block multiple clones", v1.PersistentVolumeMode(v1.PersistentVolumeBlock), 5),
```
mhenriks, line 2638: wonder how these multiple tests are working with the shared PVC?

mhenriks: Can we be more explicit about when doing efficient snapshot clone vs dumb clone?

mhenriks (commented):
Looking good!  Here's another pass

tests/cloner_test.go
akalenyu: The func tests are only doing efficient right now:
https://github.com/kubevirt/containerized-data-importer/blob/e62c96ffe0a76655466828fbfba8e646a672e381/tests/cloner_test.go#L2562

The reason I didn't do dumb was because it was hard to simulate a real env for the fallback to happen,
but I think I can give it a try now that the sourceVolumeMode stuff is in

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,486 @@
...
+
+func (r *SnapshotCloneReconciler) evaluateFallBackToHostAssistedNeeded(datavolume *cdiv1.DataVolume, pvcSpec *corev1.PersistentVolumeClaimSpec, snapshot *snapshotv1.VolumeSnapshot) (bool, error) {
+	bindingMode, err := r.getStorageClassBindingMode(pvcSpec.StorageClassName)
+	if err != nil {
+		return true, err
+	}
+
+	// Storage classes do not match means we can only do dumb cloning
+	if snapshot.Spec.VolumeSnapshotClassName == nil || *snapshot.Spec.VolumeSnapshotClassName == "" {
+		return true, fmt.Errorf("Snapshot %s/%s does not have volume snap class populated, can't clone", snapshot.Name, snapshot.Namespace)
```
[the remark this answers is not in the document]
akalenyu: Thanks!
I had to add `-checks=all` in vscode settings, under `go extension->lint flags` for them to show up for me

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: You're right, it was not my intention, I am cleaning up the "shared" PVC once the corresponding DV is done.
BTW - the creation won't actually fail. If the PVC exists, we will just attempt to use it.

Just to make sure I understand, do you mean we should switch to unique PVC per fallback dumb clone for this PR,
and follow up later with attempts to clone from shared?

BTW' - I played around with concurrent host-assisted clones (both PVC & snapshot source)
Seems we have a problem outside of this PR, the source pod gets restarted multiple times before completing
```bash
I0122 14:58:03.462143      10 prometheus.go:72] 25.53
F0122 14:58:03.536422      10 clone-source.go:41] Subprocess did not execute successfully, result is: '\x02'
disk.img
/usr/bin/tar: disk.img: Read error at byte 1119744, while reading 512 bytes: Permission denied
/usr/bin/tar: Exiting with failure status due to previous errors

NAMESPACE   NAME                  PHASE             PROGRESS   RESTARTS   AGE
default     cloned-datavolume-1   CloneInProgress   24.66%     3          2m13s
```

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,486 @@
...
+						message:   err.Error(),
+					})
+			}
+			return syncRes, err
+		}
+		pvc = targetHostAssistedPvc
+	}
+
+	if fallBackToHostAssisted {
+		if err := r.ensureExtendedToken(pvc); err != nil {
```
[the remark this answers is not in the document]
akalenyu: Great point! I see it's really possible for readyToUse to take a while:
https://github.com/kubevirt/kubevirt/pull/8640

Thing is, we need a target PVC to exist for `ensureExtendedToken`.
Should I drop waiting for readyToUse, since everything will converge eventually when it becomes ready?

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,486 @@
...
+	if syncErr != nil || syncRes.result != nil {
+		return syncRes, syncErr
+	}
+
+	pvc := syncRes.pvc
+	pvcSpec := syncRes.pvcSpec
+	datavolume := syncRes.dvMutated
+	transferName := getTransferName(datavolume)
+
+	pvcPopulated := pvcIsPopulated(pvc, datavolume)
```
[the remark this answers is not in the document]
akalenyu: I see various controllers check this differently, I just stuck close the PVC clone controller
but I think it should be possible. @arnongilboa wdyt?

tests/cloner_test.go
akalenyu: Apparently beta for this feature is 1.27
https://github.com/kubernetes/enhancements/issues/3141

So unless we enable stuff explicitly we won't have sourceVolumeMode populated for now

pkg/clone/auth.go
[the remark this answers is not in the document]
akalenyu: https://github.com/kubevirt/containerized-data-importer/issues/2552

pkg/controller/datavolume/snapshot-clone-controller.go
mhenriks: > Just to make sure I understand, do you mean we should switch to unique PVC per fallback dumb clone for this PR,
and follow up later with attempts to clone from shared?

yup

tests/cloner_test.go
akalenyu: I have something more hacky in mind, but at least we won't be tied to "local". Going to use AllowVolumeExpansion
Working on this right now

akalenyu: Added host assisted tests

> wonder how these multiple tests are working with the shared PVC?

I guess because the temp source PVC has some k8s finalizers that protect it from being removed completely?

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: So yeah shared source seems to work, but it may be ugly that we try deleting it, should we protect it with `GetPodsUsingPVCs`?

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,486 @@
...
+		return syncRes, nil
+	}
+
+	nn := types.NamespacedName{Namespace: datavolume.Spec.Source.Snapshot.Namespace, Name: datavolume.Spec.Source.Snapshot.Name}
+	snapshot := &snapshotv1.VolumeSnapshot{}
+	if err := r.client.Get(context.TODO(), nn, snapshot); err != nil {
+		return syncRes, err
+	}
+	if snapshot.Status == nil || snapshot.Status.ReadyToUse == nil || !*snapshot.Status.ReadyToUse {
+		return syncRes, fmt.Errorf("Snapshot %s/%s not ready to use, can't clone", snapshot.Name, snapshot.Namespace)
```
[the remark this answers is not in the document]
akalenyu: done

pkg/controller/datavolume/snapshot-clone-controller.go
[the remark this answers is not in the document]
akalenyu: done

pkg/controller/common/util.go
```diff
@@ -602,15 +631,38 @@ func ValidateCloneTokenDV(validator token.Validator, dv *cdiv1.DataVolume) error
...
+	return ""
+}
+
+func getTokenResourceNamePvc(sourcePvc *corev1.PersistentVolumeClaim) string {
+	if v, ok := sourcePvc.Labels[common.CDIComponentLabel]; ok && v == common.CloneFromSnapshotFallbackPVCCDILabel {
+		return "volumesnapshots"
+	}
+
+	return "persistentvolumeclaims"
 }
```
arnongilboa, line 658: using (typed?) consts might be better here for "persistentvolumeclaims" and "volumesnapshots"

pkg/controller/common/util.go
```diff
@@ -1088,3 +1140,18 @@ func NewImportDataVolume(name string) *cdiv1.DataVolume {
 		},
 	}
 }
+
+// GetCloneSourceNameAndNamespace returns the name and namespace of the cloning source
```
arnongilboa, line 1144: maybe return NamespacedName?

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -125,7 +127,9 @@ const (
 
 	annCloneType = "cdi.kubevirt.io/cloneType"
 
-	cloneControllerName = "datavolume-clone-controller"
+	pvcCloneControllerName = "datavolume-pvc-clone-controller"
+
+	snapshotCloneControllerName = "datavolume-snapshot-clone-controller"
```
arnongilboa, line 132: consts should be in their respective files

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -283,7 +278,7 @@ func (r *CloneReconcilerBase) doCrossNamespaceClone(log logr.Logger,
 	return nil, nil
 }
 
-func (r *CloneReconcilerBase) initTransfer(log logr.Logger, syncRes *dataVolumeCloneSyncResult, name string) (bool, error) {
```
arnongilboa, line 286: you have `GetCloneSourceNameAndNamespace()` so why add namespace param here and in `doCrossNamespaceClone()`?

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -577,12 +617,90 @@ func (r *CloneReconcilerBase) syncDataVolumeStatusPhaseWithEvent(
...
+	if dv.Spec.Source.PVC != nil {
+		return dv.Spec.Source.PVC.Namespace != "" && dv.Spec.Source.PVC.Namespace != dv.Namespace
+	}
+
+	if dv.Spec.Source.Snapshot != nil {
+		return dv.Spec.Source.Snapshot.Namespace != "" && dv.Spec.Source.Snapshot.Namespace != dv.Namespace
 	}
 
-	return dv.Spec.Source.PVC.Namespace != "" && dv.Spec.Source.PVC.Namespace != dv.Namespace
+	return false
```
arnongilboa, line 703: why not use `GetCloneSourceNameAndNamespace()` here?

pkg/controller/datavolume/controller-base.go
```diff
@@ -219,8 +220,12 @@ func addDataVolumeControllerCommonWatches(mgr manager.Manager, dataVolumeControl
 
 func getDataVolumeOp(dv *cdiv1.DataVolume) dataVolumeOp {
 	src := dv.Spec.Source
+
 	if (src != nil && src.PVC != nil) || dv.Spec.SourceRef != nil {
-		return dataVolumeClone
+		return dataVolumePvcClone
+	}
+	if (src != nil && src.Snapshot != nil) || dv.Spec.SourceRef != nil {
+		return dataVolumeSnapshotClone
```
arnongilboa, line 228: The `|| dv.Spec.SourceRef != nil` in both ifs makes it order dependent. Are you sure that's what we want here?

pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -235,7 +235,7 @@ func (r PvcCloneReconciler) prepare(syncRes *dataVolumeSyncResult) error {
 	if err := r.populateSourceIfSourceRef(dv); err != nil {
 		return err
 	}
-	if isCrossNamespaceClone(dv) && dv.Status.Phase == cdiv1.Succeeded {
```
arnongilboa, line 238: why cleanup when not needed?

pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -347,7 +352,8 @@ func (r *SmartCloneReconciler) getTargetPVC(dataVolume *cdiv1.DataVolume) (*core
 	return pvc, nil
 }
 
-func newPvcFromSnapshot(snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec) (*corev1.PersistentVolumeClaim, error) {
+func newPvcFromSnapshot(name string, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec) (*corev1.PersistentVolumeClaim, error) {
+	targetPvcSpecCopy := targetPvcSpec.DeepCopy()
```
arnongilboa, line 356: why do we need `targetPvcSpecCopy`?

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,488 @@
...
+	storagev1 "k8s.io/api/storage/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	"kubevirt.io/containerized-data-importer/pkg/common"
+	"kubevirt.io/containerized-data-importer/pkg/util"
+
```
arnongilboa, line 38: -

pkg/controller/datavolume/pvc-clone-controller.go
akalenyu: just moved the cross ns check to https://github.com/kubevirt/containerized-data-importer/blob/2137848f19209350b1a28fa8958955839e7f6dcf/pkg/controller/datavolume/pvc-clone-controller.go#L714-L719

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,488 @@
...
+	return nil
+}
+
+// Reconcile loop for the clone data volumes
+func (r SnapshotCloneReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithValues("DataVolume", req.NamespacedName)
+	return r.updateStatus(r.sync(log, req))
+}
+
+func (r SnapshotCloneReconciler) prepare(syncRes *dataVolumeSyncResult) error {
```
arnongilboa, line 114: looks similar to `PvcCloneReconciler.prepare()` so why not move to `CloneReconcilerBase`?

pkg/controller/datavolume/controller-base.go
akalenyu: nice catch, nope!
we must GET datasource in here now to distinct between pvc/snap?

anyway sourceref support will be in a follow up so I'll add a fixme for now

pkg/controller/datavolume/smart-clone-controller.go
akalenyu: we used to mutate the pvcSpec through the ptr (accidentally)
https://github.com/kubevirt/containerized-data-importer/blob/bfe30a89648bdf37e8fd2eafcb65a04e9885cc20/pkg/controller/datavolume/smart-clone-controller.go#L392-L400

pkg/controller/datavolume/snapshot-clone-controller.go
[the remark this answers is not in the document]
arnongilboa: it's already handled in `ReconcilerBase.sync()/handlePrePopulation()`, so it might be done there

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,488 @@
...
+	transferName := getTransferName(datavolume)
+
+	pvcPopulated := pvcIsPopulated(pvc, datavolume)
+	_, prePopulated := datavolume.Annotations[cc.AnnPrePopulated]
+
+	if pvcPopulated || prePopulated {
+		return syncRes, nil
+	}
+
+	nn := types.NamespacedName{Namespace: datavolume.Spec.Source.Snapshot.Namespace, Name: datavolume.Spec.Source.Snapshot.Name}
```
arnongilboa, line 172: another reason `GetCloneSourceNameAndNamespace()` should return NamespacedName

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,488 @@
...
+	snapshot := &snapshotv1.VolumeSnapshot{}
+	if err := r.client.Get(context.TODO(), nn, snapshot); err != nil {
+		return syncRes, err
+	}
+	if snapshot.Status == nil || snapshot.Status.ReadyToUse == nil || !*snapshot.Status.ReadyToUse {
```
arnongilboa, line 177: I think `== false` is cleaner than `!*`

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,488 @@
...
+			return syncRes, err
+		}
+		pvc = targetHostAssistedPvc
+	}
+
+	if fallBackToHostAssisted {
+		if err := r.ensureExtendedToken(pvc); err != nil {
+			return syncRes, err
+		}
+	} else {
```
arnongilboa, line 217: drop this else, and `return syncRes, syncErr` just above it

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: https://staticcheck.io/docs/checks#S1002 yells that's why I went there

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: yeah they happen to be the same but prefer to keep a specific one for each controller
not at peace with the claim that preparation will be the same for all clone controllers

pkg/controller/datavolume/snapshot-clone-controller.go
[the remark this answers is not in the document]
akalenyu: not sure i got that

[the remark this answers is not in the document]
akalenyu: should we just drop these?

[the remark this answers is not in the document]
arnongilboa: No, I think the check can be moved to `ReconcilerBase.sync()/handlePrePopulation()`

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: done, thx

pkg/controller/common/util.go
akalenyu: so I am struggling to find these plural names in API libs and feels a bit wrong to define them ourselves,
any idea what I am missing?
We use `"persistentvolumeclaims"` quite a bit outside this file as well, so I am assuming someone ran into this before me

pkg/controller/datavolume/clone-controller-base.go
akalenyu: done

pkg/controller/datavolume/clone-controller-base.go
akalenyu: I only call `GetCloneSourceNameAndNamespace` when we don't know for sure,
which is not the case in `doCrossNamespaceClone` which gets called from a pvc/snap clone controller

pkg/controller/datavolume/controller-base.go
akalenyu: added fixme

pkg/controller/common/util.go
akalenyu: I prefer to keep this one if that's fine

pkg/controller/datavolume/snapshot-clone-controller.go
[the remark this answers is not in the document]
mhenriks: yeah I guess we don't have to wait but should probably make sure that the snapshot does not have the error set

pkg/controller/datavolume/snapshot-clone-controller.go
mhenriks: May work but I'm sure is racy.  Unique source pvcs is the way to go for now

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,524 @@
...
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	return reconcile.Result{}, r.syncCloneStatusPhase(syncRes, cdiv1.CloneFromSnapshotSourceInProgress, nil)
+}
+
+func (r *SnapshotCloneReconciler) createTempHostAssistedSourcePvc(dv *cdiv1.DataVolume, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec, syncRes *dataVolumeCloneSyncResult) error {
+	tempPvcName := getTempHostAssistedSourcePvcName(dv.Name)
```
mhenriks, line 352: `dv.Name` is not unique enough.  What if `ns1/fred` and `ns2/fred` are being cloned concurrently?

akalenyu: yup, thx, handled (dv.UID)

pkg/controller/datavolume/snapshot-clone-controller.go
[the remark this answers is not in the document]
akalenyu: do you mean `snapshot.status.error`? what if some csi driver is notoriously known for setting new errors for a long time, until snapshot is ready?
we would be back to the same problem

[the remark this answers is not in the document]
mhenriks: yeah that's the field.  I think it will be very unlikely for a snapshot to be in error state then be good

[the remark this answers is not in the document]
akalenyu: done

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -0,0 +1,527 @@
...
+	if pvcPopulated || prePopulated {
+		return syncRes, nil
+	}
+
+	nn := types.NamespacedName{Namespace: datavolume.Spec.Source.Snapshot.Namespace, Name: datavolume.Spec.Source.Snapshot.Name}
+	snapshot := &snapshotv1.VolumeSnapshot{}
+	if err := r.client.Get(context.TODO(), nn, snapshot); err != nil {
+		return syncRes, err
+	}
+	if snapshot.Status == nil || snapshot.Status.ReadyToUse == nil || !*snapshot.Status.ReadyToUse {
```
mhenriks, line 179: I think in the unlikely case that `snapshot.Status == nil` we should return error and exit.  At that point we do not know if the snapshot is in error state

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -23,6 +23,7 @@ import (
 	"context"
 	"encoding/json"
```
mhenriks, line 25: No changes required in `datavolume-validate.go`?  There is a bunch of stuff in there for PVC source I believe?  The mutate webhook should be just about creating the token.  The validate webhook should do the bulk of the validation.

akalenyu: So I was considering this but looking at PVC validation, it now allows the PVC to not exist (https://github.com/kubevirt/containerized-data-importer/pull/2306) altogether, which made me unsure if I want to go there.

Anyway, I plan to be making a couple of changes to validation in the sourceRef follow up
(https://github.com/akalenyu/containerized-data-importer/commits/snapshot-sourceref)
We can keep discussing this issue there if that is okay

This PR is huge as it is and I have been stepping outside of it quite a bit 
to ensure it can progress towards making it in (https://github.com/kubevirt/containerized-data-importer/pull/2559, https://github.com/kubevirt/containerized-data-importer/pull/2544),
so I would prefer if we could not block over this

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: yep, agree, done

pkg/apiserver/webhooks/datavolume-mutate.go
mhenriks: I think we should support snapshot not existing just like pvc.  Also, I get that sourceRef is orthogonal (as long as we fail/don't do anything now).  As well as sourceVolumeMode support.  But I think we should lock down `source.Snapshot` here.  Including all validation and support of snappshot not existing.

akalenyu: WDYT about some basic validation like https://github.com/akalenyu/containerized-data-importer/commit/74c4d290d214f2e855c04011d7c121853b2803b0#diff-2082e0e19255fc87c811d8b4347d274b06376449f517cca2e34ebbbdcfaf6cee, would that work?

I just feel like anything more than that is an enhancement; correct me if I'm wrong,
validation is nice as it will reject stuff early on but eventual consistency will get us to where we want to be in the end

pkg/operator/resources/cluster/controller.go
```diff
@@ -60,7 +60,6 @@ func getControllerClusterPolicyRules() []rbacv1.PolicyRule {
 			Resources: []string{
 				"persistentvolumes",
 				"persistentvolumeclaims",
-				"volumesnapshots",
```
mhenriks, line 63: Doesn't cdi-deployment need this?

pkg/controller/common/util.go
```diff
@@ -701,6 +702,42 @@ func ValidateClone(sourcePVC *v1.PersistentVolumeClaim, spec *cdiv1.DataVolumeSp
 	return nil
 }
 
+// ValidateSnapshotClone compares a snapshot clone spec against its source snapshot to validate its creation
+func ValidateSnapshotClone(sourceSnapshot *snapshotv1.VolumeSnapshot, spec *cdiv1.DataVolumeSpec) error {
```
mhenriks, line 706: Curious why you put this here if it's only used in the webhook?

tests/cloner_test.go
```diff
@@ -2537,6 +2526,197 @@ var _ = Describe("all clone tests", func() {
 			Expect(annValue).To(Equal("true"))
 		})
 	})
+
+	var _ = Describe("Clone from volumesnapshot source", func() {
```
mhenriks, line 2530: How about some functional test for validation webhook?  clone to smaller pvc, etc?

Also test source initially not existing?:

mhenriks (commented):
Getting there!

pkg/operator/resources/cluster/controller.go
akalenyu: its listed under ```APIGroups: []string{
				"",
			},```

pkg/controller/common/util.go
akalenyu: I had to use it in the controller itself and missed that part

tests/cloner_test.go
akalenyu: Right, added

pkg/operator/resources/cluster/controller.go
mhenriks: ah okay 👍

mhenriks: I see there is another rbac for "*" in "snapshot.storage.k8s.io"

mhenriks:
Great work @akalenyu!
/lgtm
/approve

PR #2525, opened 2023-01-02
Align DV controllers with kubevirt sync/updateStatus model

arnongilboa:

Follow the model of controllers in kubevirt where `sync()` does the "work" of the controller (creating other resources, etc). The `sync()` for import/upload is just creating the pvc resource. For clones, it is more complicated depending on clone type and if cross-namespace or not. After `sync()`, `updateStatus()` is called, even if `sync()` returns an error. `updateStatus()` is conscious of all steps performed in `sync()` and updates the status of the reconciled resource based on what can be observed. The only k8s resource it creates/updates is the `DataVolume`. This split is especially nice for resources with status subresource enabled, which should be one of the ultimate goals of DV controllers refactor work.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume/clone-controller.go
+++ b/pkg/controller/datavolume/clone-controller.go
@@ -334,8 +334,8 @@ func (r CloneReconciler) Reconcile(ctx context.Context, req reconcile.Request) (
 	return r.updateStatus(r.sync(log, req))
 }
 
-func (r CloneReconciler) prepare(syncRes *dataVolumeCloneSyncResult) error {
-	dv := syncRes.dvCopy
+func (r CloneReconciler) prepare(syncRes *dataVolumeSyncResult) error {
+	dv := syncRes.dvMutated
 	if err := r.populateSourceIfSourceRef(dv); err != nil {
 		return err
 	}
@@ -366,29 +366,22 @@ func (r CloneReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, pvc *co
 
 func (r CloneReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeCloneSyncResult, error) {
 	syncRes, syncErr := r.syncClone(log, req)
-	if err := r.syncUpdateMeta(log, syncRes.dataVolumeSyncResult); err != nil {
+	if err := r.syncUpdateMeta(log, &syncRes.dataVolumeSyncResult); err != nil {
 		syncErr = err
 	}
 	return syncRes, syncErr
 }
 
 func (r CloneReconciler) syncClone(log logr.Logger, req reconcile.Request) (dataVolumeCloneSyncResult, error) {
-	var syncRes dataVolumeCloneSyncResult
-
-	cleanup := func() error {
-		return r.cleanup(&syncRes)
-	}
-	prepare := func() error {
-		return r.prepare(&syncRes)
-	}
-	syncErr := r.syncCommon(log, req, &syncRes.dataVolumeSyncResult, cleanup, prepare)
+	res, syncErr := r.syncCommon(log, req, r.cleanup, r.prepare)
+	syncRes := dataVolumeCloneSyncResult{dataVolumeSyncResult: *res}
 	if syncErr != nil || syncRes.result != nil {
 		return syncRes, syncErr
 	}
 
 	pvc := syncRes.pvc
 	pvcSpec := syncRes.pvcSpec
-	datavolume := syncRes.dvCopy
+	datavolume := syncRes.dvMutated
 	transferName := getTransferName(datavolume)
 
 	// Get the most appropiate clone strategy
@@ -539,7 +532,7 @@ func (r CloneReconciler) syncClone(log logr.Logger, req reconcile.Request) (data
 
 func (r CloneReconciler) updateStatus(syncRes dataVolumeCloneSyncResult, syncErr error) (reconcile.Result, error) {
 	if ps := syncRes.phaseSync; ps != nil {
-		if err := r.updateDataVolumeStatusPhaseWithEvent(ps.phase, syncRes.dv, syncRes.dvCopy, ps.pvc, ps.event); err != nil {
+		if err := r.updateDataVolumeStatusPhaseWithEvent(ps.phase, syncRes.dv, syncRes.dvMutated, ps.pvc, ps.event); err != nil {
 			syncErr = err
 		}
 		return getReconcileResult(syncRes.result), syncErr
@@ -673,7 +666,7 @@ func (r *CloneReconciler) reconcileCsiClonePvc(log logr.Logger,
 	transferName string) (reconcile.Result, error) {
 
 	log = log.WithName("reconcileCsiClonePvc")
-	datavolume := syncRes.dvCopy
+	datavolume := syncRes.dvMutated
 	pvcSpec := syncRes.pvcSpec
 	pvcName := datavolume.Name
 
@@ -802,7 +795,7 @@ func (r *CloneReconciler) expandPvcAfterClone(
 	if isTemp {
 		// trigger transfer and next reconcile should have pvcExists == true
 		pvc.Annotations[annReadyForTransfer] = "true"
-		pvc.Annotations[cc.AnnPopulatedFor] = syncRes.dvCopy.Name
+		pvc.Annotations[cc.AnnPopulatedFor] = syncRes.dvMutated.Name
 		if err := r.updatePVC(pvc); err != nil {
 			return reconcile.Result{}, err
 		}
@@ -828,7 +821,7 @@ func (r *CloneReconciler) reconcileSmartClonePvc(log logr.Logger,
 	transferName string,
 	snapshotClassName string) (reconcile.Result, error) {
 
-	datavolume := syncRes.dvCopy
+	datavolume := syncRes.dvMutated
 	pvcName := datavolume.Name
 
 	if isCrossNamespaceClone(datavolume) {
@@ -930,7 +923,7 @@ func (r *CloneReconciler) doCrossNamespaceClone(log logr.Logger,
 	returnWhenCloneInProgress bool,
 	selectedCloneStrategy cloneStrategy) (*reconcile.Result, error) {
 
-	datavolume := syncRes.dvCopy
+	datavolume := syncRes.dvMutated
 	initialized, err := r.initTransfer(log, syncRes, pvcName)
 	if err != nil {
 		return &reconcile.Result{}, err
@@ -987,7 +980,7 @@ func (r *CloneReconciler) sourceInUse(dv *cdiv1.DataVolume, eventReason string) 
 
 func (r *CloneReconciler) initTransfer(log logr.Logger, syncRes *dataVolumeCloneSyncResult, name string) (bool, error) {
 	initialized := true
-	dv := syncRes.dvCopy
+	dv := syncRes.dvMutated
 
 	log.Info("Initializing transfer")
 
@@ -1045,8 +1038,8 @@ func (r *CloneReconciler) initTransfer(log logr.Logger, syncRes *dataVolumeClone
 	return initialized, nil
 }
 
-func (r CloneReconciler) cleanup(syncRes *dataVolumeCloneSyncResult) error {
-	dv := syncRes.dvCopy
+func (r CloneReconciler) cleanup(syncRes *dataVolumeSyncResult) error {
+	dv := syncRes.dvMutated
 	transferName := getTransferName(dv)
 	if !cc.HasFinalizer(dv, crossNamespaceFinalizer) {
 		return nil
@@ -1178,7 +1171,7 @@ func (r *CloneReconciler) expand(log logr.Logger, syncRes *dataVolumeCloneSyncRe
 
 	if !podExists {
 		var err error
-		pod, err = r.createExpansionPod(pvc, syncRes.dvCopy, podName)
+		pod, err = r.createExpansionPod(pvc, syncRes.dvMutated, podName)
 		// Check if pod has failed and, in that case, record an event with the error
 		if podErr := cc.HandleFailedPod(err, podName, pvc, r.recorder, r.client); podErr != nil {
 			return false, podErr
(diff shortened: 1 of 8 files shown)


pkg/controller/datavolume/upload-controller.go
```diff
@@ -93,50 +92,47 @@ func NewUploadController(
...
+			r.updateDataVolumeStatusPhaseWithEvent(cdiv1.Pending, syncRes.dv, nil, nil,
 				Event{
 					eventType: corev1.EventTypeWarning,
 					reason:    cc.ErrExceededQuota,
-					message:   err.Error(),
+					message:   syncErr.Error(),
 				})
 		}
-		return err
+		return reconcile.Result{}, syncErr
```
mhenriks, line 130: Why is updateStatusCommon not called?  The main idea of this pattern is that even if there is an error during sync, the status is checked/updated if anything changes.  Do you know for sure that the status has not changed from whatever it was last?

pkg/controller/datavolume/import-controller.go
```diff
@@ -198,58 +204,56 @@ func (r ImportReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, annota
...
+			r.updateDataVolumeStatusPhaseWithEvent(cdiv1.Pending, syncRes.dv, nil, nil,
+				Event{
+					eventType: corev1.EventTypeWarning,
+					reason:    cc.ErrExceededQuota,
+					message:   syncErr.Error(),
+				})
 		}
-		*pvc = newPvc
-		return nil
+		return reconcile.Result{}, syncErr
```
mhenriks, line 241: why return without attempting to check/update status?

pkg/controller/datavolume/clone-controller.go
```diff
@@ -503,11 +524,32 @@ func (r *CloneReconciler) reconcileClone(log logr.Logger,
...
+}
+
+func (r CloneReconciler) updateStatus(syncRes dataVolumeCloneSyncResult, syncErr error) (reconcile.Result, error) {
+	modifier := getModifier(syncRes)
+	ps := syncRes.phaseSync
+	if ps != nil {
+		syncErr = r.updateDataVolumeStatusPhaseWithEvent(ps.phase, syncRes.dv, ps.pvc, modifier, ps.event)
+	}
+	if syncErr != nil {
+		return reconcile.Result{}, syncErr
```
mhenriks, line 544: Why return without attempting to check/update status?

mhenriks (commented):
The idea of the sync()/updateStatus() pattern is that we want to make sure to check/update status when sync() (the real work) returns an error and it does not appear that we are doing that in a bunch of places (I did not check everywhere)

pkg/controller/datavolume/upload-controller.go
arnongilboa: fixed

pkg/controller/datavolume/import-controller.go
arnongilboa: fixed

pkg/controller/datavolume/clone-controller.go
arnongilboa: fixed

arnongilboa:
> The idea of the sync()/updateStatus() pattern is that we want to make sure to check/update status when sync() (the real work) returns an error and it does not appear that we are doing that in a bunch of places (I did not check everywhere)

Right. Somehow missed it during cleanup.

mhenriks:
This test seems to fail consistently:  `Tests Suite: DataVolume Garbage Collection Should [test_id:8564] garbage collect dv after completion when TTL is 10s`

pkg/controller/datavolume/import-controller.go
```diff
@@ -198,58 +204,60 @@ func (r ImportReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, annota
...
-		if err != nil {
-			return err
-		}
-		if changed {
-			err = r.client.Get(context.TODO(), types.NamespacedName{Name: dv.Name, Namespace: dv.Namespace}, dv)
-			if err != nil {
-				return err
+	modifier := func(dv *cdiv1.DataVolume) {
+		if syncRes.pvc != nil && cc.GetSource(syncRes.pvc) == cc.SourceVDDK {
+			if vddkHost := syncRes.pvc.Annotations[cc.AnnVddkHostConnection]; vddkHost != "" {
```
mhenriks, line 248: One of the goals of this controller refactoring work should be to enable the status subresource on DataVolumes.   Not sure if you've thought much about that.  But it may be good to make sure that `updateStatus()` only updates the resource status.  In which case, updates like this (adding annotations) should be moved to `sync()`

arnongilboa: fixed

arnongilboa:
> This test seems to fail consistently: `Tests Suite: DataVolume Garbage Collection Should [test_id:8564] garbage collect dv after completion when TTL is 10s`

fixed

pkg/controller/datavolume/import-controller.go
mhenriks: Confirming your fix for now is to have the annotations added during `sync()` but still relying on `updateStatus()` to call update?

arnongilboa: Right, I guess we don't want to call `client.Update()` both on `sync()` and `updateStatus()`. You may also have both annotations and status updates on the same CR, so a single `client.Update()` will be called.

mhenriks: I understand that fewer api calls makes sense for efficiency.  But I think it may be nice to stick to the theme of this pr and update the resources in sync now.  Then in `emitEvent()` you can change:

`if !reflect.DeepEqual(dataVolume, dataVolumeCopy) {`

to 

`if !reflect.DeepEqual(dataVolume.Status, dataVolumeCopy.Status) {`


also, I see that `updateStatusCommon()` may annotate the DataVolume like here:  https://github.com/arnongilboa/containerized-data-importer/blob/sync_update_status/pkg/controller/datavolume/controller.go#L464


Basically, I'm thinking that it may make sense to go over the sync/updateStatus code paths now.  For `sync() `call `update()` if the resource spec/metadata changes.  Make sure `sync()` does not update any status values.  for `updateStatus()`, make sure no spec/metadata changes occur.

Then in a follow up pr, you can enable the status subresource for the DataVolume.  Change `updateStatus()` to to call `Status().Update()` and see if anything breaks.

What do you think?

arnongilboa: Sounds great. Fixing.

pkg/controller/datavolume/controller.go
```diff
@@ -236,63 +231,81 @@ func getDataVolumeOp(dv *cdiv1.DataVolume) dataVolumeOp {
...
-func (r ReconcilerBase) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	log := r.log.WithValues("DataVolume", req.NamespacedName)
-	res, syncErr := r.sync(log, req, nil, nil)
-	return r.updateStatus(log, res, syncErr)
-}
-
-func (r ReconcilerBase) sync(log logr.Logger, req reconcile.Request, cleanup, prepare func(*cdiv1.DataVolume) error) (dataVolumeSyncResult, error) {
-	var ctx dataVolumeSyncResult
-
+func (r ReconcilerBase) syncCommon(log logr.Logger, req reconcile.Request, syncRes *dataVolumeSyncResult, cleanup, prepare func() error) error {
```
mhenriks, line 234: I think it would be better to have this return `(*dataVolumeSyncResult, error).  The function will create the object, pass it to the callback functions and return it

pkg/controller/datavolume/controller.go
```diff
@@ -87,20 +87,15 @@ type Event struct {
 
 type dataVolumeSyncResult struct {
 	dv      *cdiv1.DataVolume
+	dvCopy  *cdiv1.DataVolume
```
mhenriks, line 90: maybe the naming can be improved here?  I assume one is the original and one is being mutated.  How to know which?

pkg/controller/datavolume/controller.go
```diff
@@ -236,63 +231,81 @@ func getDataVolumeOp(dv *cdiv1.DataVolume) dataVolumeOp {
...
+	}
+	return nil
+}
+
+func (r ReconcilerBase) annotate(dv *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) {
+	if pvc.Status.Phase == corev1.ClaimBound && pvcIsPopulated(pvc, dv) {
+		if dv.Annotations == nil {
+			dv.Annotations = make(map[string]string)
+		}
+		dv.Annotations[cc.AnnPrePopulated] = pvc.Name
```
mhenriks, line 307: I think it appears that each of the "derived" controllers may add their own annotations so it is okay that they each call `syncUpdateMeta` explicitly but that may not always be the case. And if so, this update would get lost.  Should syncCommon also call `syncUpdateMeta`?

mhenriks (commented):
Some things to consider

pkg/controller/datavolume/controller.go
arnongilboa: Like the naming convention all over CDI,  the original is obj and the mutated is objCopy.

pkg/controller/datavolume/controller.go
```diff
@@ -279,6 +280,15 @@ func (r ReconcilerBase) syncCommon(log logr.Logger, req reconcile.Request, syncR
 	return nil
 }
 
+func (r *ReconcilerBase) annotate(dv *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) {
```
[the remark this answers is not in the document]
arnongilboa: The idea was to keep the name generic for adding more annotations in the future. It's the single ReconcilerBase func for annotating DV, so calling it something clumsy like `addPrePopulatedAnnotationIfNeeded` seems tldr to me. Better idea?

pkg/controller/datavolume/controller.go
arnongilboa: Not sure I agree. The func will need to return the pair instead of a just error (x9 times) which is not that nice (currently the code assumes that dataVolumeSyncResult != nil), and the callbacks will also need a param which is currently not needed. Pleasee convince me why you do prefer it.

pkg/controller/datavolume/controller.go
arnongilboa: It won't get lost as the check in emitEvent() will always error on:
```
	if !reflect.DeepEqual(dataVolume.ObjectMeta, dataVolumeCopy.ObjectMeta) {
		return fmt.Errorf("meta update is not allowed in updateStatus phase")
	}
```

pkg/controller/datavolume/controller.go
mhenriks: To me:
```
var syncRes dataVolumeSyncResult
syncErr := r.syncCommon(log, req, &syncRes, nil, nil)
```

is awkward compared to:


```
syncRes, syncErr := r.syncCommon(log, req, nil, nil)
```

IMO it only makes sense to pass a parameter by reference if  1) the parameter is initialized with values or 2) it is being used in some sort of polymorphic way like `json.UnMarshal`.  Neither of those are true here

Also, not that it matters much, I see this function called only three times

pkg/controller/datavolume/controller.go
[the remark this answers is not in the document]
mhenriks: okay then add parameters for the annotation key/value.  I'm sure we already have a utility function that does what you describe though.  So make this name more specific or use the other function

pkg/controller/datavolume/controller.go
mhenriks: Cpy may be fine when the variable exists in the scope of a function but not in a struct that is passed around without context

pkg/controller/datavolume/controller.go
mhenriks: This error would not be easy to debug either, as it is returned by `updateStatus` but the metadata is mutated in `sync`.  The peril of having shared state

pkg/controller/datavolume/controller.go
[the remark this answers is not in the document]
arnongilboa: Done.

pkg/controller/datavolume/controller.go
arnongilboa: Renamed to `dvMutated`.

pkg/controller/datavolume/controller.go
arnongilboa: Adding `syncUpdateMeta` call in `syncCommon`.

pkg/controller/datavolume/controller.go
```diff
@@ -236,63 +231,89 @@ func getDataVolumeOp(dv *cdiv1.DataVolume) dataVolumeOp {
...
+		return fmt.Errorf("status update is not allowed in sync phase")
+	}
+	if !reflect.DeepEqual(syncRes.dv.ObjectMeta, syncRes.dvMutated.ObjectMeta) {
+		if err := r.updateDataVolume(syncRes.dvMutated); err != nil {
+			r.log.Error(err, "Unable to sync update dv meta", "name", syncRes.dvMutated.Name)
+			return err
+		}
+		// Needed for emitEvent() DeepEqual check
+		objMeta := syncRes.dvMutated.ObjectMeta.DeepCopy()
+		syncRes.dv.ObjectMeta = *objMeta
```
mhenriks, line 308: I think the right thing to do here is actually:
```
syncRes.dv = syncRes.dvMutated.DeepCopy()
```

mhenriks: Problem is what if the spec and metadata is updated?  You have to either specifically disallow that or change the name of this to `syncUpdate`.  Spec updates should not be out of bounds for the controller

arnongilboa: fixed

mhenriks: What do you think about calling this `syncUpdate()` since it's not just about meta

mhenriks:
/approve

@alromeros @akalenyu care to take a look?  lgtm if you like

alromeros (approved):
I like what I see, good job! Since I'll be working on top of these changes I think we can merge this and, if needed, add minor fixes and improvements on our next PRs 
/lgtm

pkg/controller/datavolume/controller.go
arnongilboa: @alromeros, ^^^

PR #2527, opened 2023-01-03
Remove v1alpha1 version from CDI CRD

maya-r:
This is the last CRD of CDI that still has v1alpha1 version.
We left it because it appears in release YAMLs, and we want update through applying YAMLs to just work.

Technical debt left over from #2407.
We needed more complicated work to watch the CDI CRD.
This CRD is created by the release YAMLs and not owned by the controller, therefore we need custom watch code for it that isn't from controller-lifecycle-operator-sdk.

I've checked early in reconcile whether this is a CDI CRD and not a CDI CR - not sure how to integrate it with controller-lifecycle-operator-sdk. Not sure whether needing Update() is normal.
Perhaps it's not too bad as we already know we'll remove this code in under a year (once we've released a version or more with this PR).

**Release note**:

```release-note
At runtime, remove v1alpha1 version from CDI CRD, the last CRD that had it.
The release YAMLs still contain the alpha version, but it will be removed at runtime.
We plan to remove the version from the release YAML in the future (after a few versions), but as long as you update through intermediate versions, it will just work without errors.
```

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -192,9 +192,9 @@ func (r *ReconcileCDI) SetController(controller controller.Controller) {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileCDI) Reconcile(_ context.Context, request reconcile.Request) (reconcile.Result, error) {
-	operatorVersion := r.namespacedArgs.OperatorVersion
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.Info("Reconciling CDI")
+	reqLogger.Info("Reconciling CDI CR")
+	operatorVersion := r.namespacedArgs.OperatorVersion
 	cr := &cdiv1.CDI{}
 	crKey := client.ObjectKey{Namespace: "", Name: request.NamespacedName.Name}
 	err := r.client.Get(context.TODO(), crKey, cr)
--- a/pkg/operator/controller/cruft.go
+++ b/pkg/operator/controller/cruft.go
@@ -306,7 +306,7 @@ func reconcileCDICRD(args *callbacks.ReconcileCallbackArgs) error {
 		return nil
 	}
 	deployment := args.CurrentObject.(*appsv1.Deployment)
-	if !isControllerDeployment(deployment) || !sdk.CheckDeploymentReady(deployment) {
+	if !isControllerDeployment(deployment) {
 		return nil
 	}
 
--- a/pkg/operator/controller/cruft.go
+++ b/pkg/operator/controller/cruft.go
@@ -31,11 +31,16 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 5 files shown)


pkg/operator/controller/controller.go
```diff
@@ -192,9 +192,22 @@ func (r *ReconcileCDI) SetController(controller controller.Controller) {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileCDI) Reconcile(_ context.Context, request reconcile.Request) (reconcile.Result, error) {
-	operatorVersion := r.namespacedArgs.OperatorVersion
+	if request.NamespacedName.Name == "cdis.cdi.kubevirt.io" {
+		return r.ReconcileCDICRD(request)
```
mhenriks, line 196: I think it will be better to have a singe Reconcile function.  The watch on CRDs can use the `GetActiveCDI()` function to queue CDI CR.  The work of `ReconcileCDICRD` can then be moved to a callback function

maya-r: I'm having trouble getting this code to work. the operator lifecycle sdk only triggers the callbacks for resources from GetAllResources, but if I add CDI CRD to the list, cdi-operator loops and calls CDI CRD and orphan object.

[Code here](https://github.com/maya-r/containerized-data-importer/tree/v1alpha1-remove-cdi-crd2).

mhenriks: I meant that you call `GetActiveCDI` here:  https://github.com/maya-r/containerized-data-importer/blob/9123579359b3c241d1c06d1b04a61d29b48ea3a2/pkg/operator/controller/cruft.go#L280-L287

That will trigger regular reconcile.  Then add a callback on any resource like cdi-deployment to do the work of checking/updating the CDI CRD

maya-r: I've written the code this way, but I can't use a callback like for the other CRDs. if I add it as a resource that operator-lifecycle-sdk monitors, it errors and fails to deploy CDI complaining about the "orphan" resource of CDI CRD (that isn't owned by the CDI deployment). It looks better, I hope that's good enough.

mhenriks: Not sure what you were trying to do but operator-lifecycle-sdk can be totally oblivious.  Take a look at the `reconcileCreateRoute` callback.  It is a callback on deployments that is executed when cdi-deployment is passed.  But that is really just a mechanism to make sure that the callback only executes once per reconcile loop.  The callback has nothing to do with that cdi-deployment.  It manages a Route that operator-lifecycle-sdk does not care about.  We can do something similar for the CDI CR.

maya-r: Done, thanks for the tip

pkg/operator/controller/cruft.go
```diff
@@ -263,3 +268,93 @@ func reconcileRemainingRelationshipLabels(args *callbacks.ReconcileCallbackArgs)
...
+
+// ReconcileCDICRD monitors the CDI CRD and removes the alpha version from it.
+// Delete after we no longer want to include CDI CRD v1alpha1 version in release YAMLs
+// Remove alpha as a version from the CDI CRD
+func reconcileCDICRD(args *callbacks.ReconcileCallbackArgs) error {
+	if args.State != callbacks.ReconcileStatePostRead {
+		return nil
+	}
+	deployment := args.CurrentObject.(*appsv1.Deployment)
+	if !isControllerDeployment(deployment) || !sdk.CheckDeploymentReady(deployment) {
```
mhenriks, line 309: not a big deal, but don't think you have to wait for the deployment to be ready

maya-r: Done

PR #2536, opened 2023-01-09
mount block pvc readonly

mhenriks:

#2410 had an unintended consequence of mounting block PVCs read/write when they were a clone source

This PR undoes that

Fixes #

**Release note**:

```release-note
Mount block PVCs readonly when source of clone operation
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -624,6 +624,7 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 					VolumeSource: corev1.VolumeSource{
 						PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
 							ClaimName: sourcePvcName,
+							ReadOnly:  sourceVolumeMode == corev1.PersistentVolumeBlock,
 						},
 					},
 				},
--- a/pkg/controller/clone-controller_test.go
+++ b/pkg/controller/clone-controller_test.go
@@ -177,7 +177,7 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		}),
 	)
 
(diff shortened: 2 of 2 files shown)


pkg/controller/clone-controller.go
```diff
@@ -624,6 +624,7 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 					VolumeSource: corev1.VolumeSource{
 						PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
 							ClaimName: sourcePvcName,
+							ReadOnly:  sourceVolumeMode == corev1.PersistentVolumeBlock,
```
akalenyu, line 627: isn't it a little confusing to read? the impression is that we set readOnly: true only for block, but that's not the case
previously we had it as the following:
https://github.com/kubevirt/containerized-data-importer/blob/4d4ad12df5673c847f2bf232607f395faf8b8712/pkg/controller/clone-controller.go#L673
and https://github.com/kubevirt/containerized-data-importer/blob/4d4ad12df5673c847f2bf232607f395faf8b8712/pkg/controller/clone-controller.go#L732-L736

akalenyu (commented):
awesome

pkg/controller/clone-controller.go
akalenyu: So I am reading the `pod.spec.volumes.persistentVolumeClaim` API wrong?
I thought it is just about enforcing the lower level readOnly fields on the container.
```readOnly Will force the ReadOnly setting in VolumeMounts. Default false.```

I mean we should go with https://github.com/kubevirt/containerized-data-importer/blob/4d4ad12df5673c847f2bf232607f395faf8b8712/pkg/controller/clone-controller.go#L670-L673
(As we have previously, no container level setting at all)

akalenyu:
/lgtm
thanks for offline explanation!

PR #2546, opened 2023-01-18
Ensure Prometheus resources exist for CDINotReady

arnongilboa:

Ensure Prometheus resources exist also when cdi-deployment is not ready. This is needed because currently when cdi deployment fails (e.g. wrong NodeSelector) the Prometheus resources are not created, so CDINotReady will not be fired although CDI is not ready.

Fixes bz 2145146

**Release note**:
```release-note
Ensure Prometheus resources exist for CDINotReady
```

### DIFF
--- a/pkg/operator/controller/callbacks.go
+++ b/pkg/operator/controller/callbacks.go
@@ -171,6 +171,7 @@ func reconcileCreatePrometheusInfra(args *callbacks.ReconcileCallbackArgs) error
 		args.Recorder.Event(cr, corev1.EventTypeWarning, createResourceFailed, fmt.Sprintf("Failed to ensure prometheus resources exists, %v", err))
 		return err
 	}
+
 	return nil
 }
 
--- a/pkg/operator/controller/callbacks.go
+++ b/pkg/operator/controller/callbacks.go
@@ -151,7 +151,8 @@ func reconcileCreatePrometheusInfra(args *callbacks.ReconcileCallbackArgs) error
 	}
 
 	deployment := args.CurrentObject.(*appsv1.Deployment)
-	if !isControllerDeployment(deployment) || !sdk.CheckDeploymentReady(deployment) {
+	// we don't check sdk.CheckDeploymentReady(deployment) since we want Prometheus to cover NotReady state as well
+	if !isControllerDeployment(deployment) {
 		return nil
 	}
 
@@ -170,7 +171,6 @@ func reconcileCreatePrometheusInfra(args *callbacks.ReconcileCallbackArgs) error
 		args.Recorder.Event(cr, corev1.EventTypeWarning, createResourceFailed, fmt.Sprintf("Failed to ensure prometheus resources exists, %v", err))
 		return err
 	}
-
 	return nil
 }
 
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -921,7 +921,7 @@ var _ = Describe("ALL Operator tests", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 
-			It("[test_id:9656] CDINotReady alert firing when CDI is not ready", func() {
+			It("[test_id:9656] Metric kubevirt_cdi_cr_ready is 0 when CDI is not ready", func() {
 				if !f.IsPrometheusAvailable() {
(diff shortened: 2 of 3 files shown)


pkg/operator/controller/callbacks.go
```diff
@@ -170,7 +171,6 @@ func reconcileCreatePrometheusInfra(args *callbacks.ReconcileCallbackArgs) error
 		args.Recorder.Event(cr, corev1.EventTypeWarning, createResourceFailed, fmt.Sprintf("Failed to ensure prometheus resources exists, %v", err))
 		return err
 	}
-
```
akalenyu, line 173: intended? afaik we mostly keep the empty line before returns

tests/operator_test.go
```diff
@@ -963,19 +887,85 @@ var _ = Describe("ALL Operator tests", func() {
...
+						Name: "cdi",
+					},
+					Spec: cr.Spec,
+				}
+				cdi.Spec.Infra.NodeSelector = map[string]string{"wrong": "wrong"}
+				cdi, err := f.CdiClient.CdiV1beta1().CDIs().Create(context.TODO(), cdi, metav1.CreateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+
+				promRule := getPrometheusRule()
+				patchPrometheusRule(promRule, "CDINotReady", "1m")
```
akalenyu, line 948: does this work? isn't the operator overriding `1m`?

akalenyu (commented):
awesome!

pkg/operator/controller/callbacks.go
arnongilboa: ok :)

tests/operator_test.go
arnongilboa: Thanks for reminding me this one I somehow forgot. How can we prevent it? scale down thr operator?

arnongilboa: Following your suggestion, the updated test checks the metric instead of the alert

tests/operator_test.go
```diff
@@ -1373,6 +1347,112 @@ var _ = Describe("ALL Operator tests", func() {
 	})
 })
 
+func getCDICrAndPods(f *framework.Framework) (*cdiv1.CDI, *corev1.PodList) {
+	cdiPods, err := f.K8sClient.CoreV1().Pods(f.CdiInstallNs).List(context.TODO(), metav1.ListOptions{})
+
+	Expect(err).ToNot(HaveOccurred(), "failed listing cdi pods")
+	Expect(len(cdiPods.Items)).To(BeNumerically(">", 0), "no cdi pods found")
+
+	cr, err := f.CdiClient.CdiV1beta1().CDIs().Get(context.TODO(), "cdi", metav1.GetOptions{})
```
awels, line 1356: So I know this code was copied from above, but it is not correct I think, this test should still work if the CDI CR is not named `cdi` there is another function that lists the CDI CRs, and returns the CR if there is only one, regardless of what the name is. This test will be skipped when we deploy CDI with HCO since the CR name will be different, and I don't think that is intended.

tests/operator_test.go
```diff
@@ -1373,6 +1347,112 @@ var _ = Describe("ALL Operator tests", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			return false
+		}, 5*time.Minute, 2*time.Second).Should(BeTrue())
+	} else {
+		Expect(errors.IsNotFound(err)).To(BeTrue())
+	}
+
+	cdi = &cdiv1.CDI{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cdi",
```
awels, line 1403: Same here, the name of the CR is hard coded to `cdi`, we should take the name from the passed in CR

awels (approved):
/approve

tests/operator_test.go
akalenyu: nice catch, this effectively means we were skipping stuff on runs where cdi.Name != "cdi" :(

awels:
one more time but failed on same issue twice in a row now.

arnongilboa:
> /test pull-containerized-data-importer-e2e-destructive one more time but failed on same issue twice in a row now.

@awels see flakiness fix in #2557

PR #2547, opened 2023-01-19
Pass DV labels to PVC

### LINKED ISSUE #2533: DataVolume Labels are not Copied to the PVC
**What happened**:
Created a DataVolume with `generateName` and some labels and annotations. The PVC was created with the DataVolume annotations but without the DataVolume labels. It was then hard to find the PVC, as it's name is not known, and there is no way to use label selector to find it, as the labels do not exist.

The DataVolume labels should be copied to the PVC, as done for the annotations.

1. Create a DataVolume with labels.
2. Read the PVC.
3. Notice that the labels do not exist in the PVC

**Additional context**:

**Environment**:
- DV specification: N/A
- Install tools: N/A
- Others: N/A

ido106:

Pass all the DataVolume labels to its created PVC. It's fixing the current behavior which passes only specific labels, aligning it with DV annotations passing to the PVC.

Fixes #2533 

**Release note**:
```release-note
Pass all the DataVolume labels to its created PVC
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -314,6 +314,7 @@ type dvController interface {
 	updateStatusPhase(pvc *corev1.PersistentVolumeClaim, dataVolumeCopy *cdiv1.DataVolume, event *Event) error
 }
 
+<<<<<<< HEAD
 func (r *ReconcilerBase) reconcile(ctx context.Context, req reconcile.Request, dvc dvController) (reconcile.Result, error) {
 	log := r.log.WithValues("DataVolume", req.NamespacedName)
 	syncRes, syncErr := dvc.sync(log, req)
@@ -340,6 +341,11 @@ func (r *ReconcilerBase) syncCommon(log logr.Logger, req reconcile.Request, clea
 func (r *ReconcilerBase) syncDvPvcState(log logr.Logger, req reconcile.Request, cleanup, prepare dvSyncStateFunc) (dvSyncState, error) {
 	syncState := dvSyncState{}
 	dv, err := r.getDataVolume(req.NamespacedName)
+=======
+func (r ReconcilerBase) sync(log logr.Logger, req reconcile.Request, cleanup, prepare dataVolumeSyncResultFunc) (*dataVolumeSyncResult, error) {
+	syncRes := &dataVolumeSyncResult{}
+	dv, err := getDataVolume(r.client, req.NamespacedName)
+>>>>>>> 67bd1f823 (Rebase and add utests)
 	if dv == nil || err != nil {
 		syncState.result = &reconcile.Result{}
 		return syncState, err
@@ -551,9 +557,9 @@ func (r *ReconcilerBase) getPVC(key types.NamespacedName) (*corev1.PersistentVol
 	return pvc, nil
 }
 
-func (r *ReconcilerBase) getDataVolume(key types.NamespacedName) (*cdiv1.DataVolume, error) {
+func getDataVolume(c client.Client, key types.NamespacedName) (*cdiv1.DataVolume, error) {
 	dv := &cdiv1.DataVolume{}
-	if err := r.client.Get(context.TODO(), key, dv); err != nil {
+	if err := c.Get(context.TODO(), key, dv); err != nil {
 		if k8serrors.IsNotFound(err) {
 			return nil, nil
 		}
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -960,21 +960,6 @@ func buildHTTPClient() *http.Client {
 	return httpClient
 }
 
-func passDataVolumeInstancetypeLabelstoPVC(dataVolumeLabels, pvcLabels map[string]string) map[string]string {
-	instancetypeLabels := []string{
-		cc.LabelDefaultInstancetype,
-		cc.LabelDefaultInstancetypeKind,
-		cc.LabelDefaultPreference,
-		cc.LabelDefaultPreferenceKind,
-	}
-	for _, label := range instancetypeLabels {
-		if dvLabel, hasLabel := dataVolumeLabels[label]; hasLabel {
-			pvcLabels[label] = dvLabel
-		}
-	}
-	return pvcLabels
-}
-
 // newPersistentVolumeClaim creates a new PVC for the DataVolume resource.
 // It also sets the appropriate OwnerReferences on the resource
 // which allows handleObject to discover the DataVolume resource
@@ -986,7 +971,9 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume, 
 	if util.ResolveVolumeMode(targetPvcSpec.VolumeMode) == corev1.PersistentVolumeFilesystem {
 		labels[common.KubePersistentVolumeFillingUpSuppressLabelKey] = common.KubePersistentVolumeFillingUpSuppressLabelValue
 	}
-	labels = passDataVolumeInstancetypeLabelstoPVC(dataVolume.GetLabels(), labels)
+	for k, v := range dataVolume.Labels {
+		labels[k] = v
+	}
 
 	annotations := make(map[string]string)
 	for k, v := range dataVolume.ObjectMeta.Annotations {
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -314,7 +314,6 @@ type dvController interface {
 	updateStatusPhase(pvc *corev1.PersistentVolumeClaim, dataVolumeCopy *cdiv1.DataVolume, event *Event) error
 }
 
-<<<<<<< HEAD
 func (r *ReconcilerBase) reconcile(ctx context.Context, req reconcile.Request, dvc dvController) (reconcile.Result, error) {
 	log := r.log.WithValues("DataVolume", req.NamespacedName)
 	syncRes, syncErr := dvc.sync(log, req)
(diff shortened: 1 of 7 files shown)


awels:
So you essentially want to pass all labels from the DV to the PVC? Any particular reason you want that? We are already passing labels from the DV to the import/upload/clone pods for things like istio and I don't think we want to pass those to the PVC.

akalenyu:
> So you essentially want to pass all labels from the DV to the PVC? Any particular reason you want that? We are already passing labels from the DV to the import/upload/clone pods for things like istio and I don't think we want to pass those to the PVC.

So we actually discussed this in the sig-storage meeting and I tried to summarize it in the issue itself;
Basically, if we look at DVs as a PVC creation template, you will want to say which labels you want on your PVC, right?

I would understand the argument that you could just grab the PVC and label it after the DV creation, but that is a bit clunky..
What if you created the DV with `generateName`? like the issue author did?
Anyway, we were already passing down annotations and could not think why we aren't giving the same treatment to labels.

pkg/controller/datavolume/controller-base.go
```diff
@@ -761,7 +746,9 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume,
 	if util.ResolveVolumeMode(targetPvcSpec.VolumeMode) == corev1.PersistentVolumeFilesystem {
 		labels[common.KubePersistentVolumeFillingUpSuppressLabelKey] = common.KubePersistentVolumeFillingUpSuppressLabelValue
 	}
-	labels = passDataVolumeInstancetypeLabelstoPVC(dataVolume.GetLabels(), labels)
+	for k, v := range dataVolume.Labels {
```
akalenyu, line 749: Note that not all flows we have are using `newPersistentVolumeClaim`
One example I could find is smart clone which creates the target PVC in
https://github.com/kubevirt/containerized-data-importer/blob/b875775732019b3f1f3c614c763bfca508c06bb5/pkg/controller/datavolume/smart-clone-controller.go#L370-L374

akalenyu (commented):
Awesome! great spot!

Would it be possible to also add a one-liner to some func tests that ensures we pass the labels?
in e.g. import/upload/clone/...

pkg/controller/datavolume/controller-base.go
arnongilboa: Good point!

arnongilboa:
@ido106 please fill in the **What this PR does / why we need it** and  **Release note** sections in the card description.

pkg/controller/datavolume/controller-base.go
ido106: Hi Alex, there is no DV in that function and the PVC is created from a VolumeSnapshot, so I don't think labels should be passed in this case. wdyt ?

mhenriks:
@ido106 any update on this?  I believe we are waiting on a functional test?

arnongilboa:
> @ido106 any update on this? I believe we are waiting on a functional test?

Ido is in his exam period, so I guess he'll continue his first PR when he's back.

pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -352,7 +352,7 @@ func (r *SmartCloneReconciler) getTargetPVC(dataVolume *cdiv1.DataVolume) (*core
 	return pvc, nil
 }
 
-func newPvcFromSnapshot(name string, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec) (*corev1.PersistentVolumeClaim, error) {
+func newPvcFromSnapshot(c client.Client, name string, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec) (*corev1.PersistentVolumeClaim, error) {
```
akalenyu, line 355: I think you may even be able to just pass
`annotations, labels map[string]string`/`dv *cdiv1.DataVolume` to this func
(the DV should be already available to you when calling `newPvcFromSnapshot`)

pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -369,6 +369,17 @@ func newPvcFromSnapshot(name string, snapshot *snapshotv1.VolumeSnapshot, target
 		common.CDILabelKey:       common.CDILabelValue,
 		common.CDIComponentLabel: common.SmartClonerCDILabel,
 	}
+
+	dv, err := getDataVolume(c, client.ObjectKeyFromObject(snapshot))
+	if err != nil {
+		return nil, err
+	}
+	if dv != nil {
+		for k, v := range dv.Labels {
```
akalenyu, line 378: It seems like we also don't pass the annotations in this case, we probably should

pkg/controller/datavolume/controller-base.go
```diff
@@ -365,9 +365,9 @@ func (r *ReconcilerBase) getPVC(dv *cdiv1.DataVolume) (*corev1.PersistentVolumeC
 	return pvc, nil
 }
 
-func (r *ReconcilerBase) getDataVolume(key types.NamespacedName) (*cdiv1.DataVolume, error) {
+func getDataVolume(c client.Client, key types.NamespacedName) (*cdiv1.DataVolume, error) {
```
akalenyu, line 368: Hmm not sure why this one needed to be changed?

akalenyu (commented):
Looking good, some things to consider, and maybe it's worth to add one liner checks in some of the functional tests?

pkg/controller/datavolume/controller-base.go
ido106: I use this function in 2 different places in the code and one of them doesn't have access to the function this way, so I changed the signature to be a regular function and requested the client as a parameter instead.

akalenyu: AFAIK you only use `func (r *SmartCloneReconciler) getDataVolume` and not this one

ido106: Thanks Alex, I tried to fix what you mentioned.
Plz let me know if there is any problem and I'll fix it tomorrow.

pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -369,22 +370,32 @@ func newPvcFromSnapshot(name string, snapshot *snapshotv1.VolumeSnapshot, target
 		common.CDILabelKey:       common.CDILabelValue,
 		common.CDIComponentLabel: common.SmartClonerCDILabel,
 	}
+
```
alromeros, line 373: This is a massive nitpick but I'd drop this empty line, as done with annotations below

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -464,14 +464,15 @@ var _ = Describe("All DataVolume Tests", func() {
...
-		It("Should pass annotation from DV to created a PVC on a DV", func() {
+		It("Should pass annotations and labels from DV to created PVC", func() {
 			dv := NewImportDataVolume("test-dv")
 			dv.SetAnnotations(make(map[string]string))
 			dv.GetAnnotations()["test-ann-1"] = "test-value-1"
 			dv.GetAnnotations()["test-ann-2"] = "test-value-2"
 			dv.GetAnnotations()[AnnSource] = "invalid phase should not copy"
 			dv.GetAnnotations()[AnnPodNetwork] = "data-network"
 			dv.GetAnnotations()[AnnPodSidecarInjection] = "false"
+			dv.Labels = map[string]string{"test": "test-label"}
```
alromeros, line 475: Nitpick, but I'd use the same convention for Annotations and Labels (Get/Set or not).

alromeros (commented):
Good job Ido! I think after your latest commit this looks great. I'm only commenting some minor nitpicks in case you end up committing again, no need to change them. Everything else looks good to me so I'll give a preventive lgtm, but let's see what other teammates have to say.

/lgtm

pkg/controller/datavolume/controller-base.go
akalenyu: Looks good, I just think we should test this in functional tests since we are now committing that CDI will always pass labels

tests/datavolume_test.go
```diff
@@ -1135,6 +1135,31 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			table.Entry("[test_id:8045]for clone DataVolume", createCloneDataVolume, fillCommand),
 		)
 
+		table.DescribeTable("Should pass all DV labels and annotations to the PVC", func(dvFunc func(string, string, string) *cdiv1.DataVolume, url string) {
```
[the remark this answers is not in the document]
ido106: Thanks Alex. I tried to fix it, hope it's ok

tests/cloner_test.go
```diff
@@ -2610,6 +2619,14 @@ var _ = Describe("all clone tests", func() {
 			size := "1Gi"
 			createSnapshot(size, nil, volumeMode)
 
+			dataVolume := utils.NewDataVolumeForCloningFromSnapshot("pass-all-labels-and-annotations-test", size, snapshot.Namespace, snapshot.Name, nil, &volumeMode)
+			dataVolume.Labels = map[string]string{"test-label-1": "test-label-value-1"}
+			dataVolume.Annotations = map[string]string{"test-annotation-1": "test-annotation-value-1"}
+			dataVolume, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, targetNs.Name, dataVolume)
+			// All labels and annotations passed
+			Expect(dataVolume.Labels["test-label-1"]).To(Equal("test-label-value-1"))
+			Expect(dataVolume.Annotations["test-annotation-1"]).To(Equal("test-annotation-value-1"))
```
arnongilboa, line 2628: You added labels and annotations to the DV, so you need to check they eere passed  to the **PVC**.

tests/cloner_test.go
```diff
@@ -282,11 +282,20 @@ var _ = Describe("all clone tests", func() {
 			DescribeTable("[test_id:1355]Should clone data across different namespaces", func(targetSize string) {
 				pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
 				pvcDef.Namespace = f.Namespace.Name
+
+				pvcDef.Labels = map[string]string{"test-label-1": "test-label-key-1"}
+				pvcDef.Annotations = map[string]string{"test-annotation-1": "test-annotation-key-1"}
+
```
arnongilboa, line 288: You should add labels and annotations to *DV*, not to PVC, and check they are passed to the *PVC*.

ido106: I had a few issues unrelated to my code that Alex and I couldn't resolve..
I did a rebase and got conflicts in places I didn't edit, in addition to some errors I couldn't solve when I ran test-functional.
I tried to push anyway

akalenyu:
/lgtm

awesome!

akalenyu:
/hold

I think I see the problem
https://github.com/kubevirt/containerized-data-importer/blob/dbc9997fd96b492e1b8f55553a36289fbc2abc23/tests/cloner_test.go#L2878-L2879
The annotations are being overwritten so we lose `AnnDeleteAfterCompletion`
We should probably just append the test annotations/labels to existing ones

akalenyu:
/hold cancel

But we have an unrelated CI issue following 1.56.0 release related to v1alpha1 manifests

PR #2551, opened 2023-01-22
Clean up when dv is gone in addition to deletionTimestamp

akalenyu:

Not sure how we do cleanup of a DV that is deleted mid-way without this?
`cleanupTransfer` for example seems to only happen if phase is succeeded.

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
/close

not a good idea, finalizer on DV is way better

PR #2556, opened 2023-01-25
Don't go further in reconcile loop after cleanup finished successfully

akalenyu:

Right now if cleanup finishes successfully we just keep going on in the reconcile loop, 
even though deletion timestamp is set.
Results in nil ptr when doing cross ns smart clone:
```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1543ec1]
 
goroutine 1133 [running]:
kubevirt.io/containerized-data-importer/pkg/controller/datavolume.(*PvcCloneReconciler).selectCloneStrategy(0xc001a69768, 0xc0050e65a0, 0x0)
        /remote-source/app/pkg/controller/datavolume/pvc-clone-controller.go:489 +0x81
```

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -260,8 +260,9 @@ func (r ReconcilerBase) sync(log logr.Logger, req reconcile.Request, cleanup, pr
 	if dv.DeletionTimestamp != nil {
 		log.Info("DataVolume marked for deletion, cleaning up")
 		if cleanup != nil {
-			err := cleanup(syncRes)
-			return syncRes, err
+			if err := cleanup(syncRes); err != nil {
+				return syncRes, err
+			}
 		}
 		syncRes.result = &reconcile.Result{}
 		return syncRes, nil


pkg/controller/datavolume/controller-base.go
```diff
@@ -260,8 +260,9 @@ func (r ReconcilerBase) sync(log logr.Logger, req reconcile.Request, cleanup, pr
 	if dv.DeletionTimestamp != nil {
 		log.Info("DataVolume marked for deletion, cleaning up")
 		if cleanup != nil {
-			err := cleanup(syncRes)
-			return syncRes, err
+			if err := cleanup(syncRes); err != nil {
+				return syncRes, err
+			}
```
mhenriks, line 265: based on the code change, shouldn't the title be "DO keep going after cleanup finished successfully"?

akalenyu: don't go further in reconcile loop is what I meant (i.e. selectCloneStrategy), let me edit the title

mhenriks: sorry but the title is still not clear to me.  the only thing that appears to have changed is that if cleanup() completes without error `syncres.result` gets initialized.

akalenyu: right, which will result in backing out of `syncClone` early enough and avoid the nil ptr.
imagine scenario where dv timestamp != nil and cleanup finishes successfully. nothing stops you from going further

PR #2557, opened 2023-01-26
Fix StorageProfile incomplete metric test flakiness

arnongilboa:

Flakiness seems to be introduced by #2546 

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -981,18 +981,24 @@ var _ = Describe("ALL Operator tests", func() {
 					Skip("This test depends on prometheus infra being available")
 				}
 
-				numAddedStorageClasses = 2
 				defaultStorageClass := utils.GetDefaultStorageClass(f.K8sClient)
 				defaultStorageClassProfile := &cdiv1.StorageProfile{}
 				err := f.CrClient.Get(context.TODO(), types.NamespacedName{Name: defaultStorageClass.Name}, defaultStorageClassProfile)
 				Expect(err).ToNot(HaveOccurred())
+
+				originalMetricVal := 0
+				Eventually(func() int {
+					originalMetricVal = getMetricValue("kubevirt_cdi_incomplete_storageprofiles_total")
+					return originalMetricVal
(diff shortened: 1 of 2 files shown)


tests/operator_test.go
```diff
@@ -981,18 +981,24 @@ var _ = Describe("ALL Operator tests", func() {
 					Skip("This test depends on prometheus infra being available")
 				}
 
-				numAddedStorageClasses = 2
 				defaultStorageClass := utils.GetDefaultStorageClass(f.K8sClient)
 				defaultStorageClassProfile := &cdiv1.StorageProfile{}
 				err := f.CrClient.Get(context.TODO(), types.NamespacedName{Name: defaultStorageClass.Name}, defaultStorageClassProfile)
 				Expect(err).ToNot(HaveOccurred())
+
+				originalMetricVal := 0
```
akalenyu, line 989: I think originally the metric values is not 0 because we have the "local" storage class

awels: But local should not trigger the warning anymore since we added it to the storage profile.

akalenyu: did we? where?

arnongilboa: Due to PR #2546 playing with CDI CR, in this test case it takes `StorageProfileReconciler` a second or two to get to its `IncompleteProfileGauge.Set(>0)`.

akalenyu (commented):
what was the flake here? IIUC the other PR introduced a test that did not mess around with storage classes at all

arnongilboa:
> what was the flake here? IIUC the other PR introduced a test that did not mess around with storage classes at all

see my comment above

tests/operator_test.go
awels: Interesting we only did it for LSO not U/S local storage.

akalenyu: yeah because you still need to alert when someone is rolling their own static PVs

PR #2559, opened 2023-01-27
Grab target PVC early on to avoid false status updates

akalenyu:

Now that we update status in the end of a loop regardless of err,
it is important that we have the PVC in place for that.
Otherwise, if we fail at cleanup/prepare, syncRes.pvc is nil,
and we get false status updates like the following:
```bash
NAMESPACE   NAME               PHASE                             PROGRESS   RESTARTS
default     cloned-target-dv   Succeeded                         N/A                   23s
default     cloned-target-dv   Pending                           N/A                   23s
default     cloned-target-dv   Pending                           N/A                   23s
default     cloned-target-dv   Pending                           N/A                   23s
NAMESPACE   NAME                 STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
golden-ns   golden-snap-source   Bound    pvc-63ed1368-9f84-44ce-8b4b-7e9e3c4574f4   8Gi        RWO            rook-ceph-block   5m59s
(Target got deleted)
```
This results in a nasty bug where the target PVC simply disappears in cross ns clones.

Fixes #

**Release note**:

```release-note
BugFix: target PVC disappears after cross namespace clone 
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -256,6 +256,10 @@ func (r ReconcilerBase) sync(log logr.Logger, req reconcile.Request, cleanup, pr
 	}
 	syncRes.dv = dv
 	syncRes.dvMutated = dv.DeepCopy()
+	syncRes.pvc, err = r.getPVC(dv)
+	if err != nil {
+		return syncRes, err
+	}
 
 	if dv.DeletionTimestamp != nil {
 		log.Info("DataVolume marked for deletion, cleaning up")
@@ -273,10 +277,6 @@ func (r ReconcilerBase) sync(log logr.Logger, req reconcile.Request, cleanup, pr
 		}
 	}
 
-	syncRes.pvc, err = r.getPVC(dv)
-	if err != nil {
-		return syncRes, err
-	}
 	if syncRes.pvc != nil {
 		if err := r.garbageCollect(syncRes, log); err != nil {
 			return syncRes, err


mhenriks:
Good catch @akalenyu but I think this issue points to a fundamental flaw with the current sync/updateStatus implementation.  

I think it would be better if there is no state shared between sync and updateStatus.  updateStatus should stand on it's own, come to it's own conclusions based on what it observes.  It is okay if it is "late to the party" and does not observe the latest changes in sync().  It will eventually converge.  This is what kubevirt does

We can get this PR in, but I like to continue the refactoring work with the above goal.  What do you think @arnongilboa @akalenyu @alromeros @awels ?

arnongilboa:
> Good catch @akalenyu but I think this issue points to a fundamental flaw with the current sync/updateStatus implementation.
> 
> I think it would be better if there is no state shared between sync and updateStatus. updateStatus should stand on it's own, come to it's own conclusions based on what it observes. It is okay if it is "late to the party" and does not observe the latest changes in sync(). It will eventually converge. This is what kubevirt does
> 
> We can get this PR in, but I like to continue the refactoring work with the above goal. What do you think @arnongilboa @akalenyu @alromeros @awels ?

If there's no state shared between sync and updateStatus, it will force us to duplicate logic and make updateStatus much smarter than it is now, as it will have to re-implement a flow quite similar to sync, which also hurts maintainability.

akalenyu:
> Good catch @akalenyu but I think this issue points to a fundamental flaw with the current sync/updateStatus implementation.
> 
> I think it would be better if there is no state shared between sync and updateStatus. updateStatus should stand on it's own, come to it's own conclusions based on what it observes. It is okay if it is "late to the party" and does not observe the latest changes in sync(). It will eventually converge. This is what kubevirt does
> 
> We can get this PR in, but I like to continue the refactoring work with the above goal. What do you think @arnongilboa @akalenyu @alromeros @awels ?

I agree that this points to a deeper flaw, consider PVC GET errors, for example;
We would still get false status updates when we should've just backed out and requeued the request

alromeros:
> Good catch @akalenyu but I think this issue points to a fundamental flaw with the current sync/updateStatus implementation.
> 
> I think it would be better if there is no state shared between sync and updateStatus. updateStatus should stand on it's own, come to it's own conclusions based on what it observes. It is okay if it is "late to the party" and does not observe the latest changes in sync(). It will eventually converge. This is what kubevirt does
> 
> We can get this PR in, but I like to continue the refactoring work with the above goal. What do you think @arnongilboa @akalenyu @alromeros @awels ?

Just like @arnongilboa, I don't like the idea of duplicating logic at all... but as long as it is a reasonable amount, it ultimately seems like the best option. I think we should prioritize robustness and just follow the standard already set by Kubevirt and keep updateStatus and sync self-sufficient.

awels:
/override pull-containerized-data-importer-e2e-destructive
Known failures, and don't want to hold up this PR.

PR #2571, opened 2023-02-02
Watch DataSources in DV clone controller

arnongilboa:

DV with sourceRef DataSource pointing to non-existing PVC, is to be reconciled when the PVC is created and populated and the DataSource is Ready.

Fixes bz #2156517

**Release note**:
```release-note
BugFix: VM with DataSource referring a non-existing PVC won't start after DataSource fix
```

### DIFF
--- a/tests/datasource_test.go
+++ b/tests/datasource_test.go
@@ -9,10 +9,12 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
+	dvc "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
 	"kubevirt.io/containerized-data-importer/tests/framework"
 	"kubevirt.io/containerized-data-importer/tests/utils"
 )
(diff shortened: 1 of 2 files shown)


tests/datasource_test.go
```diff
@@ -65,20 +67,40 @@ var _ = Describe("DataSource", func() {
...
 
+		By("Update DataSource source PVC to nonexisting one")
 		ds.Spec.Source.PVC = &cdiv1.DataVolumeSourcePVC{Namespace: f.Namespace.Name, Name: pvc1Name}
 		ds, err = f.CdiClient.CdiV1beta1().DataSources(ds.Namespace).Update(context.TODO(), ds, metav1.UpdateOptions{})
 		Expect(err).ToNot(HaveOccurred())
 		ds = waitForReadyCondition(ds, corev1.ConditionFalse, "NotFound")
 
+		By("Create clone DV with SourceRef pointing the DataSource")
+		dv := utils.NewDataVolumeWithSourceRef("clone-dv", "1Gi", ds.Namespace, ds.Name)
+		dv.Annotations[controller.AnnImmediateBinding] = "true"
```
awels, line 84: We are not using `ForceBindPvcIfDvIsWaitForFirstConsumer` anymore to ensure the right thing happens in WFFC case? I don't mind just adding the annotation as it is simpler, just asking.

arnongilboa: I personally prefer the simpler DV annotation instead of WaitForPVC & ForceBind, and it'salso used in kubevirt DIC/DVTemplates etc.

awels: Okay

awels:
And just so I am clear on what is happening. We have a DV with a DataSource, but the DataSource is not pointing to a valid PVC, so the DataSource is not in a ready phase. Then the PVC is created, this causes a reconcile in the DataSource, and it becomes ready. And now this will cause a reconcile in the DV controller to actually initiate the clone?

arnongilboa:
> And just so I am clear on what is happening. We have a DV with a DataSource, but the DataSource is not pointing to a valid PVC, so the DataSource is not in a ready phase. Then the PVC is created, this causes a reconcile in the DataSource, and it becomes ready. And now this will cause a reconcile in the DV controller to actually initiate the clone?

Exactly.

alromeros (commented):
We recently added this check in our webhook to avoid creating DVs with a DataSource with an invalid (nil) PVC:

https://github.com/kubevirt/containerized-data-importer/blob/a7606f7ce7aa3bc8afad9bd2a6092999e66c919d/pkg/apiserver/webhooks/datavolume-validate.go#L435-L441

Do you think it clashes with this PR?

arnongilboa:
> We recently added this check in our webhook to avoid creating DVs with a DataSource with an invalid (nil) PVC:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/a7606f7ce7aa3bc8afad9bd2a6092999e66c919d/pkg/apiserver/webhooks/datavolume-validate.go#L435-L441
> 
> Do you think it clashes with this PR?

We handle DataSource with non-existing PVC, not nil, but sure we can extend it to the nil case.

/hold

PR #2577, opened 2023-02-07
Fix operator_test CDI CR Gets

arnongilboa:

Get rid of the assumption that CR name is cdi, so tests won't be skippd when CR name is not cdi.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -33,7 +33,6 @@ import (
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
-	cdiClientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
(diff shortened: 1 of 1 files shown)


tests/operator_test.go
```diff
@@ -1430,17 +1392,19 @@ var _ = Describe("ALL Operator tests", func() {
...
 
+func getCDI(f *framework.Framework) *cdiv1.CDI {
+	By("Getting CDI resource")
 	cdis, err := f.CdiClient.CdiV1beta1().CDIs().List(context.TODO(), metav1.ListOptions{})
 	Expect(err).ToNot(HaveOccurred())
-	Expect(len(cdis.Items)).To(Equal(1))
-	cr := &cdis.Items[0]
-
-	return cr, cdiPods
+	Expect(cdis.Items).To(HaveLen(1))
```
ShellyKa13, line 1406: so do we prefer that test will fail because they expect to have at least 1 cdi? or should we skip if len < 1 like somewhat done before:
				if len(cdis.Items) < 1 {
					Skip("CDI CR does not exist.  Probably managed by another operator so skipping.")
				}

arnongilboa: why would you run CDI func tests if you have no CDI on the cluster? failing looks good to me...

awels (approved):
/approve

PR #2583, opened 2023-02-12
Annotation to check for statically provisioned PVs when creating DataVolumes

mhenriks:

This PR basically does the following if a DataVolume that has the `cdi.kubevirt.io/storage.checkStaticVolume` annotation is created:

1.  Check if there is PV explicitly provisioned (has claimRef name+namespace set) for the target PVC
2. If so, the "base" controller will create the target PVC in `syncCommon()`
3. The target PVC will initially have an annotation `cdi.kubevirt.io/storage.persistentVolumeList` that contains the names of the PVCs from step 1
4. When the target PVC gets bound to a PV, `syncCommon()` will check that the PVC is bound to one of the PVs in the `persistentVolumeList` annotation
5. If the bind works as expected, `cdi.kubevirt.io/storage.persistentVolumeList` is removed and `cdi.kubevirt.io/storage.populatedFor` is added to the PVC
6. If the PVC does not bind to one of the expected PVs, the PVC is deleted and we go back to step 1

The motivation for this PR is KubeVirt integration with a specific "Metro DR" solution.  That solution bundles:

[RamenDR](https://github.com/RamenDR/ramen)
[volume-replication-operator](https://github.com/csi-addons/volume-replication-operator)
[OCM](https://open-cluster-management.io/concepts/architecture/)

Fixes #

TODO

- [X] import
- [x] upload
- [x] pvc clone
- [x] snapshot clone
- [x] external populator

I've taken the liberty to do some minor refactoring along the way.

[This](https://github.com/kubevirt/containerized-data-importer/pull/2559#issuecomment-1407056241) has not been addressed and is bumming me out bigtime

**Release note**:

```release-note
cdi.kubevirt.io/storage.checkStaticVolume annotation skips volume population step if a static PV is configured for target DataVOlume PVC
```


pkg/controller/datavolume/import-controller.go
```diff
@@ -101,12 +101,29 @@ func NewImportController(
 	return datavolumeController, nil
 }
 
+// TODO find a better place for this
```
akalenyu, line 104: There's https://github.com/kubevirt/containerized-data-importer/blob/11fc97e4a64bda2d67a14b3cd26f51dbaf06c3dc/pkg/controller/datavolume/smart-clone-controller.go#L362
But I might be missing a reason that it's not suitable for us

pkg/controller/datavolume/controller-base.go
```diff
@@ -323,18 +341,115 @@ func (r ReconcilerBase) syncUpdate(log logr.Logger, syncRes *dataVolumeSyncResul
...
+
+func (r *ReconcilerBase) getAvailableVolumesForDV(syncRes *dataVolumeSyncResult, log logr.Logger) ([]string, error) {
+	pvList := &corev1.PersistentVolumeList{}
+	fields := client.MatchingFields{claimRefField: claimRefIndexKeyFunc(syncRes.dv.Namespace, syncRes.dv.Name)}
+	if err := r.client.List(context.TODO(), pvList, fields); err != nil {
+		return nil, err
+	}
+	var pvNames []string
+	for _, pv := range pvList.Items {
+		// TODO - should we be more specific here, like do the actual matching in pv controller?
```
akalenyu, line 426: Maybe something from https://pkg.go.dev/k8s.io/component-helpers@v0.26.1/storage/volume
can help here?

tests/static-volume_test.go
```diff
@@ -0,0 +1,153 @@
...
+			dv, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dvDef)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Waiting for target DV to succeed")
+			err = utils.WaitForDataVolumePhaseWithTimeout(f, f.Namespace.Name, cdiv1.Succeeded, dv.Name, 300*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+
+			pvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(dv.Namespace).Get(context.TODO(), dv.Name, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pvc.Spec.VolumeName).To(Equal(pvName))
```
akalenyu, line 144: This test is great, just one consideration, maybe we should make sure pv.CreationTimestamp < dv.CreationTimestamp

tests/utils/datavolume.go
```diff
@@ -317,6 +317,30 @@ func NewDataVolumeWithHTTPImportToBlockPV(dataVolumeName string, size string, ht
 
 // NewDataVolumeWithExternalPopulation initializes a DataVolume struct meant to be externally populated
 func NewDataVolumeWithExternalPopulation(dataVolumeName, size, storageClassName string, volumeMode corev1.PersistentVolumeMode, dataSource, dataSourceRef *corev1.TypedLocalObjectReference) *cdiv1.DataVolume {
+	dataVolume := &cdiv1.DataVolume{
```
akalenyu, line 320: Could you explain what happened here?
Usually, I thought we shift towards using the storage API mostly

mhenriks:
@akalenyu @arnongilboa what do you think?

tests/utils/datavolume.go
mhenriks: The storage api obscures the matching a bit, just have to be consistent either way (all pvc or all storage).  I took the PVC route to be explicit as possible.  Basically when I originally wrote the test I was using `pvc` to create the source then didn't get a match I used `storage` for target.

tests/static-volume_test.go
mhenriks: sure

pkg/controller/datavolume/controller-base.go
mhenriks: unfortunately that package is not as fleshed out in k8s 23 but using it anyway

pkg/controller/datavolume/import-controller.go
mhenriks: problem is that pv claimRef is not a `metav1.Object` so have to sonstruct from namespace/name

pkg/controller/datavolume/controller-base.go
```diff
@@ -112,6 +114,22 @@ func pvcIsPopulated(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) boo
...
+	if _, ok := dv.Annotations[cc.AnnCheckStaticVolume]; !ok {
+		return false
+	}
+	_, ok := pvc.Annotations[cc.AnnPersistentVolumeList]
+	return ok
+}
+
+func shouldSetPending(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) bool {
+	return checkStaticProvisionPending(pvc, dv) ||
+		(pvc == nil && dv.Status.Phase == cdiv1.PhaseUnset)
```
arnongilboa, line 130: I would add `&& dv != nil`

pkg/controller/datavolume/util.go
```diff
@@ -354,3 +355,71 @@ func getReconcileResult(result *reconcile.Result) reconcile.Result {
...
+		return fmt.Errorf("incompatible accessMode")
+	}
+
+	return nil
+}
+
+// CheckVolumeModeMismatches is a convenience method that checks volumeMode for PersistentVolume
+// and PersistentVolumeClaims
+// TODO - later versions of k8s.io/component-helpers have this function
+func CheckVolumeModeMismatches(pvcSpec *v1.PersistentVolumeClaimSpec, pvSpec *v1.PersistentVolumeSpec) bool {
```
arnongilboa, line 396: It's strange k8s (and you) choose `Mismatches` instead of being positive

pkg/controller/datavolume/util.go
```diff
@@ -354,3 +355,71 @@ func getReconcileResult(result *reconcile.Result) reconcile.Result {
...
+// TODO - later versions of k8s.io/component-helpers have this function
+func CheckAccessModes(claim *v1.PersistentVolumeClaim, volume *v1.PersistentVolume) bool {
+	pvModesMap := map[v1.PersistentVolumeAccessMode]bool{}
+	for _, mode := range volume.Spec.AccessModes {
+		pvModesMap[mode] = true
+	}
+
+	for _, mode := range claim.Spec.AccessModes {
+		_, ok := pvModesMap[mode]
+		if !ok {
```
arnongilboa, line 420: I think it's more clear to write it as:
`if hasMode := pvModesMap[mode]; hasMode`

pkg/controller/datavolume/controller-base.go
```diff
@@ -323,18 +346,123 @@ func (r ReconcilerBase) syncUpdate(log logr.Logger, syncRes *dataVolumeSyncResul
...
+	}
+
+	var volumes []string
+	if err := json.Unmarshal([]byte(volumeAnno), &volumes); err != nil {
+		return err
+	}
+
+	for _, v := range volumes {
+		if v == syncRes.pvc.Spec.VolumeName {
+			pvcCpy := syncRes.pvc.DeepCopy()
```
arnongilboa, line 400: why use a copy if we don't need `DeepEqual` check here?

pkg/controller/datavolume/controller-base.go
```diff
@@ -323,18 +346,123 @@ func (r ReconcilerBase) syncUpdate(log logr.Logger, syncRes *dataVolumeSyncResul
...
+			if err := r.updatePVC(pvcCpy); err != nil {
+				return err
+			}
+			syncRes.pvc = pvcCpy
+			return nil
+		}
+	}
+
+	// delete the pvc and hope for better luck...
+	pvcCpy := syncRes.pvc.DeepCopy()
```
arnongilboa, line 413: why copy is needed here? and why set `syncRes.pvc = pvcCpy` after the deletion instead of setting it to `nil`?

pkg/controller/datavolume/controller-base.go
```diff
@@ -532,11 +656,6 @@ func (r ReconcilerBase) updateStatusCommon(syncRes dataVolumeSyncResult, updateS
 		if err := r.reconcileProgressUpdate(dataVolumeCopy, pvc, &result); err != nil {
 			return result, err
 		}
-	} else {
-		_, ok := dataVolumeCopy.Annotations[cc.AnnPrePopulated]
-		if ok {
-			dataVolumeCopy.Status.Phase = cdiv1.Pending
-		}
```
arnongilboa, line 539: why isn't this covered in `shouldSetPending()`?

pkg/controller/datavolume/external-population-controller.go
```diff
@@ -184,35 +186,38 @@ func isSnapshotPopulation(pvc *corev1.PersistentVolumeClaim) bool {
...
 		return errors.Errorf("undefined population source")
 	}
 	pvc.Annotations[cc.AnnExternalPopulation] = "true"
 	return nil
 }
 
-func (r PopulatorReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeSyncResult, error) {
+func (r *PopulatorReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeSyncResult, error) {
 	syncRes, syncErr := r.syncExternalPopulation(log, req)
+	// TODO _ I think it is bad form that the datavolume is updated even in the case of error
```
arnongilboa, line 199: This is fixed in PR #2590

pkg/controller/datavolume/external-population-controller.go
```diff
@@ -184,35 +186,38 @@ func isSnapshotPopulation(pvc *corev1.PersistentVolumeClaim) bool {
...
 	}
-	return *syncRes, syncErr
+	return syncRes, syncErr
 }
 
-func (r PopulatorReconciler) updateStatus(syncRes dataVolumeSyncResult, syncErr error) (reconcile.Result, error) {
+func (r *PopulatorReconciler) updateStatus(syncRes dataVolumeSyncResult, syncErr error) (reconcile.Result, error) {
+	// TODO FIXME - WE SHOULD ALWAYS UPDATE STATUS
 	if syncErr != nil {
+		r.log.Info("FIXME should not return because of this", "err", syncErr)
```
arnongilboa, line 220: This is fixed in PR #2590

pkg/controller/datavolume/external-population-controller.go
```diff
@@ -84,16 +84,18 @@ func NewPopulatorController(ctx context.Context, mgr manager.Manager, log logr.L
...
 }
 
 // If dataSourceRef is set (external populator), use an empty spec.Source field
-func (r PopulatorReconciler) prepare(syncRes *dataVolumeSyncResult) error {
+func (r *PopulatorReconciler) prepare(syncRes *dataVolumeSyncResult) error {
 	if !dvUsesVolumePopulator(syncRes.dv) {
 		return errors.Errorf("undefined population source")
 	}
+	// TODO - let's revisit this
+	syncRes.dv.Spec.Source = &cdiv1.DataVolumeSource{}
```
arnongilboa, line 98: @alromeros wdyt?

pkg/controller/datavolume/import-controller.go
arnongilboa: I would put a common func in `pkg/controller/common` and get rid of the following:
```
datasource-controller.go:        getKey := func(namespace, name string) string {
datavolume/clone-controller-base.go:     getKey := func(namespace, name string) string {
datavolume/pvc-clone-controller.go:      getKey := func(namespace, name string) string {
```

pkg/controller/datavolume/import-controller.go
```diff
@@ -101,19 +101,36 @@ func NewImportController(
...
+func claimRefIndexKeyFunc(namespace, name string) string {
+	return namespace + "/" + name
+}
+
 func addDataVolumeImportControllerWatches(mgr manager.Manager, datavolumeController controller.Controller) error {
+	// TODO - these are technically shared indexes (not just used by this controller) so should live elsewhere
```
arnongilboa, line 110: Good Point. So I guess all inexing should move to controller-base?

pkg/controller/datavolume/import-controller.go
```diff
@@ -205,36 +222,41 @@ func (r ImportReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, pvc *c
...
-func (r ImportReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+func (r *ImportReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	log := r.log.WithValues("DataVolume", req.NamespacedName)
 	return r.updateStatus(r.sync(log, req))
 }
 
-func (r ImportReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeSyncResult, error) {
+func (r *ImportReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeSyncResult, error) {
 	syncRes, syncErr := r.syncImport(log, req)
+	// TODO _ I think it is bad form that the datavolume is updated even in the case of error
```
arnongilboa, line 232: This is fixed in PR #2590

tests/static-volume_test.go
akalenyu: I am a broken record about this by now, but, maybe we would sleep better if there is an md5 check at the end of this test

pkg/controller/datavolume/import-controller.go
```diff
@@ -205,36 +222,41 @@ func (r ImportReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, pvc *c
...
+	return syncRes, syncErr
 }
 
-func (r ImportReconciler) updateStatus(syncRes dataVolumeSyncResult, syncErr error) (reconcile.Result, error) {
+func (r *ImportReconciler) updateStatus(syncRes dataVolumeSyncResult, syncErr error) (reconcile.Result, error) {
+	// TODO FIXME - WE SHOULD ALWAYS UPDATE STATUS
+	// BUT THIS TEST FAILS WITHOUT IT
+	// [test_id:7737]when creating import dv with given valid url
 	if syncErr != nil {
+		r.log.Info("FIXME should not return because of this", "err", syncErr)
```
arnongilboa, line 259: This is fixed in PR #2590

pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -257,17 +257,19 @@ func (r PvcCloneReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, pvc
 	return nil
 }
 
-func (r PvcCloneReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeCloneSyncResult, error) {
+func (r *PvcCloneReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeCloneSyncResult, error) {
 	syncRes, syncErr := r.syncClone(log, req)
+	// TODO _ I think it is bad form that the datavolume is updated even in the case of error
+	// however if this function exits without calling syncUpdate() test_id:7736 and test_id:7739 fail
```
arnongilboa, line 263: This is fixed in PR #2590

tests/utils/datavolume.go
akalenyu: gotcha

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -159,17 +159,18 @@ func (r SnapshotCloneReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume,
 	return nil
 }
 
-func (r SnapshotCloneReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeCloneSyncResult, error) {
+func (r *SnapshotCloneReconciler) sync(log logr.Logger, req reconcile.Request) (dataVolumeCloneSyncResult, error) {
 	syncRes, syncErr := r.syncSnapshotClone(log, req)
+	// TODO _ I think it is bad form that the datavolume is updated even in the case of error
```
arnongilboa, line 164: This is fixed in PR #2590

pkg/controller/datavolume/upload-controller.go
```diff
@@ -101,32 +101,35 @@ func (r UploadReconciler) updateAnnotations(dataVolume *cdiv1.DataVolume, pvc *c
...
+	if err := r.handlePvcCreation(log, &syncRes, r.updateAnnotations); err != nil {
 		syncErr = err
 	}
-	return *syncRes, syncErr
+	return syncRes, syncErr
 }
 
-func (r UploadReconciler) updateStatus(syncRes dataVolumeSyncResult, syncErr error) (reconcile.Result, error) {
+func (r *UploadReconciler) updateStatus(syncRes dataVolumeSyncResult, syncErr error) (reconcile.Result, error) {
+	// TODO FIXME - WE SHOULD ALWAYS UPDATE STATUS
```
arnongilboa, line 130: This is fixed in PR #2590

pkg/controller/datavolume/external-population-controller.go
alromeros: @mhenriks what do you mean? Maybe adding a source for external populators?

tests/static-volume_test.go
mhenriks: not a bad idea

pkg/controller/datavolume/controller-base.go
mhenriks: dv will never be nil in either case and `shouldSetPending` kind of assumes the dv exists

pkg/controller/datavolume/util.go
mhenriks: eh, would rather do exactly what the lib does

pkg/controller/datavolume/controller-base.go
mhenriks: Best practice to always mutate a copy especially in this case where the pvc reference lives beyond the scope of the call.  Imagine the update call fails, that annotation will stay in the shared reference even though that is not how it looks in etcd

pkg/controller/datavolume/controller-base.go
mhenriks: Same reasoning as above, what if the delete call mutates the variable in some way and fails?  I don't think it does but I think you get my drift

pkg/controller/datavolume/external-population-controller.go
mhenriks: It just seemed awkward and unexpected we don't have to do anything in the pr.  I'll remove the comment

pkg/controller/datavolume/controller-base.go
```diff
@@ -323,18 +346,123 @@ func (r ReconcilerBase) syncUpdate(log logr.Logger, syncRes *dataVolumeSyncResul
...
+
+	// delete the pvc and hope for better luck...
+	pvcCpy := syncRes.pvc.DeepCopy()
+	if err := r.client.Delete(context.TODO(), pvcCpy, &client.DeleteOptions{}); err != nil {
+		return err
+	}
+
+	syncRes.pvc = pvcCpy
+
+	return fmt.Errorf("DataVolume bound to unexpected PV %s", syncRes.pvc.Spec.VolumeName)
```
alromeros, line 420: If I understand correctly by reading the PR description, shouldn't we return to step 1 in this case? isn't this an irrecoverable state?

mhenriks: We do.  Next reconcile there will be no pvc and we do the same steps over again

alromeros: Nice! I thought that we only returned reconcile errors in irrecoverable states.

mhenriks: Nah, whenever reconcile returns an error it is retried with exponential backoff until reconcile returns no error

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -130,17 +116,41 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
...
+		targetNamespace = ar.Request.Namespace
+	}
+
+	if targetName == "" {
+		targetName = ar.Request.Name
+	}
+
+	cloneSourceHandler, err := newCloneSourceHandler(dataVolume, wh.cdiClient)
+	if err != nil {
+		if k8serrors.IsNotFound(err) && noTokenOkay {
```
arnongilboa, line 138: Why do we continue with the following code when `noTokenOkay` and not just return?

tests/static-volume_test.go
```diff
@@ -0,0 +1,158 @@
...
+			pvName = pvc.Spec.VolumeName
+			pvStorageClass = pvc.Spec.StorageClassName
+			pvMode = pvc.Spec.VolumeMode
+
+			By("Getting source MD5")
+			md5, err := f.GetMD5(f.Namespace, pvc, utils.DefaultImagePath, 0)
+			Expect(err).ToNot(HaveOccurred())
+			err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+			Expect(err).ToNot(HaveOccurred())
+			sourceMD5 = md5
```
akalenyu, line 85: Really not big deal but I think you can get rid of this and use the md5 const https://github.com/kubevirt/containerized-data-importer/blob/3788f70d086ca8e72afafea19cc3fdd6e3005d42/tests/utils/upload.go#L42-L43

pkg/apiserver/webhooks/datavolume-mutate.go
mhenriks: Want to add the token if possible.  Still have to handle initial provisioning case where DataVolume has static pv annotation but the pv does not exist

tests/static-volume_test.go
```diff
@@ -0,0 +1,158 @@
...
+				Name: "doesnotmatter",
+			}
+			return utils.NewDataVolumeWithExternalPopulation(dvName, dvSize, *pvStorageClass, *pvMode, nil, dataSourceRef)
+		}
+
+		BeforeEach(func() {
+			By("Creating source DV")
+			// source here shouldn't matter
+			dvDef := importDef()
+			controller.AddAnnotation(dvDef, controller.AnnDeleteAfterCompletion, "false")
```
arnongilboa, line 65: why GC needs to be disabled?

Issue #2561, opened 2023-01-27
CDI download fails due to nbdkit: curl error


k8scoder192:
**What happened**:
**CDI download of large image file (380M) fails with error**
```I0127 02:55:25.226441       1 nbdkit.go:290] Log line from nbdkit: nbdkit: curl[4]: error: pread: curl_easy_perform: Failed sending data to the peer: Connection died, tried 5 times before giving up
E0127 02:55:26.347182       1 prlimit.go:174] qemu-img failed output is:
E0127 02:55:26.347210       1 prlimit.go:175]     (0.00/100%)
    (1.02/100%)
    (2.09/100%)
    (3.16/100%)
    (4.22/100%)
    (5.22/100%)
    (6.26/100%)
    (7.32/100%)
    (8.33/100%)
    (9.33/100%)
    (10.39/100%)
    (11.39/100%)
    (12.39/100%)
    (13.39/100%)
    (14.40/100%)
    (15.40/100%)


E0127 02:55:26.347218       1 prlimit.go:176] qemu-img: error while reading at byte 379387904: Input/output error
```
**Describe on importer pod output**
```    Last State:     Terminated
      Reason:       Error
      Message:      Unable to process data: Unable to convert source data to target format: Conversion to Raw failed: could not convert image to raw Log line from nbdkit: nbdkit: curl[4]: error: pread: curl_easy_perform: Failed sending data to the peer: Connection died, tried 5
times before giving up: qemu-img execution failed: exit status 1
      Exit Code:    1
      Started:      Fri, 27 Jan 2023 15:40:40 -0500
      Finished:     Fri, 27 Jan 2023 15:45:06 -0500
    Ready:          True
    Restart Count:  1
    Limits:
      cpu:     4
      memory:  10Gi
    Requests:
      cpu:     1
      memory:  8Gi
    Environment:
      IMPORTER_SOURCE:               http
      IMPORTER_ENDPOINT:             https://cloud-images.ubuntu.com/bionic/current/bionic-server-cloudimg-amd64.img
      IMPORTER_CONTENTTYPE:          kubevirt
      IMPORTER_IMAGE_SIZE:           10Gi
      OWNER_UID:                     ffccd373-9db1-4712-9d4a-4ab7ff23e006
      FILESYSTEM_OVERHEAD:           0
      INSECURE_TLS:                  false
      IMPORTER_DISK_ID:
      IMPORTER_UUID:
      IMPORTER_READY_FILE:
[...]

PR #2584, opened 2023-02-13
Apply retry filter to ndkit conversions

awels:

Reduce the number of pod retries when there are failures in downloading images. Instead retry
inside the pod. This implements suggestion from https://github.com/kubevirt/containerized-data-importer/issues/2561.
Also fix some linter warnings in same package.

Fixes #

**Release note**:

```release-note
BugFix: use retry filter in inline streaming conversion with qcow2 http endpoints.
```

### DIFF
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -100,6 +100,8 @@ func NewNbdkitCurl(nbdkitPidFile, user, password, certDir, socket string, extraH
 	for _, header := range extraHeaders {
 		pluginArgs = append(pluginArgs, fmt.Sprintf("header=%s", header))
 	}
+	// Don't do exponential retry, the container restart will be exponential
+	pluginArgs = append(pluginArgs, "retry-exponential=no")
 	for _, header := range secretExtraHeaders {
 		redactArgs = append(redactArgs, fmt.Sprintf("header=%s", header))
 	}
@@ -111,6 +113,7 @@ func NewNbdkitCurl(nbdkitPidFile, user, password, certDir, socket string, extraH
 		redactArgs: redactArgs,
 		Socket:     socket,
 	}
+	// Should be last filter
 	n.AddFilter(NbdkitRetryFilter)
 	return n
 }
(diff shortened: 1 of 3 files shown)


pkg/image/nbdkit.go
```diff
@@ -107,15 +103,16 @@ func NewNbdkitCurl(nbdkitPidFile, user, password, certDir, socket string, extraH
...
-	return &Nbdkit{
+	n := &Nbdkit{
 		NbdPidFile: nbdkitPidFile,
 		plugin:     NbdkitCurlPlugin,
 		nbdkitArgs: args,
 		pluginArgs: pluginArgs,
 		redactArgs: redactArgs,
 		Socket:     socket,
 	}
+	n.AddFilter(NbdkitRetryFilter)
```
akalenyu, line 114: maybe add a note that this should be added last?

pkg/image/qemu_test.go
```diff
@@ -301,7 +301,7 @@ var _ = Describe("Create blank image", func() {
 		size := convertQuantityToQemuSize(quantity)
 		replaceExecFunction(mockExecFunction("", "", nil, "create", "-f", "raw", "image", size), func() {
 			err = CreateBlankImage("image", quantity, false)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("no such file or directory"))
```
akalenyu, line 304: hmm are we fine with this error happening? shouldn't we pass a tempdir to CreateBlankImage and keep test the same?

akalenyu:
Looking at the func test failure, maybe it is retrying on a 404? unless the file host returns something else

pkg/image/qemu_test.go
awels: Well that error is happening because the file we pass in is not real, and since we were ignoring the error before...

akalenyu: I think it may be worth to create a temp file and avoid the err, I think we do this in other qemu tests already

awels:
Yeah, we were retrying at the container level on a 404 to start with. Now it is in the filter. I fixed the broken test by disabling exponential retries on the filter level. So now the filter retry will fail in sufficient time for the test to see a container retry and restore the file to the http server.

pkg/image/qemu_test.go
akalenyu: Let me know if you want to do that

Issue #1980, opened 2021-10-08
Importer pod fails during disk image validation


mmellin:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
> /kind enhancement

**What happened**:

When creating a Datavolume with an HTTP source and `volumeMode: Block`, the importer pod fails to validate the image file, prematurely erroring out and complaining:

```
E1008 21:26:43.748945       1 prlimit.go:176] qemu-img: error while reading at byte 880803840: Input/output error
```

**What you expected to happen**:

I expect that the image validation occurs, the image is pulled/converted properly and the importer pod runs to completion successfully.

**How to reproduce it (as minimally and precisely as possible)**:

1) Start with a single node K8s cluster running 1.20 which a node that has a single extra empty disk (ie /dev/sdb).
2) Install rook/ceph 1.7 as per their docs, and create a Block resource / storageclass with a replication factor of 1.
```
$ git clone --single-branch --branch release-1.7 https://github.com/rook/rook.git
cd rook/cluster/examples/kubernetes/ceph
kubectl create -f crds.yaml -f common.yaml -f operator.yaml
kubectl create -f cluster.yaml


$ cat <<EOF | kubectl apply -f -
apiVersion: ceph.rook.io/v1
kind: CephBlockPool
metadata:
  name: replicapool
  namespace: rook-ceph
spec:
  failureDomain: host
  replicated:
    size: 1
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: rook-ceph-block
# Change "rook-ceph" provisioner prefix to match the operator namespace if needed
provisioner: rook-ceph.rbd.csi.ceph.com
parameters:
    # clusterID is the namespace where the rook cluster is running
    clusterID: rook-ceph
    # Ceph pool into which the RBD image shall be created
    pool: replicapool

    # (optional) mapOptions is a comma-separated list of map options.
    # For krbd options refer
    # https://docs.ceph.com/docs/master/man/8/rbd/#kernel-rbd-krbd-options
    # For nbd options refer
    # https://docs.ceph.com/docs/master/man/8/rbd-nbd/#options
    # mapOptions: lock_on_read,queue_depth=1024

    # (optional) unmapOptions is a comma-separated list of unmap options.
    # For krbd options refer
    # https://docs.ceph.com/docs/master/man/8/rbd/#kernel-rbd-krbd-options
[...]

PR #2588, opened 2023-02-14
Stop forcing preallocation on block imports

alromeros:

This pull request enables deactivating preallocation on block imports.

### The bug

Due to what seemed like a [bug](https://github.com/kubevirt/containerized-data-importer/issues/1980#issuecomment-940238434) in qemu-img, we started forcing preallocation on block imports as an arbitrary way to skip that specific error. This behavior was introduced in https://github.com/kubevirt/containerized-data-importer/pull/2001. After several updates in qemu and related dependencies, the bug is no longer reproducible. 

After some research, we've found a [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1872633)  that seems to solve an oddly similar case. The described bugs and fixes are:
-  A race in the kernel solved in commit https://github.com/torvalds/linux/commit/767630c63bb23acf022adb265574996ca39a4645  (v5.11).
- QEMU not being able to handle the previous bug. This was addressed here https://github.com/qemu/qemu/commit/ece4fa9152ea37c7ebd158af330e3b20e33cf385  (v5.2.0).
- " An optimization for grossly unaligned writes that qemu-img convert does".

As a last important note, CDI 1.40 (the one used in the original [issue](https://github.com/kubevirt/containerized-data-importer/issues/1980#issuecomment-940238434)) included qemu 5.1.0, the version where the bug described in the previous bugzilla was reproduced. 

### Conclusions

Knowing this, we can draw some conclusions:
- Always preallocating block imports is an arbitrary behavior that shouldn't be forced unless explicitly necessary.
- The original bug is not CDI-related. 
- Since the bug is no longer reproducible and several related fixes are present in our current qemu binaries, we can conclude that is safe to revert the behavior introduced in https://github.com/kubevirt/containerized-data-importer/pull/2001.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2168165

https://github.com/kubevirt/kubevirt/pull/6942 should also be reverted once this PR is merged.

**Release note**:

```release-note
Stop forcing preallocation on block imports
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -111,8 +111,6 @@ func main() {
 	volumeMode := v1.PersistentVolumeBlock
 	if _, err := os.Stat(common.WriteBlockPath); os.IsNotExist(err) {
 		volumeMode = v1.PersistentVolumeFilesystem
-	} else {
-		preallocation = true
 	}
 
 	// With writeback cache mode it's possible that the process will exit before all writes have been commited to storage.


mhenriks:
#1980 is no longer reproducible even with cdi v1.40.0?

alromeros:
> #1980 is no longer reproducible even with cdi v1.40.0?

Yesterday I spent a significant amount of time trying to reproduce it and I wasn't able o do it, neither with cdi v1.40 nor manually with older `qemu-img`.

akalenyu (commented):
Really great analysis!
One note, if this backports, we should make sure the target versions are paired with a qemu-img/kernel that contain the fixes

mhenriks:
> > #1980 is no longer reproducible even with cdi v1.40.0?
> 
> Yesterday I spent a significant amount of time trying to reproduce it and I wasn't able o do it, neither with cdi v1.40 nor manually with older `qemu-img`.

Yeah, I tried myself as well.  I have updated my fedora version since initially tested so maybe have new kernel

awels:
So are we okay with this at this point? If someone runs into it, we can basically say update your kernel. From that perspective I think we should be good

PR #2601, opened 2023-02-21
Fix DV verifyConditions test

arnongilboa:
`verifyConditions()` also checks now that all the test conditions exist

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/utils/datavolume.go
+++ b/tests/utils/datavolume.go
@@ -873,25 +873,27 @@ func WaitForConditions(ci ClientsIface, dataVolumeName, namespace string, timeou
 	}, timeout, pollingInterval).Should(gomega.BeTrue())
 }
 
-// verifyConditions checks if the conditions match testConditions
+// verifyConditions checks if all the conditions exist and match testConditions
 func verifyConditions(actualConditions []cdiv1.DataVolumeCondition, testConditions []*cdiv1.DataVolumeCondition) bool {
 	for _, condition := range testConditions {
-		if condition != nil {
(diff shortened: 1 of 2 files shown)


tests/utils/datavolume.go
```diff
@@ -873,25 +873,27 @@ func WaitForConditions(ci ClientsIface, dataVolumeName, namespace string, timeou
...
-				if !strings.Contains(actualCondition.Message, condition.Message) {
-					fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: Condition.Message does not match for type: %s, message expected: [%s],  message found: [%s]\n", condition.Type, condition.Message, actualCondition.Message)
-					return false
-				}
-			}
+		if condition == nil {
+			continue
+		}
+		actualCondition := dvc.FindConditionByType(condition.Type, actualConditions)
+		if actualCondition == nil {
```
akalenyu, line 883: So the test was passing because at some point `actualCondition==nil` was true for all 3 of them?

akalenyu (commented):
Still figuring this out please bear with me lol

tests/utils/datavolume.go
arnongilboa: Even worse - it passed whenever the DV **existing** conditions matched the passed `testConditions`, so if you have several `testConditions` and none of them exists in the DV, the test would pass, and we were actully testing nothing here...

akalenyu: Yeah I think that is what I meant `actualCondition==nil` means the DV had none of them at one point

PR #2604, opened 2023-02-22
Respect bind.immediate annotation by not attempting fancy clones

akalenyu:
`bind.immediate` is an annotation that says we should bypass "real" WFFC behavior by simply starting our worker pods
wherever.
Prior to this, we were only honoring it for import/upload/host-assisted clone, but we should definitely do the same for
opt out of other fancy clone methods when it is specified:
- CSI clone
- Smart clone
- Clone from snapshot source

Fixes #

**Release note**:

```release-note
BugFix: Respect bind.immediate annotation on fancy clones by falling back to host assisted
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -45,6 +45,7 @@ import (
 	cdiv1utils "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/utils"
 	"kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned/scheme"
 	"kubevirt.io/containerized-data-importer/pkg/common"
+	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
 	"kubevirt.io/containerized-data-importer/pkg/token"
 	"kubevirt.io/containerized-data-importer/pkg/util"
 	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/api"
@@ -203,6 +204,9 @@ const (
 	AnnPodSidecarInjection = "sidecar.istio.io/inject"
 	// AnnPodSidecarInjectionDefault is the default value passed for AnnPodSidecarInjection
 	AnnPodSidecarInjectionDefault = "false"
+
+	// AnnImmediateBinding provides a const to indicate whether immediate binding should be performed on the PV (overrides global config)
+	AnnImmediateBinding = AnnAPIGroup + "/storage.bind.immediate.requested"
 
 	// CloneUniqueID is used as a special label to be used when we search for the pod
 	CloneUniqueID = "cdi.kubevirt.io/storage.clone.cloneUniqeId"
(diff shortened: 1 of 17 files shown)


pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -450,6 +450,16 @@ func (r *PvcCloneReconciler) selectCloneStrategy(datavolume *cdiv1.DataVolume, p
...
 		return NoClone, err
 	}
+	if bindingMode != nil && *bindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
+		_, isImmediateBindingRequested := datavolume.Annotations[cc.AnnImmediateBinding]
+		waitForFirstConsumerEnabled, err := cc.IsWaitForFirstConsumerEnabled(isImmediateBindingRequested, r.featureGates)
+		if err != nil {
+			return NoClone, err
+		}
+		if !waitForFirstConsumerEnabled {
+			return HostAssistedClone, nil
```
mhenriks, line 460: The title of this PR is "Respect bind.immediate annotation on fancy clones" but is this not turning a "fancy" clone into a "dumb" clone?  Kind of a misnomer if you ask me.

My expectation was that a dummy pod would consume either:

1.  the temp pvc before namespace transfer
2. the target pvc for same namespace clones

Obviously, there are tradeoffs for either solution, and I'm not suggesting which way is better, but it seems to me that the messaging is off

mhenriks (commented):
I came into this expecting something else, not saying we shouldn't do what is implemented here, but lets's talk about it and maybe fix the messaging

pkg/controller/datavolume/pvc-clone-controller.go
akalenyu: Yep agree the messaging is confusing, will adjust

pkg/controller/common/util.go
```diff
@@ -1177,3 +1181,15 @@ func GetCloneSourceNameAndNamespace(dv *cdiv1.DataVolume) (name, namespace strin
 
 	return sourceName, sourceNamespace
 }
+
+// IsWaitForFirstConsumerEnabled tells us if we should respect "real" WFFC behavior or just let our worker pods randomly spawn
+func IsWaitForFirstConsumerEnabled(isImmediateBindingRequested bool, gates featuregates.FeatureGates) (bool, error) {
```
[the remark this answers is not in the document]
akalenyu: Done

awels (commented):
So I don't have an issue with the PR, I am just wondering what the problem is with creating a dummy pod that triggers the fancy clone to complete if we detect this situation instead of falling back to host assisted clone. We could re-use the resize pod for instance to trigger this. So basically always call the resize pod if we have detected this situation.

akalenyu:
> So I don't have an issue with the PR, I am just wondering what the problem is with creating a dummy pod that triggers the fancy clone to complete if we detect this situation instead of falling back to host assisted clone. We could re-use the resize pod for instance to trigger this. So basically always call the resize pod if we have detected this situation.

Hmm things like:
- Backportability
- Another pod lifecycle to manage
- How common or encouraged is this path in the first place? Under what circumstances would you
want to cheat WFFC on a clone?

awels:
> Hmm things like:
> 
>     * Backportability
If you re-use the resize pod I think that would actually be less code than we have here. Essentially right now we are bailing when we find a WFFC restore and nothing happens. If we just continue, we should get to the resize pod, which will trigger the actual restore.
>     * Another pod lifecycle to manage
Not really, see above comment about the resize pod.
> 
>     * How common or encouraged is this path in the first place? Under what circumstances would you
>       want to cheat WFFC on a clone?
So the first thing that comes to mind is running on many/most of the cloud providers. All their storage classes are WFFC and if we want to clone (we may or may not want that) it would make sense to do that using the best method available.

awels:
So I found many people that don't understand how WFFC storage works, and they discover the annotation will simply slap it on everything. Like I said I have no issues with the PR as is. I just wanted to think about how it would work if we just triggered the clone with a random pod. Or the resize pod in particular since we already manage it.

PR #2618, opened 2023-03-05
Simplify upgrade testing, and only test v1beta1 DVs

maya-r:
Newer releases of CDI no longer contain v1alpha1 support, so we can't blindly apply v1alpha1 DVs. Any old version DV is good enough, nothing specific to v1alpha1, so let's remove that variant.

While here, make the DVs for upgrade testing always contain the annotation to not be garbage collected.
This helps simplify the cluster-sync logic.

**Release note**:

```release-note
NONE
```


maya-r:
I'm blindly using the CI for testing this change, as it's fairly time-consuming to do it locally.

PR #2624, opened 2023-03-07
Avoid CI downgrading from latest release to cherrypick PR

akalenyu:

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
> So essentially we are disabling upgrade testing from `y - 1` if we are on a release branch?

In most release branches this is a downgrade but yes currently for 1.56 we would be disabling it

> So essentially we are disabling upgrade testing from `y - 1` if we are on a release branch?

PTAL, now trying to grab the previous y release

PR #2633, opened 2023-03-09
Follow up for WFFC test suite adjustments

akalenyu:
Previous PR does not make our test suite completely WFFC-friendly,
as it turns out more follow-ups were needed.
This continues the work in https://github.com/kubevirt/containerized-data-importer/pull/2610

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/pvc.go
+++ b/tests/framework/pvc.go
@@ -370,7 +370,7 @@ func (f *Framework) NewPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolu
 		Spec: k8sv1.PodSpec{
 			// this may be causing an issue
 			TerminationGracePeriodSeconds: &[]int64{10}[0],
-			RestartPolicy:                 k8sv1.RestartPolicyNever,
+			RestartPolicy:                 k8sv1.RestartPolicyOnFailure,
 			Containers: []k8sv1.Container{
 				{
 					Name:    "runner",
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -490,6 +490,7 @@ var _ = Describe("all clone tests", func() {
 				Expect(err).ToNot(HaveOccurred())
 				targetPvc, err := utils.WaitForPVC(f.K8sClient, targetDataVolume.Namespace, targetDataVolume.Name)
 				Expect(err).ToNot(HaveOccurred())
+				f.ForceBindPvcIfDvIsWaitForFirstConsumer(targetDataVolume)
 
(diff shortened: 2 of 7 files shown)


alromeros (commented):
Nice!
/lgtm

ShellyKa13 (commented):
/lgtm

tests/framework/pvc.go
```diff
@@ -370,7 +370,6 @@ func (f *Framework) NewPodWithPVC(podName, cmd string, pvc *k8sv1.PersistentVolu
 		Spec: k8sv1.PodSpec{
 			// this may be causing an issue
 			TerminationGracePeriodSeconds: &[]int64{10}[0],
-			RestartPolicy:                 k8sv1.RestartPolicyNever,
```
alromeros, line 373: Any reason to remove this?

akalenyu: Yeah, I am trying to find out why we have it in the first place, seems to date back 5+ years
basically means verifier won't attempt to restart on transient errs

I have some tests where the verifier pod regularly hits an error without much context,
and does not recover from it

akalenyu: Changed to OnFailure so it only restarts if it exits with a return code other than 0
(default is Always - pod must always be running)

akalenyu:
> /hold
> 
> testing the changes

Just a clarification, I am testing this change against an external cluster with LVM Storage

tests/framework/pvc.go
alromeros: I think it makes perfect sense to change it to `RestartPolicyOnFailure`, but like you, I lack some context on why we had `RestartPolicyNever` before. If it's not breaking anything I'm for the change.

awels: What transient errors are we thinking. we would like to know if the verifier pod errors (like maybe the file we are looking for doesn't exist). This would simply restart the failed command, which may or may not fail again depending on if the error is transient or a real issue. If we do this, we need to make sure there are some time limits on the commands we run in the verifier pod.

akalenyu: I am hitting errors in container startup (not in executing commands on it)
Seems to be related to quick recreation of the deleted verifier pod, but not sure, error does not say much.

Is there a reason to go with `RestartPolicyNever`?

awels: We can try with the on failure policy, we will know if something breaks due to it, there is an overall suite timeout anyway.

PR #2636, opened 2023-03-13
Move pkg/clone/auth to API lib and introduce single dv.Authorize()

### LINKED ISSUE #2552: Unify auth in a single AuthorizeDataVolume
Currently, operators utilizing CDI clone API need to "know a lot" about CDI clones
to do auth properly.
Kubevirt does this, for example:
https://github.com/kubevirt/kubevirt/blob/2baf5303dddfe71c98a2510357c464a8f7bdd3ca/pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go#L214-L275
(DataSource GET, Spec.PVC lookup)

We can try to unify auth in a single AuthorizeDataVolume:

- For DataSource GETs., we need to do something similar to https://github.com/kubevirt/kubevirt/blob/2baf5303dddfe71c98a2510357c464a8f7bdd3ca/pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go#L61-L67
- The DV should be passed to us so we could easily look up Spec.PVC
- For sources other than Source.PVC AuthorizeDataVolume is a no-op?

**Additional context**:

akalenyu:
This way other projects don't have to vendor in CDI, the API lib will be enough.
While it's not a huge deal at this point, each existing dependency is an entry door for pulling in more stuff in the future by accident.
Also, these are annoying https://github.com/kubevirt/containerized-data-importer/blob/20dc7d438af11916bff3d7cdf4588b827d4d27e6/go.mod#L26-L27

Fixes #2552 

**Release note**:

```release-note
Move clone auth helpers to CDI API library
```


pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -33,8 +33,8 @@ import (
 	"k8s.io/klog/v2"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	cdiv1cloneauth "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/clone"
```
mhenriks, line 36: I think `kubevirt.io/containerized-data-importer-api/pkg/clone` would be better.  Let's keep `pkg/apis` for kubernetes apis

akalenyu: So we already have https://github.com/kubevirt/containerized-data-importer/blob/be1add08d51ebba2f62538afe6d8c13db952d712/pkg/controller/common/util.go#L45

akalenyu: makes sense, was about to do this, but then realized we have https://github.com/kubevirt/containerized-data-importer/issues/2552 which will accept a DV

mhenriks: okay then maybe it makes to refactor to do something like:

```golang
ok, err := dv.Authorize()
```

akalenyu: so `Authorize()` will go in ${version}/utils, and we will be able to have different authorization logic for different version?
I am just making sure, I have no issue with moving to `kubevirt.io/containerized-data-importer-api/pkg/clone`

mhenriks: I guess what I'm saying is do #2552 in this pr and have `Authorize` be a DataVolume member function.

akalenyu: Posted a first draft just to make sure we are fine with the direction
- Still missing authorization via SA
- Feels like the logic is clunky, any pointers are greatly appreciated
- Would aggregating the proxies into one interface make sense?

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize.go
```diff
@@ -0,0 +1,98 @@
...
+	// AnnCheckStaticVolume checks if a statically allocated PV exists before creating the target PVC.
+	// If so, PVC is still created but population is skipped
+	AnnCheckStaticVolume = core.GroupName + "/storage.checkStaticVolume"
+)
+
+var ErrNoTokenOkay = errors.New("proceeding without token is okay under the circumstances")
+
+// Authorize indicates if the creating context is authorized to create the data volume
+// For sources other than clone (import/upload/etc), this is a no-op
+func (dv *DataVolume) Authorize(requestNamespace, requestName string, sarClient SubjectAccessReviewsProxy, nsClient NamespaceProxy, dsClient DataSourceProxy, userInfo authentication.UserInfo) (bool, string, cloneSourceHandler, error) {
```
mhenriks, line 41: Maybe use the client native interfaces instead of our own proxies, so `SubjectAccessReviewsGetter`, `NamespacesGetter`, `DataSourcesGetter`?

Is that an improvement?

mhenriks (commented):
Great start!  I'm not that down on this

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize.go
mhenriks: Scratch that, can't have DataSourcesGetter in here (circular dep).  But maybe the others?  Or will that make the deps situation worse?

akalenyu: Introduced a single `type AuthorizationHelperProxy interface`, hope that is fine

> I'm not that down on this
I am not sure if you meant you're not down with the approach or don't hate it too much

staging/src/kubevirt.io/containerized-data-importer-api/go.mod
```diff
@@ -3,9 +3,11 @@ module kubevirt.io/containerized-data-importer-api
 go 1.19
 
 require (
+	github.com/kubernetes-csi/external-snapshotter/client/v6 v6.0.1
```
mhenriks, line 6: Can we get rid of this somehow?

mhenriks: Let's keep this go.mod lean and mean.  Copying string consts is fine

pkg/clone/auth.go
akalenyu, line 1: @mhenriks do you think it's fine to delete this? should we keep for backwards compatibility?

staging/src/kubevirt.io/containerized-data-importer-api/go.mod
akalenyu: done

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize_utils.go
```diff
@@ -0,0 +1,341 @@
...
+	SACloneAuthFunc   ServiceAccountCloneAuthFunc
+	SourceName        string
+	SourceNamespace   string
+}
+
+type createSarFunc func(*authorization.SubjectAccessReview) (*authorization.SubjectAccessReview, error)
+type dsGetFunc func(string, string) (*DataSource, error)
+
+// AuthorizationHelperProxy proxies calls to APIs used for DV authorization
+type AuthorizationHelperProxy interface {
```
mhenriks, line 120: Can't believe I just thought of this and I'd be remiss to not bring it up, what if our api just takes a `dynamic.Interface` instead of the proxy?

PRO:  
Less code for clients

CON: 
more code in api (translating to/from unstructured)
adding `client-go` to go.mod

Don't care too much either way just want to get your opinion

akalenyu: Would have to look up some references but the immediate concern is:
Would we still be able to restrict access similarly to how we do with proxy?
I mean, with the proxy, external entities know that CDI doesn't have access to the client itself other than
the implemented Get/Create methods

pkg/clone/auth.go
akalenyu: gentle ping

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize_utils.go
mhenriks: With the proxy, callers know exactly what calls we are making, and that may be nice, but ultimately there is RBAC in place to keep the SA from doing anything bad.  And obviously the code is available to audit and make sure we're not reading secrets or whatever.

akalenyu: Yes my concern is that we get passed a general SA that is probably not very limited,
unless you are suggesting to somehow reduce the SA

mhenriks: No I don't think it is practical to create a special client with reduced permissions just to pass to this api.  I also don't think passing the dynamic client is a security concern.

akalenyu: > I also don't think passing the dynamic client is a security concern.

I thought that is the reason we went with proxy structs in the first place

mhenriks: > > I also don't think passing the dynamic client is a security concern.
> 
> I thought the reason we went with proxy structs in the first place was to avoid passing a full-blown client

The issue is with the typed client here:  https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/client/clientset/versioned/typed/core/v1beta1/datavolume.go#L40

That would cause a circular dependency

akalenyu: Ahh right.

> I thought the reason we went with proxy structs in the first place was to avoid passing a full-blown client

We used the sarProxy prior to this PR though (and thus prior to the circular dep potential), was this^ not the main concern?

Anyway my (not 100% trustworthy) gut is telling me that whipping out a dynamic client
when we have a defined set of actions on objects we know about is not right, so I lean towards current impl

mhenriks: All past decisions are not made with perfect information.  If I could go back in time I would probably use the dynamic client.  But if you like using this interface, I don't care enough to keep belaboring the discussion.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize.go
```diff
@@ -0,0 +1,162 @@
...
+	AnnCheckStaticVolume = core.GroupName + "/storage.checkStaticVolume"
+)
+
+// ErrNoTokenOkay indicates proceeding without token is allowed
+// This error should only be of interest to entities that give out DataVolume tokens
+var ErrNoTokenOkay = errors.New("proceeding without token is okay under the circumstances")
+
+// AuthorizeUser indicates if the creating user is authorized to create the data volume
+// For sources other than clone (import/upload/etc), this is a no-op
+func (dv *DataVolume) AuthorizeUser(requestNamespace, requestName string, proxy AuthorizationHelperProxy, userInfo authentication.UserInfo) (bool, string, CloneSourceHandler, error) {
```
mhenriks, line 43: Rather than returning four vales, how about a new struct that combines the first three?

akalenyu: Yep makes sense, added
My biggest worry is still this though:
https://github.com/kubevirt/containerized-data-importer/pull/2636#discussion_r1165542038

mhenriks: I think it's fine to delete auth.go.  I seriously doubt any project other than kubevirt is importing it

akalenyu: awesome

tests/webhook_test.go
```diff
@@ -12,22 +12,31 @@ import (
...
-	"kubevirt.io/containerized-data-importer/pkg/clone"
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	cdiclientset "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"kubevirt.io/containerized-data-importer/tests/framework"
 	"kubevirt.io/containerized-data-importer/tests/utils"
 )
 
-type sarProxy struct {
-	client kubernetes.Interface
+type authProxy struct {
```
arnongilboa, line 25: why not using authProxy from datavolume-mutate.go?

akalenyu: I think it's nice that webhook_test.go will showcase all building blocks toward authorizing a data volume
plus, it feels a little "off" to export that struct from datavolume-mutate since it's not really of use to anyone but the tests

PR #2637, opened 2023-03-13
Adapt operator_test to cover OpenShift

arnongilboa:
Until recently, operator_test destructive tests were executed in CDI CI k8s lane, but they were rarely executed D/S on OpenShift cluster. To allow executing it D/S on a regular basis, as a follow-up to #2577, we fix a few test issues encountered on OpenShift cluster:
-Fix affinity to use a worker node instead of the first one
-Check only the storage-labeled pods, instead of all the pods in cdi namespace
-Check CDI UninstallStrategy to decide wheter to delete DVs beore deleting CDI
 
Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/framework/nodeplacement.go
+++ b/tests/framework/nodeplacement.go
@@ -22,13 +22,22 @@ var (
 // The values chosen are valid, but the pod will likely not be schedulable.
 func (f *Framework) TestNodePlacementValues() sdkapi.NodePlacement {
 	nodes, _ := f.K8sClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+
+	nodeName := nodes.Items[0].Name
+	for _, node := range nodes.Items {
+		if _, hasLabel := node.Labels["node-role.kubernetes.io/worker"]; hasLabel {
+			nodeName = node.Name
+			break
+		}
+	}
+
 	affinityTestValue = &v1.Affinity{
 		NodeAffinity: &v1.NodeAffinity{
 			RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
 				NodeSelectorTerms: []v1.NodeSelectorTerm{
 					{
 						MatchExpressions: []v1.NodeSelectorRequirement{
-							{Key: "kubernetes.io/hostname", Operator: v1.NodeSelectorOpIn, Values: []string{nodes.Items[0].Name}},
+							{Key: "kubernetes.io/hostname", Operator: v1.NodeSelectorOpIn, Values: []string{nodeName}},
 						},
 					},
 				},
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -274,10 +274,7 @@ var _ = Describe("ALL Operator tests", func() {
 				Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")
 				Expect(err).ToNot(HaveOccurred())
 
-				cdiPods, err = f.K8sClient.CoreV1().Pods(f.CdiInstallNs).List(context.TODO(), metav1.ListOptions{})
-
-				Expect(err).ToNot(HaveOccurred(), "failed listing cdi pods")
-				Expect(len(cdiPods.Items)).To(BeNumerically(">", 0), "no cdi pods found")
+				cdiPods = getCDIPods(f)
 			})
 
 			AfterEach(func() {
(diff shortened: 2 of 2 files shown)


tests/framework/nodeplacement.go
```diff
@@ -22,18 +22,29 @@ var (
...
+
+	nodeIdx := -1
+	for i, node := range nodes.Items {
+		if _, hasLabel := node.Labels["node-role.kubernetes.io/worker"]; hasLabel {
+			nodeIdx = i
+			break
+		}
+	}
+
+	if nodeIdx > -1 {
```
alromeros, line 34: Are we ok with using empty `affinityTestValue` when nodeIdx is -1?

tests/operator_test.go
```diff
@@ -1236,6 +1236,8 @@ var _ = Describe("ALL Operator tests", func() {
 				serverSecretNames := []string{"cdi-apiserver-server-cert", "cdi-uploadproxy-server-cert"}
 
 				ts := time.Now()
+				// Time comparison here is in seconds, so make sure there is an interval
+				time.Sleep(time.Second)
```
alromeros, line 1240: Nice!

tests/operator_test.go
```diff
@@ -420,7 +415,7 @@ var _ = Describe("ALL Operator tests", func() {
 			})
 
 			It("[test_id:4986]should remove/install CDI a number of times successfully", func() {
-				for i := 0; i < 10; i++ {
+				for i := 0; i < 5; i++ {
```
alromeros, line 418: Any reason to change this and the eventually timeout?

tests/framework/nodeplacement.go
```diff
@@ -22,18 +22,29 @@ var (
...
 	nodes, _ := f.K8sClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
-	affinityTestValue = &v1.Affinity{
-		NodeAffinity: &v1.NodeAffinity{
-			RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
-				NodeSelectorTerms: []v1.NodeSelectorTerm{
-					{
-						MatchExpressions: []v1.NodeSelectorRequirement{
-							{Key: "kubernetes.io/hostname", Operator: v1.NodeSelectorOpIn, Values: []string{nodes.Items[0].Name}},
+
+	nodeIdx := -1
```
awels, line 26: Since we are really just interested in the node name, why not just store it? that way you don't have to keep track of the index, and you can initialize it with something like this:
```go
nodeName := nodes.Items[0].Name
for _, node := range nodes.Items {
  if _, hasLabel := node.Labels["node-role.kubernetes.io/worker"]; hasLabel {
    nodeName = node.Name
    break
  }
}
```
and then you can use the nodeName in the selector.

tests/operator_test.go
awels: So this failed if it was too fast? < 1 second?

tests/operator_test.go
```diff
@@ -1400,6 +1404,9 @@ var _ = Describe("ALL Operator tests", func() {
 			})
 
 			It("should use openshift priority class if not set and available", func() {
+				if utils.IsOpenshift(f.K8sClient) {
```
awels, line 1407: Shouldn't this skip be in the previous test?

tests/framework/nodeplacement.go
awels: Or if you don't want to initalize it, leave it blank and check if it is set after the loop

tests/operator_test.go
arnongilboa: yep

tests/operator_test.go
arnongilboa: I see no reason to make it sooo long. 5 iters seems more than enough I guess?

tests/framework/nodeplacement.go
arnongilboa: sure, will fix.

tests/framework/nodeplacement.go
arnongilboa: fixed, just for consistency.

tests/operator_test.go
arnongilboa: No, in this test OpenShift already has the `"kubevirt-cluster-critical" PriorityClass`, so I see no reason to use/update it as it won't really test anything. In the previous test we are setting CDI-specific `"system-cluster-critical" PriorityClass`, and test it's passed to the pods, so I think it should not be skipped in OpenShift.

PR #2640, opened 2023-03-14
Update git config to mark rsync directory as safe

awels:
In order to avoid
```
error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
```
we add `git config --global --add safe.directory /root/go/src/kubevirt.io/containerized-data-importer` in the builder container to tell git that the ownership not matching is not a problem and the rsync can work normally

buildvcs adds version information to the built binaries, would be lost if we directly add `-buildvcs=false` to the build command.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/bazel-docker.sh
+++ b/hack/build/bazel-docker.sh
@@ -48,6 +48,8 @@ fi
 echo "Making sure output directory exists..."
 ${CDI_CRI} run -v "${BUILDER_VOLUME}:/root:rw,z" --security-opt label=disable $DISABLE_SECCOMP --rm --entrypoint "/entrypoint-bazel.sh" ${BUILDER_IMAGE} mkdir -p /root/go/src/kubevirt.io/containerized-data-importer/_out
 
+${CDI_CRI} run -v "${BUILDER_VOLUME}:/root:rw,z" --security-opt label=disable $DISABLE_SECCOMP --rm --entrypoint "/entrypoint-bazel.sh" ${BUILDER_IMAGE} git config --global --add safe.directory /root/go/src/kubevirt.io/containerized-data-importer
+
 echo "Starting rsyncd"
 # Start an rsyncd instance and make sure it gets stopped after the script exits
 RSYNC_CID_CDI=$(${CDI_CRI} run -d -v "${BUILDER_VOLUME}:/root:rw,z" --security-opt label=disable $DISABLE_SECCOMP --cap-add SYS_CHROOT --expose 873 -P --entrypoint "/entrypoint-bazel.sh" ${BUILDER_IMAGE} /usr/bin/rsync --no-detach --daemon --verbose)


awels:
yeah good point, less changes.

mhenriks:
Quick explanation why this is preferable to `-buildvcs=false` would be great

akalenyu:
> Quick explanation why this is preferable to `-buildvcs=false` would be great

I think we would just be masking the original issue that git is broken inside the builder
```bash
[root@... containerized-data-importer]# git status
fatal: detected dubious ownership in repository at '/root/go/src/kubevirt.io/containerized-data-importer'
To add an exception for this directory, call:

        git config --global --add safe.directory /root/go/src/kubevirt.io/containerized-data-importer
```

awels:
So the way I read buildcvs is that when compiling some version information is embedded in the binary. If we disable that, that information is not there or incorrect. And we probably want that information to be correct. I don't think it would be a huge deal, if we didn't have it.

mhenriks:
I am running docker and this change did not fix the issue for me:

```
➜  containerized-data-importer git:(main) ✗ make all
./hack/build/bazel-docker.sh "DOCKER_PREFIX= DOCKER_TAG= VERBOSITY= PULL_POLICY= CR_NAME= CDI_NAMESPACE= ./hack/build/build-manifests.sh"
CDI_CRI: docker, CDI_CONTAINER_BUILDCMD: docker
Making sure output directory exists...
go version go1.19.5 linux/amd64
Starting rsyncd

Rsyncing /home/shared/sandbox/go/src/kubevirt.io/containerized-data-importer to container
187d8783baea4b8fa467e55051ebf0239903b9b3673f6d4e3166e538510f5717
Starting bazel server
go version go1.19.5 linux/amd64
CDI_CRI: , CDI_CONTAINER_BUILDCMD: buildah
error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:98: manifests] Error 1
➜  containerized-data-importer git:(main) ✗
➜  containerized-data-importer git:(main) ✗
➜  containerized-data-importer git:(main) ✗ git diff
diff --git a/hack/build/bazel-docker.sh b/hack/build/bazel-docker.sh
index a5a4620c1..ef298e092 100755
--- a/hack/build/bazel-docker.sh
+++ b/hack/build/bazel-docker.sh
@@ -81,7 +81,7 @@ printf "\n"
 rsynch_fail_count=0

 _rsync() {
-    rsync -al "$@"
+    rsync -al --no-owner "$@"
 }

 echo "Rsyncing ${CDI_DIR} to container"
```

mhenriks:
I assume that deleting the docker volume will solve the issue but I wonder if there's not another less disruptive fix

awels:
Well the problem is the owner inside of the container, it is running as root in docker, so the owner should be root, that is what the --no-owner does it copies the files and makes the owner the right user inside the container, and then copying back it is the original owner again. You can fix it by execing into the container and running `git config --global --add safe.directory /root/go/src/kubevirt.io/containerized-data-importer` inside of it. That will be saved in the volume until it is removed. At which point this fix kicks in.

mhenriks:
> Well the problem is the owner inside of the container, it is running as root in docker, so the owner should be root, that is what the --no-owner does it copies the files and makes the owner the right user inside the container, and then copying back it is the original owner again. You can fix it by execing into the container and running `git config --global --add safe.directory /root/go/src/kubevirt.io/containerized-data-importer` inside of it. That will be saved in the volume until it is removed. At which point this fix kicks in.

Why not have bazel-docker exec that command then?

mhenriks:
basically, I think we should have a fix that does not require an user intervention other than a git pull.  

Changing this line is one way to do it I think.  https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23

akalenyu:
> basically, I think we should have a fix that does not require an user intervention other than a git pull.
> 
> Changing this line is one way to do it I think. https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23
> 
> Maybe execing that git command is another

Thing is, if we don't fix `git` inside the builder container we may run into issues

mhenriks:
> > basically, I think we should have a fix that does not require an user intervention other than a git pull.
> > Changing this line is one way to do it I think. https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23
> > Maybe execing that git command is another
> 
> Thing is, if we don't fix `git` inside the builder container we may run into other issues

I don't care if we change the arg to `go build` or if we do something with git, all I care about is that developers/users of this repo don't have to do anything manually after updating their sandbox

btw, I was able to `make cluster-up cluster-sync test-functional` after only updating https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23

So I don't think there will be any other issues

akalenyu:
> > > basically, I think we should have a fix that does not require an user intervention other than a git pull.
> > > Changing this line is one way to do it I think. https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23
> > > Maybe execing that git command is another
> > 
> > 
> > Thing is, if we don't fix `git` inside the builder container we may run into other issues
> 
> I don't care if we change the arg to `go build` or if we do something with git, all I care about is that developers/users of this repo don't have to do anything manually after updating their sandbox
> 
> btw, I was able to `make cluster-up cluster-sync test-functional` after only updating https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23
> 
> So I don't think there will be any other issues

Yeah I couldn't name anything that would suffer from git being broken in the builder, there may not be any issues because of that

awels:
So you basically ran the command on that line in the builder, and that fixed it? I am a little confused about what we our options are. Obviously what we have here requires a rebuild of the volume which is undesirable since you are rebuilding everything.

mhenriks:
> So you basically ran the command on that line in the builder, and that fixed it? I am a little confused about what we our options are. Obviously what we have here requires a rebuild of the volume which is undesirable since you are rebuilding everything.

@awels The only thing I've tried is adding `-buildvcs=false` here:  https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/build-manifests.sh#L23

Everything seemed to work after that.

@akalenyu suggests running `git config --global --add safe.directory /root/go/src/kubevirt.io/containerized-data-importer` in the build container.  If this is preferred for some reason, I propose it is added to `bazel-docker.sh`.  Users should not have to execute it manually.

hack/build/bazel-docker.sh
```diff
@@ -81,7 +84,7 @@ printf "\n"
 rsynch_fail_count=0
 
 _rsync() {
-    rsync -al "$@"
+    rsync -al --no-owner "$@"
```
mhenriks, line 87: no need to do this and `git config --global --add safe.directory`. I say we hold off until there is a real need

mhenriks: The invocation of `git config` can be optimized  by running it only after the volume is created.  But I think as-is is fine.

I don't see why the `--no-owner` state is needed or desired now or in the future.  Please explain why it is.  My fear is that it may have unintended consequences and why take on that risk?

PR #2575, opened 2023-02-06
Ensure host assisted concurrent clone test runs on all storages

akalenyu:
Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

**What this PR does / why we need it**:
Also, make some changes to the test to catch
https://github.com/kubevirt/containerized-data-importer/issues/2550

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

Issue #2550, opened 2023-01-22
Errors when doing concurrent host assisted clones


akalenyu:
**What happened**:
Host-assisted clone source pods keep erroring when doing concurrent dumb clones.
With 2 concurrent clones it seems to converge slowly to success.

**What you expected to happen**:
restarts are mostly edge case on host assisted clones

**How to reproduce it (as minimally and precisely as possible)**:
Patch storage profile to do host assisted cloning
```bash
# k get storageprofile rook-ceph-block -o yaml
selecting docker as container runtime
apiVersion: cdi.kubevirt.io/v1beta1
kind: StorageProfile
metadata:
...
  name: rook-ceph-block
spec:
  claimPropertySets:
  - accessModes:
    - ReadWriteOnce
    volumeMode: Filesystem
  cloneStrategy: copy
status:
  claimPropertySets:
  - accessModes:
    - ReadWriteOnce
    volumeMode: Filesystem
  cloneStrategy: copy
  provisioner: rook-ceph.rbd.csi.ceph.com
  storageClass: rook-ceph-block
```
source dv yaml
```yaml
apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: golden-snap-source
  namespace: golden-ns
spec:
  source:
      http:
         url: http://.../Fedora-Cloud-Base-34-1.2.x86_64.qcow2
  pvc:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 8Gi
```
run concurrent clones
```bash
#!/bin/bash

for (( c=0; c<2; c++ ))
do
  cat << __EOF__ | ./cluster-up/kubectl.sh create -f -
---
apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: cloned-datavolume-${c}
  namespace: default
spec:
  source:
    pvc:
      namespace: golden-ns
      name: golden-snap-source
  storage:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 9Gi
__EOF__
done
```

**Additional context**:
Errors:
```bash
I0122 14:58:03.462143      10 prometheus.go:72] 25.53
F0122 14:58:03.536422      10 clone-source.go:41] Subprocess did not execute successfully, result is: '\x02'
disk.img
/usr/bin/tar: disk.img: Read error at byte 1119744, while reading 512 bytes: Permission denied
[...]

PR #2647, opened 2023-03-19
Set pod affinity for host assisted clone

### LINKED ISSUE #2576: Set pod affinity for host assisted clone source pods
Basically, the issue in https://github.com/kubernetes/kubernetes/issues/103305 applies to CDI
in the concurrent host assisted clone use case, on storage that is not topology constrained (globally accessible from all nodes).
Consider the scenario:
- Initiate 2 concurrent host assisted clone from source-pvc that was provisioned by rook-ceph-block with `ReadWriteOnce`
- It is entirely possible that each of the pods lands on a different node, and thus one will remain pending

The suggested solution in https://github.com/hetznercloud/csi-driver/issues/319#issuecomment-1324052836

**Additional context**:
This is not a problem for topology constrained storage like HPP because scheduler will ensure
the pods land where the PV is (`volume.kubernetes.io/selected-node` ann)

ido106:

Pods using the same PVC are scheduled on different nodes. We expect them to be scheduled on the same K8S node but they don't and thus cannot be attached to the PersistentVolume. The solution is to set [pod affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity).

Fixes #2576

**Release note**:

```release-note
Set pod affinity for host assisted clone source pods, in order to schedule them on the same K8S node.
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -57,6 +57,8 @@ const (
 
 	cloneSourcePodFinalizer = "cdi.kubevirt.io/cloneSource"
 
+	hostAssistedCloneSource = "cdi.kubevirt.io/hostAssistedSourcePodCloneSource"
+
 	uploadClientCertDuration = 365 * 24 * time.Hour
 )
 
@@ -477,7 +479,7 @@ func (r *CloneReconciler) cleanup(pvc *corev1.PersistentVolumeClaim, log logr.Lo
 
 // CreateCloneSourcePod creates our cloning src pod which will be used for out of band cloning to read the contents of the src PVC
 func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *corev1.PersistentVolumeClaim, log logr.Logger) (*corev1.Pod, error) {
-	exists, sourcePvcNamespace, sourcePvcName := ParseCloneRequestAnnotation(pvc)
+	exists, _, _ := ParseCloneRequestAnnotation(pvc)
 	if !exists {
 		return nil, errors.Errorf("bad CloneRequest Annotation")
 	}
@@ -519,7 +521,7 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *co
 		sourceVolumeMode = corev1.PersistentVolumeFilesystem
 	}
 
-	pod := MakeCloneSourcePodSpec(sourceVolumeMode, image, pullPolicy, imagePullSecrets, sourcePvcName, sourcePvcNamespace, ownerKey, serverCABundle, pvc, podResourceRequirements, workloadNodePlacement)
+	pod := MakeCloneSourcePodSpec(sourceVolumeMode, image, pullPolicy, ownerKey, imagePullSecrets, serverCABundle, pvc, sourcePvc, podResourceRequirements, workloadNodePlacement)
 	util.SetRecommendedLabels(pod, r.installerLabels, "cdi-controller")
 
 	if err := r.client.Create(context.TODO(), pod); err != nil {
@@ -532,9 +534,13 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *co
 }
 
 // MakeCloneSourcePodSpec creates and returns the clone source pod spec based on the target pvc.
-func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image, pullPolicy string, imagePullSecrets []corev1.LocalObjectReference, sourcePvcName, sourcePvcNamespace, ownerRefAnno string,
-	serverCACert []byte, targetPvc *corev1.PersistentVolumeClaim, resourceRequirements *corev1.ResourceRequirements,
+func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image, pullPolicy, ownerRefAnno string, imagePullSecrets []corev1.LocalObjectReference,
+	serverCACert []byte, targetPvc, sourcePvc *corev1.PersistentVolumeClaim, resourceRequirements *corev1.ResourceRequirements,
 	workloadNodePlacement *sdkapi.NodePlacement) *corev1.Pod {
+
+	sourcePvcName := sourcePvc.GetName()
+	sourcePvcNamespace := sourcePvc.GetNamespace()
+	sourcePvcUID := string(sourcePvc.GetUID())
 
 	var ownerID string
 	cloneSourcePodName := targetPvc.Annotations[AnnCloneSourcePod]
@@ -565,6 +571,7 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 				// this label is used when searching for a pvc's cloner source pod.
 				cc.CloneUniqueID:          cloneSourcePodName,
 				common.PrometheusLabelKey: common.PrometheusLabelValue,
+				hostAssistedCloneSource:   sourcePvcUID,
 			},
 		},
 		Spec: corev1.PodSpec{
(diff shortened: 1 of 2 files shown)


tests/cloner_test.go
```diff
@@ -2911,6 +2915,8 @@ func doFileBasedCloneTest(f *framework.Framework, srcPVCDef *v1.PersistentVolume
 	targetDV.Labels["test-label-1"] = "test-label-key-1"
 	targetDV.Annotations["test-annotation-1"] = "test-annotation-key-1"
 
+	targetDV.Annotations[controller.AnnPodRetainAfterCompletion] = "true"
```
akalenyu, line 2918: This already happens in the test at https://github.com/kubevirt/containerized-data-importer/blob/597076f7d7c322579b45c191ae21c751f357a96f/tests/cloner_test.go#L119 Do we still want to keep?

awels (commented):
Can we also verify in a functional test that when we clone multiple times from the same source, and the source is RWO, that all the source pods are on the same node? We already have a multiple clone test [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/cloner_test.go#L1496-L1539)

We just need to modify it to check if all the pods are on the same node IF the source volume is RWO (which is not always the case, nfs and ceph will be RWX)

tests/cloner_test.go
ido106: I think it is enough. I will delete it and run the tests again.

pkg/controller/clone-controller.go
```diff
@@ -532,10 +534,13 @@ func (r *CloneReconciler) CreateCloneSourcePod(image, pullPolicy string, pvc *co
...
 
 // MakeCloneSourcePodSpec creates and returns the clone source pod spec based on the target pvc.
-func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image, pullPolicy string, imagePullSecrets []corev1.LocalObjectReference, sourcePvcName, sourcePvcNamespace, ownerRefAnno string,
-	serverCACert []byte, targetPvc *corev1.PersistentVolumeClaim, resourceRequirements *corev1.ResourceRequirements,
+func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image, pullPolicy string, imagePullSecrets []corev1.LocalObjectReference, ownerRefAnno string,
+	serverCACert []byte, targetPvc, sourcePvc *corev1.PersistentVolumeClaim, resourceRequirements *corev1.ResourceRequirements,
 	workloadNodePlacement *sdkapi.NodePlacement) *corev1.Pod {
 
+	sourcePvcName := sourcePvc.GetName()
+	sourcePvcNamespace := sourcePvc.GetNamespace()
```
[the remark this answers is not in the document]
ido106: `sourcePvcName` and `sourcePvcNamespace` were originally sent as parameters to the function, so I decided to pass `sourcePvc` instead and place them at the beginning of the function for backwards compatibility. That was after I used `sourcePvcUid`, but I thought about it afterwards and I really should define it in the same way like you said.

akalenyu:
> Can we also verify in a functional test that when we clone multiple times from the same source, and the source is RWO, that all the source pods are on the same node? We already have a multiple clone test [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/cloner_test.go#L1496-L1539)
> 
> We just need to modify it to check if all the pods are on the same node IF the source volume is RWO (which is not always the case, nfs and ceph will be RWX)

So this test actually doesn't test ceph with host assisted (only smart clone), that is why I made https://github.com/kubevirt/containerized-data-importer/pull/2575
(https://github.com/kubevirt/containerized-data-importer/issues/2576 is only interesting in case the storage is not topology constrained).

I canceled the hold on that PR so we can decide if we want it in

ido106:
> Changes look good, but I think you might have missed my overall comment about verifying that when we do multiple clones at once (we have tests that already do this) we should ensure that the source pods are all on the same node. I put in a link to the tests.

The problem is that this test will work anyway even without these changes. Maybe I can create 2 clones with 2 different source pvc's and check if each one was placed on the correct node ?
I added the test you asked for (locally) so I can push it but maybe we should check it another way.

tests/cloner_test.go
```diff
@@ -1813,45 +1823,57 @@ var _ = Describe("all clone tests", func() {
...
 				Expect(f.VerifyTargetPVCContentMD5(f.Namespace, utils.PersistentVolumeClaimFromDataVolume(dv), matchFile, md5sum[:32])).To(BeTrue())
+
+				pvc, err := f.K8sClient.CoreV1().PersistentVolumeClaims(dv.Namespace).Get(context.TODO(), dv.Name, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+
+				if cloneSourcePod := pvc.Annotations[ctrl.AnnCloneSourcePod]; cloneSourcePod != "" {
+					By(fmt.Sprintf("Getting pod %s/%s", dv.Namespace, cloneSourcePod))
+					pod, err := f.K8sClient.CoreV1().Pods(dv.Namespace).Get(context.TODO(), cloneSourcePod, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					podsNodeName[dv.Name] = pod.Spec.NodeName
```
awels, line 1857: Would it make sense to have the map be `map[string]bool` and on this line just say `podsNodeName[pod.Spec.NodeName]=true` and below check `Expect(podsNodeName).To(HaveLen(1))` We are essentially creating a set of node names, and ensure there is only 1 node name. This demonstrates all the pods were on the same node.

awels (commented):
Looks good just one minor comment. was having a hard time understanding what that part of the test was doing.

tests/cloner_test.go
```diff
@@ -1042,14 +1042,13 @@ var _ = Describe("all clone tests", func() {
...
+					// All pods should be in the same node except when the map is empty in smart clone
 					if len(podsNodeName) > 0 {
-						Expect(podsNodeName[targetDvs[0].Name]).To(Equal(podsNodeName[targetDvs[1].Name]))
-						Expect(podsNodeName[targetDvs[1].Name]).To(Equal(podsNodeName[targetDvs[2].Name]))
+						Expect(podsNodeName).To(HaveLen(1))
 					}
 				})
 
-				It("[rfe_id:1277][test_id:1899][crit:High][vendor:cnv-qe@redhat.com][level:component] Should allow multiple cloning operations in parallel for block devices", func() {
+				FIt("[rfe_id:1277][test_id:1899][crit:High][vendor:cnv-qe@redhat.com][level:component] Should allow multiple cloning operations in parallel for block devices", func() {
```
akalenyu, line 1051: Other than the FIts the PR looks good to me!

ido106: whoops

tests/cloner_test.go
```diff
@@ -1019,19 +1020,37 @@ var _ = Describe("all clone tests", func() {
...
+						}
+
 						By("Deleting verifier pod")
 						err = f.K8sClient.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), utils.VerifierPodName, metav1.DeleteOptions{})
 						Expect(err).ToNot(HaveOccurred())
 						_, err = utils.WaitPodDeleted(f.K8sClient, utils.VerifierPodName, f.Namespace.Name, verifyPodDeletedTimeout)
 						Expect(err).ToNot(HaveOccurred())
 					}
+
+					// All pods should be in the same node except when the map is empty in smart clone
```
akalenyu, line 1045: You should be able to use `cloneType` in this context to tell you if this is going to be smart clone or not
For example `cloneType != "snapshot"`

ido106: Ok, I changed the condition to `if cloneType == "network"`

akalenyu:

Same issue as in https://github.com/kubevirt/containerized-data-importer/issues/2550,
maybe this change make it fail more often? in that case we might want to comment out `Expect(restartCount).To(BeNumerically("<", 2))` with a TODO to remove once the issue is fixed

akalenyu:
forgot to lgtm again
/lgtm

PR #2648, opened 2023-03-20
Add nfs-subdir-external-provisioner to CapabilitiesByProvisionerKey

fruboes:
This PR adds entry on nfs-subdir-external-provisioner to CapabilitiesByProvisionerKey mapping so storageprofile created for this profile during cdi initialization will have accessModes and volumeMode correctly set.

**Release note**:
```release-note
NONE
```


akalenyu:
/ok-to-test

thanks!

PR #2650, opened 2023-03-21
Annotate DIC-created DV for immediate binding

arnongilboa:
If the storage class binding mode is WaitForFirstConsumer, and the annotation was not explicitly added to the DIC DV template, the created DV will get stuck in WFFC phase.

Fixes [bz #2166394](https://bugzilla.redhat.com/show_bug.cgi?id=2166394)

**Release note**:
```release-note
Annotate DataImportCron-created DataVolumes for immediate binding, so they will not get stuck in WaitForFirstConsumer phase.
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -943,7 +943,7 @@ func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCro
 	dv.Name = dataVolumeName
 	dv.Namespace = cron.Namespace
 	r.setDataImportCronResourceLabels(cron, dv)
-	passCronAnnotationToDv(cron, dv, cc.AnnImmediateBinding)
+	cc.AddAnnotation(dv, cc.AnnImmediateBinding, "true")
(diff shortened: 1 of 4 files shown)


awels:
Yeah, since this a DiC we don't care too much where they end up in this particular scenario. Note this only does anything in the case of WFFC storage.

akalenyu:
> Yeah, since this a DiC we don't care too much where they end up in this particular scenario. Note this only does anything in the case of WFFC storage.

So if someone wanted their WFFC golden image PVCs on a specific node, how can they achieve that?

awels:
With this it would not be possible, can't think of a reason why one would want to do that though.

awels:
Well that causes the bugzilla this is trying to solve. If someone forgets, and then tries to add it later it is not propagated on purpose. Until we get complaints from someone that they want their images on a particular node I don't think we should worry about that particular thing. It would be much more common for someone to forget to add the annotation

akalenyu:
may go away after https://github.com/kubevirt/containerized-data-importer/pull/2475
the theory is that one of the strict reconciliation tests is scaling down the controller deployment,
and that may be costing some time in the -1 check

PR #2674, opened 2023-03-28
Correctly wait for resourcequota "ready" state in tests

akalenyu:
Apparently the correct way to understand if a quota is "ready" is to look at the keys in the `used` field.
Hopefully this eliminates error like these on test runs
```bash
persistentvolumeclaims "test-pvc" is forbidden: status unknown for quota: test-storage-quota
```

Sources:
https://github.com/kubernetes/apiserver/blob/a32e26b98aff9e1dfdf0fb8d0d49d9b56287b875/pkg/admission/plugin/resourcequota/controller.go#L473-L475 https://github.com/kubernetes/kubernetes/blob/a24aef65108bb3549c9ad31f51aab8097c640447/test/e2e/apimachinery/resource_quota.go#L2152

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -462,7 +462,7 @@ func (f *Framework) CreateQuotaInSpecifiedNs(ns string, requestCPU, requestMemor
 		if err != nil {
 			return false, err
 		}
-		return len(quota.Status.Hard) == 4, nil
+		return len(quota.Status.Used) == 4, nil
 	})
 }
 
@@ -486,7 +486,14 @@ func (f *Framework) UpdateQuotaInNs(requestCPU, requestMemory, limitsCPU, limits
 	if err != nil {
 		ginkgo.Fail("Unable to set resource quota " + err.Error())
 	}
-	return err
+	return wait.PollImmediate(5*time.Second, nsDeleteTime, func() (bool, error) {
+		quota, err := f.K8sClient.CoreV1().ResourceQuotas(f.Namespace.GetName()).Get(context.TODO(), "test-quota", metav1.GetOptions{})
+		if err != nil {
+			fmt.Fprintf(ginkgo.GinkgoWriter, "ERROR: GET ResourceQuota failed once, retrying: %v\n", err.Error())
+			return false, nil
+		}
+		return len(quota.Status.Used) == 4, nil
+	})
 }
 
 // CreateStorageQuota creates a quota to limit pvc count and cumulative storage capacity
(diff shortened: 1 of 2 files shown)


tests/framework/framework.go
```diff
@@ -462,7 +462,7 @@ func (f *Framework) CreateQuotaInSpecifiedNs(ns string, requestCPU, requestMemor
 		if err != nil {
 			return false, err
 		}
-		return len(quota.Status.Hard) == 4, nil
+		return len(quota.Status.Used) == 4, nil
```
alromeros, line 465: According to https://pkg.go.dev/k8s.io/api/core/v1#ResourceQuotaStatus, 
```go
	// Used is the current observed total usage of the resource in the namespace.
```
How do we expect this value to reach 4 if we are just "getting" the quota in the PollImmediate and not creating/doing anything? I think I'm missing something because the same reasoning could be applied for `Hard`

tests/transfer_test.go
```diff
@@ -472,6 +480,7 @@ var _ = Describe("[rfe_id:5630][crit:high]ObjectTransfer tests", func() {
 			rq.Spec.Hard[corev1.ResourceRequestsStorage] = *resource.NewQuantity(bq, resource.DecimalSI)
 			rq, err = f.K8sClient.CoreV1().ResourceQuotas(targetNs.Name).Update(context.TODO(), rq, metav1.UpdateOptions{})
 			Expect(err).ToNot(HaveOccurred())
+			Eventually(quotaUpdated, 2*time.Minute, 2*time.Second).Should(BeNumerically("==", 1))
```
alromeros, line 483: Maybe also check this before updating the quota? I would also add some comments explaining what we are doing here, the test seems a little bit confusing.

alromeros (commented):
Good job with this, it looks more reasonable than using `Hard`. Just some comments, still trying to figure out how this works.

tests/framework/framework.go
alromeros: And how do we expect `Used` to already be 4 when the quota has just been created? It made sense with `Hard` since:
```go
	// Hard is the set of enforced hard limits for each named resource.
```

akalenyu: So `used` keys take a couple of seconds to be created, that's why we need to wait for that
If you follow the k8s links in the PR description, that is how the admission hook works as well,
it would only proceed if the `used` keys were created. Note we are not waiting for the usage to go up or anything, we just need the keys that track the usage to be there

tests/transfer_test.go
akalenyu: Yep, changed the test name to better reflect the steps

akalenyu: No need to check before updating, just after applying changes(create/update)

alromeros:
Thanks for the offline clarification!
/lgtm

PR #2677, opened 2023-03-30
Skip TLS crypto test when routes are used/ensure anyvolumedatasource test pvc is gone

akalenyu:

Fixes #

**Release note**:

```release-note
NONE
```


tests/external_population_test.go
```diff
@@ -44,22 +45,31 @@ var _ = Describe("Population tests", func() {
...
 	dataSourceRef := &corev1.TypedLocalObjectReference{
 		APIGroup: &apiGroup,
 		Kind:     populatorKind,
 		Name:     samplePopulatorName,
 	}
+	dummySourceRef := &corev1.TypedLocalObjectReference{
+		APIGroup: &dummyAPIGroup,
+		Kind:     "Dummy",
+		Name:     "dummyname",
+	}
```
mhenriks, line 58: isn't it an error for source and sourceRef to not match?

akalenyu: they'll match here, just switching to a dummy one so we don't trigger the testing populator

PR #2678, opened 2023-03-30
Upload populator

ShellyKa13:
    The upload populator controller can be used
    standalone without the need of datavolume.
    It reconciles pvc with upload dataSourceRef
    and uses populators API to populated the pvc
    with an upload command.
    The controller creates pvc' with upload
    annotation. After the upload completes it
    rebinds the pv to the original target pvc and
    deletes pvc prime.
    Eventually we get a bound PVC which is already
    populated.

you can read more in jira card: https://issues.redhat.com/browse/CNV-27346

Fixes #

Still missing more tests and UT

**Release note**:

```release-note
Created upload populator controller
```


tests/upload_test.go
ShellyKa13, line 1: Note: I moved the existing tests under new Contexts and adjusted the before and after a bit, you should focus your review more on the `Context("Upload population"`

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,231 @@
...
+	if err := addCommonPopulatorsWatches(mgr, uploadPopulator); err != nil {
+		return nil, err
+	}
+	if err := addUploadPopulatorWatches(mgr, uploadPopulator, log); err != nil {
+		return nil, err
+	}
+	return uploadPopulator, nil
+}
+
+func addUploadPopulatorWatches(mgr manager.Manager, c controller.Controller, log logr.Logger) error {
```
alromeros, line 77: I think these watches and indexer could also be included in `addCommonPopulatorsWatches`, maybe just adding a `typeToWatch` argument as we do here https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume/clone-controller-base.go#L697.

alromeros (commented):
Very good job Shelly, I'll continue with the review next week since I don't have that much time now, just adding a comment to consider.

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,231 @@
...
+		return r.reconcileTargetPVC(pvc)
+	}
+
+	if isPVCPrime(pvc) {
+		_, ok := pvc.Annotations[cc.AnnUploadRequest]
+		if !ok {
+			// Ignore PVCs that aren't for this populator to handle
+			return reconcile.Result{}, nil
+		}
+		return r.reconcilePVCPrime(pvc)
```
mhenriks, line 147: it is cleaner to just reconcile target pvc and have the watch will trigger target pvc reconcile when pvc prime changes

mhenriks (commented):
Great start!

Let's really think about what to name the populator resources

pkg/operator/controller/controller_test.go
```diff
@@ -1700,6 +1700,7 @@ func createNotReadyEventValidationMap() map[string]bool {
 	match[normalCreateSuccess+" *v1.CustomResourceDefinition datasources.cdi.kubevirt.io"] = false
 	match[normalCreateSuccess+" *v1.CustomResourceDefinition dataimportcrons.cdi.kubevirt.io"] = false
 	match[normalCreateSuccess+" *v1.CustomResourceDefinition objecttransfers.cdi.kubevirt.io"] = false
+	match[normalCreateSuccess+" *v1.CustomResourceDefinition uploadsource.cdi.kubevirt.io"] = false
```
akalenyu, line 1703: uploadsources?

pkg/uploadproxy/uploadproxy.go
```diff
@@ -234,22 +235,28 @@ func (app *uploadProxyApp) resolveUploadPath(pvcName, pvcNamespace, defaultPath
...
 		return "", fmt.Errorf("rejecting upload request for PVC %s - upload content-type %s is invalid", pvcName, contentType)
 	}
+
+	//If using upload populator then the upload server will have PVC' name in the url
+	if populators.IsPVCDataSourceRefKind(pvc, cdiv1.UploadSourceRef) {
+		pvcPrimeName := populators.PVCPrimeName(pvc)
+		return app.urlResolver(pvcNamespace, pvcPrimeName, path), nil
+	}
+
+	return app.urlResolver(pvcNamespace, pvcName, path), nil
```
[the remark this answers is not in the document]
ShellyKa13: as far as I see it the uploadproxy has to get the pvc prime in order to check the annotation and pod phase on the pvc. hence it has to know the pvc prime name anyways. Im not sure its so bad that it will know the format of getting pvc' name for that. you are also not revealing anything.
You are relying here on function createUploadServiceNameFromPvcName which is in the logic of the upload controller so I dont see how different it is between relying on the pvcprimename function.
If you update an annotation on the target pvc you rely on the annotation name that the populator updates so seems pretty much the same for me without the need to do an update on the target PVC and reconcile again to create the pvc'.

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	return nil
+}
+
+func addCommonPopulatorsWatches(mgr manager.Manager, c controller.Controller) error {
+	// Setup watches
+	if err := c.Watch(&source.Kind{Type: &corev1.PersistentVolumeClaim{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
ShellyKa13: what about when we watch the pod or the PV, in there you only have information for PVC' how are you queuing the target PVC in this type of watches?

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,231 @@
...
+	if err := r.client.Get(context.TODO(), req.NamespacedName, pvc); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	dataSourceRef := pvc.Spec.DataSourceRef
+	if dataSourceRef != nil {
+		if !IsPVCDataSourceRefKind(pvc, cdiv1.UploadSourceRef) || dataSourceRef.Name == "" {
```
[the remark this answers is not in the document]
ShellyKa13: I think its better to be safe and make sure again.

pkg/controller/populators/upload-populator.go
mhenriks: Populator should not care about the upload pod.  THe upload PVC (pvc prime) has all relevant status

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
mhenriks: Should not care about Pods, also don't think that PVs are important either because the PVC will get updated when binding happens

pkg/uploadproxy/uploadproxy.go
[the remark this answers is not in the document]
mhenriks: I am pretty strongly against the upload proxy knowing anything about populator implementation details

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
ShellyKa13: so you say I can remove the watch for pods and pvs then? in that case I guess we can reconcile only the target PVC

tests/framework/pvc.go
```diff
@@ -42,6 +43,18 @@ func (f *Framework) CreateBoundPVCFromDefinition(def *k8sv1.PersistentVolumeClai
 	return pvc
 }
 
+// CreateBoundPopulationPVCFromDefinition is a wrapper around utils.CreatePVCFromDefinition that also force binds pvc on
+// on WaitForFirstConsumer storage class by executing f.ForceBindIfWaitForFirstConsumer(pvc)
+func (f *Framework) CreateBoundPopulationPVCFromDefinition(def *k8sv1.PersistentVolumeClaim) *k8sv1.PersistentVolumeClaim {
```
akalenyu, line 48: Just making sure I understand:
With populators, Bound means the population is done, so there is no way the original target PVC
will ever be bound before somebody `curl`d the upload endpoint and that process completed successfully?

This was not a problem with Immediate binding storage classes because the whole `ForceBindIfWaitForFirstConsumer` logic was skipped?

akalenyu (commented):
The PR is great, just some suggestions following offline discussions
Mainly the bind.immediate annotation discussion

tests/framework/pvc.go
ShellyKa13: exactly, also the name of this function is probably not accurate, ill change that.
But yet with populators the PVC is bound only after the population, hence you cant wait here for the consumer pod to be running, I create here the consumer pod just to get the selectedNode annotation to be able to do the population

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,248 @@
...
+
+type pvcModifierFunc func(pvc *corev1.PersistentVolumeClaim, source client.Object)
+
+func (r *ReconcilerBase) createPVCPrime(pvc *corev1.PersistentVolumeClaim, source client.Object, waitForFirstConsumer bool, pvcModifier pvcModifierFunc) (*corev1.PersistentVolumeClaim, error) {
+	labels := map[string]string{
+		common.CDILabelKey: common.CDILabelValue,
+	}
+	annotations := make(map[string]string)
+	annotations[annPopulationTargetPVC] = pvc.Name
+	annotations[cc.AnnImmediateBinding] = ""
```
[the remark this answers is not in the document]
ShellyKa13: regarding the first question as you mentioned in other commit, its the other way around, the whole population process will start only when there is some consumer of the PVC that will add the selectedNode annotation, then we put the selectedNode also on the PVC' and we want it to start the upload- for that we add the immediate bind annotation. Once upload completed PVC' is bound, we rebind, and then targetPVC is bound and we delete PVC'.

Regarding using immedaitebound annotation on targetPVC:
I did test it locally it seems to work, do we want it I dont see a reason why not, but maybe @mhenriks you have a reason?

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,248 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// ReconcilerBase members
+type ReconcilerBase struct {
```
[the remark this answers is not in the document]
ShellyKa13: did you mean embed reconcile.Reconciler? I remember I did and there was an issue with that, I cant remember what. Why do you think we need to embed? we put the UploadPopulatorReconciler that we define in the NewUploadPopulator as the reconciler for the controller and it accepts it as an object that implements the reconciler interface.  So the struct implements the interface without the embedment.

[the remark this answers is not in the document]
akalenyu: I mean if this reconciler builds up on the base DV reconciler, it may be better to embed the fields instead of copying them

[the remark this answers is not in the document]
akalenyu: I see, the populators sit in a different package
Regarding embedding the reconcile interface, maybe @arnongilboa can advice why we did that

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,201 @@
...
+		// reconcile will be triggered once created
+		if k8serrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return uploadSource, nil
+}
+
+func getContentType(uploadSource *cdiv1.UploadSource) cdiv1.DataVolumeContentType {
```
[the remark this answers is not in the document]
ShellyKa13: I'll unite them and also use the code for GetContentType(pvc)

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
mhenriks: If we want to support immediate bind annotation on the target pvc OR a property in the CR would be good

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
arnongilboa: I think reconcile.Reconciler iface can be removed from ReconcilerBase. afair it was originally meant to make utest easier, but it's not needed anymore. See #2698.

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,225 @@
...
+	// Rebind PVC to target PVC
+	pv.Annotations = make(map[string]string)
+	pv.Spec.ClaimRef = &corev1.ObjectReference{
+		Namespace:       targetPVC.Namespace,
+		Name:            targetPVC.Name,
+		UID:             targetPVC.UID,
+		ResourceVersion: targetPVC.ResourceVersion,
+	}
+	r.log.V(3).Info("Rebinding PV to target PVC", "PVC", targetPVC.Name)
+	if err := r.client.Update(context.TODO(), pv); err != nil {
```
arnongilboa, line 162: Why don't we delete `pvcPrime` before we update the `PV` `ClaimRef` to `targetPVC`? are we sure it's not racy and PVC controller won't override it as `pvcPrime.Spec.VolumeName` still points the `PV`?

ShellyKa13: the volume name is updated according to the PV claimRef and not the other way around. Also if you delete the pvcPrime while it has the claimRef wouldn't the dynamically provisioned PV be deleted?
And also its whats done in k8s -> https://github.com/kubernetes-csi/lib-volume-populator/blob/master/populator-machinery/controller.go#L648-L673 :)

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
arnongilboa: I guess @mhenriks meant using something like [getDataVolumeOp](https://github.com/kubevirt/containerized-data-importer/blob/bdbf5be21c0d73ce201954231b6b00b0b8528ee8/pkg/controller/datavolume/controller-base.go#L231), so it will reconcile only the relevant PVC according to the controller (upload/import/clone...)

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,225 @@
...
+	}); err != nil {
+		return err
+	}
+
+	mapDataSourceRefToPVC := func(obj client.Object) (reqs []reconcile.Request) {
+		var pvcs corev1.PersistentVolumeClaimList
+		matchingFields := client.MatchingFields{dataSourceRefField: indexKeyFunc(obj.GetNamespace(), obj.GetName())}
+		if err := mgr.GetClient().List(context.TODO(), &pvcs, matchingFields); err != nil {
+			log.Error(err, "Unable to list PVCs", "matchingFields", matchingFields)
+			return reqs
```
arnongilboa, line 210: be consistent with the named returns here (either with reqs or without it)

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		for _, pvc := range pvcs.Items {
+			reqs = append(reqs, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: pvc.Namespace, Name: pvc.Name}})
```
arnongilboa, line 213: use `client.ObjectKeyFromObject(pvc)` instead

pkg/controller/populators/populator-base.go
arnongilboa: Cool, I learned something new. I simply played with this flow [here](https://github.com/kubevirt/containerized-data-importer/pull/2683/commits/2e23d09e3cb6d0a76f406247d07339a28342ef2a#diff-f492548a4d272a2e8c736f701a769c82d4143960f57bd0dd30f1541b08d3ac79R1810-R1835). I guess it was different in my case as I used `PersistentVolumeReclaimRetain`? I saw that if I don't delete the original PVC it will take the PV again.

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
arnongilboa: Ignore it ^^^, you do it much better.

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if cc.IsUnbound(pvc) {
+		// If PVC' doesn't exist yet, create it
+		if pvcPrime == nil {
+			pvcPrime, err = r.createPVCPrime(pvc, uploadSource, waitForFirstConsumer, updateAnnotations)
+			if err != nil {
+				r.recorder.Eventf(pvc, corev1.EventTypeWarning, errCreatingPVCPrime, err.Error())
+				return reconcile.Result{}, err
+			}
+		} else {
```
arnongilboa, line 147: better always `return reconcile.Result{}, err` in the `if` and get rid of the `else`

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,178 @@
...
+				return reconcile.Result{}, err
+			}
+		}
+		return reconcile.Result{}, nil
+	}
+
+	// Wait for the bind controller to rebind the PV
+	if pvcPrime != nil {
+		if corev1.ClaimLost != pvcPrime.Status.Phase {
+			return reconcile.Result{RequeueAfter: 2 * time.Second}, err
```
arnongilboa, line 167: You are already returning on all err != nil flows, so I guess you can return nil here?

ShellyKa13:
/hold
will wait until the [import populator PR](https://github.com/kubevirt/containerized-data-importer/pull/2690) is merged and then work on rebasing and adjusting code.

pkg/uploadproxy/uploadproxy.go
```diff
@@ -221,47 +222,56 @@ func (app *uploadProxyApp) handleUploadRequest(w http.ResponseWriter, r *http.Re
...
 		if err != nil {
 			if k8serrors.IsNotFound(err) {
 				return false, fmt.Errorf("rejecting Upload Request for PVC %s that doesn't exist", pvcName)
 			}
 
 			return false, err
 		}
+		// If using upload populator then need to check upload possibility to the PVC'
+		if populators.IsPVCDataSourceRefKind(pvc, cdiv1.VolumeUploadSourceRef) {
+			pvcPrimeName, ok := pvc.Annotations[populators.AnnPVCPrimeName]
```
mhenriks, line 261: I think this is fine but a little extra.  I don't think it is necessary to check the datasourceref.  Also the "pvc prime" naming betrays populator implementation details.

mhenriks (commented):
looking good!

pkg/uploadproxy/uploadproxy.go
ShellyKa13: a. you mean just check if the annotation exist? right it works, the way it was before required this check, I can remove.
b. we need to check that pvc prime is ready for upload same as we do for regular pvc so we need to know eventually the pvc prime name here, not sure what is an alternative, putting the service name and inferring from it the pvc prime name? seems redundant to me, I was in favor of using the pvcprimename function like I did before and refer from using all this annotation but..
what do you suggest doing otherwise?

pkg/controller/populators/upload-populator.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil
+	}
+
+	delete(pvc.Annotations, AnnPVCPrimeName)
+	return r.client.Update(context.TODO(), pvc)
+}
+
+func (r *UploadPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
+	updated, err := r.updatePVCPrimeNameAnnotation(pvc, pvcPrime.Name)
+	if updated || err != nil {
```
arnongilboa, line 145: why return if updated?

mhenriks:
/hold
/approve

let's make sure @alromeros takes another look

alromeros (commented):
Very good job! Just have some comments, most of them nitpicks. The only missing thing that may be important is the preallocation field in the CRD spec.

Maybe it's also a good idea to set AnnContentType in the target PVC somewhere in populator-base, as discussed with @mhenriks.

pkg/controller/populators/upload-populator.go
ShellyKa13: until its actually updated the upload cant start anyways, no really need to continue the reconcile loop

ShellyKa13:
@alromeros added preallocation and small fixes in the new commit. I don't want to add the content type to the target pvc in this commit, dont think its relevant

pkg/controller/populators/populator-base.go
mengkai514: It is said in the introduction that pvc prime will be deleted, but in the actual scenario, pvc prime still exists after the upload is completed. And no pv rebind phenomenon is observed. I don't understand this, can you explain it?

PR #2680, opened 2023-03-31
Allow restoreSize to be 0, and use request size if 0

### LINKED ISSUE #2679: Snapshot clone doesn't handle CSI drivers that have 0 in restoreSize of a snapshot properly
**What happened**:
When doing a snapshot clone, the part of the code that creates a PVC with the snapshot source currently relies on the restoreSize of the VolumeSnapshot. If this value is 0 then creating the PVC fails (because we are attempting to create a 0 size PVC). For various reasons it might not be feasible to get a proper value in this field by the driver. See: https://github.com/democratic-csi/democratic-csi/issues/129 We do know ahead of time what we want the size of the PVC to be, but due to some older broken CSI drivers we had to start at the restoreSize and then expand to the target size.

If we consider that behavior (target PVC size must equal restore size) a bug in the CSI driver, then we can create a PVC of the requested size regardless of the restoreSize in the snapshot.

CSI drivers that have 0 in the restoreSize can still use snapshot cloning, if we use the size of the requested DataVolume.

Have a CSI driver that can create snapshots, and returns restoreSize 0 in their snapshot. Then attempt to do a snapshot clone.

**Additional context**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 1.56 and below
- DV specification: N/A
- Install tools: N/A
- Others: N/A

awels:
Some CSI provisioner when making volume snapshots don't set the restore size because it is hard to figure out the value or the actual stored size on the storage is 0 (Copy on Write). There was a strict check in place in case of CSI drivers not allowing a restore to a size other than the restoreSize of the snapshot.

This makes the check more lenient that when we encounter a 0 size restoreSize, we can assume we can use the requested size value in the pvc

Fixes #2679 

**Release note**:

```release-note
BugFix: Fixed failing snapshot clone if CSI driver set 0 to restoreSize of volumeSnapshot
```

### DIFF
--- a/pkg/controller/datavolume/smart-clone-controller.go
+++ b/pkg/controller/datavolume/smart-clone-controller.go
@@ -356,8 +356,21 @@ func (r *SmartCloneReconciler) getTargetPVC(dataVolume *cdiv1.DataVolume) (*core
 func newPvcFromSnapshot(dv *cdiv1.DataVolume, name string, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec) (*corev1.PersistentVolumeClaim, error) {
 	targetPvcSpecCopy := targetPvcSpec.DeepCopy()
 	restoreSize := snapshot.Status.RestoreSize
-	if restoreSize == nil {
+	if restoreSize == nil || restoreSize.Sign() == -1 {
 		return nil, fmt.Errorf("snapshot has no RestoreSize")
+	}
+	if restoreSize.IsZero() {
+		// Don't error out if the restoreSize is 0. Use the size from the DataVolume instead
+		if dv.Spec.PVC != nil {
+			reqSize := dv.Spec.PVC.Resources.Requests[corev1.ResourceStorage]
+			restoreSize = &reqSize
+		} else if dv.Spec.Storage != nil {
+			reqSize := dv.Spec.Storage.Resources.Requests[corev1.ResourceStorage]
+			restoreSize = &reqSize
+		}
+		if restoreSize.IsZero() {
+			return nil, fmt.Errorf("unable to determine restore size of PVC")
+		}
 	}
 
 	key, err := cache.MetaNamespaceKeyFunc(snapshot)
--- a/pkg/controller/datavolume/smart-clone-controller_test.go
+++ b/pkg/controller/datavolume/smart-clone-controller_test.go
@@ -18,6 +18,7 @@ package datavolume
 
 import (
 	"context"
+	"fmt"
 
 	"k8s.io/apimachinery/pkg/api/resource"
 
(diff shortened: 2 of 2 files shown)


pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -356,9 +356,22 @@ func (r *SmartCloneReconciler) getTargetPVC(dataVolume *cdiv1.DataVolume) (*core
...
+	if restoreSize == nil || restoreSize.Sign() == -1 {
 		return nil, fmt.Errorf("snapshot has no RestoreSize")
 	}
+	if restoreSize.IsZero() {
+		// Don't error out if the restoreSize is 0. Use the size from the DataVolume instead
+		if dv.Spec.PVC != nil {
+			reqSize := dv.Spec.PVC.Resources.Requests[corev1.ResourceStorage]
+			restoreSize = &reqSize
+		} else if dv.Spec.Storage != nil {
+			reqSize := dv.Spec.Storage.Resources.Requests[corev1.ResourceStorage]
```
akalenyu, line 368: this function already receives `targetPvcSpec` which does the PVC/Storage API handling and provides the resulting spec

pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -356,9 +356,22 @@ func (r *SmartCloneReconciler) getTargetPVC(dataVolume *cdiv1.DataVolume) (*core
...
+			restoreSize = &reqSize
+		}
+		if restoreSize.IsZero() {
```
akalenyu, line 371: I think this is handled in our hooks https://github.com/kubevirt/containerized-data-importer/blob/92382bdffe14b8c106e16724003d1457f84d2d7c/pkg/apiserver/webhooks/datavolume-validate.go#L543-L548

pkg/controller/datavolume/smart-clone-controller.go
```diff
@@ -356,9 +356,22 @@ func (r *SmartCloneReconciler) getTargetPVC(dataVolume *cdiv1.DataVolume) (*core
 func newPvcFromSnapshot(dv *cdiv1.DataVolume, name string, snapshot *snapshotv1.VolumeSnapshot, targetPvcSpec *corev1.PersistentVolumeClaimSpec) (*corev1.PersistentVolumeClaim, error) {
 	targetPvcSpecCopy := targetPvcSpec.DeepCopy()
 	restoreSize := snapshot.Status.RestoreSize
-	if restoreSize == nil {
+	if restoreSize == nil || restoreSize.Sign() == -1 {
```
akalenyu, line 359: Are we sure we want to error out on the negative sign?
I mean, if the volumesnapshot is ready, and for some reason, the provisioner reports a negative sign, couldn't we proceed in the same way as if it was zero?

Just want to make sure we won't be back with the same issue we have with `0`

akalenyu (commented):
Great!

pkg/controller/datavolume/smart-clone-controller.go
akalenyu: Same for nil `.Status.RestoreSize`, should we be able to work with that as well?

awels: I was originally going to allow missing restoreSize, but if it is missing then that means there is a serious problem with the snapshot, and we should error out, same with a negative number.

PR #2683, opened 2023-04-04
Create PVC if possible even if the StorageClass is missing

### LINKED ISSUE #2186: No information if storage class can't be detected

>
/kind bug
> /kind enhancement

**What happened**:

I added by accident a storage class called `""`. The logs of the cdi deployment showed:

```
{"level":"error","ts":1646754170.6523237,"logger":"controller-runtime.manager.controller.datavolume-controller","msg":"Reconciler error","name":"example-69ntw-rhcos","namespace":"clusters-example","error":"unable to retrieve storage class"
```

but the datavolume status remained empty:

```
    storage:
      resources:
        requests:
          storage: 16Gi
      storageClassName: ""
  status: {}
```

I wonder if you want to indictate that something is wrong on the object itself.

**Environment**:
- Install tools:
- Others:

arnongilboa:
When `PVC` is created with `storageClassName` and the `SC` is not found, k8s looks for `PV` with the `storageClassName` for satisfying this claim. In this case k8s supports also a blank (“”, not the nil default one) `storageClassName`. To support this behavior we added:
- `DV` controller support for this flow (for both “” and non-existing `SC`)
- Condition update and event when `StorageSpec` `PVC` rendering errors and `PVC` is not created (e.g. missing both `AccessModes` and `SC`/`PV`)
- `PVC` is created even if a satisfying `SC`/`PV` doesn't exist if `spec.pvc/storage` `AccessModes` is set (otherwise k8s PVC validation fails). PVC/DV phase  will be `Pending` until a satisfying `SC`/`PV` is found
- `PV` watch to reconcile `DVs` waiting for the `PV` `storageClassName`
- `PV` `storageClassName` indexer, so we can list the relevant `PVs`

Fixes #2186

**Release note**:
```release-note
Create PVC if possible even if the StorageClass is missing; Update DV condition fire and event on PVC rendering error
```

### DIFF
--- a/pkg/apiserver/webhooks/dataimportcron-validate.go
+++ b/pkg/apiserver/webhooks/dataimportcron-validate.go
@@ -40,6 +40,8 @@ type dataImportCronValidatingWebhook struct {
 }
 
 func (wh *dataImportCronValidatingWebhook) Admit(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	var causes []metav1.StatusCause
+
 	if ar.Request.Resource.Group != cdiv1.CDIGroupVersionKind.Group || ar.Request.Resource.Resource != "dataimportcrons" {
 		klog.V(3).Infof("Got unexpected resource type %s", ar.Request.Resource.Resource)
 		return toAdmissionResponseError(fmt.Errorf("unexpected resource: %s", ar.Request.Resource.Resource))
@@ -52,8 +54,8 @@ func (wh *dataImportCronValidatingWebhook) Admit(ar admissionv1.AdmissionReview)
 		return toAdmissionResponseError(err)
 	}
 
-	causes := validateNameLength(cron.Name, validation.DNS1035LabelMaxLength)
-	if len(causes) > 0 {
+	if cause := validateNameLength(cron.Name, validation.DNS1035LabelMaxLength); cause != nil {
+		causes = append(causes, *cause)
 		return toRejectedAdmissionResponse(causes)
 	}
 
@@ -152,8 +154,8 @@ func (wh *dataImportCronValidatingWebhook) validateDataImportCronSpec(request *a
 		return causes
 	}
 
-	causes = validateNameLength(spec.ManagedDataSource, validation.DNS1035LabelMaxLength)
-	if len(causes) > 0 {
+	if cause := validateNameLength(spec.ManagedDataSource, validation.DNS1035LabelMaxLength); cause != nil {
+		causes = append(causes, *cause)
 		return causes
 	}
 
--- a/pkg/apiserver/webhooks/datavolume-validate.go
+++ b/pkg/apiserver/webhooks/datavolume-validate.go
@@ -63,16 +63,15 @@ func validateSourceURL(sourceURL string) string {
 	return ""
 }
 
-func validateNameLength(name string, maxLen int) []metav1.StatusCause {
-	var causes []metav1.StatusCause
+func validateNameLength(name string, maxLen int) *metav1.StatusCause {
 	if len(name) > maxLen {
-		causes = append(causes, metav1.StatusCause{
+		return &metav1.StatusCause{
 			Type:    metav1.CauseTypeFieldValueInvalid,
 			Message: fmt.Sprintf("Name cannot be longer than %d characters", maxLen),
 			Field:   "",
-		})
+		}
 	}
-	return causes
+	return nil
 }
 
 func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *admissionv1.AdmissionRequest, field *k8sfield.Path, spec *cdiv1.DataVolumeSpec, namespace *string) []metav1.StatusCause {
@@ -101,6 +100,10 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSpec(request *admission
 
 	cause, valid := validateStorageSize(spec, field)
 	if !valid {
+		causes = append(causes, *cause)
+		return causes
+	}
+	if cause := validateStorageClassName(spec, field); cause != nil {
 		causes = append(causes, *cause)
 		return causes
 	}
@@ -564,6 +567,22 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSourceSnapshot(snapshot
 	return nil
 }
 
+func validateStorageClassName(spec *cdiv1.DataVolumeSpec, field *k8sfield.Path) *metav1.StatusCause {
+	var sc *string
+
+	if spec.PVC != nil {
+		sc = spec.PVC.StorageClassName
+	} else if spec.Storage != nil {
+		sc = spec.Storage.StorageClassName
+	}
+
+	if sc == nil || *sc == "" {
+		return nil
+	}
+
+	return validateNameLength(*sc, kvalidation.DNS1123SubdomainMaxLength)
+}
+
 func validateStorageSize(spec *cdiv1.DataVolumeSpec, field *k8sfield.Path) (*metav1.StatusCause, bool) {
 	var name string
 	var resources v1.ResourceRequirements
@@ -633,6 +652,8 @@ func validateExternalPopulation(spec *cdiv1.DataVolumeSpec, field *k8sfield.Path
 }
 
 func (wh *dataVolumeValidatingWebhook) Admit(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	var causes []metav1.StatusCause
+
 	klog.V(3).Infof("Got AdmissionReview %+v", ar)
 
 	if err := validateDataVolumeResource(ar); err != nil {
@@ -682,9 +703,9 @@ func (wh *dataVolumeValidatingWebhook) Admit(ar admissionv1.AdmissionReview) *ad
 		}
 	}
 
-	causes := validateNameLength(dv.Name, kvalidation.DNS1123SubdomainMaxLength)
-	if len(causes) > 0 {
+	if cause := validateNameLength(dv.Name, kvalidation.DNS1123SubdomainMaxLength); cause != nil {
 		klog.Infof("rejected DataVolume admission")
+		causes = append(causes, *cause)
 		return toRejectedAdmissionResponse(causes)
 	}
 
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -354,8 +354,11 @@ func GetStorageClassByName(client client.Client, name *string) (*storagev1.Stora
 	if name != nil {
 		storageClass := &storagev1.StorageClass{}
 		if err := client.Get(context.TODO(), types.NamespacedName{Name: *name}, storageClass); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return nil, nil
+			}
 			klog.V(3).Info("Unable to retrieve storage class", "storage class name", *name)
-			return nil, errors.New("unable to retrieve storage class")
+			return nil, errors.Errorf("unable to retrieve storage class %s", *name)
 		}
 		return storageClass, nil
 	}
@@ -392,7 +395,7 @@ func GetFilesystemOverheadForStorageClass(client client.Client, storageClassName
 	}
 
 	targetStorageClass, err := GetStorageClassByName(client, storageClassName)
-	if err != nil {
+	if err != nil || targetStorageClass == nil {
 		klog.V(3).Info("Storage class", storageClassName, "not found, trying default storage class")
 		targetStorageClass, err = GetStorageClassByName(client, nil)
 		if err != nil {
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -189,6 +189,16 @@ func getIndexArgs() []indexArgs {
 				return nil
 			},
 		},
+		{
+			obj:   &corev1.PersistentVolume{},
+			field: claimStorageClassNameField,
+			extractValue: func(obj client.Object) []string {
+				if pv, ok := obj.(*corev1.PersistentVolume); ok && pv.Status.Phase == corev1.VolumeAvailable {
+					return []string{pv.Spec.StorageClassName}
+				}
+				return nil
+			},
+		},
 	}
 }
 
@@ -202,14 +212,6 @@ func CreateCommonIndexes(mgr manager.Manager) error {
 		if err := mgr.GetFieldIndexer().IndexField(context.TODO(), ia.obj, ia.field, ia.extractValue); err != nil {
 			return err
 		}
-	}
-	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &corev1.PersistentVolume{}, claimStorageClassNameField, func(obj client.Object) []string {
-		if pv, ok := obj.(*corev1.PersistentVolume); ok {
-			return []string{pv.Spec.StorageClassName}
-		}
-		return nil
-	}); err != nil {
-		return err
 	}
 	return nil
 }
@@ -319,6 +321,7 @@ func addDataVolumeControllerCommonWatches(mgr manager.Manager, dataVolumeControl
 				if storage != nil &&
 					storage.StorageClassName != nil &&
 					*storage.StorageClassName == pv.Spec.StorageClassName &&
+					pv.Status.Phase == corev1.VolumeAvailable &&
 					getDataVolumeOp(mgr.GetLogger(), &dv, mgr.GetClient()) == op {
 					reqs = append(reqs, reconcile.Request{NamespacedName: types.NamespacedName{Name: dv.Name, Namespace: dv.Namespace}})
 				}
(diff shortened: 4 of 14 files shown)


pkg/apiserver/webhooks/datavolume-validate.go
```diff
@@ -523,6 +526,28 @@ func (wh *dataVolumeValidatingWebhook) validateDataVolumeSourceSnapshot(snapshot
...
+		sc = spec.Storage.StorageClassName
+	}
+
+	if sc == nil {
+		return nil
+	}
+	if *sc == "" {
+		return &metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("storageClassName is empty"),
```
mhenriks, line 544: "" storage class name is valid

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: host-volume
spec:
  capacity:
    storage: 10Gi
  accessModes:
    - ReadWriteOnce
  hostPath:
    path: "/mnt/data"
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-pvc
spec:
  volumeMode: Filesystem
  storageClassName: ""
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
```

```bash
✗ k get pvc
NAME     STATUS   VOLUME        CAPACITY   ACCESS MODES   STORAGECLASS   AGE
my-pvc   Bound    host-volume   10Gi       RWO                           3s
```

mhenriks (commented):
storageClass is optional on a PV

user may ask for storageClass "" on a pvc and they will get a PV with no storageclass (assuming other param match of course)

pkg/apiserver/webhooks/datavolume-validate.go
awels: So had this discussion with Arnon, if you read [this](https://kubernetes.io/blog/2023/01/05/retroactive-default-storage-class/), it does say "" is equivalent to missing, due to historical reasons, but having a blank string is sort of gray at this point. This is fixing a reported issue where someone is passing a blank storage class by accident. I think we should disallow blank, and if someone wants to use the default, they can omit the field.

mhenriks: I dunno, this pr provides no way to explicitly bind to "" storageclass .  That combined with the fact that in 1.26 "" on PVC will bind to default storage class if exists, makes me believe that we should leave things as-is OR allow updating the storageclass on the DV as you can do with PVCs in 1.26

mhenriks: Finally tried this out on a 1.26 cluster.  Apparently, I misread the doc.  "" will not bind to the default storageclass.  And pvc with "" storageClassName can't be updated.

Nevertheless because "" is a legit value I don't think we should disallow it

mhenriks (commented):
This is all a little awkward and racy.  Remind me why we can't go straight to "create the pvc even if the storage class doesn't exist"?

arnongilboa:
> This is all a little awkward and racy. Remind me why we can't go straight to "create the pvc even if the storage class doesn't exist"?

We will `create the pvc even if the storage class doesn't exist` if `spec.pvc/storage` `AccessModes` is set (otherwise k8s `PVC` validation fails). `PVC/DV phase` will be `Pending` until a satisfying `SC/PV` is found.

pkg/controller/datavolume/controller-base.go
```diff
@@ -300,6 +310,30 @@ func addDataVolumeControllerCommonWatches(mgr manager.Manager, dataVolumeControl
...
+	// Watch for PV updates to reconcile the DVs waiting for available PV
+	if err := dataVolumeController.Watch(&source.Kind{Type: &corev1.PersistentVolume{}}, handler.EnqueueRequestsFromMapFunc(
+		func(obj client.Object) (reqs []reconcile.Request) {
+			pv := obj.(*corev1.PersistentVolume)
+			dvList := &cdiv1.DataVolumeList{}
+			if err := mgr.GetClient().List(context.TODO(), dvList, client.MatchingFields{dvPhaseField: ""}); err != nil {
+				return
+			}
+			for _, dv := range dvList.Items {
+				storage := dv.Spec.Storage
```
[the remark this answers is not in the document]
arnongilboa: It's not needed as we `create the pvc even if the storage class doesn't exist` if `AccessModes` is set (otherwise k8s PVC validation fails). `PVC/DV` phase will be `Pending` until a satisfying `SC/PV` is found.

pkg/controller/common/util.go
```diff
@@ -338,7 +338,7 @@ func GetStorageClassByName(client client.Client, name *string) (*storagev1.Stora
 		storageClass := &storagev1.StorageClass{}
 		if err := client.Get(context.TODO(), types.NamespacedName{Name: *name}, storageClass); err != nil {
 			klog.V(3).Info("Unable to retrieve storage class", "storage class name", *name)
-			return nil, errors.New("unable to retrieve storage class")
+			return nil, errors.New(fmt.Sprintf("unable to retrieve storage class %s", *name))
```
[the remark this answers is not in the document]
arnongilboa: fixed

mhenriks:
> > This is all a little awkward and racy. Remind me why we can't go straight to "create the pvc even if the storage class doesn't exist"?
> 
> We will `create the pvc even if the storage class doesn't exist` if `spec.pvc/storage` `AccessModes` is set (otherwise k8s `PVC` validation fails). `PVC/DV phase` will be `Pending` until a satisfying `SC/PV` is found.

Okay, that makes sense, so what about this in the description, what is the follow up?

> TODO in a follow-up PR: use k8s PVC SC/PV pending mechanism and get rid of it in CDI

pkg/controller/datavolume/util.go
```diff
@@ -151,6 +160,37 @@ func copyStorageAsPvc(log logr.Logger, storage *cdiv1.StorageSpec) *v1.Persisten
 	return pvcSpec
 }
 
+// Gets either the bound PV or an available satisfying PV
+func getPV(c client.Client, pvcSpec *v1.PersistentVolumeClaimSpec, pvcName, pvcNamespace string) (*corev1.PersistentVolume, error) {
+	if pvcSpec.StorageClassName == nil {
+		return nil, fmt.Errorf("PVC spec is missing StorageClassName")
```
mhenriks, line 166: Is this really an error?  Can't storageclass name be nil for either `spec.pvc` or `spec.storage`?

mhenriks: Is that tested?

arnongilboa: It's not an error, should return nil here. Fixed.

tests/datavolume_test.go
```diff
@@ -2305,6 +2308,190 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			err = utils.WaitForDataVolumePhase(f, dataVolume.Namespace, cdiv1.Succeeded, dataVolume.Name)
+			Expect(err).ToNot(HaveOccurred())
+		},
+			table.Entry("[test_id:XXXX]the storage class is created (PvcSpec)", testScName, createStorageClass, newDataVolumeWithPvcSpec),
+			table.Entry("[test_id:XXXX]PV with the SC name is created (PvcSpec)", testScName, createPV, newDataVolumeWithPvcSpec),
+			table.Entry("[test_id:XXXX]PV with the SC name (\"\" blank) is created (PvcSpec)", "", createPV, newDataVolumeWithPvcSpec),
+			table.Entry("[test_id:XXXX]the storage class is created (StorageSpec)", testScName, createStorageClass, newDataVolumeWithStorageSpec),
+			table.Entry("[test_id:XXXX]PV with the SC name is created (StorageSpec)", testScName, createPV, newDataVolumeWithStorageSpec),
+			table.Entry("[test_id:XXXX]PV with the SC name (\"\" blank) is created (StorageSpec)", "", createPV, newDataVolumeWithStorageSpec),
+		)
```
duyanyan, line 2492: test_ids: CNV-9926
test_ids: CNV-9927
test_ids: CNV-9928
test_ids: CNV-9929
test_ids: CNV-9930
test_ids: CNV-9931

pkg/controller/datavolume/util.go
```diff
@@ -151,6 +160,37 @@ func copyStorageAsPvc(log logr.Logger, storage *cdiv1.StorageSpec) *v1.Persisten
...
+	pvList := &v1.PersistentVolumeList{}
+	fields := client.MatchingFields{claimStorageClassNameField: *pvcSpec.StorageClassName}
+	if err := c.List(context.TODO(), pvList, fields); err != nil {
+		return nil, err
+	}
+	for _, pv := range pvList.Items {
+		if pv.Status.Phase == corev1.VolumeBound &&
+			pv.Spec.ClaimRef != nil &&
+			pv.Spec.ClaimRef.Name == pvcName &&
+			pv.Spec.ClaimRef.Namespace == pvcNamespace {
```
mhenriks, line 178: Not sure about this loop.  If a PV exists that meets this criteria, I'm pretty sure it would mean that the target PVC already exists.

I think just the following loop is sufficient

arnongilboa: fixed

pkg/apiserver/webhooks/dataimportcron-validate.go
```diff
@@ -152,8 +154,8 @@ func (wh *dataImportCronValidatingWebhook) validateDataImportCronSpec(request *a
 		return causes
 	}
 
-	causes = validateNameLength(spec.ManagedDataSource, validation.DNS1035LabelMaxLength)
-	if len(causes) > 0 {
+	if cause := validateNameLength(spec.ManagedDataSource, validation.DNS1035LabelMaxLength); cause != nil {
+		causes = append(causes, *cause)
```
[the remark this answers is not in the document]
awels: Since we moved from returning a slice of causes to a single cause in validateNameLength. This just changes the way that is handled, but functionally does look the same.

tests/datavolume_test.go
```diff
@@ -2305,6 +2308,190 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+		table.DescribeTable("import DV using StorageSpec without AccessModes, PVC is created only when", func(scName string, scFunc func(string)) {
+			By(fmt.Sprintf("verifying no storage class %s", testScName))
+			_, err := f.K8sClient.StorageV1().StorageClasses().Get(context.TODO(), scName, metav1.GetOptions{})
+			Expect(err).To(HaveOccurred())
+
+			By(fmt.Sprintf("creating new datavolume %s with StorageClassName %s", dataVolumeName, scName))
+			dataVolume := utils.NewDataVolumeWithHTTPImportAndStorageSpec(
+				dataVolumeName, "100Mi", fmt.Sprintf(utils.TinyCoreQcow2URL, f.CdiInstallNs))
+			dataVolume.Spec.Storage.StorageClassName = pointer.String(scName)
+			dataVolume.Spec.Storage.AccessModes = nil
```
awels, line 2389: I know you are being explicit here with no access modes, but no reason for this line as it will be nil if not specified.

awels (commented):
looks good, just a few minor nitpicky things.

tests/datavolume_test.go
```diff
@@ -2305,6 +2308,190 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			})
+		}
+
+		createStorageClass := func(scName string) {
+			var err error
+			By(fmt.Sprintf("creating storage class %s", scName))
+			sc := utils.DefaultStorageClass.DeepCopy()
+			sc.Name = scName
+			sc.ResourceVersion = ""
+			sc.Annotations[controller.AnnDefaultStorageClass] = "false"
```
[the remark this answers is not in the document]
arnongilboa: We DeepCopied the DefaultStorageClass, so it will be "true" if we don't touch it.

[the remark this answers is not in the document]
awels: Good point, missed that detail.

pkg/controller/datavolume/util.go
```diff
@@ -151,6 +152,41 @@ func copyStorageAsPvc(log logr.Logger, storage *cdiv1.StorageSpec) *v1.Persisten
 	return pvcSpec
 }
 
+// Update the PVC spec from either the created PVC or an available satisfying PV
+func updatePvcSpecFromPvcOrPv(c client.Client, pvcSpec *v1.PersistentVolumeClaimSpec, pvc *v1.PersistentVolumeClaim) error {
```
alromeros, line 156: I see we update the pvcSpec here but I'm struggling to see the cases where this is relevant if the PVC already exists. Why do we need to set VolumeMode and AccessMode if the PVC already has them?

pkg/controller/datavolume/util.go
```diff
@@ -77,8 +76,10 @@ func pvcFromStorage(client client.Client, recorder record.EventRecorder, log log
 	if err != nil {
 		return nil, err
 	}
-
 	if storageClass == nil {
+		if err := updatePvcSpecFromPvcOrPv(client, pvcSpec, pvc); err != nil {
```
alromeros, line 80: I think this flow is getting confusing, it'd be good to have some comments explaining why are we doing this. I'm also not a fan of the `updatePvcSpecFromPvcOrPv` name, it's not a must but I think this whole storageClass handling would benefit from being in a separate function.

alromeros (commented):
Nice job! I have some comments and questions about spec rendering, which I think is getting confusing. Other than that looks good to me.

tests/datavolume_test.go
arnongilboa: `NewDataVolumeWithHTTPImportAndStorageSpec()` sets it to `ReadWriteOnce`, so we need to set it to nil here.

pkg/controller/datavolume/util.go
arnongilboa: Simplified and renamed.

pkg/apiserver/webhooks/dataimportcron-validate.go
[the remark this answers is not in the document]
arnongilboa: Right, the previous one was a bit clumsy.

pkg/controller/datavolume/util.go
arnongilboa: Good catch! fixed.

pkg/controller/datavolume/util.go
```diff
@@ -44,85 +44,105 @@ const (
 )
 
 // renderPvcSpec creates a new PVC Spec based on either the dv.spec.pvc or dv.spec.storage section
-func renderPvcSpec(client client.Client, recorder record.EventRecorder, log logr.Logger, dv *cdiv1.DataVolume) (*v1.PersistentVolumeClaimSpec, error) {
+func renderPvcSpec(client client.Client, recorder record.EventRecorder, log logr.Logger, dv *cdiv1.DataVolume, pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaimSpec, error) {
```
alromeros, line 47: I think this would look cleaner if we do something like:

```go
	// If the PVC already exists, just return its spec
	if pvc != nil {
		return pvc.Spec.DeepCopy(), nil
	}
	if dv.Spec.PVC != nil {
		return dv.Spec.PVC.DeepCopy(), nil
	}
	if dv.Spec.Storage != nil {
		return pvcFromStorage(client, recorder, log, dv)
	}
``` 

`pvcFromStorage` would require fewer arguments too, wdyt?

pkg/controller/datavolume/util.go
```diff
@@ -44,85 +44,105 @@ const (
...
+	if pvc == nil {
+		pvcSpec = copyStorageAsPvc(log, dv.Spec.Storage)
+		if err := renderPvcSpecVolumeModeAndAccessModes(client, recorder, log, dv, pvcSpec); err != nil {
+			return nil, err
+		}
+	} else {
+		pvcSpec = pvc.Spec.DeepCopy()
+	}
+
+	if err := renderPvcSpecVolumeSize(client, dv.Spec, pvcSpec); err != nil {
```
alromeros, line 69: Related to my previous comment, we are still rendering the volume size for already existing PVCs, which I believe is not necessary. I think this would be better without PVC handling.

pkg/controller/datavolume/util.go
```diff
@@ -44,85 +44,105 @@ const (
...
+		return nil, err
+	}
+
+	return pvcSpec, nil
+}
 
+func renderPvcSpecVolumeModeAndAccessModes(client client.Client, recorder record.EventRecorder, log logr.Logger, dv *cdiv1.DataVolume, pvcSpec *v1.PersistentVolumeClaimSpec) error {
```
alromeros, line 76: Maybe `renderVolumeAndAccessModes`? I think we could remove the pvcSpec from the name, but `renderPvcSpecVolumeAndAccessModes` would also work

alromeros (commented):
Looks good! just a couple of nitpicks that I think could improve the spec rendering code

pkg/controller/datavolume/util.go
alromeros: Or do we really want to update the volume size? In that case, are we updating the PVC later to reflect the new pvcSpec?

arnongilboa: It is necessary to render the volume size for smart clones, as the DV may require volume size greater than the source PVC, so the traget PVC should be expanded.

alromeros: Oh nice, wasn't sure about that. Thanks for explaining!

pkg/controller/datavolume/util.go
arnongilboa: renderVolumeAndAccessModes sounds good to me.

PR #2687, opened 2023-04-11
update k8s libs to 1.26.

mhenriks:

Been awhile since we updated libs!

Fixes #

Couple thinks that came along with this that may be interesting...

PVC `spec.dataSourceRef` has `namespace` field now but we will not allow it.

ConfigMap based leaderelection is gone, fortunately we were doing configmap + leases for awhile, so we should be good. 

**Release note**:

```release-note
k8s 1.26.1 libs
```


awels:
hmm looks like they changed the quota message format again, and the tests are failing because the string isn't being matched.

mhenriks:
> hmm looks like they changed the quota message format again, and the tests are failing because the string isn't being matched.

Yeah controller runtime added an additional field to the log, those checks are redundant IMO so removed them

mhenriks:
@awels how do we deal with the fossa fails?

awels:
/test pull-containerized-data-importer-fossa
Fixed the fossa issues.

tests/cloner_test.go
```diff
@@ -2235,14 +2235,6 @@ var _ = Describe("all clone tests", func() {
 
 			f.ForceBindPvcIfDvIsWaitForFirstConsumer(dataVolume)
 
-			By("Verify Quota was exceeded in logs")
-			matchString := strings.Trim(fmt.Sprintf("\"namespace\": \"%s\", \"error\": \"pods \\\"cdi-upload-target-dv\\\" is forbidden: exceeded quota: test-quota, requested", targetNs.Name), " ")
```
alromeros, line 2239: Do we know if the new quota message still matches the `ErrQuotaExceeded` function that we use in the dv controllers?

go.mod
```diff
@@ -23,44 +21,44 @@ require (
...
-	k8s.io/cluster-bootstrap v0.0.0
-	k8s.io/code-generator v0.23.3
-	k8s.io/component-helpers v0.23.0
-	k8s.io/klog/v2 v2.70.1
-	k8s.io/kube-aggregator v0.23.0
-	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1
-	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
+	k8s.io/cluster-bootstrap v0.26.1
+	k8s.io/code-generator v0.26.1
+	k8s.io/component-helpers v0.26.1
```
akalenyu, line 51: can swap the copied funcs like https://github.com/kubevirt/containerized-data-importer/blob/167a90ec528aef7150bbb54778b1c7362f7a0e37/pkg/controller/datavolume/util.go#L388

tests/cloner_test.go
mhenriks: If you look at the file we still verify a condition and an event contain ErrExceededQuota

go.mod
akalenyu: we also copied CheckAccessModes :donthateme:

awels (commented):
/lgtm
/approve

PR #2690, opened 2023-04-12
Import populator

alromeros:

The import populator is a controller that handles the import of data in PVCs (still taking advantage of the import-controller flow) without the need for DataVolumes.

This controller creates an additional PVC' with import annotations. After the import process succeeds, the controller rebinds the PV to the original target PVC and deletes the PVC'.

**Special notes for your reviewer**:  Still need to commit the unit tests.

**Release note**:

```release-note
Include a new controller "import-populator" to populate PVCs with the standard CDI Import flow using a volume populator
```


alromeros:
@ShellyKa13 I cherry-picked some of your commits and tried to refactor common code so it's shared between populators in https://github.com/kubevirt/containerized-data-importer/pull/2690/commits/f51f2f7728aaa278a0ff74061442a5757593b237. Let me know what you think.

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	// Each populator reconciles the target PVC in a different way
+	if cc.IsUnbound(pvc) {
+		return populator.reconcileTargetPVC(pvc, pvcPrime)
+	}
+
+	// Making sure to clean PVC'
+	return r.reconcileCleanup(pvcPrime)
+}
```
mhenriks, line 287: maybe I missed it, but I want to make sure we are handling the case where the target pvc has deletion timestamp set.  We shouldn't create PVC' in that case.  But maybe no need to call cleanup if everything has has ownerrefs

mhenriks (commented):
Great start!  Here's my first quick pass...

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+		}
+	}
+
+	return pvcPrime, nil
+}
+
+func (r *ReconcilerBase) reconcileCleanup(pvcPrime *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
+	if pvcPrime != nil {
+		// Wait for the bind controller to rebind the PV
+		if pvcPrime.Status.Phase != corev1.ClaimLost {
```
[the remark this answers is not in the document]
alromeros: makes sense, are we sure we won't get unintended consequences? just asking before this behavior is also present in the [lib-volume-populator controller](https://github.com/kubernetes-csi/lib-volume-populator/blob/master/populator-machinery/controller.go#L684).

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -609,6 +609,90 @@ type DataImportCronList struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ImportSourceSpec `json:"spec"`
+	// +optional
+	Status ImportSourceStatus `json:"status,omitempty"`
+}
+
+// ImportSourceSpec defines the Spec field for ImportSource
+type ImportSourceSpec struct {
```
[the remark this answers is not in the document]
alromeros: Yeah I get your point... `VolumeImportSource` works for me but I'm easy to convince. I'll try to think of some names with @ShellyKa13 so we can further discuss them.

pkg/controller/populators/util.go
```diff
@@ -0,0 +1,100 @@
...
+	if ok {
+		return kind == cdiv1.UploadSourceRef
+	}
+	// TODO: Maybe use a more descriptive annotation?
+	_, ok = pvc.Annotations[cc.AnnSource]
+	if ok {
+		return kind == cdiv1.ImportSourceRef
+	}
+	return false
+}
```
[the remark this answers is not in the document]
alromeros: I agree, will try to improve this.

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+	const dataSourceRefField = "spec.dataSourceRef"
+
+	indexKeyFunc := func(namespace, name string) string {
+		return namespace + "/" + name
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &corev1.PersistentVolumeClaim{}, dataSourceRefField, func(obj client.Object) []string {
+		pvc := obj.(*corev1.PersistentVolumeClaim)
+		if IsPVCDataSourceRefKind(pvc, sourceKind) && pvc.Spec.DataSourceRef.Name != "" {
+			return []string{indexKeyFunc(obj.GetNamespace(), pvc.Spec.DataSourceRef.Name)}
```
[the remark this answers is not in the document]
alromeros: Thanks for the reminder, I forgot about that. Will take a look.

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+		return err
+	}
+
+	const dataSourceRefField = "spec.dataSourceRef"
+
+	indexKeyFunc := func(namespace, name string) string {
+		return namespace + "/" + name
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &corev1.PersistentVolumeClaim{}, dataSourceRefField, func(obj client.Object) []string {
```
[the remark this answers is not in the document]
alromeros: Even specifying the sourceKind? Why it would be a problem?

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func (r *ReconcilerBase) createPVCPrime(pvc *corev1.PersistentVolumeClaim, source client.Object, waitForFirstConsumer bool, pvcModifier pvcModifierFunc) (*corev1.PersistentVolumeClaim, error) {
+	labels := map[string]string{
+		common.CDILabelKey: common.CDILabelValue,
+	}
+	annotations := make(map[string]string)
+	annotations[cc.AnnImmediateBinding] = ""
+	if waitForFirstConsumer {
+		annotations[cc.AnnSelectedNodeName] = pvc.Annotations[AnnSelectedNode]
```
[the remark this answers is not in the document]
alromeros: Notice how I'm copying the `AnnSelectedNode` from target PVC into `AnnSelectedNodeName`  of PVC'. This annotation is then copied into the importer pod's `NodeName` in import-controller. I noticed your comment in Shelly's PR and tried improving this behavior, but I think I messed something up because it's causing the tests to fail.

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
mhenriks: I can't think of any unintended consequences, if the target is bound, our job is done I think

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
mhenriks: If the type (PVC) and field name (spec.dataSourceRef) passed to `IndexField` are the same then I'm pretty sure it will be a problem

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
alromeros: Once I get that to work I'll add proper comments and maybe change the annotation name so it's not as confusing.

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
alromeros: Thanks for the offline clarification Michael. Will fix this.

pkg/controller/populators/populator-base.go
alromeros: Any recomendation on how to handle this? Isn't enough to just check for pvc.DeletionTimestamp just after getting the pvc? For example:

```go
	// Get the target PVC
	pvc := &corev1.PersistentVolumeClaim{}
	if err := r.client.Get(context.TODO(), req.NamespacedName, pvc); err != nil {
		}
	}

        if pvc.DeletionTimestamp != nil {
                log.V(1).Info("PVC being terminated, ignoring it")
        }
```

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,335 @@
...
+			return nil, nil
+		}
+		return nil, err
+	}
+	return pv, nil
+}
+
+func (r *ReconcilerBase) handleStorageClass(pvc *corev1.PersistentVolumeClaim) (bool, bool, error) {
+	if pvc.Spec.StorageClassName == nil {
+		return false, false, nil
```
ShellyKa13, line 154: in this case dont we want to return ready true? currently we will just do nothing..

pkg/controller/populators/import-populator.go
```diff
@@ -0,0 +1,304 @@
...
+		r.recorder.Eventf(pvc, corev1.EventTypeNormal, importSucceeded, messageImportSucceeded, pvc.Name)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// Import-specific implementation of updateAnnotations
+func (r *ImportPopulatorReconciler) updateAnnotations(pvc *corev1.PersistentVolumeClaim, source client.Object) {
+	importSource := source.(*cdiv1.ImportSource)
+	annotations := pvc.Annotations
```
ShellyKa13, line 150: are we sure we want to copy all the target pvc annotations?

pkg/controller/populators/populator-base.go
alromeros: Good point, will change that

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,335 @@
...
+	return pvcPrime, nil
+}
+
+func (r *ReconcilerBase) reconcileCleanup(pvcPrime *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
+	if pvcPrime != nil {
+		if err := r.client.Delete(context.TODO(), pvcPrime); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
```
ShellyKa13, line 334: why delete the success event I had here?

alromeros: I moved it to `reconcileTargetPVC` so we can use specific events depending on the populator. Once the target PVC is bound I think it's safe to trigger the event.

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,335 @@
...
+		return nil, err
+	}
+
+	// If PVC' doesn't exist and target PVC is not bound, we should create the PVC' to start the population.
+	// We still return the nil PVC' as we'll get called again once PVC' exists.
+	// If target PVC is bound, we don't really need to populate anything.
+	if pvcPrime == nil && cc.IsUnbound(pvc) {
+		_, err := r.createPVCPrime(pvc, populationSource, waitForFirstConsumer, populator.updateAnnotations)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
alromeros: I moved it to createPVCPrime since in that function there's another creation-related event (in that way we have them in the same function).

pkg/controller/populators/import-populator.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+}
+
+// Progress reporting
+
+func (r *ImportPopulatorReconciler) updateImportProgress(podPhase string, pvc, pvcPrime *corev1.PersistentVolumeClaim) error {
+	if pvc.Annotations == nil {
+		pvc.Annotations = make(map[string]string)
+	}
+	// Just add 100.0% if pod is succeeded
```
ShellyKa13, line 245: nit pic "just *set* 100..." not add

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,335 @@
...
+		if k8serrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	// We first perform the common reconcile steps.
+	// We should only continue if we get a valid PVC'
+	pvcPrime, err := r.reconcileCommon(pvc, populator, log)
+	if err != nil {
```
ShellyKa13, line 276: technically you can check err != nil || pvcPrime == nil, then the err is nil and no need for the next if

pkg/controller/populators/populator-base.go
ShellyKa13: I think with the ownerRef then the suggestion is good

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,335 @@
...
+		}),
+	}
+	util.SetRecommendedLabels(pvcPrime, r.installerLabels, "cdi-controller")
+
+	if !isPVCOwnedByDataVolume(pvc) {
+		requestedSize, _ := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+		// disk or image size, inflate it with overhead
+		requestedSize, err := cc.InflateSizeWithOverhead(r.client, requestedSize.Value(), &pvc.Spec)
+		if err != nil {
+			return nil, err
```
ShellyKa13, line 218: since you moved the event to after the create fails should we add here also a log or event?

pkg/controller/populators/util.go
```diff
@@ -0,0 +1,108 @@
...
+	return dataSourceRef != nil && dataSourceRef.APIGroup != nil && *dataSourceRef.APIGroup == cc.AnnAPIGroup && dataSourceRef.Kind == kind
+}
+
+func isPVCPrimeDataSourceRefKind(pvc *corev1.PersistentVolumeClaim, kind string) bool {
+	owner := metav1.GetControllerOf(pvc)
+	if owner == nil || owner.Kind != "PersistentVolumeClaim" {
+		return false
+	}
+	populatorKind, _ := pvc.Annotations[AnnPopulatorKind]
+	switch populatorKind {
```
ShellyKa13, line 69: I think its better to use the consts UploaSourceRef and ImportSourceRef instead of new consts and just return populatorKind == kind

pkg/controller/populators/import-populator.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// Import-specific implementation of updateAnnotations
+func (r *ImportPopulatorReconciler) updateAnnotations(pvc *corev1.PersistentVolumeClaim, source client.Object) {
+	importSource := source.(*cdiv1.ImportSource)
+	annotations := pvc.Annotations
+	annotations[AnnPopulatorKind] = importKind
```
ShellyKa13, line 151: relates to previous comment just set cdiv1.ImportSourceRef

pkg/controller/import-controller.go
```diff
@@ -526,6 +523,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
 		scratchPvcName:    scratchPvcName,
 		vddkImageName:     vddkImageName,
 		priorityClassName: cc.GetPriorityClass(pvc),
+		selectedNodeName:  cc.GetSelectedNodeName(pvc),
```
ShellyKa13, line 526: dont really need to add another value to the args since you have the pvc in the args you can do the get when creating the pod

pkg/controller/populators/populator-base.go
alromeros: You are right, didn't notice that having the event after the function allowed us to trigger it with more error cases. Will move it back.

pkg/controller/populators/util.go
alromeros: Agree, will change it

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
alromeros: Fixed

pkg/controller/populators/import-populator.go
alromeros: I'll try to improve that.

pkg/controller/populators/util.go
```diff
@@ -43,9 +62,9 @@ func isPVCPrimeDataSourceRefKind(pvc *corev1.PersistentVolumeClaim, kind string)
 	if owner == nil || owner.Kind != "PersistentVolumeClaim" {
 		return false
 	}
-	_, ok := pvc.Annotations[cc.AnnUploadRequest]
-	if ok {
-		return kind == cdiv1.UploadSourceRef
+	populatorKind, _ := pvc.Annotations[AnnPopulatorKind]
+	if populatorKind == kind {
```
ShellyKa13, line 66: can just return populatorKind == kind

pkg/controller/populators/util.go
```diff
@@ -17,12 +32,16 @@ const (
 
 	// errCreatingPVCPrime provides a const to indicate we failed to create PVC prime for population
 	errCreatingPVCPrime = "ErrCreatingPVCPrime"
-	// createdPVCPrimeSucceesfully provides a const to indicate we created PVC prime for population
-	createdPVCPrimeSucceesfully = "CreatedPVCPrimeSucceesfully"
 
-	// AnnSelectedNode annotation is added to a PVC that has been triggered by scheduler to
-	// be dynamically provisioned. Its value is the name of the selected node.
-	AnnSelectedNode = "volume.kubernetes.io/selected-node"
```
ShellyKa13: still need in this commit this const and also the function GetSelectedNodeName

alromeros: Should be already :+1:

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,344 @@
...
+			Group:   "",
+			Version: "v1",
+			Kind:    "PersistentVolumeClaim",
+		}),
+	}
+	util.SetRecommendedLabels(pvcPrime, r.installerLabels, "cdi-controller")
+
+	// If the target PVC is owned by a DataVolume, we don't do anything since
+	// the size has been already inflated by the datavolume controller.
+	// TODO: I think it looks nicer to always inflate the size here.
```
mhenriks, line 243: Yeah ideally this layer does not know about DataVolumes.

pkg/controller/import-controller.go
```diff
@@ -1016,6 +1012,9 @@ func makeNodeImporterPodSpec(args *importerPodArgs) *corev1.Pod {
 	})
 
 	cc.SetRestrictedSecurityContext(&pod.Spec)
+	// We explicitly define a NodeName for dinamically provisioned PVCs
```
arnongilboa, line 1015: dynamically :)

pkg/controller/import-controller.go
```diff
@@ -1174,6 +1173,9 @@ func makeImporterPodSpec(args *importerPodArgs) *corev1.Pod {
 	}
 
 	cc.SetRestrictedSecurityContext(&pod.Spec)
+	// We explicitly define a NodeName for dinamically provisioned PVCs
```
arnongilboa, line 1176: dynamically :)

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+func (r *ReconcilerBase) rebindPV(targetPVC, pvcPrime *corev1.PersistentVolumeClaim) error {
+	pv, err := r.getPV(pvcPrime.Spec.VolumeName)
+	if pv == nil {
+		return err
+	}
+
+	// Examine the claimref for the PV and see if it's still bound to PVC'
+	claimRef := pv.Spec.ClaimRef
+	if claimRef.Name != pvcPrime.Name || claimRef.Namespace != pvcPrime.Namespace || claimRef.UID != pvcPrime.UID {
+		return nil
```
mhenriks, line 160: should we assert that claimRef is pointing to target here?

pkg/controller/populators/util.go
```diff
@@ -0,0 +1,94 @@
...
+	populatorKind, _ := pvc.Annotations[cc.AnnPopulatorKind]
+	return populatorKind == kind
+}
+
+// PVCPrimeName returns the name of the PVC' of a given pvc
+func PVCPrimeName(targetPVC *corev1.PersistentVolumeClaim) string {
+	return fmt.Sprintf("%s-%s", primePvcPrefix, targetPVC.UID)
+}
+
+func getPopulatorIndexKey(namespace, kind, name string) string {
```
mhenriks, line 70: what about apigroup?

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// CreateCommonPopulatorIndexes creates indexes used by all populators
+func CreateCommonPopulatorIndexes(mgr manager.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &corev1.PersistentVolumeClaim{}, dataSourceRefField, func(obj client.Object) []string {
+		pvc := obj.(*corev1.PersistentVolumeClaim)
+		dataSourceRef := pvc.Spec.DataSourceRef
+		if dataSourceRef != nil && dataSourceRef.APIGroup != nil &&
+			*dataSourceRef.APIGroup == cc.AnnAPIGroup && dataSourceRef.Name != "" {
+			return []string{getPopulatorIndexKey(obj.GetNamespace(), pvc.Spec.DataSourceRef.Kind, pvc.Spec.DataSourceRef.Name)}
```
mhenriks, line 77: since rebased on 1.26 should check `pvc.Spec.DataSourceRef.Namespace)` before using `pvc.Namespace`

pkg/controller/populators/populator-base.go
```diff
@@ -0,0 +1,337 @@
...
+const (
+	dataSourceRefField = "spec.dataSourceRef"
+)
+
+// Interface to store populator-specific methods
+type populatorController interface {
+	// Returns the specific populator CR
+	getPopulationSource(namespace, name string) (client.Object, error)
+	// Returns PVC' annotations according to the population source
+	updateAnnotations(pvc *corev1.PersistentVolumeClaim, source client.Object)
```
mhenriks, line 65: the callee can technically do anything with `pvc`, right?  So why call the function `updateAnnotations`?

mhenriks (commented):
Looking good!  more little stuff

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -618,6 +618,89 @@ type DataImportCronList struct {
...
+	Status VolumeUploadSourceStatus `json:"status,omitempty"`
+}
+
+// VolumeUploadSourceSpec defines specification for VolumeUploadSource
+type VolumeUploadSourceSpec struct {
+	ContentType DataVolumeContentType `json:"contentType,omitempty"`
+}
+
+// VolumeUploadSourceStatus provides the most recently observed status of the VolumeUploadSource
+type VolumeUploadSourceStatus struct {
```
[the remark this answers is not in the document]
alromeros: Or maybe we could include a map with all populated PVCs and their progress phase?

pkg/controller/common/util.go
```diff
@@ -917,6 +936,15 @@ func SetRestrictedSecurityContext(podSpec *v1.PodSpec) {
 	}
 }
 
+// SetNodeNameIfPopulator sets NodeName in a pod spec when the PVC is being handled by a CDI volume populator
+func SetNodeNameIfPopulator(pvc *corev1.PersistentVolumeClaim, podSpec *v1.PodSpec) {
+	_, isPopulator := pvc.Annotations[AnnPopulatorKind]
+	nodeName, _ := pvc.Annotations[AnnSelectedNode]
+	if isPopulator && nodeName != "" {
```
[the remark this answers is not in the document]
alromeros: So I previously tried other two things:
- Having an additional annotation in the PVC Prime  (`AnnSelectedNodeName`) and just copy that one into the pod NodeName. This failed because the populator controller needs to also set `AnnSelectedNode` into PVC' for it to work.
- Always copying `AnnSelectedNode` into the pod spec's NodeName. This caused lots of errors in the tests when a PVC created with the regular import flow had `AnnSelectedNode` (the DV stayed in ImportScheduled until timeout).

So I'm just copying AnnSelectedNode when the import-controller reconciles a PVC', which seems to work and doesn't mess with the standard import flow while avoiding creating another redundant annotation.

pkg/controller/common/util.go
```diff
@@ -1216,3 +1250,143 @@ func IsWaitForFirstConsumerEnabled(obj metav1.Object, gates featuregates.Feature
...
+		return retain == "true"
+	}
+	return false
+}
+
+// BuildHTTPClient generates an http client that accepts any certificate, since we are using
+// it to get prometheus data it doesn't matter if someone can intercept the data. Once we have
+// a mechanism to properly sign the server, we can update this method to get a proper client.
+func BuildHTTPClient(httpClient *http.Client) *http.Client {
+	if httpClient == nil {
```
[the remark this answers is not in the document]
alromeros: Yeah I'll take some time to improve this. I added the argument since we use a global `httpClient` in the dv controller and I didn't want to mess anything with moving the function to the common code (can't just use global variables as we were doing).

pkg/controller/populators/util.go
alromeros: When creating the index I check in the indexer function that `*dataSourceRef.APIGroup == cc.AnnAPIGroup` so I don't think it's necessary. I can move it to the key if preferred.

pkg/controller/populators/populator-base.go
alromeros: Planned to work on cross-namespace population in a separate PR but yeah I'll do it now.

pkg/controller/populators/populator-base.go
alromeros: This was to follow what we already do on the DV controller, but maybe something like `preparePVCForPopulation` or `setupPVCForPopulation` would be more descriptive.

pkg/controller/populators/populator-base.go
alromeros: If claimRef is pointing to target we would still return nil and everything would continue as expected, right? If that's the case I don't consider it necessary.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
mhenriks: Maintaining status on this resource in a non-racy way would mean creating a dedicated controller that reconciles `VolumeImportSource/VolumeUploadSource` whenever a PVC referencing populator cr is created/updated/deleted.

Or could single thread the current controller but probably don't want to do that

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -618,6 +618,89 @@ type DataImportCronList struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+type VolumeImportSource struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec VolumeImportSourceSpec `json:"spec"`
+	// +optional
+	Status VolumeImportSourceStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
alromeros: After removing `+optional` now it's mandatory to create a status in the yaml:
```
The VolumeImportSource "importsource" is invalid: status: Required value
```
In that case I think we should keep it.

pkg/controller/populators/populator-base.go
mhenriks: yes but if for some strange reason it is not pointing to the target, it is clearly a bad situation that will not fix itself and there is no indication given anywhere.  Why not log it?

pkg/controller/populators/util.go
mhenriks: I think it would be better to be in the key.  And not that it matters any, the key should probably be ordered in the same way as k8s api url `apigroup/kind/namespace/name`

pkg/controller/populators/populator-base.go
mhenriks: Or just `updatePVC` I don't really care

pkg/controller/common/util.go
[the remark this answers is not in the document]
mhenriks: I think it is unfortunate for this layer to know anything about populators.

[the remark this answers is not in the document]
alromeros: I agree but having another annotation exclusive to populators with the same info as AnnSelectedNode seemed bad too. I can change it though.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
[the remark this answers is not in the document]
mhenriks: in fairness, the "right" thing to do probably is to make the field a pointer if it really is optional.  But that is not the standard applied on many k8s resources including core types

pkg/controller/populators/util.go
alromeros: Good suggestion, I'll change it

pkg/controller/populators/populator-base.go
alromeros: Agree

alromeros: Added a log message to specify that.

pkg/controller/populators/populator-base.go
alromeros: Just committed some changes related to cross-namespace population. I've not been able to test them yet since only k8s v1.26 allows `CrossNamespaceVolumeDataSource` (kubevirtci still uses v1.25.7).

Another problem that I see is that there's no way to check if `CrossNamespaceVolumeDataSource` is enabled (still alpha), and there are more [prerequisites](https://kubernetes.io/blog/2023/01/02/cross-namespace-data-sources-alpha/#trying-it-out) than I expected, so we should probably just skip that test for the moment.

pkg/controller/populators/populator-base.go
alromeros: Updated it (ended up using `preparePVCForPopulation` but I can change it if preferred).

tests/import_test.go
```diff
@@ -1398,6 +1399,390 @@ var _ = Describe("Preallocation", func() {
...
+			// Make sure pvcPrime was deleted after upload population
+			_, err := f.FindPVC(pvcPrime.Name)
+			return err != nil && k8serrors.IsNotFound(err)
+		}, timeout, pollingInterval).Should(BeTrue())
+	})
+
+	// TODO: Only k8s from v1.26 will allow PVCs with namespaced DataSourceRef.
+	// We need to find a way to check if CrossNamespaceVolumeDataSource feature gate is enabled
+	// More information in: https://kubernetes.io/blog/2023/01/02/cross-namespace-data-sources-alpha/#trying-it-out
+	XIt("should import with VolumeImportSource from different namespace", func() {
```
mhenriks, line 1745: Supporting cross namespace is more than just enabling a couple feature gates.  The populator controller actually has to actually check the resource grants itself.  I say we get rid of this test for now

tests/import_test.go
```diff
@@ -1398,6 +1399,390 @@ var _ = Describe("Preallocation", func() {
...
+
+		By("Wait for PVC prime to be deleted")
+		Eventually(func() bool {
+			// Make sure pvcPrime was deleted after upload population
+			_, err := f.FindPVC(pvcPrime.Name)
+			return err != nil && k8serrors.IsNotFound(err)
+		}, timeout, pollingInterval).Should(BeTrue())
+	})
+})
+
```
mhenriks, line 1785: Can we add a test to make sure that the static binding works as expected?  Create a PV with claimref set for PVC.  Then create the PVC with populator.  Make sure the PVC gets bound correctly and data is not overwritten

pkg/controller/populators/import-populator_test.go
```diff
@@ -0,0 +1,422 @@
...
+	"kubevirt.io/containerized-data-importer/pkg/common"
+	. "kubevirt.io/containerized-data-importer/pkg/controller/common"
+	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
+)
+
+var (
+	dvPopulatorLog = logf.Log.WithName("import-populator-test")
+)
+
+var _ = Describe("Import populator tests", func() {
```
mhenriks, line 54: Maybe I missed it, but how about a unit test that ignores/errors cross namespace datasource?

mhenriks (commented):
Couple little things

tests/import_test.go
mhenriks: I think just adding having check is fine

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -618,6 +618,54 @@ type DataImportCronList struct {
...
+
+	Spec VolumeImportSourceSpec `json:"spec"`
+	// +optional
+	Status VolumeImportSourceStatus `json:"status"`
+}
+
+// VolumeImportSourceSpec defines the Spec field for VolumeImportSource
+type VolumeImportSourceSpec struct {
+	ContentType DataVolumeContentType `json:"contentType,omitempty"`
+	// Import sources
```
alromeros, line 639: @mhenriks want to know your opinion on this, do you think it's better to have the sources embedded directly in the spec? Or should we add a new field to store them as we do with DataVolumes.

mhenriks: Good question!  Yeah I think it makes sense to have this pseudo-union defined in it's own struct

mhenriks:
Great work @alromeros!

/approve

Let''s let @ShellyKa13 do a final once over and lgtm

pkg/controller/util.go
```diff
@@ -135,6 +135,13 @@ func newScratchPersistentVolumeClaimSpec(pvc *v1.PersistentVolumeClaim, pod *v1.
 			}
 		}
 	}
+	// When the original PVC is being handled by a populator, copy AnnSelectedNode to avoid issues with k8s scheduler
+	_, isPopulator := pvc.Annotations[cc.AnnPopulatorKind]
+	selectedNode := pvc.Annotations[cc.AnnSelectedNode]
+	if isPopulator && selectedNode != "" {
+		annotations[cc.AnnSelectedNode] = selectedNode
+	}
```
ShellyKa13, line 143: maybe also put in a function? big chunk of code for this function which is very specific to populators :)
not a must though..

tests/import_test.go
```diff
@@ -1392,6 +1393,420 @@ var _ = Describe("Preallocation", func() {
...
+		By("Forcing cleanup")
+		err = utils.DeleteVerifierPod(f.K8sClient, f.Namespace.Name)
+		Expect(err).ToNot(HaveOccurred())
+		err = f.CdiClient.CdiV1beta1().VolumeImportSources(f.Namespace.Name).Delete(context.TODO(), "import-populator-test", metav1.DeleteOptions{})
+		if err != nil && !k8serrors.IsNotFound(err) {
+			Expect(err).ToNot(HaveOccurred())
+		}
+		Eventually(func() bool {
+			// Make sure pvcPrime was deleted after upload population
```
ShellyKa13, line 1763: comment line wrong: "upload" population instead of import

ShellyKa13 (commented):
looks really good!
have several comments, mostly cosmetics, once fixed ill lgtm

pkg/controller/populators/import-populator_test.go
```diff
@@ -0,0 +1,432 @@
...
+
+			pod := CreateImporterTestPod(targetPvc, pvcPrime.Name, nil)
+			pod.Spec.Containers[0].Ports[0].ContainerPort = int32(port)
+			pod.Status.PodIP = ep.Hostname()
+			pod.Status.Phase = corev1.PodRunning
+
+			reconciler = createImportPopulatorReconciler(targetPvc, pvcPrime, pod)
+			err = reconciler.updateImportProgress(string(corev1.PodRunning), targetPvc, pvcPrime)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(targetPvc.Annotations[AnnImportProgressReporting]).To(BeEquivalentTo("13.45%"))
```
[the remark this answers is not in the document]
alromeros: Yeah I'm sure it's a github thing, looks good when you open the file.

pkg/controller/util.go
alromeros: hmmm I know I'm repeating some code here but I've tried some alternatives to reuse `SetNodeNameIfPopulator` and I think it looks uglier.

pkg/controller/populators/import-populator_test.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	// Import populator's DataSourceRef
+	apiGroup := AnnAPIGroup
+	dataSourceRef := &corev1.TypedObjectReference{
+		APIGroup: &apiGroup,
+		Kind:     cdiv1.VolumeImportSourceRef,
+		Name:     samplePopulatorName,
+	}
+	nsName := "test-import"
```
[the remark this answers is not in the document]
alromeros: This one needs to be a variable to do &nsName.

ShellyKa13:
great work!
/lgtm

PR #2691, opened 2023-04-12
Enable additional linters to golangci-lint configuration

nunnatsa:
Enable the following linters and fix their findings:
* `typecheck` - Like the front-end of a Go compiler, parses and type-checks Go code [no findings]
* `staticcheck` - (megacheck): It's a set of rules from staticcheck. It's not the same thing as the staticcheck binary. The author of staticcheck doesn't support or approve the use of staticcheck as a library inside golangci-lint. [many findings]
* `ineffassign` - Detects when assignments to existing variables are not used [1 finding]

***Note***: *Some findings were not fixed, as they need to be handled by the CDI team. These findings are disabled by the `nolint` comment + `TODO`*

```release-note
None
```

### DIFF
--- a/pkg/controller/clone-controller_test.go
+++ b/pkg/controller/clone-controller_test.go
@@ -438,7 +438,7 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		reconciler.shortTokenValidator.(*cc.FakeValidator).Params["targetName"] = "testPvc1"
 		By("Verifying no source pod exists")
 		sourcePod, err := reconciler.findCloneSourcePod(testPvc)
-		Expect(err).To(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred())
 		Expect(sourcePod).To(BeNil())
 		_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "testPvc1", Namespace: "default"}})
 		Expect(err).To(HaveOccurred())
--- a/pkg/controller/clone-controller_test.go
+++ b/pkg/controller/clone-controller_test.go
@@ -123,10 +123,12 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		By("Verifying no source pod exists")
 		sourcePod, err := reconciler.findCloneSourcePod(testPvc)
 		Expect(sourcePod).To(BeNil())
+		Expect(err).ToNot(HaveOccurred())
 		_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "testPvc1", Namespace: "default"}})
 		Expect(err).ToNot(HaveOccurred())
 		By("Verifying no source pod exists")
 		sourcePod, err = reconciler.findCloneSourcePod(testPvc)
+		Expect(err).ToNot(HaveOccurred())
 		Expect(sourcePod).To(BeNil())
 		By("Verifying the PVC now has a source pod name")
 		err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "testPvc1", Namespace: "default"}, testPvc)
@@ -152,6 +154,7 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		reconciler.shortTokenValidator.(*cc.FakeValidator).Params["targetName"] = "testPvc1"
 		By("Verifying no source pod exists")
 		sourcePod, err := reconciler.findCloneSourcePod(testPvc)
+		Expect(err).ToNot(HaveOccurred())
 		Expect(sourcePod).To(BeNil())
 		result, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "testPvc1", Namespace: "default"}})
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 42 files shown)


pkg/controller/clone-controller_test.go
```diff
@@ -701,7 +709,7 @@ func createCloneReconciler(objects ...runtime.Object) *CloneReconciler {
 	cdiv1.AddToScheme(s)
 	rec := record.NewFakeRecorder(1)
 	// Create a fake client to mock API calls.
-	cl := fake.NewFakeClientWithScheme(s, objs...)
+	cl := fake.NewFakeClientWithScheme(s, objs...) //nolint:staticcheck // todo: BuildNameToCertificate() is deprecated - check this
```
mhenriks, line 712: copy/paste error, comment should say `NewFakeClientWithScheme`

nunnatsa: Done

pkg/importer/http-datasource.go
```diff
@@ -154,6 +154,9 @@ func (hs *HTTPDataSource) Transfer(path string) (ProcessingPhase, error) {
 			return ProcessingPhaseError, err
 		}
 		size, err := util.GetAvailableSpace(path)
+		if err != nil {
```
akalenyu, line 157: I think this is why tests fail, we rely on `ErrInvalidPath` to tell us scratch space is needed

akalenyu (commented):
awesome

pkg/importer/http-datasource.go
nunnatsa: But this value assignment of the `err` var is never checked. Maybe there is a hidden bug here that this check finds.

akalenyu: I mean we should definitely keep the err check, just need to find a way to return `ErrInvalidPath` when needed
since we rely on it https://github.com/kubevirt/containerized-data-importer/blob/aef617e7c9de2c641dc949cd8109d23231406cf6/pkg/importer/http-datasource.go#L160-L163

nunnatsa: got you. and since `GetAvailableSpace` returns `-1` in case of error, you get the same behaviour. What do you think about combining the conditions? e.g.
```suggestion
		size, err := util.GetAvailableSpace(path)
		if err != nil || size <= 0 {
			//Path provided is invalid.
			return ProcessingPhaseError, ErrInvalidPath
		}
```

nunnatsa: I'm not sure. Maybe you can tidy that later, but this will not change the current behavior. https://github.com/kubevirt/containerized-data-importer/blob/f9c049f0e481d4697dfa9444db8aa64620a17e91/pkg/util/util.go#L102-L109

For any error in `GetAvailableSpace`, the function returns -1, so the the old `if size <= int64(0)` check, bahaved the same.

akalenyu: Yep right

akalenyu:

The nfs failures are known flakes, but it seems like there's a legit failure in the destructive lane

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -340,7 +340,9 @@ func (vmware *VMwareClient) FindDiskInSnapshotTree(snapshots []types.VirtualMach
 		if disk := vmware.FindDiskInSnapshot(snapshot.Snapshot, fileName); disk != nil {
 			return disk
 		}
-		return vmware.FindDiskInSnapshotTree(snapshot.ChildSnapshotList, fileName)
+		if disk := vmware.FindDiskInSnapshotTree(snapshot.ChildSnapshotList, fileName); disk != nil {
```
akalenyu, line 343: What was the linter complaining about here? or is this just about consistency?

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -717,7 +719,8 @@ func (sink *VDDKFileSink) ZeroRange(offset uint64, length uint32) error {
 	if sink.isBlock { // Try to punch a hole in block device destination
 		err = punch(offset, length)
 	} else {
-		info, err := sink.file.Stat()
```
akalenyu, line 720: What is the error for this?

pkg/importer/vddk-datasource_amd64.go
nunnatsa: @akalenyu - The linter warning was:
```
SA4004: the surrounding loop is unconditionally terminated (staticcheck)
```

And I think it's an actual bug, because before this change, the function only search for the disk in the first snapshot's ChildSnapshotList, and exited anyway, even if not found. The original version never search in more than one snapshot in the slice. 

As I'm not familiar with the domain, and the slice is always with length of 1, then maybe it would better to rewrite this function. But for now, just by reading golang code - this is a bug.

akalenyu: > before this change, the function only search for the disk in the first snapshot's ChildSnapshotList

Isn't it fine since it's recursive?

pkg/importer/vddk-datasource_amd64.go
nunnatsa: The problem is in line 720. this line creates a shadow err variable. This variable is never checked and actually ignored.

https://github.com/kubevirt/containerized-data-importer/blob/f9c049f0e481d4697dfa9444db8aa64620a17e91/pkg/importer/vddk-datasource_amd64.go#L716-L730

The linter created two warnings for that:
```
pkg/importer/vddk-datasource_amd64.go:725:5: ineffectual assignment to err (ineffassign)
                                err = syscall.Ftruncate(int(sink.file.Fd()), int64(offset+uint64(length)))
                                ^
pkg/importer/vddk-datasource_amd64.go:727:5: ineffectual assignment to err (ineffassign)
                                err = punch(offset, length)
```

pkg/importer/vddk-datasource_amd64.go
nunnatsa: Let's say you have 3 snapshots in the slice, each with 3 children. And let's say that the required disk is the third child of the third snapshot.

The function will start checking the first snapshot. The first condition will be false so it will start checking the children. First the first child, and the first condition is false again, then it returns nil, since the first child does not have children. But since it's a `return`, the rest of the children are not checked. now we again at the original call. And again, since it's a call with `return`, the 2nd and the 3rd snapshot are not checked.

PR #2693, opened 2023-04-13
[release-v1.49] backport tar handling from #2410

mhenriks:

May be helpful to backport new tar behavior to avaoid EPERM

Fixes #

https://bugzilla.redhat.com/show_bug.cgi?id=2184386

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cdi-cloner/clone-source.go
+++ b/cmd/cdi-cloner/clone-source.go
@@ -146,12 +146,38 @@ func validateMount() {
 }
 
 func newTarReader(preallocation bool) (io.ReadCloser, error) {
-	args := "cv"
+	excludeMap := map[string]struct{}{
+		"lost+found": struct{}{},
+	}
+
+	args := []string{"/usr/bin/tar", "cv"}
 	if !preallocation {
 		// -S is used to handle sparse files. It can only be used when preallocation is not requested
-		args = "S" + args
+		args = append(args, "-S")
 	}
-	cmd := exec.Command("/usr/bin/tar", args, ".")
+
+	files, err := os.ReadDir(mountPoint)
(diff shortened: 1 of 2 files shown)


mhenriks:
> Back in 1.49, we are still running as root, so that err is a bit surprising no?
> 
> https://github.com/kubevirt/containerized-data-importer/blob/c8f96f9585aa36902f49cfb554336fafdd86fcc6/pkg/controller/clone-controller.go#L592
> 
> But this does remind me of an nfs root squash misconfiguration which we recently heard about, will try to find that discussion

Yeah, that may be it but there is a detailed trace in the bz.  this change should keep that error from occurring.  not sure if some other error will pop up though

akalenyu:
> > Back in 1.49, we are still running as root, so that err is a bit surprising no?
> > https://github.com/kubevirt/containerized-data-importer/blob/c8f96f9585aa36902f49cfb554336fafdd86fcc6/pkg/controller/clone-controller.go#L592
> > 
> > But this does remind me of an nfs root squash misconfiguration which we recently heard about, will try to find that discussion
> 
> Yeah, that may be it but there is a detailed trace in the bz. this change should keep that error from occurring. not sure if some other error will pop up though

gotcha yeah it should not happen, was just wondering how we hit it in the first place,
being root. I guess when the squash is messed up it just doesn't matter that you're running as root in the container.

/lgtm
/test pull-cdi-generate-verify-1.49

akalenyu:
Apparently the generate error has been around for a while
(failed in https://github.com/kubevirt/containerized-data-importer/pull/2659)
@awels @brianmcarey Any idea what can cause this on the 1.49 lanes?

PR #2696, opened 2023-04-19
golangci-lint: Enable errcheck

nunnatsa:
Also, fix the usage of golangci-lint to actually block merges, and fix some new findings from other linters.

```release-note
None
```


mhenriks:
Thanks @nunnatsa!

/lgtm
/approve

PR #2700, opened 2023-04-25
Allow snapshots as format for DataImportCron created sources

akalenyu:
Today DataImportCron maintains up-to-date PVCs that serve as base image sources for cloning,
However, for certain storage types, we know that snapshots provide a better fit.
Snapshots are also immutable which makes them better suited to act as up-to-date OS image sources.

This PR attempts to allow DataImportCrons to manage a snapshot as the source format instead of DataVolume/PVC

Fixes #

**Release note**:

```release-note
Allow snapshots as a format for DataImportCron created sources
```


pkg/controller/dataimportcron-controller.go
```diff
@@ -420,12 +456,35 @@ func (r *DataImportCronReconciler) getImportState(ctx context.Context, cron *cdi
...
+		if !k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+		snapshot = nil
+	}
+
+	return snapshot, nil
+}
+
+func (r *DataImportCronReconciler) updateUnderlyingSourceObject(ctx context.Context, cron *cdiv1.DataImportCron, obj client.Object) error {
```
arnongilboa, line 482: mmm... isn't updateSource() good enough?

akalenyu: yeah I didn't feel so good about this too.. updateSourceObj should be better

pkg/controller/dataimportcron-controller.go
```diff
@@ -311,13 +313,36 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	dataImportCronCopy := dataImportCron.DeepCopy()
 	imports := dataImportCron.Status.CurrentImports
 	importSucceeded := false
+
+	dataVolume := dataImportCron.Spec.Template
+	explicitScName := getStorageClassFromTemplate(&dataVolume)
+	dvStorageClass, err := cc.GetStorageClassByName(r.client, explicitScName)
+	if err != nil {
+		return res, err
```
akalenyu, line 321: @arnongilboa having second thoughts about this wrt to https://github.com/kubevirt/containerized-data-importer/pull/2683. maybe better to take the sc name from the created pvc?

pkg/controller/common/util.go
```diff
@@ -1473,3 +1474,72 @@ func UpdateImageIOAnnotations(annotations map[string]string, imageio *cdiv1.Data
...
+
+	targetStorageClass, err := GetStorageClassByName(client, targetPvcStorageClassName)
+	if err != nil {
+		return "", err
+	}
+	if targetStorageClass == nil {
+		logger.Info("Target PVC's Storage Class not found")
+		return "", nil
+	}
+
```
arnongilboa, line 1495: any clue why there was another `Fetch the source storage class` before the func moved here?

pkg/controller/dataimportcron-controller.go
arnongilboa: If you want to support the case the DV SC is not found yet, or it's using the default which is not set yet, taking `sc name from the created pvc` is sure the safest, but does it fit your code?

tests/utils/volumesnapshot.go
```diff
@@ -0,0 +1,40 @@
...
+		Spec: snapshotv1.VolumeSnapshotSpec{
+			Source: snapshotv1.VolumeSnapshotSource{
+				PersistentVolumeClaimName: &sourcePvcName,
+			},
+			VolumeSnapshotClassName: snapshotClassName,
+		},
+	}
+}
+
+// WaitSnapshotReady waits until the snapshot is ready to be used
```
arnongilboa, line 30: In all your calls to this func you are Get()ting again the VolumeSnapshot, so why not return it here?

pkg/controller/dataimportcron-controller.go
```diff
@@ -327,7 +352,16 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
 	} else if pvc != nil {
-		if err = r.updatePvc(ctx, dataImportCron, pvc); err != nil {
+		// TODO: IsPopulated check? wait for corev1.Bound?
+		if err = r.updateSourceObject(ctx, dataImportCron, pvc); err != nil {
+			return res, err
+		}
+		importSucceeded = true
+		if err := r.handleCronFormat(ctx, dataImportCron, format, &dvStorageClass.Name); err != nil {
+			return res, err
+		}
```
arnongilboa, line 362: This block and `case cdiv1.Succeeded` are similar, so maybe refactor or handle them in one place?

pkg/controller/dataimportcron-controller.go
```diff
@@ -704,24 +941,48 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
+		scName := obj.GetName()
+		for _, cron := range crons.Items {
+			dataVolume := cron.Spec.Template
+			explicitScName := getStorageClassFromTemplate(&dataVolume)
+			templateSc, err := cc.GetStorageClassByName(mgr.GetClient(), explicitScName)
+			if err != nil || templateSc == nil {
+				c.GetLogger().Error(err, "Unable to get storage class", "templateSc", templateSc)
+				return
+			}
+			if templateSc.Name == scName {
```
arnongilboa, line 966: why not annotate the DIC with templateSc.Name in the reconcile (you are already getting it there), so you use an index to filter the List() and get rid of the loop.

pkg/controller/dataimportcron-controller.go
```diff
@@ -730,6 +991,22 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 	); err != nil {
 		return err
 	}
+
+	if err := c.Watch(&source.Kind{Type: &cdiv1.StorageProfile{}},
+		handler.EnqueueRequestsFromMapFunc(mapStorageProfileToCron),
+		predicate.Funcs{
+			CreateFunc: func(event.CreateEvent) bool { return false },
```
arnongilboa, line 998: are you sure create is not interesting?

pkg/controller/common/util.go
akalenyu: guess a leftover from the refactoring of https://github.com/kubevirt/containerized-data-importer/blob/release-v1.34/pkg/controller/datavolume-controller.go#L1160

tests/utils/volumesnapshot.go
akalenyu: I think it may be a little unexpected for this util to do something other than return the struct

pkg/controller/dataimportcron-controller.go
akalenyu: I can do that, but, the storage profile watch going off should be a rare occurrence, may not be worth the index?
not sure what's the overhead other than managing an annotation

pkg/controller/dataimportcron-controller.go
akalenyu: yep it is, for example, new storage class that is default being created. thx!

pkg/controller/dataimportcron-controller.go
akalenyu: done

pkg/controller/dataimportcron-controller.go
akalenyu: yeah pvc may be racy, I went with just proceeding with DV creation as before.
added a unit test for this behavior

pkg/controller/dataimportcron-controller.go
```diff
@@ -311,13 +314,40 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
+	if err != nil {
+		return res, err
+	}
+	snapshot, err := r.getSnapshot(ctx, dataImportCron, format)
+	if err != nil {
+		return res, err
+	}
+
+	handleDataVolumePopulated := func() error {
+		if err = r.updateSourceObject(ctx, dataImportCron, pvc); err != nil {
```
arnongilboa77, line 334: You mix `err = ` and `err := ` in several ifs here. I think the later is better.

pkg/controller/dataimportcron-controller.go
```diff
@@ -327,7 +357,12 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionFalse, fmt.Sprintf("Import DataVolume phase %s", dvPhase), dvPhase)
 		}
 	} else if pvc != nil {
-		if err = r.updatePvc(ctx, dataImportCron, pvc); err != nil {
+		// TODO: with plain populator PVCs (no DataVolumes) we may need to wait for corev1.Bound
+		if err := handleDataVolumePopulated(); err != nil {
+			return res, err
+		}
+	} else if snapshot != nil {
+		if err = r.updateSourceObject(ctx, dataImportCron, snapshot); err != nil {
```
arnongilboa77, line 365: again, better `err := `.

pkg/controller/dataimportcron-controller.go
```diff
@@ -420,12 +457,35 @@ func (r *DataImportCronReconciler) getImportState(ctx context.Context, cron *cdi
...
+		if !k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+		snapshot = nil
+	}
+
+	return snapshot, nil
+}
+
+func (r *DataImportCronReconciler) updateSourceObject(ctx context.Context, cron *cdiv1.DataImportCron, obj client.Object) error {
```
arnongilboa77, line 483: Not sure the word `Object` adds anything. All update functions update an object :)

pkg/controller/dataimportcron-controller.go
```diff
@@ -557,22 +634,161 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+		}
+	default:
+		return fmt.Errorf("unknown source format for snapshot")
+	}
+
+	return nil
+}
+
+func (r *DataImportCronReconciler) getSourceFormat(ctx context.Context, dataImportCron *cdiv1.DataImportCron, dvStorageClass *storagev1.StorageClass) (cdiv1.DataImportCronSourceFormat, error) {
+	format := cdiv1.DataImportCronSourceFormatPvc
```
arnongilboa77, line 767: Isn't Type/Kind better than format?

pkg/controller/dataimportcron-controller.go
```diff
@@ -420,12 +457,35 @@ func (r *DataImportCronReconciler) getImportState(ctx context.Context, cron *cdi
...
+		return nil, nil
+	}
+
+	snapName := imports[0].DataVolumeName
+	snapshot := &snapshotv1.VolumeSnapshot{}
+	if err := r.client.Get(ctx, types.NamespacedName{Namespace: cron.Namespace, Name: snapName}, snapshot); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+		snapshot = nil
```
arnongilboa77, line 477: why no explicitly `return nil, nil` ?

pkg/controller/dataimportcron-controller.go
```diff
@@ -311,13 +314,40 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
+	format, err := r.getSourceFormat(ctx, dataImportCron, dvStorageClass)
+	if err != nil {
+		return res, err
+	}
+	snapshot, err := r.getSnapshot(ctx, dataImportCron, format)
+	if err != nil {
+		return res, err
+	}
+
+	handleDataVolumePopulated := func() error {
```
arnongilboa77, line 333: maybe `handlePopulatedPvc`?

pkg/controller/dataimportcron-controller.go
arnongilboa77: Afaiu indexing won't cost too much as their values will be static and there won't too many DICs.

pkg/controller/dataimportcron-controller.go
```diff
@@ -557,22 +634,161 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
 			return err
 		}
 	} else {
-		if err := r.updatePvc(ctx, dataImportCron, pvc); err != nil {
+		if err := r.updateSourceObject(ctx, dataImportCron, snapshot); err != nil {
 			return err
 		}
+		// If snapshot exists don't create DV
+		return nil
 	}
```
arnongilboa77, line 663: Similar handling for pvc/snapshot, so either use a helper func or iterate over them.

pkg/controller/dataimportcron-controller.go
```diff
@@ -311,13 +314,40 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
+	if err != nil {
+		return res, err
+	}
+
+	handleDataVolumePopulated := func() error {
+		if err = r.updateSourceObject(ctx, dataImportCron, pvc); err != nil {
+			return err
+		}
+		importSucceeded = true
+		if err := r.handleCronFormat(ctx, dataImportCron, format, dvStorageClass); err != nil {
```
arnongilboa77, line 338: why do we need this helper (used only here)? if Snapshot just call r.handleSnapshotFormat() directly

akalenyu: just trying to be more welcoming towards new formats, I thought it's nice to have a structured entrypoint

arnongilboa77: so maybe move the handleDataVolumePopulated logic to handleCronFormat and name it accordingly

pkg/controller/dataimportcron-controller.go
akalenyu: We have actually discussed this in the sig-storage meeting and came up with format;
I think type/kind are too general in this context

pkg/controller/dataimportcron-controller.go
akalenyu: handlePopulatedPvc (coming soon) doesn't belong there imo, why is this so bad?

pkg/controller/dataimportcron-controller.go
akalenyu: done

pkg/controller/dataimportcron-controller.go
akalenyu: yep, changed to `err :=` throughout the controller

pkg/controller/dataimportcron-controller.go
akalenyu: removed

pkg/controller/dataimportcron-controller.go
akalenyu: better. thx

pkg/controller/dataimportcron-controller.go
akalenyu: changed

pkg/controller/dataimportcron-controller.go
akalenyu: done

pkg/controller/dataimportcron-controller.go
akalenyu: Reverted the index with a TODO;
As we offline discussed, we may lose the storage profile event, as relying on the storage class annotation may get racy:
Even if we set up the indexer before the controller, we would still not have the annotation yet.

Kept the annotation so we could capitalize on it later when we switch to indexing here

tests/utils/volumesnapshot.go
arnongilboa: so why not pass `snapshot` instead of `snap` to `Get()` save the additional `Get()` after this func is called?

tests/dataimportcron_test.go
```diff
@@ -382,6 +562,85 @@ var _ = Describe("DataImportCron", func() {
 			return errors.IsNotFound(err)
 		}, dataImportCronTimeout, pollingInterval).Should(BeTrue(), "cronjob first job pod was not deleted")
 	})
+
+	Context("Change source format of existing DataImportCron", func() {
+		It("[test_id:XXXX] Should switch from PVC to snapshot sources once format changes", func() {
```
duyanyan, line 567: [test_id:10034]

tests/dataimportcron_test.go
```diff
@@ -271,36 +428,59 @@ var _ = Describe("DataImportCron", func() {
...
+					break
+				}
+			}
+			Expect(found).To(BeTrue())
 		}
-		Expect(pvcFound).To(BeTrue())
-	})
+	},
+		table.Entry("[test_id:7406] with PVC sources", cdiv1.DataImportCronSourceFormatPvc),
+		table.Entry("[test_id:XXXX] with snapshot sources", cdiv1.DataImportCronSourceFormatSnapshot),
```
duyanyan, line 482: [test_id:10033]

tests/dataimportcron_test.go
```diff
@@ -208,34 +346,25 @@ var _ = Describe("DataImportCron", func() {
...
-			}, dataImportCronTimeout, pollingInterval).Should(BeTrue(), "PVCs were not deleted")
+			verifyDeletion(format)
 		}
 	},
-		table.Entry("[test_id:7403] succeed importing initial PVC from registry URL", true, false, 1),
-		table.Entry("[test_id:7414] succeed importing PVC from registry URL on source digest update", true, false, 2),
-		table.Entry("[test_id:8266] succeed deleting error DVs when importing new ones", false, true, 2),
+		table.Entry("[test_id:7403] succeed importing initial PVC from registry URL", true, false, 1, cdiv1.DataImportCronSourceFormatPvc),
+		table.Entry("[test_id:7414] succeed importing PVC from registry URL on source digest update", true, false, 2, cdiv1.DataImportCronSourceFormatPvc),
+		table.Entry("[test_id:XXXX] succeed importing initially into a snapshot from registry URL", true, false, 1, cdiv1.DataImportCronSourceFormatSnapshot),
```
duyanyan, line 356: [test_id:10031]

tests/dataimportcron_test.go
```diff
@@ -208,34 +346,25 @@ var _ = Describe("DataImportCron", func() {
...
+		table.Entry("[test_id:XXXX] succeed importing to a snapshot from registry URL on source digest update", true, false, 2, cdiv1.DataImportCronSourceFormatSnapshot),
```
duyanyan, line 357: [test_id:10032]

pkg/controller/dataimportcron-controller.go
arnongilboa, line 1: You have several funcs with `switch format` for Pvc/Snapshot: `handleCronFormat()`, `updateDataSource()`, `updateDataImportCronSuccessCondition()`. Consider using a map with format key to struct with func pointers.

pkg/controller/dataimportcron-controller.go
arnongilboa: You may append in `reqs` the crons with the expected scName annotation, plus the crons without the scName annotation yet, and it solves the race afaiu.

akalenyu:
@arnongilboa @duyanyan just rebase for now to make sure nothing breaks, addressing comments later

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -97,6 +99,21 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
...
+	"openshift-storage.rbd.csi.ceph.com": cdiv1.DataImportCronSourceFormatSnapshot,
+}
+
+// CloneStrategyByProvisionerKey defines the advised clone strategy for a provisioner
+var CloneStrategyByProvisionerKey = map[string]cdiv1.CDICloneStrategy{
+	"csi-vxflexos.dellemc.com":   cdiv1.CloneStrategyCsiClone,
+	"csi-isilon.dellemc.com":     cdiv1.CloneStrategyCsiClone,
+	"csi-powermax.dellemc.com":   cdiv1.CloneStrategyCsiClone,
+	"csi-powerstore.dellemc.com": cdiv1.CloneStrategyCsiClone,
+}
```
arnongilboa, line 115: please add also:
```
"hspc.csi.hitachi.com": cdiv1.CloneStrategyCsiClone,
"csi.hpe.com": cdiv1.CloneStrategyCsiClone,
```

akalenyu: done

pkg/controller/dataimportcron-controller.go
```diff
@@ -557,22 +639,150 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
-		dv := r.newSourceDataVolume(dataImportCron, dvName)
-		if err := r.client.Create(ctx, dv); err != nil && !k8serrors.IsAlreadyExists(err) {
+		cc.AddAnnotation(desiredSnapshot, AnnLastUseTime, time.Now().UTC().Format(time.RFC3339Nano))
+		if err := r.client.Create(ctx, desiredSnapshot); err != nil {
 			return err
 		}
 	} else {
-		if err := r.updatePvc(ctx, dataImportCron, pvc); err != nil {
-			return err
+		if currentSnapshot.Status != nil && currentSnapshot.Status.ReadyToUse != nil && *currentSnapshot.Status.ReadyToUse {
```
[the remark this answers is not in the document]
akalenyu: added

pkg/controller/dataimportcron-controller.go
```diff
@@ -557,22 +639,150 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+	case cdiv1.DataImportCronSourceFormatPvc:
+		return nil
+	case cdiv1.DataImportCronSourceFormatSnapshot:
+		return r.handleSnapshotFormat(ctx, dataImportCron, &dataImportCron.Spec.Template, dvStorageClass)
+	default:
+		return fmt.Errorf("unknown source format for snapshot")
+	}
+}
+
+func (r *DataImportCronReconciler) handleSnapshotFormat(ctx context.Context, dataImportCron *cdiv1.DataImportCron, dataVolume *cdiv1.DataVolume, dvStorageClass *storagev1.StorageClass) error {
```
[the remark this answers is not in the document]
akalenyu: yeah sure

pkg/controller/dataimportcron-controller.go
akalenyu: we can and it does, but I don't think it's much better than the current approach
a little bit more inversed logically, still no index, still List on all of them

tests/utils/volumesnapshot.go
akalenyu: ohh you are talking about `WaitSnapshotReady`, thought you meant `NewVolumeSnapshot`
I think mutating the pointer is a little weird, maybe have this func return latest snap state?

pkg/controller/dataimportcron-controller.go
```diff
@@ -557,22 +639,150 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+	case cdiv1.DataImportCronSourceFormatSnapshot:
+		snapshot := &snapshotv1.VolumeSnapshot{}
+		err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: pvcName}, snapshot)
+		if err != nil {
+			if !k8serrors.IsNotFound(err) {
+				return err
+			}
+			// Snapshot creation will trigger reconcile
+			return nil
+		}
```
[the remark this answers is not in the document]
akalenyu: because this updates the status I think it makes more sense that it comes to conclusions on it's own based on it's own observation of the cluster

pkg/controller/dataimportcron-controller.go
akalenyu: sounds good, will see if there are any surprises doing this

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
arnongilboa: I don't think that getting it again is a good practice. Even in the much larger DV controller we are getting all the CRs once in `sync()`, and once in `updateStatus()`.

pkg/controller/dataimportcron-controller.go
arnongilboa: Ok, not sure it's the best choice, but I can accept your argument...

tests/utils/volumesnapshot.go
arnongilboa: We don't mutate the pointer, we just referesh it's refered CR state by Get(..., snapshot). The line above it will sure be deleted.

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
akalenyu: Yes, I am getting it in the "updateStatus" portion of the data import cron controller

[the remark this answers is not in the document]
akalenyu: @awels Hey

We were hoping to get your thoughts on this one as the tiebreaker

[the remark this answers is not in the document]
awels: I am with Arnon on this one. If the status changed between us looking up the value earlier, and updating the status. Then that should trigger a new reconcile since we are watching the snapshotvolumes, and the next reconcile will properly update the condition to what it should be.

pkg/controller/dataimportcron-controller.go
```diff
@@ -558,22 +643,150 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
 	if err != nil {
 		return err
 	}
-	// If PVC exists don't create DV
-	pvc := &corev1.PersistentVolumeClaim{}
-	if err = r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: dvName}, pvc); err != nil {
+	dataImportCron.Status.CurrentImports = []cdiv1.ImportStatus{{DataVolumeName: dvName, Digest: digest}}
+
+	sources := []client.Object{&corev1.PersistentVolumeClaim{}, &snapshotv1.VolumeSnapshot{}}
```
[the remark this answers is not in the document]
akalenyu: Done

pkg/controller/dataimportcron-controller.go
akalenyu: Yeah, I have a working copy of this but I am hesitant to introduce it to this PR now, it's a pretty big change..
Would it be ok if I pursue this on a subsequent PR?

PR #2703, opened 2023-04-28
enable ginkgolinter and fix findings

nunnatsa:
All fixes were done automatically by the ginkgolinter cli.

```release-note
None
```

### DIFF
--- a/pkg/controller/dataimportcron-controller_test.go
+++ b/pkg/controller/dataimportcron-controller_test.go
@@ -124,7 +124,7 @@ var _ = Describe("All DataImportCron Tests", func() {
 			if dataSource != nil {
 				imports := cron.Status.CurrentImports
 				Expect(imports).ToNot(BeNil())
-				Expect(len(imports)).ToNot(BeZero())
+				Expect(imports).ToNot(BeEmpty())
 				dvName := imports[0].DataVolumeName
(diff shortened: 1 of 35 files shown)


pkg/controller/dataimportcron-controller_test.go
```diff
@@ -439,7 +439,7 @@ var _ = Describe("All DataImportCron Tests", func() {
 				Expect(err).ToNot(HaveOccurred())
 				ts := pvc.Annotations[AnnLastUseTime]
 				Expect(ts).ToNot(BeEmpty())
-				Expect(ts > lastTs).To(BeTrue())
+				Expect(strings.Compare(ts, lastTs)).Should(Equal(1))
```
arnongilboa, line 442: is it better?

nunnatsa: Currently, there is a bug in ginkgolinter (solved in v0.11.2) that replaces
```go
Expect(ts > lastTs).To(BeTrue())
```
with
```go
Expect(ts).To(BeNumerically(">", lastTs))
```

So I want to avoid false positive here.

In addition, as far as I understand, the best practice for lexicographic comparisons in go, is using  `strings.Compare()`.

nunnatsa:
rebased

PR #2704, opened 2023-04-28
Fix multiple clone functional test

awels:
The test was failing due to a type mismatch in the comparison (int vs int32).
The destructive lane is failing due to a cdi pod not comping back up in time. Added some extra logging to determine why

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -1021,7 +1021,7 @@ var _ = Describe("all clone tests", func() {
 							Expect(err).ToNot(HaveOccurred())
 							restartCount := cloner.Status.ContainerStatuses[0].RestartCount
 							fmt.Fprintf(GinkgoWriter, "INFO: restart count on clone source pod %s: %d\n", clonerPodName, restartCount)
-							Expect(restartCount).To(BeNumerically("<", 2))
+							Expect(restartCount).To(BeNumerically("<", int32(2)))
 						}
 					}
 
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -1770,30 +1770,29 @@ var _ = Describe("Import populator", func() {
 		verifyCleanup(pvc)
 
 		By("Making PV available")
-		Eventually(func() bool {
+		Eventually(func() v1.PersistentVolumePhase {
 			pv, err := f.K8sClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 			Expect(pv.Spec.ClaimRef.Namespace).To(Equal(pvc.Namespace))
 			Expect(pv.Spec.ClaimRef.Name).To(Equal(pvc.Name))
-			if pv.Status.Phase == v1.VolumeAvailable {
-				return true
+			if pv.Status.Phase != v1.VolumeAvailable {
+				pv.Spec.ClaimRef.ResourceVersion = ""
+				pv.Spec.ClaimRef.UID = ""
+				_, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
 			}
-			pv.Spec.ClaimRef.ResourceVersion = ""
-			pv.Spec.ClaimRef.UID = ""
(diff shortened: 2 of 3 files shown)


tests/cloner_test.go
```diff
@@ -1021,7 +1021,7 @@ var _ = Describe("all clone tests", func() {
 							Expect(err).ToNot(HaveOccurred())
 							restartCount := cloner.Status.ContainerStatuses[0].RestartCount
 							fmt.Fprintf(GinkgoWriter, "INFO: restart count on clone source pod %s: %d\n", clonerPodName, restartCount)
-							Expect(restartCount).To(BeNumerically("<", 2))
+							Expect(restartCount).To(BeNumerically("<", int32(2)))
```
akalenyu, line 1024: if you look at the failure https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2701/pull-containerized-data-importer-e2e-ceph/1651557222200643584
the issue here is that 2 is too high for this test to pass
Expected <int32>: 2 **to be <** <int>: 2

We have a PR pending that disables this check:
https://github.com/kubevirt/containerized-data-importer/pull/2647#issuecomment-1512887512

tests/operator_test.go
```diff
@@ -1131,8 +1131,11 @@ func waitCDI(f *framework.Framework, cr *cdiv1.CDI, cdiPods *corev1.PodList) {
 	By("Waiting for there to be as many CDI pods as before")
 	Eventually(func() bool {
 		newCdiPods = getCDIPods(f)
-		By(fmt.Sprintf("number of cdi pods: %d\n new number of cdi pods: %d\n", len(cdiPods.Items), len(newCdiPods.Items)))
-		return len(cdiPods.Items) == len(newCdiPods.Items)
+		fmt.Fprintf(GinkgoWriter, "number of cdi pods: %d\n new number of cdi pods: %d\n", len(cdiPods.Items), len(newCdiPods.Items))
+		for _, pod := range newCdiPods.Items {
+			fmt.Fprintf(GinkgoWriter, "pod %s/%s\n", pod.Namespace, pod.Name)
```
akalenyu, line 1136: awesome!

tests/import_test.go
```diff
@@ -1770,30 +1770,29 @@ var _ = Describe("Import populator", func() {
...
 			Expect(err).ToNot(HaveOccurred())
 			Expect(pv.Spec.ClaimRef.Namespace).To(Equal(pvc.Namespace))
 			Expect(pv.Spec.ClaimRef.Name).To(Equal(pvc.Name))
-			if pv.Status.Phase == v1.VolumeAvailable {
-				return true
+			if pv.Status.Phase != v1.VolumeAvailable {
+				pv.Spec.ClaimRef.ResourceVersion = ""
+				pv.Spec.ClaimRef.UID = ""
+				_, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
+				Expect(err).ToNot(HaveOccurred())
```
akalenyu, line 1782: Expecting inside an eventually block may fail the test earlier than we want to
to get around this you can do this:
https://github.com/kubevirt/containerized-data-importer/pull/2696/files#diff-e3e6528c8b5d4860f4fd89d38f0e0babb9cc0e9147c0b0b28be532e5e692c182R99

tests/cloner_test.go
awels: The exactly failure I was seeing is that the type didn't match `int` vs `int32` so I just cast it.

tests/import_test.go
awels: So I didn't change that particular thing, I just cleaned up the return value a bit so it is easier to see what actually failed. This expect if it return false it is because the update failed. Which should fail the test right there.

tests/cloner_test.go
akalenyu: the matcher should play along with that, the full output looks like:
```
{ Failure /root/go/src/kubevirt.io/containerized-data-importer/tests/cloner_test.go:965
Expected
    <int32>: 2
to be <
    <int>: 2
/root/go/src/kubevirt.io/containerized-data-importer/tests/cloner_test.go:1024}
```

PR #2705, opened 2023-04-28
Fix import populator static binding test

alromeros:

This Pull Request fixes a race condition in a test that checks for static binding in the import populator. We shouldn't worry about the PVC' being created in that case since the target PVC will be bound immediately afterward.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -1792,13 +1792,17 @@ var _ = Describe("Import populator", func() {
 		err = createHTTPImportPopulatorCR(cdiv1.DataVolumeKubeVirt, true)
 		Expect(err).ToNot(HaveOccurred())
 
-		By("Verify PVC prime is not created anymore")
-		_, err = f.FindPVC(populators.PVCPrimeName(pvc))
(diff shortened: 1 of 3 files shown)


tests/import_test.go
```diff
@@ -1792,10 +1792,6 @@ var _ = Describe("Import populator", func() {
 		err = createHTTPImportPopulatorCR(cdiv1.DataVolumeKubeVirt, true)
 		Expect(err).ToNot(HaveOccurred())
 
-		By("Verify PVC prime is not created anymore")
-		_, err = f.FindPVC(populators.PVCPrimeName(pvc))
-		Expect(k8serrors.IsNotFound(err)).To(BeTrue())
-
 		By("Verify target PVC is bound")
 		err = utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, pvc.Namespace, v1.ClaimBound, pvc.Name)
```
mhenriks, line 1796: Can we also verify that pvc is bound to `pvName`?

PR #2712, opened 2023-05-10
Propagate CDI workload NodePlacement to DIC CronJob and initial Job

arnongilboa:

Fixes bz#2196429

**Release note**:
```release-note
Propagate CDI workload NodePlacement to DataImportCron CronJob and initial Job
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -53,15 +53,12 @@ import (
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/common"
-
 	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
 	cdv "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
 	"kubevirt.io/containerized-data-importer/pkg/monitoring"
 	"kubevirt.io/containerized-data-importer/pkg/operator"
 	"kubevirt.io/containerized-data-importer/pkg/util"
 	"kubevirt.io/containerized-data-importer/pkg/util/naming"
-
-	"kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/controller/dataimportcron-controller_test.go
```diff
@@ -690,7 +690,7 @@ func createDataImportCronReconciler(objects ...runtime.Object) *DataImportCronRe
 
 func createDataImportCronReconcilerWithoutConfig(objects ...runtime.Object) *DataImportCronReconciler {
 	crd := &extv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: "dataimportcrons.cdi.kubevirt.io"}}
-	objs := []runtime.Object{crd}
+	objs := []runtime.Object{crd, cc.MakeEmptyCDICR()}
```
akalenyu, line 693: Maybe unit test that existing cronjob (carry over from upgrade etc) gets reconciled with placement values?

arnongilboa: added

pkg/controller/dataimportcron-controller.go
```diff
@@ -734,49 +730,54 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
 	}
-	desired, err := r.newCronJob(cron)
-	if err != nil {
-		return false, err
-	}
-	current, err := sdk.StripStatusFromObject(&cronJob)
-	if err != nil {
-		return false, err
-	}
-	sdk.MergeLabelsAndAnnotations(desired, current)
```
akalenyu, line 750: I see you got rid of all the SDK merging utils - are we ok with that?

akalenyu: the concern is that the deepequal will be false most of the time due to status changes/omitted spec fields

arnongilboa: No nned for the SDK merging utils. Regarding the `DeepEqual` you were right. Fixed that.

awels: Would it make sense to DeepEqual the spec instead of the entire object? Or are the annotations/labels we are interested in?

pkg/controller/dataimportcron-controller.go
```diff
@@ -813,6 +813,10 @@ func (r *DataImportCronReconciler) newCronJob(cron *cdiv1.DataImportCron) (*batc
 		volumes = append(volumes, createConfigMapVolume(ProxyCertVolName, volName))
 	}
 
+	if len(volumes) == 0 {
```
awels, line 816: instead of `volumes := []corev1.Volume{}` above you can do `var volumes []corev1.Volume` which will make it nil. Append works fine with a nil pointer. Then you don't have to do this.

awels (commented):
looks good one minor comment

pkg/controller/dataimportcron-controller.go
arnongilboa: We DeepEqual the whole object as we are intersted also in annotations/labels which may update on new version. Our CronJobs are not reconciled that often, so it won't cost too much to be on the safe side.

pkg/controller/dataimportcron-controller.go
arnongilboa: right:) fixed.

PR #2722, opened 2023-05-21
Integration of Data volume using CDI populators

ShellyKa13:
This PR integrates the use of datavolumes with import and upload populators. In case of having CSI storage instead of using the current path of creating a PVC with import/upload annotation which gets bound before the actual population, we create a PVC that has datasourceref pointing to the appropriate populator (import,upload) and create a volumeuploadsource/volumeimportsource with the right values according to the datavolume. In that case the appropriate populator with be triggered and the population with PVC' process will initiate which will result in having the target PVC bound only after the population succeeded.
This feature is crucial for our product to work seamlessly with DR solutions.
This is a valuable feature also when using datavolumes with VMs that are WFFC which will prevent the need for the current need of "doppelganger" pod to make the PVC bound before starting the VM.

added some TODOs I wish the reviewers to address
Still missing UTs

**Release note**:

```release-note
Integration of Data volume using CDI populators with CSI storage
```


pkg/controller/datavolume/controller-base.go
```diff
@@ -1130,3 +1155,15 @@ func (r *ReconcilerBase) storageClassCSIDriverExists(storageClassName *string) (
...
 }
+
+func (r *ReconcilerBase) checkPVCUsingPopulators(pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	if pvc.Spec.DataSourceRef == nil {
+		return false, nil
+	}
+	return r.storageClassCSIDriverExists(pvc.Spec.StorageClassName)
+}
+
+func (r *ReconcilerBase) usePopulatorsForPopulation(syncState *dvSyncState) (bool, error) {
```
alromeros, line 1175: Maybe change this function name to something like `shouldUseCDIPopulator`, `shouldUsePopulator`, or `checkPopulatorRequirements`? I think usePopulatorsForPopulation can sound odd.

pkg/controller/datavolume/upload-controller.go
```diff
@@ -91,6 +97,19 @@ func NewUploadController(
 	return datavolumeController, nil
 }
 
+func (r *UploadReconciler) updatePVCForPopulation(dataVolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) error {
+	if dataVolume.Spec.Source.Upload == nil {
```
alromeros, line 101: Same as with import.

alromeros (commented):
Good job! First quick review with some comments. Also wondering if we should skip fs overhead inflation in `renderPvcSpecVolumeSize` when using a populator.

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -130,11 +130,6 @@ func (r *SnapshotCloneReconciler) prepare(syncState *dvSyncState) error {
 	if err := r.populateSourceIfSourceRef(dv); err != nil {
 		return err
 	}
-	if dv.Status.Phase == cdiv1.Succeeded {
-		if err := r.cleanup(syncState); err != nil {
-			return err
-		}
-	}
```
mhenriks, line 137: surprised to see changes to this controller and pvc clone controller.  Please elaborate.  Concerrned about the removal of calls to cleanup()

pkg/controller/datavolume/controller-base.go
mhenriks: Agree

pkg/controller/datavolume/import-controller.go
```diff
@@ -160,23 +184,50 @@ func (r *ImportReconciler) sync(log logr.Logger, req reconcile.Request) (dvSyncR
...
 func (r *ImportReconciler) syncImport(log logr.Logger, req reconcile.Request) (dvSyncState, error) {
-	syncState, syncErr := r.syncCommon(log, req, nil, nil)
+	syncState, syncErr := r.syncCommon(log, req, r.cleanup, nil)
 	if syncErr != nil || syncState.result != nil {
 		return syncState, syncErr
 	}
-	if err := r.handlePvcCreation(log, &syncState, r.updateAnnotations); err != nil {
+
+	pvcModifier := r.updateAnnotations
+	if syncState.usePopulator {
```
mhenriks, line 193: @alromeros  What about DV with VDDK source, I don't think we can use populator for that (currently)?  No checkpoint support

pkg/controller/datavolume/import-controller.go
```diff
@@ -434,3 +485,63 @@ func (r *ImportReconciler) getNextCheckpoint(dataVolume *cdiv1.DataVolume, pvc *
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      importSourceName,
+			Namespace: dv.Namespace,
+		},
+		Spec: cdiv1.VolumeImportSourceSpec{
+			Source:        source,
+			ContentType:   dv.Spec.ContentType,
+			Preallocation: dv.Spec.Preallocation,
+		},
+	}
```
mhenriks, line 526: no owner ref?  if not using a finalizer on the dv, have to make sure it is cleaned up if dv is deleted

pkg/controller/datavolume/import-controller.go
```diff
@@ -160,23 +184,50 @@ func (r *ImportReconciler) sync(log logr.Logger, req reconcile.Request) (dvSyncR
...
 		syncErr = r.maybeSetPvcMultiStageAnnotation(syncState.pvc, syncState.dvMutated)
 	}
 	return syncState, syncErr
 }
 
+func (r *ImportReconciler) cleanup(syncState *dvSyncState) error {
+	dv := syncState.dvMutated
+	// This cleanup should be done if dv is marked for deletion, dv is succeeded
+	// and data volume is using the import populator
+	if dv.DeletionTimestamp == nil && dv.Status.Phase != cdiv1.Succeeded && !syncState.usePopulator {
```
mhenriks, line 216: this seems complicated to me.  If the CR is owned by the dv, then only explicitly delete when the target PVC is bound?  Seems simpler to me

pkg/controller/datavolume/controller-base.go
```diff
@@ -1125,8 +1156,35 @@ func (r *ReconcilerBase) storageClassCSIDriverExists(storageClassName *string) (
...
 	}
 
 	return true, nil
 }
+
+func (r *ReconcilerBase) checkPVCUsingPopulators(pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	if pvc.Spec.DataSourceRef == nil {
+		return false, nil
+	}
+	return r.storageClassCSIDriverExists(pvc.Spec.StorageClassName)
```
mhenriks, line 1172: rather than look up the driver again, is there an annotation on the pvc we can safely rely on?

pkg/controller/datavolume/controller-base.go
```diff
@@ -429,13 +430,14 @@ func (r *ReconcilerBase) syncDvPvcState(log logr.Logger, req reconcile.Request,
...
-				return syncState, err
-			}
+	if cleanup != nil {
+		if err := cleanup(&syncState); err != nil {
+			return syncState, err
 		}
+	}
+
+	if dv.DeletionTimestamp != nil {
+		log.Info("DataVolume marked for deletion")
```
mhenriks, line 440: can you provide some context for these changes?  seems weird to always call cleanup.  To me cleanup should only be called when the operation completes or the deleted while in progress.  If ther is come other callback required for populators, maybe make it explicit?

pkg/controller/datavolume/controller-base.go
```diff
@@ -796,10 +810,27 @@ func (r ReconcilerBase) updateStatus(req reconcile.Request, phaseSync *statusPha
 				if err != nil {
 					return reconcile.Result{}, err
 				}
-				if shouldBeMarkedWaitForFirstConsumer {
```
mhenriks, line 799: I'm wondering if the `shouldBeMarkedWaitForFirstConsumer` should be updated.  I'm not sure it should return true if the nodeName annotation has been added (at least for populators).  We should be able to start reporting the phases of pvc' at that point.  I don't think they'll get reported at all otherwise, well until the target is bound which is when succeeded

mhenriks (commented):
Good start, I think I've given you enough to think about for now

pkg/controller/datavolume/import-controller.go
alromeros: The VDDK source can still be used but I'm not too sure if these annotations are relevant without multi-sage import.

alromeros: Yeah you are right, currently sources with multi-stage import (VDDK and ImageIO using checkpoint API) should continue using the old flow. But I guess we can use the populator with regular VDDK and ImageIO sources if they lack multi-stage annotations.

pkg/controller/datavolume/import-controller.go
```diff
@@ -106,6 +111,25 @@ func addDataVolumeImportControllerWatches(mgr manager.Manager, datavolumeControl
 	return nil
 }
 
+func (r *ImportReconciler) updatePVCForPopulation(dataVolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) error {
+	if dataVolume.Spec.Source.HTTP != nil &&
```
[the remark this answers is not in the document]
ShellyKa13: I saw that there is such check in every pvcmodifier function so I did the same.. in the updateAnnotation you go over all of this and at the end if the source of none of those there is an error returned..

pkg/controller/datavolume/controller-base.go
```diff
@@ -679,6 +680,11 @@ func (r *ReconcilerBase) reconcileProgressUpdate(datavolume *cdiv1.DataVolume, p
 		datavolume.Status.Progress = "N/A"
 	}
 
+	if progress, ok := pvc.Annotations[cc.AnnImportProgressReporting]; ok {
```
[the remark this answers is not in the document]
ShellyKa13: if the pvc is using populator it should have this annotation, I guess we can add the check for the populators for the time in between the creation of the pvc' and the update of the progress on it

pkg/controller/datavolume/upload-controller.go
ShellyKa13: same answer this check is done also in updateAnnotations, wanted to keep it the same

pkg/controller/populators/import-populator.go
```diff
@@ -220,9 +246,6 @@ func (r *ImportPopulatorReconciler) updateImportProgress(podPhase string, pvc, p
 	if progressReport != "" {
 		if f, err := strconv.ParseFloat(progressReport, 64); err == nil {
 			pvc.Annotations[cc.AnnImportProgressReporting] = fmt.Sprintf("%.2f%%", f)
-			if err := r.client.Update(context.TODO(), pvc); err != nil {
```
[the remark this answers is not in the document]
ShellyKa13: Ill think how to make it better

pkg/controller/datavolume/snapshot-clone-controller.go
ShellyKa13: you can read the commit message that did this change, it explains why. if not clear after that let me know

pkg/controller/datavolume/import-controller.go
ShellyKa13: I do make sure it is deleted in the cleanup phase, but I guess adding owner ref wont hurt

pkg/controller/datavolume/import-controller.go
ShellyKa13: it is redundant to do the delete if we don't use populators, and phase succeded is basically the same as target PVC is bound, so you say to remove the deletionTimestamp check when I add the ownerRef, then sure I guess we can

pkg/controller/datavolume/import-controller.go
ShellyKa13: so need to also add here a check if len(dataVolume.Spec.Checkpoints) > 0 ? is that a good enough check?

pkg/controller/datavolume/controller-base.go
ShellyKa13: please read the first commit message

pkg/controller/datavolume/import-controller.go
alromeros: I think that should be enough but let me review the API in case I'm missing something.

pkg/controller/datavolume/controller-base.go
ShellyKa13: ok will do something else

pkg/controller/datavolume/snapshot-clone-controller.go
mhenriks: It's not the end of the world, but I think it would be better if the callback functions were dumb.  They just do a thing.  and the framework makes sure it is called at the right time.  Maybe there can be a `done` callback that is called when the DV is succeeded and `cleanup` when it's deleted?

pkg/controller/datavolume/util.go
```diff
@@ -297,12 +304,70 @@ func resolveVolumeSize(c client.Client, dvSpec cdiv1.DataVolumeSpec, pvcSpec *v1
...
+		if err != nil {
+			return false, err
+		}
+		return boolUsePopulator, nil
+	}
+	usePopulator, err := storageClassCSIDriverExists(client, log, storageClass)
+	if err != nil {
+		return false, err
+	}
+	cc.AddAnnotation(dv, cc.AnnUsePopulator, strconv.FormatBool(usePopulator))
```
mhenriks, line 383: I don't love that this function has the side effect of mutating the datavolume, perhaps it would be better to explicitly set this somehow in one place?  Currently, this function is called in multiple places

pkg/controller/populators/import-populator.go
```diff
@@ -210,15 +204,23 @@ func updateImportSourceAnnotation(pvc, pvcPrime *corev1.PersistentVolumeClaim) {
 	pvc.Annotations[cc.AnnSource] = cc.GetSource(pvcPrime)
 }
 
+func (r *ImportPopulatorReconciler) updateImportAnnotations(pvc, pvcPrime *corev1.PersistentVolumeClaim) {
+	phase := pvcPrime.Annotations[cc.AnnPodPhase]
+	if err := r.updateImportProgress(phase, pvc, pvcPrime); err != nil {
+		//TODO: currently I don't return err in case of failing to report progress, opinion?
+		// Also this will print error every 2 secs
+		klog.Errorf("Failed to update import progress for pvc %s/%s", pvc.Namespace, pvc.Name)
```
mhenriks, line 212: why klog and not the logr.Logger is my first question?

If we requeue every 2secs I guess it isn't a big deal to not handle errors but maybe would be nice

hack/build/run-linters.sh
```diff
@@ -4,4 +4,4 @@ GOLANGCI_VERSION="${GOLANGCI_VERSION:-v1.52.2}"
 
 go install "github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_VERSION}"
 
-golangci-lint run --timeout=8m ./...
+golangci-lint run --timeout=16m ./...
```
mhenriks, line 7: I did exactly the same thing in #2709

ShellyKa13: I know, just wanted to make the cdi-lint lane pass until your PR will get merged

pkg/controller/populators/import-populator.go
ShellyKa13: I didn't want to return an error cause I think we shouldn't prevent the phases to be updated just because there might be an issue with the progress reporting.
regarding the klog, I'll fix

pkg/controller/datavolume/snapshot-clone-controller.go
ShellyKa13: I think how it was before that the prepare called cleanup was yuck.
I think in the cleanup you do different cleanups according to different conditions, so I think the fact that you can decide in the cleanup callback itself when and what should be cleaned is good.. don't think that adding another done callback that will be called when dv is succeeded and will do the exact same thing as the cleanup is valuable.

pkg/controller/datavolume/util.go
ShellyKa13: I guess I can do the update itself on a line after the call to this line and only in one place in the syncDvPvcState function after updating syncState.usePopulator

hack/build/run-linters.sh
mhenriks: I meant this in a "great minds think alike" kind of way

pkg/controller/datavolume/import-controller.go
```diff
@@ -434,3 +487,72 @@ func (r *ImportReconciler) getNextCheckpoint(dataVolume *cdiv1.DataVolume, pvc *
...
+	} else if vddk := dv.Spec.Source.VDDK; vddk != nil {
+		source.VDDK = vddk
+	} else {
+		// Our dv shouldn't be without source
+		// Defaulting to Blank source
+		source.Blank = &cdiv1.DataVolumeBlankImage{}
+	}
+
+	importSourceName := volumeImportSourceName(dv)
+	importSource := &cdiv1.VolumeImportSource{
```
mhenriks, line 518: Totally forgot that the controller should be watching everything it creates, so should be watching `VolumeImportSources` owned by datavolumes.

Also, I think it would be better to see if the resource exists before creating, saves a trip to the apiserver, still have to handle already exists because cache is not perfect

pkg/controller/populators/import-populator.go
```diff
@@ -188,6 +187,14 @@ func (r *ImportPopulatorReconciler) updatePVCForPopulation(pvc *corev1.Persisten
 	annotations[cc.AnnSource] = cc.SourceNone
 }
 
+func deleteBoundConditionIfNeeded(pvc, pvcPrime *corev1.PersistentVolumeClaim) {
```
mhenriks, line 190: What is the purpose of this?

pkg/controller/populators/populator-base.go
```diff
@@ -225,17 +226,14 @@ type updatePVCAnnotationsFunc func(pvc, pvcPrime *corev1.PersistentVolumeClaim)
...
-	}
-	if preallocation, ok := pvcPrime.Annotations[cc.AnnPreallocationApplied]; ok {
-		cc.AddAnnotation(pvcCopy, cc.AnnPreallocationApplied, preallocation)
+	for k, v := range pvcPrime.GetAnnotations() {
+		if strings.Contains(k, common.CDIAnnKey) &&
+			!strings.Contains(k, cc.AnnImmediateBinding) &&
+			!strings.Contains(k, cc.AnnPopulatorKind) &&
+			!strings.Contains(k, "upload") &&
+			!strings.Contains(k, "import") {
+			cc.AddAnnotation(pvcCopy, k, v)
```
mhenriks, line 235: Would it be best to simply have a whitelist of allowed annotations?  What's your opinion?

ShellyKa13: it seems there are too much allowed annotations comparing to annotations that shouldn't be added.. and that's how I missed a bunch, I think its easier to know which we don't want to add

mhenriks: Too much?  Could probably count all the annotations we care about on two hands?  Maybe three?  They're all defined and unlikely to change.  Would be nice to have one place (or maybe one place for each source) that states "these are the annotations we care about for datavolumes".  I also clearly have a preference for whitelists over blacklists.

pkg/controller/datavolume/controller-base.go
```diff
@@ -694,6 +694,11 @@ func (r *ReconcilerBase) reconcileProgressUpdate(datavolume *cdiv1.DataVolume, p
 		return nil
 	}
 
+	// With upload we don't have report of progress and no need to requeue
+	if datavolume.Spec.Source != nil && datavolume.Spec.Source.Upload != nil {
```
mhenriks, line 700: Should base controller really be doing upload specific stuff?  Can it be moved to upload controller?

mhenriks: Or something added to import?

pkg/controller/datavolume/import-controller.go
```diff
@@ -103,6 +112,31 @@ func addDataVolumeImportControllerWatches(mgr manager.Manager, datavolumeControl
...
+}
+
+func (r *ImportReconciler) updatePVCForPopulation(dataVolume *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim) error {
+	if dataVolume.Spec.Source.HTTP != nil &&
+		dataVolume.Spec.Source.S3 != nil &&
+		dataVolume.Spec.Source.GCS != nil &&
+		dataVolume.Spec.Source.Registry != nil &&
+		dataVolume.Spec.Source.Imageio != nil &&
+		dataVolume.Spec.Source.VDDK != nil &&
+		dataVolume.Spec.Source.Blank != nil {
```
mhenriks, line 131: shouldn't all these `!=` be `==`?

pkg/controller/populators/populator-base.go
```diff
@@ -220,6 +213,31 @@ func (r *ReconcilerBase) createPVCPrime(pvc *corev1.PersistentVolumeClaim, sourc
 	return pvcPrime, nil
 }
 
+type updatePVCAnnotationsFunc func(pvc, pvcPrime *corev1.PersistentVolumeClaim)
+
+var desiredAnnotations = []string{cc.AnnPodPhase, cc.AnnPodReady, cc.AnnPodRestarts, cc.AnnPreallocationRequested, cc.AnnPreallocationApplied, cc.AnnRunningCondition, cc.AnnRunningConditionMessage, cc.AnnRunningConditionReason, cc.AnnBoundCondition, cc.AnnBoundConditionMessage, cc.AnnBoundConditionReason}
```
mhenriks, line 218: If I understand this correctly, the dv will have bound condition true if pvc' is bound.  That does not seem right to me

pkg/controller/datavolume/import-controller.go
ShellyKa13: totally right :)

pkg/controller/populators/populator-base.go
ShellyKa13: the update boundcondition on the dv doesnt update the condition to true unless the pvc is bound: 
https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume/conditions.go#L128-L143
so seems to me its ok.

mhenriks: Why bother passing over the annotations then?

ShellyKa13: I remember there was some tests checking this. I can remove this update and see how it affects our tests if it bothers you.

ShellyKa13:
/hold
hold until PR in kubevirt it merged: https://github.com/kubevirt/kubevirt/pull/9812#issuecomment-1578881464

pkg/controller/datavolume/import-controller.go
```diff
@@ -434,3 +502,82 @@ func (r *ImportReconciler) getNextCheckpoint(dataVolume *cdiv1.DataVolume, pvc *
...
+	importSource = &cdiv1.VolumeImportSource{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      importSourceName,
+			Namespace: dv.Namespace,
+		},
+		Spec: cdiv1.VolumeImportSourceSpec{
+			Source:        source,
+			ContentType:   dv.Spec.ContentType,
+			Preallocation: dv.Spec.Preallocation,
+		},
```
aglitke, line 552: Is this getting owned by the DV so that it will be cleaned up in case the DV is deleted before population succeeds?

pkg/controller/datavolume/controller-base.go
```diff
@@ -1109,24 +1169,57 @@ func (r *ReconcilerBase) handlePvcCreation(log logr.Logger, syncState *dvSyncSta
...
+	if usePopulator, ok := dv.Annotations[cc.AnnUsePopulator]; ok {
+		boolUsePopulator, err := strconv.ParseBool(usePopulator)
+		if err != nil {
+			return false, err
+		}
+		return boolUsePopulator, nil
+	}
+	// currently populators don't support retain pod annotation so don't use populators in that case
+	if retain := dv.Annotations[cc.AnnPodRetainAfterCompletion]; retain == "true" {
+		return false, nil
```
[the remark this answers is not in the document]
ShellyKa13: sure makes sense

pkg/controller/datavolume/import-controller.go
```diff
@@ -434,3 +502,82 @@ func (r *ImportReconciler) getNextCheckpoint(dataVolume *cdiv1.DataVolume, pvc *
 
 	return nil
 }
+
+func volumeImportSourceName(dv *cdiv1.DataVolume) string {
+	return fmt.Sprintf("%s-%s", volumeImportSourcePrefix, dv.UID)
```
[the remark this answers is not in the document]
ShellyKa13: we use the dv uid and a fix prefix so will be fixed size always

pkg/controller/datavolume/import-controller.go
ShellyKa13: I believe so does the PVC and no need to the import source if no pvc to populate

pkg/controller/datavolume/upload-controller.go
```diff
@@ -163,3 +233,59 @@ func (r *UploadReconciler) updateStatusPhase(pvc *corev1.PersistentVolumeClaim,
...
+	uploadSourceName := volumeUploadSourceName(syncState.dvMutated)
+	uploadSource := &cdiv1.VolumeUploadSource{}
+	exists, err := cc.GetResource(context.TODO(), r.client, syncState.dvMutated.Namespace, uploadSourceName, uploadSource)
+	if err != nil {
+		return err
+	}
+	if exists {
+		// uploadSource already exists
+		return nil
+	}
```
arnongilboa, line 251:
```suggestion
	if exists, err := cc.GetResource(...);  err != nil || exists {
		return err
	}
```

pkg/controller/datavolume/upload-controller.go
```diff
@@ -163,3 +233,59 @@ func (r *UploadReconciler) updateStatusPhase(pvc *corev1.PersistentVolumeClaim,
...
+		Spec: cdiv1.VolumeUploadSourceSpec{
+			ContentType:   syncState.dv.Spec.ContentType,
+			Preallocation: syncState.dv.Spec.Preallocation,
+		},
+	}
+
+	if err := controllerutil.SetControllerReference(syncState.dvMutated, uploadSource, r.scheme); err != nil {
+		return err
+	}
+	if err := r.client.Create(context.TODO(), uploadSource); err != nil {
```
arnongilboa, line 267: err != nil && !k8serrors.IsAlreadyExists(err)

pkg/controller/datavolume/upload-controller.go
```diff
@@ -163,3 +233,59 @@ func (r *UploadReconciler) updateStatusPhase(pvc *corev1.PersistentVolumeClaim,
...
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		if err := r.client.Delete(context.TODO(), uploadSource); err != nil {
+			if !k8serrors.IsNotFound(err) {
+				return err
+			}
+		}
+	}
```
arnongilboa, line 288: Sama as before, `Get()` is not needed before `Delete()` and after removing it use `err != nil && !k8serrors.IsAlreadyExists(err)`

pkg/controller/datavolume/util.go
```diff
@@ -303,6 +305,73 @@ func resolveVolumeSize(c client.Client, dvSpec cdiv1.DataVolumeSpec, pvcSpec *v1
 	return &requestedSize, err
 }
 
+// storageClassCSIDriverExists returns true if the passed storage class has CSI drivers available
+func storageClassCSIDriverExists(client client.Client, log logr.Logger, storageClassName *string) (bool, error) {
+	log = log.WithName("getCsiDriverForStorageClass").V(3)
```
arnongilboa, line 310: wrong logged name

pkg/controller/populators/import-populator.go
```diff
@@ -181,21 +187,35 @@ func (r *ImportPopulatorReconciler) updatePVCForPopulation(pvc *corev1.Persisten
...
+	}
+	if vddkVersion := pvcPrime.Annotations[cc.AnnVddkVersion]; vddkVersion != "" {
+		cc.AddAnnotation(pvc, cc.AnnVddkVersion, vddkVersion)
+	}
+}
+
+func (r *ImportPopulatorReconciler) updateImportAnnotations(pvc, pvcPrime *corev1.PersistentVolumeClaim) {
+	phase := pvcPrime.Annotations[cc.AnnPodPhase]
+	if err := r.updateImportProgress(phase, pvc, pvcPrime); err != nil {
+		r.log.Error(err, "Failed to update import progress for pvc %s/%s", pvc.Namespace, pvc.Name)
```
arnongilboa, line 205: why just log the error?

pkg/controller/datavolume/upload-controller.go
ShellyKa13: Ill take the Get suggestion, I like better to have the not found err in another line :)

pkg/controller/datavolume/util.go
ShellyKa13: I just moved the function but sure ill fix :)

pkg/controller/populators/import-populator.go
ShellyKa13: I dont want to prevent the phase update in case there is some issue with the progress pod

pkg/controller/datavolume/import-controller.go
```diff
@@ -434,3 +502,82 @@ func (r *ImportReconciler) getNextCheckpoint(dataVolume *cdiv1.DataVolume, pvc *
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *ImportReconciler) deleteVolumeImportSourceCR(syncState *dvSyncState) error {
+	importSourceName := volumeImportSourceName(syncState.dvMutated)
+	importSource := &cdiv1.VolumeImportSource{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: importSourceName, Namespace: syncState.dvMutated.Namespace}, importSource); err != nil {
```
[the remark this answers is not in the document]
ShellyKa13: and thats the reason why I did the get in the first place which of course I forgot so I removed it :/ I guess Ill bring it back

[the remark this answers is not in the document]
arnongilboa: Good reason, but are we really going to get multiple reconciles when `Phase == cdiv1.Succeeded`? if not then we'll have a single Delete anyway.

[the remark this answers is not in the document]
mhenriks: With garbage collection disabled, the vast majority of DV reconciles will be in Succeeded state because of cache resync.

Not a huge difference either way, but if we have a cache why not use it

alromeros (commented):
Really good job, don't have any comments other than the typo in the test. You'll probably push again anyway but I'm giving a provisional lgtm just in case 
/lgtm

pkg/controller/populators/import-populator_test.go
```diff
@@ -302,6 +307,57 @@ var _ = Describe("Import populator tests", func() {
...
+			Expect(updatedPVC.GetAnnotations()).ToNot(BeNil())
+			for _, ann := range desiredAnnotations {
+				_, ok := updatedPVC.Annotations[ann]
+				Expect(ok).To(BeTrue())
+			}
+			_, ok := updatedPVC.Annotations["undesiredAnn"]
+			Expect(ok).To(BeFalse())
+		},
+			table.Entry("with pod running phase", string(corev1.PodRunning)),
+			table.Entry("with pod succeded phase", string(corev1.PodFailed)),
```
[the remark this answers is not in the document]
ShellyKa13: done

mhenriks:
/approve
🥇

PR #2724, opened 2023-05-22
Add clone from snapshot functionalities to clone-populator

alromeros:

This Pull Request updates the clone populator to include cloning from snapshot capabilities.
For more information, check the clone populator PR (https://github.com/kubevirt/containerized-data-importer/pull/2709) and the clone from snapshot PR (https://github.com/kubevirt/containerized-data-importer/pull/2522).

**Special notes for your reviewer**: Work in progress.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/clone/planner.go
+++ b/pkg/controller/clone/planner.go
@@ -345,10 +345,11 @@ func (p *Planner) computeStrategyForSourceSnapshot(ctx context.Context, args *Ch
 	// Defaulting to host-assisted clone since it has fewer requirements
 	strategy := cdiv1.CloneStrategyHostAssisted
 
-	// Do volume snapshot validation before checking other specific stuff
 	if ok, err := p.validateTargetStorageClassAssignment(ctx, args); !ok || err != nil {
 		return nil, err
 	}
+
+	// Check that snapshot exists
 	sourceSnapshot := &snapshotv1.VolumeSnapshot{}
 	exists, err := getResource(ctx, p.Client, args.DataSource.Namespace, args.DataSource.Spec.Source.Name, sourceSnapshot)
 	if err != nil {
@@ -360,11 +361,6 @@ func (p *Planner) computeStrategyForSourceSnapshot(ctx context.Context, args *Ch
 		args.Log.V(3).Info("Source Snapshot does not exist, cannot compute strategy")
 		return nil, nil
 	}
-	valid, err := cc.IsSnapshotValidForClone(sourceSnapshot, args.Log)
-	if err != nil || !valid {
-		p.Recorder.Event(args.TargetClaim, corev1.EventTypeWarning, CloneValidationFailed, MessageCloneValidationFailed)
-		return nil, err
-	}
 
 	// Do snapshot and storage class validation
 	targetStorageClass, err := GetStorageClassForClaim(ctx, p.Client, args.TargetClaim)
@@ -374,7 +370,7 @@ func (p *Planner) computeStrategyForSourceSnapshot(ctx context.Context, args *Ch
 	if targetStorageClass == nil {
 		return nil, fmt.Errorf("target claim's storageclass doesn't exist, clone will not work")
 	}
-	valid, err = cc.ValidateSnapshotCloneProvisioners(ctx, p.Client, sourceSnapshot, targetStorageClass)
+	valid, err := cc.ValidateSnapshotCloneProvisioners(ctx, p.Client, sourceSnapshot, targetStorageClass)
 	if err != nil {
 		return nil, err
 	}
@@ -383,7 +379,7 @@ func (p *Planner) computeStrategyForSourceSnapshot(ctx context.Context, args *Ch
 		return &strategy, nil
 	}
 
-	// Lastly, do size validation to determine wether to use dumb or smart cloning
+	// Lastly, do size validation to determine whether to use dumb or smart cloning
 	valid, err = cc.ValidateSnapshotCloneSize(sourceSnapshot, &args.TargetClaim.Spec, targetStorageClass, args.Log)
 	if err != nil {
 		return nil, err
@@ -559,6 +555,10 @@ func (p *Planner) planHostAssistedFromSnapshot(ctx context.Context, args *PlanAr
 		Client:         p.Client,
 		Log:            args.Log,
 		Recorder:       p.Recorder,
+	}
+
+	if args.DataSource.Spec.PriorityClassName != nil {
+		hcp.PriorityClassName = *args.DataSource.Spec.PriorityClassName
 	}
 
 	rp := &RebindPhase{
@@ -745,16 +745,16 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
 }
 
 func createTempSourceClaim(ctx context.Context, namespace string, targetClaim *corev1.PersistentVolumeClaim, snapshot *snapshotv1.VolumeSnapshot, client client.Client) (*corev1.PersistentVolumeClaim, error) {
-	scName, err := getStorageClassNameForTempSourceClaim(ctx, snapshot, *targetClaim.Spec.StorageClassName, client)
+	scName, err := getStorageClassNameForTempSourceClaim(ctx, snapshot, client)
 	if err != nil {
 		return nil, err
 	}
 	// Get the appropriate size from the snapshot
 	targetCpy := targetClaim.DeepCopy()
-	restoreSize := snapshot.Status.RestoreSize
-	if restoreSize == nil || restoreSize.Sign() == -1 {
+	if snapshot.Status == nil || snapshot.Status.RestoreSize == nil || snapshot.Status.RestoreSize.Sign() == -1 {
 		return nil, fmt.Errorf("snapshot has no RestoreSize")
 	}
+	restoreSize := snapshot.Status.RestoreSize
 	if restoreSize.IsZero() {
 		reqSize := targetCpy.Spec.Resources.Requests[corev1.ResourceStorage]
 		restoreSize = &reqSize
@@ -777,8 +777,8 @@ func createTempSourceClaim(ctx context.Context, namespace string, targetClaim *c
 	return desiredClaim, nil
 }
 
-func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapshotv1.VolumeSnapshot, targetSCName string, client client.Client) (string, error) {
-	matches := []string{}
+func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapshotv1.VolumeSnapshot, client client.Client) (string, error) {
+	var matches []string
 
 	if snapshot.Status == nil || snapshot.Status.BoundVolumeSnapshotContentName == nil {
 		return "", fmt.Errorf("volumeSnapshotContent name not found")
@@ -787,15 +787,13 @@ func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapsh
 	if err := client.Get(ctx, types.NamespacedName{Name: *snapshot.Status.BoundVolumeSnapshotContentName}, volumeSnapshotContent); err != nil {
 		return "", err
 	}
-	// Attempting to get a storageClass compatible with the source snapshot,
-	// but different from the one used in the target claim.
-	// This allows us to be consistent with the host-assisted clone behavior.
+	// Attempting to get a storageClass compatible with the source snapshot
 	storageClasses := &storagev1.StorageClassList{}
 	if err := client.List(ctx, storageClasses); err != nil {
 		return "", err
 	}
 	for _, storageClass := range storageClasses.Items {
-		if storageClass.Provisioner == volumeSnapshotContent.Spec.Driver && storageClass.Name != targetSCName {
+		if storageClass.Provisioner == volumeSnapshotContent.Spec.Driver {
 			matches = append(matches, storageClass.Name)
 		}
 	}
--- a/pkg/controller/clone/planner.go
+++ b/pkg/controller/clone/planner.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 	"sync"
 
 	"github.com/go-logr/logr"
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -130,7 +132,10 @@ func (p *Planner) ChooseStrategy(ctx context.Context, args *ChooseStrategyArgs) 
 		args.Log.V(3).Info("Getting strategy for PVC source")
 		return p.computeStrategyForSourcePVC(ctx, args)
 	}
-
+	if IsDataSourceSnapshot(args.DataSource.Spec.Source.Kind) {
+		args.Log.V(3).Info("Getting strategy for Snapshot source")
+		return p.computeStrategyForSourceSnapshot(ctx, args)
+	}
 	return nil, fmt.Errorf("unsupported datasource")
 }
 
@@ -163,6 +168,18 @@ func (p *Planner) Plan(ctx context.Context, args *PlanArgs) ([]Phase, error) {
 			args.Log.V(3).Info("Planning csi clone from PVC")
 
 			return p.planCSIClone(ctx, args)
+		}
+	}
+
+	if IsDataSourceSnapshot(args.DataSource.Spec.Source.Kind) {
+		if args.Strategy == cdiv1.CloneStrategyHostAssisted {
+			args.Log.V(3).Info("Planning host assisted clone from Snapshot")
+
+			return p.planHostAssistedFromSnapshot(ctx, args)
+		} else if args.Strategy == cdiv1.CloneStrategySnapshot {
+			args.Log.V(3).Info("Planning Smart clone from Snapshot")
+
+			return p.planSmartCloneFromSnapshot(ctx, args)
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/controller/common/util.go
```diff
@@ -1619,3 +1620,40 @@ func ProgressFromClaim(ctx context.Context, args *ProgressFromClaimArgs) (string
...
+		if msg := snapshot.Status.Error.Message; msg != nil {
+			errMessage = *msg
+		}
+		return false, fmt.Errorf("snapshot in error state with msg: %s", errMessage)
+	}
+	return true, nil
+}
+
+// ValidateSnapshotCloneSize does proper size validation when doing a clone from snapshot operation
+func ValidateSnapshotCloneSize(snapshot *snapshotv1.VolumeSnapshot, pvcSpec *corev1.PersistentVolumeClaimSpec, targetSC *storagev1.StorageClass, log logr.Logger) (bool, error) {
```
mhenriks, line 1644: I guess you forsee these funcs being used outside of the clone package?

mhenriks: I see now 👍

pkg/controller/clone/planner.go
```diff
@@ -305,6 +320,62 @@ func (p *Planner) computeStrategyForSourcePVC(ctx context.Context, args *ChooseS
...
+		return nil, nil
+	}
+	valid, err := cc.IsSnapshotValidForClone(sourceSnapshot, args.Log)
+	if err != nil || !valid {
+		return nil, err
+	}
+
+	// Do snapshot and storage class validation
+	volumeSnapshotClass := &snapshotv1.VolumeSnapshotClass{}
+	exists, err = getResource(ctx, p.Client, args.DataSource.Namespace, *sourceSnapshot.Spec.VolumeSnapshotClassName, volumeSnapshotClass)
```
mhenriks, line 348: VolumeSnapshotClasses can come and go (like StorageClasses). It may be better to get the driver from the VolumeSnapshotContent. Then match that driver to the target PVC via StorageClass

pkg/controller/clone/planner.go
```diff
@@ -410,6 +481,131 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+		SourceName:     args.DataSource.Spec.Source.Name,
+		DesiredClaim:   sourceClaimForDumbClone,
+		OwnershipLabel: p.OwnershipLabel,
+		Client:         p.Client,
+		Log:            args.Log,
+		Recorder:       p.Recorder,
+	}
+
+	desiredClaim := createDesiredClaim(args.DataSource.Namespace, args.TargetClaim)
+	hcp := &HostClonePhase{
```
mhenriks, line 517: I think we have to do the filesystem overhead adjustment on this like in `planHostAssistedFromPVC`

pkg/controller/clone/planner.go
```diff
@@ -410,6 +481,131 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+		SourceName:     sourceClaimForDumbClone.Name,
+		DesiredClaim:   desiredClaim,
+		ImmediateBind:  true,
+		OwnershipLabel: p.OwnershipLabel,
+		Client:         p.Client,
+		Log:            args.Log,
+		Recorder:       p.Recorder,
+	}
+
+	pcp := &PrepClaimPhase{
```
mhenriks, line 529: shouldn't have to prep this, with host clone, don't have to expand or force bind

mhenriks (commented):
Great start!

pkg/controller/clone/planner.go
```diff
@@ -410,6 +481,131 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+	cfsp := &SnapshotClonePhase{
+		Owner:          args.TargetClaim,
+		Namespace:      args.DataSource.Namespace,
+		SourceName:     args.DataSource.Spec.Source.Name,
+		DesiredClaim:   sourceClaimForDumbClone,
+		OwnershipLabel: p.OwnershipLabel,
+		Client:         p.Client,
+		Log:            args.Log,
+		Recorder:       p.Recorder,
+	}
```
mhenriks, line 514: I think you may have to prep `sourceClaimForDumbClone` to force bind if source storageclass is WFFC

pkg/controller/clone/planner.go
```diff
@@ -410,6 +481,131 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+	vsc, err := GetCompatibleVolumeSnapshotClass(ctx, p.Client, args.TargetClaim)
+	if err != nil {
+		return nil, err
+	}
+	if vsc == nil {
+		return nil, fmt.Errorf("no compatible volumesnapshotclass")
+	}
+
+	// TODO: Maybe improve the logic to create the sourceClaim for host assisted clone, or even just make it more pretty
+	sourceClaimForDumbClone := createDesiredClaim(args.DataSource.Namespace, args.TargetClaim)
```
mhenriks, line 503: Pretty sure you are going to have to set a different storage class on this thing...one that is compatible with the source.  Otherwise we wouldn't be executing this strategy

pkg/controller/clone/planner.go
```diff
@@ -532,3 +736,50 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
...
+	if err := client.List(ctx, storageClasses); err != nil {
+		return nil, err
+	}
+	for _, storageClass := range storageClasses.Items {
+		if storageClass.Provisioner == volumeSnapshotContent.Spec.Driver && storageClass.Name != scName {
+			matches = append(matches, storageClass)
+		}
+	}
+	if len(matches) >= 1 {
+		scName = matches[0].Name
```
mhenriks, line 768: Maybe sort matches before returning the first one in order to be deterministic.  It's also done in the VolumeSnapshotClass lookup code

pkg/controller/clone/planner.go
```diff
@@ -305,6 +321,58 @@ func (p *Planner) computeStrategyForSourcePVC(ctx context.Context, args *ChooseS
...
+		return nil, nil
+	}
+	valid, err := cc.IsSnapshotValidForClone(sourceSnapshot, args.Log)
+	if err != nil || !valid {
+		p.Recorder.Event(args.TargetClaim, corev1.EventTypeWarning, CloneValidationFailed, MessageCloneValidationFailed)
+		return nil, err
+	}
+
+	// Do snapshot and storage class validation
+	targetStorageClass, err := cc.GetStorageClassByName(ctx, p.Client, args.TargetClaim.Spec.StorageClassName)
```
mhenriks, line 351: I don't love this function in populator context because it has some business logic we don't care about (returns default volumesnapshotclass if name is nil).  A simple get makes more sense to me but your call

pkg/controller/common/util.go
```diff
@@ -1619,3 +1620,66 @@ func ProgressFromClaim(ctx context.Context, args *ProgressFromClaimArgs) (string
...
+	return true, nil
+}
+
+// ValidateSnapshotCloneProvisioners validates the target PVC storage class against the snapshot class provisioner
+func ValidateSnapshotCloneProvisioners(ctx context.Context, c client.Client, snapshot *snapshotv1.VolumeSnapshot, storageClass *storagev1.StorageClass) (bool, error) {
+	// Do snapshot and storage class validation
+	if storageClass == nil {
+		return false, fmt.Errorf("target storage class not found")
+	}
+	volumeSnapshotContentName := snapshot.Spec.Source.VolumeSnapshotContentName
```
mhenriks, line 1670: `status.boundVolumeSnapshotContentName` is the right place.  The spec field is for manually creating Volumesnapshots.

I hit this when I tried testing this pr manually

mhenriks (commented):
needs some testing

pkg/controller/common/util.go
alromeros: Thanks! I noticed something wasn't right but couldn't find the fix.

pkg/controller/clone/planner.go
```diff
@@ -431,6 +502,141 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+	}
+	if !exists {
+		return nil, fmt.Errorf("source claim does not exist")
+	}
+
+	vsc, err := GetCompatibleVolumeSnapshotClass(ctx, p.Client, args.TargetClaim)
+	if err != nil {
+		return nil, err
+	}
+	if vsc == nil {
```
mhenriks, line 519: vsc is not used past here.  Don't think you need to get it

pkg/controller/clone/planner.go
```diff
@@ -553,3 +759,46 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
...
+	}
+	for _, storageClass := range storageClasses.Items {
+		if storageClass.Provisioner == volumeSnapshotContent.Spec.Driver && storageClass.Name != scName {
+			matches = append(matches, storageClass.Name)
+		}
+	}
+	if len(matches) > 0 {
+		sort.Strings(matches)
+		scName = matches[0]
+	}
```
mhenriks, line 788: why not break out getting scName into a separate function?

pkg/controller/clone/planner.go
```diff
@@ -431,6 +502,141 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+	if !exists {
+		return nil, fmt.Errorf("source claim does not exist")
+	}
+
+	vsc, err := GetCompatibleVolumeSnapshotClass(ctx, p.Client, args.TargetClaim)
+	if err != nil {
+		return nil, err
+	}
+	if vsc == nil {
```
mhenriks, line 599: vsc not used here either

pkg/controller/clone/planner.go
```diff
@@ -431,6 +502,141 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+		return nil, fmt.Errorf("no compatible volumesnapshotclass")
+	}
+
+	sourceClaimForDumbClone, err := createTempSourceClaim(ctx, args.DataSource.Namespace, args.TargetClaim, sourceSnapshot, p.Client)
```
mhenriks, line 523: This claim should be the same size as the `restoreSize` of the snapshot.  We don't want prep phase to expand it to the target size.

pkg/controller/clone/planner.go
mhenriks: also, should return an error if scName cannot be found.  it is dereferenced later

mhenriks (commented):
couple issues, func tests also don't compile

pkg/controller/clone/planner.go
alromeros: My idea was to use the target claim storage class if no match was found. The clone still seemed to work in that case.

pkg/controller/clone/planner.go
alromeros: We still want to prep `sourceClaimForDumbClone`, right? Is it enough to use the restoreSize when creating the claim?

pkg/controller/clone/planner.go
mhenriks: No way that can be guaranteed to work.  If it does then the "snapshot" clone method should have been chosen by the planner.  Sounds like you explicitly specified the "copy" clone type when creating the target PVC?

pkg/controller/clone/planner.go
mhenriks: yes prep is for force bind only if required

pkg/controller/common/util.go
```diff
@@ -1619,3 +1620,65 @@ func ProgressFromClaim(ctx context.Context, args *ProgressFromClaimArgs) (string
 
 	return "", nil
 }
+
+// IsSnapshotValidForClone returns true if the passed snapshot is valid for cloning
+func IsSnapshotValidForClone(snapshot *snapshotv1.VolumeSnapshot, log logr.Logger) (bool, error) {
```
mhenriks, line 1625: We actually don't care about most of this for planning purposes.  The phase will wait for the snapshot to be ready to use.  I think we only care that the snapshot content exists so we can get the driver.

I don't think the planner has to call this function at all since `ValidateSnapshotCloneProvisioners` validates the driver

pkg/controller/clone/planner.go
```diff
@@ -553,3 +743,65 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
...
+	}
+	// Attempting to get a storageClass compatible with the source snapshot,
+	// but different from the one used in the target claim.
+	// This allows us to be consistent with the host-assisted clone behavior.
+	storageClasses := &storagev1.StorageClassList{}
+	if err := client.List(ctx, storageClasses); err != nil {
+		return "", err
+	}
+	for _, storageClass := range storageClasses.Items {
+		if storageClass.Provisioner == volumeSnapshotContent.Spec.Driver && storageClass.Name != targetSCName {
```
mhenriks, line 798: why `storageClass.Name != targetSCName`?  That is totally fine

pkg/controller/clone/planner.go
mhenriks: The target storage class should not be considered when choosing.  It is possible that the chosen sc is the same as the target but that is a coincidence

pkg/controller/clone/planner.go
```diff
@@ -553,3 +743,65 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
...
 }
+
+func createTempSourceClaim(ctx context.Context, namespace string, targetClaim *corev1.PersistentVolumeClaim, snapshot *snapshotv1.VolumeSnapshot, client client.Client) (*corev1.PersistentVolumeClaim, error) {
+	scName, err := getStorageClassNameForTempSourceClaim(ctx, snapshot, *targetClaim.Spec.StorageClassName, client)
+	if err != nil {
+		return nil, err
+	}
+	// Get the appropriate size from the snapshot
+	targetCpy := targetClaim.DeepCopy()
+	restoreSize := snapshot.Status.RestoreSize
```
mhenriks, line 754: `snapshot.Status` may be nil

mhenriks (commented):
Getting there

Can you please put further updates in new commits?  Helpful when reviewing and everything gets squashed when merging to main

pkg/controller/clone/planner.go
alromeros: I can check it here again but in `IsSnapshotValidForClone` we already do some validations to ensure that the required fields are always populated.

pkg/controller/common/util.go
alromeros: But this also ensures that the snapshot has a populated status (required for the snapshot content) and it's not errored. I think it doesn't hurt to have it here.

mhenriks: At each step in the clone process, we should only do what is absolutely necessary.  There is no benefit to this.  It actually makes things worse because the controller has to do a `requeueAfter` when the planner cannot compute the strategy

pkg/controller/clone/planner_test.go
```diff
@@ -818,22 +816,6 @@ var _ = Describe("Planner test", func() {
 			validateHostClonePhase(planner, args, plan[2])
 			validateRebindPhase(planner, args, plan[3])
 		})
-
-		It("should fail planning host-assisted clone from snapshot when no valid storage class for source PVC is found", func() {
```
mhenriks: This is a case that should be tested.  It is totally possible that a snapshot is created, the storage class is deleted, and we cannot create the temp claim

pkg/controller/clone/planner.go
```diff
@@ -553,3 +738,63 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
...
+	desiredClaim.Spec.DataSource = nil
+	desiredClaim.Spec.DataSourceRef = nil
+	desiredClaim.Spec.StorageClassName = &scName
+	desiredClaim.Spec.Resources.Requests[corev1.ResourceStorage] = *restoreSize
+
+	return desiredClaim, nil
+}
+
+func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapshotv1.VolumeSnapshot, client client.Client) (string, error) {
+	matches := []string{}
```
mhenriks, line 776: nit: you don't need to initialize a zero length slice in order to work with it.  `var matches []string` would be the canonical definition in my experience.  But there is nothing wrong with this.

pkg/controller/clone/planner.go
```diff
@@ -745,16 +740,16 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
...
+	scName, err := getStorageClassNameForTempSourceClaim(ctx, snapshot, client)
 	if err != nil {
 		return nil, err
 	}
 	// Get the appropriate size from the snapshot
 	targetCpy := targetClaim.DeepCopy()
-	restoreSize := snapshot.Status.RestoreSize
-	if restoreSize == nil || restoreSize.Sign() == -1 {
+	if snapshot.Status == nil || snapshot.Status.RestoreSize == nil || snapshot.Status.RestoreSize.Sign() == -1 {
 		return nil, fmt.Errorf("snapshot has no RestoreSize")
```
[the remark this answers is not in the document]
mhenriks: I guess, if we hit this condition, this function could return a single "dummy" phase that returns `reconcile.Result{RequeueAfter: xxx}, nil`.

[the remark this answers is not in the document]
alromeros: If status is nil we would return an error in `computeStrategyForSourceSnapshot` before reaching this point, don't know if that's better. I'd be ok with adding again the checks I removed.

awels:
linter seems unhappy `gofmt -w <file>` should make it happy

pkg/controller/clone/planner.go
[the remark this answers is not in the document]
mhenriks: I'd prefer to go with my suggestion above but we don't have to do it now.  Having resoreSize is not necessary to compute strategy so let's not deal with it there

tests/clone-populator_test.go
```diff
@@ -309,6 +309,9 @@ var _ = Describe("Clone Populator tests", func() {
 			if !f.IsSnapshotStorageClassAvailable() {
 				Skip("Clone from volumesnapshot does not work without snapshot capable storage")
 			}
+			if utils.DefaultStorageClassCsiDriver == nil {
```
ShellyKa13, line 312: instead of putting it here and in the beforeEach of the clone from pvc it can be in one BeforeEach at the beginning of all the clone tests

ShellyKa13 (commented):
couple minor comments from a quick look. will have to continue tomorrow

pkg/controller/clone/planner.go
```diff
@@ -324,6 +341,55 @@ func (p *Planner) computeStrategyForSourcePVC(ctx context.Context, args *ChooseS
...
+	valid, err := cc.ValidateSnapshotCloneProvisioners(ctx, p.Client, sourceSnapshot, targetStorageClass)
+	if err != nil {
+		return nil, err
+	}
+	if !valid {
+		args.Log.V(3).Info("Provisioner differs, need to fall back to host assisted")
+		return &strategy, nil
+	}
+
+	// Lastly, do size validation to determine wether to use dumb or smart cloning
```
ShellyKa13, line 381: typo-> whether

pkg/controller/clone/planner.go
```diff
@@ -431,6 +497,125 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+	sourceSnapshot := &snapshotv1.VolumeSnapshot{}
+	exists, err := getResource(ctx, p.Client, args.DataSource.Namespace, args.DataSource.Spec.Source.Name, sourceSnapshot)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, fmt.Errorf("source claim does not exist")
+	}
+
+	sourceClaimForDumbClone, err := createTempSourceClaim(ctx, args.DataSource.Namespace, args.TargetClaim, sourceSnapshot, p.Client)
```
ShellyKa13, line 510: I wonder if this justifies another phase, if we have snapshotphase, why createpvcfromsnapshot doesnt justify a phase

pkg/controller/clone/planner.go
```diff
@@ -431,6 +497,125 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+		OwnershipLabel:  p.OwnershipLabel,
+		Client:          p.Client,
+		Log:             args.Log,
+		Recorder:        p.Recorder,
+	}
+
+	desiredClaim := createDesiredClaim(args.DataSource.Namespace, args.TargetClaim)
+	if util.ResolveVolumeMode(desiredClaim.Spec.VolumeMode) == corev1.PersistentVolumeFilesystem {
+		ds := desiredClaim.Spec.Resources.Requests[corev1.ResourceStorage]
+		is, err := cc.InflateSizeWithOverhead(ctx, p.Client, ds.Value(), &args.TargetClaim.Spec)
```
ShellyKa13, line 540: @mhenriks I hope you are prepared to handle the inflation when integrating dvs with clone populators

pkg/controller/clone/planner.go
```diff
@@ -431,6 +497,125 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
...
+		Owner:          args.TargetClaim,
+		Namespace:      sourceClaimForDumbClone.Namespace,
+		SourceName:     sourceClaimForDumbClone.Name,
+		DesiredClaim:   desiredClaim,
+		ImmediateBind:  true,
+		OwnershipLabel: p.OwnershipLabel,
+		Client:         p.Client,
+		Log:            args.Log,
+		Recorder:       p.Recorder,
+	}
```
ShellyKa13, line 557: here we dont have PriorityClassName? like done in clone from pvc with hostassited?:
	if args.DataSource.Spec.PriorityClassName != nil {
		hcp.PriorityClassName = *args.DataSource.Spec.PriorityClassName
	}

ShellyKa13 (commented):
Looks good, had some questions and suggestions

pkg/controller/clone/planner.go
mhenriks: Can't wait

pkg/controller/clone/planner.go
alromeros: Will add it, thanks!

PR #2738, opened 2023-06-02
FormatReader should support ISO file

### LINKED ISSUE #2737: HTTP Datasource can't handle s3 presigned URL for ISO file
**What happened**:
Create a DataVolume with HTTP Datasource. The HTTP URL is a s3 presigned URL, which is only signed with HTTP Method 'GET' (which means **any HTTP Method other than 'GET' is forbidden by s3 policy**)

I got error log from cdi-importer, looks like nbdkit try to do HTTP 'HEAD' on the s3 presigned URL, which is forbidden.

```
E0601 12:44:17.844208       1 importer.go:177] qemu-img: Could not open 'nbd+unix:///?socket=/tmp/nbdkit.sock': Requested export not available
, qemu-img execution failed: exit status 1 Log line from nbdkit: nbdkit: curl[1]: error: problem doing HEAD request to fetch size of URL [http://xxx/]: HTTP response code said error: The requested URL returned error: 403
Unable to convert source data to target format

kubevirt.io/containerized-data-importer/pkg/importer.(*DataProcessor).initDefaultPhases.func6
	pkg/importer/data-processor.go:242
kubevirt.io/containerized-data-importer/pkg/importer.(*DataProcessor).ProcessDataWithPause
	pkg/importer/data-processor.go:275
kubevirt.io/containerized-data-importer/pkg/importer.(*DataProcessor).ProcessData
	pkg/importer/data-processor.go:184
main.handleImport
	cmd/cdi-importer/importer.go:174
main.main
	cmd/cdi-importer/importer.go:140
runtime.main
	GOROOT/src/runtime/proc.go:250
runtime.goexit
	GOROOT/src/runtime/asm_amd64.s:1571
```

**Additional context**:
I can't use s3 datasource, because I think it's not safe to use s3 AK/SK just to download something.

**Environment**:
- CDI version: 1.56.0
- Others: nbdkit-1.30.1-2.el9

lxs137:

HTTP Datasource can't handle s3 presigned URL for ISO file

Fixes #2737

**Release note**:

```release-note
FormatReader support ISO fileheader
```

### DIFF
--- a/pkg/image/filefmt.go
+++ b/pkg/image/filefmt.go
@@ -10,10 +10,10 @@ import (
 )
 
 // MaxExpectedHdrSize defines the Size of buffer used to read file headers.
-// Note: this is the size of tar's header. If a larger number is used the tar unarchive operation
+// Note: this is the size of iso's header. If a larger number is used the tar unarchive operation
 //
 //	creates the destination file too large, by the difference between this const and 512.
-const MaxExpectedHdrSize = 512
+const MaxExpectedHdrSize = 32774
 
 // Headers provides a map for header info, key is file format, eg. "gz" or "tar", value is metadata describing the layout for this hdr
 type Headers map[string]Header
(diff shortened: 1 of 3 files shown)


mhenriks (commented):
@lxs137 I'm not sure about this fix.  Via #2737, it appears that the real issue is that nbdkit has a problem when serving up s3 signed urls.  This pr gets around that by adding explicit iso detection and saying that iso requires conversion, which they really don't.  But by saying conversion is required, we will use the go http client to download the file to scratch space.

I think it would be better to address the issue with nbdkit rather than always download raw/iso files to scratch space.  Nbdkit handles many other servers/files/protocols just fine. I don't think we should penalize all raw/iso imports by requiring scratch space just to support S3.

@rwmjones @awels

lxs137:
#2743 fix is correct

PR #2744, opened 2023-06-07
Fix test failure - missing pod following recreate of CDI

akalenyu:
From some debugging, it looks like the priority class tests don't properly wait for
redeploy to finish and thus the following tests get an initial wrong slice of cdiPods
(Consider 2 cdi-uploadproxy pods, one of which is terminating, making way to the new one)

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -610,6 +610,7 @@ var _ = Describe("ALL Operator tests", func() {
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "cdi-operator",
 						Namespace: f.CdiInstallNs,
+						Labels:    currentCdiOperatorDeployment.Labels,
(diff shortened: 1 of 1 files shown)


awels:
Okay, I guess we need to make sure those are cleaned up before moving to the next test then

akalenyu:
> Okay, I guess we need to make sure those are cleaned up before moving to the next test then

Yeah just have to find a nice way that relies on the deployment status

akalenyu:
> > Okay, I guess we need to make sure those are cleaned up before moving to the next test then
> 
> Yeah just have to find a nice way that relies on the deployment status

Should be ok now

ShellyKa13 (commented):
/lgtm

PR #2746, opened 2023-06-08
Refactor distinction between dataimportcron formats

akalenyu:
Following up on one of the review comments from the initial PR,
an interface should look better than switch cases.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -334,6 +334,10 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	if err != nil {
 		return res, err
 	}
+	handler, ok := formatHandlers[format]
+	if !ok {
+		return res, fmt.Errorf("can't find handler for format %s", format)
+	}
 
 	handlePopulatedPvc := func() error {
 		if pvc != nil {
@@ -342,7 +346,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			}
 		}
 		importSucceeded = true
-		if err := r.handleCronFormat(ctx, dataImportCron, format, dvStorageClass); err != nil {
+		if err := handler.handleSourceCreation(ctx, r, dataImportCron, &dataImportCron.Spec.Template, dvStorageClass); err != nil {
 			return err
 		}
 
@@ -390,7 +394,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		}
 	}
 
-	if err := r.updateDataSource(ctx, dataImportCron, format); err != nil {
+	if err := r.updateDataSource(ctx, dataImportCron, handler); err != nil {
 		return res, err
 	}
 
@@ -418,9 +422,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			}
 		}
 	} else if importSucceeded {
-		if err := r.updateDataImportCronSuccessCondition(ctx, dataImportCron, format, snapshot); err != nil {
-			return res, err
-		}
+		handler.updateOnSuccess(dataImportCron, snapshot)
 	} else if len(imports) > 0 {
 		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionFalse, "Import is progressing", inProgress)
 	} else {
@@ -545,7 +547,7 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
 	return nil
 }
 
-func (r *DataImportCronReconciler) updateDataSource(ctx context.Context, dataImportCron *cdiv1.DataImportCron, format cdiv1.DataImportCronSourceFormat) error {
+func (r *DataImportCronReconciler) updateDataSource(ctx context.Context, dataImportCron *cdiv1.DataImportCron, handler formatHandler) error {
 	log := r.log.WithName("updateDataSource")
 	dataSourceName := dataImportCron.Spec.ManagedDataSource
 	dataSource := &cdiv1.DataSource{}
@@ -575,7 +577,7 @@ func (r *DataImportCronReconciler) updateDataSource(ctx context.Context, dataImp
 	passCronLabelToDataSource(dataImportCron, dataSource, cc.LabelDynamicCredentialSupport)
 
 	sourcePVC := dataImportCron.Status.LastImportedPVC
-	populateDataSource(format, dataSource, sourcePVC)
+	populateDataSource(handler, dataSource, sourcePVC)
 
 	if !reflect.DeepEqual(dataSource, dataSourceCopy) {
 		if err := r.client.Update(ctx, dataSource); err != nil {
@@ -586,24 +588,12 @@ func (r *DataImportCronReconciler) updateDataSource(ctx context.Context, dataImp
 	return nil
 }
 
-func populateDataSource(format cdiv1.DataImportCronSourceFormat, dataSource *cdiv1.DataSource, sourcePVC *cdiv1.DataVolumeSourcePVC) {
+func populateDataSource(handler formatHandler, dataSource *cdiv1.DataSource, sourcePVC *cdiv1.DataVolumeSourcePVC) {
 	if sourcePVC == nil {
 		return
 	}
 
-	switch format {
-	case cdiv1.DataImportCronSourceFormatPvc:
-		dataSource.Spec.Source = cdiv1.DataSourceSource{
-			PVC: sourcePVC,
-		}
-	case cdiv1.DataImportCronSourceFormatSnapshot:
-		dataSource.Spec.Source = cdiv1.DataSourceSource{
-			Snapshot: &cdiv1.DataVolumeSourceSnapshot{
-				Namespace: sourcePVC.Namespace,
-				Name:      sourcePVC.Name,
-			},
-		}
-	}
+	handler.populateDataSourceSpec(&dataSource.Spec, sourcePVC)
 }
 
 func updateDataImportCronOnSuccess(dataImportCron *cdiv1.DataImportCron) error {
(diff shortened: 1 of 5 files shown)


pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (snapshotFormatHandler) updateOnSuccess(r *DataImportCronReconciler, ctx context.Context, dataImportCron *cdiv1.DataImportCron, pvcName string) error {
+	snapshot := &snapshotv1.VolumeSnapshot{}
+	err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: pvcName}, snapshot)
```
akalenyu, line 116: So I brought back the extra GET call since stuff get pretty messy if I try to pass along the snapshot/pvc depending on handler, any ideas to make this go away again are welcome :pray:

akalenyu: /cc @awels @arnongilboa

pkg/controller/dataimportcron-controller.go
```diff
@@ -416,7 +420,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			}
 		}
 	} else if importSucceeded {
-		if err := r.updateDataImportCronSuccessCondition(ctx, dataImportCron, format, snapshot); err != nil {
+		if err := handler.updateOnSuccess(r, ctx, dataImportCron, imports[0].DataVolumeName); err != nil {
```
arnongilboa, line 423: imports is `dataImportCron.Status.CurrentImports`, so why pass it as arg?

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,150 @@
...
+	} else {
+		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionFalse, "Snapshot of imported data is progressing", inProgress)
+	}
+
+	return nil
+}
+
+type pvcFormatHandler struct{}
+
+func (pvcFormatHandler) populateDataSourceSpec(dataSource *cdiv1.DataSource, sourcePVC *cdiv1.DataVolumeSourcePVC) {
```
arnongilboa, line 135: pass `dataSourceSpec *cdiv1.DataSourceSpec` instead

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,150 @@
...
+func (pvcFormatHandler) handleSourceCreation(_ *DataImportCronReconciler, _ context.Context, _ *cdiv1.DataImportCron, _ *cdiv1.DataVolume, _ *storagev1.StorageClass) error {
+	// No extra steps needed
+	return nil
+}
+
+func (pvcFormatHandler) updateOnSuccess(_ *DataImportCronReconciler, _ context.Context, dataImportCron *cdiv1.DataImportCron, _ string) error {
+	updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronUpToDate, corev1.ConditionTrue, "Latest import is up to date", upToDate)
+
+	return nil
+}
```
arnongilboa, line 150: afaik you don't need the _

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,150 @@
...
+var formatHandlers = map[cdiv1.DataImportCronSourceFormat]formatHandler{
+	cdiv1.DataImportCronSourceFormatSnapshot: snapshotFormatHandler{},
+	cdiv1.DataImportCronSourceFormatPvc:      pvcFormatHandler{},
+}
+
+type formatHandler interface {
+	populateDataSourceSpec(*cdiv1.DataSource, *cdiv1.DataVolumeSourcePVC)
+	handleSourceCreation(*DataImportCronReconciler, context.Context, *cdiv1.DataImportCron, *cdiv1.DataVolume, *storagev1.StorageClass) error
+	updateOnSuccess(*DataImportCronReconciler, context.Context, *cdiv1.DataImportCron, string) error
+}
```
arnongilboa, line 45: context.Context should be the first parameter of a function (linter says)

pkg/controller/dataimportcron-format-handlers.go
arnongilboa: Either pass the snapshot as `snapshot *snapshotv1.VolumeSnapshot`, or `sourceObj client.Object` if you think it's a cleaner interface. Drop the `pvcName`.

akalenyu: Yes I was going for client.Object but the casting gets really ugly, I will keep thinking about this one

arnongilboa: this one is so ugly? `snapshot := obj.(*snapshotv1.VolumeSnapshot)`

akalenyu: no, you also need the other way around (snap to obj)

arnongilboa: afaik you can pass it as is (`*VolumeSnapshot` as `client.Object`, not `*client.Object`)

akalenyu: Yes, that was my problem, thanks!
Interesting readouts:
https://stackoverflow.com/a/62146924
https://glucn.medium.com/golang-an-interface-holding-a-nil-value-is-not-nil-bb151f472cc7

pkg/controller/dataimportcron-controller.go
akalenyu: not doing that anymore

pkg/controller/dataimportcron-format-handlers.go
akalenyu: done

pkg/controller/dataimportcron-format-handlers.go
akalenyu: yep, done

akalenyu:
@arnongilboa ping

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,145 @@
...
+
+var formatHandlers = map[cdiv1.DataImportCronSourceFormat]formatHandler{
+	cdiv1.DataImportCronSourceFormatSnapshot: snapshotFormatHandler{},
+	cdiv1.DataImportCronSourceFormatPvc:      pvcFormatHandler{},
+}
+
+type formatHandler interface {
+	populateDataSourceSpec(*cdiv1.DataSourceSpec, *cdiv1.DataVolumeSourcePVC)
+	handleSourceCreation(context.Context, *DataImportCronReconciler, *cdiv1.DataImportCron, *cdiv1.DataVolume, *storagev1.StorageClass) error
+	updateOnSuccess(context.Context, *DataImportCronReconciler, *cdiv1.DataImportCron, client.Object) error
```
arnongilboa, line 43: `updateOnSuccess` returns nil on all flows, sure you want it to return error?

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := r.client.Get(context.TODO(), client.ObjectKeyFromObject(desiredSnapshot), currentSnapshot); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+		cc.AddAnnotation(desiredSnapshot, AnnLastUseTime, time.Now().UTC().Format(time.RFC3339Nano))
+		if err := r.client.Create(ctx, desiredSnapshot); err != nil {
+			return err
+		}
+	} else {
+		if cc.IsSnapshotReady(currentSnapshot) {
```
arnongilboa, line 101: maybe `else if`? one tab less :)

pkg/controller/dataimportcron-format-handlers.go
akalenyu: no reason to return error, thanks

pkg/controller/dataimportcron-format-handlers.go
akalenyu: even better, dropped the else

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,141 @@
...
+		r.log.Info("Deleting dv/pvc as snapshot is ready", "name", desiredSnapshot.Name)
+		if err := r.deleteDvPvc(ctx, desiredSnapshot.Name, desiredSnapshot.Namespace); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (snapshotFormatHandler) updateOnSuccess(ctx context.Context, r *DataImportCronReconciler, dataImportCron *cdiv1.DataImportCron, sourceObj client.Object) {
```
awels, line 113: Any reason we are passing the context and reconciler in and then not using it? I am guessing it is a leftover.

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,141 @@
...
+		PVC: sourcePVC,
+	}
+}
+
+func (pvcFormatHandler) handleSourceCreation(context.Context, *DataImportCronReconciler, *cdiv1.DataImportCron, *cdiv1.DataVolume, *storagev1.StorageClass) error {
+	// No extra steps needed
+	return nil
+}
+
+func (pvcFormatHandler) updateOnSuccess(_ context.Context, _ *DataImportCronReconciler, dataImportCron *cdiv1.DataImportCron, _ client.Object) {
```
awels, line 139: Same here, I think the interface should not be passing the context or the reconciler, and then the empty parameters are not needed.

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,141 @@
...
+			Source: snapshotv1.VolumeSnapshotSource{
+				PersistentVolumeClaimName: &dvName,
+			},
+			VolumeSnapshotClassName: &className,
+		},
+	}
+	r.setDataImportCronResourceLabels(dataImportCron, desiredSnapshot)
+
+	currentSnapshot := &snapshotv1.VolumeSnapshot{}
+	if err := r.client.Get(context.TODO(), client.ObjectKeyFromObject(desiredSnapshot), currentSnapshot); err != nil {
```
awels, line 92: Would we not want to use the context we passed in?

pkg/controller/dataimportcron-format-handlers.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+var formatHandlers = map[cdiv1.DataImportCronSourceFormat]formatHandler{
+	cdiv1.DataImportCronSourceFormatSnapshot: snapshotFormatHandler{},
+	cdiv1.DataImportCronSourceFormatPvc:      pvcFormatHandler{},
+}
+
+type formatHandler interface {
+	populateDataSourceSpec(*cdiv1.DataSourceSpec, *cdiv1.DataVolumeSourcePVC)
+	handleSourceCreation(context.Context, *DataImportCronReconciler, *cdiv1.DataImportCron, *cdiv1.DataVolume, *storagev1.StorageClass) error
```
awels, line 42: I am having a bit of a hard time with passing the reconciler into this function, I see we are using 4 things from the reconciler:
- 'Client'
- 'Log'
- 'setDataImportCronResourceLabels' (which really only cares about the `installerLabels`)
- 'deleteDvPvc' (which really only cares about the `Client`)
So we could relatively easily make deleteDvPvc not be part of the reconciler, then we would still have to pass at least 3 extra parameters to this function which also doesn't fill me with happiness.
Could we make some of these part of the implementing struct and not pass them as arguments to the function?

akalenyu: That would make the initialization of these struct handlers not static anymore
instead, they would be initialized on every reconcile trigger which may also make us unhappy
https://github.com/kubevirt/containerized-data-importer/blob/295b5ca61245bbe11064a0780026fab83395ec2d/pkg/controller/dataimportcron-format-handlers.go#L35-L38

pkg/controller/dataimportcron-format-handlers.go
akalenyu: done, thx! thats an issue

pkg/controller/dataimportcron-format-handlers.go
akalenyu: yep, thx

pkg/controller/dataimportcron-format-handlers.go
akalenyu: yep

PR #2748, opened 2023-06-12
cdi-importer: import virtualbox ova or tar archive in oneshot

lxs137:

This PR try to make CDI support import Virtualbox OVA file (tar archive is also supported) in oneshot. 
- with [nbdkit-tar-filter](https://libguestfs.org/nbdkit-tar-plugin.1.html), we can extract vmdk file in remote OVA endpoint, which make the conversion don't need scratch disk space.

Without the PR, we need to download OVA file, and extract vmdk file in it, then convert it to QCOW2 file, upload the QCOW2 file into the PVC. 

**something need help**
- if there are multiple vmdk in one ova file, should find a way to let user specify the target vmdk file.

Fixes #

**Release note**:

```release-note
cdi-importer: http-datasource can import (compressed) tar archive (or virtualbox OVA) streamly
```


awels:
Hi, thanks for the PR, looks like there are some unit tests failures. You can run the unit tests by calling `make test-unit`

hack/build/rpm-deps.sh
```diff
@@ -46,6 +46,7 @@ nbdkit-basic-filters
 nbdkit-curl-plugin
 nbdkit-xz-filter
 nbdkit-gzip-filter
+nbdkit-tar-filter
```
mhenriks, line 49: We need to understand if/how nbdkit does caching with this filter.  I bet some caching is going on as it would be pretty expensive to always open/seek to the appropriate offset in the archive.  If there is any significant on disk/memory caching I'd prefer we use simply the golang tar reader to seek to the file and download it to scratch.  Then convert from scratch to target.  Thus not getting nbdkit involved at all.

lxs137: Sorry, I can't get the point, why nbdkit cache will be the concern.
I believe nbdkit-tar-filter will only retrieve tar header block, if find the header name match "tar-entry", then get the target file data offset. (I think golang tar reader will do the same thing)

mhenriks: Looks to me like tar writes to tmp which is something we have to avoid:  https://github.com/libguestfs/nbdkit/blob/45b72f5bd8fc1b475fa130d06c86cd877bf595d5/filters/tar/tar.c#L146

But I'm not an expert on nbdkit.

@rwmjones care to comment on how curl + tar filter uses tmp space?

rwmjones: tar filter will not use tmp space (or more accurately, only a few bytes are used for a temporary file used to parse offsets printed by tar).  Here's how it works:

https://gitlab.com/nbdkit/nbdkit/-/blob/462eb6cbf605c2969f6e851bb9076f8e3961cccb/filters/tar/tar.c#L138

mhenriks:
I'd like to think about this PR more as "tar file support" then "ova support" so a couple suggestions.  If the tar archive only has one file, use that regardless of extension.  Support for other extensions like ".iso" ".raw" etc

What do you think @awels?

rwmjones:
It should be possible to tell cheaply if a tar file contains either just one file or (more realistic and still usable) the large file is contained early on.  Tar files are basically:
```
GLOBAL HEADER
FILE1 HEADER
FILE1 CONTENT
FILE2 HEADER
FILE2 CONTENT
&c
```
(There is no central index unlike zip or any modern format.)  You could read the first NN bytes (eg. 10M or whatever you consider small enough), feed that into `tar -t --block-number -f - filename` and see if it produces a result.

In fact this would be a simple modification to nbdkit-tar-file, adding some sort of "read limit". (@ebblake)

mhenriks:
FYI: single file in a tar came from a discussion with @rmohr regarding how GCE imports disk images as described here:  https://cloud.google.com/compute/docs/import/import-existing-image#requirements_for_the_image_file

rwmjones:
Patch for above posted: https://listman.redhat.com/archives/libguestfs/2023-June/031837.html

mhenriks:
> FYI: single file in a tar came from a discussion with @rmohr regarding how GCE imports disk images as described here: https://cloud.google.com/compute/docs/import/import-existing-image#requirements_for_the_image_file

Supporting "tar.gz" is probably beyond the scope of this PR though.

lxs137:
Update:
- support extract file extensions: "qcow2", "vmdk", "vdi", "vhd", "vhdx", "iso", "raw"
- if the tar archive only has one file, use that regardless of extension
- fix e2e functional tests

lxs137:
I am trying to use nbdkit serve a remote `tar.gz` archive with curl plugin, tar filter and gzip filter, but I find it will be super slow even execute qemu-img info over it.

fast:
```
nbdkit --foreground --readonly --exit-with-parent -U /tmp/nbdkit.sock --pidfile /tmp/nbdkit.pid --filter=tar -r curl url=http://xxx.tar tar-entry=xx.img &

// will cost near 0.1s
qemu-img info nbd+unix:///?socket=/tmp/nbdkit.sock
```

slow (60x):
```
nbdkit --foreground --readonly --exit-with-parent -U /tmp/nbdkit.sock --pidfile /tmp/nbdkit.pid --filter=tar --filter=gzip -r curl url=http://xxx.tar.gz tar-entry=xx.img &

// will cost near 6s
qemu-img info nbd+unix:///?socket=/tmp/nbdkit.sock
```

I find it's because when combined tar filter and gzip filter, nbdkit will try to download whole remote file. 
If I use tar filter only, nbdkit will only peek serval bytes in remote file.

@rwmjones will you take a look on it, thank you :-)

rwmjones:
`gzip` format isn't seekableᵃ.  Use `xz` format as described here: https://libguestfs.org/nbdkit-xz-filter.1.html#Getting-best-random-access-performance-from-xz

ᵃWell, it's a complicated story.  It *may be* possible to make a modifiable seekable gzip: https://rwmj.wordpress.com/2022/12/01/creating-a-modifiable-gzipped-disk-image/

lxs137:
> `gzip` format isn't seekableᵃ. Use `xz` format as described here: https://libguestfs.org/nbdkit-xz-filter.1.html#Getting-best-random-access-performance-from-xz
> 
> ᵃWell, it's a complicated story. It _may be_ possible to make a modifiable seekable gzip: https://rwmj.wordpress.com/2022/12/01/creating-a-modifiable-gzipped-disk-image/

OK, I test a `tar.xz` file with small block size, the qemu-img command run more fast (3x slow than run qemu-img over `tar` file).

If user provide an unseekable compressed tar archive (gzip, xz with large block size), run qemu-img  info command over nbdkit will be super slow.

@mhenriks What do you think?

mhenriks:
> > `gzip` format isn't seekableᵃ. Use `xz` format as described here: https://libguestfs.org/nbdkit-xz-filter.1.html#Getting-best-random-access-performance-from-xz
> > ᵃWell, it's a complicated story. It _may be_ possible to make a modifiable seekable gzip: https://rwmj.wordpress.com/2022/12/01/creating-a-modifiable-gzipped-disk-image/
> 
> OK, I test a `tar.xz` file with small block size, the qemu-img command run more fast (3x slow than run qemu-img over `tar` file).
> 
> If user provide an unseekable compressed tar archive (gzip, xz with large block size), run qemu-img info command over nbdkit will be super slow.
> 
> @mhenriks What do you think?

We prefer using scratch space with 'gz' or 'xz'.  So supporting 'tar.gz' should mean using the golang reader stack to download to scratch then qemu-img convert from there.

mhenriks:
Ho @lxs137 checking back in on this.  Are you planning to add `tar.gz` support or should we go with what's here for now?

lxs137:
> Ho @lxs137 checking back in on this. Are you planning to add `tar.gz` support or should we go with what's here for now?

Yes, I am still working on it, I think it will take a few days :-)

mhenriks:
> > Ho @lxs137 checking back in on this. Are you planning to add `tar.gz` support or should we go with what's here for now?
> 
> Yes, I am still working on it, I think it will take a few days :-)

Okay, take your time!

lxs137:
@mhenriks Sorry for my late reply, I have done all the work. :-)

- tar (inside file has known file extension): use nbdkit-tar-filter to extract inner file content
- compressed tar: FormatReaders will seek to inner file content
- tar (inside file has unknown file extension): can't handle

tools/cdi-func-test-file-host-init/main.go
```diff
@@ -52,6 +52,8 @@ func main() {
 		[]string{".qcow2", ".gz"},
 		[]string{".qcow2", ".xz"},
 		[]string{".qcow2", ".zst"},
+		[]string{".tar", ".gz"},
```
a180285, line 55: I wondering, if it can deal with file like xxx.raw.tar.gz ?

lxs137: @a180285  Sure, actually we match file by magic number instead of extension.

lxs137:
> @mhenriks Sorry for my late reply, I have done all the work. :-)
> 
> * tar (inside file has known file extension): use nbdkit-tar-filter to extract inner file content
> * compressed tar: FormatReaders will seek to inner file content
> * tar (inside file has unknown file extension): can't handle

@mhenriks Hi, do you have some time to review on this PR ~

awels:
Sorry this needs a rebase before we can review it.

lxs137:
I found we disable the nbdkit streamly importing in #2832 .
But in this PR, I try to extract the tar inner content with nbdkit filter (except compressed tar, use golang reader instead).
What do you think?
@awels @mhenriks

lxs137:
> I found we disable the nbdkit streamly importing in #2832 . But in this PR, I try to extract the tar inner content with nbdkit filter (except compressed tar, use golang reader instead). What do you think? @awels @mhenriks

Read some comments from #2809. 

Maybe this PR should hold unitl ndkit-1.35.8 is availavle in centos 9 stream, 
or split this PR into golang importing part and nbdkit importing part, we just focus on the golang importing part.

PR #2749, opened 2023-06-12
Delete old version DVs with DIC GC

ido106:

In the old versions, the garbage collection label was added to DV only. In the new version, the GC looks for this label in the PVC, so it does not find it and does not delete the DV's from the old versions

Fixes bz #2209969

**Release note**:

```release-note
Delete old version DV's with DIC garbage collector
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -804,7 +804,7 @@ func (r *DataImportCronReconciler) garbageCollectOldImports(ctx context.Context,
 		maxImports = int(*cron.Spec.ImportsToKeep)
 	}
 
-	if err := r.garbageCollectPVCs(ctx, cron.Namespace, selector, maxImports); err != nil {
+	if err := r.garbageCollectPVCs(ctx, cron.Namespace, cron.Name, selector, maxImports); err != nil {
 		return err
 	}
 	if err := r.garbageCollectSnapshots(ctx, cron.Namespace, selector, maxImports); err != nil {
@@ -814,7 +814,7 @@ func (r *DataImportCronReconciler) garbageCollectOldImports(ctx context.Context,
 	return nil
 }
 
-func (r *DataImportCronReconciler) garbageCollectPVCs(ctx context.Context, namespace string, selector labels.Selector, maxImports int) error {
+func (r *DataImportCronReconciler) garbageCollectPVCs(ctx context.Context, namespace, cronName string, selector labels.Selector, maxImports int) error {
 	pvcList := &corev1.PersistentVolumeClaimList{}
 
 	if err := r.client.List(ctx, pvcList, &client.ListOptions{Namespace: namespace, LabelSelector: selector}); err != nil {
@@ -828,6 +828,27 @@ func (r *DataImportCronReconciler) garbageCollectPVCs(ctx context.Context, names
 			r.log.Info("Deleting dv/pvc", "name", pvc.Name, "pvc.uid", pvc.UID)
 			if err := r.deleteDvPvc(ctx, pvc.Name, pvc.Namespace); err != nil {
 				return err
+			}
+		}
+	}
+
+	dvList := &cdiv1.DataVolumeList{}
+	if err := r.client.List(ctx, dvList, &client.ListOptions{Namespace: namespace, LabelSelector: selector}); err != nil {
+		return err
+	}
+
+	if len(dvList.Items) > maxImports {
+		for _, dv := range dvList.Items {
+			pvc := &corev1.PersistentVolumeClaim{}
+			if err := r.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: dv.Name}, pvc); err != nil {
+				return err
+			}
+
+			if pvc.Labels[common.DataImportCronLabel] != cronName {
+				r.log.Info("Deleting old version dv/pvc", "name", pvc.Name, "pvc.uid", pvc.UID)
+				if err := r.deleteDvPvc(ctx, dv.Name, dv.Namespace); err != nil {
+					return err
+				}
 			}
 		}
 	}
--- a/pkg/controller/dataimportcron-controller_test.go
+++ b/pkg/controller/dataimportcron-controller_test.go
@@ -706,6 +706,63 @@ var _ = Describe("All DataImportCron Tests", func() {
 			Entry("has no tag", imageStreamName, 1),
 		)
 
+		It("Should succeed garbage collecting old version DVs", func() {
+			cron = newDataImportCron(cronName)
+			importsToKeep := int32(1)
+			cron.Spec.ImportsToKeep = &importsToKeep
+			reconciler = createDataImportCronReconciler(cron)
+
+			// Labeled DV and unlabeled PVC
+			dv1 := cc.NewImportDataVolume("test-dv1")
+			dv1.Labels = map[string]string{common.DataImportCronLabel: cronName}
+			err := reconciler.client.Create(context.TODO(), dv1)
+			Expect(err).ToNot(HaveOccurred())
+
+			pvc1 := cc.CreatePvc(dv1.Name, dv1.Namespace, nil, nil)
+			err = reconciler.client.Create(context.TODO(), pvc1)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Labeled DV and PVC
+			dv2 := cc.NewImportDataVolume("test-dv2")
+			dv2.Labels = map[string]string{common.DataImportCronLabel: cronName}
+			err = reconciler.client.Create(context.TODO(), dv2)
+			Expect(err).ToNot(HaveOccurred())
+
+			pvc2 := cc.CreatePvc(dv2.Name, dv2.Namespace, nil, nil)
+			pvc2.Labels = map[string]string{common.DataImportCronLabel: cronName}
+			err = reconciler.client.Create(context.TODO(), pvc2)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Unlabeled DV and PVC
+			dv3 := cc.NewImportDataVolume("test-dv3")
+			err = reconciler.client.Create(context.TODO(), dv3)
+			Expect(err).ToNot(HaveOccurred())
+
+			pvc3 := cc.CreatePvc(dv3.Name, dv3.Namespace, nil, nil)
+			err = reconciler.client.Create(context.TODO(), pvc3)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = reconciler.garbageCollectOldImports(context.TODO(), cron)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Ensure the old version DV is deleted (labeled DV and unlabeled PVC).
+			// The labeled PVC will not be deleted here, as there is no relevant controller.
+			err = reconciler.client.Get(context.TODO(), dvKey(dv1.Name), dv1)
+			Expect(k8serrors.IsNotFound(err)).To(BeTrue())
+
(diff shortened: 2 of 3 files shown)


akalenyu:
If there is a bug opened for this, would be great to include it under `Fixes #`

ido106:
> If there is a bug opened for this, would be great to include it under `Fixes #`

There is only a Jira card
https://issues.redhat.com/browse/CNV-29106

ido106:
> > > If there is a bug opened for this, would be great to include it under `Fixes #`
> > 
> > 
> > There is only a Jira card https://issues.redhat.com/browse/CNV-29106
> 
> Yup, if you look at that issue it translates to a bugzilla

Ok, I added the Bugzilla bug number

pkg/controller/dataimportcron-controller.go
```diff
@@ -832,6 +832,24 @@ func (r *DataImportCronReconciler) garbageCollectPVCs(ctx context.Context, names
 		}
 	}
 
+	dvList := &cdiv1.DataVolumeList{}
+	if err := r.client.List(ctx, dvList, &client.ListOptions{Namespace: namespace, LabelSelector: selector}); err != nil {
+		return err
+	}
+
+	for _, dv := range dvList.Items {
```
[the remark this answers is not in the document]
ido106: At first I did that, but then we decided to delete all those DVs because they are all old.
@arnongilboa 
Maybe we'll change it back?

pkg/controller/dataimportcron-controller.go
```diff
@@ -832,6 +832,24 @@ func (r *DataImportCronReconciler) garbageCollectPVCs(ctx context.Context, names
...
+		pvc := &corev1.PersistentVolumeClaim{}
+		if err := r.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: dv.Name}, pvc); err != nil {
+			return err
+		}
+
+		if pvc.Labels[common.DataImportCronLabel] != cronName {
+			if err := r.deleteDvPvc(ctx, dv.Name, dv.Namespace); err != nil {
```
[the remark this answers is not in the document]
ido106: Okay

pkg/controller/dataimportcron-controller.go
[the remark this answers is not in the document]
arnongilboa: Yep, we would like to delete these (very) old ones and no need to sort, but we can sure add
```go
if len(dvList.Items) > maxImports
```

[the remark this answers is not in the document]
ido106: > 

ok

ido106:
The functional test works on my machine. Someone know what's the problem?

alromeros:
> The functional test works on my machine. Does someone know what the problem is?

You probably need to add  `f.ForceBindPvcIfDvIsWaitForFirstConsumer(dv)` so the DV can start the import. The test probably works in your machine if you are using a storage class without CSI drivers and immediate binding (local, for example).

ido106:
> > The functional test works on my machine. Does someone know what the problem is?
> 
> You probably need to add `f.ForceBindPvcIfDvIsWaitForFirstConsumer(dv)` before `WaitForDataVolumePhase` so the DV can start the import. The test probably works in your machine if you are using a storage class without CSI drivers and immediate binding (local, for example).
> 
> Edit: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2749/pull-containerized-data-importer-e2e-nfs/1698335348150505472 is failing in a different section, so there's probably other stuff that needs to be fixed. I can take a look later today.

Great, thanks Alvaro

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -706,6 +706,64 @@ var _ = Describe("All DataImportCron Tests", func() {
 			Entry("has no tag", imageStreamName, 1),
 		)
 
+		It("XXXXXX Should succeed garbage collecting old version DVs", func() {
```
akalenyu, line 709: Don't need test ID in unit tests (XXXXXX)

tests/dataimportcron_test.go
```diff
@@ -367,6 +368,120 @@ var _ = Describe("DataImportCron", func() {
...
+			Expect(snapshot.Status.ReadyToUse).ToNot(BeNil())
+			Expect(*snapshot.Status.ReadyToUse).To(BeTrue())
+		}
+
+		By("Check garbage collection")
+		Eventually(func() int {
+			dvList, err := f.CdiClient.CdiV1beta1().DataVolumes(ns).List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			return len(dvList.Items)
+		}, dataImportCronTimeout, pollingInterval).Should(Equal(0), "Garbage collection failed cleaning old imports")
```
akalenyu, line 442: I think this is why the NFS lane fails;
(With non-snapshot-capable storage) You will always have 1 main (non-GCd) datavolume present (latest pull) 
so you will never reach `len(dvList.Items) == 0`

Maybe using [test_id:7406] will make life easier and test runtime lower, but I am not sure,
let me know what you think

akalenyu:
> The functional test works on my machine. Does someone know what the problem is?

You should be able to reproduce the failures if you start a fresh dev cluster
with `KUBEVIRT_STORAGE=hpp`

pkg/controller/dataimportcron-controller_test.go
ido106: Fixed

tests/dataimportcron_test.go
```diff
@@ -367,6 +368,120 @@ var _ = Describe("DataImportCron", func() {
 		Entry("[test_id:8266] succeed deleting error DVs when importing new ones", false, true, 2, cdiv1.DataImportCronSourceFormatPvc),
 	)
 
+	It("[test_id:XXXX] succeed garbage collecting old version dv's", func() {
```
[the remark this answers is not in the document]
ido106: I will take a look at it

tests/dataimportcron_test.go
ido106: I will take a look at `[test_id:7406] ` and see if I can integrate the test there instead.
@arnongilboa wdyt?

ido106:
> > The functional test works on my machine. Does someone know what the problem is?
> 
> You should be able to reproduce the failures if you start a fresh dev cluster with `KUBEVIRT_STORAGE=hpp`

Great, thanks !

ido106:
Is 7406 cool now ?
@akalenyu

tests/dataimportcron_test.go
```diff
@@ -367,6 +369,120 @@ var _ = Describe("DataImportCron", func() {
 		Entry("[test_id:8266] succeed deleting error DVs when importing new ones", false, true, 2, cdiv1.DataImportCronSourceFormatPvc),
 	)
 
+	It("[test_id:XXXX] Succeed garbage collecting old version DVs", func() {
```
akalenyu, line 372: We can delete this test now that we are doing the check in 7406

tests/dataimportcron_test.go
```diff
@@ -506,6 +646,12 @@ var _ = Describe("DataImportCron", func() {
 		By("Check garbage collection")
 		switch format {
 		case cdiv1.DataImportCronSourceFormatPvc:
+			By("Check old DV garbage collection")
+			Eventually(func() bool {
+				_, err := f.CdiClient.CdiV1beta1().DataVolumes(ns).Get(context.TODO(), oldDvName, metav1.GetOptions{})
+				return err != nil
+			}, dataImportCronTimeout, pollingInterval).Should(Equal(true), "Garbage collection failed cleaning old DV")
```
akalenyu, line 653: Nit, if you want, can do something like this:
```
Eventually(func() error {
	_, err := f.CdiClient.CdiV1beta1().DataVolumes(ns).Get(context.TODO(), oldDvName, metav1.GetOptions{})
	return err
}, dataImportCronTimeout, pollingInterval).Should(Satisfy(errors.IsNotFound), "Garbage collection failed cleaning old DV")
```

akalenyu (commented):
Looks good just not sure if you meant to keep the separate test

tests/dataimportcron_test.go
ido106: Yup, I was just waiting to see if '7406' is good before deleting the old test

tests/dataimportcron_test.go
ido106: Great, I will fix it

pkg/controller/dataimportcron-controller.go
```diff
@@ -832,6 +832,27 @@ func (r *DataImportCronReconciler) garbageCollectPVCs(ctx context.Context, names
...
+
+	if len(dvList.Items) > maxImports {
+		for _, dv := range dvList.Items {
+			pvc := &corev1.PersistentVolumeClaim{}
+			if err := r.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: dv.Name}, pvc); err != nil {
+				return err
+			}
+
+			if pvc.Labels[common.DataImportCronLabel] != cronName {
+				r.log.Info("Deleting dv/pvc", "name", pvc.Name, "pvc.uid", pvc.UID)
```
arnongilboa, line 848: I think this log line should be a bit different from the one above, e.g. "Deleting old version dv/pvc"?

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -706,6 +706,64 @@ var _ = Describe("All DataImportCron Tests", func() {
 			Entry("has no tag", imageStreamName, 1),
 		)
 
+		It("Should succeed garbage collecting old version DVs", func() {
+			cron = newDataImportCron(cronName)
+			importsToKeep := int32(1)
+			cron.Spec.ImportsToKeep = &importsToKeep
+			cron.Annotations[AnnSourceDesiredDigest] = testDigest
```
arnongilboa, line 713: why do you need to set digest?

tests/dataimportcron_test.go
```diff
@@ -48,6 +48,7 @@ var _ = Describe("DataImportCron", func() {
 		ns                  string
 		scName              string
 		originalProfileSpec *cdiv1.StorageProfileSpec
+		oldDvName           = "src-garbage-0"
```
arnongilboa, line 51: oldDvName should move into the function, as nobody else is using it

tests/dataimportcron_test.go
```diff
@@ -447,7 +448,7 @@ var _ = Describe("DataImportCron", func() {
 		configureStorageProfileResultingFormat(format)
 
 		garbageSources := 3
-		for i := 0; i < garbageSources; i++ {
+		for i := 1; i <= garbageSources; i++ {
 			srcName := fmt.Sprintf("src-garbage-%d", i)
```
arnongilboa, line 452: Not sure I like this naming hack, oldDvName should have a better name than "src-garbage-0", e.g. "old-version-dv", and the loop can be reverted to 0-based.

pkg/controller/dataimportcron-controller.go
ido106: Fixed

pkg/controller/dataimportcron-controller_test.go
ido106: You're right, we don't need it. Probably left over from the old test I wrote.
Fixed

tests/dataimportcron_test.go
ido106: Okay

tests/dataimportcron_test.go
ido106: Sure, no problem. Fixed

ido106:
Do you have any more comments?
@arnongilboa

PR #2753, opened 2023-06-14
Allow dynamic linked build for non bazel build

awels:
The current script always passes the static ldflag to the compiler which will result in a static binary. We would like to be able to build dynamic libraries instead.

cdi-containerimage-server has to be static because we are copying it into the context of a container disk container which is most likely based on a scratch container and has no libraries for us to use.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/build-go.sh
+++ b/hack/build/build-go.sh
@@ -54,10 +54,14 @@ elif [ "${go_opt}" == "build" ]; then
         outLink=${BIN_DIR}/${BIN_NAME}
         rm -f ${outFile}
         rm -f ${outLink}
+        static_flag=""
+        if [ "$tgt" == "tools/cdi-containerimage-server" ]; then
+          static_flag="static"
+        fi
         (
             cd $tgt
             # Only build executables for linux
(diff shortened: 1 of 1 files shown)


hack/build/build-go.sh
```diff
@@ -54,10 +54,14 @@ elif [ "${go_opt}" == "build" ]; then
...
         rm -f ${outLink}
+        static_flag=""
+        if [ "$tgt" == "tools/cdi-containerimage-server" ]; then
+          static_flag="static"
+        fi
         (
             cd $tgt
             # Only build executables for linux
-            GOOS=linux go build -o ${outFile} -ldflags '-extldflags "static"' -ldflags "$(cdi::version::ldflags)"
+            GOOS=linux go build -o ${outFile} -tags strictfipsruntime -ldflags '-extldflags $static_flag' -ldflags "$(cdi::version::ldflags)"
```
aglitke, line 64: I guess it works to pass a naked `-extldflags` arg?

PR #2754, opened 2023-06-15
Disable DV GC by default

arnongilboa:
`DataVolume` garbage collection is a nice feature, but unfortunately it violates fundamental principle of Kubernetes. CR should not be auto-deleted when it completes its role (`Job` with `TTLSecondsAfterFinished` is an exception), and once CR was created we can assume it is there until explicitly deleted. In addition, CR should keep idempotency, so the same CR manifest can be applied multiple times, as long as it is a valid update (e.g. `DataVolume` validation webhook does not allow updating the spec).

When GC is enabled, some systems (e.g GitOps / ArgoCD) may require a workaround (DV annotation `deleteAfterCompletion = false`) to prevent GC and function correctly.

On the next kubevirt-bot Bump kubevirtci PR (with bump-cdi), it will fail on all kubevirtci lanes with tests referring DVs, as the tests `IsDataVolumeGC()` looks at `CDIConfig Spec.DataVolumeTTLSeconds` and assumes default is enabled. This should be fixed there. Disabling GC on all kubevirt CI lanes beforehand can eliminate the failures, see [PR](https://github.com/kubevirt/kubevirt/pull/9921).

Fixes #

**Release note**:
```release-note
Disable DataVolume garbage collection by default
```

### DIFF
--- a/pkg/apiserver/webhooks/datavolume-mutate.go
+++ b/pkg/apiserver/webhooks/datavolume-mutate.go
@@ -86,12 +86,11 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 		if err != nil {
 			return toAdmissionResponseError(err)
 		}
+		// Garbage collection is disabled by default
+		// Annotate DV for GC only if GC is enabled in CDIConfig and the DV is not annotated to prevent GC
 		if cc.GetDataVolumeTTLSeconds(config) >= 0 {
-			if modifiedDataVolume.Annotations == nil {
-				modifiedDataVolume.Annotations = make(map[string]string)
-			}
 			if modifiedDataVolume.Annotations[cc.AnnDeleteAfterCompletion] != "false" {
-				modifiedDataVolume.Annotations[cc.AnnDeleteAfterCompletion] = "true"
+				cc.AddAnnotation(modifiedDataVolume, cc.AnnDeleteAfterCompletion, "true")
 			}
 		}
 	}
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -1211,8 +1211,9 @@ func IsErrCacheNotStarted(err error) bool {
 }
 
 // GetDataVolumeTTLSeconds gets the current DataVolume TTL in seconds if GC is enabled, or < 0 if GC is disabled
+// Garbage collection is disabled by default
 func GetDataVolumeTTLSeconds(config *cdiv1.CDIConfig) int32 {
-	const defaultDataVolumeTTLSeconds = 0
+	const defaultDataVolumeTTLSeconds = -1
 	if config.Spec.DataVolumeTTLSeconds != nil {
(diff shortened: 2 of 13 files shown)


mhenriks:
looks okay, what's the deal with the test failures?

awels:
Looks like most failures are happening here https://github.com/kubevirt/containerized-data-importer/blob/main/tests/utils/datavolume.go#L148

arnongilboa:
> looks okay, what's the deal with the test failures?

The test was incorrect for the no-GC flow. It should check that PVC with specific UID was deleted, as upon deletion of the last imported PVC DIC creates a new one.

alromeros:
Looks good!
/lgtm

pkg/controller/common/util.go
```diff
@@ -1212,7 +1212,7 @@ func IsErrCacheNotStarted(err error) bool {
 
 // GetDataVolumeTTLSeconds gets the current DataVolume TTL in seconds if GC is enabled, or < 0 if GC is disabled
 func GetDataVolumeTTLSeconds(config *cdiv1.CDIConfig) int32 {
-	const defaultDataVolumeTTLSeconds = 0
+	const defaultDataVolumeTTLSeconds = -1
```
akalenyu, line 1215: Maybe a comment here that GC is disabled by default?

pkg/apiserver/webhooks/datavolume-mutate.go
```diff
@@ -87,11 +87,8 @@ func (wh *dataVolumeMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admi
 			return toAdmissionResponseError(err)
 		}
 		if cc.GetDataVolumeTTLSeconds(config) >= 0 {
-			if modifiedDataVolume.Annotations == nil {
-				modifiedDataVolume.Annotations = make(map[string]string)
-			}
 			if modifiedDataVolume.Annotations[cc.AnnDeleteAfterCompletion] != "false" {
-				modifiedDataVolume.Annotations[cc.AnnDeleteAfterCompletion] = "true"
+				cc.AddAnnotation(modifiedDataVolume, cc.AnnDeleteAfterCompletion, "true")
```
akalenyu, line 91: Maybe worth a comment as well, similar to `const defaultDataVolumeTTLSeconds = -1`
as these are the GC defaulting points

akalenyu (commented):
PR looks good, my main concern is versioning this.

Is this planned to merge into 1.57?
It may be ready to do so, but we should carefully decide if that's where we want it.

/cc @awels

pkg/controller/common/util.go
arnongilboa: added

arnongilboa:
> PR looks good, my main concern is versioning this.
> 
> Is this planned to merge into 1.57? It may be ready to do so, but we should carefully decide if that's where we want it.
> 
> /cc @awels

/hold

@aglitke

PR #2759, opened 2023-06-18
Fix sizeless clones for sourceRef data volumes

akalenyu:
We were hitting a panic because we're passing the original DV object to renderPvcSpec (source is nil)
instead of the mutated DV which has sourceRef converted (source not nil)

Fixes #

**Release note**:

```release-note
BugFix: sizeless clones cause panic for sourceRef data volumes
```

### DIFF
--- a/pkg/controller/datavolume/util.go
+++ b/pkg/controller/datavolume/util.go
@@ -292,7 +292,7 @@ func resolveVolumeSize(c client.Client, dvSpec cdiv1.DataVolumeSpec, pvcSpec *v1
 
 	if !found {
 		// Storage size can be empty when cloning
-		isClone := dvSpec.Source.PVC != nil
+		isClone := dvSpec.Source.PVC != nil || dvSpec.Source.Snapshot != nil
 		if isClone {
 			return &requestedSize, nil
 		}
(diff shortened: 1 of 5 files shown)


pkg/controller/datavolume/util.go
```diff
@@ -292,7 +292,7 @@ func resolveVolumeSize(c client.Client, dvSpec cdiv1.DataVolumeSpec, pvcSpec *v1
 
 	if !found {
 		// Storage size can be empty when cloning
-		isClone := dvSpec.Source.PVC != nil
+		isClone := dvSpec.Source.PVC != nil || dvSpec.Source.Snapshot != nil
```
alromeros, line 295: We don't do regular size-detection when cloning from a snapshot, are we sure this works? For example, what would happen if the target DV size is empty and the snapshot `restoreSize` is zero?

akalenyu: covered by https://github.com/kubevirt/containerized-data-importer/blob/5f85c423e16c71de32d3011eaca2545afb8c347d/pkg/controller/common/util.go#L790-L792

akalenyu:
/cc @mhenriks
I skipped the smart clone controller fix since I know you're removing that,
but I think that once your PR is in the issue should not occur for snapshot pvc clones

PR #2769, opened 2023-06-22
Handle nil ptr in dataimportcron controller

akalenyu:
We see a spike in failures for this test:
https://search.ci.kubevirt.io/?search=8033&maxAge=336h&context=1&type=bug%2Bissue%2Bjunit&name=containerized-data-importer&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job
CDI controller crashes because of a nil pointer

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -336,8 +336,10 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	}
 
 	handlePopulatedPvc := func() error {
-		if err := r.updateSource(ctx, dataImportCron, pvc); err != nil {
-			return err
+		if pvc != nil {
+			if err := r.updateSource(ctx, dataImportCron, pvc); err != nil {
+				return err
+			}
 		}
 		importSucceeded = true
 		if err := r.handleCronFormat(ctx, dataImportCron, format, dvStorageClass); err != nil {
--- a/tests/dataimportcron_test.go
+++ b/tests/dataimportcron_test.go
@@ -38,7 +38,7 @@ const (
 
 var _ = Describe("DataImportCron", func() {
 	var (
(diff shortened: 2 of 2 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -336,8 +336,10 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 	}
 
 	handlePopulatedPvc := func() error {
-		if err := r.updateSource(ctx, dataImportCron, pvc); err != nil {
-			return err
+		if pvc != nil {
```
awels, line 339: Instead of checking here, check in updateSource if the object is nil? that way other callers are also protected. If it is nil, just return nil as you can't update the object anyway.

akalenyu: We're still not sure exactly what is going on.. but once we know this fixes it we can do that

akalenyu:
@arnongilboa I noticed the failing test doesn't even create DVs because of the wrong config map name

So another theory is that one of the previous test namespaces gets deleted in such an order that
triggers this scenario

pkg/controller/dataimportcron-controller.go
akalenyu: So if we want to check for nil in updateSource (receives the `client.Object` interface),
we must use reflect, which I am not sure is worth it.. wdyt?
(reflect.ValueOf(i).IsNil())

alromeros: I personally don't see the problem with using reflect but most of the uses of `updateSource` already check the object before calling the function. I think it'd be enough to add the check here.

mhenriks: I like this fix as-is

PR #2784, opened 2023-07-02
Follow up for GC=on in some tests

akalenyu:
This may be the reason for the flakes we're seeing: https://search.ci.kubevirt.io/?search=9924&maxAge=336h&context=1&type=bug%2Bissue%2Bjunit&name=containerized-data-importer&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

With GC on we want to delete the DV (PVC won't cut it)
We may have more cases like this one scattered around, so let's keep an eye

Not sure how populators DVs handle that situation but we see that the test PV is simply gone from artifacts

Fixes #

**Release note**:

```release-note
NONE
```


ShellyKa13 (commented):
/lgtm

PR #2787, opened 2023-07-04
dataVolume: Add default instance type labels from source

### LINKED ISSUE #2782: Default instance type labels not passed from `DataVolume` source `DataSource` or `PVC` to the resulting `PVC`
**What happened**:

It could be argued that this is more of a feature request but for now I'll leave it marked as `kind/bug`.

The original passing of default instance type labels targeted the `DataImportCron` -> `DataVolume` and `DataVolume` -> `DataSource` and `PVC` use cases. This missed the common use case of an unlabelled `DataVolume` using a labelled `DataSource` or `PVC` however, where a user might expect the labels from the sources to end up on the resulting `PVC` etc.

Take the following example where we use a labelled `DataVolume` to create a labelled `PVC`. We then create another unlabelled `DataVolume` using the labelled `PVC` as a source, at present this results in a `PVC` that is unlabelled:

```
$ ./cluster-up/kubectl.sh apply -f - <<EOF
> apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: "example-import-dv"
  annotations:
    "cdi.kubevirt.io/storage.bind.immediate.requested": "true"
  labels:
    "instancetype.kubevirt.io/default-instancetype": "u1.medium"
    "instancetype.kubevirt.io/default-preference": "cirros"
spec:
  source:
      http:
         url: "https://download.cirros-cloud.net/0.6.2/cirros-0.6.2-x86_64-disk.img"
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: "1Gi"
> EOF

$ ./cluster-up/kubectl.sh get pvc/example-import-dv -o json | jq .metadata.labels 
selecting podman as container runtime
{
  "alerts.k8s.io/KubePersistentVolumeFillingUp": "disabled",
  "app": "containerized-data-importer",
  "app.kubernetes.io/component": "storage",
  "app.kubernetes.io/managed-by": "cdi-controller",
  "instancetype.kubevirt.io/default-instancetype": "u1.medium",
  "instancetype.kubevirt.io/default-preference": "cirros"
}

$ ./cluster-up/kubectl.sh apply -f - <<EOF
> apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: "example-clone-dv"
  annotations:
    "cdi.kubevirt.io/storage.bind.immediate.requested": "true"
spec:
  source:
      pvc:
        name: example-import-dv
        namespace: default
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: "1Gi"
> EOF
selecting podman as container runtime
datavolume.cdi.kubevirt.io/example-clone-dv created

$ ./cluster-up/kubectl.sh get pvc/example-clone-dv -o json | jq .metadata.labels
selecting podman as container runtime
{
  "alerts.k8s.io/KubePersistentVolumeFillingUp": "disabled",
  "app": "containerized-data-importer",
  "app.kubernetes.io/component": "storage",
  "app.kubernetes.io/managed-by": "cdi-controller"
}
```

Passing any default labels from the original source to the `DataVolume` should be enough here with the existing logic then passing these onto the created `PVC`.

As above.

**Additional context**:

Also tracked under https://github.com/kubevirt/kubevirt/issues/9894 as I'll likely land more functional test coverage in KubeVirt once a fix lands and is available in CDI.

**Environment**:
- DV specification: N/A
- Install tools: N/A
- Others: N/A

lyarwood:

https://github.com/kubevirt/containerized-data-importer/commit/f229aeb5ff69b3d89b537b39f547eb6c2bc3d5de started to pass default instance type labels from
DataImportCrons to DataVolumes and DataVolumes to any associated
destination DataSources or PVCs. As documented in issue https://github.com/kubevirt/containerized-data-importer/issues/2782 this does
not however pass these labels from the initial source of a DataVolume to
either the DataVolume or the destination DataSources or PVCs

This change corrects this by updating DataVolumes when reconciled,
adding any labels found on PVC or DataSource sources. These labels will
then be passed on to the destination PVC or DataSources by the existing
functionality highlighted above.

Note that if any default instance type labels already exist on the
DataVolume then the process is skipped as it is assumed these are
provided either directly by the user or via a DataImportCron.

Fixes #2782

Depends on https://github.com/kubevirt/containerized-data-importer/pull/2818

Functional tests still missing, will try to add them later this week but reviews welcome in the meantime!

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -467,6 +467,10 @@ func (r *ReconcilerBase) syncDvPvcState(log logr.Logger, req reconcile.Request, 
 		return syncState, err
 	}
 
+	if err = updateDataVolumeDefaultInstancetypeLabels(r.client, &syncState); err != nil {
+		return syncState, err
+	}
+
 	if syncState.pvc != nil {
 		if err := r.garbageCollect(&syncState, log); err != nil {
 			return syncState, err
--- a/pkg/controller/datavolume/util.go
+++ b/pkg/controller/datavolume/util.go
@@ -351,6 +351,55 @@ func updateDataVolumeUseCDIPopulator(syncState *dvSyncState) {
 	cc.AddAnnotation(syncState.dvMutated, cc.AnnUsePopulator, strconv.FormatBool(syncState.usePopulator))
 }
 
+func updateDataVolumeDefaultInstancetypeLabels(client client.Client, syncState *dvSyncState) error {
+	// Skip looking anything up if any default instance type labels are already present
+	dv := syncState.dvMutated
+	for _, defaultInstanceTypeLabel := range cc.DefaultInstanceTypeLabels {
+		if _, ok := dv.Labels[defaultInstanceTypeLabel]; ok {
+			return nil
+		}
+	}
+	if dv.Spec.Source != nil && dv.Spec.Source.PVC != nil {
+		pvc := &v1.PersistentVolumeClaim{}
+		key := types.NamespacedName{
+			Name:      dv.Spec.Source.PVC.Name,
+			Namespace: dv.Spec.Source.PVC.Namespace,
+		}
+		if err := client.Get(context.TODO(), key, pvc); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+		copyDefaultInstancetypeLabels(dv, pvc.Labels)
+		return nil
+	}
+	if dv.Spec.SourceRef != nil && dv.Spec.SourceRef.Namespace != nil && dv.Spec.SourceRef.Kind == cdiv1.DataVolumeDataSource {
+		ds := &cdiv1.DataSource{}
+		key := types.NamespacedName{
+			Name:      dv.Spec.SourceRef.Name,
+			Namespace: *dv.Spec.SourceRef.Namespace,
+		}
+		if err := client.Get(context.TODO(), key, ds); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+		copyDefaultInstancetypeLabels(dv, ds.Labels)
+		return nil
+	}
+	return nil
+}
+
+func copyDefaultInstancetypeLabels(dataVolume *cdiv1.DataVolume, labels map[string]string) {
+	for _, defaultInstancetypeLabel := range cc.DefaultInstanceTypeLabels {
+		if v, ok := labels[defaultInstancetypeLabel]; ok {
+			cc.AddLabel(dataVolume, defaultInstancetypeLabel, v)
+		}
+	}
+}
+
 func checkDVUsingPopulators(dv *cdiv1.DataVolume) (bool, error) {
 	usePopulator, ok := dv.Annotations[cc.AnnUsePopulator]
 	if !ok {
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -320,6 +320,14 @@ var (
 	// FilesystemMode is filesystem device mode
 	FilesystemMode = corev1.PersistentVolumeFilesystem
 
+	// DefaultInstanceTypeLabels is a list of currently supported default instance type labels
+	DefaultInstanceTypeLabels = []string{
+		LabelDefaultInstancetype,
+		LabelDefaultInstancetypeKind,
+		LabelDefaultPreference,
+		LabelDefaultPreferenceKind,
+	}
+
 	apiServerKeyOnce sync.Once
 	apiServerKey     *rsa.PrivateKey
 )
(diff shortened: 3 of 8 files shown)


arnongilboa (commented):
Why do it in the mutate webhook and not in the controller? what about cases where the source PVC does not exist yet?

lyarwood:
> Why do it in the mutate webhook and not in the controller? what about cases where the source PVC does not exist yet?

Is a non-existent source PVC supported during DataVolume submission? If it is we could move this logic to the controller but it would ultimately lead to more lookups of the source PVC/DataSource during the lifecycle of the DataVolume.

arnongilboa:
> > Why do it in the mutate webhook and not in the controller? what about cases where the source PVC does not exist yet?
> 
> Is a non-existent source PVC supported during DataVolume submission? If it is we could move this logic to the controller but it would ultimately lead to more lookups of the source PVC/DataSource during the lifecycle of the DataVolume.

See #2306 and #2571

lyarwood:
> > > Why do it in the mutate webhook and not in the controller? what about cases where the source PVC does not exist yet?
> > 
> > 
> > Is a non-existent source PVC supported during DataVolume submission? If it is we could move this logic to the controller but it would ultimately lead to more lookups of the source PVC/DataSource during the lifecycle of the DataVolume.
> 
> See #2306 and #2571

Thanks I'll rework this next week.

pkg/controller/datavolume/controller-base.go
```diff
@@ -481,6 +481,10 @@ func (r *ReconcilerBase) syncDvPvcState(log logr.Logger, req reconcile.Request,
 		r.handlePrePopulation(syncState.dvMutated, syncState.pvc)
 	}
 
+	if err = updateDataVolumeDefaultInstancetypeLabels(r.client, &syncState); err != nil {
```
lyarwood, line 484: @arnongilboa - if you have a second, does updating the `DataVolume` at this point make sense or should this be done elsewhere? I'm also ignoring errors here given you previously highlighted that the `PVC` and/or `DataSource` might not exist yet.

pkg/controller/datavolume/util.go
```diff
@@ -351,6 +351,47 @@ func updateDataVolumeUseCDIPopulator(syncState *dvSyncState) {
...
+		ds := &cdiv1.DataSource{}
+		key := types.NamespacedName{
+			Name:      dv.Spec.SourceRef.Name,
+			Namespace: *dv.Spec.SourceRef.Namespace,
+		}
+		if err := client.Get(context.TODO(), key, ds); err != nil {
+			return err
+		}
+		copyDefaultInstancetypeLabels(dv, ds.Labels)
+	}
```
arnongilboa, line 383: why check both Source and SourceRef? I guess you should return after copyDefaultInstancetypeLabels().

pkg/controller/datavolume/controller-base.go
arnongilboa: Putting it right after the other DV updates (before `if syncState.pvc != nil`) is better.
btw why return error if you ignore it without even checking which error is it?

lyarwood: Only to log the error, happy to move that down into `updateDataVolumeDefaultInstancetypeLabels` if you'd prefer.

pkg/controller/datavolume/util.go
lyarwood: We label both `DataSources` and `PVCs` so checking both is valid IMHO. I'll add the early return now.

pkg/controller/datavolume/controller-base.go
arnongilboa: I think that if !IsNotFound(err) you should `return syncState, err`

lyarwood:
@arnongilboa  if you have anytime this week this should be ready to review again, many thanks in advance!

arnongilboa:
@lyarwood please remove the `chore: Run make deps-update` commit, as it's already merged

lyarwood:
@mhenriks, @awels , if either of you have time this week would you mind taking a look at this, many thanks in advance!

pkg/controller/datavolume/util.go
```diff
@@ -351,6 +351,55 @@ func updateDataVolumeUseCDIPopulator(syncState *dvSyncState) {
...
 
+func updateDataVolumeDefaultInstancetypeLabels(client client.Client, syncState *dvSyncState) error {
+	// Skip looking anything up if any default instance type labels are already present
+	dv := syncState.dvMutated
+	for _, defaultInstanceTypeLabel := range cc.DefaultInstanceTypeLabels {
+		if _, ok := dv.Labels[defaultInstanceTypeLabel]; ok {
+			return nil
+		}
+	}
+	if dv.Spec.Source != nil && dv.Spec.Source.PVC != nil {
```
mhenriks, line 362: what about `dv.Spec.Source.Snapshot`?

mhenriks: Eventually `DataImportCrons` will create "golden snapshots" instead of "golden pvcs" for provisioners in which that is optimal

lyarwood: ACK thanks I thought you might say that, definitely something to follow up with in that case!

PR #2788, opened 2023-07-05
Update to ginkgo v2

maya-r:
We want to use new features of ginkgo v2, but it's a big enough change to warrant being a separate PR.

Had to update qe-tools as well.

**Release note**:

```release-note
NONE
```


maya-r:
WIP title because I didn't do a full test run locally. I'm using the CI for this.

akalenyu:
kubevirt reference even though I am probably late to the party: https://github.com/kubevirt/kubevirt/pull/7326

maya-r:
I'm not sure why, but the junit xml file is not generated, so the new way to handle custom reporters is not working

akalenyu:
Looks good, but the xml result of the test run is missing?
https://gcsweb.ci.kubevirt.io/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2788/pull-containerized-data-importer-e2e-ceph/1678877872149762048/
It's usually in here on a healthy run:
https://gcsweb.ci.kubevirt.io/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2795/pull-containerized-data-importer-e2e-ceph/1678871111246286848/artifacts/

There are also these AfterEach failures that are not reported as failures:
[FAILED] in [AfterEach] - /root/go/src/kubevirt.io/containerized-data-importer/tests/import_test.go:395 @ 07/12/23 00:22:24.359
S [SKIPPED] [0.022 seconds]
They seem to occur after a Skip() was called in BeforeEach

akalenyu:
> I'm not sure why, but the junit xml file is not generated, so the new way to handle custom reporters is not working

I think this is the relevant bit
https://github.com/kubevirt/kubevirt/pull/7326/commits/46ffcbf1bacd3cdf1ae074d92dac74669b113c80#diff-c65f096e307113b821ebd29ba506abcd493b3ec18eb0c925da313180c3c42f65R16

oshoval:
Hi, focusing atm only about the report,
sorry, don't remember all the details
but you can either use the deprecated reporter (that was adapted on kubevirt to support ginkgo v2, until we move to the new reporter), or even better use the new build-in one
for example like you can see here
https://github.com/kubevirt/cluster-network-addons-operator/pull/1260/commits/e0cf4b81120544e381a7b6a866986cc31655bbd1#diff-b58397a7db869638f0d4c5cd1b484f93e985ad35de25ddbd0ce77148bbf69dfdR40

but not a mix, this what i remember

automation/test.sh
```diff
@@ -110,7 +110,7 @@ make cluster-sync
 
 kubectl version
 
-ginko_params="--test-args=--ginkgo.timeout=24h --ginkgo.noColor --junit-output=${ARTIFACTS_PATH}/junit.functest.xml"
+ginko_params="--test-args=--ginkgo.timeout=24h --ginkgo.noColor --ginkgo.junit-report=${ARTIFACTS_PATH}/junit.functest.xml"
```
oshoval, line 113: note that ARTIFACT_DIR might need to be supplied in other argument,
at least this what i was the state at the past
otherwise it didnt work well

oshoval, line 113: ginkgo.noColor is also deprecated , please see the log

oshoval:
> > I'm not sure why, but the junit xml file is not generated, so the new way to handle custom reporters is not working
> 
> I think this is the relevant bit [kubevirt/kubevirt@46ffcbf#diff-c65f096e307113b821ebd29ba506abcd493b3ec18eb0c925da313180c3c42f65R16](https://github.com/kubevirt/kubevirt/commit/46ffcbf1bacd3cdf1ae074d92dac74669b113c80#diff-c65f096e307113b821ebd29ba506abcd493b3ec18eb0c925da313180c3c42f65R16)

if you can remove refs to this one would be good

automation/test.sh
maya-r: I don't understand what you're saying

oshoval: there was a problem in the past that `ARTIFACTS_PATH` had to be supplied with --output-dir or so
else it won't work if it is part of the junit-report
but if it works for you, then all good already

akalenyu:
Looks good! trying to understand why we get the same error printed twice
in test failure stdouts:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2788/pull-containerized-data-importer-e2e-ceph/1680876813745655808

akalenyu:
Would you mind adding ginkgo trace https://github.com/kubevirt/kubevirt/pull/7345/commits/a965d83941bb8baa561c51749e16927679bdc5ee
to this PR? seems like a nice addition for debugging

akalenyu:
/lgtm
/cc @awels 

I think we can consider more ginkgo v2 related changes as follow up,
For example, this should tackle the double err output:
https://github.com/kubernetes/kubernetes/blob/3e2b26ce52cacde090aba6ee593b8213cb24419e/test/e2e/framework/internal/junit/junit.go#L34-L38
The bump itself looks good, and the same set of tests seems to be run

awels (commented):
/approve

PR #2814, opened 2023-07-19
Annotate PVC with host-assisted clone fallback reason; add missing events

arnongilboa:

Fixes #

**Release note**:
```release-note
Annotate PVC with host-assisted clone fallback reason; add missing events
```


pkg/controller/clone/planner.go
```diff
@@ -126,9 +132,10 @@ func (p *Planner) AddCoreWatches(log logr.Logger) error {
 
 // ChooseStrategyArgs are args for ChooseStrategy function
 type ChooseStrategyArgs struct {
-	Log         logr.Logger
-	TargetClaim *corev1.PersistentVolumeClaim
-	DataSource  *cdiv1.VolumeCloneSource
+	Log            logr.Logger
+	TargetClaim    *corev1.PersistentVolumeClaim
+	DataSource     *cdiv1.VolumeCloneSource
+	FallbackReason string
```
mhenriks, line 138: I think it would be better to have `ChooseStrategy` return a struct that that the strategy and fallbackreason

mhenriks: Something like
```
type ChooseStrategyResult struct{
    Strategy cdiv1.CDICloneStrategy
    FallbackReason *string
}
```

arnongilboa: yep, smells better

pkg/controller/populators/clone-populator.go
```diff
@@ -48,6 +48,9 @@ const (
 	// AnnCloneError has the error string for error phase
 	AnnCloneError = "cdi.kubevirt.io/cloneError"
 
+	// AnnCloneFallbackReason has the host-assisted clone fallback reason
+	AnnCloneFallbackReason = "cdi.kubevirt.io/cloneFallbackReason"
```
mhenriks, line 52: DataVolume controller should add this annotation when using non-csi clone path

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -114,6 +115,10 @@ const (
 	RebindInProgress = "RebindInProgress"
 	// MessageRebindInProgress is a const for reporting target rebind
 	MessageRebindInProgress = "Rebinding PersistentVolumeClaim for DataVolume %s/%s"
+	// NoPopulator reports CDI populator is not used so we fallback to host-assisted cloning (reason)
+	NoPopulator = "NoPopulator"
+	// NoPopulatorMessage reports CDI populator is not used so we fallback to host-assisted cloning (message)
+	NoPopulatorMessage = "CDI populator is not used"
```
mhenriks, line 121: I think something like "in tree storage class does not support snapshot/clone" would be better

mhenriks (commented):
no functional test possible?

arnongilboa:
> no functional test possible?

added

PR #2821, opened 2023-07-27
Fix race condition in populators so we clean up PVC' after population succeeds

alromeros:

This Pull Request fixes a race condition in CDI populators so we fetch the PVC' even if the population source doesn't exist.

This new behavior allows us to correctly delete the PVC' once the population has succeeded. We assume that, if PVC' already exists, all the requirements for population have already been met, so we don't need to check them all again.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2223361

**Release note**:

```release-note
Bugfix: Clean up PVC' when population succeeds even if the population source doesn't exist
```


alromeros:
/cc @mhenriks 
I can't think of any unintended consequence of returning PVC' before the other `reconcileCommon` checks, wdyt?

PR #2823, opened 2023-07-27
Fix DIC returned reconcile.Result for ImageStream

arnongilboa:

Regression introduced in #2700

Fixes bz #2226982

**Release note**:
```release-note
Fix DataImportCron ImageStream polling per schedule
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -394,13 +394,14 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		return res, err
 	}
 
-	// We use the poller returned reconcile.Result for RequeueAfter if needed
-	// skip if we disabled schedule
+	// Skip if schedule is disabled
 	if isImageStreamSource(dataImportCron) && dataImportCron.Spec.Schedule != "" {
-		res, err := r.pollImageStreamDigest(ctx, dataImportCron)
+		// We use the poll returned reconcile.Result for RequeueAfter if needed
+		pollRes, err := r.pollImageStreamDigest(ctx, dataImportCron)
 		if err != nil {
-			return res, err
+			return pollRes, err
 		}
+		res = pollRes
 	}
 
 	desiredDigest := dataImportCron.Annotations[AnnSourceDesiredDigest]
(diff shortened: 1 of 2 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -394,10 +394,11 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
 	}
 
-	// We use the poller returned reconcile.Result for RequeueAfter if needed
-	// skip if we disabled schedule
+	// Skip if schedule is disabled
 	if isImageStreamSource(dataImportCron) && dataImportCron.Spec.Schedule != "" {
-		res, err := r.pollImageStreamDigest(ctx, dataImportCron)
+		// We use the poll returned reconcile.Result for RequeueAfter if needed
+		// res is the func returned result, so it must not be block-scoped
+		res, err = r.pollImageStreamDigest(ctx, dataImportCron)
```
akalenyu, line 401: My concern is that this is so delicate we may step on it again in the future,
maybe we should decide whether requeue happens more explicitly?

BTW the reasoning for this was to get a uniform way of assigning errors in this module (= to :=) IIRC
instead of mixing them up

akalenyu (commented):
Not insisting just a thought

pkg/controller/dataimportcron-controller.go
arnongilboa: Rewrited the code a bit. I think it's enough.

pkg/controller/dataimportcron-controller.go
```diff
@@ -394,13 +394,14 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		return res, err
 	}
 
-	// We use the poller returned reconcile.Result for RequeueAfter if needed
-	// skip if we disabled schedule
+	// Skip if schedule is disabled
 	if isImageStreamSource(dataImportCron) && dataImportCron.Spec.Schedule != "" {
-		res, err := r.pollImageStreamDigest(ctx, dataImportCron)
+		// We use the poll returned reconcile.Result for RequeueAfter if needed
+		pollRes, err := r.pollImageStreamDigest(ctx, dataImportCron)
```
awels, line 400: Would it be simpler to do
```go
res, err = r.pollImageStreamDigest(ctx, dataImportCron)
if err != nil {
  return res, err
}
```
Note the subtle change from `:=` to `=` which essentially assigns res to the result of pollImageStreamDigest. Which is logically the same as all the code here.

akalenyu: This was the original solution but my concern was that we'd trip over this again
https://github.com/kubevirt/containerized-data-importer/pull/2823#discussion_r1276461398

awels: Apparently this was the original solution, but not explicit enough with regards to how res and err are assigned. I have no issues with either approach honestly.

awels (approved):

PR #2825, opened 2023-07-28
Fix bug: apply DV's label to PVC.

a180285:
Fix bug: apply DV's label to PVC.

Fixes #

**Release note**:

```release-note
Fix bug: apply DV's label to PVC.
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -1009,7 +1009,7 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume, 
 	if util.ResolveVolumeMode(targetPvcSpec.VolumeMode) == corev1.PersistentVolumeFilesystem {
 		labels[common.KubePersistentVolumeFillingUpSuppressLabelKey] = common.KubePersistentVolumeFillingUpSuppressLabelValue
 	}
-	for k, v := range dataVolume.Labels {
+	for k, v := range dataVolume.ObjectMeta.Labels {
 		labels[k] = v
 	}
 


alromeros:
Hey @a180285, thanks for opening this PR. Were you having any unexpected consequences with the old behavior? At first glance it should be working in the same way.

a180285:
Hey, @alromeros , yes, when I am try to add labels on DV. And expect the label will be copied to PVC. But actaully not.

Here is my sample DV.yaml. The annotation is copied. But label not nopied. My cdi version is: `v1.56.0`
```yaml
kind: DataVolume
metadata:
  annotations:
    my.cloud/DefaultUsername: root
  labels:
    my.cloud/PvcType: rbdImage
  name: rbd-img-test
  namespace: user
spec:
  pvc:
    accessModes:
      - ReadWriteMany
    resources:
      requests:
        storage: 40Gi
    storageClassName: ecs-rbd-sc
    volumeMode: Block
  source:
    http:
      url: https://abc.com/a.raw.gz
```

alromeros:
That's odd, I just tested the old code and it works (for me). We even have some tests covering this behavior https://github.com/kubevirt/containerized-data-importer/pull/2646, so they would be failing if the old mechanism was buggy.

That said, I just checked our v1.56 and the code doesn't contain @ido106's fix. Since https://github.com/kubevirt/containerized-data-importer/pull/2646 was merged weeks after our latest v1.56 build I think we are missing a release. @awels wdyt?

a180285:
@alromeros , maybe you are right. I didn't compare code before and after change.
I just compared changed code in this PR. And images of v1.56.0 on quay.io.

Because I come to change from branch release-v1.56. I think it the branch release-v1.56 is equal to v1.56.0 on quay.io.

a180285:
I have just tested, this already fixed.
Image v1.56.0 on quay.io is not work.
But branch release-v1.56 works.

So this PR could be closed, I think.

alromeros:
Happy to hear that @a180285, you are probably right and we need to do a release that includes https://github.com/kubevirt/containerized-data-importer/pull/2646. Again, thanks for the PR!

Issue #2827, opened 2023-07-29
[Doc Broken] API Reference Types Definition 


a180285:
**Description**:
A clear and concise description of what the issue is.

The API page is broken: https://kubevirt.io/cdi-api-reference/main/definitions.html

**What you expected**:
A clear and concise description of what you expected.

**URL**:
Provide a link to the documentation.

https://kubevirt.io/cdi-api-reference/main/definitions.html

**Additional context**:
Add any other context about the issue here.

It broken after resource.Quantity

![image](https://github.com/kubevirt/containerized-data-importer/assets/8207732/2f0e9a45-7158-4154-ba99-cb2bd3e3d168)

![image](https://github.com/kubevirt/containerized-data-importer/assets/8207732/1fa72727-8e41-462c-9e09-83c0f0321e74)

alromeros:
Hey @a180285, thanks for the issue! Yeah, that doesn't seem right. I see the master branch of the repository seems ok https://github.com/kubevirt/api-reference/blob/3dcfdcc0084307ccbcc7b30978c86fd369138cb2/main/definitions.html#L2483, but since most of this code is automatically built the error might come from the API spec in Kubevirt. I'll take a look.

alromeros:
Okay so this behavior seems to be caused by some formatting included in the description of the Quantity API in k8s (https://github.com/kubernetes/kubernetes/blob/41da26dbe15207cbe5b6c36b48a31d2cd3344123/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L37-L57). Since these docs are automatically generated, the description is all condensed in a single line and lacks the `\n` needed to close the formatting.

I can open a PR in kubevirt and fix it, but probably will be updated to the current state every time we run `make deps-generate`. I'll let you know if I'm able to fix it.

alromeros:
Hey @rmohr, pingin you since I see you regularly commit to the api-reference repository. Do you know if we can disable the clode-block formatting from our api-reference? I see [k8s is having a similar problem with formatting](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#quantity-resource-core) but their page doesn't render the code block, so at least the rest of their documentation isn't messed up. Thanks!

rmohr:
@alromeros can't you just perform some modifications of the generated spec in the repo? Like some regexp modifications? The api-reference is automatically built and  pushed. I think it just used my bot account a long time ago.

[...]

PR #2834, opened 2023-08-02
Skip code-block formatting from api-reference

alromeros:

Our api-reference broke due to `quantity`'s description containing code-block formatting that's not rendered properly: https://kubevirt.io/cdi-api-reference/main/definitions.html.

Kubevirt [skips code-block formatting ](https://github.com/kubevirt/kubevirt/blob/main/hack/gen-swagger-doc/gen-swagger-docs.sh#L72)to avoid this behavior, so we should do the same.

Fixes # https://github.com/kubevirt/containerized-data-importer/issues/2827

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/gen-swagger-doc/gen-swagger-docs.sh
+++ b/hack/gen-swagger-doc/gen-swagger-docs.sh
@@ -66,6 +66,8 @@ ${LINK_OPERATIONS}
 __END__
 
 if [ "$OUTPUT_FORMAT" = "html" ]; then
+    # Skipping code-block formatting so it doesn't break the api-reference
+    sed -i 's|```||g' "$WORKDIR/definitions.adoc"
     # $$ has special meaning in asciidoc, we need to escape it
     sed -i 's|\$\$|+++$$+++|g' "$WORKDIR/definitions.adoc"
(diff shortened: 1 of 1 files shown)


aglitke (approved):
/lgtm

awels:
the failure looks like a race condition on when the `source-pod` is deleted. Which it should not be doing since we annotated the DV with the `retainAfterCompletion` annotation. So there might be a bug in that for the source pod.

alromeros:
> the failure looks like a race condition on when the `source-pod` is deleted. Which it should not be doing since we annotated the DV with the `retainAfterCompletion` annotation. So there might be a bug in that for the source pod.
> 
> More investigation, the target PVC has the annotation, but I see no evidence of the source pod ever getting created. The clone is not complete. The source pod name was generated, and visible on the target PVC, but no logs or events about that name at all.
> 
> Failure happened in the legacy hpp lane, no CSI and no populators.

I don't remember which test failed, do you know if the error can be replicated consistently? Any other PRs where it happened?

awels:
https://github.com/kubevirt/containerized-data-importer/blob/main/tests/datavolume_test.go#L1687-L1744 is the test that failed, in particular it failed on line 1743 in that it could not find the source pod to read some information from.

PR #2860, opened 2023-08-22
Do not fail DV reconcile if storage class is not found

arnongilboa:

Currently there is a repeating reconcile error due to exponential backoff: `DataVolume.storage spec is missing accessMode and no storageClass to choose profile`, however the error is not needed as we already have `StorageClass` watch for this case, e.g. when the `DataVolume` is using the default `StorageClass` but no default `StorageClass` is configured yet.

Fixes #

**Release note**:
```release-note
Remove DataVolume reconcile error when storage class is not found yet
```

### DIFF
--- a/pkg/controller/datavolume/import-controller_test.go
+++ b/pkg/controller/datavolume/import-controller_test.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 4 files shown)


akalenyu:
/hold
could we add a small check in test_id:8383 that we're not accumulating exponential backoff due to this?

arnongilboa:
> /hold could we add a small check in test_id:8383 that we're not accumulating exponential backoff in this scenario?

@akalenyu I think a utest verifying reconcile does not return error in this case is wiser here. wdyt?

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -85,6 +86,23 @@ var _ = Describe("All DataVolume Tests", func() {
...
 
+		DescribeTable("Should return nil when storage spec has no AccessModes and", func(scName *string) {
+			importDataVolume := newImportDataVolumeWithPvc("test-dv", nil)
+			importDataVolume.Spec.Storage = &cdiv1.StorageSpec{
+				StorageClassName: scName,
+			}
+
+			reconciler = createImportReconciler(importDataVolume)
+			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}})
+			Expect(err).ToNot(HaveOccurred())
```
akalenyu, line 97: Maybe also
`Expect(result).To(Equal(reconcile.Result{}))`
To ensure requeue/requeueafter don't get set?

arnongilboa: done

PR #2862, opened 2023-08-22
increase controller concurrency and cpu requests

mhenriks:

This commit ups the cpu request for for all our installed compopents (cdi-deployment, cdi-apiserver, cdi-uploadproxy, cdi-operator) for 10m (1% of a core) to 100m (10% of a core).

Also make all controllers handle 3 concurrent requests.

Without this change, it is pretty easy to create a large number of concurrent clone operations and get token timeout errors. Upping resource requests and concurrency addresses the issue in a very direct way.

I experimented with a bunch of other approaches including creating a controller just for refreshing tokens but this is the lowest touch (codewise) and most highly effective solution.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2216038

**Release note**:

```release-note
Increase deployment cpu requests to 100m.  Configure controllers to handle concurrent requests.
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -495,10 +495,17 @@ func (r *ReconcilerBase) syncUpdate(log logr.Logger, syncState *dvSyncState) err
 		return fmt.Errorf("status update is not allowed in sync phase")
 	}
 	if !reflect.DeepEqual(syncState.dv.ObjectMeta, syncState.dvMutated.ObjectMeta) {
+		_, ok := syncState.dv.Annotations[cc.AnnExtendedCloneToken]
+		_, ok2 := syncState.dvMutated.Annotations[cc.AnnExtendedCloneToken]
 		if err := r.updateDataVolume(syncState.dvMutated); err != nil {
 			r.log.Error(err, "Unable to sync update dv meta", "name", syncState.dvMutated.Name)
 			return err
 		}
+		if !ok && ok2 {
+			delta := time.Since(syncState.dv.ObjectMeta.CreationTimestamp.Time)
+			log.V(3).Info("Adding extended DataVolume token took", "delta", delta)
+		}
+		syncState.dv = syncState.dvMutated.DeepCopy()
 	}
 	return nil
 }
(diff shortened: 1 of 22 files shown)


pkg/controller/datavolume/controller-base.go
```diff
@@ -495,10 +495,17 @@ func (r *ReconcilerBase) syncUpdate(log logr.Logger, syncState *dvSyncState) err
...
+		_, ok2 := syncState.dvMutated.Annotations[cc.AnnExtendedCloneToken]
 		if err := r.updateDataVolume(syncState.dvMutated); err != nil {
 			r.log.Error(err, "Unable to sync update dv meta", "name", syncState.dvMutated.Name)
 			return err
 		}
+		if !ok && ok2 {
+			delta := time.Since(syncState.dv.ObjectMeta.CreationTimestamp.Time)
+			log.V(3).Info("Adding extended DataVolume token took", "delta", delta)
+		}
+		syncState.dv = syncState.dvMutated.DeepCopy()
```
alromeros, line 508: Everything looks good but I'm wondering about this section, does this introduce new behavior? Should we test it?

mhenriks: In my testing this appeared to reduce updateStatus failures/retries.  The point is that once the dv has been updated the "baseline" and mutated DV are the same

akalenyu: weird, syncUpdate is usually the last thing we do in the loop, how come something else runs updates after it

mhenriks: Good point!  But the fact is that the "fresh copy" from the cache is often stale (informer didn't receive update yet).  Would be better if updatestatus actually uses the copy returned by update in that case.  I swear that adding this assignment was reducing errors/retries.  Seems unlikely but maybe the additional time to do the assignment gives more time for the informer to update.

Anyway, worst case this assignment has no actual effect.  I can remove it if you'd like but it does seem more correct to me that "dv" and "dvMutated" should be equal at that point.  The fact that neither value is referenced again is kind of beside the point.  That may change in the future

akalenyu: > That may change in the future

That is my concern; if we ever have to look at dv/dvmutated changes to compute some status transition,
this may throw us off

mhenriks: I don't think we should spend a lot of time wondering about hypotheticals.  But what are you suggesting we do here @akalenyu?

akalenyu: Yeah I guess that's not very likely, I am fine with keeping it

PR #2864, opened 2023-08-23
Ensure priority class is assigned to pod populating pvc prime

awels:
Priority class was not being passed to the pvc prime from the PVC and thus was not being added to the importer pod populating the pvc prime. There is a list of allowed annotations that can be passed and the priority class annotation was not in it. This commit adds the annotation to the allowed list.

Cleaned up unneeded log argument to a function related to passing the annotations.

Fixes [RHBZ: 2231839](https://bugzilla.redhat.com/show_bug.cgi?id=2231839)

**Release note**:

```release-note
BugFix: Fix not passing priority class to populator pod
```

### DIFF
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -247,7 +247,7 @@ func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.
 
 // reconcile functions
 
-func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorController, log logr.Logger) (reconcile.Result, error) {
+func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorController, pvcNameLogger logr.Logger) (reconcile.Result, error) {
 	// Get the target PVC
 	pvc := &corev1.PersistentVolumeClaim{}
 	if err := r.client.Get(context.TODO(), req.NamespacedName, pvc); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/controller/populators/populator-base.go
```diff
@@ -259,7 +259,7 @@ func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorCon
 
 	// We first perform the common reconcile steps.
 	// We should only continue if we get a valid PVC'
-	pvcPrime, err := r.reconcileCommon(pvc, populator, log)
```
alromeros, line 262: Should we remove the log passed to reconcile too? I don't remember if I added this for some reason or just by mistake.

awels: Lets see if we can make it clearer this is the case otherwise I will do this again. I was like why are we passing this log, the struct has a logger.

tests/datavolume_test.go
```diff
@@ -3006,7 +3006,6 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			dataVolume := utils.NewDataVolumeWithHTTPImport(dataVolumeName, "1Gi", fmt.Sprintf(utils.TinyCoreQcow2URLRateLimit, f.CdiInstallNs))
 			By(fmt.Sprintf("creating new datavolume %s with priority class", dataVolume.Name))
 			dataVolume.Spec.PriorityClassName = "system-cluster-critical"
-			dataVolume.Annotations[controller.AnnPodRetainAfterCompletion] = "true"
```
alromeros, line 3009: Maybe remove the annotation from the next two (upload and clone) tests? Other than that looks good to me.

awels: So the cloner is problematic, as that pod disappears too quickly and the test fails..

awels:
/hold
Adding some unit tests for the upload and clone cases as well.

PR #2866, opened 2023-08-24
Audit RBAC, avoid global (*) permissions

akalenyu:
There are some permissions that are logically not needed, and some others where we can just reduce the verb set.
Trying to follow https://kubernetes.io/docs/concepts/security/rbac-good-practices/

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=2183659

**Release note**:

```release-note
BugFix: Global permissions [*] seen across CDI components
```

### DIFF
--- a/pkg/operator/resources/operator/operator.go
+++ b/pkg/operator/resources/operator/operator.go
@@ -33,7 +33,8 @@ import (
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 
 	"kubevirt.io/containerized-data-importer/pkg/operator/resources"
-	cluster "kubevirt.io/containerized-data-importer/pkg/operator/resources/cluster"
+	cdicluster "kubevirt.io/containerized-data-importer/pkg/operator/resources/cluster"
+	cdinamespaced "kubevirt.io/containerized-data-importer/pkg/operator/resources/namespaced"
 	utils "kubevirt.io/containerized-data-importer/pkg/operator/resources/utils"
 )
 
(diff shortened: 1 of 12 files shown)


pkg/operator/resources/operator/operator.go
```diff
@@ -54,7 +54,12 @@ func getClusterPolicyRules() []rbacv1.PolicyRule {
 				"clusterroles",
 			},
 			Verbs: []string{
-				"*",
```
akalenyu, line 57: @mhenriks Apparently we have been using `escalate` for a while, which means the operator
can create clusterroles with more rights than it has
https://kubernetes.io/docs/concepts/security/rbac-good-practices/#escalate-verb

pkg/operator/resources/operator/operator.go
```diff
@@ -74,41 +80,45 @@ func getClusterPolicyRules() []rbacv1.PolicyRule {
...
 		{
 			APIGroups: []string{
-				"apiextensions.k8s.io",
+				"cdi.kubevirt.io",
+				"upload.cdi.kubevirt.io",
 			},
 			Resources: []string{
-				"customresourcedefinitions",
-				"customresourcedefinitions/status",
+				"*",
```
Barakmor1, line 104: nit:
why *

akalenyu: All CDI resources (datavolumes/datasources/..)

Barakmor1: As far as i know the operator doesn't create or delete datavolumes/datasources/..

Barakmor1: Oh sorry i missed that :+1:

pkg/operator/resources/operator/operator.go
mhenriks: Yikes!

awels (approved):

PR #2873, opened 2023-08-28
Allow the usage of AnnPodRetainAfterCompletion with populators

alromeros:

When introducing populators we somehow assumed that having the PVC' in Lost phase would interfere with the behavior of `AnnPodRetainAfterCompletion`. After better testing, the pod state and logs are still fetchable after completion.

This PR allows the usage of this annotation.

Fixes #

**Special notes for your reviewer**: Let's see how tests behave.

**Release note**:

```release-note
Allow the usage of AnnPodRetainAfterCompletion with populators
```

### DIFF
--- a/pkg/controller/populators/clone-populator.go
+++ b/pkg/controller/populators/clone-populator.go
@@ -301,9 +301,15 @@ func (r *ClonePopulatorReconciler) validateCrossNamespace(pvc *corev1.Persistent
 }
 
 func (r *ClonePopulatorReconciler) reconcileDone(ctx context.Context, log logr.Logger, pvc *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
-	log.V(3).Info("executing cleanup")
-	if err := r.planner.Cleanup(ctx, log, pvc); err != nil {
-		return reconcile.Result{}, err
+	if shouldSkipCleanup(pvc) {
+		log.V(3).Info("skipping cleanup")
+		// Avoiding cleanup so we can keep clone objects for debugging purposes.
+		r.recorder.Eventf(pvc, corev1.EventTypeWarning, retainedPVCPrime, messageRetainedPVCPrime)
+	} else {
+		log.V(3).Info("executing cleanup")
+		if err := r.planner.Cleanup(ctx, log, pvc); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	log.V(1).Info("removing finalizer")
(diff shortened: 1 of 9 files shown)


pkg/controller/populators/clone-populator.go
```diff
@@ -302,8 +302,13 @@ func (r *ClonePopulatorReconciler) validateCrossNamespace(pvc *corev1.Persistent
 
 func (r *ClonePopulatorReconciler) reconcileDone(ctx context.Context, log logr.Logger, pvc *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
 	log.V(3).Info("executing cleanup")
-	if err := r.planner.Cleanup(ctx, log, pvc); err != nil {
-		return reconcile.Result{}, err
+	if pvc.Annotations[cc.AnnPodRetainAfterCompletion] == "true" {
+		// Avoiding cleanup so we can keep clone objects for debugging purposes.
+		r.recorder.Eventf(pvc, corev1.EventTypeWarning, retainedPVCPrime, messageRetainedPVCPrime)
+	} else {
+		if err := r.planner.Cleanup(ctx, log, pvc); err != nil {
```
alromeros, line 309: Hey @mhenriks, what do you think of skipping the cleanup when using `AnnPodRetainAfterCompletion`? I think that for debugging purposes we can allow keeping all the clone objects (even the PVC prime in lost state).

Another possibility is to have the PVC Prime in terminating state until the user deletes the pod, but having the PVC stuck there seems awkward to me. Maybe we could also consider removing the PVC protection finalizer to delete the PVC prime.

alromeros:
@mhenriks This could be also an opportunity to add the importer pod name annotation to the target PVC. I personally liked the idea and find it better than expecting users to build the pod name with the UID.

pkg/controller/populators/clone-populator.go
mhenriks: Only do this for host clone

mhenriks: One thing to note here though.  Pods will be in the source namespace.  With cross namespace clone, the user may not have access to them.  And I'm not sure we want to allow the user to do this in cross namespace case

alromeros: Yeah that's a good point. What about the check in the DV controller to see if we use populators or not? 

https://github.com/kubevirt/containerized-data-importer/blob/bd6335259dd12ba518634b2366be5c664567ba40/pkg/controller/datavolume/controller-base.go#L1180-L1184

Are we ok with turning to populators even if we only use `AnnPodRetainAfterCompletion` with host assisted clone?

alromeros: I guess we are since it's basically the same behavior without populators.

mhenriks: Without populators, the target pod is in the target namespace.  With populators both pods are in the source then rebind happens.  The reason for having the target pod in the source is to better support k8s native cross namespace auth (when it goes live).  In that case there will be no clone token.  So can't do cross namespace host clone

pkg/controller/populators/clone-populator.go
```diff
@@ -472,3 +478,18 @@ func getSavedCloneStrategy(obj client.Object) *cdiv1.CDICloneStrategy {
...
 	cc.AddAnnotation(obj, cc.AnnCloneType, string(strategy))
 }
+
+func shouldSkipCleanup(pvc *corev1.PersistentVolumeClaim) bool {
+	// We can skip cleanup to keep objects for debugging purposes if:
+	//   - AnnPodRetainAfterCompletion annotation is set to true. This means that the user explicitly wants
+	//     to keep the pods.
+	//   - Clone is host-assisted, which is the only clone type with transfer pods worth debugging.
+	//   - Clone occurs in a single namespace, so we avoid retaining pods in a namespace we don't have right to access.
+	dataSourceNamespace, ok := pvc.Annotations[AnnDataSourceNamespace]
```
mhenriks, line 488: Let's future proof this for when cross namespace datasource goes live.  In that case `pvc.spec.dataCource.Namespace` may be set

alromeros:
Backporting this as requested in Jira
/cherrypick release-v1.57

PR #2879, opened 2023-08-29
Fix gingko install in builder

awels:
Created a broken builder by using the wrong path for ginkgo since we are using the v2 version now.
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -40,7 +40,7 @@ RUN mkdir -p /gimme && curl -sL https://raw.githubusercontent.com/travis-ci/gimm
 RUN \
 	source /etc/profile.d/gimme.sh && \
 	eval $(go env) && \
-	go install github.com/onsi/ginkgo/ginkgo@v2.12.0 && \
+	go install github.com/onsi/ginkgo/v2@v2.12.0 && \
 	go install golang.org/x/tools/cmd/goimports@latest && \
 	go install mvdan.cc/sh/cmd/shfmt@latest && \
 	go install github.com/mattn/goveralls@latest && \


hack/build/docker/builder/Dockerfile
```diff
@@ -40,7 +40,7 @@ RUN mkdir -p /gimme && curl -sL https://raw.githubusercontent.com/travis-ci/gimm
 RUN \
 	source /etc/profile.d/gimme.sh && \
 	eval $(go env) && \
-	go install github.com/onsi/ginkgo/ginkgo@v2.12.0 && \
+	go install github.com/onsi/ginkgo/v2@v2.12.0 && \
```
akalenyu, line 43: I think it's better to pin this to the vendored ginkgo version
https://github.com/kubevirt/containerized-data-importer/blob/973096abdda78ee26494be6b4d80699db15fe641/go.mod#L24
https://onsi.github.io/ginkgo/#installing-ginkgo
(`you must make sure the version of the ginkgo cli you install is the same as the version of Ginkgo in your go.mod file.`)

awels: Well this is the builder. That needs to get updated first, there will be a follow up PR where we start using the builder, and in that PR I will make sure the ginkgo version in go.mod matches the one in the builder.

awels: Note our current builder is on v1 of ginkgo, so doesn't match the go.mod at all.

awels: So apparently that still wasn't the right path even though it worked for me locally. Gotta do one more PR to really fix it... /facepalm

PR #2881, opened 2023-08-30
HTTP import with DIC

ido106:

Allow using the CDI DataImportCron framework (auto import to golden image namespace, DataSource update, garbage collection of old PVCs etc.) when the user wants to use HTTP import instead of registry import. It requires the user to either implement a customized poller (or manually) updating the DataImportCron desired digest annotation, and triggering new import.

Fixes #

**Release note**:

```release-note
Allow using the CDI DataImportCron framework with HTTP imports.
```

### DIFF
--- a/pkg/apiserver/webhooks/dataimportcron-validate.go
+++ b/pkg/apiserver/webhooks/dataimportcron-validate.go
@@ -90,10 +90,28 @@ func (wh *dataImportCronValidatingWebhook) Admit(ar admissionv1.AdmissionReview)
 func (wh *dataImportCronValidatingWebhook) validateDataImportCronSpec(request *admissionv1.AdmissionRequest, field *k8sfield.Path, spec *cdiv1.DataImportCronSpec, namespace *string) []metav1.StatusCause {
 	var causes []metav1.StatusCause
 
-	if spec.Template.Spec.Source == nil || spec.Template.Spec.Source.Registry == nil {
+	if spec.Template.Spec.Source == nil || (spec.Template.Spec.Source.Registry == nil && spec.Template.Spec.Source.HTTP == nil) {
 		causes = append(causes, metav1.StatusCause{
 			Type:    metav1.CauseTypeFieldValueInvalid,
-			Message: "Missing registry source",
+			Message: "Missing registry or HTTP source",
+			Field:   field.Child("Template").String(),
+		})
+		return causes
+	}
+
+	if spec.Template.Spec.Source.Registry != nil && spec.Template.Spec.Source.HTTP != nil {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "Only one source type can be selected: Registry or HTTP",
+			Field:   field.Child("Template").String(),
+		})
+		return causes
+	}
+
+	if spec.Template.Spec.Source.HTTP != nil && spec.Schedule != "" {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "Schedule must be empty when using HTTP import",
 			Field:   field.Child("Template").String(),
 		})
 		return causes
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"reflect"
 	"sort"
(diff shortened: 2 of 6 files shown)


ido106:
> /ok-to-test

Actually I still have to change and add some stuff, so we can wait with the tests in the meantime. 
I'm writing a design to it right now so we can discuss it before I continue.

pkg/apiserver/webhooks/dataimportcron-validate.go
```diff
@@ -90,10 +90,28 @@ func (wh *dataImportCronValidatingWebhook) Admit(ar admissionv1.AdmissionReview)
...
+			Message: "Only one source type can be selected: Registry or HTTP",
+			Field:   field.Child("Template").String(),
+		})
+		return causes
+	}
+
+	if spec.Template.Spec.Source.HTTP != nil && spec.Schedule != "" {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "Schedule must be empty when using HTTP import",
```
arnongilboa, line 114: "Schedule must be empty when using HTTP import"?

pkg/controller/dataimportcron-controller.go
```diff
@@ -117,6 +109,14 @@ const (
...
 	cronJobUIDSuffixLength      = 8
 	defaultImportsToKeepPerCron = 3
+
+	// ErrDataSourceAlreadyManaged provides a const to indicate DataSource already managed error
+	ErrDataSourceAlreadyManaged = "ErrDataSourceAlreadyManaged"
+	// MessageDataSourceAlreadyManaged provides a const to form DataSource already managed error message
+	MessageDataSourceAlreadyManaged = "DataSource %s is already managed by DataImportCron %s"
+
+	prometheusNsLabel       = "ns"
+	prometheusCronNameLabel = "cron_name"
```
arnongilboa, line 119: Why moved from const to var?

pkg/controller/dataimportcron-controller.go
```diff
@@ -178,7 +178,7 @@ func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron
 		}
 		return nil
 	}
-	if !isURLSource(dataImportCron) {
+	if !isURLSource(dataImportCron) && !isHTTPSource(dataImportCron) {
```
arnongilboa, line 181: Names are not clear as they smell the same - Maybe isRegistryURLSource?

pkg/controller/dataimportcron-controller.go
```diff
@@ -259,16 +259,22 @@ func splitImageStreamName(imageStreamName string) (string, string, error) {
 	return "", "", errors.Errorf("Illegal ImageStream name %s", imageStreamName)
 }
 
-func (r *DataImportCronReconciler) pollImageStreamDigest(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
+func (r *DataImportCronReconciler) pollDigest(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
```
arnongilboa, line 262: Why renamed? it is for image streams and the diff is commented out.

pkg/controller/dataimportcron-controller.go
```diff
@@ -1172,6 +1249,94 @@ func InitPollerPodSpec(c client.Client, cron *cdiv1.DataImportCron, podSpec *cor
 		)
 	}
 
+	err = createPodSpec(container, insecureTLS, *cdiConfig, c, podSpec, volumes)
+
+	return err
+}
+
+func initPollerPodSpecHTTP(c client.Client, cron *cdiv1.DataImportCron, podSpec *corev1.PodSpec, image string, pullPolicy corev1.PullPolicy, log logr.Logger) error {
```
arnongilboa, line 1257: No idea what you tried to do here but cloning the poller won't work.

pkg/apiserver/webhooks/dataimportcron-validate.go
ido106: I added it now because that's what you wrote in the Jira card

pkg/controller/dataimportcron-controller.go
ido106: I saw 2 const blocks so I merged them.
I saw that it appears here as a var too but for me locally it is as a const block. weird

pkg/controller/dataimportcron-controller.go
ido106: > Names are not clear as they smell the same - Maybe isRegistryURLSource?

Sure.

pkg/controller/dataimportcron-controller.go
ido106: I tried to implement it differently to the HTTP import as we initially wanted with the timestamp of the last successful if-modified-since request, but then I changed it to digest as the URL suffix as suggested in the Jira card so I moved those changes to a comment.
Don't know if to keep them

pkg/controller/dataimportcron-controller.go
ido106: What exactly won't work? I haven't added complex logic to the poller yet, I just adjusted the code for each type of import.
When using Registry I need `getCronRegistrySource` and with HTTP I need `getCronHTTPSource`, in the Registry the URL is of type `*string` and in HTTP it is of type `string`. So for now I kept the code that was the same and moved the different code to another function. This will probably change later.
If I leave the original code I get an error in the unit test: `Cron with no registry source test-cron.`
I tried going through it again but couldn't find anything else at the moment so I'll try again when I get back, maybe with more information/reviews

pkg/apiserver/webhooks/dataimportcron-validate.go
arnongilboa: Yep, I think it was mentioned because we can't use the current poller, so until we provide a polling solution the schedule is useless.

pkg/controller/dataimportcron-controller.go
arnongilboa: All you do here is using the existing poller pod, which definately won't work.

PR #2882, opened 2023-08-31
Allow modifying log verbosity in CDI

alromeros:

This PR aims to make the debugging process of CDI easier, allowing changes in log verbosity from the CDI config.

It adds a new field in the `CDIConfig` API to allow specifying a different log verbosity level to re-initialize all loggers. Changes in this field will restart all CDI components to initialize all loggers with the new value.

The new API looks like this:

```yaml
Kind: CDI
# ...
spec:
  config:
      logVerbosity: 4
```

I opted for a simpler API compared to [the one used in kubevirt](https://github.com/kubevirt/kubevirt/pull/4672/files#diff-5469726ef6ef072d9258db4593eb8811363099d9181a4e5dbd119af7b2b1414eR1614). I think the kubevirt API is overkill for CDI, but I can always change it if preferred.

**Release note**:

```release-note
Allow modifying log verbosity in CDI
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -4935,6 +4935,11 @@
        "default": ""
       }
      },
+     "logVerbosity": {
+      "description": "LogVerbosity overrides the default verbosity level used to initialize loggers",
+      "type": "integer",
+      "format": "int32"
+     },
      "podResourceRequirements": {
       "description": "ResourceRequirements describes the compute resource requirements.",
       "$ref": "#/definitions/v1.ResourceRequirements"
--- a/doc/debug.md
+++ b/doc/debug.md
@@ -28,3 +28,21 @@ spec:
       requests:
         storage: 1Gi
 ```
+
+## Log verbosity
+
+Different levels of verbosity are used in CDI to control the amount of information that is logged. The verbosity level of logs in CDI can be adjusted using a dedicated field in CDIConfig. This feature enables users to control the amount of detail displayed in logs, ranging from minimal to detailed debugging information.
+
+For example:
+
+```yaml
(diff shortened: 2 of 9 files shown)


alromeros:
@akalenyu wondering if it'd be useful to have individual verbosity levels for all components as they have in kubevirt, wdyt? At first glance seems overkill for CDI.

akalenyu:
> @akalenyu wondering if it'd be useful to have individual verbosity levels for all components as they have in kubevirt, wdyt? At first glance seems overkill for CDI.

Yeah, I think we can start off just with the global verbosity value, just have to make sure we're
fine for additions in the future.. @mhenriks WDYT? would a struct with .globalVerbosity be better for that?

alromeros:
> > @akalenyu wondering if it'd be useful to have individual verbosity levels for all components as they have in kubevirt, wdyt? At first glance seems overkill for CDI.
> 
> Yeah, I think we can start off just with the global verbosity value, just have to make sure we're fine for additions in the future.. @mhenriks WDYT? would a struct with .globalVerbosity be better for that?

Yeah I like the idea of having a struct so it's more maintainable, Let's see what others think.

mhenriks:
I think this is fine for infra components but what about worker pods?

awels:
The workloads get their verbosity from the controllers that create them. So the import controller creates pods and grabs the verbosity from the cdi-deployment.

alromeros:
@mhenriks @awels are we ok with having a single verbosity value? Or do you prefer to use a struct as proposed by @akalenyu so we can add more fields in the future?
/hold

awels:
I am fine with a single verbosity value. I am not sure there is much value in a more granular approach.

PR #2895, opened 2023-09-11
WIP forklift: add forklift volume populators images

bennyz:

This is preliminary work to build and include the populators from forklift[1], for ovirt and for openstack.
To do this, this PR first updates the ovirt-imageio dependencies, as the current version does not include the ovirt-img tool used by the ovirt volume populator.

Then, the bazel build files are created in order to build the images. This includes, generating rpm dependencies for ovirt, pycurl and six which are required by ovirt-img.
The openstack volume populator uses the gophercloud[2] Go SDK to interact with openstack, so this go dependency is also brought in.

Lastly, build scripts and relevant operator code was modified to include the new images.

Fixes #

There is no aarch64 build for ovirt-engine-sdk so this is still an outstanding issue

**Release note**:

```release-note

```

[1] https://github.com/kubev2v/forklift
[2] https://github.com/gophercloud/gophercloud


BUILD.bazel
```diff
@@ -168,6 +168,24 @@ container_push(
...
+)
+
+container_push(
+    name = "push-openstack-populator",
+    format = "Docker",
+    image = "//cmd/openstack-populator:openstack-populator-image",
+    registry = "$(container_prefix)",
+    repository = "openstack-populator",
+    tag = "$(container_tag)",
+)
```
mhenriks, line 187: Have you considered adding binaries to the cdi-importer image instead of creating two new images?

mhenriks (commented):
Good start!  Looks like you have chose to create two new images for the populator "worker" pods.  I'm not against that, but just curious about the pros/cons of that approach vs adding code to cdi-importer.  It's not a huge deal, but more images is just more things to manage.  As you've seen, there is a non-trivial amount of code needed to set the image names in yamls/csvs/etc in upstream and downstream configurations.

Can you talk more about next steps and how the work will be split up?

bennyz:
> Good start! Looks like you have chose to create two new images for the populator "worker" pods. I'm not against that, but just curious about the pros/cons of that approach vs adding code to cdi-importer. It's not a huge deal, but more images is just more things to manage. As you've seen, there is a non-trivial amount of code needed to set the image names in yamls/csvs/etc in upstream and downstream configurations.
> 
> Can you talk more about next steps and how the work will be split up?

Thank you for looking into this!

Integrating the populator images code into cdi-importer was brought up in the meeting we had at the time, the argument against it would that it would be best to keep things separate as much as possible, and the other argument is that the ovirt populator has some RPM dependencies[1] and that would "pollute" the cdi-importer image further.
However, given that we can already get ovirt-img with the dependency bump I introduced here, there are only two extra python dependencies to introduce, so if you feel that incorporating the populators into cdi-importer would be better (and I do see a lot of benefits to do that given the amount of code needed to separate the image), I am willing to give it a go

As for the next, we'd need a controller to run the binaries (or binary, if we go with the single binary approach), my thinking was to add an additional forklift-controller in `pkg/controller/populators`, I thought about separating this into a separate package, but populator-base.go has too much useful infra that can be reused so I believe it should be in that package 

[1] https://github.com/kubevirt/containerized-data-importer/blob/31295bac25becb255fc28b4352d6ce4ee4910aa4/hack/build/rpm-deps.sh#L72
(qemu-img should already be there, the other two not)

mhenriks:
> Integrating the populator images code into cdi-importer was brought up in the meeting we had at the time, the argument against it would that it would be best to keep things separate as much as possible, and the other argument is that the ovirt populator has some RPM dependencies[1] and that would "pollute" the cdi-importer image further. However, given that we can already get ovirt-img with the dependency bump I introduced here, there are only two extra python dependencies to introduce, so if you feel that incorporating the populators into cdi-importer would be better (and I do see a lot of benefits to do that given the amount of code needed to separate the image), I am willing to give it a go

I think my vote would be to add to `cdi-importer` rather than create two new images.  what do you think @awels and @akalenyu ?

akalenyu:
> d a controller to run the binaries (or binary, if we go with the single binary approach), my thinking was to add an additional forklift-controller in `pkg/controller/populat`

So `ovirt-img` is already in the importer image today? If that's the case then I think adding to cdi-imporer is fine.
We can compare the resulting image to see if the python libs bloat it

bennyz:
> > d a controller to run the binaries (or binary, if we go with the single binary approach), my thinking was to add an additional forklift-controller in `pkg/controller/populat`
> 
> So `ovirt-img` is already in the importer image today? If that's the case then I think adding to cdi-imporer is fine. We can compare the resulting image to see if the python libs bloat it

not with the current version of `ovirt-imageio`, this PR bumps its version to one that does include it. I will post a PR with a unified image, we'll see how it goes

akalenyu:
> > > d a controller to run the binaries (or binary, if we go with the single binary approach), my thinking was to add an additional forklift-controller in `pkg/controller/populat`
> > 
> > 
> > So `ovirt-img` is already in the importer image today? If that's the case then I think adding to cdi-imporer is fine. We can compare the resulting image to see if the python libs bloat it
> 
> not with the current version of `ovirt-imageio`, this PR bumps its version to one that does include it. I will post a PR with a unified image, we'll see how it goes

I use https://github.com/wagoodman/dive to inspect container images, maybe it could also be helpful in comparison here

bennyz:
@mhenriks @akalenyu 

I started looking into this and I have some things I am not sure about how you'd want to approach this:
https://github.com/kubevirt/containerized-data-importer/blob/20cced8381b4f53337db6247b9317702a38f9510/cmd/cdi-importer/importer.go#L136

Did you mean to tap into this main with a new `source` or something similar or did you have something else in mind with regards to how to implement this?

I'm trying to understand what would be the general flow before I start doing the image related changes

bennyz:
Closing in favor of https://github.com/kubevirt/containerized-data-importer/pull/2947

PR #2896, opened 2023-09-12
Bump controller runtime to avoid mistakenly defaulting to wrong mapper

akalenyu:
Today, controller runtime mistakenly ignores the inherited Manager default
dynamic mapper and uses a discovery mapper instead:
https://github.com/kubernetes-sigs/controller-runtime/pull/2491
This means that if some CRD was not available on the cdi-controller startup,
Even if it got installed after, we would still get IsNoMatch when trying to access it.

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=2237470

**Release note**:

```release-note
BugFix: If a CRD wasn't available during controller startup, our client would keep failing to get the objects,
mistakenly returning NoMatchError even if the CRD was installed since
```


akalenyu:
/hold
let's get thoughts first, maybe we can live with discovery mapper peacefully for a while

alromeros:
Just wondering if this PR would still be necessary if your controller-runtime fix merges. As I understand, your PR https://github.com/kubernetes-sigs/controller-runtime/pull/2491 fixes the issue in controller runtime v0.14, which is used in our older maintained releases. We don't have to worry about future releases since v0.15-0.16 doesn't have this issue and we plan to update soon. Is this PR still necessary, then?

akalenyu:
> Just wondering if this PR would still be necessary if your controller-runtime fix merges. As I understand, your PR [kubernetes-sigs/controller-runtime#2491](https://github.com/kubernetes-sigs/controller-runtime/pull/2491) fixes the issue in controller runtime v0.14, which is used in our older maintained releases. We don't have to worry about future releases since v0.15-0.16 doesn't have this issue and we plan to update soon. Is this PR still necessary, then?

So we have the Bugzilla linked in this PR that triggered this investigation to begin with;
Basically, OpenShift routes were not available for some reason at the controller startup time, and since we use
the "wrong" mapper (non-dynamic one) our client can never get routes anymore until the controller pod is restarted.
There could be other pain points with using a non-dynamic mapper, or, we may be able to live with it
until controller runtime PR merges, a 0.14.z is made, and that is used in CDI.

PR #2898, opened 2023-09-14
Allow StorageProfile to use a specific VolumeSnapshotClass

arnongilboa:
When there are several `VolumeSnapshotClasses` of the same provisioner but with different parameters, we would like to allow `StorageProfile` choose `VolumeSnapshotClass` for its `StorageClass`, so we won't choose the wrong one.

Fixes bz #2219774

**Release note**:
```release-note
Allow StorageProfile to use a specific VolumeSnapshotClass
```

### DIFF
--- a/pkg/controller/clone/common.go
+++ b/pkg/controller/clone/common.go
@@ -145,6 +145,23 @@ func GetStorageClassForClaim(ctx context.Context, c client.Client, pvc *corev1.P
 	return nil, nil
 }
 
+func getSnapshotClassForClaim(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (*string, error) {
+	if pvc.Spec.StorageClassName == nil || *pvc.Spec.StorageClassName == "" {
+		return nil, nil
+	}
+
+	sp := &cdiv1.StorageProfile{}
+	exists, err := getResource(ctx, c, "", *pvc.Spec.StorageClassName, sp)
+	if err != nil {
+		return nil, err
+	}
+	if exists {
+		return sp.Status.SnapshotClass, nil
+	}
+
+	return nil, nil
+}
+
 // GetDriverFromVolume returns the CSI driver name for a PVC
 func GetDriverFromVolume(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (*string, error) {
 	if pvc.Spec.VolumeName == "" {
@@ -208,19 +225,43 @@ func GetCommonDriver(ctx context.Context, c client.Client, pvcs ...*corev1.Persi
 	return result, nil
 }
 
+func getCommonSnapshotClass(ctx context.Context, c client.Client, pvcs ...*corev1.PersistentVolumeClaim) (*string, error) {
+	var result *string
+
+	for _, pvc := range pvcs {
+		sc, err := getSnapshotClassForClaim(ctx, c, pvc)
+		if err != nil {
+			return nil, err
+		}
+		if sc == nil {
+			return nil, nil
+		}
+		if result == nil {
+			result = sc
+		} else if *result != *sc {
+			return nil, nil
+		}
+	}
+
+	return result, nil
+}
+
 // GetCompatibleVolumeSnapshotClass returns a VolumeSnapshotClass name that works for all PVCs
 func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, pvcs ...*corev1.PersistentVolumeClaim) (*string, error) {
 	driver, err := GetCommonDriver(ctx, c, pvcs...)
 	if err != nil {
 		return nil, err
 	}
-
 	if driver == nil {
 		return nil, nil
 	}
 
-	//FIXME: may pass snapshotClassName if common to all pvcs?
-	return cc.GetVolumeSnapshotClass(context.TODO(), c, *driver, nil)
+	snapshotClassName, err := getCommonSnapshotClass(ctx, c, pvcs...)
+	if err != nil {
+		return nil, err
+	}
+
+	return cc.GetVolumeSnapshotClass(context.TODO(), c, *driver, snapshotClassName)
 }
 
 // SameVolumeMode returns true if all pvcs have the same volume mode
--- a/pkg/controller/clone/common.go
+++ b/pkg/controller/clone/common.go
@@ -146,12 +146,13 @@ func GetStorageClassForClaim(ctx context.Context, c client.Client, pvc *corev1.P
 }
 
 func getSnapshotClassForClaim(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (*string, error) {
-	if pvc.Spec.StorageClassName == nil || *pvc.Spec.StorageClassName == "" {
-		return nil, nil
+	sc, err := cc.GetStorageClassByNameWithK8sFallback(ctx, c, pvc.Spec.StorageClassName)
+	if err != nil || sc == nil {
+		return nil, err
 	}
 
 	sp := &cdiv1.StorageProfile{}
-	exists, err := getResource(ctx, c, "", *pvc.Spec.StorageClassName, sp)
+	exists, err := getResource(ctx, c, "", sc.Name, sp)
 	if err != nil {
 		return nil, err
 	}
@@ -247,7 +248,9 @@ func getCommonSnapshotClass(ctx context.Context, c client.Client, pvcs ...*corev
 }
 
 // GetCompatibleVolumeSnapshotClass returns a VolumeSnapshotClass name that works for all PVCs
-func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, log logr.Logger, pvcs ...*corev1.PersistentVolumeClaim) (*string, error) {
+// Note the last PVC passed is considered the target PVC for logs and events
+func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, log logr.Logger, recorder record.EventRecorder, pvcs ...*corev1.PersistentVolumeClaim) (*string, error) {
+	targetClaim := pvcs[len(pvcs)-1]
 	driver, err := GetCommonDriver(ctx, c, pvcs...)
 	if err != nil {
 		return nil, err
@@ -261,7 +264,7 @@ func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, log 
 		return nil, err
 	}
 
-	return cc.GetVolumeSnapshotClass(context.TODO(), c, *driver, snapshotClassName, log)
+	return cc.GetVolumeSnapshotClass(context.TODO(), c, targetClaim, *driver, snapshotClassName, log, recorder)
 }
 
 // SameVolumeMode returns true if all pvcs have the same volume mode
--- a/pkg/controller/clone/common.go
+++ b/pkg/controller/clone/common.go
@@ -247,7 +247,7 @@ func getCommonSnapshotClass(ctx context.Context, c client.Client, pvcs ...*corev
 }
 
 // GetCompatibleVolumeSnapshotClass returns a VolumeSnapshotClass name that works for all PVCs
-func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, pvcs ...*corev1.PersistentVolumeClaim) (*string, error) {
+func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, log logr.Logger, pvcs ...*corev1.PersistentVolumeClaim) (*string, error) {
 	driver, err := GetCommonDriver(ctx, c, pvcs...)
 	if err != nil {
 		return nil, err
@@ -261,7 +261,7 @@ func GetCompatibleVolumeSnapshotClass(ctx context.Context, c client.Client, pvcs
 		return nil, err
 	}
 
-	return cc.GetVolumeSnapshotClass(context.TODO(), c, *driver, snapshotClassName)
+	return cc.GetVolumeSnapshotClass(context.TODO(), c, *driver, snapshotClassName, log)
 }
 
 // SameVolumeMode returns true if all pvcs have the same volume mode
--- a/pkg/controller/clone/common.go
+++ b/pkg/controller/clone/common.go
@@ -3,14 +3,11 @@ package clone
 import (
 	"context"
 	"fmt"
-	"sort"
 
 	"github.com/go-logr/logr"
-	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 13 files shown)


pkg/controller/common/util.go
```diff
@@ -1803,7 +1803,7 @@ func ValidateSnapshotCloneProvisioners(ctx context.Context, c client.Client, sna
 }
 
 // GetSnapshotClassForSmartClone looks up the snapshot class based on the storage class
-func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName *string, log logr.Logger, client client.Client) (string, error) {
+func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName, snapshotClassName *string, log logr.Logger, client client.Client) (string, error) {
```
akalenyu, line 1806: Maybe we should define something more deterministic for the `else {` case?
Right now we're just picking the first match, but maybe we can define something like
- StorageProfile
- [Default](https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/#the-volumesnapshotclass-resource) cluster snap class
- first match in sorted slice

akalenyu, line 1806: Am I missing something, or does this need to change too?
https://github.com/kubevirt/containerized-data-importer/blob/581dfb91c7699a587335ff63fe1120958a776886/pkg/controller/clone/common.go#L214-L215

pkg/controller/storageprofile-controller.go
```diff
@@ -86,10 +86,13 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
 
 	storageProfile.Status.StorageClass = &sc.Name
 	storageProfile.Status.Provisioner = &sc.Provisioner
-	snapClass, err := cc.GetSnapshotClassForSmartClone("", &sc.Name, r.log, r.client)
+	snapClass, err := cc.GetSnapshotClassForSmartClone("", &sc.Name, storageProfile.Spec.SnapshotClass, r.log, r.client)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	if snapClass != "" {
+		storageProfile.Status.SnapshotClass = &snapClass
```
akalenyu, line 94: Just making sure, status will always be populated with the vsc that is going to be used?
So an Admin can only change the spec unless the one that got picked up is wrong?

akalenyu (commented):
PR looks good and makes sense according to the sig storage call,
just some questions

/cc @aglitke for some more eyes on the approach

pkg/controller/common/util.go
arnongilboa: Sure we can add the default cluster snap class (if there is one) instead of the first match.

arnongilboa: Yes it should, and I guess they should be refactored to reuse the same code.

akalenyu: I can see how its more tricky in the populator case, since you probably don't want to start looking up storage profiles here

aglitke: Agree but you can only pick the default SnapClass if the provisioner matches.   So the order is something like:
* StorageProfile
* If there is a single snap class matching the provisioner (easy case): Use single matching snap class
* If there are multiple matches and there is a default snap class with matching provisioner: Default cluster snap class
* If there are multiple matches and no matching default snap class: choose first match in sorted slice

pkg/controller/common/util.go
```diff
@@ -1820,25 +1823,59 @@ func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName *str
...
+			dvName, "snapshot class", *vscName)
+		return *vscName, nil
 	}
 
 	logger.Info("Could not match snapshotter with storage class, falling back to host assisted clone")
 	return "", nil
 }
 
+// GetVolumeSnapshotClass looks up the snapshot class based on the driver and an optional specific name
+func GetVolumeSnapshotClass(ctx context.Context, c client.Client, driver string, snapshotClassName *string) (*string, error) {
```
aglitke, line 1841: Let's make sure we log why we chose a particular snapshot class.  This reminds me of the work we are doing to help people understand why we needed to fall back to host-assisted clone.

aglitke (commented):
Approach looks good.  I made some comments.

pkg/controller/common/util.go
arnongilboa: @aglitke that's exactly what implemented.

pkg/controller/common/util.go
arnongilboa: Added

pkg/controller/common/util.go
akalenyu: I am more concerned about the coupling of storage profiles and our populators
@alromeros Did we ever mean for our set of populators to be used standalone?
Or is this fine to look up storage profiles?

alromeros: > @alromeros Did we ever mean for our set of populators to be used standalone?

Yeah... so it depends on what we mean by standalone usage. We first merged the populators without DV support and ensured it was well-tested and reliable as a standalone feature. However they'll always be dependent on CDI controllers, so don't know if relying on storage profiles would be against this design decision.

alromeros: Yeah, we plan to keep depending on all import, upload, and clone controllers. So I guess that if the main rationale behind populators was always to stop depending on DVs (and we are close to achieving that), there's no problem with resorting to storage profiles since they were never an issue to begin with.

akalenyu (commented):
So apparently I lied and default-snapshot-class is a per-provisioner setting!
I don't think it's clear from k8s [docs](https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/#the-volumesnapshotclass-resource) but I stumbled across the code

https://github.com/kubernetes-csi/external-snapshotter/blob/ff71329d8cc08dca53194f8d4568ed53e69ecb82/pkg/common-controller/snapshot_controller.go#L1362-L1374

I guess this makes the situation solvable without any additions?
@arnongilboa @aglitke

arnongilboa:
> So apparently I lied and default-snapshot-class is a per-provisioner setting! I don't think it's clear from k8s [docs](https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/#the-volumesnapshotclass-resource) but I stumbled across the code
> 
> https://github.com/kubernetes-csi/external-snapshotter/blob/ff71329d8cc08dca53194f8d4568ed53e69ecb82/pkg/common-controller/snapshot_controller.go#L1362-L1374
> 
> I guess this makes the situation solvable without any additions? @arnongilboa @aglitke

@akalenyu  but the whole idea (see the bz) was allowing to use different VolumeSnapshotClasses with the same provisioner, so how does default-per-provisioner help here?

akalenyu:
> > So apparently I lied and default-snapshot-class is a per-provisioner setting! I don't think it's clear from k8s [docs](https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/#the-volumesnapshotclass-resource) but I stumbled across the code
> > https://github.com/kubernetes-csi/external-snapshotter/blob/ff71329d8cc08dca53194f8d4568ed53e69ecb82/pkg/common-controller/snapshot_controller.go#L1362-L1374
> > I guess this makes the situation solvable without any additions? @arnongilboa @aglitke
> 
> @akalenyu but the whole idea (see the bz) was allowing to use different VolumeSnapshotClasses with the same provisioner, so how does default-per-provisioner help here?

We could avoid the storageprofile field and just add respecting for the default snapclass, which they can always change as needed

arnongilboa:
> > > So apparently I lied and default-snapshot-class is a per-provisioner setting! I don't think it's clear from k8s [docs](https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/#the-volumesnapshotclass-resource) but I stumbled across the code
> > > https://github.com/kubernetes-csi/external-snapshotter/blob/ff71329d8cc08dca53194f8d4568ed53e69ecb82/pkg/common-controller/snapshot_controller.go#L1362-L1374
> > > I guess this makes the situation solvable without any additions? @arnongilboa @aglitke
> > 
> > 
> > @akalenyu but the whole idea (see the bz) was allowing to use different VolumeSnapshotClasses with the same provisioner, so how does default-per-provisioner help here?
> 
> Right we would still need changes. But we could avoid the storageprofile field and just add respecting for the default snapclass, which they can always change as needed

but the user wants to use two different VolumeSnapshotClasses with the same provisioner `in parallel`, so patching the default with the desired value per PVC creation seems like a an ugly hack.

akalenyu:
> > > > So apparently I lied and default-snapshot-class is a per-provisioner setting! I don't think it's clear from k8s [docs](https://kubernetes.io/docs/concepts/storage/volume-snapshot-classes/#the-volumesnapshotclass-resource) but I stumbled across the code
> > > > https://github.com/kubernetes-csi/external-snapshotter/blob/ff71329d8cc08dca53194f8d4568ed53e69ecb82/pkg/common-controller/snapshot_controller.go#L1362-L1374
> > > > I guess this makes the situation solvable without any additions? @arnongilboa @aglitke
> > > 
> > > 
> > > @akalenyu but the whole idea (see the bz) was allowing to use different VolumeSnapshotClasses with the same provisioner, so how does default-per-provisioner help here?
> > 
> > 
> > Right we would still need changes. But we could avoid the storageprofile field and just add respecting for the default snapclass, which they can always change as needed
> 
> but the user wants to use two different VolumeSnapshotClasses with the same provisioner `in parallel`, so patching the default with the desired value per PVC creation seems like a an ugly hack.

Yep, in that case we still need a storage profile field, it's too ugly to keep patching them

pkg/controller/common/util.go
```diff
@@ -1820,25 +1823,63 @@ func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName *str
...
+// GetVolumeSnapshotClass looks up the snapshot class based on the driver and an optional specific name
+func GetVolumeSnapshotClass(ctx context.Context, c client.Client, driver string, snapshotClassName *string, log logr.Logger) (*string, error) {
+	logger := log.WithName("GetVolumeSnapshotClass").V(3)
+	if snapshotClassName != nil {
+		vsc := &snapshotv1.VolumeSnapshotClass{}
+		if err := c.Get(context.TODO(), types.NamespacedName{Name: *snapshotClassName}, vsc); err != nil {
+			return nil, err
+		}
+		if vsc.Driver == driver {
+			logger.Info("VolumeSnapshotClass selected according to StorageProfile", "name", vsc.Name)
```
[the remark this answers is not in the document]
arnongilboa: Choosing a specific VolumeSnapshotClass is a simple good-path logic and a rare use case, so I think logging it is more than enough (unlike the fallback to host-assisted clone where we added pvc event and anotation). @aglitke wdyt?

pkg/controller/clone/common.go
```diff
@@ -148,6 +145,23 @@ func GetStorageClassForClaim(ctx context.Context, c client.Client, pvc *corev1.P
 	return nil, nil
 }
 
+func getSnapshotClassForClaim(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (*string, error) {
+	if pvc.Spec.StorageClassName == nil || *pvc.Spec.StorageClassName == "" {
+		return nil, nil
```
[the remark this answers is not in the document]
arnongilboa: Fixed

pkg/controller/common/util.go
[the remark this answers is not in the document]
aglitke: Since this decision was made based on something other than the PVC/DV spec, I think we should err on the side of over-communicating this decision.  I agree with @awels .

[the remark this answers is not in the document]
awels: I guess the problem here is that we don't have the PVC so it is difficult to communicate this information.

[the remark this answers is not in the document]
awels: But we could pass in the name of the PVC, and use that to then look up the PVC and emit the event.

awels (commented):
looks good minus the issue about emitting an event.

awels (approved):

pkg/controller/storageprofile-controller.go
```diff
@@ -346,16 +353,18 @@ func NewStorageProfileController(mgr manager.Manager, log logr.Logger, installer
...
+		recorder:        mgr.GetEventRecorderFor(storageProfileControllerName),
 		scheme:          mgr.GetScheme(),
-		log:             log.WithName("storageprofile-controller"),
+		log:             log.WithName(storageProfileControllerName),
 		installerLabels: installerLabels,
 	}
 
 	storageProfileController, err := controller.New(
-		"storageprofile-controller",
+		dataImportControllerName,
```
akalenyu, line 367: dataImportControllerName?

pkg/controller/common/util.go
```diff
@@ -1887,25 +1898,74 @@ func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName *str
...
+		}
+		return nil, err
+	}
+
+	var candidates []string
+	for _, vsc := range vscList.Items {
+		if vsc.Driver == driver {
+			if vsc.Annotations[AnnDefaultSnapshotClass] == "true" {
+				logEvent(MessageDefaultVolumeSnapshotClassSelected, vsc.Name)
+				return &vsc.Name, nil
```
akalenyu, line 1954: Usually pointers to loop vars are an endless path of pain and sadness https://go.dev/blog/loopvar-preview

pkg/controller/common/util.go
```diff
@@ -1887,25 +1898,74 @@ func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName *str
...
+				pvc.Name, "snapshot class", *vscName)
 		}
+		return *vscName, nil
 	}
 
 	logger.Info("Could not match snapshotter with storage class, falling back to host assisted clone")
 	return "", nil
 }
 
+// GetVolumeSnapshotClass looks up the snapshot class based on the driver and an optional specific name
```
akalenyu, line 1917: Should we add a comment for the defined fallback behavior?

pkg/controller/common/util.go
```diff
@@ -1887,25 +1898,74 @@ func GetSnapshotClassForSmartClone(dvName string, targetPvcStorageClassName *str
...
+			if vsc.Annotations[AnnDefaultSnapshotClass] == "true" {
+				logEvent(MessageDefaultVolumeSnapshotClassSelected, vsc.Name)
+				return &vsc.Name, nil
+			}
+			candidates = append(candidates, vsc.Name)
+		}
+	}
+
+	if len(candidates) > 0 {
+		sort.Strings(candidates)
```
akalenyu, line 1961: Not sure what's the best behavior here,  I don't think it's a huge deal but have to make sure we know this
k8s does this:
https://github.com/kubevirt/containerized-data-importer/blob/539280843c306e415ec072dd87a7181ed1e364cf/pkg/controller/common/util.go#L520-L529
but the external snapshotter behaves differently:
https://github.com/kubernetes-csi/external-snapshotter/blob/ff71329d8cc08dca53194f8d4568ed53e69ecb82/pkg/common-controller/snapshot_controller.go#L1362-L1381

akalenyu (commented):
Looks good, just mainly concerned about the loop var pointer being a recipe for trouble

pkg/controller/storageprofile-controller.go
arnongilboa: good catch:)

pkg/controller/common/util.go
arnongilboa: The k8s and external snapshotter code are for getting the default storage/snapshot class, but we deal here with what to return when there is no such default.

pkg/controller/common/util.go
arnongilboa: sure, added.

pkg/controller/common/util.go
arnongilboa: fixed.

PR #2900, opened 2023-09-18
Fix possible nil ptr in snapshot/pvc clone controllers

akalenyu:
We slap on the clone-without-source index key regardless of source.PVC/source.Snapshot so
it was possible for a DV with PVC source to get queued for the snapshot controller
(and hence the nil ptr).

The test addition demonstrates a real use case where this would happen.

Fixes #
https://issues.redhat.com/browse/CNV-33002

**Release note**:

```release-note
BugFix: nil ptr in snapshot clone controller
```

### DIFF
--- a/pkg/controller/datavolume/clone-controller-base.go
+++ b/pkg/controller/datavolume/clone-controller-base.go
@@ -524,7 +524,7 @@ func isCrossNamespaceClone(dv *cdiv1.DataVolume) bool {
 }
 
 // addCloneWithoutSourceWatch reconciles clones created without source once the matching PVC is created
-func addCloneWithoutSourceWatch(mgr manager.Manager, datavolumeController controller.Controller, typeToWatch client.Object, indexingKey string) error {
+func addCloneWithoutSourceWatch(mgr manager.Manager, datavolumeController controller.Controller, typeToWatch client.Object, indexingKey string, op dataVolumeOp) error {
 	getKey := func(namespace, name string) string {
 		return namespace + "/" + name
 	}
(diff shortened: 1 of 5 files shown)


tests/dataimportcron_test.go
```diff
@@ -691,6 +692,12 @@ var _ = Describe("DataImportCron", func() {
 			err = f.CrClient.List(context.TODO(), snapshots, &client.ListOptions{Namespace: ns})
 			Expect(err).ToNot(HaveOccurred())
 			Expect(snapshots.Items).To(BeEmpty())
+			// Ensure existing PVC clones from this source don't mess up future ones
+			cloneDV := utils.NewDataVolumeForImageCloningAndStorageSpec("target-dv-from-pvc", size, ns, currentImportDv, nil, nil)
+			cloneDV, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, cloneDV)
+			Expect(err).ToNot(HaveOccurred())
+			f.ForceBindPvcIfDvIsWaitForFirstConsumer(cloneDV)
+			err = utils.WaitForDataVolumePhase(f, cloneDV.Namespace, cdiv1.Succeeded, cloneDV.Name)
```
arnongilboa, line 700: Not sure I understand the need for this check

akalenyu: This is the use case that triggers this nil ptr;

A cloned DV from a previous source PVC named `foo` exists
You upgrade CDI, now the dataimprotcron controller has a snapshot source `foo` created
This causes a DV with source.pvc to be queued in the snapshot clone controller

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -524,7 +524,7 @@ func isCrossNamespaceClone(dv *cdiv1.DataVolume) bool {
 }
 
 // addCloneWithoutSourceWatch reconciles clones created without source once the matching PVC is created
-func addCloneWithoutSourceWatch(mgr manager.Manager, datavolumeController controller.Controller, typeToWatch client.Object, indexingKey string) error {
+func addCloneWithoutSourceWatch(mgr manager.Manager, datavolumeController controller.Controller, typeToWatch client.Object, indexingKey string, op dataVolumeOp) error {
```
awels, line 527: What does `Op` mean?

akalenyu: It's the datavolume requested operation so one of these
https://github.com/kubevirt/containerized-data-importer/blob/c51229623bd2fb68acca6060291bfebcd6400ede/pkg/controller/datavolume/controller-base.go#L157-L164

akalenyu: So `op` is consistent across the project, should I refactor?

awels (approved):

PR #2908, opened 2023-09-26
Switch CI lanes to 1.28

akalenyu:
Let's start testing newer k8s in CI

Fixes #

**Release note**:

```release-note
NONE
```


automation/destructive.sh
```diff
@@ -18,7 +18,7 @@
 #
 
 set -ex
-export TARGET=k8s-1.26-centos9
+export TARGET=k8s-1.28
```
oshoval, line 21: wdyt about running here
`find cluster-provision/k8s/* -maxdepth 0 -type d -printf '%f\n'`
and taking the last line ?

and on the 1.27 ones take one before last line
and then it aint need to be changed anymore

btw if you prefer that the code will show the selected version instead an expression
then maybe this can be a helper script that replaces all the needed places in this commit, based the the above expression

akalenyu: I prefer to leave this manual, since I don't think we ever defined the version calculation formally, and I don't want to rely on the folder structure changing in the future

oshoval: folder structure didnt change for years, and i dont think will, but up to you

akalenyu:
/cc @awels
sonarcloud lying?

awels:
I need to override those in sonar cloud. I did it for all the other k8s versions already.

awels:
Okay sonarcloud is happy again.

awels:
Does look like a pretty consistent failure on the destructive lane.

akalenyu:
> Does look like a pretty consistent failure on the destructive lane.

Yeah, passed only once

tests/operator_test.go
```diff
@@ -887,15 +888,18 @@ var _ = Describe("ALL Operator tests", func() {
 					serverSecrets := getSecrets(serverSecretNames)
 
 					for _, s := range append(caSecrets, serverSecrets...) {
+						fmt.Fprintf(GinkgoWriter, "Loop 1\n")
```
awels, line 891: Might want to take these debug statements out?

akalenyu: I actually intended to keep them, very hard to debug otherwise

awels: Okay, could you make them a little more descriptive for our future selves then, loop 1,2,3 doesn't mean much without context. I certainly don't understand what they mean.

akalenyu: Done

awels (approved):

PR #2909, opened 2023-09-26
[WIP] Check content-type in importer to avoid unexpected imports

alromeros:

In some network environments, proxy servers or firewalls may block access to certain external resources.  Although we'd expect them to return a 4xx (Forbidden/Unauthorized) error code, some of these servers return an HTML error page and a 200 error code.  Currently, CDI checks if the content is compressed or a qcow2 image.  If neither of these are detected it assumes it is a raw disk image.

Even though these servers are misbehaving we can provide a better experience.  When importing an kubevirt disk image using a DataVolume, we can also check the HTTP content-type header.  If the media type is reported as 'text/html' (or text/*) we can assume that the server has sent a notice page instead of the requested content.  We can fail this request and log the content in the importer log.

This pull request implements the described behavior.

Fixes #

Still testing the code, will probably update later.

**Release note**:

```release-note
Check content-type to avoid unexpected imports
```


akalenyu:
So IIUC the tests fail because our local test servers return `text/*` content type?

alromeros:
> So IIUC the tests fail because our local test servers return `text/*` content type?

That's it, still have to check if it's our proxy who's behaving badly or just something we should allow (maybe just avoid text/html type).

alromeros:
Following some offline discussions, I've been toying with the idea of having a Prometheus metric to get important import information into the PVC or DataVolume (like this specific case). However, I think this might be outside the scope of the PR so we can merge this and groom the idea a bit further. Rejecting the import seems overkill, I think CDI is functioning as expected right know.

pkg/importer/http-datasource.go
```diff
@@ -44,10 +44,12 @@ import (
...
-	tempFile         = "tmpimage"
-	nbdkitPid        = "/tmp/nbdkit.pid"
-	nbdkitSocket     = "/tmp/nbdkit.sock"
-	defaultUserAgent = "cdi-golang-importer"
+	tempFile          = "tmpimage"
+	nbdkitPid         = "/tmp/nbdkit.pid"
+	nbdkitSocket      = "/tmp/nbdkit.sock"
+	defaultUserAgent  = "cdi-golang-importer"
+	httpContentType   = "Content-Type"
+	httpContentLenght = "Content-Length"
```
akalenyu, line 52: Lenght typo

akalenyu (commented):
/approve

pkg/importer/http-datasource.go
alromeros: Oops, fixed!

alromeros:
/hold
Let's see if I can improve the testing a little

PR #2913, opened 2023-10-01
Default storage class for virtualization purposes

akalenyu:
Storage providers in the field may supply many ways of utilizing a back end storage solution;  
They do so by providing different Kubernetes StorageClass parameters.

With the growth of kubevirt adoptions, we may sometimes see a certain combination of storage class parameters is preferrable for VM workloads.
This PR presents a possible way of steering kubevirt users
towards using such identified storage class.

Fixes #

Was considering not doing functional tests but this feels too important.
There is little to no penalty because the entire block of tests I'm adding takes less than 30 seconds.
(No wait for dv success)

**Release note**:

```release-note
Default virtualization storage class that takes precedence over k8s default storage class
```

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -369,6 +369,7 @@ func getTokenPrivateKey() *rsa.PrivateKey {
 func registerMetrics() {
 	metrics.Registry.MustRegister(controller.IncompleteProfileGauge)
 	controller.IncompleteProfileGauge.Set(-1)
+	metrics.Registry.MustRegister(controller.DefaultVirtStorageClassesGauge)
 	metrics.Registry.MustRegister(controller.DataImportCronOutdatedGauge)
 }
 
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"reflect"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -211,6 +212,8 @@ const (
 
 	//AnnDefaultStorageClass is the annotation indicating that a storage class is the default one.
 	AnnDefaultStorageClass = "storageclass.kubernetes.io/is-default-class"
+	// AnnDefaultVirtStorageClass is the annotation indicating that a storage class is the default one for virtualization purposes
+	AnnDefaultVirtStorageClass = "storageclass.kubevirt.io/is-default-virt-class"
 
 	// AnnOpenShiftImageLookup is the annotation for OpenShift image stream lookup
 	AnnOpenShiftImageLookup = "alpha.image.policy.openshift.io/resolve-names"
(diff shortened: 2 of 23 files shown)


pkg/controller/common/util.go
```diff
@@ -211,6 +212,8 @@ const (
 
 	//AnnDefaultStorageClass is the annotation indicating that a storage class is the default one.
 	AnnDefaultStorageClass = "storageclass.kubernetes.io/is-default-class"
+	// AnnDefaultVirtStorageClass is the annotation indicating that a storage class is the default one for virtualization purposes
+	AnnDefaultVirtStorageClass = "storageclass.kubevirt.io/is-default-class"
```
aglitke, line 216: Maybe this should be storageclass.kubevirt.io/is-default-virt-class to avoid confusion with the standard k8s annotation.  I know we have a different URI but still.

akalenyu: Was just following the pattern
`snapshot.storage.kubernetes.io/is-default-class`
`ingressclass.kubernetes.io/is-default-class`
`storageclass.kubernetes.io/is-default-class`

akalenyu: I am up for extra clarity though so just let me know if this still makes sense

pkg/controller/common/util.go
```diff
@@ -434,38 +437,97 @@ func GetVolumeMode(pvc *corev1.PersistentVolumeClaim) corev1.PersistentVolumeMod
...
+
+	return storageClass, nil
+}
+
+// GetStorageClassByNameWithVirtFallback looks up the storage class based on the name
+// If name is nil, it looks for the following, in this order:
+// default kubevirt storage class (if the caller is interested) storageclass.kubevirt.io/is-default-class
+// default k8s storage class storageclass.kubernetes.io/is-default-class
+// If no storage class is found, returns nil
+func GetStorageClassByNameWithVirtFallback(ctx context.Context, client client.Client, name *string, considerVirtSc bool) (*storagev1.StorageClass, error) {
```
mhenriks, line 466: Why both `GetStorageClassByName` and `GetStorageClassByNameWithVirtFallback`?  I'm sure there's a good reason.  Prob worth adding a comment why 

Also, bool params should be avoided when possible.  Maybe it would be better to have function take a DataVolume as a param instead?

pkg/controller/common/util.go
akalenyu: Done

cmd/cdi-controller/controller.go
```diff
@@ -369,6 +369,7 @@ func getTokenPrivateKey() *rsa.PrivateKey {
 func registerMetrics() {
 	metrics.Registry.MustRegister(controller.IncompleteProfileGauge)
 	controller.IncompleteProfileGauge.Set(-1)
+	metrics.Registry.MustRegister(controller.DefaultVirtStorageClassesGauge)
```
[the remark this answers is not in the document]
akalenyu: Good point, not sure if anyone would look up storage profiles though

pkg/controller/common/util.go
akalenyu: > Why both `GetStorageClassByName` and `GetStorageClassByNameWithVirtFallback`? I'm sure there's a good reason. Prob worth adding a comment why

Renamed to better communicate the reason.
Basically, there will still be flows where you don't want to take content type into account, quick example is when using the `PVC` API

> Also, bool params should be avoided when possible. Maybe it would be better to have function take a DataVolume as a param instead?

Yes definitely. Hopefully this is better now, whole DV was a bit overkill but content type should be fine

pkg/controller/common/util.go
```diff
@@ -434,38 +437,97 @@ func GetVolumeMode(pvc *corev1.PersistentVolumeClaim) corev1.PersistentVolumeMod
 	return util.ResolveVolumeMode(pvc.Spec.VolumeMode)
 }
 
-// GetStorageClassByName looks up the storage class based on the name. If no storage class is found returns nil
-func GetStorageClassByName(ctx context.Context, client client.Client, name *string) (*storagev1.StorageClass, error) {
+// GetStorageClassByNameWithK8sFallback looks up the storage class based on the name
+// If name is nil, it looks for the default k8s storage class storageclass.kubernetes.io/is-default-class
+// If no storage class is found, returns nil
+func GetStorageClassByNameWithK8sFallback(ctx context.Context, client client.Client, name *string) (*storagev1.StorageClass, error) {
+	if name == nil {
```
awels, line 444: Can't you just call `GetStorageClassByNameWithVirtFallback(ctx, client, name, cdiv1.DataVolumeArchive)` or if you think that is confusing, have a 3rd function that is called from these two. The amount of duplicated code is like 95% between these two functions.

awels (commented):
looks good, just one minor code duplication question

pkg/controller/common/util.go
akalenyu: Yep definitely, thanks, pushed a change

awels (approved):

PR #2917, opened 2023-10-04
Respect wffc override for blank block disks

### LINKED ISSUE #2915: Blank disk with block volume mode stuck in ImportScheduled for WFFC storage class
**What happened**:
Created a blank DV with block volume mode for WFFC storage class, and found that it is stuck in ImportScheduled phase

Expect the disk to be in Succeeded phase.

1. Have a storage class that supports block mode and WFFC
2. Create a blank DV as below:
```
kind: DataVolume
metadata:
  name: empty-disk
spec:
  source:
    blank: {}
  storage:
    accessModes:
    - ReadWriteMany
    resources:
      requests:
        storage: 10Gi
    storageClassName: block-sc
    volumeMode: Block
```
3. Observe that the dv is stuck in ImportScheduled phase.

**Additional context**:

Adding the annotation `cdi.kubevirt.io/storage.usePopulator: "false"`  to the dv allows the import to fall back to the old way and bypass the issue. 
    
**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.57.0

akalenyu:
We currently don't support the wffc override for blank block disks,
while there may be some use cases where that is desired.

Fixes #2915 

**Release note**:

```release-note
BugFix: wffc override not respected for blank block disks
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -215,20 +215,6 @@ func (r *ImportReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 		return reconcile.Result{}, nil
 	}
 
-	// In case this is a request to create a blank disk on a block device, we do not create a pod.
-	// we just mark the DV as successful
-	volumeMode := cc.GetVolumeMode(pvc)
-	if volumeMode == corev1.PersistentVolumeBlock && pvc.GetAnnotations()[cc.AnnSource] == cc.SourceNone && pvc.GetAnnotations()[cc.AnnPreallocationRequested] != "true" {
-		log.V(1).Info("attempting to create blank disk for block mode, this is a no-op, marking pvc with pod-phase succeeded")
-		if pvc.GetAnnotations() == nil {
-			pvc.SetAnnotations(make(map[string]string, 0))
-		}
-		pvc.GetAnnotations()[cc.AnnPodPhase] = string(corev1.PodSucceeded)
-		if err := r.updatePVC(pvc, log); err != nil {
-			return reconcile.Result{}, errors.WithMessage(err, fmt.Sprintf("could not update pvc %q annotation and/or label", pvc.Name))
-		}
-		return reconcile.Result{}, nil
-	}
 	return r.reconcilePvc(pvc, log)
 }
 
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -587,31 +587,34 @@ var _ = Describe("Importer Test Suite-Block_device", func() {
 
 	})
 
-	It("[test_id:4972]Should create blank raw image for block PV", func() {
+	DescribeTable("Should create blank raw image for block PV", func(consumer bool) {
 		if !f.IsBlockVolumeStorageClassAvailable() {
 			Skip("Storage Class for block volume is not available")
 		}
 		dv := utils.NewDataVolumeForBlankRawImageBlock("create-blank-image-to-block-pvc", "500Mi", f.BlockSCName)
+		if !consumer {
+			controller.AddAnnotation(dv, controller.AnnImmediateBinding, "true")
+		}
 		dv, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
 		Expect(err).ToNot(HaveOccurred())
 
 		By("verifying pvc was created")
 		pvc, err := utils.WaitForPVC(f.K8sClient, dv.Namespace, dv.Name)
 		Expect(err).ToNot(HaveOccurred())
-		f.ForceBindIfWaitForFirstConsumer(pvc)
+		if consumer {
+			f.ForceBindIfWaitForFirstConsumer(pvc)
+		}
 
 		By("Waiting for import to be completed")
 		err = utils.WaitForDataVolumePhase(f, f.Namespace.Name, cdiv1.Succeeded, dv.Name)
 		Expect(err).ToNot(HaveOccurred(), "Datavolume not in phase succeeded in time")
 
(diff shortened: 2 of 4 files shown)


tests/import_test.go
```diff
@@ -611,6 +616,15 @@ var _ = Describe("Importer Test Suite-Block_device", func() {
...
 		}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(BeTrue())
 		Expect(err).ToNot(HaveOccurred())
+
+		err = utils.WaitForPersistentVolumeClaimPhase(f.K8sClient, pvc.Namespace, v1.ClaimBound, pvc.Name)
+		Expect(err).ToNot(HaveOccurred())
+	},
+		Entry("[test_id:4972] with first consumer", true),
+		Entry("with bind immediate annotation", false),
+	)
+	It("Should create blank raw image for block PV", func() {
```
alromeros, line 626: Code leftover, looks good otherwise. Thanks for the PR!

akalenyu: whoops thx

pkg/controller/import-controller.go
```diff
@@ -220,12 +220,11 @@ func (r *ImportReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 	volumeMode := cc.GetVolumeMode(pvc)
 	if volumeMode == corev1.PersistentVolumeBlock && pvc.GetAnnotations()[cc.AnnSource] == cc.SourceNone && pvc.GetAnnotations()[cc.AnnPreallocationRequested] != "true" {
 		log.V(1).Info("attempting to create blank disk for block mode, this is a no-op, marking pvc with pod-phase succeeded")
-		if pvc.GetAnnotations() == nil {
-			pvc.SetAnnotations(make(map[string]string, 0))
-		}
-		pvc.GetAnnotations()[cc.AnnPodPhase] = string(corev1.PodSucceeded)
-		if err := r.updatePVC(pvc, log); err != nil {
-			return reconcile.Result{}, errors.WithMessage(err, fmt.Sprintf("could not update pvc %q annotation and/or label", pvc.Name))
+		if pvc.Status.Phase == corev1.ClaimBound {
```
awels, line 223: So I am not sure about this. We purposely mark the PVC succeeded in the blank block case regardless of if it is bound or not. This is so that KubeVirt won't think it can't use the PVC. Once we have a consumer kubernetes will bind the PVC and everything resolves itself.

This was pre-populator logic. Now I don't really see how populators would affect blank block volumes at all, since there is no population. Maybe I am missing something.

akalenyu: > This is so that KubeVirt won't think it can't use the PVC

KubeVirt should be fine with WFFC Pending PVCs/DV:
https://github.com/kubevirt/kubevirt/pull/9812

awels: Right, so I am trying to understand why we want this particular check here. I don't think we even want to use populators, unless maybe pre-allocation is set to true.

akalenyu: The problem is if we stop using populators with this flow, we're back at the doppelganger pod solution

awels: For the blank volume without pre-allocation case, there is no need for the doppel ganger. It is 'populated' already considering there is nothing we have to do. It just gets bound when the VM starts, but it is still exactly the way we want it. With pre-allocation, I would expect our populator to attempt the pr-eallocation at the appropriate time.

akalenyu:
/hold

Taking a step back. This works with HonorWaitForFirstConsumer enabled creating a bind pod.
The only thing that is broken is using `bind.immediate` with block blank DVs (because no consumer will be created)

pkg/controller/import-controller.go
lzang: In my case, the PVC is not in bound state. It is in WFFC. 
If the issue is because I didn't enable HonorWaitForFirstConsumer feature gate, is there any side effect or behavior change of enabling this feature gate?

akalenyu: So we've been going with this feature for quite a few releases now,
it basically means we respect WFFC binding by not creating any worker pods until PVCs get scheduled.

Is there any particular use case you have that would require bypassing WFFC behavior,
and bind PVCs randomly? Or will these blank disks be consumed by a VM?

akalenyu:
@lzang @awels PTAL

pkg/controller/import-controller.go
```diff
@@ -215,20 +215,6 @@ func (r *ImportReconciler) Reconcile(_ context.Context, req reconcile.Request) (
...
-		log.V(1).Info("attempting to create blank disk for block mode, this is a no-op, marking pvc with pod-phase succeeded")
-		if pvc.GetAnnotations() == nil {
-			pvc.SetAnnotations(make(map[string]string, 0))
-		}
-		pvc.GetAnnotations()[cc.AnnPodPhase] = string(corev1.PodSucceeded)
-		if err := r.updatePVC(pvc, log); err != nil {
-			return reconcile.Result{}, errors.WithMessage(err, fmt.Sprintf("could not update pvc %q annotation and/or label", pvc.Name))
-		}
-		return reconcile.Result{}, nil
-	}
```
lzang, line 231: I don't get why this is not needed any more from looking at the CL.  Could you please clarify why? 
Does this logic work for both populator case and non-populator case?

awels: Essentially this was a short cut for one particular case (blank disk, no preallocation) and removing it will cause a pod to be created, but per the rest of this PR, that pod won't do anything. What it will do is follow the rest of the logic when creating a population pod, which should work for both populator and non populator cases, as well as HonorWaitForFirstConsumer set or not.

PR #2928, opened 2023-10-16
Improve DataVolume status reporting with populators

alromeros:

With the addition of populators into CDI, we introduced a minor divergence in the DataVolume status reporting between the new and the old flow to avoid checking PVC Prime from the legacy controllers.

Though it would be ideal to ignore the PVC Prime in the DV layer completely, there are certain cases where the reported status does not match the current behavior (check https://bugzilla.redhat.com/show_bug.cgi?id=2241637 for an example).

This PR aims to update the status reporting mechanism in import and upload controllers so both the old flow and the populator flow match.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2241637

It would be ideal to completely ignore the PVC Prime from the DV layer, but I think this solution will help us with all the specific cases where we are reporting a status that doesn't match CDI behavior. Let me know what you think.

**Release note**:

```release-note
Bugfix: Improve DataVolume status reporting with populators
```

### DIFF
--- a/pkg/controller/datavolume/upload-controller.go
+++ b/pkg/controller/datavolume/upload-controller.go
@@ -196,19 +196,37 @@ func isPVCUploadPopulation(pvc *corev1.PersistentVolumeClaim) bool {
 	return populators.IsPVCDataSourceRefKind(pvc, cdiv1.VolumeUploadSourceRef)
 }
 
+func (r *UploadReconciler) shouldUpdateStatusPhase(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) (bool, error) {
+	pvcCopy := pvc.DeepCopy()
+	if isPVCUploadPopulation(pvcCopy) {
+		// Better to play it safe and check the PVC Prime too
+		// before updating DV phase.
+		nn := types.NamespacedName{Namespace: pvcCopy.Namespace, Name: populators.PVCPrimeName(pvcCopy)}
+		err := r.client.Get(context.TODO(), nn, pvcCopy)
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
+		}
+	}
+	_, ok := pvcCopy.Annotations[cc.AnnUploadRequest]
+	return ok && pvcCopy.Status.Phase == corev1.ClaimBound && !pvcIsPopulated(pvcCopy, dv), nil
+}
+
 func (r *UploadReconciler) updateStatusPhase(pvc *corev1.PersistentVolumeClaim, dataVolumeCopy *cdiv1.DataVolume, event *Event) error {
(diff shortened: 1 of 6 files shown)


alromeros:
/cc @mhenriks
I've been trying to avoid checking the PVC Prime but I think this is a good solution for all the corner cases where we report `ImportScheduled` or `UploadScheduled` with an indefinitely pending PVC Prime. Let me know what you think.

alromeros:
I'm wondering how to address the error that's popping up in the [tests](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/2928/pull-containerized-data-importer-e2e-ceph-wffc/1713851829361577984). There's this specific case where we test a failed import in which the PVC Prime gets bound after the target PVC is updated for the last time. Since we don't trigger reconciles for PVC Prime updates, we don't update the status phase to `ImportScheduled`. It's just a minor cosmetic divergence from the non-populator flow, but still a bummer since all this PR is to address these divergences. Don't know if there's a way to solve this unless we add a watch to reconcile PVC Primes, which seems overkill for this specific case. Will continue with it tomorrow.

pkg/controller/datavolume/upload-controller.go
```diff
@@ -196,19 +196,37 @@ func isPVCUploadPopulation(pvc *corev1.PersistentVolumeClaim) bool {
 	return populators.IsPVCDataSourceRefKind(pvc, cdiv1.VolumeUploadSourceRef)
 }
 
+func (r *UploadReconciler) shouldUpdateStatusPhase(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) (bool, error) {
```
mhenriks, line 199: Is this function duplicated in import controller?

alromeros: There's common code but one function uses `isPVCImportPopulation` and `AnnImportPod` and the other the upload equivalent. I can try to have one shared function but I think this practice is somewhat common in our controllers.

pkg/controller/populators/import-populator.go
```diff
@@ -248,8 +248,16 @@ func (r *ImportPopulatorReconciler) updateImportProgress(podPhase string, pvc, p
 	if err != nil {
 		return err
 	}
+	if importPod == nil {
+		_, ok := pvc.Annotations[cc.AnnPopulatorProgress]
+		// Initialize the progress once PVC Prime is bound
+		if !ok && cc.IsBound(pvcPrime) {
+			cc.AddAnnotation(pvc, cc.AnnPopulatorProgress, "N/A")
```
alromeros, line 255: This is a nasty patch to force a reconcile in the DV controller when the PVC' gets bound disguised as an annotation to initialize progress. It should fix the error in the tests, and it matches the behavior in the DV controller. After spending hours on this I think this is the best-looking solution.

Issue #2515, opened 2022-12-21
clone_progress metric doesn't meet the metrics naming conventions


sradco:
**What happened**:
clone_progress metric doesn't meet the metrics naming conventions.
- https://prometheus.io/docs/instrumenting/writing_exporters/#naming
- https://sdk.operatorframework.io/docs/best-practices/observability-best-practices/

This makes it hard for users to identify the metric and to what operator it is related to.
It also 

We should deprecate the clone_progress metric and separate it into 2 metrics:
1. A metric that reports the how much was already processed
2. A metric that reports the remaining bytes of the clone that needs to be processed.

Similar to the kubevirt_migrate_vmi_data_processed_bytes and kubevirt_migrate_vmi_data_remaining_bytes metrics.

This way the percentage can be calculated by dividing the 2 metrics as it is recommended in Prometheus 
"Expose ratios, not percentages. Even better, specify a counter for each of the two components of the ratio."

**What you expected to happen**:
The metrics should follow the best practices so that the operator users can find them and use them correctly.

**How to reproduce it (as minimally and precisely as possible)**:
Steps to reproduce the behavior.

**Additional context**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): N/A
- Kubernetes version (use `kubectl version`): N/A
- DV specification: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

alromeros:
Hello @sradco, thanks for opening this issue!

I created a Jira card https://issues.redhat.com/browse/CNV-26873 to track this.

Since we are planning on doing a big refactor/development in CDI to implement custom K8s populators per source type, that would probably be a good opportunity to modify the cloning metrics to match the naming conventions.

avlitman:
Please make sure to update [`import_progress](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/image/qemu_test.go#L247)` and `[test_progress](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/util/prometheus/prometheus_test.go#L25)` to fit the new naming for `clone_progress`

avlitman:
[...]

PR #2929, opened 2023-10-18
Move clone_progress to separate list

avlitman:
Since clone_progress metric used for internal purpose only we should not document it in metrics.md or test it in the metrics linter. therefore clone_progress moved to separate list (not used by the doc and linter generators).

Fixes https://github.com/kubevirt/containerized-data-importer/issues/2515
Jira-ticket: https://issues.redhat.com/browse/CNV-33647

This fix was discussed with CDI developer.

**Release note**:

```
The "clone_progress" metric has been removed from the project documentation. We have decided to exclude this metric from the documentation as it is intended for internal use only.
```


cmd/cdi-cloner/clone-source.go
```diff
@@ -96,8 +96,8 @@ func startPrometheus() {
 func createProgressReader(readCloser io.ReadCloser, ownerUID string, totalBytes uint64) io.ReadCloser {
 	progress := prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: monitoring.MetricOptsList[monitoring.CloneProgress].Name,
```
sradco, line 99: @avlitman  Why should we keep them here at all?

avlitman:
Please do not merge this PR before @sradco approves.

PR #2940, opened 2023-10-25
Ensure AnnPVCPrimeName annotation is always removed after upload succeeds

alromeros:

In our existing workflow, we aim to remove the AnnPVCPrimeName annotation from the target PVC once the upload populator succeeds. 

However, our current implementation of `updatePVCWithPVCPrimeAnnotations` only updates the PVC annotations when there is a modification to the object. This means that, since `AnnPVCPrimeName` annotation is removed in a separate function, there are scenarios where this change isn't reflected in the PVC as expected.

This PR aims to fix this behavior by removing the annotation inside `updatePVCWithPVCPrimeAnnotations`.  

Partial fix of https://bugzilla.redhat.com/show_bug.cgi?id=2241658.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/populators/upload-populator.go
+++ b/pkg/controller/populators/upload-populator.go
@@ -131,12 +131,14 @@ func (r *UploadPopulatorReconciler) updatePVCPrimeNameAnnotation(pvc *corev1.Per
 	return true, nil
 }
 
-func removePVCPrimeNameAnnotation(pvc *corev1.PersistentVolumeClaim) {
+func (r *UploadPopulatorReconciler) updateUploadAnnotations(pvc *corev1.PersistentVolumeClaim, pvcPrime *corev1.PersistentVolumeClaim) {
 	if _, ok := pvc.Annotations[AnnPVCPrimeName]; !ok {
 		return
 	}
-
-	delete(pvc.Annotations, AnnPVCPrimeName)
+	// Delete the PVC Prime annotation once the pod is succeeded
+	if pvcPrime.Annotations[cc.AnnPodPhase] == string(corev1.PodSucceeded) {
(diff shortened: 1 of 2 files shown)


alromeros:
The failures look really odd

pkg/controller/populators/upload-populator.go
```diff
@@ -131,12 +131,14 @@ func (r *UploadPopulatorReconciler) updatePVCPrimeNameAnnotation(pvc *corev1.Per
...
-func removePVCPrimeNameAnnotation(pvc *corev1.PersistentVolumeClaim) {
+func (r *UploadPopulatorReconciler) updateUploadAnnotations(pvc *corev1.PersistentVolumeClaim, pvcPrime *corev1.PersistentVolumeClaim) {
 	if _, ok := pvc.Annotations[AnnPVCPrimeName]; !ok {
 		return
 	}
-
-	delete(pvc.Annotations, AnnPVCPrimeName)
+	// Delete the PVC Prime annotation once the pod is succeeded
+	if pvcPrime.Annotations[cc.AnnPodPhase] == string(corev1.PodSucceeded) {
+		delete(pvc.Annotations, AnnPVCPrimeName)
```
mhenriks, line 140: perhaps a unit test for this?

alromeros: Okay. This flow was already tested but just added a new one to specifically cover `r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateUploadAnnotations)`. Still having lots of upload errors in the tests though, I'm clueless about what's failing.

alromeros:
I think the failures were unrelated to the PR and it was just a matter of bad timing after consequent retests. This should be ready.

PR #2944, opened 2023-10-26
Mount block PVC rw in clone source pods

akalenyu:
Some CSI drivers [0] don't allow mounting block ro, so they simply cannot perform host assisted clones.
Host assisted clones are important as they are a last resort in some cases like switching over to a new storage provisioner.

This may or may not have some performance implications (parallel cloning from single source)
but we are choosing compatibility over performance in this case.

[0] https://github.com/dell/csi-unity/blob/76cb78decd0c7e0ae3bded9668e7fbe9b39dd2ff/service/node.go#L420-L424

Fixes #
https://issues.redhat.com/browse/CNV-33859

**Release note**:

```release-note
BugFix: Some provisioners don't allow mounting block PVC ro
```

### DIFF
--- a/pkg/controller/clone-controller_test.go
+++ b/pkg/controller/clone-controller_test.go
@@ -178,6 +178,41 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		}),
 	)
 
+	It("Should create new source pod if source PVC is in use by other host assisted source pod", func() {
+		testPvc := cc.CreatePvc("testPvc1", "default", map[string]string{
+			cc.AnnCloneRequest:   "default/source",
+			cc.AnnPodReady:       "true",
+			cc.AnnCloneToken:     "foobaz",
+			AnnUploadClientName:  "uploadclient",
+			cc.AnnCloneSourcePod: "default-testPvc1-source-pod"}, nil)
+		sourcePvc := cc.CreatePvc("source", "default", map[string]string{}, nil)
+		pod := podUsingBlockPVC(sourcePvc, false)
+		pod.Name = "other-target-pvc-uid" + common.ClonerSourcePodNameSuffix
+		pod.Labels = map[string]string{
+			common.CDIComponentLabel: common.ClonerSourcePodName,
+		}
+		reconciler = createCloneReconciler(testPvc, sourcePvc, pod)
+		By("Setting up the match token")
+		reconciler.multiTokenValidator.ShortTokenValidator.(*cc.FakeValidator).Match = "foobaz"
+		reconciler.multiTokenValidator.ShortTokenValidator.(*cc.FakeValidator).Name = "source"
+		reconciler.multiTokenValidator.ShortTokenValidator.(*cc.FakeValidator).Namespace = "default"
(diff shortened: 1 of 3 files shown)


pkg/controller/common/util.go
```diff
@@ -603,6 +603,12 @@ func GetPodsUsingPVCs(ctx context.Context, c client.Client, namespace string, na
 									onlyReadOnly = false
 								}
 							}
+							for _, vm := range c.VolumeDevices {
+								if vm.Name == volume.Name {
+									// Node level rw mount and container can't mount block device ro
+									onlyReadOnly = false
+								}
+							}
```
akalenyu, line 611: @mhenriks @awels I think this check was always broken for pods with volumeDevices

awels (approved):

pkg/controller/common/util.go
```diff
@@ -612,6 +618,12 @@ func GetPodsUsingPVCs(ctx context.Context, c client.Client, namespace string, na
 						// all mounts must be ro
 						addPod = false
 					}
+					if strings.HasSuffix(pod.Name, common.ClonerSourcePodNameSuffix) && pod.Labels != nil &&
+						pod.Labels[common.CDIComponentLabel] == common.ClonerSourcePodName {
+						// Host assisted clone source pod only reads from source
+						// But some drivers disallow mounting a block PVC ReadOnly
+						addPod = false
```
alromeros, line 625: Several consecutive conditions lead to the same result (false addPod), would it be possible to clean-up this a little so we have it more condenssed? Maybe something like a `shouldAddPod` function? Not a must but I think this code is getting hard to follow.

pkg/controller/clone-controller_test.go
```diff
@@ -178,6 +178,41 @@ var _ = Describe("Clone controller reconcile loop", func() {
 		}),
 	)
 
+	It("Should create new source pod if source PVC is in use by other host assisted source pod", func() {
```
alromeros, line 181: Nice test

pkg/controller/common/util.go
akalenyu: Yeah the whole function can be refactored but I prefer to stay on the safe side since this is going to backport all the way to 0.55

mhenriks:
/hold
let me look at this

pkg/controller/common/util.go
mhenriks: Kind of.  This will only cover cases where block PVCs are in the pod spec read/write but not referred to by a container.

pkg/controller/common/util.go
```diff
@@ -612,6 +618,12 @@ func GetPodsUsingPVCs(ctx context.Context, c client.Client, namespace string, na
 						// all mounts must be ro
 						addPod = false
 					}
+					if strings.HasSuffix(pod.Name, common.ClonerSourcePodNameSuffix) && pod.Labels != nil &&
+						pod.Labels[common.CDIComponentLabel] == common.ClonerSourcePodName {
```
mhenriks, line 622: `pod.Labels != nil` is not required.   `pod.Labels[common.CDIComponentLabel] == common.ClonerSourcePodName` can still be executed when pod.Labels is nil

PR #2954, opened 2023-10-31
Add support for linkerd injection annotations on datavolumes

### LINKED ISSUE #2948: Unable to mesh CDI importer pod with linkerd
**What happened**:

Attempted to mesh the CDI component as per linkerd documentation, adding `linkerd.io/inject: enabled` annotation to CDI `importer` pods. The problem is the life cycle of sidecar containers and is something [Kubernetes is looking to address](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/). It is currently feature gated and not GA. AFAIK, Istio suffers the same problem, because a k8s Job is not considered complete until all containers have stopped running.

```
Containers:
  linkerd-proxy:
...
    State:          Running
      Started:      Mon, 30 Oct 2023 15:33:21 +0000
    Ready:          True
...
  importer:
...
    State:          Terminated
      Reason:       Completed
      Message:      Import Complete
      Exit Code:    0
      Started:      Mon, 30 Oct 2023 15:33:22 +0000
      Finished:     Mon, 30 Oct 2023 15:33:35 +0000
    Ready:          False
...
  server:
...
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Mon, 30 Oct 2023 15:33:22 +0000
      Finished:     Mon, 30 Oct 2023 15:33:36 +0000
    Ready:          False
...
```

Should be able to mesh cdi importer pods with linkerd.

1. Deploy linkerd control plane using the helm chart as per the docs https://linkerd.io/2.14/tasks/install-helm/.
2. Create a VM that downloads disk image using CDI (i.e. download container image).
3. Mesh the CDI importer Pod with annotation: `linkerd.io/inject: enabled`.
4. Wait for pod to complete container image download. Will see that the `linkerd-proxy` continues to run indefinitely.

**Additional context**:

Related to https://github.com/kubevirt/kubevirt/issues/10556.

Kubernetes are actively looking to address this although currently behind feature gate. There are ways around this at the moment and this [article](https://tech.ingrid.com/on-linkerd-and-kubernetes-jobs/) explains it pretty well.

As per linkerds suggestion, on completion of importer, the container should call `curl -X POST http://localhost:4191/shutdown` to shutdown the linkerd proxy container.

The alternate solution is to allow the annotation `linkerd.io/inject: enabled` to propagate the annotation to importer pod as done with istio. Although the first option is preferred as it allows us to mesh the pod.

**Environment**:
- DV specification: N/A
- Install tools: N/A
- Others: N/A

bc185174:

The PR allows users to set `linkerd.io/inject` annotation on `DataVolumes` to disable/enable linkerd side car injection (as per the documentation https://linkerd.io/2.14/features/proxy-injection/).

Fixes #2948 

**Release note**:

```release-note
Add support for linkerd injection annotations on DataVolumes
```


tests/import_test.go
```diff
@@ -325,7 +325,8 @@ var _ = Describe("[Istio] Namespace sidecar injection", func() {
 		By(fmt.Sprintf("Create new datavolume %s", dataVolume.Name))
 		// We set the Immediate Binding annotation to true, to eliminate creation of the consumer pod, which will also fail due to the Istio sidecar.
 		dataVolume.Annotations[controller.AnnImmediateBinding] = "true"
-		dataVolume.Annotations[controller.AnnPodSidecarInjection] = "true"
+		dataVolume.Annotations[controller.AnnPodSidecarInjectionIstio] = "true"
+		dataVolume.Annotations[controller.AnnPodSidecarInjectionLinkerd] = "true"
```
akalenyu, line 329: Wondering if it would make sense to have a separate permutation of the test for Linkerd.
Likely not - because only one of these would be present on a cluster, right?
If so, we could add a comment about this

akalenyu (commented):
Looks good! thank you, minor comment

tests/import_test.go
bc185174: Either would work yes, but you're right - only one of these meshes _should_ be present on a cluster. Nothing would stop you from deploying both meshes but reckon you'll get some unexpected behaviour. Generally its one or the other. If think its necessary I can add a comment on this?

PR #2963, opened 2023-11-02
Add doc for fallback to host-assisted cloning

arnongilboa:

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/doc/efficient-cloning.md
+++ b/doc/efficient-cloning.md
@@ -27,6 +27,26 @@ _The required prerequisites in order to trigger efficient cloning methods_
 
 *Note: Data Volume Cloning can work together with namespace transfer and size expansion*  
 
+### Fallback to Host-Assisted Cloning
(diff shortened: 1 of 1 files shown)


doc/efficient-cloning.md
```diff
@@ -27,6 +27,26 @@ _The required prerequisites in order to trigger efficient cloning methods_
 
 *Note: Data Volume Cloning can work together with namespace transfer and size expansion*  
 
+### Fallback to Host-Assisted Cloning
+Whenever the above efficient cloning prerequisites are not met, we fallback to host-assisted cloning (`cloneType: copy`). In that case, the target PVC is annotated with the fallback reason and an event is emitted.
```
alromeros, line 31: I think we can complete this more with a description of how host-assisted clone works. Noticed that we don't have any documentation that focuses on this clone type (as we do with smart and csi), so maybe we could leave that for a follow-up and just reference that documentation here.

arnongilboa: Sure, added a line, and there are already some details [here](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/clone-datavolume.md).

PR #2976, opened 2023-11-09
Work around common snapshot controller issue in tests

akalenyu:
This issue prevents the common csi snapshot controller from removing it's PVC finalizer:
https://github.com/kubernetes-csi/external-snapshotter/issues/957

This causes us some pain in testing:
https://search.ci.kubevirt.io/?search=dataimportcron_test.go%3A138&maxAge=336h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job
so let's try working around it.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/dataimportcron_test.go
+++ b/tests/dataimportcron_test.go
@@ -134,7 +134,16 @@ var _ = Describe("DataImportCron", func() {
 				},
 			}
 			snapshot = utils.WaitSnapshotReady(f.CrClient, snapshot)
-			deleted, err := utils.WaitPVCDeleted(f.K8sClient, name, ns, 2*time.Minute)
+			deleted, err := utils.WaitPVCDeleted(f.K8sClient, name, ns, 30*time.Second)
+			if err != nil {
+				// work around https://github.com/kubernetes-csi/external-snapshotter/issues/957
+				// it does converge after the resync period of snapshot controller (15mins)
+				cc.AddAnnotation(snapshot, "workaround", "triggersync")
+				err = f.CrClient.Update(context.TODO(), snapshot)
+				Expect(err).ToNot(HaveOccurred())
+				// try again
+				deleted, err = utils.WaitPVCDeleted(f.K8sClient, name, ns, 30*time.Second)
+			}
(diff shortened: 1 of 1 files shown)


tests/dataimportcron_test.go
```diff
@@ -134,7 +134,16 @@ var _ = Describe("DataImportCron", func() {
 				},
 			}
 			snapshot = utils.WaitSnapshotReady(f.CrClient, snapshot)
-			deleted, err := utils.WaitPVCDeleted(f.K8sClient, name, ns, 2*time.Minute)
+			deleted, err := utils.WaitPVCDeleted(f.K8sClient, name, ns, 30*time.Second)
+			if err != nil {
+				// work around https://github.com/kubernetes-csi/external-snapshotter/issues/957
+				// it does converge after the resync period of snapshot controller (15mins)
+				cc.AddAnnotation(snapshot, "workaround", "triggersync")
+				err = f.CrClient.Update(context.TODO(), snapshot)
```
awels, line 142: So you are essentially just re-triggering a reconcile so it cleans up, what happens if that gets the same error, maybe retry a few times instead of just once? I don't know if that is overkill or not.

akalenyu: Yeah good call I am not sure. It fails on
```
E1109 05:15:43.755196       1 snapshot_controller.go:191] error check and remove PVC finalizer for snapshot [datasource-test-63c30e708034]: snapshot controller failed to update datasource-test-63c30e708034 on API server: Operation cannot be fulfilled on persistentvolumeclaims "datasource-test-63c30e708034": the object has been modified; please apply your changes to the latest version and try again
```

akalenyu: I think we won't need to retry more since 30 seconds have passed so it had plenty of times to catch up

alromeros:
Seems that we can't do any better here, looks good. Maybe use a shorter timeout in the second waitPVCDeleted? Just a suggestion though.
/lgtm

PR #2980, opened 2023-11-12
Adjust functest suite for running an allowed set of tests in parallel

akalenyu:
Currently, our entire test suite runs serially.
This means it takes a good while just to get through a PR feedback cycle - 5-6hrs.
Most of our tests don't have to be serial, as they won't collide with other tests.

Others, namely, ones that alter global config (CDIConfig/cluster PVs/storage classes) are more challenging
to parallelize.

This is an attempt to parallelize some of our tests.

Important to ensure before proceeding:
- [ ] Number of ran tests in report doesn't change
- [ ] Fail some tests (artificially if needed), ensure artifacts collected correctly for both serial and parallel, no overwrites (https://github.com/kubevirt/kubevirt/pull/10562/files#r1391451698)

Fixes #

One example for something that needs more thought is `ClonerBehavior`. This sub-suite is run 3 times for each cloning strategy;
We adjust the default cluster storage class for a specific clone strategy, so we can run the behavioral suite on each of those.
This can't be trivially made parallel as it will collide with other tests that use the default storage class (potentially all of them lol)

**Release note**:

```release-note
Run a subset of functional tests in parallel
```


mhenriks:
/lgtm
/approve

Great work @akalenyu!

PR #2947, opened 2023-10-30
forklift: add forklift populators to cdi-importer

bennyz:
Add forklift populators, ovirt-populator and openstack-populator to the cdi-importer image, this includes:
- Add RPM dependencies to cdi-importer for ovirt-populator (ovirt-imageio-client, ovirt-engine-sdk)
- Executables for the populators under cmd/openstack-populator and cmd/ovirt-populator, and use them in cdi-importer-image
- OpenStack go SDK library - gophercloud - added for the OpenStack populator

Changed according to discussion in https://github.com/kubevirt/containerized-data-importer/pull/2895

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
add forklift populators apps to cdi-importer image
```

PR #2983, opened 2023-11-13
forklift: introduce forklift controller

bennyz:
A controller to run the forklift populators, it will handle the population for the OvirtVolumePopulator and OpenstackVolumePopulator kinds

Example DVs:
```yaml
kind: DataVolume
metadata:
  name: my-populator-dv
  annotations:
spec:
  storage:
    dataSourceRef:
        apiGroup: forklift.konveyor.io
        kind: OvirtVolumePopulator
        name: test
    resources:
      requests:
        storage: 1G
    accessMode:
      - ReadWriteOnce
    storageClassName: nfs-csi
```

```yaml
apiVersion: forklift.konveyor.io/v1beta1
kind: OvirtVolumePopulator
metadata:
  name: test
  namespace: default
spec:
  engineUrl: https://ovirt.fqdn
  engineSecretName: ovirt-secret
  diskId: 134db723-7ba2-47c4-b19a-975383ff1491
```

Fixes #

depends on https://github.com/kubevirt/containerized-data-importer/pull/2947
**Release note**:

```release-note
Introduce a controller to handle forklift's volume populators (ovirt, openstack)
```

TODO:
- [ ] Make sure all annotations are properly set
- [ ] Handle pod cleanup
- [ ] Complete tests


bennyz:
@akalenyu could you have a preliminary look?

akalenyu:
> @akalenyu could you have a preliminary look?

So this is fine for review, regardless of the ovirt populator still not being in?

bennyz:
> > @akalenyu could you have a preliminary look?
> 
> So this is fine for review, regardless of the ovirt populator still not being in?

This fetches the ovirt populator from our repo, as this is seems the direction we'll be going for, and even if not, switching back to the original approach will not be difficult and won't affect the core controller code too much, just the boilerplate code around

hack/build/config.sh
```diff
@@ -61,6 +61,9 @@ QUAY_REPOSITORY=${QUAY_REPOSITORY:-cdi-operatorhub}
 QUAY_NAMESPACE=${QUAY_NAMESPACE:-kubevirt}
 CDI_LOGO_PATH=${CDI_LOGO_PATH:-"assets/cdi_logo.png"}
 
+# oVirt populator image from forklift
+OVIRT_POPULATOR_IMAGE_NAME=${OVIRT_POPULATOR_IMAGE_NAME:-"quay.io/kubev2v/ovirt-populator:release-v2.5.4"}
```
akalenyu, line 65: cc @alromeros @mhenriks 
my worry is productizing this in downstream

akalenyu:
Do we want this in draft mode? I think it's ready for review

hack/build/config.sh
akalenyu: WDYT about including & using the sources under CDI? at least in our scripts
On other distributions (downstream) the env var will be set to something else and that will get tested instead.

bennyz: no problem

akalenyu:
I think I can give this PR testing rights once the WIP is removed from title & release note added in PR description

bennyz:
> I think I can give this PR testing rights once the WIP is removed from title & release note added in PR description

I see it doesn't even compile anymore after recent changes, I'll fix the issues retest and proceed

hack/build/config.sh
akalenyu: this is not implemented yet, right?

bennyz: not yet, just to clarify, we want to build it in CDI?
something like was done here: https://github.com/kubevirt/containerized-data-importer/pull/2895 ?

akalenyu: yeah, any drawbacks? lets us stay completely in charge upstream if anything breaks down

akalenyu: not sure why you linked the closed PR, the one that merged also had the ovirt populator sources in it:
https://github.com/kubevirt/containerized-data-importer/pull/2947

bennyz: in the closed one ovirt-populator is built separately which is what I assume we want here, while in https://github.com/kubevirt/containerized-data-importer/pull/2947 it's part of cdi-importer

akalenyu: ah, I see. maybe for consistency let's keep it non separate? wdyt?

bennyz: published, only did a basic sanity check but looks ok

bennyz:
```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  annotations:
    volume.beta.kubernetes.io/storage-provisioner: nfs.csi.k8s.io
    cdi.kubevirt.io/createdForDataVolume: f7963dbd-3d39-455e-ae96-e43033d45367
    cdi.kubevirt.io/storage.usePopulator: 'true'
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"cdi.kubevirt.io/v1beta1","kind":"DataVolume","metadata":{"annotations":{},"name":"ovirt-dv","namespace":"default"},"spec":{"storage":{"accessMode":["ReadWriteOnce"],"dataSourceRef":{"apiGroup":"forklift.konveyor.io","kind":"OvirtVolumePopulator","name":"ovirt-pop"},"resources":{"requests":{"storage":"13G"}},"storageClassName":"nfs-csi"}}}
    cdi.kubevirt.io/storage.populator.progress: 7.81%
    volume.kubernetes.io/storage-provisioner: nfs.csi.k8s.io
    cdi.kubevirt.io/externalPopulation: 'true'
    cdi.kubevirt.io/storage.pod.phase: Running
    cdi.kubevirt.io/storage.contentType: kubevirt
  resourceVersion: '6493875'
  name: ovirt-dv
  uid: 3e9f76f9-3ca2-4587-8d13-30e350cf0e57
  creationTimestamp: '2024-04-03T07:29:21Z'
  namespace: default
  ownerReferences:
    - apiVersion: cdi.kubevirt.io/v1beta1
      kind: DataVolume
      name: ovirt-dv
      uid: f7963dbd-3d39-455e-ae96-e43033d45367
      controller: true
      blockOwnerDeletion: true
  finalizers:
  labels:
    alerts.k8s.io/KubePersistentVolumeFillingUp: disabled
    app.kubernetes.io/part-of: hyperconverged-cluster
    app.kubernetes.io/version: 4.15.0
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: '13756873279'
  storageClassName: nfs-csi
  volumeMode: Filesystem
  dataSource:
    apiGroup: forklift.konveyor.io
    kind: OvirtVolumePopulator
    name: ovirt-pop
  dataSourceRef:
    apiGroup: forklift.konveyor.io
    kind: OvirtVolumePopulator
    name: ovirt-pop
status:
  phase: Pending
```

This is what a PVC that's being populated looks like

akalenyu:

/cc @alromeros
Looks like this failure made it here as well (don't think this PR does anything to help trigger it)

akalenyu:
/test all
just getting some cycles in to potentially identify other issues with the rpm bump. not necessarily related to the logic in this PR

hack/build/resource-generator.sh
```diff
@@ -45,6 +45,7 @@ function generateResourceManifest() {
             -apiserver-image="${DOCKER_PREFIX}/${APISERVER_IMAGE_NAME}:${DOCKER_TAG}" \
             -uploadproxy-image="${DOCKER_PREFIX}/${UPLOADPROXY_IMAGE_NAME}:${DOCKER_TAG}" \
             -uploadserver-image="${DOCKER_PREFIX}/${UPLOADSERVER_IMAGE_NAME}:${DOCKER_TAG}" \
+            -ovirt-populator-image="${OVIRT_POPULATOR_IMAGE_NAME}" \
```
akalenyu, line 48: `OVIRT_POPULATOR_IMAGE_NAME` should be "cdi-importer" at this point but that won't resolve to a proper image, no? what am I missing?

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,584 @@
...
+
+// ForkliftPopulatorReconciler members
+type ForkliftPopulatorReconciler struct {
+	ReconcilerBase
+	importerImage       string
+	ovirtPopulatorImage string
+}
+
+// NewForkliftPopulator creates a new instance of the forklift controller.
+func NewForkliftPopulator(
```
akalenyu, line 58: Do we want to add a few e2e assertions that the populator gets used instead of legacy?

hack/build/resource-generator.sh
bennyz: yeah, for some reason the generated manifest still had the original image name, and was not updated. In my tests I changed the env var manually in openshift because HCO didn't properly update the manifests, not sure why though. Anyway, this part is a bit tricky for downstream

akalenyu: > yeah, for some reason the generated manifest still had the original image name

The CI runs fine, which makes me question whether the forklift populator is in play at all?
I think one thing that is missing for sure is `DOCKER_PREFIX` here, so the image should be broken

bennyz: The controller doesn't run so nothing pulls this image I guess

akalenyu: Doesn't the functionality in this PR get exercised in e2e?

akalenyu: I am fine if not, can do in a follow up (if at all, it's just what we did with our internal flows, we converted existing DV operations to create populator CRs behind the scenes)

bennyz: > Doesn't the functionality in this PR get exercised in e2e? i.e. DataVolumes that used to do legacy now create the populator CRs?

I haven't added anything for it, so the controller probably runs, but the image won't be pulled because it doesn't invoke anything

I can try to add it later unless it's trivial, but doesn't look like it

bennyz: I've added some code to normalize the name of the image now

akalenyu: Can't we just have the same treatment for it as the others?
`-uploadserver-image="${DOCKER_PREFIX}/${UPLOADSERVER_IMAGE_NAME}:${DOCKER_TAG}" \`

bennyz: it wouldn't be an issue with the downstream image? assuming it would receive a full external image URI

if not, then sure, we can do that

akalenyu: I think if this works today for upload image etc should be fine for this one as well

bennyz: yes, but they are built as part of this project no?
my concern is that if we set OVIRT_IMAGE_POPULATOR=https://internal.registry/mtv/ovirt-populator:v4.16

then the URI would end up messed up, but I don't know what the downstream build process would look like so perhaps it's ok

akalenyu: same here, maybe @tiraboschi (hi!) knows, anyhow - we can always fix the scripts to work with downstream

bennyz: ok, then I'll revert it back

bennyz:
Retested all the recent changes using the forklift migration flow, no issues seen

pkg/controller/populators/forklift-populator.go
akalenyu: So these populators are not replacing legacy DV flows, rather, they're new functionality that did not exist in CDI before?
If that's true, I guess this is what makes it tricky to test here?

bennyz: yes, you could say the ovirt populator can replace the imageio data source, but there is no feature parity. The populator does not support checkpoints like the DV datasource for example, so it will still be used for warm migration from ovirt

I would be interested in having tests for this other than unit tests, but given that it doesn't seem trivial to do and we need some buffer to test this with forklift, I'd do it separately

bennyz:
@akalenyu was the logic reviewed?

hack/update-codegen.sh
```diff
@@ -42,14 +42,18 @@ chmod u+x "${CODEGEN_PKG}/generate-internal-groups.sh"
...
-    "core:v1alpha1 core:v1beta1 upload:v1beta1" \
+    "core:v1alpha1 core:v1beta1 upload:v1beta1 forklift:v1beta1" \
     --go-header-file ${SCRIPT_ROOT}/hack/custom-boilerplate.go.txt
 
 echo "Generating swagger doc"
 swagger-doc -in ${SCRIPT_ROOT}/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1alpha1/types.go
 
 swagger-doc -in ${SCRIPT_ROOT}/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
 swagger-doc -in ${SCRIPT_ROOT}/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/upload/v1beta1/types.go
+swagger-doc -in ${SCRIPT_ROOT}/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/upload/v1beta1/types.go
```
akalenyu, line 53: duplicated line?

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
```diff
@@ -0,0 +1,86 @@
...
+	Status OvirtVolumePopulatorStatus `json:"status"`
+}
+
+// OvirtVolumePopulatorSpec is the spec of the OvirtVolumePopulator CR
+type OvirtVolumePopulatorSpec struct {
+	EngineURL        string `json:"engineUrl"`
+	EngineSecretName string `json:"engineSecretName"`
+	DiskID           string `json:"diskId"`
+	// The network attachment definition that should be used for disk transfer.
+	TransferNetwork *core.ObjectReference `json:"transferNetwork,omitempty"`
```
akalenyu, line 31: Are we consistent with kubevirt/kubevirt wrt specifying the network here?
If that's a NetworkAttachmentDefinition, I believe kubevirt uses plain string nad_name?

akalenyu (commented):
> @akalenyu was the logic reviewed?

Yeah the logic looks fine, my main worry was/is e2e, which I understand is not trivial.
Do we have a plan for setting up testing for this? any help needed with setting up the test containers?

/cc @mhenriks
on API & populator related bits to make sure this is consistent with other populators

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
bennyz: forklift used the term transfer network, so that's the reason for the name, I have no problem changing it to nad_name

hack/update-codegen.sh
bennyz: yes, will fix

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
akalenyu: the name is fine I meant the type

bennyz: ah ok, yeah it's the same type used in https://github.com/kubev2v/forklift/blob/main/pkg/apis/forklift/v1beta1/plan.go#L44

anyway, it's good that you've mentioned it because I realize I don't actually use it in the controller, which I should

akalenyu: I meant if it's consistent with kubevirt - https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#multus

bennyz:
> > @akalenyu was the logic reviewed?
> 
> Yeah the logic looks fine, my main worry was/is e2e, which I understand is not trivial. Do we have a plan for setting up testing for this? any help needed with setting up the test containers?
> 
> /cc @mhenriks on API & populator related bits to make sure this is consistent with other populators
We will have e2e testing via forklift. For oVirt perhaps we can use fakeovirt (https://github.com/kubev2v/fakeovirt) which we use in our CI. For openstack we use a real openstack setup with packstack, it adds a significant amount to the run and consumes a lot of resources, so I don't think it'd viable here

cmd/openstack-populator/openstack-populator_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	server := httptest.NewUnstartedServer(mux)
+	server.Listener = listener
+
+	server.Start()
+
+	return server, identityServerURL, port, nil
+}
+
+func TestPopulate(t *testing.T) {
```
akalenyu, line 97: I see the test is being exercised but you probably want it integrated with ginkgo like the others https://github.com/kubevirt/containerized-data-importer/blob/0c15a1c96cefad360235cb56473240ce04b7d18b/cmd/cdi-cloner/clone-source_suite_test.go#L1-L13

akalenyu (commented):
Minor comment which we can address later, won't block over it
Otherwise things look good, thanks!

/approve

akalenyu:
For linter fixes, take a look at the PR description here
https://github.com/kubevirt/containerized-data-importer/pull/3179

cmd/openstack-populator/openstack-populator_test.go
bennyz: changed to ginkgo

bennyz:
> For linter fixes, take a look at the PR description here #3179

done, thanks

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+	case "OvirtVolumePopulator":
+		return &v1beta1.OvirtVolumePopulator{}, nil
+	case "OpenstackVolumePopulator":
+		return &v1beta1.OpenstackVolumePopulator{}, nil
+	default:
+		return nil, fmt.Errorf("unknown populator type %T", pvc.Spec.DataSourceRef.Kind)
+	}
+}
+
+func isPVCForkliftKind(pvc *corev1.PersistentVolumeClaim) bool {
```
awels, line 242: So I don't understand this function, what I am expecting based on the name is that we check the `dataSourceRef` to have the right api group, and that it has a name for the populator, and that the Kind is in the supportedPopulators map. But when I look at this:

```go
if dataSourceRef.APIGroup != nil &&
  *dataSourceRef.APIGroup == apiGroup &&
  dataSourceRef.Name != "" {
    return false
}
```
It will pass if the APIGroup is nil or if the APIGroup doesn't match. And it fails if the name is a value. Is that the correct behavior we want? Seems strange to me.

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+	if err := c.Watch(source.Kind(mgr.GetCache(), &corev1.PersistentVolumeClaim{}), handler.EnqueueRequestsFromMapFunc(
+		func(_ context.Context, obj client.Object) []reconcile.Request {
+			pvc := obj.(*corev1.PersistentVolumeClaim)
+			if isPVCForkliftKind(pvc) {
+				pvcKey := types.NamespacedName{Namespace: pvc.Namespace, Name: pvc.Name}
+				return []reconcile.Request{{NamespacedName: pvcKey}}
+			}
+
+			// TODO(benny) check dataSourceRef PVC prime
+			if isPVCForkliftKind(pvc) {
```
awels, line 108: How is this ever getting called. If the condition is true, the above if will return out of the calling function.

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+	}
+
+	return nil
+}
+
+// Reconcile the reconcile loop for the PVC with DataSourceRef of OvirtVolumePopulator or OpenstackVolumePopulator kind
+func (r *ForkliftPopulatorReconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithValues("PVC", req.NamespacedName)
+	log.V(1).Info("reconciling Forklift PVCs")
+	return r.reconcile(req, r, log)
```
awels, line 147: Why not simply pass the context passed in, to the reconcile function. In there you are creating new `context.TODO()` where you can just pass the existing context around.

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+
+	dataSourceRef := pvc.Spec.DataSourceRef
+
+	if !isPVCForkliftKind(pvc) {
+		log.V(1).Info("reconciled unexpected PVC, ignoring")
+		return nil, nil
+	}
+
+	// TODO: Remove this check once we support cross-namespace dataSourceRef
+	if dataSourceRef.Namespace != nil {
```
awels, line 205: What about if the namespace is the same namespace as the PVC, that would not be cross namespace right?

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+func (r *ForkliftPopulatorReconciler) getImportPod(pvc *corev1.PersistentVolumeClaim, importPodName string) (*corev1.Pod, error) {
+	pod := &corev1.Pod{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: importPodName, Namespace: pvc.GetNamespace()}, pod); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+		return nil, nil
+	}
+
+	if !metav1.IsControlledBy(pod, pvc) && !cc.IsImageStream(pvc) {
```
awels, line 414: I don't think we care about image streams do we?

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+		return nil, errors.Errorf("Pod is not owned by PVC")
+	}
+	return pod, nil
+}
+
+func (r *ForkliftPopulatorReconciler) createPopulatorPod(pvcPrime, pvc *corev1.PersistentVolumeClaim) error {
+	var rawBlock bool
+	if nil != pvc.Spec.VolumeMode && corev1.PersistentVolumeBlock == *pvc.Spec.VolumeMode {
```
awels, line 422: I am surprised the linter is not unhappy with this. Can we flip the arguments to `pvc.Spec.VolumeMode != nil && *pvc.Spec.VolumeMode == corev1.PersistentVolumeBlock`?

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,593 @@
...
+		return reqs
+	}
+
+	if err := c.Watch(source.Kind(mgr.GetCache(), sourceType),
+		handler.EnqueueRequestsFromMapFunc(mapDataSourceRefToPVC),
+	); err != nil {
+		return err
+	}
+
+	return nil
```
awels, line 140: Maybe also watch pods that have an ownerRef of a PVC that has the right APIGroup/Kind? That way there is no need to requeue after creating the pod. Once the pod status updates it should trigger a reconcile here.

awels (changes requested):

pkg/controller/populators/forklift-populator.go
bennyz: I took this from https://github.com/kubevirt/containerized-data-importer/blob/55dd0d0110754e17af3473a2c5622dcc10c90b2f/pkg/controller/populators/populator-base.go#L328

pkg/controller/populators/forklift-populator.go
bennyz: you're right and I couldn't understand how it worked until I saw that's because of apiGroup const

pkg/controller/populators/forklift-populator.go
bennyz: yeah it should watch the prime pvc

awels: But if you look at the code, the if statement above is identical in the check. And it returns out of the function. So if it matches, the second if is never called, and if it doesn't match, the second if will also never match.

pkg/controller/populators/forklift-populator.go
bennyz: that's how it's done in all places I've seen https://github.com/kubevirt/containerized-data-importer/blob/55dd0d0110754e17af3473a2c5622dcc10c90b2f/pkg/controller/populators/populator-base.go#L278

But I've changed

pkg/controller/populators/forklift-populator.go
bennyz: yes, I took it from the import-populator, dropped

pkg/controller/populators/forklift-populator.go
awels: I think that is wrong as well, it should check if != nil and not equals the namespace of the PVC.

pkg/controller/populators/forklift-populator.go
bennyz: yes, I meant that it's going to be changed to watch the prime pvc like the other populators do

pkg/controller/populators/forklift-populator.go
bennyz: Added, @alromeros can you confirm?

pkg/controller/populators/forklift-populator.go
bennyz: added, although I need to check this some more, not entirely sure I got it right

mhenriks:
/hold

Looks good but would like to do a deep dive before we merge this.  Thanks for your patience @bennyz

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,634 @@
...
+}
+
+func isPVCForkliftKind(pvc *corev1.PersistentVolumeClaim) bool {
+	dataSourceRef := pvc.Spec.DataSourceRef
+	if dataSourceRef == nil {
+		return false
+	}
+
+	if (dataSourceRef.APIGroup != nil && *dataSourceRef.APIGroup != apiGroup) ||
+		dataSourceRef.Name == "" {
```
mhenriks, line 279: yeah should also check `dataSourceRef.Kind`

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
```diff
@@ -0,0 +1,85 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:shortName={ovvp,ovvps}
+type OvirtVolumePopulator struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec OvirtVolumePopulatorSpec `json:"spec"`
+	// +optional
+	Status OvirtVolumePopulatorStatus `json:"status"`
```
mhenriks, line 21: For all fields that are optional should probably add "omitempty" to json tags as well.  I don't think the json generation code looks at "// + optional" think that's just code gen tools

mhenriks (commented):
looks great, couple minor changes for your consideration

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,634 @@
...
+	); err != nil {
+		return err
+	}
+
+	mapDataSourceRefToPVC := func(_ context.Context, obj client.Object) (reqs []reconcile.Request) {
+		var pvcs corev1.PersistentVolumeClaimList
+		matchingFields := client.MatchingFields{
+			dataSourceRefField: getPopulatorIndexKey(apiGroup, sourceKind, obj.GetNamespace(), obj.GetName()),
+		}
+		if err := mgr.GetClient().List(context.TODO(), &pvcs, matchingFields); err != nil {
```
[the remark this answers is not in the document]
bennyz: changed

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,634 @@
...
+			}
+
+			// Get pod owner reference
+			owner := metav1.GetControllerOf(pod)
+			if owner == nil {
+				return nil
+			}
+
+			pvcPrime := &corev1.PersistentVolumeClaim{}
+			err := mgr.GetClient().Get(context.TODO(), types.NamespacedName{Namespace: pod.Namespace, Name: owner.Name}, pvcPrime)
```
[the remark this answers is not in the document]
bennyz: changed

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,634 @@
...
+	}
+
+	if err := r.client.Create(context.TODO(), &pod); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func makePopulatePodSpec(pvcPrimeName, secretName string) corev1.PodSpec {
```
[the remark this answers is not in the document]
bennyz: ack, added

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,634 @@
...
+	if owner == nil || owner.Kind != "PersistentVolumeClaim" {
+		return false
+	}
+
+	populatorKind := pvc.Annotations[cc.AnnPopulatorKind]
+	return populatorKind == "forklift"
+}
+
+func (r *ForkliftPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
+	pvcCopy := pvc.DeepCopy()
```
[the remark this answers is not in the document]
bennyz: dropped

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,634 @@
...
+		}); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *ForkliftPopulatorReconciler) updatePVCPrime(pvc, pvcPrime *corev1.PersistentVolumeClaim) error {
+	pvcCopy := pvc.DeepCopy()
```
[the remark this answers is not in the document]
bennyz: dropped

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
bennyz: added

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,633 @@
...
+				},
+			},
+		},
+		SecurityContext: &corev1.PodSecurityContext{
+			FSGroup: ptr.To[int64](107),
+			SeccompProfile: &corev1.SeccompProfile{
+				Type: corev1.SeccompProfileTypeRuntimeDefault,
+			},
+		},
+		RestartPolicy: corev1.RestartPolicyNever,
```
bennyz, line 591: Should be `RestartPolicyOnFailure`

mhenriks:
@awels you good with this?  I think I'm good

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
```diff
@@ -0,0 +1,85 @@
...
+	Status OpenstackVolumePopulatorStatus `json:"status,omitempty"`
+}
+
+// OpenstackVolumePopulatorSpec is the spec of the OpenstackVolumePopulator CR
+type OpenstackVolumePopulatorSpec struct {
+	IdentityURL string `json:"identityUrl"`
+	SecretRef   string `json:"secretRef"`
+	ImageID     string `json:"imageId"`
+	// The network attachment definition that should be used for disk transfer.
+	TransferNetwork string `json:"transferNetwork,omitempty"`
```
awels, line 70: Since this is an optional field, I believe this should be a `*string` so we can distinguish between a blank value and not set. In this case not a huge deal, but in general optional fields should be pointers

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
```diff
@@ -0,0 +1,85 @@
...
+	Status OvirtVolumePopulatorStatus `json:"status,omitempty"`
+}
+
+// OvirtVolumePopulatorSpec is the spec of the OvirtVolumePopulator CR
+type OvirtVolumePopulatorSpec struct {
+	EngineURL string `json:"engineUrl"`
+	SecretRef string `json:"secretRef"`
+	DiskID    string `json:"diskId"`
+	// The network attachment definition that should be used for disk transfer.
+	TransferNetwork string `json:"transferNetwork,omitempty"`
```
awels, line 30: Since this is an optional field, I believe this should be a *string so we can distinguish between a blank value and not set. In this case not a huge deal, but in general optional fields should be pointers

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,644 @@
...
+			if err != nil {
+				return nil
+			}
+
+			// Check if the owner is a PVC prime
+			if !isPVCPrimeForkliftKind(pvcPrime) && pvcPrime.DeletionTimestamp == nil {
+				return nil
+			}
+
+			return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}}}
```
awels, line 143: If I am reading the reconciler right, it is expecting to get pvc name, in particular the real PVC, not the prime PVC. This returns the pod name, I believe you should look up the owner of the pvcPrime, and return that name.

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,644 @@
...
+		r.recorder.Eventf(pvc, corev1.EventTypeNormal, importSucceeded, messageImportSucceeded, pvc.Name)
+	}
+
+	if pod.DeletionTimestamp == nil && cc.ShouldDeletePod(pvcPrime) {
+		if err := r.client.Delete(context.TODO(), &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      podName,
+				Namespace: pvc.GetNamespace(),
+			},
+		}); err != nil {
```
awels, line 372: Minor optimization here, you can check if the err is not found, and if so the pod is already gone, and you can just return nil since what we wanted already happened. It is not a big deal, since it will simply reconcile again, and find out it is gone and be happy, just saves a reconcile with a simple check.

awels (commented):
A few small things that should be easy to fix.

pkg/controller/populators/forklift-populator.go
```diff
@@ -0,0 +1,644 @@
...
+		if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvc); err != nil {
+			return reconcile.Result{}, err
+		}
+	default:
+		//Should never happen
+		return reconcile.Result{}, fmt.Errorf("unknown pod phase %s", phase)
+	}
+
+	if err := r.updatePVCPrime(pvc, pvcPrimeCopy); err != nil {
+		return reconcile.Result{}, nil
```
akalenyu, line 359: why change to `return reconcile.Result{}, nil` instead of returning the err?

bennyz: forgot to revert back after experimentation

akalenyu:
@bennyz could you please elaborate a little about the recent changes, specifically this diff:
https://github.com/kubevirt/containerized-data-importer/compare/e8a2b5dbcfc98af88e88605de89877aa79fe34bc..e688d25817e9aca39c08ba5c1077b291b189e053

bennyz:
> @bennyz could you please elaborate a little about the recent changes, specifically this diff: https://github.com/kubevirt/containerized-data-importer/compare/e8a2b5dbcfc98af88e88605de89877aa79fe34bc..e688d25817e9aca39c08ba5c1077b291b189e053

Some of this will change in the next revision (based on the comments I've received), but this is basically two parts:
1. Add the restart count, which also required some fixes in the UT
2. After adding the restart count annotation, I noticed there is race with the updates of the annotation (the last update was sometimes missed), and I've changed the reconciliation return to match the other populator controller

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1/types.go
bennyz: done

pkg/controller/populators/forklift-populator.go
bennyz: yup, changed

pkg/controller/populators/forklift-populator.go
bennyz: added

bennyz:
can someone rerun tests?

awels (approved):

PR #2985, opened 2023-11-14
Prevent upload populator from hot looping prime annotation

awels:
This hot loop was causing the upload populator to constantly add and remove the annotation causing the node controller to always be outdated and never update the pvc to be bound causing the upload test to fail randomly.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/pvc.go
+++ b/tests/utils/pvc.go
@@ -213,7 +213,7 @@ func NewBlockPVCDefinition(pvcName string, size string, annotations, labels map[
 func WaitForPersistentVolumeClaimPhase(clientSet *kubernetes.Clientset, namespace string, phase k8sv1.PersistentVolumeClaimPhase, pvcName string) error {
 	err := wait.PollImmediate(pvcPollInterval, pvcPhaseTime, func() (bool, error) {
 		pvc, err := clientSet.CoreV1().PersistentVolumeClaims(namespace).Get(context.TODO(), pvcName, metav1.GetOptions{})
-		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: Checking PVC phase: %s\n", string(pvc.Status.Phase))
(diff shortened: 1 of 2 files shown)


tests/utils/pvc.go
```diff
@@ -213,7 +213,7 @@ func NewBlockPVCDefinition(pvcName string, size string, annotations, labels map[
 func WaitForPersistentVolumeClaimPhase(clientSet *kubernetes.Clientset, namespace string, phase k8sv1.PersistentVolumeClaimPhase, pvcName string) error {
 	err := wait.PollImmediate(pvcPollInterval, pvcPhaseTime, func() (bool, error) {
 		pvc, err := clientSet.CoreV1().PersistentVolumeClaims(namespace).Get(context.TODO(), pvcName, metav1.GetOptions{})
-		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: Checking PVC phase: %s\n", string(pvc.Status.Phase))
+		fmt.Fprintf(ginkgo.GinkgoWriter, "INFO: Checking PVC phase: %s, resource version %s\n", string(pvc.Status.Phase), pvc.ResourceVersion)
```
akalenyu, line 216: Should we add a `Consistently()` block to the failing test, that ensures the PVC doesn't get updated from a certain point onwards?

akalenyu (commented):
Looks good, just worried this doesn't protect us completely from code that re-introduces these hotloops

tests/utils/pvc.go
awels: Can you elaborate, I don't quite understand what you mean

akalenyu: I mean we are missing a functional test that ensures we are not hotlooping PVC updates

awels: Yeah, we are, just need to identify how to reliably test that. As we saw the revisionVersion goes up relatively fast even if we are not hot looping

PR #3017, opened 2023-11-30
Use direct io with qemu-img convert if pod is OOMKilled

akalenyu:
For a while now we have been switching between cache=none (direct io) and cache=writeback (page cache) for qemu-img's writes.

We have settled on cache=writeback for quite some time since https://github.com/kubevirt/containerized-data-importer/pull/1919,
however, this has proven to be problematic;
Our pod's live in a constrained memory environment (default limit 600M).
cgroupsv1 compares utilization of page cache vs the host's dirty_ratio.
This means that on a standard system (30% dirty ratio) pages only get forced to disk at 0.3 * HOST_MEM (basically never),
easily triggering OOM on hosts with lots of free memory.

cgroupsv2 does come to the rescue here:
- It considers dirty_ratio against CGROUP_MEM (forced to disk much faster, 0.3 * 600M)
- Has a new memory.high knob that throttles instead of OOM killing
Sadly, k8s is yet to capitalize on memory.high since this feature is still alpha:
https://kubernetes.io/blog/2023/05/05/qos-memory-resources/
Leaving us with no way to avoid frequent OOMs.

This PR introduces a cache=trynone behavior, in which writes bypass page cache (cache=none) if the target supports it,
otherwise, fall back to cache=writeback (use page cache).
This env var will automatically be set in case the pod fails once with an OOMKilled.

Note, there have previously been [issues](https://bugzilla.redhat.com/show_bug.cgi?id=2017262) where target did not support O_DIRECT. A quick example is tmpfs (ram-based). This is why we must have a fallback in case cache=none is not an option.

Fixes #

**Release note**:

```release-note
Use direct io (cache=none) with qemu-img convert if target supports it
```

### DIFF
--- a/hack/build/bazel-docker.sh
+++ b/hack/build/bazel-docker.sh
@@ -121,6 +121,8 @@ fi
 if [ -n "$DOCKER_CA_CERT_FILE" ]; then
     volumes="$volumes -v ${DOCKER_CA_CERT_FILE}:${DOCKERIZED_CUSTOM_CA_PATH}:ro,z"
 fi
+# add tmpfs path for testing purposes
+volumes="$volumes --tmpfs /mnt/cditmpfs"
 
 # Ensure that a bazel server is running
 if [ -z "$(${CDI_CRI} ps --format '{{.Names}}' | grep ${BAZEL_BUILDER_SERVER})" ]; then
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -101,6 +101,7 @@ type importPodEnvVar struct {
 	certConfigMapProxy string
 	extraHeaders       []string
 	secretExtraHeaders []string
+	cacheMode          string
 }
 
 type importerPodArgs struct {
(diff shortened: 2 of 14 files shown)


akalenyu:
/hold
There is not much appetite to change the cache mode yet again or add an API to choose it.  We will follow the k8s API referenced above since it will address the issue properly.

germanovm:
And here another place that has the same issues, this 'dd' is writing to memory to preallocate the disk. So its not just the qemu-img family of commands doing it. 

https://github.com/kubevirt/containerized-data-importer/blob/25dcbab2358493c02145ee5f5d34a2ec9f9ca7fc/pkg/image/qemu.go#L297

akalenyu:
> And here another place that has the same issues, this 'dd' is writing to memory to preallocate the disk. So its not just the qemu-img family of commands doing it.
> 
> https://github.com/kubevirt/containerized-data-importer/blob/25dcbab2358493c02145ee5f5d34a2ec9f9ca7fc/pkg/image/qemu.go#L297

Added. thanks!

pkg/controller/import-controller.go
```diff
@@ -378,25 +379,37 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
...
 		log.V(1).Info("Pod requires scratch space, terminating pod, and restarting with scratch space", "pod.Name", pod.Name)
 	}
+	podModificationsNeeded := scratchSpaceRequired
 
-	if pod.Status.ContainerStatuses != nil &&
-		pod.Status.ContainerStatuses[0].State.Terminated != nil &&
-		pod.Status.ContainerStatuses[0].State.Terminated.ExitCode > 0 {
-		log.Info("Pod termination code", "pod.Name", pod.Name, "ExitCode", pod.Status.ContainerStatuses[0].State.Terminated.ExitCode)
-		r.recorder.Event(pvc, corev1.EventTypeWarning, ErrImportFailedPVC, pod.Status.ContainerStatuses[0].State.Terminated.Message)
+	if statuses := pod.Status.ContainerStatuses; statuses != nil {
```
mhenriks, line 384: I think `len(statuses)` > 0 would be better than `statuses != nil`

hack/build/bazel-docker.sh
```diff
@@ -121,6 +121,8 @@ fi
 if [ -n "$DOCKER_CA_CERT_FILE" ]; then
     volumes="$volumes -v ${DOCKER_CA_CERT_FILE}:${DOCKERIZED_CUSTOM_CA_PATH}:ro,z"
 fi
+# add tmpfs path for testing purposes
+volumes="$volumes --tmpfs /mnt/cditmpfs"
```
mhenriks, line 125: were you planning to use this in a func test?

mhenriks (commented):
What would it take to func test this?  Is that a reasonable request?

hack/build/bazel-docker.sh
akalenyu: No, it's used in a unit test container

akalenyu:
> What would it take to func test this? Is that a reasonable request?

It'll be clunky and racy, we'd have to trick the CRI to return this reason. So I'm thinking privileged container for node access at least.

I am not sure it'll boost our confidence significantly over the unit test, do you think the unit test isn't covering this?

akalenyu:
> > > What would it take to func test this? Is that a reasonable request?
> > 
> > 
> > It'll be clunky and racy, we'd have to trick the CRI to return this reason. So I'm thinking privileged container for node access at least.
> > I am not sure it'll boost our confidence significantly over the unit test, do you think the unit test isn't covering this?
> 
> Was assuming you could create a tmpfs storageclass and hostpath PV on the tmpfs volume. Don't see how that would be racy but maybe there are other issues?

Yeah the only way the logic added gets exercised is if an OOMKilled appears on the pod

pkg/image/qemu.go
```diff
@@ -137,6 +141,33 @@ func convertToRaw(src, dest string, preallocate bool) error {
 	return nil
 }
 
+func getCacheMode(path string) (string, error) {
+	if os.Getenv(common.CacheMode) != common.CacheModeTryNone {
```
mhenriks, line 145: Does this package look at environment variables anywhere else?  if not then maybe it wiould be better to have this as a function param?

akalenyu: Yeah better to keep env checking in cmd/. Posted the change

PR #3018, opened 2023-12-01
Update output-bazel-arch.sh to be inclusive of s390x

### LINKED ISSUE #3010: enable build-the-builder script to build container that will work for s390x
Built in to the CDI builder helper container is a script `output-bazel-arch.sh` which excludes support for non-ARM non-x86 architectures. Also, the Dockerfile assumes there is a pre-built binary available in https://github.com/bazelbuild/bazel/releases/download which in the case of s390x architecture there is not. 

 The 'new feature' here would be to add s390x to the output-bazel-arch.sh list of supported architecture, e.g:

```
PLATFORM=$(uname -m)
case ${PLATFORM} in
x86_64* | i?86_64* | amd64*)
    ARCH="x86_64"
    ;;
s390x)
    ARCH="s390x"
    ;;
aarch64* | arm64*)
    ARCH="arm64"
    ;;
*)
    echo "invalid Arch, only support x86_64, aarch64"
    exit 1
    ;;
esac

echo $ARCH

```

and obtain the pre-built bazel binary for s390x from an alternative release download repository

An alternative would be to provide the script to build bazel on the local host before copying it to the builder container

cfilleke:
Remove key obstacle to s390x support for KubeVirt CDI Builder container 

 s390x architecture enablement in the CDI build-the-builder container. Removes one obstacle.

Partially Fixes #3010

Will still need a way to download a trusted up-to-date bazel executable for s390x into the CDI builder container and integration of all the rc files and rpm updates in order to have a fully functioning KubeVirt CDI build system. Please see https://github.com/bazelbuild/bazel/issues/10408

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/docker/builder/output-bazel-arch.sh
+++ b/hack/build/docker/builder/output-bazel-arch.sh
@@ -3,11 +3,14 @@ case ${PLATFORM} in
 x86_64* | i?86_64* | amd64*)
     ARCH="x86_64"
     ;;
+s390x)
+    ARCH="s390x"
+    ;;
 aarch64* | arm64*)
     ARCH="arm64"
     ;;
 *)
-    echo "invalid Arch, only support x86_64, aarch64"
+    echo "invalid Arch, only support x86_64, aarch64, s390x"
     exit 1
     ;;
 esac


awels:
Thanks for the PR, looks like we need some support from bazel to get this over the finish line and supported. We will monitor, but I would like to hold off on this until we can get bazel to cooperate better.

cfilleke:
/close 

this issue will be addressed in the s390x cross-compile enablement, PR coming soon.

PR #3023, opened 2023-12-04
Update kubevirtci, 1.29 lanes

akalenyu:
May help with nfs lane flaking:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3017/pull-containerized-data-importer-e2e-nfs/1731266718799499264
Looks like the nfs server is working but mount commands from nfs csi against it are timing out.

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
> @akalenyu do we still want to go forward with this? Looks like nfs still failed. Maybe there is a newer (better!) kubevirtci since this is a little old?

So the nfs failure here is the same one that occurred in kubevirt https://github.com/kubevirt/kubevirt/pull/10815 and
motivated me to propose https://github.com/kubevirt/containerized-data-importer/pull/3017.

I still don't know why nfs writing started to consistently breach the dirty_ratio
but I think the problem sits somewhere around the fsync calls to writeout to nfs being slow/failing

PR #3024, opened 2023-12-04
[WIP] Try tackle log flakes by delaying log rotation

akalenyu:
Ultimately we probably want to shift away from tests that depend on logs

Fixes #

**Release note**:

```release-note
NONE
```


xpivarc:
> @brianmcarey @xpivarc ever had this kind of problem in kubevirt/kubevirt? where logs "disappear" from debugging manifest

I do not think so.

akalenyu:
fyi @tiraboschi

tiraboschi:
> fyi @tiraboschi

The speed of the serial console is capped at 115200 bps, so in order to generate 20MiB we need to flood the serial console for 1456 seconds (24 minutes).
I think we can definitively exclude we are reaching this limit for the guest console logs during our tests.

PR #2384, opened 2022-08-09
[WIP] debug flake where cross namespace finalizer remains

akalenyu:
Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

**What this PR does / why we need it**:
Should probably identify which test this is

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #3029, opened 2023-12-12
Feature Gate and annotation for PVCs to be "adopted" by DataVolumes

mhenriks:

In many DR/backup scenarios, a PVC may exist before a DataVolume.  Usually that results in an error when the DataVolume is created.  This PR provides two new mechanisms to avoid that error and allow the DataVolume controller to "adopt" an existing PVC.  Adoption of a PVC means that the DataVolume will immediately go to Succeeded state and not attempt to write any data to the PVC.

**DataVolumeClaimAdoption** feature gate:  This is a global setting to configure claim adoption for all PVCs in the system

**cdi.kubevirt.io/allowClaimAdoption** PVC annotation: This overrides any feature gate setting and and allows for one-off claim adoption.  It is also possible to add this annotation to PVCs indirectly when creating a DataVolume (PVCs inherit DataVolume annotations).  But there are some subtleties there.  To avoid a chicken/egg problem must be applied after the new DataVolume has succeeded.

Fixes #

**Release note**:

```release-note
DataVolume supports PVC adoption via DataVolumeClaimAdoption feature gate and cdi.kubevirt.io/allowClaimAdoption annotation on PVC
```

### DIFF
--- a/cmd/cdi-apiserver/apiserver.go
+++ b/cmd/cdi-apiserver/apiserver.go
@@ -32,8 +32,10 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	aggregatorclient "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/apiserver"
 	cdiclient "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"kubevirt.io/containerized-data-importer/pkg/common"
@@ -119,6 +121,15 @@ func main() {
 
 	snapClient := snapclient.NewForConfigOrDie(cfg)
 
+	cluster, err := cluster.New(cfg)
+	if err != nil {
+		klog.Fatalf("Unable to create controller runtime cluster: %v\n", errors.WithStack(err))
+	}
+
+	if err := cdiv1.AddToScheme(cluster.GetScheme()); err != nil {
+		klog.Fatalf("Unable to add to scheme: %v\n", errors.WithStack(err))
+	}
+
 	ctx := signals.SetupSignalHandler()
 
 	authConfigWatcher, err := apiserver.NewAuthConfigWatcher(ctx, client)
@@ -147,6 +158,7 @@ func main() {
 		aggregatorClient,
 		cdiClient,
 		snapClient,
+		cluster.GetClient(),
 		authorizor,
 		authConfigWatcher,
 		cdiConfigTLSWatcher,
@@ -155,6 +167,12 @@ func main() {
 	if err != nil {
 		klog.Fatalf("Upload api failed to initialize: %v\n", errors.WithStack(err))
 	}
+
+	go func() {
+		if err := cluster.Start(ctx); err != nil {
+			klog.Errorf("cluster failed: %v\n", errors.WithStack(err))
+		}
+	}()
 
 	go func() {
 		if err := certWatcher.Start(ctx.Done()); err != nil {
--- a/pkg/apiserver/webhooks/datavolume-validate_test.go
+++ b/pkg/apiserver/webhooks/datavolume-validate_test.go
@@ -265,7 +265,7 @@ var _ = Describe("Validating Webhook", func() {
 			Expect(resp.Allowed).To(BeTrue())
 		})
 
-		It("should accept DataVolume with unbound PVC and adoption featurgate set", func() {
+		It("should reject DataVolume with unbound PVC and adoption featurgate set", func() {
 			dataVolume := newHTTPDataVolume("testDV", "http://www.example.com")
 			pvc := &corev1.PersistentVolumeClaim{
 				ObjectMeta: metav1.ObjectMeta{
--- a/pkg/apiserver/webhooks/datavolume-validate_test.go
+++ b/pkg/apiserver/webhooks/datavolume-validate_test.go
@@ -235,7 +235,7 @@ var _ = Describe("Validating Webhook", func() {
 			Expect(resp.Allowed).To(BeFalse())
 		})
 
-		It("should reject DataVolume with unbound PVC and adoption annotation", func() {
+		It("should accept DataVolume with unbound PVC and adoption annotation", func() {
 			dataVolume := newHTTPDataVolume("testDV", "http://www.example.com")
 			pvc := &corev1.PersistentVolumeClaim{
 				ObjectMeta: metav1.ObjectMeta{
@@ -248,7 +248,7 @@ var _ = Describe("Validating Webhook", func() {
 				Spec: *dataVolume.Spec.PVC,
 			}
 			resp := validateDataVolumeCreate(dataVolume, pvc)
-			Expect(resp.Allowed).To(BeFalse())
+			Expect(resp.Allowed).To(BeTrue())
 		})
 
 		It("should accept DataVolume with PVC and adoption featurgate set", func() {
@@ -265,7 +265,7 @@ var _ = Describe("Validating Webhook", func() {
 			Expect(resp.Allowed).To(BeTrue())
 		})
 
-		It("should reject DataVolume with unbound PVC and adoption featurgate set", func() {
+		It("should accept DataVolume with unbound PVC and adoption featurgate set", func() {
 			dataVolume := newHTTPDataVolume("testDV", "http://www.example.com")
 			pvc := &corev1.PersistentVolumeClaim{
 				ObjectMeta: metav1.ObjectMeta{
@@ -275,7 +275,7 @@ var _ = Describe("Validating Webhook", func() {
 				Spec: *dataVolume.Spec.PVC,
 			}
 			resp := validateDataVolumeCreateEx(dataVolume, []runtime.Object{pvc}, nil, nil, []string{"DataVolumeClaimAdoption"})
-			Expect(resp.Allowed).To(BeFalse())
+			Expect(resp.Allowed).To(BeTrue())
 		})
 
 		It("should reject DataVolume with PVC adoption annotation false and featuregate set", func() {
--- a/pkg/apiserver/webhooks/datavolume-validate_test.go
+++ b/pkg/apiserver/webhooks/datavolume-validate_test.go
@@ -29,6 +29,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	admissionv1 "k8s.io/api/admission/v1"
@@ -247,6 +248,49 @@ var _ = Describe("Validating Webhook", func() {
 				Spec: *dataVolume.Spec.PVC,
 			}
 			resp := validateDataVolumeCreate(dataVolume, pvc)
+			Expect(resp.Allowed).To(BeFalse())
+		})
+
+		It("should accept DataVolume with PVC and adoption featurgate set", func() {
+			dataVolume := newHTTPDataVolume("testDV", "http://www.example.com")
+			pvc := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      dataVolume.Name,
+					Namespace: dataVolume.Namespace,
+				},
+				Spec: *dataVolume.Spec.PVC,
+			}
+			pvc.Spec.VolumeName = "pv"
+			resp := validateDataVolumeCreateEx(dataVolume, []runtime.Object{pvc}, nil, nil, []string{"DataVolumeClaimAdoption"})
+			Expect(resp.Allowed).To(BeTrue())
+		})
+
+		It("should accept DataVolume with unbound PVC and adoption featurgate set", func() {
+			dataVolume := newHTTPDataVolume("testDV", "http://www.example.com")
+			pvc := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      dataVolume.Name,
+					Namespace: dataVolume.Namespace,
+				},
+				Spec: *dataVolume.Spec.PVC,
+			}
+			resp := validateDataVolumeCreateEx(dataVolume, []runtime.Object{pvc}, nil, nil, []string{"DataVolumeClaimAdoption"})
+			Expect(resp.Allowed).To(BeFalse())
+		})
+
+		It("should reject DataVolume with PVC adoption annotation false and featuregate set", func() {
+			dataVolume := newHTTPDataVolume("testDV", "http://www.example.com")
+			pvc := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      dataVolume.Name,
+					Namespace: dataVolume.Namespace,
+					Annotations: map[string]string{
+						"cdi.kubevirt.io/allowClaimAdoption": "false",
+					},
+				},
+				Spec: *dataVolume.Spec.PVC,
+			}
+			resp := validateDataVolumeCreateEx(dataVolume, []runtime.Object{pvc}, nil, nil, []string{"DataVolumeClaimAdoption"})
 			Expect(resp.Allowed).To(BeFalse())
 		})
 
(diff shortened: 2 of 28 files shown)


akalenyu:
> /retest-required

are we not worried about that failure? seems to be a fresh one for this PR
https://search.ci.kubevirt.io/?search=denied+the+request%3A++Destination+PVC+.*+already+exists&maxAge=336h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

pkg/feature-gates/feature-gates.go
```diff
@@ -14,12 +14,19 @@ import (
 const (
 	// HonorWaitForFirstConsumer - if enabled will not schedule worker pods on a storage with WaitForFirstConsumer binding mode
 	HonorWaitForFirstConsumer = "HonorWaitForFirstConsumer"
+
+	// DataVolumeClaimAdoption - if enabled will allow PVC to be adopted by a DataVolume
+	// it is not an error if PVC of sam name exists before DataVolume is created
+	DataVolumeClaimAdoption = "DataVolumeClaimAdoption"
```
mhenriks, line 20: Do we like this name?

pkg/controller/common/util.go
```diff
@@ -320,6 +320,9 @@ const (
 	AnnEventSourceKind = "cdi.kubevirt.io/events.source.kind"
 	// AnnEventSource is the source that should be related to events (namespace/name)
 	AnnEventSource = "cdi.kubevirt.io/events.source"
+
+	// AnnAllowClaimAdoption is the annotation that allows a claim to be adopted by a DataVolume
+	AnnAllowClaimAdoption = "cdi.kubevirt.io/allowClaimAdoption"
```
mhenriks, line 325: Do we like this value?  `allowClaimAdoption` does not use dotted naming like others?

pkg/controller/common/util.go
```diff
@@ -1194,6 +1197,9 @@ func CreatePvcInStorageClass(name, ns string, storageClassName *string, annotati
 		},
 	}
 	pvc.Status.Capacity = pvc.Spec.Resources.Requests.DeepCopy()
+	if pvc.Status.Phase == corev1.ClaimBound {
+		pvc.Spec.VolumeName = "pv-" + string(pvc.UID)
+	}
```
mhenriks, line 1202: This function is only used in unit tests.  Weird that it is here, right?

awels: Yeah that is weird, I guess its in this package because a bunch of other packages use it. But I think it could be in util_test.go if only tests use it.

pkg/controller/common/util.go
```diff
@@ -2088,3 +2094,30 @@ func SetPvcAllowedAnnotations(obj metav1.Object, pvc *corev1.PersistentVolumeCla
...
+func ClaimIsPopulatedForDataVolume(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) bool {
+	return pvc.Annotations[AnnPopulatedFor] == dv.Name
+}
+
+// ClaimAllowsAdoption returns true if the PVC may be adopted
+func ClaimAllowsAdoption(c client.Client, pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	anno, ok := pvc.Annotations[AnnAllowClaimAdoption]
+	// if annotation exists, go with that regardless of featuregate
+	if ok {
+		val, _ := strconv.ParseBool(anno)
```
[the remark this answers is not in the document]
awels: I think maybe `allow` and `deny` as values makes more sense, with the name of the annotation not having allow in it.

pkg/controller/datavolume/controller-base.go
```diff
@@ -631,7 +673,14 @@ func (r *ReconcilerBase) validatePVC(dv *cdiv1.DataVolume, pvc *corev1.Persisten
...
 	// a warning to the event recorder and return
 	if !metav1.IsControlledBy(pvc, dv) {
-		if pvcIsPopulated(pvc, dv) {
+		requiresNoWork, err := r.pvcRequiresNoWork(pvc, dv)
+		if err != nil {
+			return err
+		}
+		if requiresNoWork {
+			if cc.IsUnbound(pvc) {
+				return fmt.Errorf("unbound populated/adoptable PVC %s/%s", pvc.Namespace, pvc.Name)
```
awels, line 682: Can we set a condition or emit an event on the datavolume in this case. We want to make this discoverable without looking in the logs.

pkg/controller/datavolume/controller-base.go
```diff
@@ -1192,3 +1249,13 @@ func (r *ReconcilerBase) shouldUseCDIPopulator(syncState *dvSyncState) (bool, er
 
 	return usePopulator, nil
 }
+
+func (r *ReconcilerBase) pvcRequiresNoWork(pvc *corev1.PersistentVolumeClaim, dv *cdiv1.DataVolume) (bool, error) {
```
awels, line 1253: Can we flip the name and logic for this to `pvcRequiresWork` I am seeing a lot of double negatives in the rest of the code `if !pvcRequiresNoWork` which means `if pvcRequiresWork`. This would make it easier to read and understand.

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -763,6 +789,48 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, pvc)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pvc.OwnerReferences).To(HaveLen(1))
+			or := pvc.OwnerReferences[0]
+			Expect(or.UID).To(Equal(dv.UID))
+
+			err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, dv)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(dv.Status.Phase).To(Equal(cdiv1.Succeeded))
+			Expect(string(dv.Status.Progress)).To(Equal("N/A"))
```
awels, line 810: I am fairly certain there is some code that sets progress to 100% in case of the phase being succeeded. Did this not happening yet in this case?

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -763,6 +789,48 @@ var _ = Describe("All DataVolume Tests", func() {
...
```
awels, line 831: I am fairly certain there is some code that sets progress to 100% in case of the phase being succeeded. Did this not happening yet in this case?

pkg/controller/datavolume/upload-controller_test.go
```diff
@@ -204,6 +204,60 @@ var _ = Describe("All DataVolume Tests", func() {
 		Expect(string(dv.Status.Progress)).To(Equal("N/A"))
 	})
 
+	It("Should adopt a PVC (with annotation)", func() {
+		annotations := map[string]string{"cdi.kubevirt.io/allowClaimAdoption": "true"}
```
awels, line 208: Maybe use the constant for the annotation name?

pkg/controller/datavolume/upload-controller_test.go
```diff
@@ -204,6 +204,60 @@ var _ = Describe("All DataVolume Tests", func() {
...
+		Expect(or.UID).To(Equal(dv.UID))
+
+		err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, dv)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(dv.Status.Phase).To(Equal(cdiv1.Succeeded))
+		Expect(string(dv.Status.Progress)).To(Equal("N/A"))
+	})
+
+	It("Should NOT adopt a unbound PVC", func() {
+		annotations := map[string]string{"cdi.kubevirt.io/allowClaimAdoption": "true"}
```
awels, line 250: Use the const for the annotation name?

tests/adopt-volume_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	controller "kubevirt.io/containerized-data-importer/pkg/controller/common"
+	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
+	"kubevirt.io/containerized-data-importer/tests/framework"
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+var _ = Describe("PVC adoption tests", Serial, func() {
+	f := framework.NewFramework("transfer-test")
```
awels, line 21: `adoption-test`?

awels:
The failure seems relevant, it is rejecting the creation of the DV because the PVC already exists.

pkg/apiserver/webhooks/datavolume-validate_test.go
```diff
@@ -195,10 +196,104 @@ var _ = Describe("Validating Webhook", func() {
...
+					Namespace: dataVolume.Namespace,
+				},
+				Spec: *dataVolume.Spec.PVC,
+			}
+			pvc.Spec.VolumeName = "pv"
+			resp := validateDataVolumeCreateEx(dataVolume, []runtime.Object{pvc}, nil, nil, []string{"DataVolumeClaimAdoption"})
+			Expect(resp.Allowed).To(BeTrue())
+		})
+
+		It("should accept DataVolume with unbound PVC and adoption featurgate set", func() {
```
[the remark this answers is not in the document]
mhenriks: Text should say reject.  That's what the test does.  Is rejected because unbound

pkg/controller/common/util.go
mhenriks: I potentially like "allow" as a value but still not sure about naming.  Any other ideas?  ss @akalenyu

pkg/controller/common/util.go
mhenriks: unfortunately it doesn't appear that the function will be importable from other packages if we do that though

pkg/controller/datavolume/controller-base.go
mhenriks: It shouldn't really happen, the webhook makes sure the PVC is bound

pkg/controller/common/util.go
awels: That is probably why the function is in here.

pkg/controller/datavolume/controller-base.go
awels: What about WFFC type storage? It would still be nice if we can bubble up severe issues like this to the user.

pkg/controller/datavolume/import-controller_test.go
mhenriks: Not for all operations.  Progress is the value of the annotation (if it exists) otherwise N/A.  I could add the annotation to the PVC but I don't think it really matters much here

pkg/controller/datavolume/import-controller_test.go
mhenriks: See comment above.

pkg/controller/datavolume/upload-controller_test.go
mhenriks: I actually think it's technically better to not use the shared consts in tests.  Can easily detect when you break an existing API.  We (myself included) are not consistent about this though.

pkg/controller/datavolume/upload-controller_test.go
mhenriks: See above

pkg/controller/datavolume/upload-controller_test.go
awels: Well in other tests in the PR you do use the const, so for consistency at least in the PR I think it makes sense.

pkg/controller/datavolume/controller-base.go
mhenriks: Yeah I think I have to remove this restriction

pkg/controller/datavolume/upload-controller_test.go
mhenriks: fair enough

pkg/controller/common/util.go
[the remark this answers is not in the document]
alromeros: Not allowing adoption has been the expected behavior from the beginning, I think it makes sense to just treat the annotation existing as true. It also feels more consistent with other CDI annotations. Just my thoughts though.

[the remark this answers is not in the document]
alromeros: However, using booleans would allow explicitly disabling adoption even when the feature gate is enabled, right? That would make some testing scenarios more straightforward. In that case I think I prefer using `allow` and `deny`.

pkg/controller/datavolume/controller-base.go
mhenriks: Changed to `pvcRequiresWork`

pkg/controller/common/util.go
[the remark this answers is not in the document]
mhenriks: Good point @alromeros

pkg/apiserver/webhooks/datavolume-validate_test.go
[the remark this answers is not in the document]
awels: So I am confused now, the text matches the result of the test now, but you said it should be rejected, but it is not accepted.

akalenyu:
If you put this to draft because of the nfs lane, it likely has nothing to do with this PR. We have a thread about this but for some reason in some runs the NFS server breaks and never recovers. Check out the nfs csi daemon logs

tests/datavolume_test.go
```diff
@@ -40,7 +40,7 @@ const (
 	shortTimeout        = 30 * time.Second
 )
 
-var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", func() {
+var _ = FDescribe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", func() {
```
akalenyu, line 43: You need some more changes to make this achievable, here's a reference
https://github.com/kubevirt/containerized-data-importer/pull/2384/files#diff-ce25e369a9df9edc525425cc144fc8c40599fff3cc1e5058b64336d71af67162

cmd/cdi-apiserver/apiserver.go
```diff
@@ -119,6 +121,15 @@ func main() {
 
 	snapClient := snapclient.NewForConfigOrDie(cfg)
 
+	mgr, err := manager.New(cfg, manager.Options{})
```
akalenyu, line 124: sure you want to create a manager just for the sake of client? @arnongilboa is initializing a cached cr client in same app in https://github.com/kubevirt/containerized-data-importer/pull/2813/files#diff-1189f871222be9b4a380b421d75ed238831eca4554a236bfa61f15247e25ac99R128

tests/adopt-volume_test.go
```diff
@@ -0,0 +1,218 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	controller "kubevirt.io/containerized-data-importer/pkg/controller/common"
+	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
+	"kubevirt.io/containerized-data-importer/tests/framework"
+	"kubevirt.io/containerized-data-importer/tests/utils"
+)
+
+var _ = Describe("PVC adoption tests", Serial, func() {
```
akalenyu, line 20: these have to be serial just because of the feature gate, right?

cmd/cdi-apiserver/apiserver.go
mhenriks: Yeah I don't think there is any additional overhead using a manager and it seems cleaner to me

mhenriks: To me, the manager is the entrypoint for the Controller Runtime API.  And last time I went through the manager code, I don't recall it doing much other than creating the client and starting controllers.  So if there are no controllers I don't see any issues.  Are you aware of any unnecessary work/data being created?

akalenyu: At the very least I recall it creates a metrics listener, anyway, I may be the only one finding this weird
cc @awels wdyt

tests/adopt-volume_test.go
akalenyu: according to that block's before/after each it will flip the feature gate too, and thus potentially mess up other tests. what am I missing

cmd/cdi-apiserver/apiserver.go
mhenriks: Okay, if metrics server is started, I'll change to create client directly

tests/adopt-volume_test.go
mhenriks: Yeah good catch, forgot about that, doesn't actually flip the featuregate now but will if/when it is the default

cmd/cdi-apiserver/apiserver.go
```diff
@@ -119,6 +121,15 @@ func main() {
 
 	snapClient := snapclient.NewForConfigOrDie(cfg)
 
+	cluster, err := cluster.New(cfg)
+	if err != nil {
+		klog.Fatalf("Unable to create controller runtime cluster: %v\n", errors.WithStack(err))
+	}
+
+	if err := cdiv1.AddToScheme(cluster.GetScheme()); err != nil {
```
akalenyu, line 129: Aren't some schemes missing here? for example https://github.com/kubevirt/containerized-data-importer/blob/a663355f98938a4136ddabe60f599a8b77ef7fcb/cmd/cdi-controller/controller.go#L67 what am I missing

mhenriks: Not needed yet, the client just gets cdi resources

awels (approved):
/approve

PR #3033, opened 2023-12-13
Add test for checking user-facing resources can be manipulated by non-cluster-admin

akalenyu:
This should keep us from introducing user-facing resources that cannot be
manipulated by non-cluster-admin.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/rbac_test.go
+++ b/tests/rbac_test.go
@@ -2,6 +2,8 @@ package tests
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


tests/rbac_test.go
```diff
@@ -147,6 +149,34 @@ var _ = Describe("Aggregated role in-action tests", Serial, func() {
 		Entry("[test_id:3949]can do everything with edit", "edit"),
 	)
 
+	DescribeTable("check all user facing resources can be manipulated by non-cluster-admin", func(user string) {
+		var namespacedCDIAPIResourcesPlural []string
+		// Also retrieves API resources that are not served via CRDs
+		outputAPIResources, err := f.RunKubectlCommand("api-resources", "--namespaced", "-o", "name")
+		Expect(err).ToNot(HaveOccurred(), "ERR: %s, OUT: %s", err, outputAPIResources)
+		for _, apiResource := range strings.Split(strings.TrimSpace(outputAPIResources), "\n") {
+			if strings.Contains(apiResource, "cdi.kubevirt.io") {
```
awels, line 158: maybe use `HasSuffix` instead of `Contains`. limits the matches a bit.

tests/rbac_test.go
```diff
@@ -147,6 +149,34 @@ var _ = Describe("Aggregated role in-action tests", Serial, func() {
...
+
+		for _, resource := range namespacedCDIAPIResourcesPlural {
+			sa := fmt.Sprintf("system:serviceaccount:%s:%s", f.Namespace.Name, user)
+			result, err := f.RunKubectlCommand("auth", "can-i", "--as", sa, "*", resource, "--namespace", f.Namespace.Name)
+			Expect(err).ToNot(HaveOccurred(), "no permission for %s, result: %s", resource, result)
+			Expect(strings.TrimSpace(result)).To(Equal("yes"))
+		}
+	},
+		Entry("[test_id:XXXX]for admin", "admin"),
+		Entry("[test_id:XXXX]for edit", "edit"),
```
awels, line 177: What about the `view` user? They have much more limited access. But probably should still have get/list/watch on most resources.

tests/rbac_test.go
```diff
@@ -147,6 +149,34 @@ var _ = Describe("Aggregated role in-action tests", Serial, func() {
 		Entry("[test_id:3949]can do everything with edit", "edit"),
 	)
 
+	DescribeTable("check all user facing resources can be manipulated by non-cluster-admin", func(user string) {
```
awels, line 152: Maybe pass in a map or list of resources and rbac rules that a user should be allowed to do. This way we are checking all the aggregate roles and not just the admin ones.

tests/rbac_test.go
akalenyu: Good point. I took a stab at that and only "uploadtokenrequests" diverges. The cluster wide resources may need extra adjustments too

tests/rbac_test.go
akalenyu: I think we want to avoid passing a list of resources, and by doing that, we protect the project from introducing unusable new APIs in the future

tests/rbac_test.go
akalenyu: Makes sense, updated for now

PR #3037, opened 2023-12-19
Refactor e2e kubernetes reporter

akalenyu:
- Move to separate package
- Extract common code
- Ensure the reporting fires right after an It() block instead of (current behavior) after other AfterEach blocks.

This also fixes broken reporting when failures occurred on both the node 1 parallel/serial run

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -5,10 +5,11 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strconv"
 	"testing"
 	"time"
 
-	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/ginkgo/v2"
 	ginkgo_reporters "github.com/onsi/ginkgo/v2/reporters"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


tests/tests_suite_test.go
```diff
@@ -166,4 +175,13 @@ func BuildTestSuite() {
 			Expect(err).ToNot(HaveOccurred())
 		}
 	})
+
+	JustAfterEach(func() {
```
akalenyu, line 179: This will make sure we get artifacts that got deleted in AfterEach blocks
/cc @alromeros

awels (approved):
Just a note, its a battle between should we . import gingko and gomega or not. I see both, and the static linters will complain if we do. I am super passionate about either one, but we should be consistent, and we are not.

akalenyu:
> Just a note, its a battle between should we . import gingko and gomega or not. I see both, and the static linters will complain if we do. I am super passionate about either one, but we should be consistent, and we are not.

Definitely don't want to make it worse. updated to dot import

awels (approved):

alromeros:
> Ensure the reporting fires right after an It() block instead of (current behavior) after other AfterEach blocks
Thanks for the PR, Alex. This will be really useful in some cases.
/lgtm

PR #3044, opened 2024-01-02
[WIP] Avoid race condition during importer termination

alromeros:

This pull request sets the grace period to 0 when deleting the importer pod after exiting for scratchSpaceNeeded. The 30s default grace period caused some issues during termination while the pod was already restarting after scratchSpaceNeeded.

```sh
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 2
        exitCode: 42
        exitCode: 2
        exitCode: 42
        exitCode: 2
        exitCode: 0
        exitCode: 0
```
After the fix:
```sh
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 0
        exitCode: 0
```

I'm unsure about adding tests to this. Would appreciate recommendations. 

**Release note**:

```release-note
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -452,9 +452,16 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 }
 
 func (r *ImportReconciler) cleanup(pvc *corev1.PersistentVolumeClaim, pod *corev1.Pod, log logr.Logger) error {
-	if err := r.client.Delete(context.TODO(), pod); cc.IgnoreNotFound(err) != nil {
+	deleteOptions := &client.DeleteOptions{}
+	// Setting grace period to 0 if pod already has a deletion timestamp.
+	// This allows avoiding race conditions during pod termination.
+	if pod.DeletionTimestamp != nil {
+		deleteOptions.GracePeriodSeconds = ptr.To[int64](0)
+	}
+	if err := r.client.Delete(context.TODO(), pod, deleteOptions); cc.IgnoreNotFound(err) != nil {
 		return err
 	}
+
 	if cc.HasFinalizer(pvc, importPodImageStreamFinalizer) {
 		cc.RemoveFinalizer(pvc, importPodImageStreamFinalizer)
 		if err := r.updatePVC(pvc, log); err != nil {


alromeros:
/hold
Let's check the test failures

pkg/controller/import-controller.go
```diff
@@ -439,9 +439,17 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 }
 
 func (r *ImportReconciler) cleanup(pvc *corev1.PersistentVolumeClaim, pod *corev1.Pod, log logr.Logger) error {
-	if err := r.client.Delete(context.TODO(), pod); cc.IgnoreNotFound(err) != nil {
+	deleteOptions := &client.DeleteOptions{}
+	// Setting grace period to 0 if pod already has deletion timestamp.
+	// This allows avoiding race conditions during pod termination.
+	if pod.DeletionTimestamp != nil {
+		gracePeriod := int64(0)
+		deleteOptions.GracePeriodSeconds = &gracePeriod
```
awels, line 447: Can probably do `deleteOptions.GracePeriodSeconds = pointer.Int64Ptr(0)` instead.

awels (commented):
Have you tried simply always setting grace period to 0 regardless of the deletionTimestamp on the pod? At this point in the code, we want the pod to disappear, and the importer is not holding any resources it can't release if it disappears. Would make the logic simpler.

pkg/controller/import-controller.go
akalenyu: Actually following https://github.com/kubevirt/containerized-data-importer/pull/3009/commits/aa123c07faf5224ffb81cbfe1805c4eb707b6b0b it's `ptr.To[int64](0)`

akalenyu (commented):
So I wonder if simply deleting a pod during its grace period is what triggers the appearance of this weird exit code.
In that case, do we want to be working around that?
I can't think of a particular thing we're losing here by bypassing the grace period but generally I think it's better that pods get cleaned up gracefully

pkg/controller/import-controller.go
awels: Either way works, looks like we have both in the code base. I would suggest we standardize on one.

akalenyu: Where? we shouldn't have `k8s.io/utils/pointer` anymore

awels:
> So I wonder if simply deleting a pod during its grace period is what triggers the appearance of this weird exit code. In that case, do we want to be working around that (bug?)? I can't think of a particular thing we're losing here by bypassing the grace period but generally I think it's better that pods get cleaned up gracefully

I agree but in this case we know the process is already exited (for scratch space). So there really is no difference IMO. There are no resources we need to close before exiting the process. Maybe there is a problem if we have a istio sidecar, not sure.

pkg/controller/import-controller.go
awels: Oh I had not pulled the latest, after rebase I see no trace of `pointer.`

alromeros:
> > So I wonder if simply deleting a pod during its grace period is what triggers the appearance of this weird exit code. In that case, do we want to be working around that (bug?)? I can't think of a particular thing we're losing here by bypassing the grace period but generally I think it's better that pods get cleaned up gracefully
> 
> I agree but in this case we know the process is already exited (for scratch space). So there really is no difference IMO. There are no resources we need to close before exiting the process. Maybe there is a problem if we have a istio sidecar, not sure.

Don't we clean-up the pod here once import is completed? Using 0 grace period seems safe when we want to delete the pod after scratch space needed, but after completion it seems safer to use the default one.

alromeros:
Setting the delete grace period to 0 caused some unwanted behavior with container being stuck on the ceph lane. Closing this since I found a better alternative.

PR #3053, opened 2024-01-08
document enablement of podman socket in README.md

### LINKED ISSUE #3052: Document how to enable podman sockets to run cluster-up
**Description**:
if your CLI is podman, get 
```
./cluster-up/up.sh
no working container runtime found. Neither docker nor podman seems to work.
make: *** [Makefile:108: cluster-up] Error 1
```

**What you expected**:
running `make cluster-up` should result in a running cluster with kubevirt-cli operator installed and ready to test. 
This requires that the unix socket in `/run/podman/podman.socket` be enabled. This is not documented. 

**URL**:
[documentation for podman socket activation](https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md)

So at minimum need a link to the above documentation and an exhortation in cluster-up/README.md to at least run 
```
systemctl start podman.socket
```
as root in order to enable `/run/podman/podman.sock` 

**Additional context**:
Add any other context about the issue here.

cfilleke:
This addresses issue #3052

The enablement of podman socket on the test host should be documented. 

Fixes #3052 

we ran into this setting up an external CI/CD system for s390x but it applies to all platforms where `podman` is in use over `docker`.  

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/README.md
+++ b/cluster-up/README.md
@@ -4,6 +4,15 @@ This directory provides a wrapper around gocli. It can be vendored into other
 git repos and integrated to provide in the kubevirt well-known cluster commands
 like `make cluster-up` and `make cluster-down`.
 
+In order to use `make cluster-up` with `podman` you need to enable podman sockets by running 
+```
+systemctl start podman.socket
+```
+as root. If you need this configuration to persist beyond a system reboot, also:
+```
+systemctl enable podman.socket
+```
+
 In order to properly use it, one has to vendor this folder from a git tag,
 which can be found on the github release page.
 


akalenyu (commented):
So this is actually a generated file, which gets updated together with kubevirtci bumps at `hack/update-kubevirtci.sh`

I am totally up for adding this though, maybe in `hack/README.md`?

awels:
Oh yeah everything under cluster-up comes from the [kubevirtci](https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/README.md) repository, maybe open this PR over there, so that when we update kubevirtci over here it automatically gets added?

cfilleke:
Not sure if this needs to be updated in both kubevirt/containerized-data-importer and kubevirt/kubevirtci or just [kubevirtci](https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/README.md) -- possibly both, as this prerequisite will remain undocumented in CDI until this file is re-generated out of kubevirtci, which doesn't happen instantaneously.

awels:
Well it needs to get updated in kubevirtci first. Once that is merged, we can update kubevirtci in CDI (and also others like kubevirt). This is relevant for more than just CDI IMO, so it makes sense to start with kubevirtci, and have it propagate from there. Even if it takes a bit of time to update kubevirtci in CDI.

awels:
Note we will be updating kubevirtci relatively soon when k8s 1.29 becomes available for test lanes. So it does make sense to put this change in the kubevirtci repo now.

PR #3056, opened 2024-01-09
Fix populators not repopulating pvc after it was deleted

ShellyKa13:
This PR should fix this bug:
Jira-ticket: https://issues.redhat.com/browse/CNV-35085

Fixes #CNV-35085

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -138,6 +138,33 @@ var _ = Describe("all clone tests", func() {
 			Expect(uploader.DeletionTimestamp).To(BeNil())
 		})
 
+		It("should recreate and reclone target pvc if it was deleted", func() {
+			pvcDef := utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
+			pvcDef.Namespace = f.Namespace.Name
+			sourcePvc = f.CreateAndPopulateSourcePVC(pvcDef, sourcePodFillerName, fillCommand+testFile+"; chmod 660 "+testBaseDir+testFile)
+			dvName := "target-dv"
+			doFileBasedCloneTest(f, pvcDef, f.Namespace, dvName, "1Gi")
+
+			targetPVC, err := f.FindPVC(dvName)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Delete target PVC")
+			err = utils.DeletePVC(f.K8sClient, f.Namespace.Name, dvName)
+			Expect(err).ToNot(HaveOccurred())
+
+			deleted, err := f.WaitPVCDeletedByUID(targetPVC, time.Minute)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(deleted).To(BeTrue())
+
+			targetPVC, err = utils.WaitForPVC(f.K8sClient, f.Namespace.Name, dvName)
+			Expect(err).ToNot(HaveOccurred())
+			f.ForceBindIfWaitForFirstConsumer(targetPVC)
+
+			By("Verify target PVC is bound again")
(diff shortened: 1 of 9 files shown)


tests/cloner_test.go
```diff
@@ -138,6 +138,33 @@ var _ = Describe("all clone tests", func() {
 			Expect(uploader.DeletionTimestamp).To(BeNil())
 		})
 
+		It("should recreate and reclone target pvc if it was deleted", func() {
```
akalenyu, line 141: So this test will actually run Serially since it's under `Describe("[rfe_id:1277][crit:high][vendor:cnv-qe@redhat.com][level:component]Cloner Test Suite", Serial, func()`
Can we maybe move it around so it can run parallel, or is there a constraint I am missing?

tests/import_test.go
```diff
@@ -2080,6 +2080,56 @@ var _ = Describe("Import populator", func() {
...
+		dv, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dataVolume)
+		Expect(err).ToNot(HaveOccurred())
+
+		pvc, err = utils.WaitForPVC(f.K8sClient, dv.Namespace, dv.Name)
+		Expect(err).ToNot(HaveOccurred())
+		f.ForceBindIfWaitForFirstConsumer(pvc)
+
+		By("Verify PVC prime was created")
+		pvcPrime, err = utils.WaitForPVC(f.K8sClient, pvc.Namespace, populators.PVCPrimeName(pvc))
+		Expect(err).ToNot(HaveOccurred())
```
arnongilboa, line 2095: Just a dumb question (I guess this is done in many tests), can't PVC prime already be deleted in this point if import was quick enough to complete?

tests/cloner_test.go
ShellyKa13: yes sure thanks for noticing

tests/import_test.go
ShellyKa13: yeah I agree, I did several changes in this test, since I found the deletePVCByUID then its safe enough to see that the new created PVC became bound again.. (I had a race there with the get). Ill simplify the test more.

tests/cloner_test.go
ShellyKa13: I actually dont see there is the serial decorator there, are you sure? @akalenyu

alromeros:
Looks really good. Maybe add an upload functional test?

akalenyu:
> Looks really good. Maybe add an upload functional test?

Random question. Did we not ever have tests that check the PVC recreated & repopulated upon deletion?
Do those exists, and for some reason not use populators?

ShellyKa13:
@akalenyu hmm from what I saw I didnt find any..
@alromeros the pipeline of upload populator is like the import populator, so I thought that UT for both and functest for import is enough..

akalenyu:
> hmm from what I saw I didnt find any..

There's https://github.com/kubevirt/containerized-data-importer/blob/4188a8689e560964b171e94d72947faa44d4bb01/tests/datavolume_test.go#L3023

ShellyKa13:
> > hmm from what I saw I didnt find any..
> 
> There's
> 
> https://github.com/kubevirt/containerized-data-importer/blob/4188a8689e560964b171e94d72947faa44d4bb01/tests/datavolume_test.go#L3023

Yeah this is different since we delete the pvc during the import so the dv haven't reached succeeded state yet thatss why we didnt catch the bug

ShellyKa13:
@akalenyu @alromeros @arnongilboa can you review again?

akalenyu:
cc @mhenriks 
I am fine with the test changes I was suggesting

jpeimer:
Needs a backport to 1.58 as well

PR #3057, opened 2024-01-10
Fix DataImportCron PVC GC race and test flakiness

arnongilboa:
PVC should be timestamped in creation and not only upon import completion, as it might be mistakenly GCed. LRU sort will choose PVC with empty timestamp as the first candidate for deletion. The PVC will be recreated by the controller and eventually timestamped, so this bug was hidden for a while.

Fixes CNV-36896

**Release note**:
```release-note
Fix DataImportCron PVC timestamping for garbage collection
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -57,7 +57,7 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
 	cdv "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
-	"kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"
+	metrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"
 	"kubevirt.io/containerized-data-importer/pkg/operator"
 	"kubevirt.io/containerized-data-importer/pkg/util"
 	"kubevirt.io/containerized-data-importer/pkg/util/naming"
@@ -1254,6 +1254,7 @@ func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCro
 	dv.Namespace = cron.Namespace
 	r.setDataImportCronResourceLabels(cron, dv)
 	cc.AddAnnotation(dv, cc.AnnImmediateBinding, "true")
+	cc.AddAnnotation(dv, AnnLastUseTime, time.Now().UTC().Format(time.RFC3339Nano))
 	passCronAnnotationToDv(cron, dv, cc.AnnPodRetainAfterCompletion)
 
 	for _, defaultInstanceTypeLabel := range cc.DefaultInstanceTypeLabels {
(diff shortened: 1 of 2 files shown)


tests/dataimportcron_test.go
```diff
@@ -512,9 +512,11 @@ var _ = Describe("DataImportCron", Serial, func() {
...
 			Expect(err).ToNot(HaveOccurred())
 
-			pvcList, err := f.K8sClient.CoreV1().PersistentVolumeClaims(ns).List(context.TODO(), metav1.ListOptions{})
-			Expect(err).ToNot(HaveOccurred())
-			Expect(pvcList.Items).To(HaveLen(garbageSources + 1))
+			Eventually(func() int {
+				pvcList, err := f.K8sClient.CoreV1().PersistentVolumeClaims(ns).List(context.TODO(), metav1.ListOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return len(pvcList.Items)
+			}, dataImportCronTimeout, pollingInterval).Should(Equal(garbageSources + 1))
```
akalenyu, line 519: Why are we switching to eventually here?

pkg/controller/dataimportcron-controller.go
```diff
@@ -57,7 +57,7 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
 	cdv "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
-	"kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"
+	metrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"
```
akalenyu, line 60: isn't the package already named metrics? is this nicer?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1254,6 +1254,7 @@ func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCro
 	dv.Namespace = cron.Namespace
 	r.setDataImportCronResourceLabels(cron, dv)
 	cc.AddAnnotation(dv, cc.AnnImmediateBinding, "true")
+	cc.AddAnnotation(dv, AnnLastUseTime, time.Now().UTC().Format(time.RFC3339Nano))
```
akalenyu, line 1257: At first I was worried older versions won't pass annotations but the problem was only with labels. https://github.com/kubevirt/containerized-data-importer/issues/2533

pkg/controller/dataimportcron-controller.go
arnongilboa: Right. No clue why my VS go formatter insists to add it only for this specific import. Any guess?

tests/dataimportcron_test.go
arnongilboa: The flakiness was because sometime `pvcList` still included a pvc-prime which was not deleted yet. `Eventually` fixes this flakiness.

akalenyu: oh okay cool. consider using `.Should(HaveLen(x))` here

pkg/controller/dataimportcron-controller.go
akalenyu: Same for me, let's keep it so we're not breaking some consensus

akalenyu: I think the problem is that this package is not named `cdi-controller`, it's named `metrics`
"kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"

tests/dataimportcron_test.go
arnongilboa: fixed

pkg/controller/dataimportcron-controller.go
akalenyu: cc @avlitman not sure if this is the intention

ShellyKa13 (commented):
/lgtm

PR #3060, opened 2024-01-11
[WIP] Set importer restart policy to never to avoid race condition during termination

alromeros:

This is an alternative to https://github.com/kubevirt/containerized-data-importer/pull/3044.

This pull request changes the importer restart policy from "on failure" to "never". The old behavior caused some issues during termination while the pod was already restarting after scratchSpaceNeeded.

```sh
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 2
        exitCode: 42
        exitCode: 2
        exitCode: 42
        exitCode: 2
        exitCode: 0
        exitCode: 0
```

After the fix:

```sh
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 0
        exitCode: 0
```

Let's see how tests behave and decide which option is better.

**Release note**:

```release-note
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -187,7 +187,7 @@ func handleImport(
 			if err := util.WriteTerminationMessage(common.ScratchSpaceRequired); err != nil {
 				klog.Errorf("%+v", err)
 			}
-			return common.ScratchSpaceNeededExitCode
+			return 0
 		}
 		err = util.WriteTerminationMessage(fmt.Sprintf("Unable to process data: %v", err.Error()))
 		if err != nil {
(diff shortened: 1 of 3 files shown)


alromeros:
/hold

Let's see how well tests behave and decide whether https://github.com/kubevirt/containerized-data-importer/pull/3044 or this is a better option.

akalenyu:
> /hold
> 
> Let's see how well tests behave and decide whether #3044 or this is a better option.

I honestly don't know about this approach.. consider intermittent `err != nil`s in the importer execution, like, HTTP flukes for example. We do want to be eventually consistent and keep trying on those

alromeros:
> > /hold
> > Let's see how well tests behave and decide whether #3044 or this is a better option.
> 
> I honestly don't know about this approach.. consider intermittent `err != nil`s in the importer execution, like, HTTP flukes for example. We do want to be eventually consistent and keep trying on those

I kinda agree, I want to have some discussion around this but even if we manually restart the importer, this method seems more buggy and prone to breaking than setting the 0 grace period.

alromeros:
I think I found a better alternative that's less intrusive to our tests. Closing this and opening a new PR.

PR #3068, opened 2024-01-15
Refactor recording-rules and alerts code

avlitman:
Following the work started in https://github.com/kubevirt/kubevirt/pull/10044 , and according to the https://github.com/kubevirt/community/pull/219 proposal, this PR refactors monitoring recording rules and alerts

Jira-Ticket: https://issues.redhat.com/browse/CNV-34288

**Release note**:

```
Refactor recording-rules and alerts code.

```


avlitman:
@sradco @machadovilaca @akalenyu CI is passing now and rules are verified manually, will appreciate your review (:

machadovilaca:
@avlitman 
 I think the release note message is not very pretty, maybe updating it to the PR title would look good

0
machadovilaca, line 1: you can remove this

tools/metricsdocs/metricsdocs.go
```diff
@@ -0,0 +1,54 @@
...
+
+	err = cdiMetrics.SetupMetrics()
+	if err != nil {
+		panic(err)
+	}
+
+	if err := rules.SetupRules("test"); err != nil {
+		panic(err)
+	}
+	docsString := docs.BuildMetricsDocsWithCustomTemplate(operatorMetrics.ListMetrics(), rules.ListRecordingRules(), tpl)
```
machadovilaca, line 52: nit: empty line between

pkg/operator/controller/prometheus.go
```diff
@@ -88,6 +79,10 @@ func ensurePrometheusResourcesExist(ctx context.Context, c client.Client, scheme
 			return err
 		}
 
+		switch rule := desired.(type) {
+		case *promv1.PrometheusRule:
+			log.Info("[DEBUG]", "desired prometheus object spec", rule.Spec)
+		}
```
machadovilaca, line 85: leftover?

avlitman:
> @avlitman I think the release note message is not very pretty, maybe updating it to the PR title would look good

Fixed @machadovilaca thanks!

pkg/operator/controller/controller.go
```diff
@@ -39,7 +39,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
-	metrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/operator-controller"
+	"kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/operator-controller"
```
machadovilaca, line 42:
```suggestion
	metrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/operator-controller"
```

should be named as metrics, because later you use `metrics.SetupMetrics()`, no?

pkg/monitoring/rules/rules.go
```diff
@@ -0,0 +1,54 @@
...
+	return nil
+}
+
+// BuildPrometheusRule creates a PrometheusRule in a namespace.
+func BuildPrometheusRule(namespace string) (*promv1.PrometheusRule, error) {
+	rules, err := operatorrules.BuildPrometheusRule(
+		ruleName,
+		namespace,
+		map[string]string{
+			common.CDIComponentLabel:  "",
```
machadovilaca, line 35: empty value?

pkg/monitoring/rules/rules.go
```diff
@@ -0,0 +1,54 @@
...
+			common.PrometheusLabelKey: common.PrometheusLabelValue,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return rules, nil
```
machadovilaca, line 43: you can just return `operatorrules.BuildPrometheusRule...`

machadovilaca (changes requested):

0
avlitman: ohh it;s be mistake removing

pkg/monitoring/rules/rules.go
avlitman: they did it like this before...

avlitman: see https://github.com/kubevirt/containerized-data-importer/blob/b33addca41bcc0eac76dc4cd37602f66b8cf0f71/pkg/operator/controller/prometheus.go#L285

pkg/monitoring/rules/rules.go
avlitman: Not sure since they expect *promv1.PrometheusRule:
"Cannot use 'operatorrules.BuildPrometheusRule' (type func(name string, namespace string, labels map[string]string) (*promv1.PrometheusRule, error)) as the type *promv1.PrometheusRule"

pkg/operator/controller/controller.go
avlitman: but the pkg name is metrics so it's duplicate to give it an alias named metrics

pkg/operator/controller/prometheus.go
avlitman: wow I don't know where is came from, removing.

tools/metricsdocs/metricsdocs.go
avlitman: added a line before and after this.

pkg/operator/controller/controller.go
machadovilaca: the package name should be the name of the folder by convention I think, but since this wasn't changed in this PR let's leave it as is for now

pkg/monitoring/rules/rules.go
machadovilaca: okay

avlitman: but I guess we can remove this maybe I'm not sure

pkg/monitoring/rules/rules.go
machadovilaca: could also be improved there

`operatorrules.BuildPrometheusRule` already returns `rules, error`
`return operatorrules.BuildPrometheusRule, err` should be return `operatorrules.BuildPrometheusRule`

pkg/monitoring/rules/rules.go
machadovilaca: let's leave it as was

pkg/operator/controller/controller.go
avlitman: Alex also mentioned it's not needed in Arnon's pr see: https://github.com/kubevirt/containerized-data-importer/pull/3057#discussion_r1448044345

avlitman: for me it worked manually

pkg/monitoring/rules/rules.go
avlitman: thanks a lot! fixed

pkg/monitoring/rules/rules.go
```diff
@@ -0,0 +1,49 @@
...
+
+// SetupRules initializes recording and alert rules in a namespace.
+func SetupRules(namespace string) error {
+	if err := recordingrules.Register(namespace); err != nil {
+		return err
+	}
+
+	if err := alerts.Register(namespace); err != nil {
+		return err
+	}
```
akalenyu, line 24: I think you want these (or the Register func content itself) wrapped up in a https://pkg.go.dev/sync#Once
Since there is basically no guarantee that the caller won't run SetupRules multiple times

How does the k8s metrics helper lib achieve this?

pkg/operator/controller/controller.go
akalenyu: It is actually needed - https://github.com/kubevirt/containerized-data-importer/pull/3057#discussion_r1448696072
> the package name should be the name of the folder by convention I think

I think we should tackle this, in this PR or later so we don't continue to have this slip into other people's PRs

pkg/operator/controller/controller_test.go
```diff
@@ -285,14 +288,10 @@ var _ = Describe("Controller", func() {
...
-				runbookURLTemplate := getRunbookURLTemplate()
+				runbookURLTemplate := alerts.GetRunbookURLTemplate()
 
-				rule := &promv1.PrometheusRule{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "prometheus-cdi-rules",
-						Namespace: cdiNamespace,
-					},
-				}
+				rule, err := rules.BuildPrometheusRule(cdiNamespace)
```
akalenyu, line 293: So this PrometheusRule is defined solely to call a GET, do we need the entire rule spec?

pkg/operator/controller/controller_test.go
```diff
@@ -313,7 +312,7 @@ var _ = Describe("Controller", func() {
 					},
 				}
 
-				Expect(rule.Spec.Groups[0].Rules).To(ContainElement(cdiDownAlert))
+				Expect(rule.Spec.Groups[1].Rules).To(ContainElement(cdiDownAlert))
```
akalenyu, line 315: rookie question here, having a split to groups between recording&alerting is cleaner?

akalenyu (commented):
looks good mainly concerned about protecting calling register multiple times

pkg/monitoring/rules/rules.go
avlitman: we do it the same for all operators at the moment, we have to set it up in order to list it...
@machadovilaca WDYT?

pkg/operator/controller/controller_test.go
avlitman: Yes it is needed and recommended

pkg/operator/controller/controller_test.go
avlitman: not sure I understand what you mean here sorry

avlitman:
@akalenyu "I think we should tackle this, in this PR or later so we don't continue to have this slip into other people's PRs" you mean remove it? or stay as is now?

pkg/monitoring/rules/rules.go
machadovilaca: end result is the same as using `prometheus.Register` directly, we tried to avoid different usage logic, it will error on duplicate registration

akalenyu: I mean this Register(), which just appends
https://github.com/kubevirt/containerized-data-importer/pull/3068/files#diff-1fd933a1e8d6a957bbdb03b4e22f73c2b114766e15a46162de2dc20a90725f2eR6

@avlitman initially had the call to SetupMetrics inside the reconcile loop, which overwhelmed the PrometheusRule resource, which we probably want to protect against

machadovilaca: ah yes, should be idempotent in the dependency

pkg/operator/controller/controller.go
avlitman: so remove the metrics alias or not? (:

pkg/operator/controller/controller_test.go
avlitman: Fixed

pkg/operator/controller/prometheus.go
```diff
@@ -134,166 +125,15 @@ func isPrometheusDeployed(logger logr.Logger, c client.Client, namespace string)
...
-			Groups: []promv1.RuleGroup{
-				{
-					Name:  "cdi.rules",
-					Rules: append(getRecordRules(namespace), getAlertRules(runbookURLTemplate)...),
-				},
-			},
-		},
+		ObjectMeta: promRule.ObjectMeta,
+		Spec:       promRule.Spec,
 	}
```
machadovilaca, line 137: this function can be deleted, and its usage replaced by `rules.BuildPrometheusRule(namespace)`, you are creating a new `PrometheusRule` with the contents of another `PrometheusRule`

pkg/monitoring/rules/alerts/operator.go
```diff
@@ -0,0 +1,92 @@
...
+								 sum(kubevirt_cdi_storageprofile_info{virtdefault="true",rwx="true",smartclone="true"} or on() vector(0)) == 0`),
+		For: (*promv1.Duration)(ptr.To("5m")),
+		Annotations: map[string]string{
+			"summary": "Default storage class has no smart clone or ReadWriteMany",
+		},
+		Labels: getLabels("warning", "none"),
+	},
+}
+
+func getLabels(severity string, impact string) map[string]string {
```
avlitman, line 87: do not use this function

pkg/monitoring/rules/rules.go
```diff
@@ -0,0 +1,54 @@
...
+			common.CDIComponentLabel:  "",
+			common.PrometheusLabelKey: common.PrometheusLabelValue,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return rules, nil
+}
```
machadovilaca, line 44:
```suggestion
func BuildPrometheusRule(namespace string) (*promv1.PrometheusRule, error) {
	return operatorrules.BuildPrometheusRule(
		ruleName,
		namespace,
		map[string]string{
			common.CDIComponentLabel:  "",
			common.PrometheusLabelKey: common.PrometheusLabelValue,
		},
	)
}
```

machadovilaca (changes requested):

pkg/monitoring/rules/rules.go
avlitman: fixed

pkg/monitoring/rules/rules.go
avlitman: fixed in the lib.

cmd/cdi-controller/controller.go
```diff
@@ -5,6 +5,7 @@ import (
 	"crypto/rsa"
 	"flag"
 	"fmt"
+	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
machadovilaca, line 8: should be in the second import group

pkg/monitoring/rules/rules.go
machadovilaca: I think you forgot to update here

pkg/operator/controller/controller_test.go
```diff
@@ -25,6 +25,9 @@ import (
 	"strconv"
 	"strings"
 
+	"kubevirt.io/containerized-data-importer/pkg/monitoring/rules"
+	"kubevirt.io/containerized-data-importer/pkg/monitoring/rules/alerts"
```
machadovilaca, line 29: should be grouped with the other `kubevirt.io/containerized-data-importer` imports

machadovilaca (changes requested):

akalenyu (commented):
Looks good! just some confusions

avlitman:
@akalenyu Thanks Alex I will fix all small comments in separate pr

cmd/cdi-controller/controller.go
```diff
@@ -67,6 +68,7 @@ var (
 	resourcesSchemeFuncs   = []func(*apiruntime.Scheme) error{
 		clientgoscheme.AddToScheme,
 		cdiv1.AddToScheme,
+		promv1.AddToScheme,
```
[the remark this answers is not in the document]
avlitman: I doesn't know the type if it's not added, don't remember where need to check

pkg/operator/controller/prometheus.go
```diff
@@ -134,166 +125,15 @@ func isPrometheusDeployed(logger logr.Logger, c client.Client, namespace string)
...
-			},
-		),
-	}
-}
-
 func newPrometheusRule(namespace string) *promv1.PrometheusRule {
-	runbookURLTemplate := getRunbookURLTemplate()
+	promRule, err := rules.BuildPrometheusRule(namespace)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
avlitman: let's discuses it later and I will fix in a separate pr if needed

pkg/controller/dataimportcron-conditions.go
```diff
@@ -24,7 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
-	"kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"
+	metrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-controller"
```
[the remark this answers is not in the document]
avlitman: I don't know if I ran it correctly but I don't see an error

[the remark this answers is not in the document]
akalenyu: > Yes we changed it in this pr as you asked, and the metrics is an alias we prefer to use. trying to check the test also..

Ok makes sense

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
akalenyu: So this one is a bit worrying, I don't think the cdi controllers should ever work against the Prometheus resources,
only the operator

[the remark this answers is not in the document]
avlitman: Can you explain more why it's a problem? It's just to add the type no?

pkg/operator/controller/prometheus.go
avlitman: We wrapped it in a func

pkg/monitoring/rules/rules.go
avlitman: I sent on private, here again: https://github.com/machadovilaca/operator-observability/commits/main/

pkg/controller/dataimportcron-conditions.go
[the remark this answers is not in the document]
avlitman: you prefer to change to the package name?

cmd/cdi-controller/controller.go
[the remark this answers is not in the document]
avlitman: Fixed (:

pkg/operator/controller/prometheus.go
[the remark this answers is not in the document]
avlitman: @akalenyu fixed your comments just waiting for an answer about this one. Thanks 🙏🏼

akalenyu:
/approve
/lgtm

thanks!

pkg/controller/dataimportcron-conditions.go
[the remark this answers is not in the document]
akalenyu: No, sounds fine

pkg/operator/controller/prometheus.go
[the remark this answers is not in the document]
avlitman: +1 will handle that in a follow-up pr

PR #3070, opened 2024-01-17
Introduce customizeComponents option

yaroslavborbat:
We currently cannot customize the components that cdi-operator deploys, but this is possible in KubeVirt.

The ability to configure resources for CDI in the same way as for KubeVirt
```
kind: CDI
metadata:
  name: cdi
spec:
  customizeComponents:
    flags:
      api: {"v": "3"}
      controller: {"v": "3"}
      uploadProxy: {"v": "3"}
    patches:
    - resourceType: Deployment
      resourceName: cdi-apiserver
      patch: '[{"op":"replace","path":"/spec/template/spec/imagePullPolicy","value":"Always"}]'
      type: json
```

Fixes #

**Release note**:

```release-note
Enhancement: Added option customizeComponents
```


awels:
@yaroslavborbat Hi, thanks for the PR, the idea looks really good, and a quick glance at the PR I don't see anything major. I will have @mhenriks also take a look at this PR. Could you add at least one functional test to show that the customization works. This will help us when maintaining the feature.

/test all

awels:
`make test-lint` will tell you all the things the linter is unhappy about. To run some tests, and see if cdi comes up, you can:
```bash
export KUBEVIRT_PROVIDER=k8s-1.27
make cluster-up
make cluster-sync
```
That should bring up a local cluster, build CDI, and deploy it to that cluster. That will tell you if it compiles and brings up CDI. After that you can `make test-functional` to run the functional tests (note this will take look, so I suggest you use focus to run only a subset).

yaroslavborbat:
@awels Hi, yes, I will be able to add a test soon

yaroslavborbat:
@awels 
hi, I'm having some problems with the e2e tests.
For example, with pull-containerized-data-importer-e2e-nfs:
`make: *** [Makefile:125: cluster-sync-cdi] Error 1`
Could you help me with this?
In my local sandbox, the following commands work fine
```
make cluster-up
make cluster-sync
```

akalenyu:
> @awels hi, I'm having some problems with the e2e tests. For example, with pull-containerized-data-importer-e2e-nfs: `make: *** [Makefile:125: cluster-sync-cdi] Error 1` Could you help me with this? In my local sandbox, the following commands work fine
> 
> ```
> make cluster-up
> make cluster-sync
> ```

Hey, I'll try to take a look

pkg/operator/resources/namespaced/controller.go
```diff
@@ -175,14 +171,14 @@ func createControllerServiceAccount() *corev1.ServiceAccount {
 
 func createControllerDeployment(controllerImage, importerImage, clonerImage, uploadServerImage, verbosity, pullPolicy string, imagePullSecrets []corev1.LocalObjectReference, priorityClassName string, infraNodePlacement *sdkapi.NodePlacement, replicas int32) *appsv1.Deployment {
 	defaultMode := corev1.ConfigMapVolumeSourceDefaultMode
-	deployment := utils.CreateDeployment(controllerResourceName, "app", "containerized-data-importer", common.ControllerServiceAccountName, imagePullSecrets, int32(1), infraNodePlacement)
+	deployment := utils.CreateDeployment(common.CDIControllerResourceName, common.CDIComponentLabel, common.CDIControllerResourceName, common.ControllerServiceAccountName, imagePullSecrets, int32(1), infraNodePlacement)
```
akalenyu, line 174: So what happens is that this is breaking the upgrade path (which is exercised by CI):
```
2024-01-23T11:52:20Z	ERROR	cdi-operator		{"Request.Namespace": "", "Request.Name": "cdi", "error": "Deployment.apps \"cdi-deployment\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\"cdi.kubevirt.io\":\"cdi-deployment\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable"}
```

This PR changes the immutable deployment field `spec.selector` and thus the operator fails to reconcile the (already existing) cdi-deployment resource that is carried over from the previous version.

My suggestion is to stick to the existing selector we have

mhenriks: @yaroslavborbat you can test upgrade locally by setting

```bash
export UPGRADE_FROM=v1.58.0

make cluster-sync
```

mhenriks:
@yaroslavborbat did a quick once-over and did not see anything concerning.  Nice job!

pkg/operator/resources/namespaced/controller.go
yaroslavborbat: Thank you all for your help. I fixed the problem and now all the tests have been successful

mhenriks:
/lgtm

cc @akalenyu

mhenriks:
I don't love functional tests that mutate the install, but maybe it makes sense here?  What are your thoughts @akalenyu?

tests/operator_test.go
```diff
@@ -1118,9 +1227,9 @@ var _ = Describe("ALL Operator tests", func() {
 				// Deployment
 				verifyPodPriorityClass(cdiDeploymentPodPrefix, string(osUserCrit.Name), common.CDILabelSelector)
 				// API server
-				verifyPodPriorityClass(cdiApiServerPodPrefix, string(osUserCrit.Name), "")
+				verifyPodPriorityClass(cdiApiServerPodPrefix, string(osUserCrit.Name), common.CDILabelSelector)
 				// Upload server
-				verifyPodPriorityClass(cdiUploadProxyPodPrefix, string(osUserCrit.Name), "")
+				verifyPodPriorityClass(cdiUploadProxyPodPrefix, string(osUserCrit.Name), common.CDILabelSelector)
```
akalenyu, line 1232: I am fine with this, correct me if I'm wrong, but this is not strictly needed, and was not broken before this PR?
(just about being more precise?)

akalenyu:
> I don't love functional tests that mutate the install, but maybe it makes sense here? What are your thoughts @akalenyu?

I have to check if there's an AfterEach that protects from a failure in this test borking the test run

tests/operator_test.go
```diff
@@ -879,6 +880,114 @@ var _ = Describe("ALL Operator tests", func() {
...
+						if err != nil || depl.GetAnnotations()[cc.AnnCdiCustomizeComponentHash] == "" {
+							return false
+						}
+					}
+					By("Patches applied")
+					return true
+				}, 5*time.Minute, 1*time.Second).Should(BeTrue())
+
+				By("Verify patches of cdi-apiserver")
+				Eventually(func() bool {
```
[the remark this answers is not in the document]
yaroslavborbat: i did it

pkg/common/common.go
```diff
@@ -24,6 +24,15 @@ const (
 	// CDIOperatorName is the CDI operator name
 	CDIOperatorName = "cdi-operator"
 
+	// CDIControllerResourceName is the CDI controller resource name
+	CDIControllerResourceName = "cdi-deployment"
```
nunnatsa, line 28: this change actually rename the existing deployment from "cdi-controller" to "cdi-deplyment".

@yaroslavborbat is it on purpose? A typo? This change is not documented in the PR description.

ruthcamarillo:
Anybody else having issues with adding the patch to their cdi and having it roll out to all the components? 
When I edit and add this patch it is not recognized and my edits are canceled, and when I just added the readOnlyRootFilesystem to the deployment it rolled out but did not update the rest of the componenets

yaroslavborbat:
@ruthcamarillo Hi,
The problem is that the cdi-deployment creates files in the file system, so if you put it in read-only mode, it won't start
https://github.com/kubevirt/containerized-data-importer/blob/d08e24523d904cb70324fdf57535c100b81ce8ca/cmd/cdi-controller/controller.go#L348
This can be solved by adding empty-dir mounted in /tmp.
```yaml
    - patch: |
        {
          "spec": {
            "template": {
              "spec": {
                "containers": [
                  {
                    "name": "cdi-controller",
                    "securityContext": {
                      "readOnlyRootFilesystem": true
                    },
                    "volumeMounts": [
                      {
                        "mountPath": "/tmp",
                        "name": "tmp"
                      }
                    ]
                  }
                ],
                "volumes": [
                  {
                    "name": "tmp",
                    "emptyDir": {}
                  }
                ]
              }
            }
          }
        }
      resourceName: cdi-deployment
      resourceType: Deployment
      type: strategic
```
I checked on this patch and everything worked for me.
I think other components may have the same problem and it can be solved in a similar way

PR #3072, opened 2024-01-17
Fix DataImportCron import DataVolume creation when last import is not found

arnongilboa:
The `imports` local variable was not updated correctly, and later referenced, so another reconcile was required for the DV to be created.

Fixes CNV-33973

**Release note**:
```release-note
Fix DataImportCron import DataVolume creation when last import is not found
```


awels:
/lgtm
@akalenyu can you review as well

PR #3078, opened 2024-01-18
update k8s & related libraries to 1.28

akalenyu:
There are a few non trivial changes laid out in the commit descriptions

Fixes #

**Release note**:

```release-note
k8s 1.28 dependencies
```


go.mod
```diff
@@ -170,36 +169,36 @@ replace (
 	github.com/openshift/library-go => github.com/mhenriks/library-go v0.0.0-20230310153733-63d38b55bd5a
```
mhenriks, line 169: `openshift/api` (looks like `release-4.15` is the appropriate branch for k8s 1.28) and `openshift/library-go` should probably be updated as well.  library-go is tricky as this was never merged:  https://github.com/openshift/library-go/pull/1486

This is obviously pretty annoying.  I guess I'll update my fork again for now

mhenriks (commented):
Should we do 1.29?  That is what openshift is using for 4.16

akalenyu:
> Should we do 1.29? That is what openshift is using for 4.16

Yeah, I plan to do that as well but we'll need a bump to go 1.21 first
Anyway, this PR is too big as it is, it would be nice to do that in follow up

mhenriks:
> > Should we do 1.29? That is what openshift is using for 4.16
> 
> Yeah, I plan to do that as well but we'll need a bump to go 1.21 first Anyway, this PR is too big as it is, it would be nice to do that in follow up

Why can't we go straight to 1.29?

akalenyu:
> > > Should we do 1.29? That is what openshift is using for 4.16
> > 
> > 
> > Yeah, I plan to do that as well but we'll need a bump to go 1.21 first Anyway, this PR is too big as it is, it would be nice to do that in follow up
> 
> Why can't we go straight to 1.29?

Because we are still on go 1.20

go.mod
akalenyu: Let me know when that is done and I can update both

mhenriks: Looks like last week the change was made on master and backported to release-4.16 in another PR (https://github.com/openshift/library-go/commit/434fa21f5e457d51f436c4b3804c7803db1da42d)

So if you want to go straight to 1.29 you can use release-4.16 branch in library-go
If you insist on doing 1.28 first (sounds like more work to me) then you can use `github.com/mhenriks/library-go export-certrotation-4.15`

cmd/cdi-operator/operator.go
```diff
@@ -84,7 +85,11 @@ func main() {
 	}
 
 	managerOpts := manager.Options{
-		Namespace:                  namespace,
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				namespace: {},
```
alromeros, line 90: No need to include `namespace` anywhere?

cmd/cdi-controller/controller.go
```diff
@@ -372,24 +385,53 @@ func getTokenPrivateKey() *rsa.PrivateKey {
...
-func getNewManagerCache(cdiNamespace string) cache.NewCacheFunc {
+func getCacheOptions(apiClient client.Client, cdiNamespace string) cache.Options {
 	namespaceSelector := fields.Set{"metadata.namespace": cdiNamespace}.AsSelector()
-	return cache.BuilderWithOptions(
-		cache.Options{
-			SelectorsByObject: cache.SelectorsByObject{
-				&networkingv1.Ingress{}: {
-					Field: namespaceSelector,
-				},
-				&routev1.Route{}: {
```
alromeros, line 383: Just making sure, we don't want route anymore?

alromeros (commented):
Awesome, thanks for dividing it into commits. Just a couple of questions.

cmd/cdi-operator/operator.go
akalenyu: wdym? it's included `namespace: {},`

cmd/cdi-controller/controller.go
akalenyu: We conditionally add `&routev1.Route{}` if running on openshift

akalenyu: here's some context - https://github.com/kubernetes-sigs/controller-runtime/issues/2456

cmd/cdi-operator/operator.go
alromeros: I mean the namespace we get above (`namespace := util.GetNamespace()`). Just wondering if we don't have to assign it somewhere as we did before.

akalenyu: it's included as the key in `map[string]cache.Config`

go.mod
akalenyu: We can't - https://github.com/openshift/library-go/blob/master/go.mod#L3, same for ctrl runtime

akalenyu: Do you mean a few separate PRs to bump go/update builder image, right?

akalenyu: Anyway, this is huge as it is, if more (1.28->1.29) adaptations are needed this PR will just be all over the place
so I prefer to take the 1.28->1.29 stuff out

alromeros:
I agree this PR is already too large, I'm ok with doing the 1.28->1.29 stuff in a different one
/lgtm

PR #3080, opened 2024-01-19
Allow Deployed CDI to get out of Error Phase

mhenriks:

Was impossible for CDI to get out of `Error` phase once in it because `GetActiveCDI()` filtered out resources in `Error` state.  This change allows the function to return success if there is a single CDI in error state.

Fixes #

https://issues.redhat.com/browse/CNV-36904

If you look at attached issues, what appears to have happened is that CDI somehow got in an error state (not sure how but not that important).  Once in error state, cdi-operator reconcile loop would always error out.  Eventually certs would expire and you can't do anything CDI related.  Not good!  This makes things resilient.

**Release note**:

```release-note
BugFix:  Allow Deployed CDI to get out of Error Phase
```

### DIFF
--- a/pkg/controller/common/util_test.go
+++ b/pkg/controller/common/util_test.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/api"
 )
 
 var _ = Describe("GetRequestedImageSize", func() {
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/cr-manager.go
```diff
@@ -72,7 +72,7 @@ func (r *ReconcileCDI) GetDependantResourcesListObjects() []client.ObjectList {
 func (r *ReconcileCDI) IsCreating(_ client.Object) (bool, error) {
 	configMap, err := r.getConfigMap()
 	if err != nil {
-		return true, nil
```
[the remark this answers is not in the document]
mhenriks: Yeah that is my assumption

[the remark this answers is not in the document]
akalenyu: nice catch

pkg/controller/common/util_test.go
```diff
@@ -234,6 +235,69 @@ var _ = Describe("Rebind", func() {
...
+			)
+			cdi, err := GetActiveCDI(context.Background(), client)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cdi).ToNot(BeNil())
+		})
+
+		It("Should return success with single active one error", func() {
+			client := CreateClient(
+				createCDI("cdi1", sdkapi.PhaseDeployed),
+				createCDI("cdi2", sdkapi.PhaseError),
```
akalenyu, line 272: Do we want to ever allow more than 1 CDI CR?

mhenriks: Given that CDI resource is cluster scope, two in "Deployed" state would be very bad

akalenyu: I am saying don't allow 2 of any state

mhenriks: We can't currently stop CDI resources from being created.  That would require that cdi-operator run a validating webhook.  And even with a webhook it is an inherently racy operation

akalenyu: Yeah I mean just return err in this case so we don't ever get further than GetActiveCDI
when that's the case

mhenriks: How would that be better?  If CDI A is successfully deployed at t0 and someone creates CDI B at t1 why break reconcile for CDI A?

akalenyu: I believe this assumption is made multiple times across the project so if you do allow this,
bad things will happen for CDI A regardless

mhenriks: Please explain what bad things would happen and how?

mhenriks: I wrote that operator and that function with the intention that there would only be one "active" (deployed phase) CDI and there could be many CDIs in error state

akalenyu: oh, I see, even prior to this PR we did not error over this case

akalenyu (commented):
/approve

PR #3081, opened 2024-01-19
Change CDI webhook failure policy to Ignore

mhenriks:

If cdi-apiserver gets in a bad state it may not allow CDI to be deleted.  Maybe when CDI webhook is served by cdi-operator we can put this back to Fail

Fixes #

https://issues.redhat.com/browse/CNV-36904

In the attached case, webhooks had to be manually deleted in order to uninstall CDI

**Release note**:

```release-note
Change CDI webhook failure policy to Ignore
```


akalenyu:
so the bug is about random unavailability of the apiserver?

akalenyu:
/approve

as we anyway want users to be able to get out of these situations

PR #3096, opened 2024-02-01
Fix StorageProfile ClaimPropertySets validation

arnongilboa:
We have no reason to allow setting `StorageProfile` `ClaimPropertySet` missing either `VolumeMode` or `AccessModes`.

Jira-ticket: https://issues.redhat.com/browse/CNV-37991

**Release note**:
```release-note
BugFix: StorageProfile ClaimPropertySets validation
```

### DIFF
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -101,8 +102,8 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
 
 	if len(storageProfile.Spec.ClaimPropertySets) > 0 {
 		for _, cps := range storageProfile.Spec.ClaimPropertySets {
-			if len(cps.AccessModes) == 0 && cps.VolumeMode != nil {
-				err = fmt.Errorf("must provide access mode for volume mode: %s", *cps.VolumeMode)
+			if cps.VolumeMode == nil || len(cps.AccessModes) == 0 {
+				err = errors.New("each ClaimPropertySet must provide both volume mode and access modes")
 				log.Error(err, "Unable to update StorageProfile")
 				return reconcile.Result{}, err
 			}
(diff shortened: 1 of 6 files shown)


pkg/controller/storageprofile-controller.go
```diff
@@ -101,8 +102,8 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
 
 	if len(storageProfile.Spec.ClaimPropertySets) > 0 {
 		for _, cps := range storageProfile.Spec.ClaimPropertySets {
-			if len(cps.AccessModes) == 0 && cps.VolumeMode != nil {
-				err = fmt.Errorf("must provide access mode for volume mode: %s", *cps.VolumeMode)
+			if cps.VolumeMode == nil || len(cps.AccessModes) == 0 {
```
mhenriks, line 105: `cps.VolumeMode == nil` is redundant.  If it is nil len(cps.AccessModes) will return 0

arnongilboa: How is `cps.VolumeMode == nil` related to  `len(cps.AccessModes)`?

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -445,12 +445,11 @@ type StorageProfileStatus struct {
 type ClaimPropertySet struct {
 	// AccessModes contains the desired access modes the volume should have.
 	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#access-modes-1
-	// +optional
-	AccessModes []corev1.PersistentVolumeAccessMode `json:"accessModes,omitempty" protobuf:"bytes,1,rep,name=accessModes,casttype=PersistentVolumeAccessMode"`
+	AccessModes []corev1.PersistentVolumeAccessMode `json:"accessModes"`
```
arnongilboa, line 448: afaik there's no `kubebuilder:validation` marker for validating enum values for a slice (`AccessModes`) like addded for `VolumeMode`. The only ways I can think of is either by a validating webhook, or by adding an alias type, e.g.:
```go
+// +kubebuilder:validation:Enum="ReadWriteOnce";"ReadOnlyMany";"ReadWriteMany";"ReadWriteOncePod"
+type ClaimPropertySetAccessMode corev1.PersistentVolumeAccessMode
```
Both solutions aren't so nice.

awels: I don't mind the second option, creating a type that is just the same, and adding the validation there.

awels (approved):

pkg/controller/storageprofile-controller.go
mhenriks: lol it's not

PR #3101, opened 2024-02-14
feat: Make importer datasource communication explicit

0xFelix:

Make the communication of datasources in the importer explicit by adding
a GetTerminationMessage method to the DataSourceInterface.
Then use this method to communicate additional information to the import
controller once the importer pod has terminated, instead of writing
additional data to the termination message in the Close method of
datasources.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -1,6 +1,8 @@
 package common
 
 import (
+	"encoding/json"
+	"fmt"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -327,4 +329,32 @@ var SyncUploadFormPaths = []string{
 var AsyncUploadFormPaths = []string{
 	UploadFormAsync,
 	"/v1alpha1/upload-form-async",
+}
+
+// VddkInfo holds VDDK version and connection information returned by an importer pod
+type VddkInfo struct {
+	Version string
+	Host    string
+}
+
+// TerminationMessage contains data to be serialized and used as the termination message of the importer.
+type TerminationMessage struct {
+	ScratchSpaceRequired *bool             `json:"scratchSpaceRequired,omitempty"`
+	PreallocationApplied *bool             `json:"preallocationApplied,omitempty"`
+	VddkInfo             *VddkInfo         `json:"vddkInfo,omitempty"`
+	Labels               map[string]string `json:"labels,omitempty"`
+}
+
+func (it *TerminationMessage) String() (string, error) {
+	msg, err := json.Marshal(it)
+	if err != nil {
+		return "", err
+	}
+
+	// Messages longer than 4096 are truncated by kubelet
+	if length := len(msg); length > 4096 {
+		return "", fmt.Errorf("Termination message length %d exceeds maximum length of 4096 bytes", length)
+	}
+
+	return string(msg), nil
 }
--- a/pkg/common/common_test.go
+++ b/pkg/common/common_test.go
@@ -1,0 +1,43 @@
+package common
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/utils/ptr"
+)
+
+var _ = Describe("TerminationMessage", func() {
+	It("Should successfully serialize a TerminationMessage", func() {
+		termMsg := TerminationMessage{
+			VddkInfo: &VddkInfo{
+				Version: "testversion",
+				Host:    "testhost",
+			},
+			Labels: map[string]string{
+				"testlabel": "testvalue",
+			},
+			PreallocationApplied: ptr.To(true),
+		}
+
+		serialized, err := termMsg.String()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(serialized).To(Equal(`{"preallocationApplied":true,"vddkInfo":{"Version":"testversion","Host":"testhost"},"labels":{"testlabel":"testvalue"}}`))
+	})
+
+	It("Should fail if serialized data is longer than 4096 bytes", func() {
+		const length = 5000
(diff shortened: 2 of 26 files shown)


awels:
Sorry haven't had a chance to really look at this. I will try to do so later today.

pkg/common/common_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("TerminationMessage", func() {
+	It("Should fail if serialized data is longer than 4096 bytes", func() {
```
awels, line 11: Testing any of the happy paths?

pkg/controller/import-controller.go
```diff
@@ -366,7 +366,12 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 
 	log.V(1).Info("Updating PVC from pod")
 	anno := pvc.GetAnnotations()
-	setAnnotationsFromPodWithPrefix(anno, pod, cc.AnnRunningCondition)
+
+	termMsg, err := parseTerminationMessage(pod)
+	if err != nil {
+		log.V(3).Info("Ignoring failure to parse termination message")
```
awels, line 372: Maybe at least print the actual reason the parsing failed?

awels (commented):
Looks pretty good a few questions.

pkg/controller/import-controller.go
0xFelix: ACK

pkg/common/common_test.go
0xFelix: Added a test for the happy path.

0xFelix:
@awels Addressed your comments.

pkg/common/common.go
```diff
@@ -329,3 +331,30 @@ var AsyncUploadFormPaths = []string{
...
 }
+
+// VddkInfo holds VDDK version and connection information returned by an importer pod
+type VddkInfo struct {
+	Version string
+	Host    string
+}
+
+// TerminationMessage contains data to be serialized and used as the termination message of the importer.
+type TerminationMessage struct {
```
alromeros, line 342: This is a great idea, it could come in handy to transfer unexpected but non-error import info such as some odd content types. We had an attempt to improve the import behavior when importing html/text but the solution was a bit lackluster: https://github.com/kubevirt/containerized-data-importer/pull/2909. Just a comment though, that would be out of scope for the PR.

alromeros (commented):
I think this is a great idea, thanks for opening the PR. It looks good to me but as discussed in https://github.com/kubevirt/containerized-data-importer/pull/3116 I think we should first merge the importer bugfix and adapt the behavior here.

tests/datavolume_test.go
```diff
@@ -1013,7 +1013,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 				runningCondition: &cdiv1.DataVolumeCondition{
 					Type:    cdiv1.DataVolumeRunning,
 					Status:  v1.ConditionFalse,
-					Message: "Import Complete; VDDK: {\"Version\":\"1.2.3\",\"Host\":\"esx.test\"}",
```
akalenyu, line 1016: Probably missing something trivial, why are we dropping this from the expected output, shouldn't the version and host get propagated correctly?

0xFelix: After importing the importer Pod writes the extended termination message to its termination log. Once the import controller sees it and was sucessfully able to parse the message, the value written to the message annotation will be just `Import Complete` instead of the full serialized JSON string. The import controller will still set the appropriate annotations for VDDK version and host.

pkg/common/common.go
0xFelix: Yes, exactly what I'm doing in #3103

tests/datavolume_test.go
akalenyu: I see. @mrnold hey! do you happen to know if anyone was treating this message as an API on the CDI consuming side?

pkg/controller/import-controller.go
```diff
@@ -366,28 +366,32 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
...
+	if scratchSpaceRequired {
+		log.V(1).Info("Pod requires scratch space, terminating pod, and restarting with scratch space", "pod.Name", pod.Name)
+	}
+
 	if pod.Status.ContainerStatuses != nil &&
-		pod.Status.ContainerStatuses[0].State.Terminated != nil {
-		if message := pod.Status.ContainerStatuses[0].State.Terminated.Message; strings.Contains(message, common.ScratchSpaceRequired) {
-			log.V(1).Info("Pod requires scratch space, terminating pod, and restarting with scratch space", "pod.Name", pod.Name)
-			scratchExitCode = true
-			anno[cc.AnnRequiresScratch] = "true"
```
akalenyu, line 377: why drop setting this annotation? AFAIK it's needed for things to continue properly

0xFelix: Move it to `utils.go` / `setAnnotationsFromPodWithPrefix`, where other annotations are set.

https://github.com/kubevirt/containerized-data-importer/blob/c88d3ba94fd3253e9c31eeb7ceb728f1f95bcf66/pkg/controller/util.go#L305

akalenyu: :+1:

akalenyu (approved):
Looks good! thank you

0xFelix:
zv

mhenriks:
Very cool

/approve

PR #3103, opened 2024-02-21
feat(import): Set PVC labels derived from env vars on containerdisks

0xFelix:

This allows to set labels on PVCs derived from env vars containing `KUBEVIRT_IO_` on containerdisks. Env vars on containerdisks were chosen because they can be accessed in both pullMethods pod and node. While the pod pullMethod is talking to the registry and can query for the `Labels` field in the image config, the pullMethod node cannot access the image config. However, the server running in the context of the containerdisk can still see all associated env vars of the containerdisk.

```
skopeo inspect docker://quay.io/containerdisks/fedora
{
    "Name": "quay.io/containerdisks/fedora",
    "Digest": "sha256:bbc58d0576a236bafe4917c0fbe40701c5164344a499af257ab61b29dacde57c",
    "RepoTags": [
        "35-2111250200",
        "35-2112061510",
        "35-1.2",
        "35",
        "36-2205130206",
        "36-1.5",
        "36",
        "37-2211160206",
        "37-1.7",
        "37",
        "38-2304200207",
        "39-2311080206",
        "39-2401311200",
        "39-1.5",
        "39",
        "latest",
        "38-2401311200",
        "38-1.6",
        "38"
    ],
    "Created": "0001-01-01T00:00:00Z",
    "DockerVersion": "",
    "Labels": {
        "instancetype.kubevirt.io/default-instancetype": "u1.small",
        "instancetype.kubevirt.io/default-preference": "fedora",
        "shasum": "ab5be5058c5c839528a7d6373934e0ce5ad6c8f80bd71ed3390032027da52f37"
    },
    "Architecture": "amd64",
    "Os": "",
    "Layers": [
        "sha256:f06e97b1240fefbb716a8ef4158fe20c0cf92057e625fb79c539334ba1bbd5ee"
    ],
    "LayersData": [
        {
            "MIMEType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "Digest": "sha256:f06e97b1240fefbb716a8ef4158fe20c0cf92057e625fb79c539334ba1bbd5ee",
            "Size": 542252872,
            "Annotations": null
        }
    ],
    "Env": [
        "SHASUM=ab5be5058c5c839528a7d6373934e0ce5ad6c8f80bd71ed3390032027da52f37",
        "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_INSTANCETYPE=u1.small",
        "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_PREFERENCE=fedora"
    ]
}
```

Create a DataVolume from that:

```
kind: DataVolume
metadata:
  name: "example-import-dv"
  annotations:
spec:
  source:
      registry:
         url: "docker://quay.io/containerdisks/fedora:latest"
  storage:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: "10Gi"
```

The resulting PVC will look like this, pay attention to its labels:

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    cdi.kubevirt.io/storage.condition.running: "false"
    cdi.kubevirt.io/storage.condition.running.message: Import Complete
    cdi.kubevirt.io/storage.condition.running.reason: Completed
    cdi.kubevirt.io/storage.contentType: kubevirt
    cdi.kubevirt.io/storage.import.endpoint: docker://quay.io/containerdisks/fedora:latest
    cdi.kubevirt.io/storage.import.importPodName: importer-example-import-dv
    cdi.kubevirt.io/storage.import.requiresScratch: "false"
    cdi.kubevirt.io/storage.import.source: registry
    cdi.kubevirt.io/storage.pod.phase: Succeeded
    cdi.kubevirt.io/storage.pod.restarts: "0"
    cdi.kubevirt.io/storage.preallocation.requested: "false"
    cdi.kubevirt.io/storage.usePopulator: "false"
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
  labels:
    alerts.k8s.io/KubePersistentVolumeFillingUp: disabled
    instancetype.kubevirt.io/default-instancetype: u1.small
    instancetype.kubevirt.io/default-preference: fedora
  name: example-import-dv
[...]
```

Fixes #

This is based on #3101, merge that first

**Release note**:

```release-note
The containerdisk importer can now set PVC labels derived from env vars on containerdisks
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -410,11 +410,14 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 		delete(anno, cc.AnnBoundConditionReason)
 	}
 
+	if pvc.GetLabels() == nil {
+		pvc.SetLabels(make(map[string]string, 0))
+	}
 	if !checkIfLabelExists(pvc, common.CDILabelKey, common.CDILabelValue) {
-		if pvc.GetLabels() == nil {
-			pvc.SetLabels(make(map[string]string, 0))
-		}
 		pvc.GetLabels()[common.CDILabelKey] = common.CDILabelValue
+	}
+	if cc.IsPVCComplete(pvc) {
+		setLabelsFromTerminationMessage(pvc.GetLabels(), termMsg)
 	}
 
 	if !reflect.DeepEqual(currentPvcCopy, pvc) {
--- a/pkg/controller/populators/import-populator.go
+++ b/pkg/controller/populators/import-populator.go
@@ -156,9 +156,11 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 			break
 		}
 
-		// Once the import is succeeded, we rebind the PV from PVC' to target PVC
-		if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvc); err != nil {
-			return reconcile.Result{}, err
+		if cc.IsPVCComplete(pvcPrime) && cc.IsUnbound(pvc) {
+			// Once the import is succeeded, we rebind the PV from PVC to target PVC
+			if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvc); err != nil {
+				return reconcile.Result{}, err
+			}
 		}
 	}
 
--- a/pkg/controller/populators/import-populator.go
+++ b/pkg/controller/populators/import-populator.go
@@ -139,7 +139,7 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		if err = cc.MaybeSetPvcMultiStageAnnotation(pvcPrime, r.getCheckpointArgs(source)); err != nil {
 			return reconcile.Result{}, err
 		}
-		if err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
+		if _, err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
 			return reconcile.Result{}, err
 		}
 		// We requeue to keep reporting progress
@@ -162,11 +162,13 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		}
 	}
 
-	err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations)
-	if err != nil {
+	if pvcCopy, err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
 		return reconcile.Result{}, err
 	}
 	if cc.IsPVCComplete(pvcPrime) && !cc.IsMultiStageImportInProgress(pvc) {
+		if err = r.updatePVCWithPVCPrimeLabels(pvcCopy, pvcPrime.GetLabels()); err != nil {
+			return reconcile.Result{}, err
+		}
 		r.recorder.Eventf(pvc, corev1.EventTypeNormal, importSucceeded, messageImportSucceeded, pvc.Name)
 	}
 
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -288,12 +288,17 @@ func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorCon
 	}
 
 	// Each populator reconciles the target PVC in a different way
-	if cc.IsUnbound(pvc) || !cc.IsPVCComplete(pvc) || cc.IsMultiStageImportInProgress(pvc) {
-		return populator.reconcileTargetPVC(pvc, pvcPrime)
+	res, err := populator.reconcileTargetPVC(pvc, pvcPrime)
+	if err != nil {
+		return res, err
 	}
 
-	// Making sure to clean PVC'
-	return r.reconcileCleanup(pvcPrime)
+	// Making sure to clean PVC once it is complete
+	if cc.IsPVCComplete(pvc) && !cc.IsMultiStageImportInProgress(pvc) {
+		res, err = r.reconcileCleanup(pvcPrime)
+	}
+
+	return res, err
 }
 
 func (r *ReconcilerBase) reconcileCommon(pvc *corev1.PersistentVolumeClaim, populator populatorController, pvcNameLogger logr.Logger) (*corev1.PersistentVolumeClaim, error) {
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -19,6 +19,7 @@ package populators
 import (
 	"context"
 	"reflect"
+	"regexp"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -58,6 +59,10 @@ const (
 	dataSourceRefField = "spec.dataSourceRef"
 
 	uidField = "metadata.uid"
+)
+
+var (
+	validLabelsMatch = regexp.MustCompile(`^([\w.]+\.kubevirt.io|kubevirt.io)/\w+$`)
 )
 
 // Interface to store populator-specific methods
@@ -223,7 +228,7 @@ var desiredAnnotations = []string{cc.AnnPodPhase, cc.AnnPodReady, cc.AnnPodResta
 	cc.AnnPreallocationRequested, cc.AnnPreallocationApplied, cc.AnnCurrentCheckpoint, cc.AnnMultiStageImportDone,
 	cc.AnnRunningCondition, cc.AnnRunningConditionMessage, cc.AnnRunningConditionReason}
 
-func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.PersistentVolumeClaim, updateFunc updatePVCAnnotationsFunc) error {
+func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.PersistentVolumeClaim, updateFunc updatePVCAnnotationsFunc) (*corev1.PersistentVolumeClaim, error) {
 	pvcCopy := pvc.DeepCopy()
 	for _, ann := range desiredAnnotations {
 		if value, ok := pvcPrime.GetAnnotations()[ann]; ok {
@@ -241,10 +246,25 @@ func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.
 	if !reflect.DeepEqual(pvc.ObjectMeta, pvcCopy.ObjectMeta) {
 		err := r.client.Update(context.TODO(), pvcCopy)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
 
+	return pvcCopy, nil
+}
+
+func (r *ReconcilerBase) updatePVCWithPVCPrimeLabels(pvc *corev1.PersistentVolumeClaim, pvcPrimeLabels map[string]string) error {
+	pvcCopy := pvc.DeepCopy()
+	for label, value := range pvcPrimeLabels {
+		if _, found := pvcCopy.GetLabels()[label]; !found && validLabelsMatch.MatchString(label) {
+			cc.AddLabel(pvcCopy, label, value)
+		}
+	}
+	if !reflect.DeepEqual(pvc.ObjectMeta, pvcCopy.ObjectMeta) {
+		if err := r.client.Update(context.TODO(), pvcCopy); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
--- a/pkg/controller/populators/upload-populator.go
+++ b/pkg/controller/populators/upload-populator.go
@@ -159,9 +159,11 @@ func (r *UploadPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		// We'll get called later once it succeeds
 		r.recorder.Eventf(pvc, corev1.EventTypeWarning, errUploadFailed, fmt.Sprintf(messageUploadFailed, pvc.Name))
 	case string(corev1.PodSucceeded):
-		// Once the upload is succeeded, we rebind the PV from PVC' to target PVC
-		if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvcCopy); err != nil {
-			return reconcile.Result{}, err
+		if cc.IsPVCComplete(pvcPrime) && cc.IsUnbound(pvc) {
+			// Once the upload is succeeded, we rebind the PV from PVC' to target PVC
+			if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvcCopy); err != nil {
+				return reconcile.Result{}, err
+			}
 		}
 	}
 
--- a/pkg/controller/populators/upload-populator.go
+++ b/pkg/controller/populators/upload-populator.go
@@ -165,7 +165,7 @@ func (r *UploadPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		}
 	}
 
-	err := r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateUploadAnnotations)
+	_, err := r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateUploadAnnotations)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -332,6 +332,17 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *v1.Pod, termMs
 	}
 }
 
+func setLabelsFromTerminationMessage(labels map[string]string, termMsg *common.TerminationMessage) {
+	if labels == nil || termMsg == nil {
+		return
+	}
+	for k, v := range termMsg.Labels {
+		if _, found := labels[k]; !found {
+			labels[k] = v
+		}
+	}
+}
+
 func simplifyKnownMessage(msg string) string {
 	if strings.Contains(msg, "is larger than the reported available") ||
 		strings.Contains(msg, "no space left on device") ||
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/fs"
@@ -196,7 +197,19 @@ func (hs *HTTPDataSource) GetURL() *url.URL {
 
 // GetTerminationMessage returns data to be serialized and used as the termination message of the importer.
 func (hs *HTTPDataSource) GetTerminationMessage() *common.TerminationMessage {
-	return nil
+	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod != string(cdiv1.RegistryPullNode) {
+		return nil
+	}
+
+	info, err := getServerInfo(hs.ctx, fmt.Sprintf("%s://%s/info", hs.endpoint.Scheme, hs.endpoint.Host))
+	if err != nil {
+		klog.Errorf("%+v", err)
+		return nil
+	}
+
+	return &common.TerminationMessage{
+		Labels: envsToLabels(info.Env),
+	}
 }
 
 // Close all readers.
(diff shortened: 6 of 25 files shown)


pkg/controller/populators/populator-base.go
```diff
@@ -241,11 +246,26 @@ func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.
...
-			return err
+			return nil, err
 		}
 	}
 
-	return nil
+	return pvcCopy, nil
+}
+
+func (r *ReconcilerBase) updatePVCWithPVCPrimeLabels(pvc, pvcPrime *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
```
ksimon1, line 256: Does it have to return PVC? The pvc is not used in this commit (only err is used)

pkg/controller/util.go
```diff
@@ -332,6 +332,14 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *v1.Pod, termMs
 	}
 }
 
+func setLabelsFromTerminationMessage(labels map[string]string, termMsg *common.TerminationMessage) {
+	if labels != nil && termMsg != nil {
```
ksimon1, line 336:
```
if labels == nil || termMsg == nil {
  return
}
```
would be more readable

pkg/importer/http-datasource.go
```diff
@@ -196,7 +197,19 @@ func (hs *HTTPDataSource) GetURL() *url.URL {
 
 // GetTerminationMessage returns data to be serialized and used as the termination message of the importer.
 func (hs *HTTPDataSource) GetTerminationMessage() *common.TerminationMessage {
-	return nil
+	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod != string(cdiv1.RegistryPullNode) {
```
ksimon1, line 200: If the func ends with error can it somehow affect the code?

pkg/controller/populators/populator-base.go
0xFelix: Not strictly, but I stuck to the coding style of `updatePVCWithPVCPrimeAnnotation` above.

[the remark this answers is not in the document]
0xFelix: It could be simplified, I just made it similar to `updatePVCWithPVCPrimeAnnotation`. Will update it.

pkg/importer/http-datasource.go
0xFelix: The termination message will be `nil`, which is handled by the import controller. It just would lead to the additional information provided by the HTTP datasource getting lost (in this case the labels).

akalenyu:
> /retest-required

These look like legitimate flakes: [#1](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3103/pull-containerized-data-importer-e2e-istio/1768213306465587200), [#2](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3103/pull-containerized-data-importer-e2e-upg/1768213307702906880)

0xFelix:
Yeah, I'm wondering what could be the issue. Not able to reproduce it locally right now.

pkg/importer/http-datasource.go
```diff
@@ -196,7 +197,19 @@ func (hs *HTTPDataSource) GetURL() *url.URL {
...
+		return nil
+	}
+
+	info, err := getServerInfo(hs.ctx, fmt.Sprintf("%s://%s/info", hs.endpoint.Scheme, hs.endpoint.Host))
+	if err != nil {
+		klog.Errorf("%+v", err)
+		return nil
```
akalenyu, line 207: Do we want to differentiate between fatal errors such as `no such endpoint` and other intermittent network errors that can be retried by restarting the pod?

Not insisting just have to raise this

pkg/controller/populators/import-populator.go
```diff
@@ -162,11 +162,13 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		}
 	}
 
-	err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations)
-	if err != nil {
+	if pvcCopy, err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
 		return reconcile.Result{}, err
 	}
 	if cc.IsPVCComplete(pvcPrime) && !cc.IsMultiStageImportInProgress(pvc) {
+		if err = r.updatePVCWithPVCPrimeLabels(pvcCopy, pvcPrime.GetLabels()); err != nil {
```
akalenyu, line 171: As we discussed offline we probably have a bigger issue here in that we'd never make our way back to this part on a requeue, since the PVC may already be succeeded.
I think we can let the loop through if labels aren't there for now and fix properly later

pkg/controller/import-controller.go
```diff
@@ -410,12 +410,15 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
...
+	if pvc.GetLabels() == nil {
+		pvc.SetLabels(make(map[string]string, 0))
+	}
 	if !checkIfLabelExists(pvc, common.CDILabelKey, common.CDILabelValue) {
-		if pvc.GetLabels() == nil {
-			pvc.SetLabels(make(map[string]string, 0))
-		}
 		pvc.GetLabels()[common.CDILabelKey] = common.CDILabelValue
 	}
+	if cc.IsPVCComplete(pvc) {
```
akalenyu, line 419: hopefully, there's not a similar problem in the import controller as with the import populator where this will not be executed on an update conflict requeue

pkg/controller/populators/populator-base.go
```diff
@@ -223,7 +228,7 @@ var desiredAnnotations = []string{cc.AnnPodPhase, cc.AnnPodReady, cc.AnnPodResta
 	cc.AnnPreallocationRequested, cc.AnnPreallocationApplied, cc.AnnCurrentCheckpoint, cc.AnnMultiStageImportDone,
 	cc.AnnRunningCondition, cc.AnnRunningConditionMessage, cc.AnnRunningConditionReason}
 
-func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.PersistentVolumeClaim, updateFunc updatePVCAnnotationsFunc) error {
+func (r *ReconcilerBase) updatePVCWithPVCPrimeAnnotations(pvc, pvcPrime *corev1.PersistentVolumeClaim, updateFunc updatePVCAnnotationsFunc) (*corev1.PersistentVolumeClaim, error) {
```
akalenyu, line 231: kind of weird to have different return values between `updatePVCWithPVCPrimeAnnotations` and `updatePVCWithPVCPrimeLabels` no?

pkg/importer/http-datasource.go
```diff
@@ -196,7 +197,19 @@ func (hs *HTTPDataSource) GetURL() *url.URL {
...
+	}
+
+	return &common.TerminationMessage{
+		Labels: envsToLabels(info.Env),
```
akalenyu, line 211: any reason to worry about `nil`s here? (and later)
from a quick look I don't think so but may deserve a unit test

tests/import_test.go
```diff
@@ -2134,6 +2134,67 @@ var _ = Describe("Import populator", func() {
...
+
+	AfterEach(func() {
+		By("Delete DV")
+		err := utils.DeleteDataVolume(f.CdiClient, f.Namespace.Name, dataVolume.Name)
+		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() bool {
+			_, err := f.K8sClient.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
+			return k8serrors.IsNotFound(err)
+		}, timeout, pollingInterval).Should(BeTrue())
```
akalenyu, line 2161: any reason for this? after each test we delete the per-test unique namespace

tests/import_test.go
```diff
@@ -2134,6 +2134,67 @@ var _ = Describe("Import populator", func() {
...
+	})
+
+	DescribeTable("Import should add KUBEVIRT_IO_ env vars to PVC labels when source is registry", func(pullMethod cdiv1.RegistryPullMethod, urlFn func() string) {
```
akalenyu, line 2164: I think we can even integrate the label check into existing tests, but then it would bug me that something this big
doesn't have it's own test.. WDYT

pkg/controller/populators/import-populator.go
```diff
@@ -156,17 +156,21 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 			break
 		}
 
-		// Once the import is succeeded, we rebind the PV from PVC' to target PVC
-		if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvc); err != nil {
-			return reconcile.Result{}, err
+		if cc.IsPVCComplete(pvcPrime) {
```
alromeros, line 159: I'd add here `&& cc.IsUnbound(pvc)` to avoid running Rebind when the PVC is bound.

pkg/controller/populators/upload-populator.go
```diff
@@ -159,13 +159,15 @@ func (r *UploadPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		// We'll get called later once it succeeds
 		r.recorder.Eventf(pvc, corev1.EventTypeWarning, errUploadFailed, fmt.Sprintf(messageUploadFailed, pvc.Name))
 	case string(corev1.PodSucceeded):
-		// Once the upload is succeeded, we rebind the PV from PVC' to target PVC
-		if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvcCopy); err != nil {
-			return reconcile.Result{}, err
+		if cc.IsPVCComplete(pvcPrime) {
```
alromeros, line 162: Same here, I think we should check here `&& cc.IsUnbound(pvc)`

pkg/importer/util_test.go
```diff
@@ -136,6 +136,37 @@ var _ = Describe("Clean dir", func() {
...
+var _ = Describe("Env vars to labels", func() {
+	It("Should convert KUBEVIRT_IO_ env vars to labels with values", func() {
+		envs := []string{
+			"INSTANCETYPE_KUBEVIRT_IO_DEFAULT_INSTANCETYPE=u1.small",
+			"INSTANCETYPE_KUBEVIRT_IO_DEFAULT_PREFERENCE=fedora",
+		}
+		labels := envsToLabels(envs)
+		Expect(labels).To(HaveLen(2))
+		Expect(labels).To(HaveKeyWithValue("instancetype.kubevirt.io/default-instancetype", "u1.small"))
+		Expect(labels).To(HaveKeyWithValue("instancetype.kubevirt.io/default-preference", "fedora"))
```
codingben, line 148: Can you please create an issue to update documentation, that tells us the difference between the usage of environment variables and labels? like we discussed offline in the meeting earlier today.

pkg/controller/populators/populator-base.go
```diff
@@ -268,12 +288,17 @@ func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorCon
...
 	// Each populator reconciles the target PVC in a different way
-	if cc.IsUnbound(pvc) || !cc.IsPVCComplete(pvc) || cc.IsMultiStageImportInProgress(pvc) {
-		return populator.reconcileTargetPVC(pvc, pvcPrime)
+	res, err := populator.reconcileTargetPVC(pvc, pvcPrime)
+	if err != nil {
+		return res, err
+	}
+
+	// Making sure to clean PVC once it is complete
+	if cc.IsPVCComplete(pvc) {
```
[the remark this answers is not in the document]
0xFelix: Looks like it does indeed! Pushed a new rev of the PR.

pkg/controller/populators/populator-base.go
```diff
@@ -268,12 +288,17 @@ func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorCon
 	}
 
 	// Each populator reconciles the target PVC in a different way
-	if cc.IsUnbound(pvc) || !cc.IsPVCComplete(pvc) || cc.IsMultiStageImportInProgress(pvc) {
-		return populator.reconcileTargetPVC(pvc, pvcPrime)
+	res, err := populator.reconcileTargetPVC(pvc, pvcPrime)
```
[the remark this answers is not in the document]
0xFelix: I think it has to run on every reconcile cycle, because otherwise the import-populator will not be able to retry setting all annotations and labels on the target PVC if import-controller modified the target PVC at the same time and the API refuses to accept the older object on update.

pkg/controller/populators/import-populator.go
0xFelix: ACK

pkg/controller/populators/upload-populator.go
0xFelix: ACK

pkg/controller/populators/import-populator.go
0xFelix: Looks to be fixed by https://github.com/kubevirt/containerized-data-importer/pull/3103/commits/c24717c44ac17f436eb9dbd5886afaf3106563ea

pkg/controller/import-controller.go
0xFelix: Let's see what CI says on the latest rev

pkg/controller/populators/populator-base.go
0xFelix: In an earlier rev of the PR they looked the same, but @ksimon1 suggested making it simpler. Leaving it up to you to decide. https://github.com/kubevirt/containerized-data-importer/pull/3103#discussion_r1524503076

pkg/importer/http-datasource.go
0xFelix: ACK, added a unit test for that.

tests/import_test.go
0xFelix: ACK, dropped.

tests/import_test.go
0xFelix: I'd keep it separate to be honest. As you said, it's a feature and IMO it warrants its own test.

pkg/importer/util_test.go
0xFelix: Will do once this is merged.

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
akalenyu: I agree with @alromeros, I would feel much better if we only go in here on specific occasions such as the labels not being there

We have a bigger problem here that can be taken outside of the PR

pkg/importer/http-datasource.go
akalenyu: > How would it trigger the restart, exit with a code != 0?

That was my thinking, yes (you would just return an error here and that would exit with 1, IIRC)
But thinking more it's pretty complex to start differentiating the error types here so maybe leave as is

pkg/controller/import-controller.go
akalenyu: Reading through the code I am expecting the same issue

pkg/controller/populators/populator-base.go
akalenyu: it's fine can keep

pkg/controller/util.go
```diff
@@ -332,6 +332,15 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *v1.Pod, termMs
...
 
+func setLabelsFromTerminationMessage(labels map[string]string, termMsg *common.TerminationMessage) {
+	if labels == nil || termMsg == nil {
+		return
+	}
+	for k, v := range termMsg.Labels {
+		labels[k] = v
+	}
+}
+
```
mhenriks, line 343: I understand returning if termMsg is nil but why labels?  I think it is a little awkward to always expect that to be allocated.  Maybe do something similar to append builtin?

```golang
func setLabelsFromTerminationMessage(labels map[string]string, termMsg *common.TerminationMessage) map[string]string {
	if ermMsg == nil {
		return labels
	}
        if labels == nil && len(termMsg.Labels) > 0 {
           labels = make(map[string]string)
        }
	for k, v := range termMsg.Labels {
		labels[k] = v
	}
        return labels
}
```

tools/cdi-containerimage-server/main.go
```diff
@@ -46,8 +49,21 @@ func main() {
...
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/info", func(w http.ResponseWriter, _ *http.Request) {
+		info := common.ServerInfo{
+			Env: os.Environ(),
+		}
+		w.Header().Set("Content-Type", "application/json")
+		err := json.NewEncoder(w).Encode(info)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
mhenriks, line 61: Not sure you can return the status code after you started writing to the response

pkg/controller/util.go
0xFelix: I kept it in line with `setAnnotationsFromPodWithPrefix`, which also does not `make` a new map, if the passed in `anno` variable is `nil`. However, it does not even check if `anno` is `nil`. In L413 above the call to `setLabelsFromTerminationMessage` labels are populated, if they are `nil`. IMO it's good practice to still avoid writing to a `nil` map in the func. If you want however, I can add `make` to the func. WDYT?

https://github.com/kubevirt/containerized-data-importer/blob/775c43deead68d7b7106480561ad8fee9b0d8774/pkg/controller/import-controller.go#L413-L421

tools/cdi-containerimage-server/main.go
0xFelix: It looks like a commonly used pattern to me. Also, I took a peek into the code of `Encode` and it uses an internal buffer and only writes to an `io.Writer` when marshalling succeeded and no more errors can occur.

See https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/stream.go;drc=dac9b9ddbd5160c5f4552410f5f8281bd5eed38c;l=230

0xFelix: Not sure what happens in case of failures when writing to the response though? But I'm not sure if in that case there is anything left to do with the response.

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
0xFelix: @mhenriks Would you be okay with leaving it as it is right now? The tests do not flake anymore and to me it looks correct. IMO adding a short-circuit workaround will be unnecessarily complex.

pkg/importer/http-datasource.go
0xFelix: Leaving as is for now.

pkg/controller/populators/populator-base.go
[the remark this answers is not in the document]
akalenyu: @alromeros @mhenriks if you're both okay with this, I would be too. Just making sure we don't get hotloops over this, similar to https://github.com/kubevirt/containerized-data-importer/pull/2985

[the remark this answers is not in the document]
alromeros: I have similar concerns, but other than that everything is looking good to me. Since all these populator changes are required due to `updatePVCWithPVCPrimeAnnotations` failing with the `failed to update object` error we can improve this a bit when changing the `update` to `patch` (as I plan to do in the near future).

akalenyu:
> /test pull-containerized-data-importer-e2e-hpp-previous

don't hate me but I think this [failure](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3103/pull-containerized-data-importer-e2e-hpp-previous/1770086551783477248) is related to the changes;
notice [here](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3103/pull-containerized-data-importer-e2e-hpp-previous/1770086551783477248/artifacts/1/1_dvs.log) how the "Completed" reason is not there

0xFelix:
Not hating you, I think you are right.

Until I had time to investigate:
/hold

pkg/controller/util.go
mhenriks: I just think it is weird to have it both ways, either assume labels is not nil and use it without checking, or allocate/return it.  But I think it is better to allocate/return because could be hiding bugs in the other two cases (caller doesn't know have to pass in an allocated map).  However using the map without checking is slightly better because will probably catch the bug in testing

tools/cdi-containerimage-server/main.go
mhenriks: More hair splitting but if Encode is buffering everything then it would be better to call Marshall ourselves making it more likely to send the correct response code

akalenyu:
@0xFelix Would you like to have this merged and other follow ups done async?

mhenriks:
> @0xFelix Would you like to have this merged and other follow ups done async?

Yeah let's get this in.  Thanks for the great work @0xFelix 
/approve

pkg/controller/util.go
0xFelix: Alright, I'll think about it and might open a follow up PR.

tools/cdi-containerimage-server/main.go
0xFelix: It is hair splitting, to be honest I'd like to keep it as is.

0xFelix:
/hold cancel

Let's get it in!

PR #3106, opened 2024-02-21
Watch DIC-orphan cronjobs and cleanup them

arnongilboa:

Jira-ticket: https://issues.redhat.com/browse/CNV-35787

**Release note**:
```release-note
BugFix: Watch DataImportCron-orphan cronjobs and cleanup them
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -1036,6 +1036,27 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	//FIXME: nicer cron name parsing
+	mapCronJobToCron := func(_ context.Context, obj client.Object) []reconcile.Request {
+		cronLabel := obj.GetLabels()[common.DataImportCronLabel]
+		cronNsName := strings.Split(cronLabel, ".")
+		if len(cronNsName) != 2 || len(cronNsName[0]) == 0 || len(cronNsName[1]) == 0 {
+			return nil
+		}
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: cronNsName[0], Name: cronNsName[1]}}}
+	}
+
+	if err := c.Watch(source.Kind(mgr.GetCache(), &batchv1.CronJob{}),
+		handler.EnqueueRequestsFromMapFunc(mapCronJobToCron),
+		predicate.Funcs{
+			CreateFunc: func(event.CreateEvent) bool { return true },
+			DeleteFunc: func(event.DeleteEvent) bool { return false },
+			UpdateFunc: func(event.UpdateEvent) bool { return true },
+		},
+	); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -41,7 +41,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -898,30 +897,19 @@ func (r *DataImportCronReconciler) cleanup(ctx context.Context, cron types.Names
 }
 
 func (r *DataImportCronReconciler) deleteJobs(ctx context.Context, cron types.NamespacedName) error {
-	deletePropagationBackground := metav1.DeletePropagationBackground
-	deleteOpts := &client.DeleteOptions{PropagationPolicy: &deletePropagationBackground}
-	selector, err := getSelector(map[string]string{common.DataImportCronLabel: getCronJobLabelValue(cron.Namespace, cron.Name)})
+	deleteOpts := client.DeleteOptions{PropagationPolicy: ptr.To[metav1.DeletionPropagation](metav1.DeletePropagationBackground)}
+	selector, err := getSelector(map[string]string{common.DataImportCronNsLabel: cron.Namespace, common.DataImportCronLabel: cron.Name})
 	if err != nil {
 		return err
 	}
-	cronJobList := &batchv1.CronJobList{}
-	if err := r.client.List(ctx, cronJobList, &client.ListOptions{Namespace: r.cdiNamespace, LabelSelector: selector}); err != nil {
+	opts := &client.DeleteAllOfOptions{ListOptions: client.ListOptions{Namespace: r.cdiNamespace, LabelSelector: selector}, DeleteOptions: deleteOpts}
+	if err := r.client.DeleteAllOf(ctx, &batchv1.CronJob{}, opts); err != nil {
 		return err
 	}
-	for _, cronJob := range cronJobList.Items {
-		if err := r.client.Delete(ctx, &cronJob, deleteOpts); cc.IgnoreNotFound(err) != nil {
-			return err
-		}
-	}
-	jobList := &batchv1.JobList{}
-	if err := r.client.List(ctx, jobList, &client.ListOptions{Namespace: r.cdiNamespace, LabelSelector: selector}); err != nil {
+	if err := r.client.DeleteAllOf(ctx, &batchv1.Job{}, opts); err != nil {
 		return err
 	}
-	for _, job := range jobList.Items {
-		if err := r.client.Delete(ctx, &job, deleteOpts); cc.IgnoreNotFound(err) != nil {
-			return err
-		}
-	}
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -1036,6 +1036,27 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	//FIXME: nicer cron name parsing
+	mapCronJobToCron := func(_ context.Context, obj client.Object) []reconcile.Request {
+		cronLabel := obj.GetLabels()[common.DataImportCronLabel]
+		cronNsName := strings.Split(cronLabel, ".")
+		if len(cronNsName) != 2 || len(cronNsName[0]) == 0 || len(cronNsName[1]) == 0 {
```
akalenyu, line 1043: Are `len(cronNsName[0]) == 0 || len(cronNsName[1]) == 0` actually possible?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1036,6 +1036,27 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
+			return nil
+		}
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: cronNsName[0], Name: cronNsName[1]}}}
+	}
+
+	if err := c.Watch(source.Kind(mgr.GetCache(), &batchv1.CronJob{}),
```
akalenyu, line 1049: Do we want to watch other resources too? `Job` maybe?

pkg/controller/dataimportcron-controller.go
```diff
@@ -900,7 +900,7 @@ func (r *DataImportCronReconciler) cleanup(ctx context.Context, cron types.Names
 func (r *DataImportCronReconciler) deleteJobs(ctx context.Context, cron types.NamespacedName) error {
 	deletePropagationBackground := metav1.DeletePropagationBackground
 	deleteOpts := &client.DeleteOptions{PropagationPolicy: &deletePropagationBackground}
-	selector, err := getSelector(map[string]string{common.DataImportCronLabel: getCronJobLabelValue(cron.Namespace, cron.Name)})
+	selector, err := getSelector(map[string]string{common.DataImportCronLabel: GetCronJobLabelValue(cron.Namespace, cron.Name)})
```
akalenyu, line 903: I think a little further down https://github.com/kubevirt/containerized-data-importer/blob/98d0caeb56b7fc23963eeca2e3db8329f8506d23/pkg/controller/dataimportcron-controller.go#L907-L910 can be replaced with a direct delete? I may be wrong but seems to be equivalent

tests/utils/dataimportcron.go
```diff
@@ -64,3 +75,18 @@ func NewDataImportCronWithStorageSpec(name, size, schedule, dataSource string, i
 		},
 	}
 }
+
+// VerifyCronJobCleanup verifies the DataImportCron cronjob was deleted
+func VerifyCronJobCleanup(k8sClient *kubernetes.Clientset, cronJobNamespace, cronNamespace, cronName string) {
```
akalenyu, line 80: I think we are better off not including this debugging utility in the code base; It's probably not that expensive, but, on the logical level we know the cleanup part of the reconcile loop has been there for ages and is covered by unit tests, and we are not aware of any problems with it.
We've identified a scenario where it could get skipped, and the watches should solve that.

pkg/controller/dataimportcron-controller.go
akalenyu: How come it's needed for jobs?

arnongilboa: There might be several jobs (according to the cronjob spec: 1 Active, 1 Successful, 1 Failed) from the cronjob, and also the initial job.

arnongilboa: On second thought, we have TTLSecondsAfterFinished set (10sec) for all jobs, so they will be deleted anyway and not sure we must delete them explicitly earlier.

pkg/controller/dataimportcron-controller.go
akalenyu: sounds good

pkg/controller/dataimportcron-controller.go
arnongilboa: @akalenyu regarding the cronjob deletion loop, since we may get there after the DIC is already deleted (like in the bug), we cannot delete the cronjob by name as it's based on the DIC UID
https://github.com/kubevirt/containerized-data-importer/blob/42ec627e3593c45027c2ebffb32f26a182c9cdee/pkg/controller/dataimportcron-controller.go#L1337-L1340
so unfortunately I think we should keep this loop.

akalenyu: Not sure I understand the problem. We can always convert the label value on the cron job to the deleted cron's name,
which is all we need to use DeleteAllOf with a label selector. No?

arnongilboa: Sure, we can use DeleteAllOf (fixing it), but note the convertion mentioned is currently buggy as the label may truncate the name
https://github.com/kubevirt/containerized-data-importer/blob/42ec627e3593c45027c2ebffb32f26a182c9cdee/pkg/controller/dataimportcron-controller.go#L1351-L1358

pkg/controller/dataimportcron-controller.go
arnongilboa: Not really, unless somebody annotated the cronjob with "." or something, but we can ignore such case.
Note the whole conversion is buggy as already commented.

pkg/controller/dataimportcron-controller.go
arnongilboa: No. As said, the List and Delete loop is replaced with DeleteAllOf. The bug in this originial PR is in the other direction: converting the watched cronjob label (which is possibly truncated) to reconcile ns/name (of its DIC). Fixing it as well.

pkg/controller/dataimportcron-controller.go
```diff
@@ -1235,7 +1238,8 @@ func (r *DataImportCronReconciler) setJobCommon(cron *cdiv1.DataImportCron, obj
 	}
 	util.SetRecommendedLabels(obj, r.installerLabels, common.CDIControllerName)
 	labels := obj.GetLabels()
-	labels[common.DataImportCronLabel] = getCronJobLabelValue(cron.Namespace, cron.Name)
+	labels[common.DataImportCronNsLabel] = cron.Namespace
```
akalenyu, line 1241: Are we worried about upgrade-carried resources not having this on them? maybe worth a small unit test that an existing cronjob without this label gets it on update?

arnongilboa: In upgrade it will set this label (assuming the DIC exists). Can sure add it to the unit tests.

pkg/controller/dataimportcron-controller.go
```diff
@@ -898,30 +897,19 @@ func (r *DataImportCronReconciler) cleanup(ctx context.Context, cron types.Names
 }
 
 func (r *DataImportCronReconciler) deleteJobs(ctx context.Context, cron types.NamespacedName) error {
-	deletePropagationBackground := metav1.DeletePropagationBackground
-	deleteOpts := &client.DeleteOptions{PropagationPolicy: &deletePropagationBackground}
-	selector, err := getSelector(map[string]string{common.DataImportCronLabel: getCronJobLabelValue(cron.Namespace, cron.Name)})
+	deleteOpts := client.DeleteOptions{PropagationPolicy: ptr.To[metav1.DeletionPropagation](metav1.DeletePropagationBackground)}
```
akalenyu, line 900: Brace yourself for a ptr.To conflict in release branches, but this is definitely better

pkg/controller/dataimportcron-controller.go
```diff
@@ -1036,6 +1024,21 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	mapCronJobToCron := func(_ context.Context, obj client.Object) []reconcile.Request {
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: obj.GetLabels()[common.DataImportCronNsLabel], Name: getCronName(obj)}}}
```
akalenyu, line 1028: hmm may want to back out if `obj.GetLabels()[common.DataImportCronNsLabel]` is nil? not sure that's possible but could get messy in the reconcile loop

pkg/controller/dataimportcron-controller.go
```diff
@@ -1036,6 +1024,21 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
+	}
+
+	if err := c.Watch(source.Kind(mgr.GetCache(), &batchv1.CronJob{}),
+		handler.EnqueueRequestsFromMapFunc(mapCronJobToCron),
+		predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool { return getCronName(e.Object) != "" },
```
akalenyu, line 1034: So only triggering on creation is enough?

arnongilboa: Yep, to fix this issue we only care specifically about the first time the cronjob is visibile to the new instance of DIC controller. This is "create" although cronjob might have been created long time before.

tests/utils/dataimportcron.go
arnongilboa: Removed the func test, added utest instead.

akalenyu (approved):
/approve

PR #3116, opened 2024-02-29
Avoid race condition during importer termination

alromeros:

This pull request sets the exit code when scratch space is required to 0 so, when deleting the importer pod after exiting for scratchSpaceNeeded, we don't need to deal with unwanted automatic restarts (RestartPolicyOnFailure). This conflicting behavior caused some issues during termination while the pod was already restarting after scratchSpaceNeeded.

```sh
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 42
        exitCode: 2
        exitCode: 42
        exitCode: 2
        exitCode: 42
        exitCode: 2
        exitCode: 0
        exitCode: 0
```
After the fix:
```sh
        exitCode: 0
        exitCode: 0
        exitCode: 0
        exitCode: 0
        exitCode: 0
        exitCode: 0
```

I think this is a better solution than https://github.com/kubevirt/containerized-data-importer/pull/3044 and https://github.com/kubevirt/containerized-data-importer/pull/3060. Let me know what you think.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -372,15 +372,14 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 
 	scratchExitCode := false
 	if pod.Status.ContainerStatuses != nil &&
-		pod.Status.ContainerStatuses[0].LastTerminationState.Terminated != nil &&
-		pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode > 0 {
-		log.Info("Pod termination code", "pod.Name", pod.Name, "ExitCode", pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode)
-		if pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode == common.ScratchSpaceNeededExitCode {
+		pod.Status.ContainerStatuses[0].State.Terminated != nil {
+		if message := pod.Status.ContainerStatuses[0].State.Terminated.Message; strings.Contains(message, common.ScratchSpaceRequired) {
 			log.V(1).Info("Pod requires scratch space, terminating pod, and restarting with scratch space", "pod.Name", pod.Name)
 			scratchExitCode = true
 			anno[cc.AnnRequiresScratch] = "true"
-		} else {
-			r.recorder.Event(pvc, corev1.EventTypeWarning, ErrImportFailedPVC, pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.Message)
+		} else if pod.Status.ContainerStatuses[0].State.Terminated.ExitCode > 0 {
+			log.Info("Pod termination code", "pod.Name", pod.Name, "ExitCode", pod.Status.ContainerStatuses[0].State.Terminated.ExitCode)
+			r.recorder.Event(pvc, corev1.EventTypeWarning, ErrImportFailedPVC, pod.Status.ContainerStatuses[0].State.Terminated.Message)
 		}
 	}
 
--- a/pkg/controller/import-controller_test.go
+++ b/pkg/controller/import-controller_test.go
@@ -531,7 +531,7 @@ var _ = Describe("Update PVC from POD", func() {
 		Expect(resPvc.GetAnnotations()[cc.AnnRunningConditionReason]).To(Equal("Explosion"))
 	})
 
-	It("Should NOT update phase on PVC, if pod exited with error state that is scratchspace exit", func() {
+	It("Should NOT update phase on PVC, if pod exited with scratchspace required msg", func() {
 		pvc := cc.CreatePvcInStorageClass("testPvc1", "default", &testStorageClass, map[string]string{cc.AnnEndpoint: testEndPoint, cc.AnnPodPhase: string(corev1.PodRunning)}, nil, corev1.ClaimBound)
 		scratchPvcName := &corev1.PersistentVolumeClaim{}
 		scratchPvcName.Name = "testPvc1-scratch"
@@ -540,17 +540,10 @@ var _ = Describe("Update PVC from POD", func() {
 			Phase: corev1.PodPending,
 			ContainerStatuses: []corev1.ContainerStatus{
 				{
-					LastTerminationState: corev1.ContainerState{
-						Terminated: &corev1.ContainerStateTerminated{
-							ExitCode: common.ScratchSpaceNeededExitCode,
-							Message:  "scratch space needed",
-						},
-					},
 					State: v1.ContainerState{
 						Terminated: &corev1.ContainerStateTerminated{
-							ExitCode: 1,
-							Message:  "I went poof",
-							Reason:   "Explosion",
+							ExitCode: 0,
+							Message:  common.ScratchSpaceRequired,
 						},
 					},
 				},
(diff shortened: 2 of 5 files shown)


alromeros:
Third PR trying to accomplish this but I think this is the best solution yet. https://github.com/kubevirt/containerized-data-importer/pull/3060 led to nasty behavior when pod was failing with errors other than scratch space required and https://github.com/kubevirt/containerized-data-importer/pull/3044 0 grace period wasn't behaving well leaving containers stuck on the ceph lane.

akalenyu (commented):
I think this is a really good idea!

You are saying the way we have it today (non-zero) is plain wrong,
as up until now we were restarting indefinitely over the scratch exit code?

akalenyu:
BTW - there is work to standardize the communication between importer & controllers at https://github.com/kubevirt/containerized-data-importer/pull/3103

alromeros:
> I think this is a really good idea!
> 
> You are saying the way we have it today (non-zero) is plain wrong, as up until now we were restarting indefinitely over the scratch exit code?

Yeah, when exiting with non-zero the pod keeps restarting indifinitely even when we just want to manually delete and recreate it. It isn't really an issue as it solves itself automatically but the deletion/creation is definitely racy.

pkg/controller/import-controller.go
```diff
@@ -372,15 +372,14 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 
 	scratchExitCode := false
 	if pod.Status.ContainerStatuses != nil &&
-		pod.Status.ContainerStatuses[0].LastTerminationState.Terminated != nil &&
-		pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode > 0 {
-		log.Info("Pod termination code", "pod.Name", pod.Name, "ExitCode", pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode)
-		if pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode == common.ScratchSpaceNeededExitCode {
+		pod.Status.ContainerStatuses[0].State.Terminated != nil {
```
akalenyu, line 375: @0xFelix I am pretty sure there is no collision between this and https://github.com/kubevirt/containerized-data-importer/pull/3103/commits/8462345840cdd85410a4907fa9bd924326b17cd5, just making sure

pkg/controller/import-controller.go
```diff
@@ -372,15 +372,14 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 
 	scratchExitCode := false
 	if pod.Status.ContainerStatuses != nil &&
-		pod.Status.ContainerStatuses[0].LastTerminationState.Terminated != nil &&
-		pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode > 0 {
-		log.Info("Pod termination code", "pod.Name", pod.Name, "ExitCode", pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode)
-		if pod.Status.ContainerStatuses[0].LastTerminationState.Terminated.ExitCode == common.ScratchSpaceNeededExitCode {
```
akalenyu, line 378: So `LastTerminationState` is not being set because requiring scratch now results in a "successful" pod?

pkg/controller/import-controller_test.go
```diff
@@ -540,17 +540,10 @@ var _ = Describe("Update PVC from POD", func() {
...
-					LastTerminationState: corev1.ContainerState{
-						Terminated: &corev1.ContainerStateTerminated{
-							ExitCode: common.ScratchSpaceNeededExitCode,
-							Message:  "scratch space needed",
-						},
-					},
 					State: v1.ContainerState{
 						Terminated: &corev1.ContainerStateTerminated{
-							ExitCode: 1,
-							Message:  "I went poof",
```
akalenyu, line 552: Why drop the message and reason? If I follow the code correctly they'll still get set

akalenyu (commented):
Just questions

/approve
/cc @mhenriks

pkg/controller/import-controller.go
alromeros: Right

alromeros:
/hold
Testing locally, works fine with rook-ceph-block but there is still an error code in the importer when using local storage class. Might be expected but I'll first make sure.

0xFelix (commented):
Actually, I think there is collision between this and https://github.com/kubevirt/containerized-data-importer/pull/3101. If the importer wants to signal that scratch space is required IMO it would be cleaner to do that with the proposed communication struct instead of matching the termination message string.

0xFelix:
During my work on #3101 and #3103 I also noticed that exit codes other than 0 do not work well in Kubernetes.

alromeros:
/unhold

@akalenyu Fixed the issue commented above: Returning 0 as normally clashed with cleanup functions that assume the imported file will be there during regular termination. Using `os.Exit(0)` instead avoids the issue.

akalenyu:
> /unhold
> 
> @akalenyu Fixed the issue commented above: Returning 0 as normally clashed with cleanup functions that assume the imported file will be there during regular termination. Using `os.Exit(0)` instead avoids the issue.

Ah, I see... how come this didn't fail any of the e2e tests?

alromeros:
> > /unhold
> > @akalenyu Fixed the issue commented above: Returning 0 as normally clashed with cleanup functions that assume the imported file will be there during regular termination. Using `os.Exit(0)` instead avoids the issue.
> 
> Ah, I see... how come this didn't fail any of the e2e tests?

Since the failing function is in `defer`, the scratch space required termination message was still written. After a couple of restarts, the pod with scratch space was created and the import succeeded anyway.

0xFelix:
Is there a chance for this to be part of #3101? IMO both this PR and #3101 try to solve very similar issues.

alromeros:
> Is there a chance for this to be part of #3101? IMO both this PR and #3101 try to solve very similar issues.

Hey @0xFelix, yeah it makes sense to eventually integrate this into #3101, probably using a specific field (`ScratchSpaceRequired` boolean maybe?) to avoid parsing the msg in the controller. That said, I would prefer to prioritize merging this first since this is part of a bugfix we plan to backport to v1.57, and it'd be safer to avoid backporting a larger feature. Thanks!

0xFelix:
Alright, makes sense if you want to backport this.

alromeros:
Added a new commit (https://github.com/kubevirt/containerized-data-importer/pull/3116/commits/3db5b724e4dd9612191a002a7d2f1930c50ca163) to address a flake caused by the new behavior:

The test relied on the assumption that deleting the img from the http server would always cause the DV to restart at least once. However, with the new faster recovery time in the importer, a race condition happened where the file was deleted with the download already started, which just caused the polling to keep retrying without failing. Since we recreate the file fast enough, there was no time for the DV to error,  so no restart was needed.

This test could also rely on false positives since the importer pod failing for scratch space always caused the DV to restart.

To fix this I deleted the part of the test that checks for dv restarts to be >= 1, and to make sure that everything is working as expected I added a md5sum check at the end.

akalenyu:
> Added a new commit ([3db5b72](https://github.com/kubevirt/containerized-data-importer/commit/3db5b724e4dd9612191a002a7d2f1930c50ca163)) to address a flake caused by the new behavior:
> 
> The test relied on the assumption that deleting the img from the http server would always cause the DV to restart at least once. However, with the new faster recovery time in the importer, a race condition happened where the file was deleted with the download already started, which just caused the polling to keep retrying without failing. Since we recreate the file fast enough, there was no time for the DV to error, so no restart was needed.
> 
> This test could also rely on false positives since the importer pod failing for scratch space always caused the DV to restart.
> 
> To fix this I deleted the part of the test that checks for dv restarts to be >= 1, and to make sure that everything is working as expected I added a md5sum check at the end.

Makes sense

akalenyu:
/test pull-containerized-data-importer-fossa
weird one, fossa yells about `github.com/gorhill/cronexpr` which has been around for ages

alromeros:
There's been some conversations around the possibility of not backporting this. I'll leave it like this until we make a decision.

akalenyu:
> There's been some conversations around the possibility of not backporting this. I'll leave it like this until we make a decision.

Yeah not a fan of backporting this, it's a subtle change, so unless someone is desperately asking for a backport I wouldn't do it

PR #3118, opened 2024-03-01
Improve readability of importer pod's creation code

alromeros:

Importer pod's creation code always seemed a bit redundant, with two functions (`makeNodeImporterPodSpec` and `makeImporterPodSpec`) sharing common code and logic.

This Pull Request aims to clean that code a bit, in an attempt to improve readability.

Summary of changes:

- `makeNodeImporterPodSpec` and `makeImporterPodSpec` have been merged into a single function.
- Most of the import-type specific logic has been moved to container, volume, and initContainer-specific functions.
- No need of `setImporterPodCommons` since now most of the code is shared between node and regular import.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -837,16 +837,15 @@ func createImporterPod(ctx context.Context, log logr.Logger, client client.Clien
 		return nil, err
 	}
 
-	var pod *corev1.Pod
-	if cc.GetSource(args.pvc) == cc.SourceRegistry && args.pvc.Annotations[cc.AnnRegistryImportMethod] == string(cdiv1.RegistryPullNode) {
+	if isRegistryNodeImport(args) {
 		args.importImage, err = getRegistryImportImage(args.pvc)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


alromeros:
I think this is already an improvement but I would like to clean `makeImporterContainerSpec` and `makeImporterVolumeSpec` a bit if possible. These functions are too large but I like to have this logic separate from the main `makeImporterPodSpec` function.

ShellyKa13 (commented):
I think it ok but omg this was one of the hardest reviews, I cant tell if I missed something, not sure what could have made this PR clearer, but its really hard.. need someone to take another look
/lgtm

alromeros:
> I think it ok but omg this was one of the hardest reviews, I cant tell if I missed something, not sure what could have made this PR clearer, but its really hard.. need someone to take another look /lgtm

Thanks for the review @ShellyKa13. I agree that it's hard to review but I'm confident the change is for the better. I've spent some time trying to clean up the container and volume creation functions even more but other than some cosmetic adjustments I wasn't able to do a lot.

PR #3123, opened 2024-03-04
Clean up leftover snapshot sources from DataImportCron tests

akalenyu:
The dataimportcron tests may affect existing crons in the environment (in addition to the ones deployed by testing)
so we are better off cleaning up after ourselves.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -1042,6 +1042,26 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	if err := mgr.GetClient().List(context.TODO(), &snapshotv1.VolumeSnapshotList{}); err != nil {
+		if meta.IsNoMatchError(err) {
+			// Back out if there's no point to attempt watch
+			return nil
+		}
+		if !cc.IsErrCacheNotStarted(err) {
+			return err
+		}
+	}
+	if err := c.Watch(source.Kind(mgr.GetCache(), &snapshotv1.VolumeSnapshot{}),
+		handler.EnqueueRequestsFromMapFunc(mapSourceObjectToCron),
+		predicate.Funcs{
+			CreateFunc: func(event.CreateEvent) bool { return false },
+			UpdateFunc: func(event.UpdateEvent) bool { return false },
+			DeleteFunc: func(e event.DeleteEvent) bool { return getCronName(e.Object) != "" },
+		},
+	); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/tests/dataimportcron_test.go
+++ b/tests/dataimportcron_test.go
@@ -15,7 +15,9 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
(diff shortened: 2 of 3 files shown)


akalenyu:
/hold
testing

tests/dataimportcron_test.go
```diff
@@ -72,6 +73,16 @@ var _ = Describe("DataImportCron", Serial, func() {
 
 		By("[AfterEach] Restore the profile")
 		Expect(utils.UpdateStorageProfile(f.CrClient, scName, *originalProfileSpec)).Should(Succeed())
+
+		if format := originalProfileSpec.DataImportCronSourceFormat; format != nil && *format == cdiv1.DataImportCronSourceFormatSnapshot {
+			return
+		}
```
arnongilboa, line 79: why not cleanup anyway?

akalenyu: true, done

tests/dataimportcron_test.go
```diff
@@ -72,6 +74,16 @@ var _ = Describe("DataImportCron", Serial, func() {
 
 		By("[AfterEach] Restore the profile")
 		Expect(utils.UpdateStorageProfile(f.CrClient, scName, *originalProfileSpec)).Should(Succeed())
+
+		// Clean up existing dataimportcrons in the environment that we might have switched
+		l, err := labels.Parse(common.DataImportCronLabel)
+		Expect(err).ToNot(HaveOccurred())
+		opts := &client.DeleteAllOfOptions{ListOptions: client.ListOptions{Namespace: metav1.NamespaceAll, LabelSelector: l}}
```
alromeros, line 81: I think I'm missing something here, since we use NamespaceAll and the generic `DataImportCronLabel`, couldn't this trigger the deletion of non-testing importcrons too?

akalenyu: This is exactly what we're trying to achieve - these tests will flip existing non-testing-importcrons to snapshots

pkg/controller/dataimportcron-controller.go
```diff
@@ -1042,6 +1042,26 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
...
+		if !cc.IsErrCacheNotStarted(err) {
+			return err
+		}
+	}
+	if err := c.Watch(source.Kind(mgr.GetCache(), &snapshotv1.VolumeSnapshot{}),
+		handler.EnqueueRequestsFromMapFunc(mapSourceObjectToCron),
+		predicate.Funcs{
+			CreateFunc: func(event.CreateEvent) bool { return false },
+			UpdateFunc: func(event.UpdateEvent) bool { return false },
+			DeleteFunc: func(e event.DeleteEvent) bool { return getCronName(e.Object) != "" },
```
mhenriks, line 1059: In what case is this helping?  Who/what is deleting the VolumeSnapshot?

akalenyu: This is what our tests do as well and "forget" to clean up after themselves

tests/dataimportcron_test.go
```diff
@@ -72,6 +74,23 @@ var _ = Describe("DataImportCron", Serial, func() {
...
+			Not(HaveOccurred()),
+			Satisfy(meta.IsNoMatchError),
+		))
+		for i := range snapshots.Items {
+			err = f.CrClient.Delete(context.TODO(), &snapshots.Items[i])
+			Expect(err).To(Or(
+				Not(HaveOccurred()),
+				Satisfy(meta.IsNoMatchError),
+			))
+		}
```
mhenriks, line 93: Question about the scope of this cleanup.  It is being applied to all tests.  Why not delete PVCs?  Is only one test or set of tests requiring snapshot cleanup?  If so it may be better to limit this cleanup to a sub context

akalenyu: Most tests in this suite have to go through this clean up, it's a pretty cheap op so don't think we care too much about deleting them in between

mhenriks: Why not explicitly delete the PVCs then?

akalenyu: We can, there's just no reason to since we allow switching from pvc->snap sources

akalenyu: Yeah `snap->pvc` transition not supported automatically

PR #3136, opened 2024-03-19
cdi.kubevirt.io/allowClaimAdoption annotation breaking "regular" operations

mhenriks:
Have to better track when a PVC was created for a particular DV (populate it) or created for someone else (adopt it)

Let's do this by adding a new annotation to a PVC created by the DataVolumeController

cdi.kubevirt.io/createdForDataVolume: "uid of datavolume"

Fixes #

https://issues.redhat.com/browse/CNV-39618

**Release note**:

```release-note
BugFix for CNV-39618 - cdi.kubevirt.io/allowClaimAdoption annotation broken
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -329,6 +329,9 @@ const (
 
 	// AnnCdiCustomizeComponentHash annotation is a hash of all customizations that live under spec.CustomizeComponents
 	AnnCdiCustomizeComponentHash = AnnAPIGroup + "/customizer-identifier"
+
+	// AnnCreatedForDataVolume stores the UID of the datavolume that the PVC was created for
+	AnnCreatedForDataVolume = AnnAPIGroup + "/createdForDataVolume"
 )
 
 // Size-detection pod error codes
@@ -2139,7 +2142,11 @@ func AllowClaimAdoption(c client.Client, pvc *corev1.PersistentVolumeClaim, dv *
 	if pvc == nil || dv == nil {
 		return false, nil
 	}
-	anno, ok := dv.Annotations[AnnAllowClaimAdoption]
+	anno, ok := pvc.Annotations[AnnCreatedForDataVolume]
+	if ok && anno == string(dv.UID) {
+		return false, nil
+	}
+	anno, ok = dv.Annotations[AnnAllowClaimAdoption]
 	// if annotation exists, go with that regardless of featuregate
 	if ok {
 		val, _ := strconv.ParseBool(anno)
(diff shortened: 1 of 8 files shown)


pkg/controller/common/util.go
```diff
@@ -2139,7 +2142,11 @@ func AllowClaimAdoption(c client.Client, pvc *corev1.PersistentVolumeClaim, dv *
 	if pvc == nil || dv == nil {
 		return false, nil
 	}
-	anno, ok := dv.Annotations[AnnAllowClaimAdoption]
+	anno, ok := pvc.Annotations[AnnCreatedForDataVolume]
+	if ok && anno == string(dv.UID) {
```
alromeros, line 2146: Maybe I misunderstood the annotation but shouldn't we allow adoption when the annotation matches the dv UID?

pkg/controller/common/util.go
```diff
@@ -329,6 +329,9 @@ const (
 
 	// AnnCdiCustomizeComponentHash annotation is a hash of all customizations that live under spec.CustomizeComponents
 	AnnCdiCustomizeComponentHash = AnnAPIGroup + "/customizer-identifier"
+
+	// AnnCreatedForDataVolume stores the UID of the datavolume that the PVC was created for
+	AnnCreatedForDataVolume = AnnAPIGroup + "/createdForDataVolume"
```
akalenyu, line 334: Would this work on DVs from previous versions? I recall this argument from https://github.com/kubevirt/containerized-data-importer/pull/3105

nirs:
@mhenriks can this change affect the cases when we do want to adopt the PVC? do we need testing with ramen or it is fully covered in the existing tests?

pkg/controller/datavolume/controller-base.go
```diff
@@ -1128,6 +1128,7 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume,
 		annotations[cc.AnnPriorityClassName] = dataVolume.Spec.PriorityClassName
 	}
 	annotations[cc.AnnPreallocationRequested] = strconv.FormatBool(cc.GetPreallocation(context.TODO(), r.client, dataVolume.Spec.Preallocation))
+	annotations[cc.AnnCreatedForDataVolume] = string(dataVolume.UID)
```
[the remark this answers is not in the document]
akalenyu: Ah, the DV would have a different UID on recovery?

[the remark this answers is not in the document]
mhenriks: Correct

pkg/controller/common/util.go
mhenriks: All pvcs without this annotation (created prev version) are adoptable

pkg/controller/common/util.go
mhenriks: No if UID matches we want to do regular population.  During DR restore the DV will have a different UID

mhenriks:
> @mhenriks can this change affect the cases when we do want to adopt the PVC? do we need testing with ramen or it is fully covered in the existing tests?

@nirs Should be no change required with ramen

pkg/controller/common/util.go
alromeros: Aw right, that makes total sense. Thanks.

nirs:
@mhenriks I see that 1.58.2 was released 2 days ago so it does not include this fix. Which release will include this fix?

mhenriks:
> @mhenriks I see that 1.58.2 was released 2 days ago so it does not include this fix. Which release will include this fix?

1.58.3 or 1.59.0 whichever comes first

nirs:
Thanks @mhenriks, @akalenyu!

PR #3147, opened 2024-03-22
cleanup: Return modified labels added from termination message

0xFelix:

Instead of changing the labels map that is passed in as parameter to the
setLabelsFromTerminationMessage function, it now returns a modified copy
of the passed in labels map and is renamed to
addLabelsFromTerminationMessage. If the passed in map is nil a new map
will be allocated and initialized.

Follow up to #3103 

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -417,7 +417,7 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 		pvc.GetLabels()[common.CDILabelKey] = common.CDILabelValue
 	}
 	if cc.IsPVCComplete(pvc) {
-		setLabelsFromTerminationMessage(pvc.GetLabels(), termMsg)
+		pvc.SetLabels(addLabelsFromTerminationMessage(pvc.GetLabels(), termMsg))
 	}
 
 	if !reflect.DeepEqual(currentPvcCopy, pvc) {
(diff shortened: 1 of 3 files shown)


pkg/controller/import-controller.go
```diff
@@ -417,7 +417,7 @@ func (r *ImportReconciler) updatePvcFromPod(pvc *corev1.PersistentVolumeClaim, p
 		pvc.GetLabels()[common.CDILabelKey] = common.CDILabelValue
 	}
 	if cc.IsPVCComplete(pvc) {
-		setLabelsFromTerminationMessage(pvc.GetLabels(), termMsg)
+		pvc.SetLabels(setLabelsFromTerminationMessage(pvc.GetLabels(), termMsg))
```
alromeros, line 420: Nit: `setLabelsFromTerminationMessage` sounds slightly innaccurate since it's no longer setting the labels, but returning them instead. This is personal preferrence though, just a thought.

alromeros (approved):
Looks good, thanks!

pkg/controller/import-controller.go
0xFelix: Do you think `addLabelsFromTerminationMessage` sound more accurate?

0xFelix: Renamed it and updated the PR.

alromeros: Nice, thank you!

mhenriks:
/approve
Thanks @0xFelix!

PR #3155, opened 2024-03-27
Fix volume/access mode inferring for temp host assisted source PVC in snapshot clones

akalenyu:
Sometimes with snapshot cloning we have to fall back to host assisted.
To do this, we create a temporary restore from the snapshot and initiate a host assisted clone
from that -> target PVC.
The issue this commit fixes is that we set the wrong access/volume modes on this temporary restore PVC
(we set it to the target's).

Fixes #

**Release note**:

```release-note
BugFix: Clone from snapshot - fix volume/access mode inferring for temp host assisted source PVC
```

### DIFF
--- a/pkg/controller/clone/planner.go
+++ b/pkg/controller/clone/planner.go
@@ -327,7 +327,7 @@ func (p *Planner) computeStrategyForSourcePVC(ctx context.Context, args *ChooseS
 		strategy = *cs
 	} else if args.TargetClaim.Spec.StorageClassName != nil {
 		sp := &cdiv1.StorageProfile{}
-		exists, err := getResource(ctx, p.Client, "", *args.TargetClaim.Spec.StorageClassName, sp)
+		exists, err := getResource(ctx, p.Client, metav1.NamespaceNone, *args.TargetClaim.Spec.StorageClassName, sp)
 		if err != nil {
 			return nil, err
 		}
@@ -546,7 +546,7 @@ func (p *Planner) planHostAssistedFromSnapshot(ctx context.Context, args *PlanAr
 		return nil, fmt.Errorf("source claim does not exist")
 	}
 
-	sourceClaimForDumbClone, err := createTempSourceClaim(ctx, args.DataSource.Namespace, args.TargetClaim, sourceSnapshot, p.Client)
+	sourceClaimForDumbClone, err := createTempSourceClaim(ctx, args.Log, args.DataSource.Namespace, args.TargetClaim, sourceSnapshot, p.Client)
 	if err != nil {
 		return nil, err
 	}
@@ -775,13 +775,25 @@ func createDesiredClaim(namespace string, targetClaim *corev1.PersistentVolumeCl
 	return desiredClaim
 }
 
-func createTempSourceClaim(ctx context.Context, namespace string, targetClaim *corev1.PersistentVolumeClaim, snapshot *snapshotv1.VolumeSnapshot, client client.Client) (*corev1.PersistentVolumeClaim, error) {
-	scName, err := getStorageClassNameForTempSourceClaim(ctx, snapshot, client)
+func createTempSourceClaim(ctx context.Context, log logr.Logger, namespace string, targetClaim *corev1.PersistentVolumeClaim, snapshot *snapshotv1.VolumeSnapshot, client client.Client) (*corev1.PersistentVolumeClaim, error) {
+	if snapshot.Status == nil || snapshot.Status.BoundVolumeSnapshotContentName == nil {
+		return nil, fmt.Errorf("volumeSnapshotContent name not found")
+	}
+	vsc := &snapshotv1.VolumeSnapshotContent{}
+	if err := client.Get(ctx, types.NamespacedName{Name: *snapshot.Status.BoundVolumeSnapshotContentName}, vsc); err != nil {
+		return nil, err
+	}
+	scName, err := getStorageClassNameForTempSourceClaim(ctx, vsc, client)
+	if err != nil {
+		return nil, err
+	}
+	targetCpy := targetClaim.DeepCopy()
+	fallbackVolumeMode := targetCpy.Spec.VolumeMode
+	volumeMode, err := getVolumeModeForTempSourceClaim(log, snapshot, vsc, fallbackVolumeMode)
 	if err != nil {
 		return nil, err
 	}
 	// Get the appropriate size from the snapshot
-	targetCpy := targetClaim.DeepCopy()
 	if snapshot.Status == nil || snapshot.Status.RestoreSize == nil || snapshot.Status.RestoreSize.Sign() == -1 {
 		return nil, fmt.Errorf("snapshot has no RestoreSize")
 	}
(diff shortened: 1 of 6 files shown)


ShellyKa13 (commented):
Over all logs great some small suggestions

akalenyu:
/hold

approach is broken when doing filesystem->filesystem (just one example) when storage profile prefers block

pkg/controller/dataimportcron-controller.go
```diff
@@ -1354,3 +1369,42 @@ func GetInitialJobName(cron *cdiv1.DataImportCron) string {
...
+			VolumeMode:       ptr.To(cdiv1.PersistentVolumeFromStorageProfile),
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					// Doesn't matter
+					corev1.ResourceStorage: resource.MustParse("1Gi"),
+				},
+			},
+		},
+	}
+	if err := dvc.RenderPvc(ctx, client, inferredPvc); err != nil {
```
akalenyu, line 1392: @arnongilboa this will probably not work for backports, wdyt about exporting `pvcFromStorage`?

akalenyu: the alternative would be to duplicate it, or,
to not support the upgrade path (only set source vol mode on new snapshots)

pkg/controller/clone/planner.go
```diff
@@ -834,3 +847,28 @@ func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapsh
...
+	if !exists {
+		// err?
+		return &fallback, nil
+	}
+	for _, cps := range sp.Status.ClaimPropertySets {
+		if cps.VolumeMode != nil {
+			return cps.VolumeMode, nil
+		}
+	}
+
```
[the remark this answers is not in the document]
akalenyu: dropping the storage profile guessing bit. our snapshots will be annotated with the source vol mode

pkg/controller/clone/planner.go
```diff
@@ -834,3 +847,28 @@ func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapsh
...
+	sp := &cdiv1.StorageProfile{}
+	exists, err := getResource(ctx, client, "", scName, sp)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		// err?
+		return &fallback, nil
```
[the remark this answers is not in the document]
akalenyu: dropped this bit too

pkg/controller/clone/planner.go
```diff
@@ -791,40 +798,46 @@ func createTempSourceClaim(ctx context.Context, namespace string, targetClaim *c
 		restoreSize = &reqSize
 	}
 
+	volumeMode, err := getVolumeModeForTempSourceClaim(ctx, vsc, client, scName)
```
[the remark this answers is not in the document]
akalenyu: done

pkg/controller/dataimportcron-controller.go
arnongilboa: Exporting `pvcFromStorage` just for backports seems odd. Duplicating `RenderPvc` in the backport seems  cleaner imho.

akalenyu: How bad of a backport would it be? I already see some APIs changes that would need to be pulled in

arnongilboa: afaik the api changes are minimal. Will `pvcFromStorage` export eliminate them?

akalenyu: `cdiv1.PersistentVolumeFromStorageProfile` at least, yes

I think that for the backport duplicating some storage profile helpers actually is less aggressive than renderpvc/pvcFromStorage, wdyt?
Would hate touching these funcs on a backport

arnongilboa: smells like the right direction.

akalenyu: > Duplicating `RenderPvc` in the backport seems cleaner imho.

I guess this was also your suggestion, I just misunderstood thinking you wanted to reintroduce it in older releases fully

arnongilboa: I meant duplicating the helper funcs you need, not specifically this one.

akalenyu:
ready for review @ShellyKa13 @arnongilboa

pkg/controller/dataimportcron-controller.go
```diff
@@ -702,6 +714,9 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 		},
 	}
 	r.setDataImportCronResourceLabels(dataImportCron, desiredSnapshot)
+	if pvc.Spec.VolumeMode != nil {
+		cc.AddAnnotation(desiredSnapshot, cc.AnnSourceVolumeMode, string(*pvc.Spec.VolumeMode))
```
ShellyKa13, line 718: If Im not missing anything the annotation will only be added in case we dont find the snapshot in line 722 and then we will create it.. if so we should do this change inside the `if`

ShellyKa13 (commented):
left some comments and I have a question:
This will allow a sort of a work around for existing snapshots? cause other then DIC snapshots no one actually puts the annotation right?

pkg/controller/dataimportcron-controller.go
akalenyu: this path sets the ann. on newly created snapshots, existing ones will be handled at 360

ShellyKa13: yeah I meant that unless you get inside the if in line 722, this will not actually do anything, so it should be under it and not before it

pkg/controller/clone/planner_test.go
```diff
@@ -846,6 +846,52 @@ var _ = Describe("Planner test", func() {
...
+				Expect(plan[0].(*SnapshotClonePhase).DesiredClaim.Spec.AccessModes).To(ConsistOf(
+					[]corev1.PersistentVolumeAccessMode{
+						corev1.ReadWriteOnce,
+					},
+				))
+				Expect(plan[0].(*SnapshotClonePhase).DesiredClaim.Spec.DataSource).To(BeNil())
+				Expect(plan[0].(*SnapshotClonePhase).DesiredClaim.Spec.DataSourceRef).To(BeNil())
+			},
+				Entry("when volumesnapshotcontent has source volume mode", []runtime.Object{volumeSnapshotContentWithSourceVolumeMode(), createStorageClass()}, corev1.PersistentVolumeMode("dummy"), createSourceSnapshot(sourceName, "test-snapshot-content-name", "vsc")),
+				Entry("when volumesnapshotcontent has no source volume mode but annotated with AnnSourceVolumeMode", []runtime.Object{createDefaultVolumeSnapshotContent(), createStorageClass()}, corev1.PersistentVolumeMode("dummyfromann"), snapWithSourceVolumeModeAnnotation()),
```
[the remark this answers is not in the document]
akalenyu: do we? it's kind of a chaos situation that we don't want anyone to depend on

pkg/controller/dataimportcron-controller.go
akalenyu: added

pkg/controller/clone/planner.go
```diff
@@ -834,3 +848,18 @@ func getStorageClassNameForTempSourceClaim(ctx context.Context, snapshot *snapsh
...
+		// Since 1.29 we should always return here
+		// Older versions did not populate this field and thus need more care
+		return vsc.Spec.SourceVolumeMode, nil
+	}
+
+	if v, ok := snapshot.Annotations[cc.AnnSourceVolumeMode]; ok {
+		mode := corev1.PersistentVolumeMode(v)
+		return &mode, nil
+	}
+
```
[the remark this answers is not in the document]
akalenyu: added

pkg/controller/clone/planner_test.go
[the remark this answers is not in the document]
ShellyKa13: This is a unit test which checks all the possible code scenarios, currently although not desired using the fallback volume is an option and so we should check we do reach that point if no other way..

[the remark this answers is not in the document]
akalenyu: added

ShellyKa13 (approved):
/lgtm

PR #3156, opened 2024-03-27
Double number of e2e parallel nodes from 3 to 6

akalenyu:
We have some feedback from testing CDI on a different platform that this can work tests-wise
(but of course, not resource-wise, which we will see here).

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
/cc @awels 
I think this is stable & gives us a back a good few minutes (10-15ish)

PR #3158, opened 2024-03-28
Fixes bazel build system for those using bazelisk

bc185174:

- Adds `.bazelversion` to ensure we have a good version of bazel chosen when building the repo. Users with bazel > 5.4.1 will hit build errors or those on 7.x.x will attempt to enable bazel modules.
- Adds standard `.user.bazelrc` and `.ci.bazelrc` options for users to configured.
- Adds bazel dirs to `.gitignore` so they are not checked into git.

Fixes #

N/A

Bazel is built with reproducibility in mind. Library, test, and binary builds are incremental, so there is not really any need to rely on external build containers since the binaries should always produce the same result.

I know its not always possible but should try to avoid `cgo`. These types of builds are complicated and often you'll hit build issues and break cross compiling. Not to mention its slower. Couple good article reads below on these topics but generally if you can avoid it, don't use it or at a last resort isolate the `cgo` dependencies.

https://dave.cheney.net/2016/01/18/cgo-is-not-go
https://relistan.com/cgo-when-and-when-not-to-use-it

**Release note**:

```release-note
NONE
```


mhenriks:
@bc185174 can you explain how your comments on reproducibility and cgo are relevant here?

bc185174:
> @bc185174 can you explain how your comments on reproducibility and cgo are relevant here?

Yeah sure. They're not really relevant to this PRs fix but just a general observation when I was trying to run tests/build binaries locally. I didn't realise you had build containers to test in! I think the build container is great for workflows, but devs shouldn't need this to run and build things locally.

- Ensure users are using a good version of bazel for the repo (bazelisk is amazing for this).
- `bazel run //:gazelle` shouldn't produce a diff.
- `bazel build //...` should not produce build errors.

Must admit I'm not up to scratch with why you use cgo as there are good reasons to! But there are pretty decent go libraries out there now trying to replace the need for calling C libraries and removing that dependency. See you use cgo for libguestfs.org/libnbd - https://github.com/pojntfx/go-nbd for example.

_Side-note_: k8s sig recommends against their fake client now with preference of using envtest as you are using a real client and a real API server.

akalenyu:
> _Side-note_: k8s sig recommends against their fake client now with preference of using envtest as you are using a real client and a real API server.

Would you mind linking the source? sounds really interesting

bc185174:
> > _Side-note_: k8s sig recommends against their fake client now with preference of using envtest as you are using a real client and a real API server.
> 
> Would you mind linking the source? sounds really interesting

Sure https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake#pkg-overview. Kube-builder have quite a useful reference as well https://book.kubebuilder.io/reference/envtest.html.

awels:
/remove-lifecycle stale
Just checking in with @akalenyu and @mhenriks to see what is missing here to get this over the finish line. Its been in here for a while.

PR #3163, opened 2024-04-04
Stop handling error phase as "unknown" in clone-populator

alromeros:

Until now, the clone populator handled the error phase as "unknown". This triggered a minor log flood with the `Unknown populator phase` message, which wasn't descriptive enough.

To avoid this, we are now handling the phase as the others but still avoiding changing DV phase for consistency with the non-populator flow. If necessary, now a proper event will be recorded in the DataVolume. This addresses this card: https://issues.redhat.com/browse/CNV-35802.

Fixes # https://issues.redhat.com/browse/CNV-35802 (partially)

Still in progress, let's see how tests behave.

**Release note**:

```release-note
Stop handling error phase as "unknown" in clone-populator
```

### DIFF
--- a/pkg/controller/datavolume/clone-controller-base.go
+++ b/pkg/controller/datavolume/clone-controller-base.go
@@ -426,7 +426,7 @@ var populatorPhaseMap = map[string]cdiv1.DataVolumePhase{
 	clone.RebindPhaseName:        cdiv1.RebindInProgress,
 	clone.SnapshotClonePhaseName: cdiv1.CloneFromSnapshotSourceInProgress,
 	clone.SnapshotPhaseName:      cdiv1.SnapshotForSmartCloneInProgress,
-	//clone.ErrorPhaseName:         cdiv1.Error, // Want to hold off on this for now
+	clone.ErrorPhaseName:         cdiv1.Failed,
 }
 
 func (r *CloneReconcilerBase) updateStatusPhaseForPopulator(pvc *corev1.PersistentVolumeClaim, dataVolumeCopy *cdiv1.DataVolume, event *Event) error {
@@ -437,7 +437,10 @@ func (r *CloneReconcilerBase) updateStatusPhaseForPopulator(pvc *corev1.Persiste
 		//dataVolumeCopy.Status.Phase = cdiv1.Unknown // hold off on this for now
 		return nil
 	}
-	dataVolumeCopy.Status.Phase = dvPhase
+	// Avoid setting DV to failed for consistency with non-populator flow
+	if dvPhase != cdiv1.Failed {
+		dataVolumeCopy.Status.Phase = dvPhase
+	}
 	r.setEventForPhase(dataVolumeCopy, dvPhase, event)
 	return nil
 }


pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -278,9 +278,9 @@ func (r *PvcCloneReconciler) syncClone(log logr.Logger, req reconcile.Request) (
 			return syncRes, nil
 		}
 
-		// Always call detect size, it will handle the case where size is specified
-		// and detection pod not necessary
-		if datavolume.Spec.Storage != nil {
+		// If the target's size is not specified, we can extract that value from the source PVC
+		targetRequest, hasTargetRequest := pvcSpec.Resources.Requests[corev1.ResourceStorage]
+		if !hasTargetRequest || targetRequest.IsZero() {
```
arnongilboa, line 283: Looks good, but maybe add a utest for it? also the func sure needs to be refactored... so many tabs I get a headache :)

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -437,7 +437,10 @@ func (r *CloneReconcilerBase) updateStatusPhaseForPopulator(pvc *corev1.Persiste
 		//dataVolumeCopy.Status.Phase = cdiv1.Unknown // hold off on this for now
 		return nil
 	}
-	dataVolumeCopy.Status.Phase = dvPhase
+	// Avoid setting DV to failed for consistency with non-populator flow
+	if dvPhase != cdiv1.Failed {
+		dataVolumeCopy.Status.Phase = dvPhase
+	}
```
arnongilboa, line 443: Not sure I understand why not use the `Failed` phase, as there are other phases (e.g. `RebindInProgress`) not used in the non-populator flow afaik.

alromeros: That's true, however, the populator-specific phases are transitory so they don't mess with tests/flow that much. Adding a failed phase at the end seems kinda harsh and (most importantly) breaks consistency with DataVolume behavior in non-populator flow.

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -426,7 +426,7 @@ var populatorPhaseMap = map[string]cdiv1.DataVolumePhase{
 	clone.RebindPhaseName:        cdiv1.RebindInProgress,
 	clone.SnapshotClonePhaseName: cdiv1.CloneFromSnapshotSourceInProgress,
 	clone.SnapshotPhaseName:      cdiv1.SnapshotForSmartCloneInProgress,
-	//clone.ErrorPhaseName:         cdiv1.Error, // Want to hold off on this for now
+	clone.ErrorPhaseName:         cdiv1.Failed,
```
ShellyKa13, line 429: Sorry but I really didnt understand this change and commit and how 'unknown' is related to this change..
Also if we do get to Error phase in the clone, with this change nothing will be shown in the dv?

alromeros: We use this map as a chart to get the appropriate DV phase depending on the AnnClonePhase annotation. We didn't have anything for error so we returned empty and the case was handled as an unknown populator phase: https://github.com/kubevirt/containerized-data-importer/blob/e7e0586e0cd3d37c90baafe8dcd48565371b3805/pkg/controller/datavolume/clone-controller-base.go#L436

This change does not affect the DV phase because we never set it to `Failed` when a clone fails, but it allows us to record an event to the DV that could help debug (check `updateStatusPhaseForPopulator`).

ShellyKa13: so Error phase is a terminal phase? cant come back from it? how come we have it in populator but not in dv?
Why was it a problem to have error phase in dv? I would have preferred to have some indication also in the dv status..

alromeros: We might be able to come back for an error phase, that's why we are cautious and avoid setting the DV to `Failed` unless we are sure it's terminal. We use annotations for populator phases and that's different than an object phase, but still may need some improvements. 

> Why was it a problem to have error phase in dv? 

We usually try to avoid it unless we are 100% sure it's terminal. However, this PR is just trying to improve that behavior without being too intrusive. With populators, we've been trying to be consistent with non-populator DV phases from the beginning.

ShellyKa13: Sorry I still dont understand why not do:
`clone.ErrorPhaseName:         cdiv1.Error`
and have this updated in the dv? how is this any different then any other populator specific phase?
and lets say we dont want to update in the dv why not do 
`if dvPhase != cdiv1.Error {`
How is Failed related to all this? as you said error!=failed

ShellyKa13 (commented):
/lgtm
Thanks for the discussion :)

PR #3165, opened 2024-04-05
Support IPv6 in controller GetMetricsURL

EduardGomezEscandell:
Joining hostname:port with string concatenation leads to wrong URLs when the host is an IPv6:
```
HOST    PORT   Sprintf    CORRECT
::1     8000   ::1:8000   [::1]:8000
```
To avoid this issue, it's best to use `net.JoinHostPort`.

It allows the metrics IP address to be IPv6.

This type of issue can be discovered running the following command:
```
    golangci-lint run ./... --enable nosprintfhostport
```

**Release note**:
```release-note
Support IPv6 for controller metrics URL.
```


pkg/controller/common/util_test.go
```diff
@@ -300,6 +300,52 @@ var _ = Describe("Rebind", func() {
...
+		pod := makePod("::1", true)
+		url, err := GetMetricsURL(pod)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(url).To(Equal("https://[::1]:8080/metrics"))
+	})
+
+	It("Should fail when there is no metrics port", func() {
+		pod := makePod("127.0.0.1", false)
+		_, err := GetMetricsURL(pod)
+		Expect(err).To(HaveOccurred())
```
akalenyu, line 345: Just a nit, if you want to, and since you know exactly which error to expect, you could assert
that you indeed get back this specific error
`Expect(err).To(MatchError(os.ErrNotExist))`

akalenyu (commented):
Looks great!
/approve

pkg/controller/common/util_test.go
EduardGomezEscandell: It is actually a regular string error: https://github.com/kubevirt/containerized-data-importer/blob/afb269737bfc90cad4f7eccfdf6d157a60b2c482/pkg/controller/common/util.go#L1632

akalenyu: Yeah, you'd have to take it out to a var to perform the assertion

EduardGomezEscandell: Not sure it's worth the hassle then

PR #3178, opened 2024-04-10
Enable nakedret linter

EduardGomezEscandell:

> Checks that functions with naked returns are not longer than a maximum
> size (can be zero).

https://github.com/alexkohler/nakedret

Relevant from the GO style guide:
> Naked returns are okay if the function is a handful of lines. Once
> it’s a medium sized function, be explicit with your return values.

https://go.dev/wiki/CodeReviewComments#named-result-parameters

I think a naked return is never easier to read than a regular return,
so I set the linter to always complain about it. Disagreements are
welcome.

**Release note**:

```release-note
Enable nakedret linter and populate all naked returns.
```

### DIFF
--- a/cmd/cdi-cloner/clone-source.go
+++ b/cmd/cdi-cloner/clone-source.go
@@ -33,15 +33,20 @@ type execReader struct {
 	stderr io.ReadCloser
 }
 
-func (er *execReader) Read(p []byte) (n int, err error) {
-	n, err = er.stdout.Read(p)
-	if errors.Is(err, io.EOF) {
-		if err2 := er.cmd.Wait(); err2 != nil {
-			errBytes, _ := io.ReadAll(er.stderr)
-			klog.Fatalf("Subprocess did not execute successfully, result is: %q\n%s", er.cmd.ProcessState.ExitCode(), string(errBytes))
-		}
+func (er *execReader) Read(p []byte) (int, error) {
+	n, err := er.stdout.Read(p)
+	if err == nil {
+		return n, nil
+	} else if !errors.Is(err, io.EOF) {
+		return n, err
 	}
-	return
+
+	if err := er.cmd.Wait(); err != nil {
+		errBytes, _ := io.ReadAll(er.stderr)
+		klog.Fatalf("Subprocess did not execute successfully, result is: %q\n%s", er.cmd.ProcessState.ExitCode(), string(errBytes))
+	}
+
+	return n, io.EOF
 }
 
 func (er *execReader) Close() error {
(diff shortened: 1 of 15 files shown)


cmd/openstack-populator/openstack-populator.go
```diff
@@ -186,14 +186,14 @@ func writeData(reader io.ReadCloser, file *os.File, imageID string, progress *pr
 
 func getAuthType() (authType clientconfig.AuthType, err error) {
 	if configuredAuthType := getStringFromSecret(authTypeString); configuredAuthType == "" {
-		authType = clientconfig.AuthPassword
+		return clientconfig.AuthPassword, nil
```
alromeros, line 189: When returning inside a condition we usually try avoiding unnecessary nesting to improve readability. In this case we can leave the else if since it's using `configuredAuthType`, but I'd drop the else:

```go
if configuredAuthType := getStringFromSecret(authTypeString); configuredAuthType == "" {
		return clientconfig.AuthPassword, nil
} else if supportedAuthType, found := supportedAuthTypes[configuredAuthType]; found {
		return supportedAuthType, nil
}
err = errors.New(unsupportedAuthTypeErrStr)
klog.Fatal(err.Error(), "authType", configuredAuthType)
return "", err
```

cmd/cdi-cloner/clone-source.go
```diff
@@ -209,7 +209,7 @@ func getInputStream(preallocation bool) (rc io.ReadCloser) {
 	default:
 		klog.Fatalf("Invalid content-type %q", contentType)
 	}
-	return
+	return rc
```
alromeros, line 212: Since we are being explicit with the return I think it'd be better to drop the named return parameter in most of these functions.

alromeros (commented):
I agree with the changes, just a couple of suggestions. Mainly, since we are being explicit with the return, there's no point in naming the return parameter so I think it's better to just drop it.

cmd/cdi-cloner/clone-source.go
EduardGomezEscandell: Ok sounds good. Many of these functions can use a refactor.

cmd/openstack-populator/openstack-populator.go
EduardGomezEscandell: Makes sense! I was hesitant to touch more than the return type, but I'll allow myself to refactor a bit around these.

pkg/controller/common/util.go
```diff
@@ -1494,16 +1494,14 @@ func NewImportDataVolume(name string) *cdiv1.DataVolume {
 func GetCloneSourceInfo(dv *cdiv1.DataVolume) (sourceType, sourceName, sourceNamespace string) {
```
alromeros, line 1494: This function still has unused named return parameters

pkg/controller/clone-controller.go
```diff
@@ -730,17 +730,15 @@ func ParseCloneRequestAnnotation(pvc *corev1.PersistentVolumeClaim) (exists bool
 	var ann string
 	ann, exists = pvc.Annotations[cc.AnnCloneRequest]
 	if !exists {
-		return
+		return false, "", ""
```
alromeros, line 733: Unused named return parameters

cmd/cdi-cloner/clone-source.go
```diff
@@ -193,23 +196,24 @@ func newTarReader(preallocation bool) (io.ReadCloser, error) {
...
 	case "blockdevice-clone":
-		rc, err = os.Open(mountPoint)
+		rc, err := os.Open(mountPoint)
 		if err != nil {
 			klog.Fatalf("Error opening block device %q: %+v", mountPoint, err)
 		}
+		return rc
 	default:
 		klog.Fatalf("Invalid content-type %q", contentType)
+		return nil
```
alromeros, line 215: Nit: Not a must but I'd leave this return outside the switch, imo it's better always to have an unnested final return.

alromeros (changes requested):
Looks good! Not a must but there are a couple of functions with unused return parameters, better to clean them up.

pkg/controller/common/util.go
EduardGomezEscandell: I kept these on purpose. The returned types are all strings, so I think the named return types serve as documentation.

pkg/controller/clone-controller.go
EduardGomezEscandell: Same thing, I kept them to avoid confusions about name/namespace ordering

EduardGomezEscandell:
retest

PR #3179, opened 2024-04-10
Enable autoformatting linters

EduardGomezEscandell:
This PR adds a few linters that autoformat the code. This should help make the codebase more consistent in style, as is the Go way. 
- `Gofmt` checks whether code was gofmt-ed. By default this tool runs with -s option to check for code simplification.
- `Whitespace` is a linter that checks for unnecessary newlines at the start and end of functions, if, for, etc.
- `Gci` controls Go package import order and makes it always deterministic. This one produces a big diff though.

**How to run them**
In CI runs, these will not autoformat and instead will just complain. Locally, you can run them with:
```
golangci-lint run ./... --fix
```

If you're using VSCode, you can add these options so the it runs when you save. I'm sure other editors have similar options.
```jsonc
"go.lintOnSave": "file", // Optional, makes it faster
"go.lintTool": "golangci-lint",
"go.lintFlags": [ "--fix" ],
```

The only manual changes are ones made to `.golangci.yaml`. Everythin else is auto-formatted.
I recommend reviewing this PR commit by commit as the result is very large.

**Release note**:

```release-note
Enable autoformatting linters and format the project
```


EduardGomezEscandell:
If the amount of changes is too scary, I can get rid of the GCI linter as that's where the majority of them come from.

alromeros:
Thanks for the PR! This improves consistency and looks good to me, let's see what others think.

mhenriks:
>Locally, you can run them with:
 golangci-lint run ./... --fix

Can we add a make target that runs this in the builder container?

EduardGomezEscandell:
> Can we add a make target that runs this in the builder container?

Already on it #3193 :smiley: 

The idea is to add this to `make format`.

PR #3184, opened 2024-04-11
Reduce amount of logs fetched by tests that assert on control plane pod logs

EduardGomezEscandell:
Several tests get information about the state of the cluster by parsing the logs.

Up until now, we had a loop fetching the entire log history and searching for a specific string on it. Local testing showed running the tests (focussed on a single entry in the table) would cause 8.1 MiB of log data to be fetched and parsed per test (with a subsequent enormous error message on failure. On certain test failure we could expect 364 MiB of logs per failed test (2.8 MiB/get * 0.5 gets/second * 270 second timeout).

I reduced this by exploiting two factors:
  1. Consectuive GETs need only request new logs (8.1Mib -> 2.8MiB)
  2. No need to get logs older than the test itself (2.7MiB -> 60 KiB)

I used this solution for all tests that showed up when grepping for `RunKubectlCommand("logs"`.

Which issue(s) this PR fixes:
Fixes https://issues.redhat.com/browse/CNV-39821

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/api_validation_test.go
+++ b/tests/api_validation_test.go
@@ -322,7 +322,7 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
 			By(fmt.Sprintf("Waiting for datavolume to match phase %s", string(cdiv1.ImportInProgress)))
 			err = utils.WaitForDataVolumePhase(f, f.Namespace.Name, cdiv1.ImportInProgress, dataVolumeName)
 			if err != nil {
-				f.PrintControllerLog()
+				fmt.Fprintf(GinkgoWriter, "Failed to wait for DataVolume phase: %v", err)
 				dv, dverr := f.CdiClient.CdiV1beta1().DataVolumes(f.Namespace.Name).Get(context.TODO(), dataVolumeName, metav1.GetOptions{})
 				Expect(dverr).ToNot(HaveOccurred(), "datavolume %s phase %s", dv.Name, dv.Status.Phase)
 			}
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -2597,7 +2597,7 @@ var _ = Describe("all clone tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 			annValue, preallocationAnnotationFound, err := utils.WaitForPVCAnnotation(f.K8sClient, targetDataVolume.Namespace, targetPvc, controller.AnnPreallocationRequested)
 			if err != nil {
-				f.PrintControllerLog()
+				fmt.Fprintf(GinkgoWriter, "Failed to wait for PVC annotation: %v", err)
 			}
 			Expect(err).ToNot(HaveOccurred())
 			Expect(preallocationAnnotationFound).To(BeTrue())
@@ -2607,7 +2607,7 @@ var _ = Describe("all clone tests", func() {
 
 			annValue, preallocationAnnotationFound, err = utils.WaitForPVCAnnotation(f.K8sClient, targetDataVolume.Namespace, targetPvc, controller.AnnPreallocationApplied)
 			if err != nil {
-				f.PrintControllerLog()
+				fmt.Fprintf(GinkgoWriter, "Failed to wait for PVC annotation: %v", err)
 			}
 			Expect(err).ToNot(HaveOccurred())
 			Expect(preallocationAnnotationFound).To(BeTrue())
@@ -3043,7 +3043,7 @@ func completeClone(f *framework.Framework, targetNs *v1.Namespace, targetPvc *v1
 		By("Verify the clone annotation is on the target PVC")
 		_, cloneAnnotationFound, err := utils.WaitForPVCAnnotation(f.K8sClient, targetNs.Name, targetPvc, controller.AnnCloneOf)
 		if err != nil {
-			f.PrintControllerLog()
+			fmt.Fprintf(GinkgoWriter, "Failed to wait for PVC annotation: %v", err)
 		}
 		Expect(err).ToNot(HaveOccurred())
 		Expect(cloneAnnotationFound).To(BeTrue())
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -3141,6 +3141,8 @@ func completeClone(f *framework.Framework, targetNs *v1.Namespace, targetPvc *v1
 }
 
 func cloneOfAnnoExistenceTest(f *framework.Framework, targetNamespaceName string) {
+	beforeClone := time.Now()
+
 	// Create targetPvc
 	By(fmt.Sprintf("Creating target pvc: %s/target-pvc", targetNamespaceName))
 	pvcName := "target-pvc"
@@ -3157,22 +3159,16 @@ func cloneOfAnnoExistenceTest(f *framework.Framework, targetNamespaceName string
 	f.ForceBindIfWaitForFirstConsumer(targetPvc)
 
 	By("Checking no cloning pods were created")
-
-	matchString := fmt.Sprintf("{\"PVC\": {\"name\":\"%s\",\"namespace\":\"%s\"}, \"isUpload\": false, \"isCloneTarget\": true, \"isBound\": true, \"podSucceededFromPVC\": true, \"deletionTimeStamp set?\": false}", pvcName, f.Namespace.Name)
-	Eventually(func() bool {
-		log, err := f.RunKubectlCommand("logs", f.ControllerPod.Name, "-n", f.CdiInstallNs)
-		Expect(err).NotTo(HaveOccurred())
-		return strings.Contains(log, matchString)
-	}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(BeTrue())
-	Expect(err).ToNot(HaveOccurred())
-
-	By("Checking logs explicitly skips PVC")
-	Eventually(func() bool {
-		log, err := f.RunKubectlCommand("logs", f.ControllerPod.Name, "-n", f.CdiInstallNs)
-		Expect(err).NotTo(HaveOccurred())
-		return strings.Contains(log, fmt.Sprintf("{\"PVC\": {\"name\":\"%s\",\"namespace\":\"%s\"}, \"checkPVC(AnnCloneRequest)\": true, \"NOT has annotation(AnnCloneOf)\": false, \"isBound\": true, \"has finalizer?\": false}", pvcName, targetNamespaceName))
-	}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(BeTrue())
-	Expect(err).ToNot(HaveOccurred())
+	Eventually(func() (string, error) {
+		out, err := f.K8sClient.CoreV1().
+			Pods(f.CdiInstallNs).
+			GetLogs(f.ControllerPod.Name, &corev1.PodLogOptions{SinceTime: &metav1.Time{Time: beforeClone}}).
+			DoRaw(context.Background())
+		return string(out), err
+	}, time.Minute, time.Second).Should(And(
+		ContainSubstring(fmt.Sprintf(`{"PVC": {"name":%q,"namespace":%q}, "isUpload": false, "isCloneTarget": true, "isBound": true, "podSucceededFromPVC": true, "deletionTimeStamp set?": false}`, pvcName, f.Namespace.Name)),
+		ContainSubstring(fmt.Sprintf(`{"PVC": {"name":%q,"namespace":%q}, "checkPVC(AnnCloneRequest)": true, "NOT has annotation(AnnCloneOf)": false, "isBound": true, "has finalizer?": false}`, pvcName, targetNamespaceName)),
+	))
 }
 
 func cleanDv(f *framework.Framework, dv *cdiv1.DataVolume) {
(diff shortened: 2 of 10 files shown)


alromeros:
I think we are seeing some actual flakes, holding just in case
/hold

alromeros:
Regarding the test failure, the [importer pod](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3184/pull-containerized-data-importer-e2e-istio/1783828829056274432/artifacts/1/pods/serial_1_cdi-e2e-tests-importer-gm65g_importer-prime-6d20e74b-1304-434b-a176-6e7ca728bbaa-importer.log) in the failed tests contains one of the expected logs (502 Bad Gateway) so I think there's some issue with the matching mechanism. I'm trying to identify it while reviewing.

EduardGomezEscandell:
Rebased main, no ne changes

tests/utils/logs.go
```diff
@@ -0,0 +1,157 @@
...
+	since := opts.Since
+	for {
+		select {
+		case <-ctx.Done():
+			return logNotFoundErr(opts, ctx.Err(), fullLogs)
+		case <-tk.C:
+		}
+
+		logOpt := &core.PodLogOptions{
+			SinceTime: &meta.Time{Time: since},
```
alromeros, line 116: I think the issue might be here? In the api (https://pkg.go.dev/k8s.io/api/core/v1#PodLogOptions) there's a field (`Previous`) that allows returning logs from previously failed containers. Since some of these tests read logs from pods that fail maybe we should set this to true. Not 100% sure though.

EduardGomezEscandell:
Seems like `pull-containerized-data-importer-e2e-destructive` is failing because no log is being fetched. The logs always return the error:
```
the server rejected our request for an unknown reason (get pods cdi-deployment-b5f4f866-rbslq)
```

EduardGomezEscandell:
The error in the unit tests is solved here: #3256

EduardGomezEscandell:
It's the same test failing everywhere:

https://github.com/kubevirt/containerized-data-importer/blob/a19e449cc111a17092ec00a06746051e798b38d2/tests/operator_test.go#L1175-L1184

It fails to fetch the logs for some reason, the error message is:
```
the server rejected our request for an unknown reason (get pods cdi-deployment-df4db985c-kbcfj)
```

Note that manually running the command works:
```
edu@fedora:~/containerized-data-importer$ k logs --namespace cdi cdi-deployment-df4db985c-kbcfj | grep "successfully acquired lease"
I0514 14:15:04.339068       1 leaderelection.go:260] successfully acquired lease cdi/cdi-controller-leader-election-helper
```

alromeros:
> It's the same test failing everywhere:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/a19e449cc111a17092ec00a06746051e798b38d2/tests/operator_test.go#L1175-L1184
> 
> It fails to fetch the logs for some reason, the error message is:
> 
> ```
> the server rejected our request for an unknown reason (get pods cdi-deployment-df4db985c-kbcfj)
> ```
> 
> Note that manually running the command works:
> 
> ```
> edu@fedora:~/containerized-data-importer$ k logs --namespace cdi cdi-deployment-df4db985c-kbcfj | grep "successfully acquired lease"
> I0514 14:15:04.339068       1 leaderelection.go:260] successfully acquired lease cdi/cdi-controller-leader-election-helper
> ```

hmmm maybe this is a transient error in the cluster that we can't really predict? For example, when we call `Update` in the tests we always do it inside a `Eventually` statement to bypass some unpredictable errors that don't depend on us (for example, [here](https://github.com/kubevirt/containerized-data-importer/blob/main/tests/import_test.go#L323)). Don't know if this case would be similar.

alromeros:
hmm I just checked how `GetLogs` is called in Kubevirt and it's never inside an `Eventually`... so I'm not really sure what could be happening here.

EduardGomezEscandell:
The problematic test was one where we expected the GET log to fail (with a 502 error). GetLogs indeed failed, but it does not report the problem, so the test failed.

The solution was to use the raw `kubectl` command for this case only. Everywhere else we test for succesfull calls to GetLogs so they can use the proper API.

alromeros:
> The problematic test was one where we expected the GET log to fail (with a 502 error). GetLogs indeed failed, but it does not report the problem, so the test failed.
> 
> The solution was to use the raw `kubectl` command for this case only. Everywhere else we test for succesfull calls to GetLogs so they can use the proper API.

That's odd, why would GetLog not return logs containing errors when that's precisely the point of getting logs? There may be some way to make this api work, but this already looks good. Thanks for the PR!

/lgtm

EduardGomezEscandell:
> That's odd, why would GetLog not return logs containing errors when that's precisely the point of getting logs? There may be some way to make this api work, but this already looks good. Thanks for the PR!

Well the GET request fails with 502 so there are no errors to show. I suspect we're asserting against the error message produced by `kubectl`.

tests/api_validation_test.go
```diff
@@ -308,6 +309,8 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
 		)
 
 		It("[test_id:4895][posneg:positive]report progress while importing 1024Mi PVC", func() {
+			testBegin := time.Now()
```
akalenyu, line 312: Should be able to use `CurrentSpecReport().StartTime`

tests/api_validation_test.go
```diff
@@ -322,7 +325,7 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
 			By(fmt.Sprintf("Waiting for datavolume to match phase %s", string(cdiv1.ImportInProgress)))
 			err = utils.WaitForDataVolumePhase(f, f.Namespace.Name, cdiv1.ImportInProgress, dataVolumeName)
 			if err != nil {
-				f.PrintControllerLog()
+				f.PrintControllerLog(testBegin)
```
akalenyu, line 328: I think `PrintControllerLog` isn't really needed anymore, from a quick check, it dates back 6 years
And we also have these type of logs under artifacts
https://gcsweb.ci.kubevirt.io/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3285/pull-containerized-data-importer-e2e-ceph/1793274221523111936/artifacts/1/pods/

akalenyu (commented):
Thank you for the PR, looks really good!

I have a few concerns:
- Test complexity/mem usage trade off
(I thought we'd be okay looping over GetLogs via a regular Eventually, as long as we set a start time, and use a variable scoped to the eventually loop)
- Asserting on logs in general usually indicates that we are doing something wrong (missing an event for something important) so I'd hoped we could get rid of some of them in favor of events for important things or a simple removal if they're not needed

EduardGomezEscandell:
> I thought we'd be okay looping over GetLogs via a regular Eventually, as long as we set a start time, and use a variable scoped to the eventually loop

I added a bit of boilerplate so that the full logs are printed on failure, which I think is worth the trade-off.

I didn't use a regular Eventually because I saw that the `test/utils` package never imports Gomega, so I wanted to follow that pattern. If that is not a real restriction, the code can be simplified quite a bit by using a regular gomega.Eventually.

EduardGomezEscandell:
Ok I just saw that you indeed import Gomega into test/utils sometimes, I should've checked twice :sweat: . I'll simplify this.

akalenyu:
> I added a bit of boilerplate so that the full logs are printed on failure, which I think is worth the trade-off.

The logs should be available for debugging in https://gcsweb.ci.kubevirt.io/gcs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3285/pull-containerized-data-importer-e2e-ceph/1793274221523111936/artifacts/1/pods/ unless ofc you delete the resources actively in the test itself
which we try not to do (each test cleans up after itself transparently via a namespace DELETE call)

akalenyu:
> If that is not a real restriction, the code can be simplified quite a bit by using a regular gomega.Eventually.

How bad would it be if we directly used Eventually in the tests? I mean I'm okay with wrapping it but have to question it a little since it's usually enough with all the expressive language around matchers

EduardGomezEscandell:
> How bad would it be if we directly used Eventually in the tests? I mean I'm okay with wrapping it but have to question it a little since it's usually enough with all the expressive language around matchers

I'm looking at it and yeah it's very doable. If I don't update the `since` variable, we fetch a bit more log history but in exchange we always get the full relevant log and so no need for a wrapper.

tests/api_validation_test.go
```diff
@@ -322,7 +323,7 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
 			By(fmt.Sprintf("Waiting for datavolume to match phase %s", string(cdiv1.ImportInProgress)))
 			err = utils.WaitForDataVolumePhase(f, f.Namespace.Name, cdiv1.ImportInProgress, dataVolumeName)
 			if err != nil {
-				f.PrintControllerLog()
+				klog.Warningf("failed to wait for DataVolume phase: %v", err)
```
akalenyu, line 326: Not the worst thing ever but pretty suprised to see klog in the tests package, usually have to use the ginkgo writer https://github.com/kubevirt/containerized-data-importer/blob/7c93c54bac55dc4becdad354ab9be7c1fa8bf34a/tests/api_validation_test.go#L269

tests/cloner_test.go
```diff
@@ -2280,6 +2285,7 @@ var _ = Describe("all clone tests", func() {
 		})
 
 		It("[test_id:3999] Create a data volume and then clone it and verify retry count", Serial, func() {
+
```
akalenyu, line 2288: past remnant I guess lol

tests/upload_test.go
```diff
@@ -321,24 +323,17 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+				return f.K8s().CoreV1().
+					Pods(uploadPod.Namespace).
+					GetLogs(uploadPod.Name, &core.PodLogOptions{SinceTime: &meta.Time{Time: CurrentSpecReport().StartTime}}).
+					DoRaw(context.Background())
+			}, controllerSkipPVCCompleteTimeout, assertionPollInterval).Should(Or(
+				ContainSubstring("is larger than the reported available"),
+				ContainSubstring("no space left on device"),
+				ContainSubstring("qemu-img execution failed"),
+				ContainSubstring("calculated new size is < than current size, not resizing"),
+			))
```
akalenyu, line 336: beautiful

tests/upload_test.go
```diff
@@ -1184,11 +1179,12 @@ var _ = Describe("CDIConfig manipulation upload tests", Serial, func() {
...
 
 		By("Verify Quota was exceeded in logs")
-		matchString := "pods \\\"cdi-upload-upload-test\\\" is forbidden: exceeded quota: test-quota, requested"
-		Eventually(func() string {
-			log, err := f.RunKubectlCommand("logs", f.ControllerPod.Name, "-n", f.CdiInstallNs)
-			Expect(err).NotTo(HaveOccurred())
-			return log
+		matchString := `pods \"cdi-upload-upload-test\" is forbidden: exceeded quota: test-quota, requested`
+		Eventually(func() ([]byte, error) {
+			return f.K8s().CoreV1().
```
akalenyu, line 1184: is `f.K8s()` a thing?

EduardGomezEscandell: Yep, `Frameworks` embeds `Clients` which has a `K8s` method:

https://github.com/kubevirt/containerized-data-importer/blob/4a64af4bdb4d1215c71008dc6b9aca3c7e2c94f9/tests/framework/framework.go#L103-L105

tests/cloner_test.go
EduardGomezEscandell: Yeah, I'm cleaning up quite a bit still.

tests/api_validation_test.go
EduardGomezEscandell: I'll replace it.

tests/upload_test.go
akalenyu: So I see we mostly use `f.K8sClient`, maybe want to align to that?

EduardGomezEscandell: Sounds good.

tests/import_test.go
```diff
@@ -342,16 +344,17 @@ var _ = Describe("[Istio] Namespace sidecar injection", Serial, func() {
...
-			if strings.Contains(log, "HTTP request errored") {
-				return true
-			}
-			if strings.Contains(log, "502 Bad Gateway") {
-				return true
-			}
-			return false
-		}, time.Minute, pollingInterval).Should(BeTrue())
+		Eventually(func() string {
+			// Cannot use GetLogs API because it does not return the error (i only says 'the server rejected our request for an unknown reason')
```
akalenyu, line 348: If you annotate the DV in this test with `AnnDeleteAfterCompletion` you should be able to get rid of the expectation for `502 Bad Gateway` (I assume it's for the case where importer gets deleted quickly?)

EduardGomezEscandell: I don't really know the logic behind this assertions, I simply refactored the test :sweat_smile:. I'll try that.

akalenyu: Actually, I got it wrong, this 502 code appears regularly in importer log text.
In that case, I am not sure why we can't use GetLogs?

EduardGomezEscandell: When I use GetLogs, I get a generic error message saying `the server rejected our request for an unknown reason (get pods cdi-deployment-b5f4f866-rbslq)`.

What I suspect is that this test Eventually is not asserting on the pod's log, but rather on the error message produced by the `kubectl` binary.

I tried to replicate it in my machine but I don't know how to run this test (seems like I need `KUBEVIRT_DEPLOY_ISTIO` but just setting the environment variable is not enough).

EduardGomezEscandell: Thanks! I'll work on fixing this then.

tests/operator_test.go
```diff
@@ -1172,13 +1167,19 @@ var _ = Describe("ALL Operator tests", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+
 				By("Ensuring this pod is the leader")
-				Eventually(func() bool {
-					controllerPod, err := utils.FindPodByPrefix(f.K8sClient, f.CdiInstallNs, cdiDeploymentPodPrefix, common.CDILabelSelector)
-					Expect(err).ToNot(HaveOccurred())
-					log := getLog(f, controllerPod.Name)
-					return checkLogForRegEx(logIsLeaderRegex, log)
-				}, 2*time.Minute, 1*time.Second).Should(BeTrue())
+				matchString := fmt.Sprintf("Successfully completed fsync(%s) syscall", common.WriteBlockPath)
```
akalenyu, line 1175: copy paste err? the fsync test is definitely not in operator_test.go

EduardGomezEscandell: That's right :sweat_smile:

alromeros:
Good job!
/lgtm

PR #3186, opened 2024-04-11
fix: propagating mesh annotations to transient pods

bc185174:

Mesh annotations are not replicating to pods set by the DV/PVCs. Allows users to unmesh these pods by default.

- Prep claim pod.
- Size detection pod.
- Importer pod.
- Clone source pod.
- Upload pod.

Fixes #

**Release note**:

```release-note
Mesh annotations now copied from the DV/PVCs to the following pods.

- Prep claim pod.
- Size detection pod.
- Importer pod.
- Clone source pod.
- Upload pod.
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -548,10 +548,8 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 			Name:      cloneSourcePodName,
 			Namespace: sourcePvcNamespace,
 			Annotations: map[string]string{
-				cc.AnnCreatedBy:                  "yes",
-				AnnOwnerRef:                      ownerRefAnno,
-				cc.AnnPodSidecarInjectionIstio:   sourcePvc.Annotations[cc.AnnPodSidecarInjectionIstio],
-				cc.AnnPodSidecarInjectionLinkerd: sourcePvc.Annotations[cc.AnnPodSidecarInjectionLinkerd],
+				cc.AnnCreatedBy: "yes",
+				AnnOwnerRef:     ownerRefAnno,
 			},
 			Labels: map[string]string{
 				common.CDILabelKey:       common.CDILabelValue, //filtered by the podInformer
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -548,8 +548,10 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 			Name:      cloneSourcePodName,
 			Namespace: sourcePvcNamespace,
 			Annotations: map[string]string{
-				cc.AnnCreatedBy: "yes",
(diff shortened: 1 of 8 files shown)


mhenriks:
Any disadvantage to adding these annotations all the time?  Could they ever be undesirable?  Could they ever somehow conflict?  If so me may want to add a param to CDI when we want annotations.  If the answer to both questions is no then I guess this approach fine

mhenriks:
> Not really no - by default they are disabled and only overwritten by the end users.

Sorry but I'm not quite parsing this

What do you mean by "disabled"?  Obviously if there is no service mesh the annotations will have no effect.  Is that what you mean?

Also confused about "overwritten".  When and why?

bc185174:
> > Not really no - by default they are disabled and only overwritten by the end users.
> 
> Sorry but I'm not quite parsing this
> 
> What do you mean by "disabled"? Obviously if there is no service mesh the annotations will have no effect. Is that what you mean?
> 
> Also confused about "overwritten". When and why?

Apologies - by default those annotations are set to `false` (istio) or `disabled` (linkerd) regardless of whether user has added them or not. They'll only change if a user applies the annotation, essentially patching and overwriting the default value.

pkg/controller/clone-controller.go
```diff
@@ -548,8 +548,10 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 			Name:      cloneSourcePodName,
 			Namespace: sourcePvcNamespace,
 			Annotations: map[string]string{
-				cc.AnnCreatedBy: "yes",
-				AnnOwnerRef:     ownerRefAnno,
+				cc.AnnCreatedBy:                  "yes",
+				AnnOwnerRef:                      ownerRefAnno,
+				cc.AnnPodSidecarInjectionIstio:   sourcePvc.Annotations[cc.AnnPodSidecarInjectionIstio],
```
akalenyu, line 553: You should be able to use `SetPvcAllowedAnnotations` for passing annotations from the data volume to the transfer pods

bc185174: Yeah - both prep-claim and the size detection pod were missing this.

akalenyu: Yeah looks much better now

pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -677,13 +677,14 @@ func (r *PvcCloneReconciler) makeSizeDetectionPodSpec(
 		pod.Annotations[cc.AnnOwnerUID] = string(dv.UID)
 	}
 
+	cc.SetPvcAllowedAnnotations(pod, sourcePvc)
```
akalenyu, line 680: I think you want targetPvc/targetDv instead of source here

akalenyu:
/hold
small issue
https://github.com/kubevirt/containerized-data-importer/pull/3186/files#r1565763489

PR #3191, opened 2024-04-11
#3183 follow-up: Move golang.org/x/net dependency from replace to require

alromeros:

https://github.com/kubevirt/containerized-data-importer/pull/3183 bumped the `golang.org/x/net` dependency to prevent CVE-2023-45288. We should have these dependencies under `require` instead of  `replace`, so this PR aims to fix that.

Fixes #

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -136,7 +136,7 @@ require (
 	golang.org/x/crypto v0.21.0 // indirect
 	golang.org/x/exp v0.0.0-20230713183714-613f0c0eb8a1 // indirect
 	golang.org/x/mod v0.12.0 // indirect
-	golang.org/x/net v0.21.0 // indirect
+	golang.org/x/net v0.23.0 // indirect
```
mhenriks, line 139: rather than change this generated indirect block I think it would be better to add our explicit require block here:  https://github.com/kubevirt/containerized-data-importer/blob/a2bc007da371c1c3fef8470f65e29cf53e1b8fda/go.mod#L5-L66

alromeros: I may be missing something but `make deps-update` keeps removing that change and adding it back to the indirect blok.

mhenriks: ah okay cool

PR #3193, opened 2024-04-12
Improve usability of Golangci-lint

EduardGomezEscandell:
Adds golangci-lint to the `make format` target, so it runs in a container.
It also removes redundant formatters that are also bundled inside golangci-lint (gofmt, govet, goimports).

**Release note**:

```release-note
NONE
```


hack/build/bazel-docker.sh
```diff
@@ -107,7 +107,7 @@ volumes="-v ${BUILDER_VOLUME}:/root:rw,z"
 mkdir -p "${HOME}/.docker"
 volumes="$volumes -v ${HOME}/.docker:/root/.docker:ro,z"
 
-if [[ $CDI_CRI = podman* ]] && [[ -f "${XDG_RUNTIME_DIR-}/containers/auth.json" ]]; then
+if [[ $CDI_CRI == podman* ]] && [[ -f "${XDG_RUNTIME_DIR-}/containers/auth.json" ]]; then
```
awels, line 110: So from what I understand, maybe do an `eq` instead would be better? I am afraid doing a single `=` is an assignment instead of a comparison.

EduardGomezEscandell: I did not touch this, it's the result of the sh formatter. I simply ran `make format` before making any changes.

Suposedly they're the same thing https://stackoverflow.com/questions/20449543/shell-equality-operators-eq

EduardGomezEscandell: I can still change it if you prefer so.

akalenyu:
/hold
just trying to orchestrate merge queue, will remove this very soon

akalenyu:
> /hold just trying to orchestrate merge queue, will remove this very soon

/hold cancel
Actually no good reason to hold this - very little changes are made to source code files

PR #3194, opened 2024-04-14
Add new DV reason: ImagePullFailed

nunnatsa:
## What this PR does / why we need it
When the image pull fails, the DataVolume Running condition, will have the Reason field of `ImagePullFailed`, to allow better error handling in code that uses it.

Also, moved the `StreamDataToFile` function from the `util` package to the `importer` package, and made to non-exported. This is done to allow calling the new `NewErrImagePullFailed` function from within `streamDataToFile`, and to avoid import cycle. In addition, the `streamDataToFile` is only used in the `importer` package and so it's better to define it there.

### Release note
```release-note
When the image pull fails, the DataVolume Running condition, will have the Reason field of "ImagePullFailed"
```

```jira-ticket
https://issues.redhat.com/browse/CNV-39603
```

### DIFF
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/utils/ptr"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
 	"kubevirt.io/containerized-data-importer/pkg/image"
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/utils/ptr"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	controller "kubevirt.io/containerized-data-importer/pkg/controller/common"
 	dvc "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
@@ -2196,6 +2197,41 @@ var _ = Describe("Containerdisk envs to PVC labels", func() {
 		Entry("with pullMethod pod", cdiv1.RegistryPullPod, tinyCoreRegistryURL, false),
 		Entry("with pullMethod node", cdiv1.RegistryPullNode, trustedRegistryURL, false),
 		Entry("with pullMethod node", cdiv1.RegistryPullNode, trustedRegistryIS, true),
+	)
+})
+
+var _ = Describe("pull image failure", func() {
+	var (
+		f = framework.NewFramework(namespacePrefix)
+	)
+
+	DescribeTable(`Should fail with "ImagePullFailed" reason if failed to pull image`, func(url string, pullMethod cdiv1.RegistryPullMethod) {
+		dv := utils.NewDataVolumeWithRegistryImport("failed-to-pull-image", "10Gi", "docker://"+url)
+		if dv.Annotations == nil {
+			dv.Annotations = make(map[string]string)
+		}
+		dv.Annotations[controller.AnnImmediateBinding] = "true"
+		dv.Spec.Source.Registry.PullMethod = &pullMethod
+
+		var err error
+		dv, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = utils.WaitForPVC(f.K8sClient, dv.Namespace, dv.Name)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Verify ImagePullFailed running condition")
+		runningCondition := &cdiv1.DataVolumeCondition{
+			Type:    cdiv1.DataVolumeRunning,
+			Status:  v1.ConditionFalse,
+			Message: common.ImagePullFailureText,
+			Reason:  "ImagePullFailed",
+		}
+		utils.WaitForConditions(f, dv.Name, f.Namespace.Name, controllerSkipPVCCompleteTimeout, assertionPollInterval, runningCondition)
+
+	},
+		Entry("pull method = pod", "myregistry/myorg/myimage:wrongtag", cdiv1.RegistryPullPod),
+		Entry("pull method = node", "myregistry/myorg/myimage:wrongtag", cdiv1.RegistryPullNode),
 	)
 })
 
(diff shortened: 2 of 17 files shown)


cmd/cdi-importer/importer.go
```diff
@@ -179,6 +180,13 @@ func handleImport(
 	err := processor.ProcessData()
 
 	scratchSpaceRequired := errors.Is(err, importer.ErrRequiresScratchSpace)
+
+	ipfErr := &importer.ErrImagePullFailed{}
+	isImagePullFailure := errors.As(err, &ipfErr)
+	if isImagePullFailure {
```
awels, line 186: Why not just do
```go
if errors.As(err, &ipfErr) {
  err = ipfErr
}
```
Also if the `As` is true, isn't err already the right type? I am not entirely I understand what is going on here.

tests/import_test.go
```diff
@@ -89,6 +90,34 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+			Type:    cdiv1.DataVolumeRunning,
+			Status:  v1.ConditionFalse,
+			Message: common.ImagePullFailureText,
+			Reason:  "ImagePullFailed",
+		}
+		utils.WaitForConditions(f, dv.Name, f.Namespace.Name, controllerSkipPVCCompleteTimeout, assertionPollInterval, runningCondition)
+
+	},
+		Entry("insecure registry", "registry:5000/myimage:latest"),
+		Entry("non exists image", "quay.io/capk/ubuntu-2004-container-disk:wrong"),
```
awels, line 118: Should not be a real issue, but maybe reference the internal registry here with an invalid image. This now assumes you have access to the internet, which may not always be valid.

awels (commented):
Looks like most of this PR is not actually related to returning the error message, but in moving the StreamDataToFile function. Other than that some minor comments.

cmd/cdi-importer/importer.go
nunnatsa: We don't need the boolean here. leftover from another implementation during the development. Will fix.

The `As` function does not change `err` but only populate `ipfErr`

tests/import_test.go
nunnatsa: Any other idea?

cmd/cdi-importer/importer.go
nunnatsa: In 2nd thought, I don't need this code at all. I was trying to write a new field in the `TerminationMessage` object, but this is not working anyway for other reasons. 

I'm now relaying on the error text, that is still there.

nunnatsa: removed this code

akalenyu (commented):
Thanks!
Had some minor comments but what's more interesting to me is why we can't use the termination message API in case of failures.
Is there something missing in it's implementation? I won't ofcourse block over this

nunnatsa:
> Thanks! Had some minor comments but what's more interesting to me is why we can't use the termination message API in case of failures. Is there something missing in it's implementation? I won't ofcourse block over this

the termination message API (object) is only available for pod that was terminated with no error. For error cases, the pod write unstructed text, and I didn't want to change this behavior because it's way out of the scope of this PR.

tests/import_test.go
```diff
@@ -89,6 +90,34 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
...
+		runningCondition := &cdiv1.DataVolumeCondition{
+			Type:    cdiv1.DataVolumeRunning,
+			Status:  v1.ConditionFalse,
+			Message: common.ImagePullFailureText,
+			Reason:  "ImagePullFailed",
+		}
+		utils.WaitForConditions(f, dv.Name, f.Namespace.Name, controllerSkipPVCCompleteTimeout, assertionPollInterval, runningCondition)
+
+	},
+		Entry("insecure registry", "registry:5000/myimage:latest"),
```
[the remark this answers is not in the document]
nunnatsa: I wanted to catch to cases here that should be easy to reproduce: insecure registry and non exist image tag. But I don't think the test will fail outside of kvci, because it will still won't be able to pull the image, for other reasons (e.g. registry not found)

nunnatsa:
> e, but in moving the StreamDataToFile function. Other than that some minor comments.

Eventually, yes. But this is not the "why". I moved the function because I wanted to call the `NewErrImagePullFailed` function from it, and this function is defined in the importer package. Then I found out that the `StreamDataToFile` function only called from the importer package. So to avoid import cycle, I moved the function, but this is not the reason for this change.

tests/import_test.go
```diff
@@ -89,6 +90,34 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		Entry("fail given a large physical size QCOW2 file", utils.LargePhysicalDiskQcow),
 	)
 
+	DescribeTable(`Should fail with "ImagePullFailed" reason if failed to pull image`, func(url string) {
```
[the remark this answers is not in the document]
nunnatsa: done

tests/import_test.go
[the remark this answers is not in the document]
akalenyu: yeah it fails on
```
transport.go:78] Could not create image reference: pinging container registry registry:5000: Get "https://registry:5000/v2/": dial tcp: lookup registry on 172.30.0.10:53: no such host
```
so should return your new err.

but how come this tests insecure registry? AFAIK the kubevirtci VM has this registry whitelisted

akalenyu (commented):
Looks good! thanks

Correct me if I'm wrong but this PR does not handle `pullMethod: node` (this is fine),
which I thought is quite common on OpenShift. Are you okay with that?

tests/import_test.go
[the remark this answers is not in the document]
nunnatsa: There is a field in CDI CR for that. I'm not sure what is it setting in kvci, but locally this is what I'm getting.

nunnatsa:
^^^ fixed new linter warning

tests/import_test.go
```diff
@@ -2199,6 +2200,41 @@ var _ = Describe("Containerdisk envs to PVC labels", func() {
...
+			Type:    cdiv1.DataVolumeRunning,
+			Status:  v1.ConditionFalse,
+			Message: common.ImagePullFailureText,
+			Reason:  "ImagePullFailed",
+		}
+		utils.WaitForConditions(f, dv.Name, f.Namespace.Name, controllerSkipPVCCompleteTimeout, assertionPollInterval, runningCondition)
+
+	},
+		Entry("pull method = pod", "myregistry/myorg/myimage:wrongtag", cdiv1.RegistryPullPod),
+		Entry("pull method = node", "myregistry/myorg/myimage:wrongtag", cdiv1.RegistryPullNode),
```
akalenyu, line 2234: :+1:

akalenyu (commented):
/approve

PR #2832, opened 2023-08-01
Always use scratchspace when importing and converting

awels:
**What this PR does / why we need it**:
nbdkit is slow in combination with qemu-img when importing and converting a qcow2 image. For now when converting always use scratch space, this is significantly faster. Once nbdkit 1.35.8 is available we can revisit this because that version will improve nbdkit performance significantly.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2809 

**Special notes for your reviewer**:
Also removes old version of nbdkit left over in WORKSPACE file, as well as fixing invalid number of arguments in some error returns

**Release note**:

```release-note
BugFix: use scratch space for all http conversion imports, to speed up the process.
```

Issue #2809, opened 2023-07-15
CDI import from URL is significantly slower than a manual wget + virtctl image-upload


cloudymax:
**What happened**:
There is a massive performance variance between CDI image import methods which makes the declarative workflow for PVCs/DataVolumes unusable. I'm REALLY trying to make Kubevirt work as a replacement for qemu wrapped in bash but the slowness of drive creation is making it impossible to achieve buy-in from others.

You can find my full setup here: https://github.com/cloudymax/argocd-nvidia-lab/tree/main/kubevirt

Disk configs are specifically here: https://github.com/small-hack/argocd/tree/main/kubevirt/disks

CDI configuration is here: https://github.com/small-hack/argocd/tree/main/kubevirt/cdi

This has been tested across multiple hardware types:
- Hetzner AX102 Instance with R97950X3D + NVME (Raid0)
- Dell XPS with i7-11700 + NVME (no raid)
- Hetzner instance with i7 7700 + SATA6 SSDs (Raid0)

- Has been tested on K3s (bare-metal and in a KVM VPS)

- Has been tested with local-path and Longhorn storage classes

- The CDI has been tested with nodePort and LoadBalancer type
- I have increased the resources available to the CDI importer but it had no effect, the process will not consume more resources.

Method: wget + virtctl
Storage Class: local-path
Time: 22 seconds

Method: wget + virtctl
Storage Class: longhorn
Time: 56 seconds

Command used:

```bash
export VOLUME_NAME=debian12-pvc
export NAMESPACE="default"
export STORAGE_CLASS="longhorn"
export ACCESS_MODE="ReadWriteMany"
export IMAGE_URL="https://cloud.debian.org/images/cloud/bookworm/daily/latest/debian-12-generic-amd64-daily.qcow2"
export IMAGE_PATH=debian-12-generic-amd64-daily.qcow2
export VOLUME_TYPE=pvc
export SIZE=120Gi
export PROXY_ADDRESS=$(kubectl get svc cdi-uploadproxy-loadbalancer -n cdi -o json | jq --raw-output '.spec.clusterIP')
# $(kubectl get svc cdi-uploadproxy -n cdi -o json | jq --raw-output 

time wget -O $IMAGE_PATH $IMAGE_URL && \
time virtctl image-upload $VOLUME_TYPE $VOLUME_NAME \
    --size=$SIZE \
    --image-path=$IMAGE_PATH \
[...]

PR #2351, opened 2022-07-06
Don't use nbdkit for xz or gz

mhenriks:
Signed-off-by: Michael Henriksen <mhenriks@redhat.com>

**What this PR does / why we need it**:

Nbdkit is not a good fit for gz because it downloads/unzips the file to tempdir.

It is also not a good fit for xz, as discussed in #2288.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2288

https://bugzilla.redhat.com/show_bug.cgi?id=2089391

**Special notes for your reviewer**:

**Release note**:

```release-note
NBDKit no longer used for .gz or .xz files
```

PR #3212, opened 2024-04-19
[WIP] Revert to using nbdkit for http conversion imports

alromeros:

Due to some performance issues discussed in https://github.com/kubevirt/containerized-data-importer/issues/2809, we stopped using nbdkit for most http imports. This new behavior introduced minor differences in some specific flows, such as stopping the [conversion](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/data-processor.go#L268) of uncompressed raw images.

The conversion process made the actual size of raw images significantly smaller, probably because of qemu's handling of sparse images. The import of raw images without conversion ended up causing failures in some tests, as imported images had a significant increase in size.

Since nbdkit performance issues have been addressed in v1.35.8 (and now we use v1.36.2), this pull request aims to revert to the old behavior.

**Example:** 

Fresh image import before this PR:

```sh
sh-5.1$ qemu-img info disk.img 
image: disk.img
file format: raw
virtual size: 70 GiB (75161927680 bytes)
disk size: 55 GiB
```

Fresh image import after this PR (due to convert):

```sh
$ qemu-img info disk.img 
image: disk.img
file format: raw
virtual size: 70 GiB (75161927680 bytes)
disk size: 9.95 GiB
```

Fixes # https://issues.redhat.com/browse/CNV-36026

Check https://github.com/kubevirt/containerized-data-importer/issues/2809 and https://github.com/kubevirt/containerized-data-importer/pull/2832 for more context about the original change and why it's safe to revert now.

If we prefer to keep this behavior and avoid using nbdkit, an alternative would be to use scratch for uncompressed raw images.

**Release note**:

```release-note
Bugfix: Use nbdkit for http conversion imports
```

### DIFF
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -133,20 +133,20 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	if hs.contentType == cdiv1.DataVolumeArchive {
 		return ProcessingPhaseTransferDataDir, nil
 	}
-	if !hs.readers.Convert {
-		return ProcessingPhaseTransferDataFile, nil
-	}
-	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod == string(cdiv1.RegistryPullNode) {
-		hs.url, _ = url.Parse(fmt.Sprintf("nbd+unix:///?socket=%s", nbdkitSocket))
-		if err = hs.n.StartNbdkit(hs.endpoint.String()); err != nil {
-			return ProcessingPhaseError, err
+	if hs.readers.Convert {
+		if hs.brokenForQemuImg || hs.readers.Archived || hs.customCA != "" {
+			return ProcessingPhaseTransferScratch, nil
 		}
-		return ProcessingPhaseConvert, nil
+	} else {
+		if hs.readers.Archived || hs.customCA != "" {
+			return ProcessingPhaseTransferDataFile, nil
+		}
 	}
-	// removing check for hs.brokenForQemuImg, and always assuming it is true
-	// revert once we are able to get nbdkit 1.35.8, which contains a fix for the
-	// slow download speed.
-	return ProcessingPhaseTransferScratch, nil
+	hs.url, _ = url.Parse(fmt.Sprintf("nbd+unix:///?socket=%s", nbdkitSocket))
+	if err = hs.n.StartNbdkit(hs.endpoint.String()); err != nil {
+		return ProcessingPhaseError, err
+	}
+	return ProcessingPhaseConvert, nil
 }
 
 // Transfer is called to transfer the data from the source to a scratch location.
--- a/pkg/importer/http-datasource_test.go
+++ b/pkg/importer/http-datasource_test.go
@@ -94,11 +94,16 @@ var _ = Describe("Http data source", func() {
 		if !wantErr {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(expectedPhase).To(Equal(newPhase))
+			if newPhase == ProcessingPhaseConvert {
+				expectURL, err := url.Parse("nbd+unix:///?socket=/tmp/nbdkit.sock")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(expectURL).To(Equal(dp.GetURL()))
+			}
 		} else {
 			Expect(err).To(HaveOccurred())
 		}
 	},
-		Entry("return TransferScratch phase ", cirrosFileName, cdiv1.DataVolumeKubeVirt, ProcessingPhaseTransferScratch, cirrosData, false),
+		Entry("return Convert phase ", cirrosFileName, cdiv1.DataVolumeKubeVirt, ProcessingPhaseConvert, cirrosData, false),
 		Entry("return TransferTarget with archive content type but not archive endpoint ", cirrosFileName, cdiv1.DataVolumeArchive, ProcessingPhaseTransferDataDir, cirrosData, false),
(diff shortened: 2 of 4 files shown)


pkg/importer/http-datasource.go
```diff
@@ -133,20 +133,20 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
...
-		hs.url, _ = url.Parse(fmt.Sprintf("nbd+unix:///?socket=%s", nbdkitSocket))
-		if err = hs.n.StartNbdkit(hs.endpoint.String()); err != nil {
-			return ProcessingPhaseError, err
+	if hs.readers.Convert {
+		if hs.brokenForQemuImg || hs.readers.Archived || hs.customCA != "" {
+			return ProcessingPhaseTransferScratch, nil
 		}
-		return ProcessingPhaseConvert, nil
+	} else {
+		if hs.readers.Archived || hs.customCA != "" {
```
mhenriks, line 141: Don't ever want to use nbdkit with archive see #2351

mhenriks: So we are good here

pkg/importer/http-datasource.go
```diff
@@ -133,20 +133,20 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
...
+			return ProcessingPhaseTransferDataFile, nil
```
mhenriks, line 142: Should return ProcessingPhaseTransferScratch here if we want to have proper preallocation

akalenyu (commented):
So https://github.com/kubevirt/containerized-data-importer/blob/3abfa1967bb89e1c615063e6d94a16df17dcb352/pkg/importer/data-processor.go#L271 is not true? raw->raw conversion is not just a copy?

I also think we have some testing of sparse images in e2e so worth to see if we can reduce this from the Windows case

akalenyu:
> So
> 
> https://github.com/kubevirt/containerized-data-importer/blob/3abfa1967bb89e1c615063e6d94a16df17dcb352/pkg/importer/data-processor.go#L271
> 
> is not true? raw->raw conversion is not just a copy?
> I also think we have some testing of sparse images in e2e so worth to see if we can reduce this from the Windows case

So looking into this further, I think our error is that we have flows where we don't use `qemu-img convert` to seal the deal before declaring a VM disk image is "ready" to be used.
qemu-img encapsulates a lot of knowledge in itself and this is just one manifestation of something we were missing from it, which is that `io.Copy` [doesn't](https://stackoverflow.com/a/43040498) preserve sparseness just like a regular `cp` call wouldn't if one doesn't specify `--sparse=always`.

I think we should always go through with `qemu-img convert` before marking a (content_type=kubevirt) image ready.

Regarding e2e, I tried uploading a sparse `cirros.raw` in a GH PR and it doesn't preserve the sparseness as well,
so have to come up with a W/A to test against this.

alromeros:
> So looking into this further, I think our error is that we have flows where we don't use `qemu-img convert` to seal the deal before declaring a VM disk image is "ready" to be used. qemu-img encapsulates a lot of knowledge in itself and this is just one manifestation of something we were missing from it, which is that `io.Copy` [doesn't](https://stackoverflow.com/a/43040498) preserve sparseness just like a regular `cp` call wouldn't if one doesn't specify `--sparse=always`.
> 
> I think we should always go through with `qemu-img convert` before marking a (content_type=kubevirt) image ready.
> 
Thanks for the analysis, @akalenyu. Yeah, at least according to this [flowchart](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/http-datasource.go#L58 ) we should be converting all kubevirt imgs, something we aren't doing. I think the safest way to accomplish this PR in a backportable way is to just avoid conversion of archived files and transfer to scratch every other.

I think I'll first do that in this PR so we can backport it safely and then consider using nbdkit again in a follow-up.

pkg/importer/http-datasource.go
alromeros: I saw this comment earlier and understood the opposite https://github.com/kubevirt/containerized-data-importer/pull/2832#discussion_r1283698347, should we return transferScratch when preallocation is false or true? Anyway, I'm thinking of simplifying this PR to just address this bug so we can backport it safely and reconsider the usage of nbdkit again in a follow-up.

akalenyu:
> > So looking into this further, I think our error is that we have flows where we don't use `qemu-img convert` to seal the deal before declaring a VM disk image is "ready" to be used. qemu-img encapsulates a lot of knowledge in itself and this is just one manifestation of something we were missing from it, which is that `io.Copy` [doesn't](https://stackoverflow.com/a/43040498) preserve sparseness just like a regular `cp` call wouldn't if one doesn't specify `--sparse=always`.
> > I think we should always go through with `qemu-img convert` before marking a (content_type=kubevirt) image ready.
> 
> Thanks for the analysis, @akalenyu. Yeah, at least according to this [flowchart](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/http-datasource.go#L58) we should be converting all kubevirt imgs, something we aren't doing. I think the safest way to accomplish this PR in a backportable way is to just avoid conversion of archived files and transfer to scratch every other.
> 
> I think I'll first do that in this PR so we can backport it safely and then consider using nbdkit again in a follow-up.

Yeah but then using scratch for raw is also painful (2x windows image)... let's see what others think

alromeros:
Closing this PR following the conversation during the sig-storage meeting: For consistency, we've decided to prioritize the scratch space flow for most imports as it's reliable and will allow us to keep using qemu-img convert. Will open a follow-up with a simple fix in the importer flow.

PR #2845, opened 2023-08-13
Don't use scratch space for registry node pull imports

akalenyu:
**What this PR does / why we need it**:
When importing via node container runtime, we always have the image available in the container filesystem,
so there should be no reason to use scratch space.
This manifests itself in the form of a bug where we loop over
```bash
E0813 13:32:38.443088       1 data-processor.go:251] scratch space required and none found
E0813 13:32:38.443102       1 importer.go:181] scratch space required and none found
```
On registry node pull imports when images are not raw. Testing for this is added in this PR

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #
https://issues.redhat.com/browse/CNV-32010

**Special notes for your reviewer**:

**Release note**:

```release-note
BugFix: non-raw registry node pull imports not progressing
```

PR #3219, opened 2024-04-22
Improve http import flow to decide whether to use scratch space or not

alromeros:

Follow up for https://github.com/kubevirt/containerized-data-importer/pull/3212.

Due to some performance issues discussed in https://github.com/kubevirt/containerized-data-importer/issues/2809, we stopped using nbdkit for most http imports and decided to use our custom scratch space method for most imports. This new behavior introduced minor differences in some specific flows, such as stopping the [conversion](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/data-processor.go#L268) of uncompressed raw images.

The conversion process made the actual size of raw images significantly smaller, probably because of qemu's handling of sparse images. The import of raw images without conversion ended up causing failures in some tests, as imported images had a significant increase in size.

This PR aims to fix this behavior by using scratch space with more import flows: All archived kubevirt imgs will be transfered direcly to file, while all non-archived kubevirt imgs will be imported to scratch.  

**Example:** 

Fresh image import before this PR:

```sh
sh-5.1$ qemu-img info disk.img 
image: disk.img
file format: raw
virtual size: 70 GiB (75161927680 bytes)
disk size: 55 GiB
```

Fresh image import after this PR (due to convert):

```sh
$ qemu-img info disk.img 
image: disk.img
file format: raw
virtual size: 70 GiB (75161927680 bytes)
disk size: 9.95 GiB
```

Fixes # https://issues.redhat.com/browse/CNV-36026

Check https://github.com/kubevirt/containerized-data-importer/issues/2809 and https://github.com/kubevirt/containerized-data-importer/pull/2832 for more context about the original change.
**Release note**:

```release-note
Bugfix: Use scratch space when importing non-archived images 
```

### DIFF
--- a/pkg/importer/format-readers.go
+++ b/pkg/importer/format-readers.go
@@ -174,15 +174,15 @@ func (fr *FormatReaders) fileFormatSelector(hdr *image.Header) {
 			fr.Archived = true
 			fr.ArchiveZstd = true
 		}
-	case "qcow2":
-		r, err = fr.qcow2NopReader(hdr)
-		fr.Convert = true
 	case "xz":
 		r, err = fr.xzReader()
 		if err == nil {
 			fr.Archived = true
 			fr.ArchiveXz = true
 		}
+	case "qcow2":
+		r, err = fr.qcow2NopReader(hdr)
+		fr.Convert = true
 	case "vmdk":
 		r = nil
 		fr.Convert = true
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -133,9 +133,6 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	if hs.contentType == cdiv1.DataVolumeArchive {
 		return ProcessingPhaseTransferDataDir, nil
 	}
-	if !hs.readers.Convert {
-		return ProcessingPhaseTransferDataFile, nil
-	}
 	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod == string(cdiv1.RegistryPullNode) {
 		hs.url, _ = url.Parse(fmt.Sprintf("nbd+unix:///?socket=%s", nbdkitSocket))
 		if err = hs.n.StartNbdkit(hs.endpoint.String()); err != nil {
@@ -143,9 +140,6 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 		}
 		return ProcessingPhaseConvert, nil
 	}
-	// removing check for hs.brokenForQemuImg, and always assuming it is true
-	// revert once we are able to get nbdkit 1.35.8, which contains a fix for the
-	// slow download speed.
 	return ProcessingPhaseTransferScratch, nil
 }
 
--- a/pkg/importer/http-datasource_test.go
+++ b/pkg/importer/http-datasource_test.go
@@ -103,12 +103,12 @@ var _ = Describe("Http data source", func() {
 		Entry("return TransferTarget with archive content type and archive endpoint ", diskimageTarFileName, cdiv1.DataVolumeArchive, ProcessingPhaseTransferDataDir, diskimageArchiveData, false),
 	)
 
-	It("calling info with raw gz image should return TransferDataFile", func() {
+	It("calling info with raw gz image should return TransferScratch", func() {
 		dp, err = NewHTTPDataSource(ts.URL+"/"+tinyCoreGz, "", "", "", cdiv1.DataVolumeKubeVirt)
 		Expect(err).NotTo(HaveOccurred())
 		newPhase, err := dp.Info()
 		Expect(err).NotTo(HaveOccurred())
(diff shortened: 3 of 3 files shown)


pkg/importer/format-readers.go
```diff
@@ -174,15 +174,15 @@ func (fr *FormatReaders) fileFormatSelector(hdr *image.Header) {
...
-	case "qcow2":
-		r, err = fr.qcow2NopReader(hdr)
-		fr.Convert = true
 	case "xz":
 		r, err = fr.xzReader()
 		if err == nil {
 			fr.Archived = true
 			fr.ArchiveXz = true
 		}
+	case "qcow2":
```
alromeros, line 183: Note for reviewer: Just moving this to group archived/converted formats together.

pkg/importer/http-datasource.go
```diff
@@ -133,7 +133,7 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	if hs.contentType == cdiv1.DataVolumeArchive {
 		return ProcessingPhaseTransferDataDir, nil
 	}
-	if !hs.readers.Convert {
+	if hs.readers.Archived {
 		return ProcessingPhaseTransferDataFile, nil
```
mhenriks, line 137: I say we transfer to scratch for archived as well so I think we can delete this `if`

mhenriks:
wonder if the number of cases in which we don't use scratch for import (I think just registry node pull?) is so limited that we should get rid of the "recreate pod with scratch space" flow?  import controller can be in charge of deciding when scratch is NOT needed

pkg/importer/http-datasource.go
alromeros: I fear we might uncover issues like the one this PR addressed (https://github.com/kubevirt/containerized-data-importer/pull/2845), but I'm ok with the change. Let's see how tests behave.

pkg/importer/http-datasource.go
```diff
@@ -133,19 +133,13 @@ func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	if hs.contentType == cdiv1.DataVolumeArchive {
 		return ProcessingPhaseTransferDataDir, nil
 	}
-	if !hs.readers.Convert {
-		return ProcessingPhaseTransferDataFile, nil
-	}
 	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod == string(cdiv1.RegistryPullNode) {
```
mhenriks, line 136: Planning to try getting rid of nbdkit here?  I don't think it's necessary

alromeros: Didn't this PR (https://github.com/kubevirt/containerized-data-importer/pull/2845) address some bug for using scratch space with pull node imports? @akalenyu wdyt?

mhenriks: The major change in that PR is returning "convert" for registry node pull.  Can still do that.  I just don't think we need nbdkit

alromeros: I didn't have much time to test this today but I added a commit to get rid of nbdkit for pull node. Let's see how tests behave and I'll do more testing tomorrow.

alromeros: @mhenriks seems the version of qemu-img we use doesn't support running convert/info on http endpoints, that's why we need either scratch space or nbdkit. It works on my local version though (odd since I'm using an older version), but since we expect to backport this I think we should keep using nbdkit here.

mhenriks: That does not sound correct to me.  What is the error you see?  Can you check the logs to see what args we are we passing to qemu-img?

alromeros: Just a simple `qemu-img info --output=json http://localhost:8100/tinyCore.iso` returns `qemu-img: Could not open 'http://localhost:8100/tinyCore.iso': Unknown protocol 'http'`. I've tried creating a debugging pod and manually running the command inside and it happens the same. However, running it locally works.

mhenriks: Ah the container must not have `qemu-block-curl` package.

Since updating rpms in old branches is annoying.  We can hold off getting rid of nbdkit in this pr but I think we should get rid of it in a subsequent pr that is not backported

alromeros:
> wonder if the number of cases in which we don't use scratch for import (I think just registry node pull?) is so limited that we should get rid of the "recreate pod with scratch space" flow? import controller can be in charge of deciding when scratch is NOT needed

I like that idea, other than archive imports, vddk and registry node pull I think we can use scratch for every import type. We can leave that for other PR but good think to consider for sure.

mhenriks:
/lgtm
/approve

Thanks @alromeros, this PR is something that can be backported.  For the furure we should:
1.  Get rid of nbdkit, add `qemu-img-block-curl`
2. Looks like import-controller can determine exactly when scratch space is required so let's get rid of the "pod restarts to get scratch space" machinery

PR #3221, opened 2024-04-26
Claim Adoption Doc

mhenriks:

Let's document how to do claim adoption

Fixes #

https://issues.redhat.com/browse/CNV-40262

**Release note**:

```release-note
NONE
```

### DIFF
--- a/doc/datavolume-claim-adoption.md
+++ b/doc/datavolume-claim-adoption.md
@@ -1,0 +1,52 @@
+# DataVolume Claim Adoption
+
+## Summary
+
+Historically, it has been an error to create a `DataVolume` with the same name as an existing `PersistentVolumeClaim`.
+
+```bash
+✗ k get dv dv1
+Error from server (NotFound): datavolumes.cdi.kubevirt.io "dv1" not found
+
+k get pvc dv1
+NAME   STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
+dv1    Bound    pvc-de8b1186-fdd1-4372-8135-9c904ecf73bc   7Gi        RWX            rook-ceph-block   9m7s
+
+✗ k create -f dv.yaml
+Error from server: error when creating "dv.yaml": admission webhook "datavolume-validate.cdi.kubevirt.io" denied the request:  Destination PVC default/dv1 already exists
+```
+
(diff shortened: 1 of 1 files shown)


doc/datavolume-claim-adoption.md
```diff
@@ -0,0 +1,52 @@
...
+```bash
+✗ k get dv dv1
+Error from server (NotFound): datavolumes.cdi.kubevirt.io "dv1" not found
+
+k get pvc dv1
+NAME   STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
+dv1    Bound    pvc-de8b1186-fdd1-4372-8135-9c904ecf73bc   7Gi        RWX            rook-ceph-block   9m7s
+
+✗ k create -f dv.yaml
+Error from server: error when creating "dv3.yaml": admission webhook "datavolume-validate.cdi.kubevirt.io" denied the request:  Destination PVC default/dv1 already exists
```
awels, line 16: Where did dv3.yaml come from? the example is calling `-f dv.yaml`

doc/datavolume-claim-adoption.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+However in certain cases, this behavior is not desirable.  It is often problematic when restoring from a backup.  In that case, the restore process will typically create a new `PersistentVolumeClaim` from a `VolumeSnapshot` and then fail when the corresponding `DataVolume` is created later.  In order to have a proper running application it is important to ensure all resources and their ownership relations are restored correctly.
+
+We now offer two options for `DataVolumes` to "adopt" existing `PersistentVolumeClaims`.  When a `DataVolume` adopts an existing claim, the DataVolume is instantly marked `Succeeded` and an `OwnershipReference` is added to the `PersistentVolumeClaim`.
+
+## cdi.kubevirt.io/allowClaimAdoption Annotation
+
+Add this annoation to your DataVolume when you want to support claim adoption on a granular level.
```
awels, line 25: Typo `annoation`

awels (commented):
Random question, if I enable the feature gate, can I force it to NOT adopt by putting the annotation with `false` on a DataVolume?

mhenriks:
> Random question, if I enable the feature gate, can I force it to NOT adopt by putting the annotation with false on a DataVolume?

Yeah if the annotation exists and the value is anything but 'true' the DV will not adopt the PVC

doc/datavolume-claim-adoption.md
mhenriks: ha good catch locally my file was named dv3.yaml

awels:
/lgtm
it might be a good idea to mention that if you set the annotation to false it will not adopt the claim even if the feature gate is set. Other than that looks good.

PR #3232, opened 2024-05-02
CDIStorageProfilesIncomplete should not impact operator health

arnongilboa:
The alert caused OpenShift Virtualization operator status go degraded.

In addition add stork-snapshot to unsupported provisioners list, as stork snapshot provisioner does not work with CDI, and shouldn't cause CDIStorageProfilesIncomplete alert.

Fixes # https://issues.redhat.com/browse/CNV-39569

**Release note**:
```release-note
BugFix: CDIStorageProfilesIncomplete should not impact operator health
```

### DIFF
--- a/pkg/monitoring/rules/alerts/operator.go
+++ b/pkg/monitoring/rules/alerts/operator.go
@@ -53,7 +53,7 @@ var operatorAlerts = []promv1.Rule{
 		},
 		Labels: map[string]string{
 			severityAlertLabelKey:        "info",
-			operatorHealthImpactLabelKey: "warning",
+			operatorHealthImpactLabelKey: "none",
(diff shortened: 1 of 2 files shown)


arnongilboa:
> CDIStorageProfilesIncomplete

both are dealing with CDIStorageProfilesIncomplete unnecessary noises, and both fixes are originated by the same bug ;)

arnongilboa:
> The changes look good.
> 
> These two changes are completely unrelated. They ideally should be in separate PRs but I understand there is CI overhead to consider so not opposed to them merging as-is.

both are dealing with CDIStorageProfilesIncomplete unnecessary noises, and both fixes are originated by the same bug ;)

PR #3241, opened 2024-05-03
Enable revive linter

EduardGomezEscandell:
Enables the revive linter:
> Fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint.
 - https://golangci-lint.run/usage/linters/#revive

Golint itself has been deprecated for 3 years.

PR #3193 removes golint from the `make lint-metrics`target.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/vddk-datasource_test.go
+++ b/pkg/importer/vddk-datasource_test.go
@@ -666,14 +666,14 @@ var _ = Describe("VDDK get block status", func() {
 		flagSetting := true
 		currentMockNbdFunctions.BlockStatus = func(length uint64, offset uint64, callback libnbd.ExtentCallback, optargs *libnbd.BlockStatusOptargs) error {
 			err := 0
-			len := uint32(MaxBlockStatusLength)
+			L := uint32(MaxBlockStatusLength)
 			if length < MaxBlockStatusLength {
-				len = uint32(length)
+				L = uint32(length)
 			}
 			if flagSetting {
-				callback("base:allocation", offset, []uint32{len, 0}, &err)
+				callback("base:allocation", offset, []uint32{L, 0}, &err)
 			} else {
-				callback("base:allocation", offset, []uint32{len, 3}, &err)
+				callback("base:allocation", offset, []uint32{L, 3}, &err)
 			}
 			flagSetting = !flagSetting
 			return nil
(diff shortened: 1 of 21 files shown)


pkg/importer/vddk-datasource_test.go
```diff
@@ -666,14 +666,14 @@ var _ = Describe("VDDK get block status", func() {
 		flagSetting := true
 		currentMockNbdFunctions.BlockStatus = func(length uint64, offset uint64, callback libnbd.ExtentCallback, optargs *libnbd.BlockStatusOptargs) error {
 			err := 0
-			len := uint32(MaxBlockStatusLength)
+			L := uint32(MaxBlockStatusLength)
```
awels, line 669: why use an upper case L instead of a lower case one?

tools/cdi-func-test-proxy/main.go
```diff
@@ -216,19 +213,10 @@ func main() {
...
-	startServer(httpPortWithAuth, withBasicAuth, false, wg)
-
-	wg.Add(1)
-	startServer(httpsPort, noBasicAuth, true, wg)
-
-	wg.Add(1)
-	startServer(httpsPortWithAuth, withBasicAuth, true, wg)
-
-	wg.Wait()
+	select {}
```
awels, line 221: This doesn't feel semantically the same as the wait groups. Am I missing something? Also I think the log.Fatal in startServer is probably a mistake on my part.

pkg/importer/vddk-datasource_test.go
EduardGomezEscandell: The linter dislikes shadowing built-in variables and functions such as `len`, `max`, etc.

The uppercase L is just my lack of originality

tools/cdi-func-test-proxy/main.go
EduardGomezEscandell: Both things block forever until one of the `startServer` exits the process with `log.Fatal`. The calls to `wg.Done()` are unreachable, so this wg was overkill.

I can rework this to not use log.Fatal and go back to workgroups if you prefer. In that case they wouldn't be overkill.

pkg/importer/vddk-datasource_test.go
awels: Right it is just that normally upper cass variables indicate 'public' type variables. Which of course is not a thing inside a function, it is just confusing. I would just use lower case l and we can call it a day.

EduardGomezEscandell: Sounds good.

awels (approved):
/approve

PR #3242, opened 2024-05-03
VDDK: Fix NBD status coalescing for large blocks.

### LINKED ISSUE #3054: 【VDDK】No new block status data at offset
**What happened**:
When I copy continuous data exceeding 2GB from VMware disk, I noticed that the log keeps printing "No new block status data at offset". After analyzing the code, I found an error in the logic for retrieving blocks in the GetBlockStatus function.

It is expected to be able to complete the copying process normally.

There is an error in the following code block. The value of lastOffset is incorrect when checking whether to merge the previous block.
vddk-datasource_amd64.go：

		for i := 0; i < len(extents); i += 2 {
			length, flags := extents[i], extents[i+1]
			if blocks != nil {
				last := len(blocks) - 1
				lastBlock := blocks[last]
				lastFlags := lastBlock.Flags
				**lastOffset := lastBlock.Offset + uint64(lastBlock.Length)**
				if lastFlags == flags && lastOffset == offset {
					// Merge with previous block
					blocks[last] = &BlockStatusData{
						Offset: lastBlock.Offset,
						Length: lastBlock.Length,
						Flags:  lastFlags,
					}
				} else {
					blocks = append(blocks, &BlockStatusData{Offset: offset, Length: length, Flags: flags})
				}
			} else {
				blocks = append(blocks, &BlockStatusData{Offset: offset, Length: length, Flags: flags})
			}
			offset += uint64(length)
		}

I modified it as follows, and it has been verified to run normally:

		for i := 0; i < len(extents); i += 2 {
			length, flags := extents[i], extents[i+1]
			if blocks != nil {
				last := len(blocks) - 1
				lastBlock := blocks[last]
				lastFlags := lastBlock.Flags
				**lastOffset := lastBlock.Offset**
				if lastFlags == flags && lastOffset == offset {
					// Merge with previous block
					blocks[last] = &BlockStatusData{
						Offset: lastBlock.Offset,
						**Length: lastBlock.Length + length,**
						Flags:  lastFlags,
					}
				} else {
					blocks = append(blocks, &BlockStatusData{Offset: offset, Length: length, Flags: flags})
				}
			} else {
				blocks = append(blocks, &BlockStatusData{Offset: offset, Length: length, Flags: flags})
			}
			offset += uint64(length)
		}

Please help me confirm if this modification is okay. Thank you.

**Additional context**:

**Environment**:
- DV specification: N/A
- Install tools: N/A
- Others: N/A

mrnold:
This pull request fixes the logic error pointed out in #3054, where the VDDK importer tries to merge smaller blocks into one big one, when mapping out the NBD status flags (basically zero, hole, or data) from the source disk. This code was incorrectly neglecting to add the length of the current block to the length of the previous block, so the combined result had the wrong length.

This pull request also avoids a potential integer overflow by changing the type of the struct field that receives this block length from a uint32 to an int64. It further changes the offset field in the same struct from a uint64 to an int64, to avoid a bunch of haphazard integer conversions. There are a few new conversions, but the net result is a bit more confidence-inspiring. 

This pull request also adds unit tests for GetBlockStatus, and fixes and alignment issue in the multi-stage VDDK documentation that tripped me up when trying to remember how to test this stuff.

Fixes #3054

Manually tested single VDDK imports and multi-stage VDDK imports with at least 6GB of changed data, and compared checksums of the results to make sure deltas got applied correctly.

**Release note**:
```release-note
Fix issue #3054, VDDK failure with "No new block status data at offset" messages.
```

### DIFF
--- a/pkg/importer/vddk-datasource_amd64.go
+++ b/pkg/importer/vddk-datasource_amd64.go
@@ -586,9 +586,7 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
 		} else {
 			length = uint64(missingLength)
 		}
-		err := handle.BlockStatus(length, uint64(lastOffset), updateBlocksCallback, &fixedOptArgs)
-		if err != nil {
-			klog.Errorf("Error getting block status at offset %d, returning whole block instead. Error was: %v", lastOffset, err)
+		createWholeBlock := func() []*BlockStatusData {
 			block := &BlockStatusData{
 				Offset: uint64(extent.Start),
 				Length: uint32(extent.Length),
@@ -597,10 +595,16 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
 			blocks = []*BlockStatusData{block}
 			return blocks
 		}
+		err := handle.BlockStatus(length, uint64(lastOffset), updateBlocksCallback, &fixedOptArgs)
+		if err != nil {
+			klog.Errorf("Error getting block status at offset %d, returning whole block instead. Error was: %v", lastOffset, err)
+			return createWholeBlock()
+		}
 		last := len(blocks) - 1
 		newOffset := blocks[last].Offset + uint64(blocks[last].Length)
 		if uint64(lastOffset) == newOffset {
-			klog.Infof("No new block status data at offset %d", newOffset)
+			klog.Infof("No new block status data at offset %d, returning whole block.", newOffset)
+			return createWholeBlock()
 		}
 		lastOffset = int64(newOffset)
 	}
--- a/pkg/importer/vddk-datasource_amd64.go
+++ b/pkg/importer/vddk-datasource_amd64.go
@@ -554,7 +554,7 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
 					// Merge with previous block
 					blocks[last] = &BlockStatusData{
 						Offset: lastBlock.Offset,
-						Length: lastBlock.Length,
+						Length: lastBlock.Length + length,
 						Flags:  lastFlags,
 					}
 				} else {
--- a/pkg/importer/vddk-datasource_amd64.go
+++ b/pkg/importer/vddk-datasource_amd64.go
@@ -26,6 +26,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"net/url"
 	"os"
 	"regexp"
@@ -501,7 +502,7 @@ const MaxPreadLengthVC = (2 << 20)
 // MaxPreadLength is the maxmimum read size to request from VMware. Default to
 // the larger option, and reduce it in createVddkDataSource when connecting to
 // vCenter endpoints.
-var MaxPreadLength uint32 = MaxPreadLengthESX
+var MaxPreadLength int = MaxPreadLengthESX
 
 // NbdOperations provides a mockable interface for the things needed from libnbd.
 type NbdOperations interface {
@@ -513,8 +514,8 @@ type NbdOperations interface {
 
 // BlockStatusData holds zero/hole status for one block of data
 type BlockStatusData struct {
-	Offset uint64
-	Length uint32
+	Offset int64
+	Length int64
 	Flags  uint32
 }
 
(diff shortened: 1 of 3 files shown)


rwmjones:
Adding @ebblake

rwmjones (commented):
[Edit: I meant this to be attached to the first commit in the series, but here we are]

More of a general comment as this is the first time I've seen this mock NBD module.  Why not use something like this for mocking NBD instead?  ie. Use libnbd, but connected to an [nbdkit-data-plugin](https://libguestfs.org/nbdkit-data-plugin.1.html) instance.
```
package libnbd
h, err := Create()
err = h.ConnectCommand([]string{
    "nbdkit", "-s", "--exit-with-parent", "data",
    "1 @8192 2"
})
```
You'll need to read the nbdkit-data-plugin manual to understand the test string, but the important point is it'll allow you to populate a disk with some sparse test data, letting you test the actual block status APIs.

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -554,7 +554,7 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
 					// Merge with previous block
 					blocks[last] = &BlockStatusData{
 						Offset: lastBlock.Offset,
-						Length: lastBlock.Length,
+						Length: lastBlock.Length + length,
```
rwmjones, line 564: This seems like the meat of the change, looks good.

pkg/importer/vddk-datasource_test.go
```diff
@@ -638,6 +638,63 @@ var _ = Describe("VDDK get block status", func() {
...
+			Expect(block.Offset).To(Equal(uint64(index * MaxBlockStatusLength)))
+			Expect(block.Length).To(Equal(uint32(MaxBlockStatusLength)))
+			if index%2 == 0 {
+				Expect(block.Flags).To(Equal(uint32(0)))
+			} else {
+				Expect(block.Flags).To(Equal(uint32(3)))
+			}
+		}
+	})
 })
```
rwmjones, line 698: With a mocking module that just used nbdkit-data-plugin in a subprocess, you could test the various scenarious by supplying different data plugin strings, eg:
```
"1 @65536 2"         # allocated 32K, sparse 32K, allocated 32K
"1 @1048576 2"    # allocated 32K, sparse 1M-32K, allocated 32K
```

rwmjones (commented):
libnbd 1.18 adds a new [BlockStatus64](https://libguestfs.org/nbd_block_status_64.3.html) API.  Example of usage: https://gitlab.com/nbdkit/libnbd/-/blob/master/golang/libnbd_465_block_status_64_test.go?ref_type=heads

mrnold:
> [Edit: I meant this to be attached to the first commit in the series, but here we are]
> 
> More of a general comment as this is the first time I've seen this mock NBD module. Why not use something like this for mocking NBD instead? ie. Use libnbd, but connected to a [nbdkit-data-plugin](https://libguestfs.org/nbdkit-data-plugin.1.html) instance as a child process of libnbd (and hence of the test harness).
> 
> ```
> import "libguestfs.org/libnbd"
> h, err := Create()
> err = h.ConnectCommand([]string{
>     "nbdkit", "-s", "--exit-with-parent", "data",
>     "1 @8192 2"
> })
> ```
> 
> You'll need to read the nbdkit-data-plugin manual to understand the test string, but the important point is it'll allow you to populate a disk with some sparse test data, letting you test the actual block status APIs.

The first commit is just the simplest method I could find for unit testing, and the same technique is used for testing other things like VMware API access in the same file. This data plugin looks ideal for the VDDK functional tests in tests/datavolume_tests.go though, right now they get fixed data from a CirrOS image.



> libnbd 1.18 adds a new [BlockStatus64](https://libguestfs.org/nbd_block_status_64.3.html) API. Example of usage: https://gitlab.com/nbdkit/libnbd/-/blob/master/golang/libnbd_465_block_status_64_test.go?ref_type=heads

This looks useful too, it would take out at least one more integer cast. It's not a huge deal at the moment because the maximum block status request size is set to 2GB for this importer, but it's good to know it could be bumped up significantly. I think CDI is even using 1.18 at runtime, it would just need a Go vendor update to pick it up for development use.

Thanks for taking a look at this. I think this particular code does not get used much, but I didn't want to leave it sitting here broken forever for the one person that reported a problem.

rwmjones:
Sure thing, didn't want to suggest that the code is wrong, it was just a possible way to simplify it.  However if it's also used for testing other VDDK functions, ie ones not used by nbdkit, then it's a little less compelling to change it.

pkg/importer/vddk-datasource_amd64.go
mhenriks: @mrnold looks like `BlockStatusData.Length` is a uint32.  Do we have to be concerned about adding an int64 to that?

mhenriks: okay so length initially comes in as a uint32 but don't we still have to be concerned about overflow?

mrnold: Yes, overflow is a concern, so this pull request changes BlockStatusData.Length to be an int64 in commit 27e07d36a87c52611c1fef4e6a9f5ceb2814c3bf.

mhenriks: Ah okay missed that

PR #3244, opened 2024-05-06
Update DataVolume doc references from the 'pvc' API to the 'storage' API

EduardGomezEscandell:

Updates `DataVolume` examples and documentation so they use the `storage` API instead of the `pvc` API, except for a couple of places where the use of `pvc` is relevant (to contrast it to the `storage` API).

Furthermore:
- Removed the `ReadWriteOnce` argument as it is set by default when using the storage API.
- Removed the storage resources requests when cloning from a PVC.

https://issues.redhat.com/browse/CNV-38189

**Release note**:

```release-note
NONE
```

### DIFF
--- a/doc/clone-datavolume.md
+++ b/doc/clone-datavolume.md
@@ -22,8 +22,6 @@ spec:
       namespace: source-ns
       name: source-datavolume
   storage:
-    accessModes:
-      - ReadWriteOnce
     resources:
       requests:
         storage: 500Mi
--- a/doc/clone-datavolume.md
+++ b/doc/clone-datavolume.md
@@ -21,7 +21,7 @@ spec:
     pvc:
       namespace: source-ns
       name: source-datavolume
-  storage:
+  storage: {}
 ```
 
 Deploy the DataVolume manifest:
--- a/doc/clone-datavolume.md
+++ b/doc/clone-datavolume.md
@@ -22,9 +22,6 @@ spec:
       namespace: source-ns
       name: source-datavolume
   storage:
-    resources:
-      requests:
-        storage: 500Mi
 ```
 
 Deploy the DataVolume manifest:
--- a/doc/clone-datavolume.md
+++ b/doc/clone-datavolume.md
@@ -21,7 +21,7 @@ spec:
     pvc:
       namespace: source-ns
       name: source-datavolume
-  pvc:
+  storage:
     accessModes:
       - ReadWriteOnce
     resources:
--- a/doc/datavolumes.md
+++ b/doc/datavolumes.md
@@ -54,7 +54,7 @@ spec:
          url: "https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img" # S3 or GCS
          secretRef: "" # Optional
          certConfigMap: "" # Optional
-  pvc:
+  storage:
     accessModes:
       - ReadWriteOnce
     resources:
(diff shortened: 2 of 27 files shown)


alromeros:
Thanks for the PR! I think this looks good but I'd update the `datavolumes.md` doc a bit so we prioritize the `## Target Storage/PVC` section more. We could move it above the `source` section and emphasize the possibility of dropping the accessMode if we use the storage api. That justifies the changes a bit more IMO.

EduardGomezEscandell:
@alromeros Sounds good. I also rephrased it a bit to highlight its conveniences.

> [!TIP]
> I also used this sort of banner, I think it looks cool

doc/clone-datavolume.md
```diff
@@ -21,12 +21,7 @@ spec:
     pvc:
       namespace: source-ns
       name: source-datavolume
-  pvc:
-    accessModes:
-      - ReadWriteOnce
-    resources:
-      requests:
-        storage: 500Mi
+  storage:
```
aglitke, line 24: does this need a `{}` in order to be valid yaml?

doc/datavolumes.md
```diff
@@ -38,6 +38,99 @@ The following statuses are possible.
...
+  pvc:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
+```
+
+### Storage
+The `storage` type is similar to `pvc` but it allows you to ommit some parameters.
```
aglitke, line 72: s/ommit/omit/

doc/datavolumes.md
```diff
@@ -38,6 +38,99 @@ The following statuses are possible.
...
+
+If you skip the `volumeMode` parameter, CDI will search for a default value in the
+StorageProfile. See also: [storage profile documentation](storageprofile.md).
+
+If you skip the `storageClassName` name parameter, CDI will prioritize the default
+virtualization storage class over k8s' default. You can define your default
+virtualization storage class by annotating it with
+`storageclass.kubevirt.io/is-default-virt-class` set to `"true"`.
+
+If you request storage with `volumeMode: fileSystem`, CDI will take the file system
```
aglitke, line 86: We should word this slightly differently so say that fs overhead is applied whenever filesystem volumeMode is used in conjunction with the storage api (whether specified explicitly or chosen automatically).  As you currently have it it sounds like this happens only if you include `volumeMode: FileSystem` in the spec.

doc/datavolumes.md
```diff
@@ -38,6 +38,99 @@ The following statuses are possible.
 * Failed: The operation has failed.
 * Unknown: Unknown status.
 
+## Target Storage/PVC
```
aglitke, line 41: Thanks for adding this!  We needed something like this.

doc/clone-datavolume.md
EduardGomezEscandell: Good catch :)

doc/datavolumes.md
EduardGomezEscandell: Makes sense :+1:

EduardGomezEscandell:
pull-containerized-data-importer-e2e-upg [failed during set-up](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3244/pull-containerized-data-importer-e2e-upg/1790335126022918144) and not during the test

/restest

EduardGomezEscandell:
pull-containerized-data-importer-e2e-upg [failed during set-up](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3244/pull-containerized-data-importer-e2e-upg/1790335126022918144) and not during the test

/retest

awels (approved):
/approve

kgoldbla (commented):
looks fine

PR #3245, opened 2024-05-06
Fix cluster-sync wait_cdi_crd_installed timeout check

EduardGomezEscandell:
Fixes a line in the cluster-up script that was writing to file `./0` instead of performing a greater-than comparison.

I always had this suspicious file in my git diff and after a couple greps for `> 0`, I found the culprit.

**Release note**:

```release-note
NONE
```


akalenyu:
/approve
thanks!

PR #3249, opened 2024-05-08
Fix DELL storagecapabilities

ido106:

The [current storagecapabilities ](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/storagecapabilities/storagecapabilities.go#L295)for Dell PowerMax, PowerFlex, PowerScale and PowerStore are defined as: 
`{ {rwx, block}, {rwo, block}, {rwo,file} } `

According to the [DELL documentation](https://dell.github.io/csm-docs/docs/csidriver/), it should be:

PowerMax:
`{ {rwx, block}, {rwo, block}, {rwo, file}, {rox, block} } `
PowerFlex: 
`{ {rwx, block}, {rwo, block}, {rwo, file}, {rox, block}, {rox, file} } `
PowerScale  (block volume mode is not supported):
`{ {rwx, file}, {rwo, file}, {rox, file} } `
PowerStore: 
`{ {rwx, block}, {rwo, block}, {rwo, file}, {rox, block} } `

If the `nfs` protocol is used, it is `{ {rwx, block}, {rwo, block}, {rwo, file} } ` for all types (except for PowerScale)

Fixes #[CNV-41538](https://issues.redhat.com/browse/CNV-41538)

**Release note**:
```release-note
Fix DELL storagecapabilities
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -50,28 +50,32 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"openshift-storage.cephfs.csi.ceph.com": {{rwx, file}},
 	// LINSTOR
 	"linstor.csi.linbit.com": createLinstorCapabilities(),
-	// dell-unity-csi
-	"csi-unity.dellemc.com": createDellUnityCapabilities(),
-	// PowerFlex
-	"csi-vxflexos.dellemc.com": createDellPowerCapabilities(),
-	// PowerScale
-	"csi-isilon.dellemc.com": createDellPowerCapabilities(),
-	// PowerMax
-	"csi-powermax.dellemc.com": createDellPowerCapabilities(),
-	// PowerStore
-	"csi-powerstore.dellemc.com": createDellPowerCapabilities(),
+	// DELL Unity XT
+	"csi-unity.dellemc.com":     createDellUnityCapabilities(),
+	"csi-unity.dellemc.com/nfs": createAllFSCapabilities(),
+	// DELL PowerFlex
+	"csi-vxflexos.dellemc.com":     createDellPowerFlexCapabilities(),
+	"csi-vxflexos.dellemc.com/nfs": createAllFSCapabilities(),
+	// DELL PowerScale
+	"csi-isilon.dellemc.com": createAllFSCapabilities(),
+	// DELL PowerMax
+	"csi-powermax.dellemc.com":     createDellPowerMaxCapabilities(),
+	"csi-powermax.dellemc.com/nfs": createAllFSCapabilities(),
+	// DELL PowerStore
+	"csi-powerstore.dellemc.com":     createDellPowerStoreCapabilities(),
+	"csi-powerstore.dellemc.com/nfs": createAllFSCapabilities(),
 	// storageos
 	"kubernetes.io/storageos": {{rwo, file}},
 	"storageos":               {{rwo, file}},
-	//AWSElasticBlockStore
+	// AWSElasticBlockStore
 	"kubernetes.io/aws-ebs": {{rwo, block}},
 	"ebs.csi.aws.com":       {{rwo, block}},
-	//AWSElasticFileSystem
+	// AWSElasticFileSystem
 	"efs.csi.aws.com": {{rwx, file}, {rwo, file}},
-	//Azure disk
+	// Azure disk
 	"kubernetes.io/azure-disk": {{rwo, block}},
 	"disk.csi.azure.com":       {{rwo, block}},
-	//Azure file
+	// Azure file
 	"kubernetes.io/azure-file": {{rwx, file}},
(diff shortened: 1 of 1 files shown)


arnongilboa (commented):
@ido106 , please keep the capabilities sorted as the existing ones:
```
		{rwx, block},
		{rwx, file},
		{rwo, block},
		{rwo, file},
		{rox, block},
		{rox, file},

```

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -289,11 +289,39 @@ func createDellUnityCapabilities() []StorageCapabilities {
 	}
 }
 
-func createDellPowerCapabilities() []StorageCapabilities {
+func createDellPowerMaxCapabilities() []StorageCapabilities {
 	return []StorageCapabilities{
 		{rwx, block},
```
arnongilboa, line 294: @akalenyu according to the Dell doc PowerMax/PoweFlwex/PowerStore suppport RWX/ROX for Filesystem volume mode **with NFS only**.  Shouldn't we add them to the capabilities?

akalenyu: Yeah, there's a good chance this driver needs us to differentiate over one of storage class parameters,
for example, I see FibreChannel/ISCSI/NFS examples in here:
https://github.com/dell/csi-powermax/tree/main/samples/storageclass
Which could mean that somebody using powermax-nfs will fail to create block volumes altogether https://github.com/kubevirt/containerized-data-importer/blob/842342452f5028098453d43f7aa4bb4ec0d67e67/pkg/storagecapabilities/storagecapabilities.go#L292-L294 But I don't see anything obvious to key off in the storage class examples.

TBH usually it's best to have the storage provider make the recommendation

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -278,6 +282,42 @@ var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageCl
...
+	"csi-vxflexos.dellemc.com": func(sc *storagev1.StorageClass) string {
+		// https://github.com/dell/csi-powerflex/tree/main/samples/storageclass
+		switch strings.ToLower(sc.Parameters["csi.storage.k8s.io/fstype"]) {
+		case "nfs":
+			return "csi-vxflexos.dellemc.com/nfs"
+		default:
+			return "csi-vxflexos.dellemc.com"
+		}
+	},
+	"csi-powermax.dellemc.com": func(sc *storagev1.StorageClass) string {
```
ido106, line 312: If it's ok now I'll try to add the last three to a function, because they have the same string key (`csi.storage.k8s.io/fstype`)

pkg/storagecapabilities/storagecapabilities.go
akalenyu: So the way they indicate this is nfs is via `csi.storage.k8s.io/fstype`?
This is a little surprising since this key is normally used as a configurable to specify which filesystem you want created on top of the block device (ext4/xfs/...) For example - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md

ido106: After checking I think you are right. I can't find anything else that mentions `nfs` or `protocol` in the code in our context (except [this](https://github.com/dell/csi-powerflex/blob/7b8d376691aef258233c69f2afcab41f9974ef1a/service/node.go#L151) but I don't think that's what we're looking for). Only DELL unity mentions the protocol explicitly as in the example [here](https://github.com/dell/csi-unity/blob/1f42af327f4130df65c5532f6029559e4ab579b5/samples/storageclass/unity-nfs.yaml#L50), but all the others do not.
Maybe we'll just ask a contact from DELL what the string is, if we have one?

akalenyu (commented):
/approve

ido106:
At last we used the `csi.storage.k8s.io/fstype` parameter because this is how DELL differentiate between the types:

> Hi [@iaharon](https://redhat-internal.slack.com/team/U06CUQ5ENSJ),
That parameter is not uniform among the drivers. IMHO to detect NFS we should rely on [csi.storage.k8s.io/fstype](http://csi.storage.k8s.io/fstype) which is more of a canonical name : https://kubernetes-csi.github.io/docs/external-provisioner.html#storageclass-parameters

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -280,12 +284,55 @@ var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageCl
 		}
 	},
 	"csi.vsphere.vmware.com": func(sc *storagev1.StorageClass) string {
-		fsType := sc.Parameters["csi.storage.k8s.io/fstype"]
+		fsType := getFSType(sc)
```
ido106, line 287: Just added `getFSType` func in the last push

PR #3251, opened 2024-05-08
Update bazel-build-builder.sh

### LINKED ISSUE #3248: Parameterize hard coded environment variables in bazel-build-builder.sh
Currently the bazel-build-builder.sh script has the registry and conditions under which it is run hard-coded.
This is awkward for the development and test of new builds for new architectures. 
Also it just seems wrong to have the target registry and run conditions hard coded like that

Suggest we make the current hard-coded environment variables the default values, so we can change them on the fly in test if we wish.  Also, the conditions under which the builder image will be re-built are rather absolute.  Suggest we have a boolean environment variable 'ADHOC' which, if set to true, will bypass the test for whether this is a post-submit job which otherwise prevents the builder image from being built or pushed.

We've been considering restructuring CDI build system to be more like the kubevirt build system, and documenting a single consistent set of environment variables, but that's a much larger project. This fix will at least allow us to make easier progress getting there. 

Since bazel-build-builder.sh also fails if it can't push the image to a repo, it would be nice to separate out the build function from the push function, but that can be a separate issue. 

**Additional context**:
Although this feature request is made in support of upstreaming an s390x port, it is not s390x specific

cfilleke:
- uses DOCKER_PREFIX to set the UNTAGGED_BUILDER_IMAGE 
 - separates out the logic to calculate whether this is a POST_SUBMIT job from the test to execute builder build or not
 - tests to see if ADHOC_BUILDER is set so we can more easily build test builder images in development by setting this env var
 - see https://github.com/kubevirt/containerized-data-importer/issues/3248

When porting CDI to other platforms or debugging the builder image, we need to do ADHOC_BUILDER images for test. By adding a test to see if this environment variable is set in addition to testing whether this is a POST_SUBMIT build, we can more easily build test CDI builder images. Using the existing DOCKER_PREFIX as set in the config.sh script, the registry and repo is set more consistently with the build system. 

Fixes #3248 

This is an incremental step towards a more consistent build system. 

**Release note**:

```release-note
Fixes #3248 
```

### DIFF
--- a/hack/build/bazel-build-builder.sh
+++ b/hack/build/bazel-build-builder.sh
@@ -28,15 +28,29 @@ if [ "${CDI_CONTAINER_BUILDCMD}" = "buildah" ]; then
     fi
 fi
 
-if ! git diff-index --quiet HEAD~1 hack/build/docker; then
-    #Since this only runs during the post-submit job, the PR will have squashed into a single
-    #commit and we can use HEAD~1 to compare.
+# When this runs during the post-submit job, the PR will have squashed into a single
+# commit and we can use HEAD~1 to compare. The exit code of the git diff will therefore
+# be 0 in this case, so negating it indicates that yes this is a post-submit job and
+# we should re-build the builder. Separating out this logic from the test for clarity
+
+git diff-index --quiet HEAD~1 hack/build/docker
+POST_SUBMIT=$((1-$?))
+
+# The other circumstance in which we need to build the builder image is
+# in the course of test and development of the builder image itself. 
+# we'll signal this circumstance by setting the env variable ADHOC_BUILDER
+#
+# also instead of hard-coding the UNTAGGED_BUILDER_IMAGE, we're going
+# to use DOCKER_PREFIX as it is set in config.sh and used elsewhere in 
+# the build system, to introduce a little more consistency  
+
+if ((POST_SUBMIT)) || [ x"${ADHOC_BUILDER}" != "x" ] ; then
     BUILDER_SPEC="${BUILD_DIR}/docker/builder"
-    UNTAGGED_BUILDER_IMAGE=quay.io/kubevirt/kubevirt-cdi-bazel-builder
+    UNTAGGED_BUILDER_IMAGE=$(DOCKER_PREFIX)/kubevirt-cdi-bazel-builder
     BUILDER_TAG=$(date +"%y%m%d%H%M")-$(git rev-parse --short HEAD)
     BUILDER_MANIFEST=${UNTAGGED_BUILDER_IMAGE}:${BUILDER_TAG}
-    echo "$DOCKER_PREFIX:$DOCKER_TAG"
-
+    echo "export BUILDER_IMAGE=$BUILDER_MANIFEST"
+    
     #Build the encapsulated compile and test container
     if [ "${CDI_CONTAINER_BUILDCMD}" = "buildah" ]; then
         (cd ${BUILDER_SPEC} && buildah build ${BUILDAH_PLATFORM_FLAG} --manifest ${BUILDER_MANIFEST} .)


hack/build/bazel-build-builder.sh
```diff
@@ -28,15 +28,29 @@ if [ "${CDI_CONTAINER_BUILDCMD}" = "buildah" ]; then
...
+
+# The other circumstance in which we need to build the builder image is
+# in the course of test and development of the builder image itself. 
+# we'll signal this circumstance by setting the env variable ADHOC_BUILDER
+#
+# also instead of hard-coding the UNTAGGED_BUILDER_IMAGE, we're going
+# to use DOCKER_PREFIX as it is set in config.sh and used elsewhere in 
+# the build system, to introduce a little more consistency  
+
+if ((POST_SUBMIT)) || [ x"${ADHOC_BUILDER}" != "x" ] ; then
```
awels, line 47: Why not use `-eq` and `-z` or `-n` so something like
```bash
git diff-index --quiet HEAD~1 hack/build/docker

if [[ $? -eq 0 ]] || [[ -n ${ADHOC_BUILDER} ]]; then
...
fi
```

mhenriks: @awels if you feel strongly about this feel free to hold the PR

awels: I don't feel super strongly about it. It is not like this is run a ton of times and requires optimization. We are creating subshells with the `$((1-$?))` where a simple number comparison will do.

cfilleke:
I wonder why pull-containerized-data-importer-e22-upg keeps failing. First it was a timeout, but now we see

```
 + make cluster-sync
./cluster-sync/clean.sh
selecting podman as container runtime
Cleaning up ...
error: the server doesn't have a resource type "ot"
error: the server doesn't have a resource type "dv"
error: the server doesn't have a resource type "dataimportcron"
error: the server doesn't have a resource type "cdi"
No resources found
No resources found
No resources found ```

This seems to have nothing to do with the bazel-build-builder.sh script

**Comment by awels** [comment]
The error appears to be
```
 Error in download_and_extract: java.io.IOException: Error downloading [https://mirror.bazel.build/openjdk/azul-zulu11.50.19-ca-jdk11.0.12/zulu11.50.19-ca-jdk11.0.12-linux_x64.tar.gz] to /root/.cache/bazel/_bazel_root/e03bf9038fc5089c0dbc8615812d9838/external/remotejdk11_linux/temp18280869944155108110/zulu11.50.19-ca-jdk11.0.12-linux_x64.tar.gz: Bytes read 177470144 but wanted 202599005
ERROR: /root/.cache/bazel/_bazel_root/e03bf9038fc5089c0dbc8615812d9838/external/bazel_tools/tools/jdk/BUILD:29:19: @bazel_tools//tools/jdk:current_java_runtime depends on @remotejdk11_linux//:jdk in repository @remotejdk11_linux which failed to fetch. no such package '@remotejdk11_linux//': java.io.IOException: Error downloading [https://mirror.bazel.build/openjdk/azul-zulu11.50.19-ca-jdk11.0.12/zulu11.50.19-ca-jdk11.0.12-linux_x64.tar.gz] to /root/.cache/bazel/_bazel_root/e03bf9038fc5089c0dbc8615812d9838/external/remotejdk11_linux/temp18280869944155108110/zulu11.50.19-ca-jdk11.0.12-linux_x64.tar.gz: Bytes read 177470144 but wanted 202599005
ERROR: Analysis of target '//cmd/cdi-cloner:go_default_test' failed; build aborted: 
```

Maybe a broken mirror?

cfilleke:
It's got that look about it, doesn't it? Who would we alert to check on this, Alex? 
@awels

awels:
will try one more time, if not I can override since all the other lanes passed.

PR #3252, opened 2024-05-11
Make upload client/server certs configurable

### LINKED ISSUE #2755: Add ability to set cdi-uploadserver-client-cert certificate renewal period
We deploy containerized-data-importer along with kubevirt in our k8s clusters. By default both applications generate internal certificates for 24hours. In our case it isn't critical to renew certificates so often and we decided to increase it via certConfig option, for example:
```
        certConfig:
          ca:
            duration: 87660h
            renewBefore: 86940h
          server:
            duration: 87660h
            renewBefore: 86940h
 ```
This works fine for `cdi-apiserver-server-cert` and `cdi-uploadproxy-server-cert`, but not for `cdi-uploadserver-client-cert`. Please correct me if I am wrong, but the time [is hardcoded here](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/operator/resources/cert/factory.go#L120). There is no `Configurable: true` for this certificate, that's why the setting isn't working for it. Am I missing something?

I would like to be able to set duration and renewBefore options for this certificate as well. Probably it will be enough just to add `Configurable: true` in the code.

### PR DESCRIPTION
**Pull request by mhenriks**

Prior to this PR, we used static "long term" certs for uploadservers and upload clients (uploadproxy, clone source pods).  With this change, uploadservers will inherit the same duration/renew parameters as cdi-apiserver and cdi-uploadproxy.  This PR also adds the ability to explicitly configure client certs on the CDI resource.

Since we don't want to give cdi-deployment permission to "get secrets" in all namespaces, rotating uploadservers is a little complicated.  Uploadservers may be configured to have a deadline.  The deadline is a time in which the uploadserver should exit (if there is not currently an upload in progress).  The deadline corresponds to the `RenewBefore` parameter in `CertConfig`.  Upload-controller will detect when pods exit via deadline and create new certs + pod.

Fixes #2755

https://issues.redhat.com/browse/CNV-11164

**Release note**:

```release-note
Make upload client/server certs configurable and rotate uploadserver pods
```


awels (approved):
one little nitpick, but feel free to ignore and approve.

mhenriks:
@awels I added another commit to enable deadline/rotation for clone target pods.  I forgot about the case when a clone source may be in use.  When that happens, there may be a significant time gap between the target and source pods are created.  My initial assumption was that rotate was not necessary since source would be sending data immediately

awels (approved):

PR #3254, opened 2024-05-12
Progress metrics refactor and rename

arnongilboa:
`clone_progress` and `openstack_populator_progress` and `ovirt_progress` metrics are not in the monitoring package. These metrics are not with correct metrics names, based on the kubevirt and Prometheus metrics naming conventions, they are not documented and are not located under `/pkg/monitoring`. After the code refactoring we should not have Prometheus metrics in other places in the code, other than the `/monitoring/metrics` package, and metrics should be registered using operator-observability package.

In addition, we aligned the progress metric names with the metric naming linter rules: `kubevirt_cdi_clone_progress_total`, `kubevirt_cdi_openstack_populator_progress_total`, `kubevirt_cdi_ovirt_progress_total`.

Fixes # https://issues.redhat.com/browse/CNV-36845

**Release note**:
```release-note
NONE
```

### DIFF
--- a/doc/metrics.md
+++ b/doc/metrics.md
@@ -3,6 +3,9 @@
 ### kubevirt_cdi_clone_pods_high_restart
 The number of CDI clone pods with high restart count. Type: Gauge.
 
+### kubevirt_cdi_clone_progress_total
+The clone progress in percentage. Type: Counter.
+
 ### kubevirt_cdi_cr_ready
 CDI install ready. Type: Gauge.
 
@@ -15,8 +18,14 @@ Number of DataVolumes pending for default storage class to be configured. Type: 
 ### kubevirt_cdi_import_pods_high_restart
 The number of CDI import pods with high restart count. Type: Gauge.
 
+### kubevirt_cdi_openstack_populator_progress_total
+Progress of volume population. Type: Counter.
+
 ### kubevirt_cdi_operator_up
 CDI operator status. Type: Gauge.
+
+### kubevirt_cdi_ovirt_progress_total
+Progress of volume population. Type: Counter.
 
 ### kubevirt_cdi_storageprofile_info
 `StorageProfiles` info labels: `storageclass`, `provisioner`, `complete` indicates if all storage profiles recommended PVC settings are complete, `default` indicates if it's the Kubernetes default storage class, `virtdefault` indicates if it's the default virtualization storage class, `rwx` indicates if the storage class supports `ReadWriteMany`, `smartclone` indicates if it supports snapshot or CSI based clone. Type: Gauge.
--- a/pkg/monitoring/metrics/cdi-controller/metrics.go
+++ b/pkg/monitoring/metrics/cdi-controller/metrics.go
@@ -15,8 +15,3 @@ func SetupMetrics() error {
 		dataVolumeMetrics,
 	)
 }
-
-// ListMetrics registered prometheus metrics
-func ListMetrics() []operatormetrics.Metric {
-	return operatormetrics.ListMetrics()
-}
--- a/pkg/monitoring/metrics/openstack-populator/populator_metrics.go
+++ b/pkg/monitoring/metrics/openstack-populator/populator_metrics.go
@@ -1,0 +1,34 @@
+package openstackpopulator
+
+import (
+	"github.com/machadovilaca/operator-observability/pkg/operatormetrics"
+	ioprometheusclient "github.com/prometheus/client_model/go"
+)
+
+var (
+	populatorMetrics = []operatormetrics.Metric{
+		populatorProgress,
+	}
+
+	populatorProgress = operatormetrics.NewCounterVec(
+		operatormetrics.MetricOpts{
+			Name: "openstack_populator_progress",
+			Help: "Progress of volume population",
+		},
+		[]string{"ownerUID"},
+	)
+)
+
+// AddPopulatorProgress adds value to the populatorProgress metric
+func AddPopulatorProgress(labelValue string, value float64) {
+	populatorProgress.WithLabelValues(labelValue).Add(value)
+}
+
+// GetPopulatorProgress returns the populatorProgress value
+func GetPopulatorProgress(labelValue string) (float64, error) {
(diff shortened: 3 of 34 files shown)


pkg/monitoring/metrics/openstack-populator/populator_metrics.go
```diff
@@ -0,0 +1,25 @@
...
+
+var (
+	populatorMetrics = []operatormetrics.Metric{
+		populatorProgress,
+	}
+
+	populatorProgress = operatormetrics.NewGaugeVec(
+		operatormetrics.MetricOpts{
+			//Subsystem: "volume_populators",
+			Name: "openstack_volume_populator",
```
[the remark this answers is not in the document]
sradco: If we are generating the metric it should have `kubevirt_cdi_` prefix

pkg/monitoring/metrics/ovirt-populator/populator-metrics.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var (
+	populatorMetrics = []operatormetrics.Metric{
+		populatorProgress,
+	}
+
+	populatorProgress = operatormetrics.NewCounterVec(
+		operatormetrics.MetricOpts{
+			Name: "ovirt_progress",
```
sradco, line 15: Same. The metric requires kubevirt_cdi_ prefix

pkg/monitoring/metrics/openstack-populator/populator_metrics.go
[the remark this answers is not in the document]
assafad: Also missing `_total` suffix, since it's a counter.

pkg/monitoring/metrics/ovirt-populator/populator-metrics.go
assafad: Also missing _total suffix, since it's a counter.

assafad:
@arnongilboa We have a metric name linter, can you please list the new metrics in 
https://github.com/kubevirt/containerized-data-importer/blob/main/tools/prom-metrics-collector/metrics_json_generator.go? You can then use it locally with `make lint-metrics` and see what names you need to fix. 
@machadovilaca @avlitman @sradco FYI, when new metrics groups are added, the developer need to add the new group to the metrics names linter.

assafad:
Unrelated to the PR, we have a small issue in https://github.com/kubevirt/containerized-data-importer/blob/d2f3903e974b53a69e42d810f6507f783851408a/tools/prom-metrics-collector/metrics_json_generator.go#L37
Which uses https://github.com/kubevirt/containerized-data-importer/blob/d2f3903e974b53a69e42d810f6507f783851408a/tools/prom-metrics-collector/metrics_json_generator.go#L11
and this lists only one group of metrics.
We need to use `"github.com/machadovilaca/operator-observability/pkg/operatormetrics"` and then `operatormetrics.ListMetrics()`, which lists all registered metrics. We will open a PR for that.

pkg/monitoring/metrics/ovirt-populator/populator-metrics.go
arnongilboa: @bennyz wdyt?

bennyz:
/lgtm

for forklift populators

pkg/monitoring/metrics/operator-controller/metrics.go
```diff
@@ -13,8 +13,3 @@ func SetupMetrics() error {
 		operatorMetrics,
 	)
 }
-
-// ListMetrics registered prometheus metrics
-func ListMetrics() []operatormetrics.Metric {
-	return operatormetrics.ListMetrics()
-}
```
assafad, line 20: @arnongilboa is removing all of those because they all do the same: `return operatormetrics.ListMetrics()`. 
Why did we do it instead of calling it directly?

assafad:
Thanks @arnongilboa! LGTM :)
Over to @machadovilaca, as he also had a few comments.

pkg/monitoring/metrics/cdi-controller/metrics.go
```diff
@@ -15,8 +15,3 @@ func SetupMetrics() error {
 		dataVolumeMetrics,
 	)
 }
-
-// ListMetrics registered prometheus metrics
-func ListMetrics() []operatormetrics.Metric {
-	return operatormetrics.ListMetrics()
-}
```
assafad, line 22: @machadovilaca, @arnongilboa is removing all of those because they all do the same: `return operatormetrics.ListMetrics()`. 
Why did we do it in the first place, instead of calling it directly?

awels (approved):
Will wait for the metrics folks to give their blessing, but code wise looks good to me.

machadovilaca (commented):
nit: I think we could avoid calling `SetupMetrics` in the `init` functions, but let's not block this because of that

doc/metrics.md
```diff
@@ -15,9 +18,15 @@ Number of DataVolumes pending for default storage class to be configured. Type:
 ### kubevirt_cdi_import_pods_high_restart
 The number of CDI import pods with high restart count. Type: Gauge.
 
+### kubevirt_cdi_openstack_populator_progress_total
+Progress of volume population. Type: Counter.
+
 ### kubevirt_cdi_operator_up
 CDI operator status. Type: Gauge.
 
+### kubevirt_cdi_ovirt_progress_total
```
sradco, line 27: @arnongilboa shouldn't the metric name include populator like it is for openstack?

sradco: @arnongilboa you update the metrics description to be clearer? The are the same for both the openstack and ovirt metrics

arnongilboa: @bennyz please consider doing it in a follow-up PR

sradco: Sure. I'm OK with that.

sradco:
/lgtm

kubevirt_cdi_ovirt_progress_total metric name and description would be updated in a follow-up PR.
This PR is urgent to unblock the metrics code refactoring linter PR.

PR #3265, opened 2024-05-16
Fix metrics name linter

assafad:
Before this PR, the metrics name linter only listed one group of metrics. It listed metrics with: https://github.com/kubevirt/containerized-data-importer/blob/d2f3903e974b53a69e42d810f6507f783851408a/tools/prom-metrics-collector/metrics_json_generator.go#L37
Which uses https://github.com/kubevirt/containerized-data-importer/blob/d2f3903e974b53a69e42d810f6507f783851408a/tools/prom-metrics-collector/metrics_json_generator.go#L11
and this lists only one group of metrics.
We need to use `"github.com/machadovilaca/operator-observability/pkg/operatormetrics"` and then `operatormetrics.ListMetrics()`, which lists all registered metrics.

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/operator/resources/crds_generated.go
+++ b/pkg/operator/resources/crds_generated.go
@@ -6866,7 +6866,7 @@ status:
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.13.0
+    controller-gen.kubebuilder.io/version: v0.14.0
   creationTimestamp: null
   name: openstackvolumepopulators.forklift.konveyor.io
 spec:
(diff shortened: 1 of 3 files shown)


EduardGomezEscandell:
Hi, it seems like the diff does not match the description of this PR.

You renamed the import alias but the URL is still `kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/operator-controller`

/hold

assafad:
FYI, I ran `make generate` locally and it changed `pkg/operator/resources/crds_generated.go`.

EduardGomezEscandell:
> FYI, I ran make generate locally and it changed pkg/operator/resources/crds_generated.go.

Should be fixed as of #3255

Did you rebase recently?

assafad:
/close
There isn't an issue. https://github.com/kubevirt/containerized-data-importer/blob/d2f3903e974b53a69e42d810f6507f783851408a/pkg/monitoring/metrics/cdi-controller/metrics.go#L20-L22

is using `operatormetrics.ListMetrics()`, so it will list all metrics. This makes me think this func completely redundant. 
Will be handeled in https://github.com/kubevirt/containerized-data-importer/pull/3254.

PR #3268, opened 2024-05-17
Filter out generated files from coverage report

EduardGomezEscandell:
As of Go 1.22, packages with no tests are included in the coverage report. For this project, it meant that the coverage plummeted from 61% down to 17%. This is not because of a lack of tests, but rather the inclusion of >60k lines of generated code that are now considered 'relevant lines'. The project size went from 22k to 90k relevant lines, diluting our coverage.

This PR filters out all the generated files from the coverage report, both in Goveralls and locally when running `make coverage`. This is done programmatically rather than using an allowlist/denylist so that we don't need to maintain it.

After this PR, the number of relevant lines becomes 27k, still greater than the pre-Go1.22 metric. This means that we have around 5k lines of code inside packages that have zero tests.

**Release note**:

```release-note
NONE
```


EduardGomezEscandell:
/hold

I added the -x flag to a script for better debugging, must be removed before merging.

PR #3271, opened 2024-05-19
Onboard Infinibox CSI driver

akalenyu:
Reference - https://github.com/Infinidat/infinibox-csi-driver/tree/develop/deploy/examples
alongside some offline chats with recommendations.

Fixes #

**Release note**:

```release-note
Onboard Infinibox CSI driver
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -105,6 +105,9 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"manila.csi.openstack.org": {{rwx, file}},
 	// ovirt csi
 	"csi.ovirt.org": createRWOBlockAndFilesystemCapabilities(),
+	// Infinidat
+	"infinibox-csi-driver/iscsiorfibrechannel": {{rwx, block}, {rwo, block}, {rwo, file}},
+	"infinibox-csi-driver/nfs":                 {{rwx, file}, {rwo, file}},
 }
 
 // SourceFormatsByProvisionerKey defines the advised data import cron source format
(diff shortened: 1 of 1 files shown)


pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -262,6 +267,17 @@ var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageCl
 		}
 		return "UNKNOWN"
 	},
+	"infinibox-csi-driver": func(sc *storagev1.StorageClass) string {
```
akalenyu, line 270: @awels wasn't there a guideline to structuring the provisioner string? this one seems off but definitely the correct one

awels: So in general if there is a `/` in the provisioner string, it is in an in tree driver. And csi drivers will have all `.` in them. It is weird to see a csi driver with a `/` in there. But there are no rules against it I guess.

awels: I guess in this case the `/` is differentiating the type of back-end used for the storage class.

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -229,7 +234,7 @@ func storageProvisionerKey(sc *storagev1.StorageClass) string {
 
 var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageClass) string{
 	"pxd.openstorage.org": func(sc *storagev1.StorageClass) string {
-		//https://docs.portworx.com/portworx-install-with-kubernetes/storage-operations/create-pvcs/create-shared-pvcs/
+		// https://docs.portworx.com/portworx-install-with-kubernetes/storage-operations/create-pvcs/create-shared-pvcs/
```
ido106, line 237: All these links are broken

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -262,6 +267,17 @@ var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageCl
 		}
 		return "UNKNOWN"
 	},
+	"infinibox-csi-driver": func(sc *storagev1.StorageClass) string {
+		// https://github.com/Infinidat/infinibox-csi-driver/tree/develop/deploy/examples
+		switch sc.Parameters["storage_protocol"] {
```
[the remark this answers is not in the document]
akalenyu: how come? they use all lower case

pkg/storagecapabilities/storagecapabilities.go
akalenyu: yeah since this tends to change maybe I could just add a comment that portworx docs served as the guideline?

ido106: I also think so

pkg/storagecapabilities/storagecapabilities.go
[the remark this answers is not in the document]
ido106: I saw an example in another repo that used upper case ('NFS'), and they handled it in the code with `lower()`. I was not sure if it's the same situation here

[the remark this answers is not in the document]
akalenyu: yeah unfortunately all of these parameters are not unified under a clear API, so each provisioner does things differently..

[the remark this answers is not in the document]
akalenyu: I can say that for infinibox they definitely only do lower case

PR #3280, opened 2024-05-21
forklift: fix race between prime pvc and pod removal

bennyz:
Because there is time until deletionTimestamp is set on the PVC prime, and since the populator pod is deleted first we might reconcile again when the PVC is not yet marked for deletion, causing the populator pod to be recreated.

This change reorders the operation, to first wait until the PVC is marked for deletion, and then delete the populator pod.

Fixes #

**Release note**:

```release-note
Fix race in the forklift populator that might prevent cleanup
```

### DIFF
--- a/pkg/controller/populators/forklift-populator.go
+++ b/pkg/controller/populators/forklift-populator.go
@@ -136,7 +136,7 @@ func addWatchers(mgr manager.Manager, c controller.Controller, log logr.Logger, 
 			}
 
 			// Check if the owner is a PVC prime
-			if !isPVCPrimeForkliftKind(pvcPrime) && pvcPrime.DeletionTimestamp == nil {
+			if !isPVCPrimeForkliftKind(pvcPrime) || pvcPrime.DeletionTimestamp != nil {
 				return nil
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/populators/forklift-populator.go
```diff
@@ -212,6 +207,10 @@ func (r *ForkliftPopulatorReconciler) reconcile(ctx context.Context, req reconci
 		res, err = r.reconcileCleanup(pvcPrime)
 	}
 
+	if pvcPrime.DeletionTimestamp != nil {
+		res, err = r.deletePopulatorPod(fmt.Sprintf("%s-%s", populatorPodPrefix, pvc.UID), pvc, pvcPrime)
```
mhenriks, line 211: in what cases would pvc prime be deleted before the pod?

bennyz: `res, err = r.reconcileCleanup(pvcPrime)` in L#207 marks it for deletion

mhenriks: Why not have `reconcileCleanup` delete the pod as well then?

bennyz:
> Is this change covered by a unit test? I ask because coverage for forklift-populator.go is very low at 24.9%

I'll add

mhenriks:
/lgtm
/approve

@bennyz nice bump in test coverage, more would be welcome in subsequent prs!

PR #3285, opened 2024-05-22
Add watch for DataImportCron-labeled PVCs deletion

arnongilboa:
When the `DataImportCron` last import `DV` is manually deleted, the controller reconciles, but due to k8s default background cascading deletion, the `PVC` may still temporarily exist, so the controller will not re-create the `DV` even after the `PVC` is deleted, unless it reconciles due to other watched CR like `DataSource`. In the scenario of CNV-39688, since we move from `PVC` source format to snapshot, the `DataSource` won’t update until a snapshot is created, which will never happen. To solve it we add a watch for deletion of DataImportCron-labeled `PVCs`. The change was tested locally, but since it required two storage classes it’s currently out of scope for the existing CI lanes, so it requires tier-2 test.

Fixes # https://issues.redhat.com/browse/CNV-39688

**Release note**:
```release-note
BugFix: On deletion of the last import DV of DataImportCron with Snapshot source format, re-create an import DV
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -1027,6 +1027,17 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	if err := c.Watch(source.Kind(mgr.GetCache(), &corev1.PersistentVolumeClaim{}),
+		handler.EnqueueRequestsFromMapFunc(mapSourceObjectToCron),
+		predicate.Funcs{
+			CreateFunc: func(event.CreateEvent) bool { return false },
+			UpdateFunc: func(event.UpdateEvent) bool { return false },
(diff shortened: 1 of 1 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -1077,6 +1077,17 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	if err := c.Watch(source.Kind(mgr.GetCache(), &corev1.PersistentVolumeClaim{}),
```
akalenyu, line 1080: May want to place this above the snapshot watch, since it has a https://github.com/kubevirt/containerized-data-importer/blob/4e7e6833fa7bbf9566c4abd2996c3a231b2a034f/pkg/controller/dataimportcron-controller.go#L1062-L1063

arnongilboa: Good catch, fixed.

ShellyKa13 (approved):
/lgtm

PR #3291, opened 2024-05-27
[WIP] Improve test resilience by wrapping updates around eventually statements

alromeros:

This Pull Request aims to improve test resilience by wrapping PVC, PV and DataVolume updates inside `eventually` statements.

Fixes # https://issues.redhat.com/browse/CNV-40014

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -381,8 +381,10 @@ var _ = Describe("all clone tests", func() {
 								sourcePV.SetLabels(make(map[string]string))
 							}
 							sourcePV.GetLabels()["source-pv"] = "yes"
-							sourcePV, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), sourcePV, metav1.UpdateOptions{})
-							Expect(err).ToNot(HaveOccurred())
+							Eventually(func() bool {
+								sourcePV, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), sourcePV, metav1.UpdateOptions{})
+								return err == nil
+							}, timeout, pollingInterval).Should(BeTrue())
 						}
 					} else if targetPV == nil {
 						pvNode := pv.Spec.NodeAffinity.Required.NodeSelectorTerms[0].MatchExpressions[0].Values[0]
@@ -394,8 +396,11 @@ var _ = Describe("all clone tests", func() {
 								targetPV.SetLabels(make(map[string]string))
 							}
 							targetPV.GetLabels()["target-pv"] = "yes"
-							targetPV, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), targetPV, metav1.UpdateOptions{})
-							Expect(err).ToNot(HaveOccurred())
+
+							Eventually(func() bool {
+								targetPV, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), targetPV, metav1.UpdateOptions{})
+								return err == nil
+							}, timeout, pollingInterval).Should(BeTrue())
 							break
 						}
 					}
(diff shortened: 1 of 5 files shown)


tests/static-volume_test.go
```diff
@@ -89,7 +89,11 @@ var _ = Describe("checkStaticVolume tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 			pv.Spec.PersistentVolumeReclaimPolicy = corev1.PersistentVolumeReclaimRetain
 			_, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
-			Expect(err).ToNot(HaveOccurred())
+			// We shouldn't make the test fail if there's a conflict with the update request.
+			// These errors are usually transient and should be fixed in subsequent retries.
```
akalenyu, line 93: I don't think we will retry over this failing with `errors.IsConflict`, unless I'm missing something

tests/static-volume_test.go
```diff
@@ -107,7 +111,11 @@ var _ = Describe("checkStaticVolume tests", func() {
 				pv.Spec.ClaimRef.ResourceVersion = ""
 				pv.Spec.ClaimRef.UID = ""
 				_, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
-				Expect(err).ToNot(HaveOccurred())
+				// We shouldn't make the test fail if there's a conflict with the update request.
+				// These errors are usually transient and should be fixed in subsequent retries.
+				if !errors.IsConflict(err) {
+					Expect(err).ToNot(HaveOccurred())
+				}
```
akalenyu, line 118: You could do something like this which will just retry transparently https://github.com/kubevirt/containerized-data-importer/blob/35f9f60ac65309c546a217f5bf9c0992ce1100d8/tests/dataimportcron_test.go#L240-L242 https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

tests/static-volume_test.go
alromeros: If this fails with `IsConflict` it'll then retry the update in the `Eventually` below, but won't have the updated reclaim policy. I'll fix it.

tests/cloner_test.go
```diff
@@ -1374,8 +1379,11 @@ var _ = Describe("all clone tests", func() {
 				sourcePvc, err = f.K8sClient.CoreV1().PersistentVolumeClaims(sourcePvc.Namespace).Get(context.TODO(), sourcePvc.Name, metav1.GetOptions{})
 				Expect(err).ToNot(HaveOccurred())
 				sourcePvc.Annotations[dvc.AnnSourceCapacity] = "400Mi"
-				_, err = f.K8sClient.CoreV1().PersistentVolumeClaims(sourcePvc.Namespace).Update(context.TODO(), sourcePvc, metav1.UpdateOptions{})
-				Expect(err).ToNot(HaveOccurred())
+
+				Eventually(func(g Gomega) {
+					_, err = f.K8sClient.CoreV1().PersistentVolumeClaims(sourcePvc.Namespace).Update(context.TODO(), sourcePvc, metav1.UpdateOptions{})
+					g.Expect(err).ToNot(HaveOccurred())
+				}, timeout, pollingInterval).Should(Succeed())
```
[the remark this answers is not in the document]
akalenyu: Yeah in this case I'd say your suggestion makes more sense
`g.Expect(err).ToNot(HaveOccurred())` is usually for the cases where you don't return the same err
value from the eventually func

tests/cloner_test.go
```diff
@@ -381,8 +381,12 @@ var _ = Describe("all clone tests", func() {
 								sourcePV.SetLabels(make(map[string]string))
 							}
 							sourcePV.GetLabels()["source-pv"] = "yes"
-							sourcePV, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), sourcePV, metav1.UpdateOptions{})
-							Expect(err).ToNot(HaveOccurred())
+							Eventually(func() error {
+								// We shouldn't make the test fail if there's a conflict with the update request.
+								// These errors are usually transient and should be fixed in subsequent retries.
+								sourcePV, err = f.K8sClient.CoreV1().PersistentVolumes().Update(context.TODO(), sourcePV, metav1.UpdateOptions{})
```
akalenyu, line 387: Should we refresh the sourcePV via a GET call? I'm thinking that if this is a conflict, it'll never resolve unless we get the updated resource?

alromeros: Oops good catch

akalenyu: Unless I am missing something, this iteration does not refresh the PV object from the API server

tests/cloner_test.go
```diff
@@ -3308,15 +3323,26 @@ func EnableGcAndAnnotateLegacyDv(f *framework.Framework, dvName, dvNamespace str
...
+
 	VerifyNoGC(f, dvName, dvNamespace)
 
 	By("Add true DeleteAfterCompletion annotation to DV")
 	controller.AddAnnotation(dv, controller.AnnDeleteAfterCompletion, "true")
-	_, err = f.CdiClient.CdiV1beta1().DataVolumes(dvNamespace).Update(context.TODO(), dv, metav1.UpdateOptions{})
-	Expect(err).ToNot(HaveOccurred())
+
+	Eventually(func() error {
+		dv, err = f.CdiClient.CdiV1beta1().DataVolumes(dvNamespace).Update(context.TODO(), dv, metav1.UpdateOptions{})
```
akalenyu, line 3342: we don't refresh the datavolume here, so it'll probably keep failing if a conflict occurs
does this test fail often with a conflict err?

alromeros: Oops my fault again, It doesn't fail but yeah, I should make sure to refresh it.

akalenyu (commented):
/approve

PR #3293, opened 2024-05-28
tests: fix regex in import progress test

akalenyu:
This test currently passes since the regex will always catch 100%, hiding a bug we have in progress reporting.

Fixes #

**Release note**:

```release-note
NONE
```


tests/datavolume_test.go
```diff
@@ -2836,13 +2835,12 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			Eventually(func(g Gomega) string {
 				dv, err := f.CdiClient.CdiV1beta1().DataVolumes(f.Namespace.Name).Get(context.TODO(), dataVolume.Name, metav1.GetOptions{})
-				Expect(err).ToNot(HaveOccurred())
-				progress := dv.Status.Progress
-				return progressRegExp.MatchString(string(progress))
-			}, timeout, pollingInterval).Should(BeTrue())
+				g.Expect(err).ToNot(HaveOccurred())
+				fmt.Fprintf(GinkgoWriter, "INFO: progress: %s\n", dv.Status.Progress)
+				return string(dv.Status.Progress)
+			}, timeout, pollingInterval).Should(MatchRegexp(`^([1-9]{1,2})(\.\d+)?%$`), "DataVolume is not reporting import progress")
```
arnongilboa, line 2843: How can you be sure you'll catch it before it gets to 100%?

akalenyu: This is a rate limited endpoint capped at 100k, so, once we fix our bugs we are guaranteed to catch progress here

awels: Why not fix the test? Or is it an actual bug in the code?

akalenyu: It's a bug in the code,
Now, that we do scratch basically all across the board, the progress is only reported for the convert from scratch to target (which in many cases is very quick). This effectively means we report no progress.

We used to also report progress for raw downloads but not anymore since https://github.com/kubevirt/containerized-data-importer/pull/3219/files#diff-1c34533700a33099790249b85a35cf4aaef045bd7c649e100e55750cac1f6c83L136-L138

awels: So essentially we need to add the progress reader to when we are copying stuff to the scratch space? Should be fairly simple to do.

akalenyu: Hmm would it be correct progress reporting though? Wouldn't the 100% progress put the DV to succeeded before it really is?

awels: That is the question right. there are 2 parts to the progress. 1. The copy over the network to scratch, and 2. The copy/conversion from scratch to the target. Both could potentially take a period of time. And how do you represent that? In many cases the bulk of the time is the transfer over the network. But in some cases the copy from scratch to target can take some time as well.
Maybe split it 80/20 or something like that? 0-80% for network transfer, and 81-100% for copy? I don't know it is a difficult one to estimate.

akalenyu: Yeah I think this requires more thought..
/hold

but in general yes

> Should be fairly simple to do.

this really was simple to get the progress for the network transfer,
and I pushed a change to do that meanwhile so we can see how it goes through CI

jpeimer:
Oh, we actually don't have 1.60 yet

PR #3295, opened 2024-05-29
forklift: rename API group

bennyz:

Since forklift 2.7 has to support OCP 4.15 we'll have 2 competing controllers, this PR changes the API group for CDI's so forklift can distinguish between them.

Fixes #

**Release note**:

```release-note
Rename API group for forklift controllers to forklift.cdi.kubevirt.io

```

### DIFF
--- a/doc/cdi-populators.md
+++ b/doc/cdi-populators.md
@@ -154,12 +154,12 @@ spec:
   dataSourceRef:
     name: ovirt-pop
     kind: OvirtVolumePopulator
-    apiGroup: forklift.cdi.konveyor.io
+    apiGroup: forklift.cdi.kubevirt.io
   resources:
     requests:
       storage: 13G
 ---
-apiVersion: forklift.cdi.konveyor.io/v1beta1
+apiVersion: forklift.cdi.kubevirt.io/v1beta1
 kind: OvirtVolumePopulator
 metadata:
   name: ovirt-pop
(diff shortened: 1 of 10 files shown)


doc/cdi-populators.md
```diff
@@ -154,12 +154,12 @@ spec:
   dataSourceRef:
     name: ovirt-pop
     kind: OvirtVolumePopulator
-    apiGroup: forklift.konveyor.io
+    apiGroup: forklift.cdi.konveyor.io
```
mhenriks, line 157: wouldn't `cdi.forklift.konveyor.io` be more appropriate?

aglitke: agree.

aglitke (commented):
Should we add a quick test that CDI ignores a CR that uses the old API group?

doc/cdi-populators.md
bennyz: done

bennyz:
> Should we add a quick test that CDI ignores a CR that uses the old API group?

added

bennyz:
@mhenriks @aglitke I had to revert this back to forklift.cdi.konveyor.io, @akalenyu and I found this breaks code generation, because prefixing with `cdi.` make it generate an already existing alias for the package which would be `cdiv1beta1`, so we can't use it as a prefix.

mhenriks:
Why not just add this into `cdi.kubevirt.io`?  Or `forklift.cdi.kubevirt.io` if we want to keep separate?

cc @aglitke @akalenyu @awels

akalenyu:
> Why not just add this into `cdi.kubevirt.io`? Or `forklift.cdi.kubevirt.io` if we want to keep separate?
> 
> cc @aglitke @akalenyu @awels

Yeah dropping the conveyor part works for me, but I do think it deserves a package so slight preference is `forklift.cdi.kubevirt.io`

akalenyu:
@bennyz could you also change release note?

bennyz:
> @bennyz could you also change release note?

done

PR #3298, opened 2024-05-30
Enable intrange linter

EduardGomezEscandell:
This linter ensures we use the new range-over-int construct

OLD:
```go
    for i:=0; i<N; i++ {}
```

NEW:
```go
    for i := range N {}
```
Often it can be simplified further:
```go
    for range N {}
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,6 +9,7 @@ linters:
   - gofmt
   - goimports
   - govet
+  - intrange
   - misspell
   - nakedret
   - revive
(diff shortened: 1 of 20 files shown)


akalenyu:
Pretty aggressive to start forcing the new style, no?

EduardGomezEscandell:
Maybe, I guess this is just a proposal.

Feel free to reject it, this took 10 minutes so it's not like there's a big sunk cost.

EduardGomezEscandell:
Closing this because I don't want to deal with all the failures, and the interest may not be there.

PR #3301, opened 2024-06-02
[release-v1.58] Add watch for DataImportCron-labeled PVCs deletion

arnongilboa:
Manual backport of #3285

When the DataImportCron last import DV is manually deleted, the controller reconciles, but due to k8s default background cascading deletion, the PVC may still temporarily exist, so the controller will not re-create the DV even after the PVC is deleted, unless it reconciles due to other watched CR like DataSource. In the scenario of CNV-39688, since we move from pvc source format to snapshot, the DataSource won’t update until a snapshot is created, which will never happen. To solve it we add a watch for deletion of DataImportCron-labeled PVCs. The change was tested locally, but since it required two storage classes it’s currently out of scope for the existing CI lanes, so it requires tier-2 test.

Fixes # https://issues.redhat.com/browse/CNV-42129

**Release note**:
```release-note
BugFix: On deletion of the last import DV of DataImportCron with Snapshot source format, re-create an import DV
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -1041,6 +1041,17 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 		return err
 	}
 
+	if err := c.Watch(&source.Kind{Type: &corev1.PersistentVolumeClaim{}},
+		handler.EnqueueRequestsFromMapFunc(mapSourceObjectToCron),
+		predicate.Funcs{
+			CreateFunc: func(event.CreateEvent) bool { return false },
+			UpdateFunc: func(event.UpdateEvent) bool { return false },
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
If you grab https://github.com/kubevirt/containerized-data-importer/pull/3123/commits/7360fb1027c692569296da09c66b387954ea5ca4 I think you'll get a clean backport
(It's low priority for a backport IMO but still has some value)

arnongilboa:
> If you grab [7360fb1](https://github.com/kubevirt/containerized-data-importer/commit/7360fb1027c692569296da09c66b387954ea5ca4) I think you'll get a clean backport (It's low priority for a backport IMO but still has some value)

It won't. In main we use source.Kind() helper instead of source.Kind{}, so the patch will fail.

PR #3310, opened 2024-06-09
Suppress CDIDefaultStorageClassDegraded on SNO

arnongilboa:
On single-node OpenShift, even if none of the default/virt default storage classes supports `ReadWriteMany` (but supports smart clone), we will not fire the `CDIDefaultStorageClassDegraded` alert. We added `degraded` label to `kubevirt_cdi_storageprofile_info` to simplify the alert expression.

jira-ticket: https://issues.redhat.com/browse/CNV-40665

**Release note**:
```release-note
Suppress CDIDefaultStorageClassDegraded alert on SNO
```

### DIFF
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/go-logr/logr"
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
+	ocpconfigv1 "github.com/openshift/api/config/v1"
 	"github.com/prometheus/client_golang/prometheus"
 
 	v1 "k8s.io/api/core/v1"
@@ -48,6 +49,7 @@ const (
 	counterLabelVirtDefault      = "virtdefault"
 	counterLabelRWX              = "rwx"
 	counterLabelSmartClone       = "smartclone"
+	counterLabelDegraded         = "degraded"
 )
 
 // StorageProfileReconciler members
@@ -296,13 +298,26 @@ func (r *StorageProfileReconciler) computeMetrics(profile *cdiv1.StorageProfile,
 		return err
 	}
 
+	isSNO := false
+	clusterInfra := &ocpconfigv1.Infrastructure{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, clusterInfra); err != nil {
+		if !meta.IsNoMatchError(err) && !k8serrors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		isSNO = clusterInfra.Status.ControlPlaneTopology == ocpconfigv1.SingleReplicaTopologyMode &&
+			clusterInfra.Status.InfrastructureTopology == ocpconfigv1.SingleReplicaTopologyMode
+	}
+
+	isDegraded := (!isSNO && !isRWX) || !isSmartClone
+
 	// Setting the labeled Gauge to 1 will not delete older metric, so we need to explicitly delete them
 	scLabels := prometheus.Labels{counterLabelStorageClass: storageClass, counterLabelProvisioner: provisioner}
 	metricsDeleted := metrics.DeleteStorageProfileStatus(scLabels)
-	scLabels = createLabels(storageClass, provisioner, isComplete, isDefault, isVirtDefault, isRWX, isSmartClone)
+	scLabels = createLabels(storageClass, provisioner, isComplete, isDefault, isVirtDefault, isRWX, isSmartClone, isDegraded)
 	metrics.SetStorageProfileStatus(scLabels, 1)
-	r.log.Info(fmt.Sprintf("Set metric:%s complete:%t default:%t vdefault:%t rwx:%t smartclone:%t (deleted %d)",
-		storageClass, isComplete, isDefault, isVirtDefault, isRWX, isSmartClone, metricsDeleted))
+	r.log.Info(fmt.Sprintf("Set metric:%s complete:%t default:%t vdefault:%t rwx:%t smartclone:%t degraded:%t (deleted %d)",
+		storageClass, isComplete, isDefault, isVirtDefault, isRWX, isSmartClone, isDegraded, metricsDeleted))
 
 	return nil
 }
@@ -335,7 +350,7 @@ func (r *StorageProfileReconciler) hasSmartClone(sp *cdiv1.StorageProfile) (bool
 	return false, nil
 }
 
-func createLabels(storageClass, provisioner string, isComplete, isDefault, isVirtDefault, isRWX, isSmartClone bool) prometheus.Labels {
+func createLabels(storageClass, provisioner string, isComplete, isDefault, isVirtDefault, isRWX, isSmartClone, isDegraded bool) prometheus.Labels {
 	return prometheus.Labels{
 		counterLabelStorageClass: storageClass,
 		counterLabelProvisioner:  provisioner,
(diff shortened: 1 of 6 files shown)


pkg/controller/storageprofile-controller.go
```diff
@@ -296,13 +298,26 @@ func (r *StorageProfileReconciler) computeMetrics(profile *cdiv1.StorageProfile,
...
+	isSNO := false
+	clusterInfra := &ocpconfigv1.Infrastructure{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, clusterInfra); err != nil {
+		if !meta.IsNoMatchError(err) && !k8serrors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		isSNO = clusterInfra.Status.ControlPlaneTopology == ocpconfigv1.SingleReplicaTopologyMode &&
+			clusterInfra.Status.InfrastructureTopology == ocpconfigv1.SingleReplicaTopologyMode
+	}
```
akalenyu, line 310: So with LVMS this check was quite tricky to rely on, since you could be doing "SNO" with multiple workers but your images (and thus VMs) would end up on same node, so for our purposes it wasn't "SNO".

Do we have the same concern here? I want to make sure we don't take away the alert from cases where it's actually helpful

akalenyu: cc @awels wdyt

arnongilboa: Discussing it with @aglitke , we currently decide to suppress only in the case of a single node, so afaiu in the mentioned LVMS case we will still alert if no RWX.

akalenyu: The problem is (I think) that this API does not actually guarantee 1 single node.
So someone doing 3-node SNO setup, trying to properly set up live migration, will not receive the alert
We may or may not be okay with that - but something to consider

arnongilboa: The [doc](https://docs.openshift.com/container-platform/4.15/operators/operator_sdk/osdk-ha-sno.html) says that's the way to detect a single-node, and afaik the other alternative for detecting SNO requires adding cluster rbac for node “list” and “watch”, which we currently prefer not to add.

akalenyu: Yeah as I thought they use the word "single" but it doesn't actually guarantee a single node:

- When the control plane replica count is less than 3, the controlPlaneTopology status is set to SingleReplica. Otherwise, it is set to HighlyAvailable.

- When the worker replica count is 0, the control plane nodes are also configured as workers. Therefore, the infrastructureTopology status will be the same as the controlPlaneTopology status.

akalenyu: Again I am not strictly against just not firing the alert for anything that's not HA, but have to raise this point

arnongilboa: btw why is it so *bad* to add cluster rbac for node “list” and “watch”? Sure CDI is a completely different animal, but HCO has this rbac.

awels: When adding RBAC we always have to consider why are we adding RBAC. And honestly some alert doesn't seem like the right thing to add RBAC for.

arnongilboa: @awels understood. And what about @akalenyu argument here?

awels: I think this was @mhenriks point, that it is not easy to get right to figure out if we are truly in a single node environment or not. There are all kinds of edge cases where we won't alert with this and probably should. I do still think it is an improvement over always alerting in the wrong case.

awels: That is another reason why adding the nodes is not a good idea. You could have multiple nodes, and some of them are not marked as worker nodes, and you now have only one node that can run workloads. Getting that stuff right is very tricky. I think this is 'good enough'. People with those edge cases should know what they are doing.

awels (approved):

pkg/controller/storageprofile-controller_test.go
```diff
@@ -484,12 +487,60 @@ var _ = Describe("Storage profile controller reconcile loop", func() {
...
-		labels := createLabels(storageClassName, provisioner, false, true, false, false, false)
+		labels := createLabels(storageClassName, provisioner, isComplete, true, false, false, false, true)
 		Expect(int(metrics.GetStorageProfileStatus(labels))).To(Equal(count))
 	},
-		Entry("Noobaa (not supported)", storagecapabilities.ProvisionerNoobaa, 0),
-		Entry("Unknown provisioner", "unknown-provisioner", 1),
+		Entry("Noobaa (not supported)", storagecapabilities.ProvisionerNoobaa, false, 0),
+		Entry("Noobaa (not supported)", storagecapabilities.ProvisionerNoobaa, true, 1),
+		Entry("Unknown provisioner", "unknown-provisioner", false, 1),
+		Entry("Unknown provisioner", "unknown-provisioner", true, 0),
```
akalenyu, line 496: Same entry name will be confusing when one of these fails

akalenyu:
/approve
/hold
unhold once you're happy with https://github.com/kubevirt/containerized-data-importer/pull/3310#discussion_r1641869644

pkg/controller/storageprofile-controller_test.go
arnongilboa: Sure, these entries are not needed and can be merged. Fixed.

akalenyu:
> @arnongilboa: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-containerized-data-importer-e2e-nfs	[45a28d6](https://github.com/kubevirt/containerized-data-importer/commit/45a28d6dc29ac39b90c329c8b23a5ce9c1f87e68)	[link](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3310/pull-containerized-data-importer-e2e-nfs/1802358998075183104)	true	`/test pull-containerized-data-importer-e2e-nfs`

@arnongilboa I think this is a real bug, the "Error" condition reason will not appear anymore after https://github.com/kubevirt/containerized-data-importer/pull/3194

arnongilboa:
> > @arnongilboa: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > Test name	Commit	Details	Required	Rerun command
> > pull-containerized-data-importer-e2e-nfs	[45a28d6](https://github.com/kubevirt/containerized-data-importer/commit/45a28d6dc29ac39b90c329c8b23a5ce9c1f87e68)	[link](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3310/pull-containerized-data-importer-e2e-nfs/1802358998075183104)	true	`/test pull-containerized-data-importer-e2e-nfs`
> 
> @arnongilboa I think this is a real bug, the "Error" condition reason will not appear anymore after #3194

Shouldn't it fail on all PRs? it has nothing to do with this one and should be fixed separately.

akalenyu:
> > > @arnongilboa: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > > Test name	Commit	Details	Required	Rerun command
> > > pull-containerized-data-importer-e2e-nfs	[45a28d6](https://github.com/kubevirt/containerized-data-importer/commit/45a28d6dc29ac39b90c329c8b23a5ce9c1f87e68)	[link](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3310/pull-containerized-data-importer-e2e-nfs/1802358998075183104)	true	`/test pull-containerized-data-importer-e2e-nfs`
> > 
> > 
> > @arnongilboa I think this is a real bug, the "Error" condition reason will not appear anymore after #3194
> 
> Shouldn't it fail on all PRs? it has nothing to do with this one and should be fixed separately.

Maybe there's something wrong with the test? It's not related to this PR and should have started failing since that change

PR #3314, opened 2024-06-12
Expose Upload Proxy certificate in CDI Config status

EduardGomezEscandell:
As a VM owner I want to upload a VM disk to my cluster without needing to use an unsecured TLS connection. I want to import the upload proxy cert into my client.

https://issues.redhat.com/browse/CNV-21214

None of the config controller code is covered by tests, I tried to add test for `watchUploadProxyCA` but it was too much boilerplate and I didn't know what I was doing.

The reconciler is well tested (at least the happy paths) so I could add my own test easily.

**Release note**:

```release-note
Expose the upload-proxy's certificate in the CDI config status
```

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -17,6 +17,7 @@ import (
 	"go.uber.org/zap/zapcore"
 
 	batchv1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -407,6 +408,9 @@ func getCacheOptions(apiClient client.Client, cdiNamespace string) cache.Options
 				Field: namespaceSelector,
 			},
 			&batchv1.Job{}: {
+				Field: namespaceSelector,
+			},
+			&v1.ConfigMap{}: {
 				Field: namespaceSelector,
 			},
 		},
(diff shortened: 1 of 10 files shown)


EduardGomezEscandell:
Seems like fossa is complaining about docker's licence? I did not add or update this dependency :man_shrugging:

EduardGomezEscandell:
Anyways this is ready for review @awels

EduardGomezEscandell:
> Looks pretty good to me, I am just missing the regular K8S Ingress. We do look up the Ingress certificate in VMExport.

We cover this implicitly I believe. The UploadProxyURL that we use is set right before in `reconcileUploadProxyURL`. It is set to the route if present, and otherwise the ingress.

awels:
Right but the Ingress CA cert is in a different config map IIRC. I would have to look at the export code to see what I did there. I do remember the logic being quite different from what I did for Routes.

awels:
I am talking about this bit of [code](https://github.com/kubevirt/kubevirt/blob/main/pkg/storage/export/export/links.go#L183-L203) in vmexport. The code assuming the secret exists in the namespace kubevirt is installed in. In this case it would be the namespace CDI is installed in.

awels (approved):
/approve
Thanks for all the work!

cmd/cdi-controller/controller.go
```diff
@@ -409,6 +410,9 @@ func getCacheOptions(apiClient client.Client, cdiNamespace string) cache.Options
 			&batchv1.Job{}: {
 				Field: namespaceSelector,
 			},
+			&v1.ConfigMap{}: {
+				Field: namespaceSelector,
+			},
```
akalenyu, line 415: I think the same treatment may be needed for Secrets, or, could use uncached client for both instead and drop this, since we're only ever going to GET on those (which should be cheap).
@mhenriks WDYT?

akalenyu: ping @awels @mhenriks

PR #3320, opened 2024-06-13
forklift: add missing rbac

bennyz:

Add missing RBAC for ovirt and openstack volume populator CRDs

Fixes #

**Release note**:

```release-note
Add missing RBAC for ovirt and openstack volume populator CRDs

```

### DIFF
--- a/pkg/operator/resources/cluster/controller.go
+++ b/pkg/operator/resources/cluster/controller.go
@@ -275,9 +275,7 @@ func getControllerClusterPolicyRules() []rbacv1.PolicyRule {
 				"openstackvolumepopulators",
 			},
 			Verbs: []string{
-				"get",
-				"list",
-				"watch",
+				"*",
 			},
 		},
 	}
--- a/pkg/operator/resources/cluster/controller.go
+++ b/pkg/operator/resources/cluster/controller.go
@@ -165,9 +165,11 @@ func getControllerClusterPolicyRules() []rbacv1.PolicyRule {
 				"get",
 			},
 		},
+		// "*" permissions are usually bad, but it makes sense for CDI to have them as it controls everything related
 		{
 			APIGroups: []string{
 				"cdi.kubevirt.io",
+				"forklift.cdi.kubevirt.io",
 			},
 			Resources: []string{
 				"*",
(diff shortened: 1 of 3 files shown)


pkg/operator/resources/cluster/controller.go
```diff
@@ -274,9 +274,7 @@ func getControllerClusterPolicyRules() []rbacv1.PolicyRule {
 				"openstackvolumepopulators",
 			},
 			Verbs: []string{
-				"get",
-				"list",
-				"watch",
+				"*",
```
stesrn, line 277: @bennyz hi. I remember having global perms was not recommended. Refer - https://issues.redhat.com/browse/CNV-22907 and https://issues.redhat.com/browse/CNV-32812

bennyz: I think I have to do this to because of how `pkg/operator/resources/cluster/rbac.go` is set

There is also:
```
		{
			APIGroups: []string{
				"cdi.kubevirt.io",
			},
			Resources: []string{
				"*",
			},
			Verbs: []string{
				"*",
			},
},
```
in L#169

But I'm open to suggestions

akalenyu:
/test all

> I remember having global perms was not recommended

Yup, there should be a unit test for this.

pkg/operator/resources/cluster/controller.go
akalenyu: Would you mind unifying those two with a comment that explains the motivation?
Some thing along the lines off, "*" permissions are bad, but logically it's clear that CDI can do anything on CDI resources

pkg/operator/resources/cluster/rbac.go
```diff
@@ -137,6 +168,20 @@ func createConfigReaderClusterRole(name string) *rbacv1.ClusterRole {
...
+				"ovirtvolumepopulators",
+				"openstackvolumepopulators",
+			},
+			Verbs: []string{
+				"get",
+				"list",
+				"watch",
+			},
+		},
 	}
```
akalenyu, line 185: I don't think you need this?

aglitke:
@awels This PR seems stalled.  Do we still want to have it merged?

awels:
We do, but we need to address the `*` permissions. We have a unit test that disallows this. We should be explicit in the permissions we want to give otherwise security will complain. Alex already pointed this out.

awels:
/lifecycle frozen
We do want the `*` issue fixed.

PR #3321, opened 2024-06-13
Update quota.md

cccsss01:
The override Defaults didn't work for me, I updated them to what did work for me however i'm not sure how to check them, as the status of the CDIS object still shows the old settings.  Adding information that states how to update it would be beneficial.

I also added docs that modifies the CDI directly.

### DIFF
--- a/doc/quota.md
+++ b/doc/quota.md
@@ -25,7 +25,7 @@ spec:
 
 ```
 
-Once the CDIConfig object is updated, the status section of the object will reflect that values that will be used to pass to the pods. [limits and requests](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#motivation-for-default-memory-limits-and-requests) are explained in the kubernetes documentation.
+Once the CDIConfig object is updated, the status section of the object will reflect that values that will be used to pass to the pods. [limits and requests](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#motivation-for-default-memory-limits-and-requests) are explained in the kubernetes documentation.  The CDIConfig object is updated by editing the `spec.config` section of the CDI resource.
 
 ## Modify CDI directly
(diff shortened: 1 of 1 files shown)


doc/quota.md
```diff
@@ -14,14 +14,44 @@ metadata:
...
+      limits:
+        cpu: "4"
+        memory: "1Gi"
+      requests:
+        cpu: "1"
+        memory: "250Mi"
+
 ```
+
 Once the CDIConfig object is updated, the status section of the object will reflect that values that will be used to pass to the pods. [limits and requests](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#motivation-for-default-memory-limits-and-requests) are explained in the kubernetes documentation.
```
mhenriks, line 28: This is an old doc, thanks for helping to make it current.  I think it would be worth mentioning that the CDIConfig singleton resource is read only and it is updated via `spec.config` section of the CDI custom resource.

Once the CDI resource is updated, CDI config `status.defaultPodResourceRequirements` should reflect the changes

cccsss01: I'm not saying I disagree, though I'm not sure I'd word it exactly like that and I'd prefer not to guess.  If you want to add the changes to this PR, put it in a comment and I'll add it, if it's a blocker, not a blocker I'm indifferent.

mhenriks, line 28:
```suggestion
Once the CDIConfig object is updated, the status section of the object will reflect that values that will be used to pass to the pods. [limits and requests](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#motivation-for-default-memory-limits-and-requests) are explained in the kubernetes documentation.  The CDIConfig object is updated by editing the `spec.config` section of the CDI resource.
```

cccsss01:
I'm thinking this is a problem with the unit tests?

PR #3331, opened 2024-06-27
Remove datavolume clone source validation

ShellyKa13:
We want our product to be more eventual consistent. In order to achieve that we want to remove webhooks that checks for stuff the can be racy. Some checks already existed in the reconcile loops since we need to check the same things for when the source doesnt exist yet. In the snapshot clone controller the validations were moved to be only when doing legacy clone although the should be checked always.

Fixes # Jira-ticket: https://issues.redhat.com/browse/CNV-41610

**Release note**:

```release-note
Remove datavolume clone source validation from webhook
```


pkg/controller/common/util.go
```diff
@@ -1010,39 +1010,25 @@ func ValidateClone(sourcePVC *corev1.PersistentVolumeClaim, spec *cdiv1.DataVolu
 	return nil
 }
 
-// ValidateSnapshotClone compares a snapshot clone spec against its source snapshot to validate its creation
```
mhenriks, line 1013: Based on the commit description, we are adding some functionality (additional validation with populators) that was missing and not directly related to the goal of this PR.  Can you speak on that a little more and perhaps give an example of previous undesirable behavior?

ShellyKa13: Well this change is related to the goal of the PR since with the webhook we do check this validations for all kinds of clone (populator/legacy). Once webhook removed I noticed that the same validations are only done in the legacy code part and not in the populators code part. Now since we allowed to create a clone without a source, this means that the validation should have also happened in the reconcile since in the webhook we skipped source checks when the source didnt exist yet. which means for clones without source once the source created you could have had a source that the restore size was too small compared to the target requested size, but you will not present any error and you would continue with the clone anyways.
So this change fixes that, but also addresses normal flow where instead of rejecting in the webhook any clone, we should do it in the reconcile (populator and legacy).

ShellyKa13: I added to the commit some better explanations too

mhenriks: Clone populator has to do it's own validation.  If source does not exist, it should wait.  If there is any error, such as target is smaller than source, it _should_ set the following annotations on the PVC:

```yaml
cdi.kubevirt.io/clonePhase: Error
cdi.kubevirt.io/cloneError: "error reason"
```

Maybe it would be better for datavolume controller to not do validataion in populator case and instead check these annotations on the PVC to report errors back to the user?

I totally get that some checks may be missing in the populator right now though.

ShellyKa13: I definitely plane on having a followup PR that checks these validations in the populators only flow.
Since I want to do it in a follow up PR, I want to merge this PR with the validations on the dv with populators path since I'm removing them from the webhook and want to cover that part until then.

mhenriks: Okay

Issue #3335, opened 2024-07-04
Include functional test focus capabilities


Davo911:
**Is your feature request related to a problem? Please describe**:

The `CDI_E2E_FOCUS` enviromental variable, to run specific functional tests, is implemented in the `test.sh` of the CI/CD, but is not working when `make test-functional`.
___
**Describe the solution you'd like**:

When running `make test-functional` with `CDI_E2E_FOCUS` set, I'd like to only run a subset of functional tests.
Furthermore I'd propose to also make use of the `--focus-file` flag of ginkgo, to be able to run a single suite.

PR #3337, opened 2024-07-06
Move and extend functional test focus flag

Davo911:

Moves the ginkgo focus capabilities to `run-functional-tests.sh`, analogue to the kubevirt repo, so that `CDI_E2E_FOCUS` can be used with `make test-functional`. 
Also with this a go-file path can be passed to `CDI_E2E_FOCUS`  to focus a specific suite using the `--focus-file` ginkgo flag.

---

**Which issue(s) this PR fixes** **:
Fixes Include functional test focus capabilities #3335 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -116,10 +116,6 @@ kubectl version
 
 ginko_params="--test-args=--ginkgo.no-color --ginkgo.junit-report=${ARTIFACTS_PATH}/junit.functest.xml"
 
-if [[ -n "$CDI_E2E_FOCUS" ]]; then
-  ginko_params="${ginko_params} --ginkgo.focus=${CDI_E2E_FOCUS}"
-fi
-
 if [[ -n "$CDI_E2E_SKIP" ]]; then
   ginko_params="${ginko_params} --ginkgo.skip=${CDI_E2E_SKIP}"
 fi
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -116,10 +116,6 @@ kubectl version
 
 ginko_params="--test-args=--ginkgo.no-color --ginkgo.junit-report=${ARTIFACTS_PATH}/junit.functest.xml"
 
-if [[ -n "$CDI_E2E_SKIP" ]]; then
-  ginko_params="${ginko_params} --ginkgo.skip=${CDI_E2E_SKIP}"
-fi
-
 if [[ -n "$CDI_DV_GC" ]]; then
     kubectl patch cdi $CDI_NAMESPACE --type merge -p '{"spec": {"config": {"dataVolumeTTLSeconds": '$CDI_DV_GC'}}}'
 fi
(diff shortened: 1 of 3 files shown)


akalenyu:
/test all
thanks for the PR! is there any specific use case you were going for? are you running CDI tests in some other CI environment?

Davo911:
Hey Alex,
I'm part of the multiarch effort to enable CDI on s390x and for this, running tests locally and having better control over specific test-suites could be benificial.

As this is my first PR here in the repo, I'd also view this as kind of intro PR to get to know the place :)

/retest

automation/test.sh
```diff
@@ -116,10 +116,6 @@ kubectl version
 
 ginko_params="--test-args=--ginkgo.no-color --ginkgo.junit-report=${ARTIFACTS_PATH}/junit.functest.xml"
 
-if [[ -n "$CDI_E2E_FOCUS" ]]; then
-  ginko_params="${ginko_params} --ginkgo.focus=${CDI_E2E_FOCUS}"
-fi
-
 if [[ -n "$CDI_E2E_SKIP" ]]; then
   ginko_params="${ginko_params} --ginkgo.skip=${CDI_E2E_SKIP}"
 fi
```
akalenyu, line 121: should we maybe move this one over to `run-functional-tests.sh` as well?

akalenyu (commented):
Looks good, minor suggestion

Just to be sure I checked the lanes to see if the amount of tests ran has changed, remained exactly the same :+1:

automation/test.sh
Davo911: You're right!

Davo911: Not sure about moving CDI_DV_GC check, but that would also be outside of the PR scope here, i think
https://github.com/kubevirt/containerized-data-importer/blob/da3d7a146e5b6883da5f6c5593e5ee929639772c/automation/test.sh#L127-L129

hack/build/run-functional-tests.sh
```diff
@@ -95,11 +95,21 @@ fi
 
 (
     export TESTS_WORKDIR=${CDI_DIR}/tests
-    ginkgo_args="--trace --timeout=8h"
+    ginkgo_args="--trace --timeout=8h --v"
+
+    if [[ -n "$CDI_E2E_SKIP" ]]; then
+        ginkgo_args="${ginkgo_args} --ginkgo.skip=${CDI_E2E_SKIP}"
```
akalenyu, line 101: I think we can use --skip since we now supply it to the ginkgo cli, similarly to how you used --focus

automation/test.sh
akalenyu: may be missing something but I don't see a reason tot move this
I suggested to move the skip part since it was natural for it to be next to the focus

hack/build/run-functional-tests.sh
akalenyu: I see you resolved this btw?

PR #3339, opened 2024-07-07
Make restart count test serial

akalenyu:
This test asserts that there are no failing datavolumes in the cluster, so it can't run in parallel to negative tests, and thus has to remain serial.

Fixes #

**Release note**:

```release-note
NONE
```


mhenriks:
nice catch
/lgtm
/approve

PR #3342, opened 2024-07-10
Flaky certrotation tests

mhenriks:

Synchronization between fake client and lister could cause flakes

Test assuming creates would be seen immedietly but updates would not

Solution is to forget about validating anything library-go package may do

It's not our code anyway!

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/controller/certrotation.go
+++ b/pkg/operator/controller/certrotation.go
@@ -196,14 +196,27 @@ func (cm *certManager) ensureCertConfig(secret *corev1.Secret, certConfig cdicer
 	return secret, nil
 }
 
-func (cm *certManager) createSecret(namespace, name string) (*corev1.Secret, error) {
+func (cm *certManager) createOrGetSecret(namespace, name string) (*corev1.Secret, error) {
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
 		},
 	}
 
-	return cm.k8sClient.CoreV1().Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+	secret, err := cm.k8sClient.CoreV1().Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+	if err != nil {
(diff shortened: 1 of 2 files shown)


akalenyu (commented):
Looks good. Are we maybe interested in something like https://github.com/kubevirt/kubevirt/blob/392a8c5b1af36eff340d83d9eec123bf8438bd00/pkg/storage/export/export/export_test.go#L195 too?
I've seen this also flake with `secrets "cdi-uploadserver-client-cert" already exists`:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3336/pull-cdi-goveralls/1809334945819136000

mhenriks:
> Looks good. Are we maybe interested in something like https://github.com/kubevirt/kubevirt/blob/392a8c5b1af36eff340d83d9eec123bf8438bd00/pkg/storage/export/export/export_test.go#L195 too?

What's the benefit?  Fake informers are indirectly still being made under the covers

> I've seen this also flake with secrets "cdi-uploadserver-client-cert" already exists:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3336/pull-cdi-goveralls/1809334945819136000

yeah we can handle that

akalenyu (commented):
/approve

akalenyu:
> What's the benefit? Fake informers are indirectly still being made under the covers

I thought the fake source has some special impl. that would make this more unit-test appropriate, but I guess that's not the case. I don't recall us having to deal with this kind of problem in kubevirt unit tests?

mhenriks:
> > What's the benefit? Fake informers are indirectly still being made under the covers
> 
> I thought the fake source has some special impl. that would make this more unit-test appropriate, but I guess that's not the case. I don't recall us having to deal with this kind of problem in kubevirt unit tests?

I think it is kind of unusual (in a unit test) for an application to get a cached resource, update it, then read it from the cache AGAIN and expect to see the update, which is what is happening here.  Usually just read from the cache once

PR #3346, opened 2024-07-11
Add KubeSAN capabilities

stefanha:
KubeSAN is a CSI plugin for creating volumes on a shared SAN LUN: https://gitlab.com/kubesan/kubesan

This PR makes CDI work out of the box with the KubeSAN CSI plugin.

Fixes #

**Release note**:

```release-note
Added KubeSAN CSI plugin storage capabilities so the CDI StorageProfile is automatically populated.
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -118,6 +118,8 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	// huawei
 	"csi.huawei.com":     createAllButRWXFileCapabilities(),
 	"csi.huawei.com/nfs": createAllFSCapabilities(),
+	// KubeSAN
+	"kubesan.gitlab.io": {{rwx, block}, {rox, block}, {rwo, block}, {rwo, file}},
 }
 
 // SourceFormatsByProvisionerKey defines the advised data import cron source format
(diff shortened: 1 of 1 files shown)


albertofaria:
FWIW, LGTM.

akalenyu (commented):
looks good - any particular thoughts about the third point here?
https://github.com/kubevirt/containerized-data-importer/blob/76f9c0a4233ece912cde58d7666af866270a3d6c/doc/onboarding-storage-provisioners.md?plain=1#L13-L19 it's totally fine if that's a no for now.

akalenyu:
/test pull-cdi-unit-test
known flake

awels:
@akalenyu On the 3rd point, I don't think we know yet. I would stick with the default for now, which should be PVCs and we can determine later if snapshots is better.
/lgtm

stefanha:
@akalenyu @awels KubeSAN currently does not benefit from cloning from a snapshot. Regardless of whether the CSI plugin performs a CSI volume clone operation or a clone from a snapshot, the performance is the same.

awels:
So the idea is that creating a snapshot takes a long time, but restoring from a snapshot could be much faster. Which is why we added the ability to clone from a snapshot instead of a PVC. This is for the use case where one wants to create a lot of disks from a single source when creating a lot of VMs at once. If there is no benefit to this for KubeSAN and it sounds like right now there isn't. Then keeping the default is the correct choice. The lgtm stands.

Issue #2982, opened 2023-11-13
tar extraction fails when tarfile has relative links


dropte:
**What happened**:
When extracting a Tarball with top level relative links, extraction will fail. Chaging the tarball to contain only absolute links makes it succeed.

**What you expected to happen**:
DataVolume should be created and return success

**How to reproduce it (as minimally and precisely as possible)**:
Steps to reproduce the behavior.
Create a tarball from a directory:
```
mkdir example
touch example/example
tar -cf example.tar -C example .
cd example
tar -cf ../example_norel.tar  *
cd ..
```

Host the files via http. 
Create DataVolumes:
```
apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: import-archive-datavolume-rel
spec:
  source:
      http:
         url: "https://webhost/example.tar" 
  contentType: archive
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 100Mi
```
```
apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: import-archive-datavolume-norel
spec:
  source:
      http:
         url: "https://webhost/example_norel.tar" 
  contentType: archive
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 100Mi
```
output from importer for rel:
```
I1113 14:07:31.425317       1 importer.go:103] Starting importer
I1113 14:07:31.425360       1 importer.go:172] begin import process
I1113 14:07:31.895947       1 data-processor.go:356] Calculating available size
I1113 14:07:31.895972       1 data-processor.go:368] Checking out file system volume size.
I1113 14:07:31.895987       1 data-processor.go:376] Request image size not empty.
I1113 14:07:31.895998       1 data-processor.go:381] Target size 96112640.
I1113 14:07:31.896031       1 data-processor.go:255] New phase: TransferDataDir
I1113 14:07:31.896045       1 util.go:207] begin untar to /data...
I1113 14:07:31.896050       1 util.go:213] running untar cmd: [/usr/bin/tar --preserve-permissions --no-same-owner -xvC /data]
E1113 14:07:31.897677       1 util.go:222] exit status 2
E1113 14:07:31.897695       1 data-processor.go:251] exit status 2
unable to untar files from endpoint
kubevirt.io/containerized-data-importer/pkg/importer.(*HTTPDataSource).Transfer
[...]

PR #3347, opened 2024-07-17

arnongilboa:
```
```

```
```

jira-ticket: https://issues.redhat.com/browse/CNV-43105

**Release note**:
```release-note
```


pkg/controller/clone-controller.go
```diff
@@ -627,6 +627,7 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 					VolumeSource: corev1.VolumeSource{
 						PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
 							ClaimName: sourcePvcName,
+							ReadOnly:  true,
```
awels, line 630: Can you set this to true only for filesystem volumes? We have had provisioners that didn't like read only block volumes. Around line 712 we have a check that sets readonly on the volumeMount already.

akalenyu (commented):
Is there maybe a change we could make to our invocation of `tar` that would discard those changes? maybe something like https://github.com/kubevirt/containerized-data-importer/issues/2982#issuecomment-1810433172?

It's just that we've seen provisioners in the field before simply reject read only mounts - https://github.com/kubevirt/containerized-data-importer/pull/2944

akalenyu:
> created with specific `securityContext.fsGroup` it updates the ownership of the volume files including `disk.img`

Are we sure this is true? all our pods should use the same one -  `fsGroup: 107`
Maybe selinuxcontext is the issue? from a conversation with @jsafrane I can see that's also taken care of
```
Typically, Pods use restricted or similar SCC and that makes sure all Pods in a single namespace with that SCC have the same SELinux label.
kubevirt-controller SCC has SELinux RunAsAny , which makes the Pod author responsible for filing the label. If unset, then random one is used.
```

pkg/controller/clone-controller.go
akalenyu: this came up last time when we faced an issue with read only in the host assisted path
the conclusion was that we'll always prefer compatibility in the host assisted case, so imposing a requirement here (even if only for fs) is risky business imo

mhenriks: I recall getting permission errors with this setting on ceph filesystem volumes.  Was hoping that readOnly set here would be sufficient:  https://github.com/kubevirt/containerized-data-importer/blob/3426b9c2a939e0da69a7be72dee37288b9676cd3/pkg/controller/clone-controller.go#L712

arnongilboa:
> Is there maybe a change we could make to our invocation of `tar` that would discard those changes? maybe something like [#2982 (comment)](https://github.com/kubevirt/containerized-data-importer/issues/2982#issuecomment-1810433172)?
> 

Tried the  --ignore-failed-read option but it didn't help. I see no tar option to ignore file ctime updates during its execution.

> It's just that we've seen provisioners in the field before simply reject read only mounts - #2944

Even if we do it only for filesystem?

arnongilboa:
> > created with specific `securityContext.fsGroup` it updates the ownership of the volume files including `disk.img`
> 
> Are we sure this is true? all our pods should use the same one - `fsGroup: 107` Maybe selinuxcontext is the issue? from a conversation with @jsafrane I can see that's also taken care of
> 

Checked again and the file ctime (stat Change) is updated during the concurrent clone tar execution, but I don't see any metadata changes (ownership or access permissions). Maybe each clone source mount update to the same `fsGroup: 107` is somehow causing the ctime update?
```
  File: disk.img
  Size: 32212254720	Blocks: 4557008    IO Block: 4096   regular file
Device: 804h/2052d	Inode: 429930422   Links: 1
Access: (0660/-rw-rw----)  Uid: (  107/     107)   Gid: (  107/ UNKNOWN)
Access: 2024-07-17 08:49:24.104100515 +0000
Modify: 2024-07-17 08:49:24.026099494 +0000
Change: 2024-07-17 08:55:25.315757750 +0000
 Birth: 2024-07-17 08:48:54.608714154 +0000
```

> ```
> Typically, Pods use restricted or similar SCC and that makes sure all Pods in a single namespace with that SCC have the same SELinux label.
> ```

akalenyu:
> Tried the --ignore-failed-read option but it didn't help. I see no tar option to ignore file ctime updates during its execution.

I believe what you're after is `--warning=no-file-changed` together with
```
-m, --touch
    Don't extract file modified time.
```
during extraction time.
But that may corrupt the file

> Even if we do it only for filesystem?

I'm not familiar with a provisioner that rejects this, but I think we just don't want to impose a new requirement on host assisted clones
https://github.com/kubevirt/containerized-data-importer/pull/3347#discussion_r1681625826

BTW - does this happen in other provisioners?

arnongilboa:
> > Tried the --ignore-failed-read option but it didn't help. I see no tar option to ignore file ctime updates during its execution.
> 
> I believe what you're after is `--warning=no-file-changed` together with

This is just about the warning. It will still fail with error code 1.
We use 1.34 and it seems to be fixed in in [1.35](https://www.gnu.org/software/tar/):
```
- Warn file changed as we read it less often. Formerly, tar warned if the file's size or ctime changed. However, this generated a false positive if tar read a file while another process hard-linked to it, changing its ctime. Now, tar warns if the file's size, mtime, user ID, group ID, or mode changes. Although neither heuristic is perfect, the new one should work better in practice.
- Fix --ignore-failed-read to ignore file-changed read errors as far as exit status is concerned. You can now suppress file-changed issues entirely with --ignore-failed-read --warning=no-file-changed.
```

> 
> ```
> -m, --touch
>     Don't extract file modified time.
> ```
> 
> during extraction time. But that may corrupt the file

The ctime seems to be updated by the other concurrent clones during the tar, so that won't help

> 
> > Even if we do it only for filesystem?
> 
> I'm not familiar with a provisioner that rejects this, but I think we just don't want to impose a new requirement on host assisted clones [#3347 (comment)](https://github.com/kubevirt/containerized-data-importer/pull/3347#discussion_r1681625826)
> 
> BTW - does this happen in other provisioners?

I'll check and update.

akalenyu:
> It will still fail with error code 1.

Really? interesting.. so all that does is hide the log? that seems weird, no?

> We use 1.34 and it seems to be fixed in in [1.35](https://www.gnu.org/software/tar/):

So do I understand correctly that the issue is alleviated with 1.35?

arnongilboa:
> > It will still fail with error code 1.
> 
> Really? interesting.. so all that does is hide the log? that seems weird, no?
> 

Yep, that's what the doc says, but in 1.35 you may combine it with --ignore-failed-read as detailed in the bug fixes.

> > We use 1.34 and it seems to be fixed in in [1.35](https://www.gnu.org/software/tar/):
> 
> So do I understand correctly that the issue is alleviated with 1.35?

Hopefully. How can I test with 1.35 when I build locally?

awels:
> Hopefully. How can I test with 1.35 when I build locally?
In the WORKSPACE file find the rpm for tar. A quick search reveals it is named `tar-2__1.34-6.el9.x86_64` change it to the version you want (don't forget to update the sha2456 to match). That should then install 1.35 when you build the container.

pkg/controller/clone-controller.go
akalenyu: you're right
https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3347/pull-containerized-data-importer-e2e-ceph/1813599133605302272/artifacts/1/pods/serial_2_cdi-e2e-tests-cloner-hk8fs_b17707b3-105a-45e3-af55-a162ff6b5497-source-pod-cdi-clone-source_previous.log

arnongilboa: saw that, looking for other directions.

arnongilboa:
@akalenyu should we consider `make rpm-deps` instead of updating just the tar rpm?

akalenyu:
> @akalenyu should we consider `make rpm-deps` instead of updating just the tar rpm?

Hey, yeah `make rpm-deps` shouldn't hurt. We had one recently which did not catch the tar bump, which I assume is very new?

arnongilboa:
> > @akalenyu should we consider `make rpm-deps` instead of updating just the tar rpm?
> 
> Hey, yeah `make rpm-deps` shouldn't hurt. We had one recently which did not catch the tar bump, which I assume is very new?

Yep, the tar build was per my request. However,  see the top commit,  `make rpm-deps` updates a lot of other rpms as well. Are we ok with it?

akalenyu:
> > > @akalenyu should we consider `make rpm-deps` instead of updating just the tar rpm?
> > 
> > 
> > Hey, yeah `make rpm-deps` shouldn't hurt. We had one recently which did not catch the tar bump, which I assume is very new?
> 
> Yep, the tar build was per my request. However, see the top commit, `make rpm-deps` updates a lot of other rpms as well. Are we ok with it?

Yeah, it should be just fine to bring in new packages from time to time
/approve

PR #3348, opened 2024-07-18
Fix race condition in multi-stage import logic

alromeros:

This Pull Request attempts to fix a race condition in the multi-stage import logic when updating checkpoint annotations for a PVC.

**Explanation:**

Once a multi-stage import pod succeeds and the current checkpoint is copied, we update the checkpoint and pod annotations to manually trigger a new importer pod with the updated data.

Cases where the pod has succeeded but still hasn't been deleted are possible. These situations can lead to irrecoverable states where the checkpoint annotations are updated, but the pod annotations remain outdated (still in the succeeded phase). This discrepancy halts progress in both the import-controller and the multi-stage import logic.

Fixes # https://issues.redhat.com/browse/CNV-40054
Fixes # https://issues.redhat.com/browse/CNV-40015 (tentatively)

I've been unable to replicate the race condition locally so let's consider this solution and wait to see if test failures stop.

**Release note**:

```release-note
Fix race condition when updating warm import checkpoints 
```


awels:
Okay lets assume this works and see if it improves stability.
/lgtm
/approve

PR #3349, opened 2024-07-18
s390x Unit Test Enablement

### LINKED ISSUE #3341: Prlimit Unit Test fails on s390x
**What happened**:
In #3226  it was discovered, that `prlimit_test.go` fails on s390x, when using a specific go version. The go version was then bumped to `1.22.3` and the Issue was closed, but the Problem is not yet resolved.

```
• [FAILED] [0.010 seconds]
Process Limits exec [It] command success with real limits
/root/go/src/kubevirt.io/containerized-data-importer/pkg/system/prlimit_test.go:62

  [FAILED] Expected
      <string>: unexpected fault address 0x180000
       fatal error: fault
       [signal SIGSEGV: segmentation violation code=0x2 addr=0x180000 pc=0x18002a]
       goroutine 1 gp=0xc0000021c0 m=0 mp=0x8241a0 [running, locked to thread]:
       runtime.throw({0x4edfa2...
```

There are currently 2 possible fixes for this:

1. omitting the `-coverprofile=.coverprofile` flag from the go-test command
https://github.com/kubevirt/containerized-data-importer/blob/364d582f76d5f50b23ecc1321537908d07139767/hack/build/run-unit-tests.sh#L26
2. increasing the memory allocation from 1GB(1 << 30) to 2GB(1 << 31)
https://github.com/kubevirt/containerized-data-importer/blob/364d582f76d5f50b23ecc1321537908d07139767/pkg/system/prlimit_test.go#L62

The second option seems to be the correct one, but not sure as both fixes do not seem to be related to me.

---

Unit Tests Succeeding

On a s390x LPAR:
1. clone `main`
2. export BUILDER_IMAGE=`<ported builder image for s390x>`
3. Either
    a). `go test -coverprofile=.coverprofile ./pkg/system/...`
    b). `./hack/build/bazel-docker.sh ./hack/build/run-unit-tests.sh ./pkg/system/..."`

**Environment**:
- Go version: 1.22.3
- OS (e.g. from /etc/os-release): RHEL 9.4
- Kernel (e.g. `uname -a`): 5.14.0-427.22.1.el9_4.s390x

Davo911:
Enable running unit tests on IBM Z Platform, by introducing a higher memory limit, when running on s390x.

Fixes #3341 

This fix is a temporary workaround, until the underlying issue is solved.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/system/prlimit_test.go
+++ b/pkg/system/prlimit_test.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -38,6 +39,13 @@ import (
 var _ = Describe("Process Limits", func() {
 	limits := &ProcessLimitValues{1, 1}
 	nullLimiter := newTestProcessLimiter(nil, nil)
+
+	realLimits := &ProcessLimitValues{1 << 30, 10}
+
+	//workaround for https://github.com/kubevirt/containerized-data-importer/issues/3341
+	if runtime.GOARCH == "s390x" {
+		realLimits = &ProcessLimitValues{1 << 31, 10}
+	}
 
 	DescribeTable("exec", func(commandOverride func(context.Context, string, ...string) *exec.Cmd, limiter ProcessLimiter, limits *ProcessLimitValues, command, output, errString string, args ...string) {
 		replaceExecCommandContext(commandOverride, func() {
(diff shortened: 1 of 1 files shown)


awels:
Hi if you could sign the commits that would make the DCO bot happy, you can run `git commit --amend -s` and it should automatically sign the commit

awels:
@Davo911 what needs to happen to get the tests to run? Do we need a new lane that points to the s390x test worker?

Davo911:
Hey @awels!
This is highly dependend on the enablement PR #3357.
s390x test worker with cdi deployed is now ready to go. 
In the next step the kubeconfig just needs to be encrypted and handed over to you guys. 
(This goes through @brianmcarey if I understood it correctly?)

https://github.com/kubevirt/project-infra/pull/3552 will be the prow job for unit testing

Davo911:
> Hmmm so the current unit test does not like this change, which is weird, since it should not have changed since the env variable is not set.

And interesingly the same memory allocation issue appeared for `amd64`:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3349/pull-cdi-unit-test/1818300000342380544#1:build-log.txt%3A1206-1240

awels:
Yeah, I am confused why that is happening. It is not like the value changed on the x86_64 run. It should not fail there.

Davo911:
Can we rerun the test, as I'm not able to recreate the issue on my `amd64` machine. (I cannot /retest I guess)
Apart from that, there seems to be a bigger underlying memory issue here, that needs to be investigated at one point.

brianmcarey:
The PR has ok-to-test so I think you should be able to trigger with the following - 

/test pull-cdi-unit-test

Davo911:
Awesome, thank you! Seems like it worked!

/retest

awels:
I think the /ok-to-test was disabled? I haven't seen it put the label on here at all.

brianmcarey:
> I think the /ok-to-test was disabled? I haven't seen it put the label on here at all.

Yes you are right - I had forgotten

awels:
/lgtm
/approve
Might have been a flake on that unit test.

PR #3353, opened 2024-07-24
Handle storage class changes automatically in cron controller

akalenyu:
Until now we've never performed the necessary adjustments when the desired storage class changed.
Instead, we documented it and relied on the user to perform the cleanup of actual sources.
This commit starts doing that on behalf of the user.

Fixes #

**Release note**:

```release-note
Handle desired storage class changes automatically in cron controller by deleting old storage class sources
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -43,6 +43,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	openapicommon "k8s.io/kube-openapi/pkg/common"
 	"k8s.io/utils/ptr"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


akalenyu:
cross posting comments from offline discussion;

TBH - I don't love it. We're essentially making it seamless to switch the storage provider for all VMs in the cluster...
This should be seamless for the **end user,** but the **admin** managing these should understand precisely what changing the storage class for them means
(And follow up by deleting sources as acknowledgement).

@arnongilboa @aglitke I know you've been hesitant as well in https://github.com/kubevirt/containerized-data-importer/pull/3326
where we essentially solved a subset of the problem where the "old" sc is not capable of provisioning volumes at all (stay Pending)

pkg/controller/dataimportcron-controller.go
```diff
@@ -673,6 +681,38 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+			return err
+		}
+	}
+	for _, src := range sources {
+		if err := r.client.Get(ctx, client.ObjectKeyFromObject(src), src); err == nil || !k8serrors.IsNotFound(err) {
+			return fmt.Errorf("waiting for old sources to get cleaned up: %w", err)
+		}
+	}
+	// Only update desired storage class once garbage collection went through
+	annPatch := fmt.Sprintf(`[{"op":"add","path":"/metadata/annotations/%s","value":"%s" }]`, strings.ReplaceAll(AnnStorageClass, "/", "~1"), desiredStorageClass)
```
mhenriks, line 707: check out `EscapeJSONPointer` in apimachinery lib

akalenyu: nice ty

awels (approved):

PR #3360, opened 2024-07-31
pkg/importer: Increase number of nbdkit lines logged

rwmjones:
30 is far too small.  For "chatty" plugins like VDDK many more debugging lines will be printed for simple open calls.  For debugging it is vital that we do not lose this information.  Ideally I would set this to infinity.

Related: https://issues.redhat.com/browse/CNV-44894

**Release note**:

```release-note
VDDK datasource: Increase number of nbdkit lines logged
```

CC @mrnold

### DIFF
--- a/pkg/importer/vddk-datasource_amd64.go
+++ b/pkg/importer/vddk-datasource_amd64.go
@@ -65,7 +65,7 @@ var newNbdKitLogWatcher = createNbdKitLogWatcher
 const (
 	nbdUnixSocket = "/tmp/nbd.sock"
 	nbdPidFile    = "/tmp/nbd.pid"
-	maxLogLines   = 30
+	maxLogLines   = 1000
 )
 
 var vddkVersion string


akalenyu (commented):
Looks good, if you want, you have the `nbdkit` cmd stdout in https://github.com/kubevirt/containerized-data-importer/blob/a9f2836c7dbf6a1e1c5e4a0a6559080b9525eda6/pkg/importer/vddk-datasource_amd64.go#L146 so you could just write the entire thing into `/tmp/nbdkit.log` by dropping https://github.com/kubevirt/containerized-data-importer/blob/a9f2836c7dbf6a1e1c5e4a0a6559080b9525eda6/pkg/image/nbdkit.go#L293-L296 in the scanner loop
(unless theres some implication in doing that)

mrnold:
> Looks good, if you want, you have the `nbdkit` cmd stdout in `output *bufio.Reader`
> 
> https://github.com/kubevirt/containerized-data-importer/blob/a9f2836c7dbf6a1e1c5e4a0a6559080b9525eda6/pkg/importer/vddk-datasource_amd64.go#L146
> so you could just write the entire thing into `/tmp/nbdkit.log` by dropping
> 
> https://github.com/kubevirt/containerized-data-importer/blob/a9f2836c7dbf6a1e1c5e4a0a6559080b9525eda6/pkg/image/nbdkit.go#L293-L296
> in the scanner loop
> (unless theres some implication in doing that)

This is probably the way to go, to see everything from nbdkit/VDDK get logged. I think there is a password printed in plain text in there somewhere though, it might be a good idea to filter that out.

rwmjones:
> I think there is a password printed in plain text in there somewhere though, it might be a good idea to filter that out.

The password is passed to nbdkit directly on the command line here:

https://github.com/kubevirt/containerized-data-importer/blob/1ac608717e3cb50d306e92ba1aabaeb804b2e5eb/pkg/image/nbdkit.go#L96

This is bad in several ways:

 * You expose the password to commands like `ps` (assuming someone has a shell inside the container).
 * It gets printed in nbdkit debug output, which you're filtering.
 * If the password starts with `-` or `+` then you likely have a security hole.

You can avoid all that by writing the password into a temporary file (ensure that the file is unreadable by other), and then passing `password=+/path/to/temp/file` on the command line instead.  If you don't want the temp file, you can also use inherited file descriptors and pipes but it's a bit harder to implement.

More details here (scroll down a bit to where it talks about the password parameter):

https://libguestfs.org/nbdkit-vddk-plugin.1.html#PARAMETERS

rwmjones:
The test fails *seem* unrelated ...?

akalenyu:
> The test fails _seem_ unrelated ...?

correct:
https://github.com/kubevirt/containerized-data-importer/pull/3024

akalenyu:
> Looks good, if you want, you have the `nbdkit` cmd stdout in `output *bufio.Reader`
> 
> https://github.com/kubevirt/containerized-data-importer/blob/a9f2836c7dbf6a1e1c5e4a0a6559080b9525eda6/pkg/importer/vddk-datasource_amd64.go#L146
> 
> so you could just write the entire thing into `/tmp/nbdkit.log` by dropping
> https://github.com/kubevirt/containerized-data-importer/blob/a9f2836c7dbf6a1e1c5e4a0a6559080b9525eda6/pkg/image/nbdkit.go#L293-L296
> 
> in the scanner loop
> (unless theres some implication in doing that)

Let me know if you plan to take this suggestion

rwmjones:
> Let me know if you plan to take this suggestion

I didn't really understand what exactly was meant.

akalenyu:
> > Let me know if you plan to take this suggestion
> 
> I didn't really understand what exactly was meant.
> 
> For the immediate customer bug we're facing, getting more log lines out from the customer's container is the most important thing here.

I am suggesting to dump the entire log into a file

rwmjones:
How can we see the log then?  Currently the customer runs `oc logs <podname>` and they can see (unfortunately only) 30 lines of the log.

arnongilboa:
> > > Let me know if you plan to take this suggestion
> > 
> > 
> > I didn't really understand what exactly was meant.
> > For the immediate customer bug we're facing, getting more log lines out from the customer's container is the most important thing here.
> 
> I am suggesting to dump the entire log into a file

You can't `oc rsh` to the pod if it failed/completed, so how would you get the nbd log if pod fails too quick?  we can keep a debug pod running but not sure that's the best way to go. I guess that's the reason we put it in the pod log.

akalenyu:
> How can we see the log then? Currently the customer runs `oc logs <podname>` and they can see (unfortunately only) 30 lines of the log.

You would have to watch that file in the container. Let me know if you want this merged as is

rwmjones:
@mrnold If you're happy with this, then so I am.  It's a minimal change to improve the situation and collect the logs, perhaps not the long term final fix.

The password issue noted above is also quite important.  I may have a go at fixing that if I get time today.

akalenyu:
/approve
/hold

@mrnold unhold when ready

mrnold:
/unhold

I think it's fine to keep this pull request to the one small change, and implement the other suggestions separately.

fabiand:
@akalenyu can we get this change backported to 4.15 and 4.14 (at least)?

akalenyu:
sure
/cherrypick release-v1.59

PR #3362, opened 2024-08-01
Fix PVC deletion logic in populators

alromeros:

This Pull Request ensures that we only proceed with PVC' deletion after confirming the target PVC is rebound.

Fixes # https://issues.redhat.com/browse/CNV-45479

**Release note**:

```release-note
Bugfix: Fix PVC deletion logic in CDI populators
```

### DIFF
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -295,7 +295,7 @@ func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorCon
 		return res, err
 	}
 
-	// Making sure to clean PVC once it is complete
+	// Making sure to clean PVC' once population is completed
 	if cc.IsPVCComplete(pvc) && !cc.IsMultiStageImportInProgress(pvc) {
 		res, err = r.reconcileCleanup(pvcPrime)
 	}
@@ -357,14 +357,16 @@ func (r *ReconcilerBase) reconcileCommon(pvc *corev1.PersistentVolumeClaim, popu
 }
 
 func (r *ReconcilerBase) reconcileCleanup(pvcPrime *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
-	if pvcPrime != nil {
-		if pvcPrime.Annotations[cc.AnnPodRetainAfterCompletion] == "true" {
-			// Retaining PVC' in Lost state. We can then keep the pod for debugging purposes.
-			r.recorder.Eventf(pvcPrime, corev1.EventTypeWarning, retainedPVCPrime, messageRetainedPVCPrime)
-		} else {
-			if err := r.client.Delete(context.TODO(), pvcPrime); err != nil {
-				return reconcile.Result{}, err
-			}
+	// If exists, make sure PVC' is rebound before deletion
+	if pvcPrime == nil || pvcPrime.Status.Phase != corev1.ClaimLost {
+		return reconcile.Result{}, nil
+	}
+	if pvcPrime.Annotations[cc.AnnPodRetainAfterCompletion] == "true" {
+		// Retaining PVC' in Lost state. We can then keep the pod for debugging purposes.
+		r.recorder.Eventf(pvcPrime, corev1.EventTypeWarning, retainedPVCPrime, messageRetainedPVCPrime)
+	} else {
+		if err := r.client.Delete(context.TODO(), pvcPrime); err != nil {
+			return reconcile.Result{}, err
 		}
 	}
 	return reconcile.Result{}, nil
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -296,7 +296,7 @@ func (r *ReconcilerBase) reconcile(req reconcile.Request, populator populatorCon
 	}
 
 	// Making sure to clean PVC' once population is completed
-	if cc.IsPVCComplete(pvc) && !cc.IsMultiStageImportInProgress(pvc) {
+	if cc.IsPVCComplete(pvc) && cc.IsBound(pvc) && !cc.IsMultiStageImportInProgress(pvc) {
 		res, err = r.reconcileCleanup(pvcPrime)
 	}
 
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -342,10 +342,10 @@ func (r *ReconcilerBase) reconcileCommon(pvc *corev1.PersistentVolumeClaim, popu
 		return nil, err
 	}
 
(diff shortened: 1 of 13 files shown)


alromeros:
/cc @mhenriks 
Thanks for the debugging! I'm surprised we didn't check this before, especially considering we used [this](https://github.com/kubernetes-csi/lib-volume-populator/blob/master/populator-machinery/controller.go#L879) as reference for our populators

pkg/controller/populators/populator-base.go
```diff
@@ -358,6 +358,10 @@ func (r *ReconcilerBase) reconcileCommon(pvc *corev1.PersistentVolumeClaim, popu
 
 func (r *ReconcilerBase) reconcileCleanup(pvcPrime *corev1.PersistentVolumeClaim) (reconcile.Result, error) {
 	if pvcPrime != nil {
```
akalenyu, line 360: not related to this pr but you could really improve the indentation in this func if you return early when `pvcPrime == nil`

alromeros: Wanted to avoid redundant `returns` but I agree, I'll change it.

pkg/controller/populators/populator-base.go
```diff
@@ -357,14 +357,16 @@ func (r *ReconcilerBase) reconcileCommon(pvc *corev1.PersistentVolumeClaim, popu
...
-	if pvcPrime != nil {
-		if pvcPrime.Annotations[cc.AnnPodRetainAfterCompletion] == "true" {
-			// Retaining PVC' in Lost state. We can then keep the pod for debugging purposes.
-			r.recorder.Eventf(pvcPrime, corev1.EventTypeWarning, retainedPVCPrime, messageRetainedPVCPrime)
-		} else {
-			if err := r.client.Delete(context.TODO(), pvcPrime); err != nil {
-				return reconcile.Result{}, err
-			}
+	// If exists, make sure PVC' is rebound before deletion
+	if pvcPrime == nil || pvcPrime.Status.Phase != corev1.ClaimLost {
```
mhenriks, line 361: I think this is better but can we can guarantee that pvcPrime being lost means that that the target pvc is bound?

My concern is that we could still hit this if target is unbound:  https://github.com/alromeros/containerized-data-importer/blob/fix-pvc-prime-deletion/pkg/controller/populators/populator-base.go#L348-L349

alromeros: That's a good point, I think the lost PVC' is a good indicator but `IsUnbound` isn't. The issue with `IsUnbound` is that it checks `pvc.Spec.VolumeName` (a field updated by our populator) instead of the status, which should be the definitive source.

I think changing the implementation of `IsUnbound` to `pvc.Status.Phase != corev1.ClaimBound` should fix this issue.

alromeros: Updated `IsBound` so we check the status instead of the spec. Maybe this solution is overkill (or there was good reasons to check the spec instead), but makes more sense to me.

mhenriks: > The issue with IsUnbound is that it checks pvc.Spec.VolumeName (a field updated by our populator)

I thought we only updated the PV claimRef?

alromeros: Still, the status update is the [last rebind step](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go#L1121) and we had issues in the past where we assumed a pvc to be bound but had outdated status (https://github.com/kubevirt/containerized-data-importer/pull/3126), so I think updating `IsBound` so we check the status is an improvement.

mhenriks: Not sure I can speak for every case in which `IsBound` is called but with populators I think having `IsBound` return `false` when the claim is `Lost` could be problematic

alromeros: I think we don't have any case where we actually check `IsBound`/`IsUnbound` with the pvc prime, so the target pvc should never be lost. That said, I can revert the last commit to avoid issues and just add an additional check in the necessary condition.

akalenyu:
This may need a backport - I am seeing it fail in 1.59 https://github.com/kubevirt/containerized-data-importer/pull/3372

pkg/controller/populators/populator-base.go
alromeros: I'm having second thoughts about changing this back. Checking the spec instead of the status has already provoqued several unwanted behaviors and this fix seems to be more accurate than other options. I'd personally wouldn't worry about the lost PVC' because we don't do this check with pvc primes.

mhenriks: > I'm having second thoughts about changing this back. Checking the spec instead of the status has already provoqued several unwanted behaviors and this fix seems to be more accurate than other options. 

Can you speak more about the unwanted behaviors?

> I'd personally wouldn't worry about the lost PVC' because we don't do this check with pvc primes.

I what thinking about the target becoming lost somehow.  If that happens, `IsBound` will return false and pvc' will be created.  And that pvc' will never be able to be rebound to the target

alromeros: One of them was this bug in clone populator where we assumed bound PVC but it still had unpopulated status: https://github.com/kubevirt/containerized-data-importer/pull/3126.
Another is this one, where we delete PVC' when target PVC is still unbound (despite having populated spec.VolumeName).

Maybe the change is overkill so I can add a single check before deleting the PVC' and make sure to address this specific case.

alromeros: However, having a function called `IsBound` that doesn't reliably tell you if the PVC is bound and that forces you to check the status alongside is really bad imo. Maybe we should do:

```go
func IsBound(pvc *corev1.PersistentVolumeClaim) {
    return pvc != nil && (pvc.Status.Phase == corev1.ClaimBound || pvc.Status.Phase == corev1.ClaimLost)
}
```

wdyt? I'm also ok with simply adding the check before deleting PVC' though.

mhenriks: Ah okay I forgot about #3126

Maybe we can have two functions:
1.  `IsBound` returns true when status == `Bound`
2. `IsRebindable` returns true when status is empty or Pending

Populator only creates pvc' when target `IsRebindable`

mhenriks: Question about `pvcPrime.Status.Phase != corev1.ClaimLost` part if this `if` statement.  What if user wants to delete while a long import is in progress. Will this keep prv' around until the import is complete?

alromeros: Added a new `IsLost` check when we manage the pvc' creation. I think that should work

PR #3363, opened 2024-08-01
nbdkit: Use password=+temporary to send passwords securely

rwmjones:
Passing passwords on the command line is not secure.  nbdkit lets you pass them via several secure mechanisms, for this I used the password=+temporary method (where "temporary" is a filename).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -84,7 +84,7 @@ func NewNbdkit(plugin NbdkitPlugin, nbdkitPidFile string) *Nbdkit {
 }
 
 // NewNbdkitCurl creates a new Nbdkit instance with the curl plugin
-func NewNbdkitCurl(nbdkitPidFile, user, password, certDir, socket string, extraHeaders, secretExtraHeaders []string) NbdkitOperation {
+func NewNbdkitCurl(nbdkitPidFile, user, password, certDir, socket string, extraHeaders, secretExtraHeaders []string) (NbdkitOperation, error) {
 	var pluginArgs []string
 	var redactArgs []string
 	args := []string{"-r"}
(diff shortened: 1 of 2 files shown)


rwmjones:
@mrnold 

Note I didn't even compile test this as I'm unable to do that here.

pkg/image/nbdkit.go
```diff
@@ -161,6 +169,24 @@ func NewNbdkitVddk(nbdkitPidFile, socket, server, username, password, thumbprint
 	return n, nil
 }
 
+func writePasswordFile(password string) (string, error) {
+	f, err := os.CreateTemp("", "password")
+	if err != nil {
+		return nil, err
+	}
```
cfergeau, line 176: You need `defer f.Close()` here so that the file is closed if the `Write` fails. Iirc even with the `defer` added, you can keep the explicit `f.Close()` at the end of the function as this will allow you to catch `Close()` errors.

pkg/image/nbdkit.go
```diff
@@ -161,6 +169,24 @@ func NewNbdkitVddk(nbdkitPidFile, socket, server, username, password, thumbprint
...
+	// This would be nice but we don't want to delete it until
+	// program exit.
+	// defer os.Remove(f.Name())
+
+	if _, err := f.Write([]byte(password)); err != nil {
+		return nil, err
```
cfergeau, line 182: `nil` is not a valid `string`, this should be `return "", err`

rwmjones:
NB I marked this as "ready for review" even though it hasn't even been compile tested.

rwmjones:
Well that was one way to compile test it ... Let me fix all those problems.

rwmjones:
Note the failed tests apply to the old version (d6cb213).  The current version is 85019a4 which builds locally for me.

awels:
This seems to be the failure at this point.

```
pkg/importer/http-datasource_test.go:48:22: cannot use image.NewMockNbdkitCurl (value of type func(nbdkitPidFile string, user string, password string, certDir string, socket string, extraHeaders []string, secretExtraHeaders []string) "kubevirt.io/containerized-data-importer/pkg/image".NbdkitOperation) as func(nbdkitPidFile string, user string, password string, certDir string, socket string, extraHeaders []string, secretExtraHeaders []string) ("kubevirt.io/containerized-data-importer/pkg/image".NbdkitOperation, error) value in assignment```

rwmjones:
I built it locally *and* ran the unit tests this time.

mrnold:
In case no one else has tested it for real yet, I have gotten some runtime with this patch applied to my local CDI, and it has been working fine. It also fixes the plaintext passwords getting logged, as far as I can tell.

PR #3370, opened 2024-08-07
Take target PVC selector into account when creating PVC prime

### LINKED ISSUE #3369: cdi Label matching bound pv does not take effect
**What happened**:
cdi specifies that the label matching binding pv is not valid. The matchlabels in pv and cdi are the same.

It is hoped that the cdi label can match the pv normally, so that the special node is scheduled

**The following are the cdi and pv contents**
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/provisioned-by: local.csi.openebs.io
  name: pvc-34fd945a-2634-4ffb-a7d4-94a9ac631234
  labels:
    openebs.io/nodename: cmcc-test-worker-01
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 60Gi
  csi:
    driver: local.csi.openebs.io
    fsType: xfs
    volumeAttributes:
      openebs.io/cas-type: localpv-lvm
      openebs.io/volgroup: localdatak8s
    volumeHandle: pvc-34fd945a-2634-4ffb-a7d4-94a9ac631234
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: openebs.io/nodename
          operator: In
          values:
          - cmcc-test-worker-01
  persistentVolumeReclaimPolicy: Delete
  storageClassName: openebs-lvmpv
  volumeMode: Filesystem
---
kind: DataVolume
metadata:
  name: "jammy-test3"
  annotations:
spec:
  source:
    http:
      url: "http://cloudimage.test.com:8000/jammy-server-cloudimg-amd64.img"
  pvc:
    resources:
      requests:
        storage: "60Gi"
    storageClassName: "openebs-lvmpv"
    accessModes:
    - ReadWriteOnce
    selector:
      matchLabels:
        openebs.io/nodename: cmcc-test-worker-01

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.59.0
- Kubernetes version (use `kubectl version`): 1.26.15
- DV specification: N/A
- OS (e.g. from /etc/os-release): ubuntu22.04 LTS
- Kernel (e.g. `uname -a`): 5.19.0-50-generic

akalenyu:
Today we completely disregard a selector being set on the target PVC,
which results in incorrect binding for anyone looking to use that.

Fixes #3369 

**Release note**:

```release-note
BugFix: Take target PVC selector into account when creating PVC prime
```

### DIFF
--- a/pkg/controller/populators/populator-base.go
+++ b/pkg/controller/populators/populator-base.go
@@ -199,6 +199,7 @@ func (r *ReconcilerBase) createPVCPrime(pvc *corev1.PersistentVolumeClaim, sourc
 			Resources:        pvc.Spec.Resources,
 			StorageClassName: pvc.Spec.StorageClassName,
 			VolumeMode:       pvc.Spec.VolumeMode,
+			Selector:         pvc.Spec.Selector,
 		},
 	}
 	pvcPrime.OwnerReferences = []metav1.OwnerReference{
--- a/pkg/controller/populators/import-populator_test.go
+++ b/pkg/controller/populators/import-populator_test.go
@@ -498,6 +498,30 @@ var _ = Describe("Import populator tests", func() {
 			Expect(pvcPrime.GetAnnotations()[AnnCurrentCheckpoint]).To(Equal("current"))
 			Expect(pvcPrime.GetAnnotations()[AnnFinalCheckpoint]).To(Equal("true"))
 		})
+
+		It("Should set selector on PVC prime", func() {
+			targetPvc := CreatePvcInStorageClass(targetPvcName, metav1.NamespaceDefault, &sc.Name, nil, nil, corev1.ClaimPending)
+			targetPvc.Spec.DataSourceRef = dataSourceRef
+			selector := metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"foo": "bar",
+				},
+			}
+			targetPvc.Spec.Selector = &selector
+			volumeImportSource := getVolumeImportSource(true, metav1.NamespaceDefault)
+
+			By("Reconcile")
+			reconciler = createImportPopulatorReconciler(targetPvc, volumeImportSource, sc)
+			result, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: targetPvcName, Namespace: metav1.NamespaceDefault}})
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(result).To(Not(BeNil()))
+
+			By("Checking PVC' selector")
+			pvcPrime, err := reconciler.getPVCPrime(targetPvc)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pvcPrime).ToNot(BeNil())
+			Expect(*pvcPrime.Spec.Selector).To(Equal(selector))
+		})
 	})
 
 	var _ = Describe("Import populator progress report", func() {


pkg/controller/populators/populator-base.go
```diff
@@ -199,6 +199,7 @@ func (r *ReconcilerBase) createPVCPrime(pvc *corev1.PersistentVolumeClaim, sourc
 			Resources:        pvc.Spec.Resources,
 			StorageClassName: pvc.Spec.StorageClassName,
 			VolumeMode:       pvc.Spec.VolumeMode,
+			Selector:         pvc.Spec.Selector,
```
mhenriks, line 202: Confirming we are already good on populator clone flows?

mhenriks (commented):
Func test may be tough for this one, huh?

pkg/controller/populators/populator-base.go
akalenyu: Good point, I haven't checked

akalenyu:
> Func test may be tough for this one, huh?

I think the unit test covers the entire thing, no? anyway it's doable

pkg/controller/populators/populator-base.go
akalenyu: Yeah looks fine. The temp restore PVC for snapshot cloning will not use the selector but that's not the target PVC, just in intermediary

zhangmingxian:
The function I want to achieve is to select a special node through the pvc selector when importing the image, referring to the following url, but it has not been implemented. Uh, can you tell me what to do? First contact with kubevirt.
  #https://github.com/kubevirt/containerized-data-importer/blob/main/doc/local-storage-selector.md

akalenyu:
> The function I want to achieve is to select a special node through the pvc selector when importing the image, referring to the following url, but it has not been implemented. Uh, can you tell me what to do? First contact with kubevirt. #https://github.com/kubevirt/containerized-data-importer/blob/main/doc/local-storage-selector.md

Yeah sure, this PR is handling that. Could you please check if this works for you? Would you like me to publish an image with the changes?
Normally dynamic provisioners would refuse to provision a claim with selector:
```
failed to provision volume with StorageClass "nfs-csi": claim Selector is not supported
```

akalenyu:
/hold
until we're sure the use case works in practice

awels:
> The function I want to achieve is to select a special node through the pvc selector when importing the image, referring to the following url, but it has not been implemented. Uh, can you tell me what to do? First contact with kubevirt. #https://github.com/kubevirt/containerized-data-importer/blob/main/doc/local-storage-selector.md

So that document was written a long time ago when there were no good dynamic provisioners availabe. As you can see it references kubernetes 1.11 which is ancient. In general you don't want to select nodes manually, but let the scheduler do this job for you. Is there a particular reason you want to select a node?

zhangmingxian:
> > The function I want to achieve is to select a special node through the pvc selector when importing the image, referring to the following url, but it has not been implemented. Uh, can you tell me what to do? First contact with kubevirt. #https://github.com/kubevirt/containerized-data-importer/blob/main/doc/local-storage-selector.md
> 
> So that document was written a long time ago when there were no good dynamic provisioners availabe. As you can see it references kubernetes 1.11 which is ancient. In general you don't want to select nodes manually, but let the scheduler do this job for you. Is there a particular reason you want to select a node?

My usage scenario is that there are multiple nodes with different nvidia graphics cards, and when I want to import images through datavolume, I select different nodes, so that I can create virtual machines with different Gpus

awels:
But openebs is a WaitForFirstConsumer storage class, you don't need to specify the node. You create the datavolume, and it won't actually populate until there is a workload that uses it. That workload is scheduled by the scheduler based on whatever node is the most appropriate for the workload requirements. Once the VM is scheduled, then CDI will populate the volume with the VM disk. There is no manual selecting of the node required in this scenario.

zhangmingxian:
> But openebs is a WaitForFirstConsumer storage class, you don't need to specify the node. You create the datavolume, and it won't actually populate until there is a workload that uses it. That workload is scheduled by the scheduler based on whatever node is the most appropriate for the workload requirements. Once the VM is scheduled, then CDI will populate the volume with the VM disk. There is no manual selecting of the node required in this scenario.

Yes, openebs is the WaitForFirstConsumer storage class. I refer to its lvm scheme, which uses the local disk space on the node. So is there any way to specify the node when pr imports the virtual machine image to pvc? Thank you.

awels:
If you just put a node selector on the VM, then the VM gets scheduled on that node, and the wait for first consumer will take care of getting the disk image on that node. But it is an anti pattern to do that in kubernetes, just let the scheduler figure it out.

zhangmingxian:
> If you just put a node selector on the VM, then the VM gets scheduled on that node, and the wait for first consumer will take care of getting the disk image on that node. But it is an anti pattern to do that in kubernetes, just let the scheduler figure it out.

Uh, well, maybe I wasn't quite right. Thank you. I'll think of other options.

awels:
I think we can close this, since using WFFC should solve the issue encountered.

zhangmingxian:
The root cause is the content of the annotation, so just comment out the annotation

PR #3377, opened 2024-08-08
feat: Copy labels from source to DataSource

0xFelix:

Copy labels from the source of a DataSource (DV/PVC/VolumeSnapshot) to the labels of the DataSource. This allows to make use of labels which were originally defined as env vars on a containerdisk.

Fixes #

**Release note**:

```release-note
Copy labels from the source of a DataSource (DV/PVC/VolumeSnapshot)
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -382,6 +382,8 @@ var (
 		AnnPriorityClassName:          "",
 		AnnPodMultusDefaultNetwork:    "",
 	}
+
+	validLabelsMatch = regexp.MustCompile(`^([\w.]+\.kubevirt.io|kubevirt.io)/[\w-]+$`)
 )
 
 // FakeValidator is a fake token validator
@@ -2026,6 +2028,16 @@ func CopyAllowedAnnotations(srcObj, dstObj metav1.Object) {
 		if val != "" {
 			klog.V(1).Info("Applying annotation", "Name", dstObj.GetName(), ann, val)
 			AddAnnotation(dstObj, ann, val)
+		}
+	}
+}
+
+// CopyAllowedLabels copies allowed labels matching the validLabelsMatch regexp from the
+// source map to the destination object allowing overwrites
+func CopyAllowedLabels(srcLabels map[string]string, dstObj metav1.Object, overwrite bool) {
+	for label, value := range srcLabels {
+		if _, found := dstObj.GetLabels()[label]; (!found || overwrite) && validLabelsMatch.MatchString(label) {
+			AddLabel(dstObj, label, value)
 		}
 	}
 }
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -105,6 +105,8 @@ const (
 	defaultImportsToKeepPerCron = 3
 )
 
+var ErrNotManagedByCron = errors.New("DataSource is not managed by this DataImportCron")
+
 // Reconcile loop for DataImportCronReconciler
 func (r *DataImportCronReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	dataImportCron := &cdiv1.DataImportCron{}
@@ -377,6 +379,15 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 				cc.AddAnnotation(snapshot, cc.AnnSourceVolumeMode, string(*volMode))
 			}
 		}
+		// Copy labels found on dataSource to the existing snapshot in case of upgrades.
+		dataSource, err := r.getDataSource(ctx, dataImportCron)
+		if err != nil {
+			if !k8serrors.IsNotFound(err) && !errors.Is(err, ErrNotManagedByCron) {
+				return res, err
+			}
+		} else {
+			cc.CopyAllowedLabels(dataSource.Labels, snapshot, true)
+		}
 		if err := r.updateSource(ctx, dataImportCron, snapshot); err != nil {
 			return res, err
 		}
@@ -494,6 +505,20 @@ func (r *DataImportCronReconciler) getSnapshot(ctx context.Context, cron *cdiv1.
 	return snapshot, nil
 }
 
+func (r *DataImportCronReconciler) getDataSource(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (*cdiv1.DataSource, error) {
+	dataSourceName := dataImportCron.Spec.ManagedDataSource
+	dataSource := &cdiv1.DataSource{}
+	if err := r.client.Get(ctx, types.NamespacedName{Namespace: dataImportCron.Namespace, Name: dataSourceName}, dataSource); err != nil {
+		return nil, err
+	}
+	if dataSource.Labels[common.DataImportCronLabel] != dataImportCron.Name {
+		log := r.log.WithName("getCronManagedDataSource")
+		log.Info("DataSource has no DataImportCron label or is not managed by cron, so it is not updated", "name", dataSourceName, "uid", dataSource.UID, "cron", dataImportCron.Name)
+		return nil, ErrNotManagedByCron
+	}
+	return dataSource, nil
+}
+
 func (r *DataImportCronReconciler) updateSource(ctx context.Context, cron *cdiv1.DataImportCron, obj client.Object) error {
 	objCopy := obj.DeepCopyObject()
 	cc.AddAnnotation(obj, AnnLastUseTime, time.Now().UTC().Format(time.RFC3339Nano))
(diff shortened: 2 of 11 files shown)


tests/datasource_test.go
```diff
@@ -114,6 +138,72 @@ var _ = Describe("DataSource", func() {
...
+		ds, err := f.CdiClient.CdiV1beta1().DataSources(ds.Namespace).Create(context.TODO(), ds, metav1.CreateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		ds = waitForReadyCondition(ds, corev1.ConditionTrue, "Ready")
+
+		By("Check for labels on DataSource")
+		ds, err = f.CdiClient.CdiV1beta1().DataSources(ds.Namespace).Get(context.TODO(), ds.Name, metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ds.Labels).To(HaveKeyWithValue(testKubevirtIoKey, testKubevirtIoValue))
+		Expect(ds.Labels).To(HaveKeyWithValue(testInstancetypeKubevirtIoKey, testInstancetypeKubevirtIoValue))
+		Expect(ds.Labels).To(HaveKeyWithValue(testKubevirtIoKeyExisting, testKubevirtIoNewValueExisting))
```
akalenyu, line 157: I think you're not guaranteed to have the labels when reaching "Ready=true". You could use eventually here

pkg/controller/common/util.go
```diff
@@ -2030,6 +2032,16 @@ func CopyAllowedAnnotations(srcObj, dstObj metav1.Object) {
 	}
 }
 
+// CopyAllowedLabels copies allowed labels matching the validLabelsMatch regexp from the
+// source map to the destination object allowing overwrites
+func CopyAllowedLabels(srcLabels map[string]string, dstObj metav1.Object, overwrite bool) {
```
akalenyu, line 2037: sure you want to support "overwrite" here? seems to me like it's not really necessary but may be wrong

pkg/controller/dataimportcron-controller.go
```diff
@@ -719,6 +712,7 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 		},
 	}
 	r.setDataImportCronResourceLabels(dataImportCron, desiredSnapshot)
+	cc.CopyAllowedLabels(pvc.GetLabels(), desiredSnapshot, false)
```
akalenyu, line 715: So this is already great, but I think to be truly backwards compatible (user carries a snapshot object over upgrade) we should also update an existing snapshot with the label. Something like https://github.com/kubevirt/containerized-data-importer/pull/3155/commits/755fe0bc6a612b55c426b3186a51ea12e5274ce2#diff-786a9fcaa47bbf5a42fae757e2e41d06f646e2e8b778e9c17a6ab595ef9a52fbR360-R373
Maybe this is why the functional test fails? not sure. Have to take a deeper look, very odd

tests/dataimportcron_test.go
```diff
@@ -154,6 +156,7 @@ var _ = Describe("DataImportCron", Serial, func() {
 				},
 			}
 			snapshot = utils.WaitSnapshotReady(f.CrClient, snapshot)
+			Expect(snapshot.Labels).To(HaveKeyWithValue(testKubevirtIoKey, testKubevirtIoValue))
```
akalenyu, line 159: Again not sure how this fails. Maybe it's possible to reach DV succeeded before the labels are copied from prime? does this consistently pass locally?

tests/datasource_test.go
```diff
@@ -114,6 +138,77 @@ var _ = Describe("DataSource", func() {
...
+			},
+			func() {
+				createDv(pvc1Name, testURL(), map[string]string{
+					testKubevirtIoKey:             testKubevirtIoValue,
+					testInstancetypeKubevirtIoKey: testInstancetypeKubevirtIoValue,
+					testKubevirtIoKeyExisting:     testKubevirtIoNewValueExisting,
+				})
+				err := utils.WaitForDataVolumePhase(f, f.Namespace.Name, cdiv1.Succeeded, pvc1Name)
+				Expect(err).ToNot(HaveOccurred())
+				err = utils.DeleteDataVolumeOrphanPVC(f.CdiClient, f.Namespace.Name, pvc1Name)
```
akalenyu, line 189: Seems a bit too clunky, maybe just create a populator standalone PVC? https://github.com/kubevirt/containerized-data-importer/blob/0f46feceb2fcefd1a8145c9773d92c6007b490f6/doc/cdi-populators.md?plain=1#L21

pkg/controller/datasource-controller.go
```diff
@@ -296,7 +305,8 @@ func addDataSourceControllerWatches(mgr manager.Manager, c controller.Controller
 			CreateFunc: func(e event.TypedCreateEvent[*snapshotv1.VolumeSnapshot]) bool { return true },
 			DeleteFunc: func(e event.TypedDeleteEvent[*snapshotv1.VolumeSnapshot]) bool { return true },
 			UpdateFunc: func(e event.TypedUpdateEvent[*snapshotv1.VolumeSnapshot]) bool {
-				return !reflect.DeepEqual(e.ObjectOld.Status, e.ObjectNew.Status)
+				return !reflect.DeepEqual(e.ObjectOld.Status, e.ObjectNew.Status) ||
+					!reflect.DeepEqual(e.ObjectOld.Labels, e.ObjectNew.Labels)
```
akalenyu, line 309: :+1:

akalenyu (commented):
Looks good! main worry is ofc the func test failure/flake which essentially means a user snap could be missing the labels

akalenyu:
ignore the nfs lane, this is a known issue, but the ceph lane failure is surely something we care about lol

pkg/controller/common/util.go
0xFelix: IIUC we need overwrite to allow overriding labels on a DataSource when the env vars/labels of the containerdisk changed.

tests/datasource_test.go
0xFelix: ACK, will change it

pkg/controller/common/util.go
akalenyu: Yeah I am saying just always overwrite to avoid the extra bool param. Unless that's a bad idea

pkg/controller/dataimportcron-controller.go
0xFelix: It should not be needed since in this case the labels should remain on an existing `DataSource`. There would also be no way to regain this information without reimporting again.

tests/dataimportcron_test.go
0xFelix: Added 2fa5f271be170bfb0f5f795006061b558b132faf to hopefully mitigate this.

pkg/controller/common/util.go
0xFelix: Ah I see, this func is shared between multiple places and some expect labels to not be overwritten.

akalenyu:
/test all
just want to see that we got rid of the flakes

0xFelix:
Fail of NFS lane was unrelated

0xFelix:
Failures are unrelated to the changes, the cluster failed to come up.

/retest

pkg/controller/populators/import-populator.go
```diff
@@ -160,20 +160,20 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
...
+				return reconcile.Result{}, err
+			}
+			if err := cc.Rebind(context.TODO(), r.client, pvcPrime, pvcCopy); err != nil {
 				return reconcile.Result{}, err
 			}
 		}
 	}
 
-	if pvcCopy, err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
+	if _, err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
```
akalenyu, line 173: Do we want the same treatment for annotations? https://github.com/kubevirt/containerized-data-importer/pull/3381#issuecomment-2286207076

pkg/controller/dataimportcron-controller.go
```diff
@@ -733,7 +727,14 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
...
 		}
 	} else {
-		if cc.IsSnapshotReady(currentSnapshot) {
+		updatedSnapshot := currentSnapshot.DeepCopy()
+		cc.CopyAllowedLabels(pvc.GetLabels(), updatedSnapshot, true)
+		if !reflect.DeepEqual(currentSnapshot, updatedSnapshot) {
+			if err := r.client.Update(ctx, updatedSnapshot); err != nil {
+				return err
+			}
+		}
```
akalenyu, line 736: I don't think this would solve the case for an existing snapshots with no labels, maybe we could just do something similar to https://github.com/kubevirt/containerized-data-importer/pull/3155/commits/755fe0bc6a612b55c426b3186a51ea12e5274ce2 by copying the labels from the data source to the snapshot. This would only be needed for a brief time window before the cron gets a sha update.

0xFelix: Updated the approach to match 755fe0b, PTAL!

pkg/controller/dataimportcron-controller.go
```diff
@@ -377,6 +379,11 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 				cc.AddAnnotation(snapshot, cc.AnnSourceVolumeMode, string(*volMode))
 			}
 		}
+		// Copy labels found on dataSource to the existing snapshot in case of upgrades.
+		dataSource, err := r.getDataSource(ctx, dataImportCron)
+		if err == nil {
```
akalenyu, line 384: I think we're "forgetting" to requeue in case the GET fails for a random network err

akalenyu (commented):
/approve

thanks!

PR #3380, opened 2024-08-08
Inject cdi-uploadproxy CA cert into user created routes

mhenriks:

Users want to create their own routes but certificate rotation make it difficult to keep them up to date

To take advantage of this feature, routes must meet the following restrictions:

1.  Be in the same namespace as cdi-operator
2. have `operator.cdi.kubevirt.io/injectUploadProxyCert: "true"` annotation

Fixes #

https://issues.redhat.com/browse/CNV-31492

**Release note**:

```release-note
Inject cdi-uploadproxy CA cert into user created routes
```

### DIFF
--- a/pkg/operator/controller/callbacks.go
+++ b/pkg/operator/controller/callbacks.go
@@ -52,7 +52,6 @@ import (
 func addReconcileCallbacks(r *ReconcileCDI) {
 	r.reconciler.AddCallback(&appsv1.Deployment{}, reconcileDeleteControllerDeployment)
 	r.reconciler.AddCallback(&corev1.ServiceAccount{}, reconcileSCC)
-	r.reconciler.AddCallback(&appsv1.Deployment{}, reconcileCreateRoute)
 	r.reconciler.AddCallback(&appsv1.Deployment{}, reconcileCreatePrometheusInfra)
 	r.reconciler.AddCallback(&appsv1.Deployment{}, reconcileRemainingRelationshipLabels)
 	r.reconciler.AddCallback(&appsv1.Deployment{}, reconcileDeleteDeprecatedResources)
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/callbacks.go
```diff
@@ -126,6 +128,10 @@ func reconcileCreateRoute(args *callbacks.ReconcileCallbackArgs) error {
 	}
 	args.Recorder.Event(cr, corev1.EventTypeNormal, createResourceSuccess, "Successfully ensured upload proxy route exists")
 
+	if err := updateUserRoutes(context.TODO(), args.Logger, args.Client, args.Recorder); err != nil {
```
akalenyu, line 131: huh, apparently args.Client is uncached https://github.com/kubevirt/containerized-data-importer/blob/36f1d191924abd0d2297269b6fd7ea89bfbb3048/vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/callbacks/callback-dispatcher.go#L143-L145 I think we could use cached client since https://github.com/kubevirt/containerized-data-importer/pull/3314/commits/a0e0afddcafd4b7277af38f25d3e369aba944bdd but not a huge deal

akalenyu (commented):
/approve

pkg/operator/controller/callbacks.go
mhenriks: Yeah I'd like to revisit the whole "uncached" thing in a separate PR. I think it is a remnant or old controller runtime days when could only have namespace or cluster level watches on everything. More configurable now

akalenyu: So in the operator that's already taken care of https://github.com/kubevirt/containerized-data-importer/blob/36f1d191924abd0d2297269b6fd7ea89bfbb3048/cmd/cdi-operator/operator.go#L90-L94 your client only ever caches the operator namespace, but I am not sure how you could pass the ctrl runtime cached client to the callback. Seems like a bug in the sdk tbh so yes outside of this PR

PR #3382, opened 2024-08-12
cross-compile docker builder for s390x

### LINKED ISSUE #3289: Add s390x cross-build libraries to builder container docker file
The problem this solves is enabling a cross-build of CDI s390x containers on x86. 

The libraries for s390x cross-compilation are maintained in the epel repositories, so the preferred solution is to `dnf install` these libraries directly into the x86 builder container for CDI, i.e. putting these lines into the Dockerfile:

```
RUN \
       if test "${ARCH}" = amd64; then 
       dnf install -y --setopt=install_weak_deps=False  --installroot /usr/s390x-linux-gnu/sys-root --forcearch s390x --releasever 9 \
          glibc-devel \
          glibc-static && \
      dnf install -y epel-release && \
      dnf install -y --setopt=install_weak_deps=False \
          gcc-s390x-linux-gnu && \
      dnf clean -y all;
fi
```

It wasn't clear, at first, whether the support for s390x cross-compiles should be statically built into the builder container as in kubevirt, or should be installed on the fly during the build of the CDI images as is done for cross-compile of arm containers. 

**Additional context**:
in order to _only_ build the cross-compile support for s390x into the x86 builder container, need to check the architecture of the platform as in the kubevirt builder build.

cfilleke:

Incorporates cross-compile build tools for s390x in x86 CDI builder container 

Fixes #3289

**Release note**:

```release-note
enables cross compile tools for s390x for CDI builder
```

### DIFF
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -52,6 +52,16 @@ COPY output-bazel-arch.sh /output-bazel-arch.sh
 
 RUN curl -L -o /usr/bin/bazel https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-$(sh /output-bazel-arch.sh) && chmod u+x /usr/bin/bazel
 
+RUN if test "${ARCH}" = amd64; then \
+      dnf install -y --setopt=install_weak_deps=False  --installroot /usr/s390x-linux-gnu/sys-root --forcearch s390x --releasever 9 \
+          glibc-devel \
+          glibc-static && \
+      dnf install -y epel-release && \
+      dnf install -y --setopt=install_weak_deps=False \
+          gcc-s390x-linux-gnu && \
+      dnf clean -y all; \
+    fi
+
 # Until we use a version including the fix for this Bazel issue:
 # https://github.com/bazelbuild/bazel/issues/11554
 RUN ln -s /usr/bin/python3 /usr/bin/python
--- a/hack/build/docker/builder/Dockerfile
+++ b/hack/build/docker/builder/Dockerfile
@@ -53,7 +53,7 @@ COPY output-bazel-arch.sh /output-bazel-arch.sh
 RUN curl -L -o /usr/bin/bazel https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-$(sh /output-bazel-arch.sh) && chmod u+x /usr/bin/bazel
 
 
-RUN if test "${ARCH}" = x86_64; then \
+RUN if test "$(sh /output-bazel-arch.sh)" = x86_64; then \
       dnf install -y --setopt=install_weak_deps=False  --installroot /usr/s390x-linux-gnu/sys-root --forcearch s390x --releasever 9 \
           glibc-devel \
           glibc-static && \
--- a/hack/build/docker/builder/output-bazel-arch.sh
+++ b/hack/build/docker/builder/output-bazel-arch.sh
@@ -10,7 +10,7 @@ s390x)
     ARCH="s390x"
     ;;
 *)
-    echo "invalid Arch, only support x86_64, aarch64"
+    echo "invalid Arch, only support x86_64, aarch64, s390x"
     exit 1
     ;;
 esac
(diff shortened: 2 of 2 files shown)


hack/build/docker/builder/Dockerfile
```diff
@@ -52,6 +52,16 @@ COPY output-bazel-arch.sh /output-bazel-arch.sh
 
 RUN curl -L -o /usr/bin/bazel https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-$(sh /output-bazel-arch.sh) && chmod u+x /usr/bin/bazel
 
+RUN if test "${ARCH}" = amd64; then \
```
akalenyu, line 55: hmm, `amd64`?

jschintag: There is no
```
ARG ARCH
```
declaration in the Dockerfile, so this should be an unset variable.

Additionally you would need to declare it as ``--build-arg ARCH=<architecture>`.

I think easier in this case would be to simply use the [pre-defined build variable](https://docs.docker.com/build/building/variables/#pre-defined-build-arguments) `TARGETARCH`

jschintag (commented):
Could you please squash your 2 commits, they have the same commit message anyway

cfilleke:
ARCH is set in the script output-bazel-arch.sh and should be positive only if ARCH is x86 not s390x

hack/build/docker/builder/Dockerfile
```diff
@@ -52,7 +52,19 @@ COPY output-bazel-arch.sh /output-bazel-arch.sh
 
 RUN curl -L -o /usr/bin/bazel https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-$(sh /output-bazel-arch.sh) && chmod u+x /usr/bin/bazel
 
-# Until we use a ver including the fix for this Bazel issue:
+
+RUN if test "${ARCH}" = x86_64; then \
```
jschintag, line 56: I don't think the environment is kept between build steps and besides, the script is never sourced anyway.
What i would propose then is to replace the Variable with the script.
```suggestion
RUN if test "$(sh /output-bazel-arch.sh)" = x86_64; then \
```

akalenyu: Isn't `x86_64` a typo?

cfilleke:

expect it to fail with `Error response from daemon: cannot listen on the TCP port: listen tcp4 :34061: bind: address already in use` again?  Is there a way to resolve that @awels ?

awels:
No that is a race condition when two clusters are started at the same time. Sometimes you end up with the same port in use. Just restart it.

cfilleke:
All tests passed except 'tide' and this apears to now require only /lgtm /approve now @mhenriks and @alromeros  and this is a size S but please feel free to ask me questions during your review.

hack/build/docker/builder/Dockerfile
```diff
@@ -52,7 +52,19 @@ COPY output-bazel-arch.sh /output-bazel-arch.sh
 
 RUN curl -L -o /usr/bin/bazel https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-$(sh /output-bazel-arch.sh) && chmod u+x /usr/bin/bazel
 
-# Until we use a ver including the fix for this Bazel issue:
+
+RUN if test "$(sh /output-bazel-arch.sh)" = x86_64; then \
```
akalenyu, line 56: @cfilleke am I missing something or this should not be x86_64?

cfilleke: When the builder container has architecture x86_64, then we add cross-compile libraries, so that when the builder container is running on x86, it builds, via cross-compilation using the libraries included,  CDI operator containers with the s390x exec architecture.  That would be case 'C' in the attached diagram. 

Screenshot 2024-08-21 at 4 04 07 PM


Theoretically, you could have all the CDI builder containers enabled to do cross-compilation to produce CDI operators of any other architecture; however, I think the logic behind having only the x86_64 builder enabled to do cross-compilation is to limit this to the cases we actually need.

akalenyu: gotcha - I'm happy if you just include a short comment about this above the condition

PR #3384, opened 2024-08-13
Ensure scratch space is large enough to hold disk image

awels:
It is possible the source image is fully allocated and in that case the scratch space would not be large enough to hold the disk image. This takes fsOverhead into account when calculating the scratch space size.

Fixes #

**Release note**:

```release-note
Increased size of scratch space to take fs overhead into account.
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -159,7 +159,7 @@ func newScratchPersistentVolumeClaimSpec(pvc *corev1.PersistentVolumeClaim, pod 
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: []corev1.PersistentVolumeAccessMode{"ReadWriteOnce"},
-			Resources:   pvc.Spec.Resources,
+			Resources:   *pvc.Spec.Resources.DeepCopy(),
 		},
 	}
 	if storageClassName != "" {
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -29,6 +29,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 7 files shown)


pkg/controller/util.go
```diff
@@ -170,6 +171,17 @@ func newScratchPersistentVolumeClaimSpec(pvc *corev1.PersistentVolumeClaim, pod
...
+	sizeRequest := scratchPvcSpec.Spec.Resources.Requests[corev1.ResourceStorage]
+	fsOverhead, err := GetFilesystemOverhead(context.TODO(), client, scratchPvcSpec)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get filesystem overhead for scratch PVC")
+	}
+	fsOverheadFloat, _ := strconv.ParseFloat(string(fsOverhead), 64)
+	usableSpaceRaw := util.CalculateOverheadSpace(fsOverheadFloat, sizeRequest.Value())
+
+	scratchPvcSpec.Spec.Resources.Requests[corev1.ResourceStorage] = *resource.NewScaledQuantity(usableSpaceRaw, 0)
+
```
mhenriks, line 184: Correct me if I'm wrong but for FileSystem access mode if target storageclass and scratch storagclass are the same you should be safe to use the same size as target?

awels: Right, but there is nothing requiring this, I need to take a look at this again, I was originally only doing this if the target was block, but it is possible scratch and target are both filesystem but different overhead. So I need to calculate the size we are targeting in the target, and then based on that calculate the appriopriate scratch space size.

tests/utils/datavolume.go
```diff
@@ -374,7 +374,7 @@ func NewDataVolumeWithHTTPImportToBlockPV(dataVolumeName string, size string, ht
 					URL: httpURL,
 				},
 			},
-			PVC: &k8sv1.PersistentVolumeClaimSpec{
+			Storage: &cdiv1.StorageSpec{
```
akalenyu, line 377: This is quite a challenge to all of the tests that use this API (could even hurt their integrity, if they don't intend to test that)
I'd prefer if we don't do it in this PR at least

awels: Well part of the issue is the current state is also not working reliably. And it was constantly failing on this PR.

akalenyu: In that case could just change the spec for those tests, but IIUC the main issue was us mutating the pointer

awels: There are 2 issues. 1. The size was not large enough causing a 100% failure rate. Once I made it a little larger, there was the mutating the pointer issue that got fixed with the deepcopy.

tests/cloner_test.go
```diff
@@ -516,7 +516,7 @@ var _ = Describe("all clone tests", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				volumeMode := v1.PersistentVolumeFilesystem
-				targetDV := utils.NewDataVolumeForImageCloning("target-dv", "1.1Gi", sourcePvc.Namespace, sourcePvc.Name, nil, &volumeMode)
+				targetDV := utils.NewDataVolumeForImageCloning("target-dv", "1.2Gi", sourcePvc.Namespace, sourcePvc.Name, nil, &volumeMode)
```
akalenyu, line 519: TBH I don't fully understand this. Ceph CSI will give you back 2Gi whether you do 1.1 or 1.2

akalenyu (commented):
/lgtm

PR #3385, opened 2024-08-13
VDDK: pass snapshot ID through to nbdkit.

mrnold:
As discussed in [CNV-44894](https://issues.redhat.com/browse/CNV-44894), some VMs show `Error 13 (You do not have access rights to this file) ` when attempting to access snapshots for a warm migration. This mostly appears to affect Windows VMs with VMware guest tools installed. Passing the `snapshot` parameter to nbdkit avoids the error and allows delta copies to be applied to a multi-stage data volume.

This pull request adds the `snapshot` parameter to the nbdkit command line when a VDDK data volume is given a checkpoint to copy. It also adds a `transports` parameter to avoid a crash in the VDDK (see [RHEL-54377](https://issues.redhat.com/browse/RHEL-54377)).

Fixes [CNV-44894](https://issues.redhat.com/browse/CNV-44894)

**Release note**:
```release-note
Add nbdkit command line parameters to improve reliability of multi-stage VDDK imports.
```

### DIFF
--- a/pkg/importer/vddk-datasource_test.go
+++ b/pkg/importer/vddk-datasource_test.go
@@ -279,9 +279,9 @@ var _ = Describe("VDDK data source", func() {
 		var returnedDiskName string
 
 		newVddkDataSource = createVddkDataSource
-		newNbdKitWrapper = func(vmware *VMwareClient, fileName string) (*NbdKitWrapper, error) {
+		newNbdKitWrapper = func(vmware *VMwareClient, fileName, snapshot string) (*NbdKitWrapper, error) {
 			returnedDiskName = fileName
-			return createMockNbdKitWrapper(vmware, fileName)
+			return createMockNbdKitWrapper(vmware, fileName, snapshot)
 		}
 		currentVMwareFunctions.Properties = func(ctx context.Context, ref types.ManagedObjectReference, property []string, result interface{}) error {
 			switch out := result.(type) {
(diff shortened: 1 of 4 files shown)


pkg/importer/vddk-datasource_test.go
```diff
@@ -315,6 +315,38 @@ var _ = Describe("VDDK data source", func() {
...
+					out.Snapshot = createSnapshots(expectedSnapshot, "snapshot-11")
+				}
+			case *mo.VirtualMachineSnapshot:
+				out.Config = *createVirtualDiskConfig("snapshot-disk", 123456)
+				disk := out.Config.Hardware.Device[0].(*types.VirtualDisk)
+				parent := disk.Backing.(*types.VirtualDiskFlatVer1BackingInfo).Parent
+				parent.FileName = "snapshot-root"
+			}
+			return nil
+		}
```
arnongilboa, line 343: Not sure I understand this

arnongilboa:
@mrnold please check the consistently failing vddk func tests.

akalenyu:
Sonar is yelling about the amount of parameters
image
Maybe passing an options struct to it would make sense?

However, there's a slight chance the errors in the VDDK test are not related:
```
Error: failed get datavolume: dv-vddk-warm-import-test, client rate limiter Wait returned an error: context deadline exceeded
```
I will hit retest and worst case, if these are consistent, they'll show up again
/retest

pkg/importer/vddk-datasource_test.go
arnongilboa: Thanks!

tools/vddk-test/vddk-test-plugin.c
```diff
@@ -20,24 +20,27 @@
...
+int expected_arg_count = 7;
 
 void fakevddk_close(void *handle) {
     close(*((int *) handle));
 }
 
 int fakevddk_config(const char *key, const char *value) {
     arg_count++;
+    if (strcmp(key, "snapshot") == 0) {
+        expected_arg_count = 8;
```
akalenyu, line 33: So IIUC this commit fixes the e2e test lanes, right?
What's weird is that the failures manifested themselves as rate limit errors:
```
Error: failed get datavolume: dv-vddk-warm-import-test, client rate limiter Wait returned an error: context deadline exceeded
```
Do you happen to have any insight around that?

mrnold: I don't know exactly what rate limiter is showing up here, but it makes sense that the test exceeds some deadline somewhere. I think these tests don't look at the importer pod at all, they just create a data volume and wait for it to finish. So if the importer is consistently crashing, the data volume never finishes, and the test times out.

akalenyu: Right, so, in this case, the importer was crashing over this change missing?

> I don't know exactly what rate limiter is showing up here

That would be the client that tries to get said DataVolume

mrnold: Yes, this is what fixes the E2E tests. The importer was using this mock VDDK plugin, and one of the only things it checks is the number of arguments. So it was failing repeatedly without showing anything in the test logs.

akalenyu: Thanks for the explanation, might bug you a little about the mock plugin offline

akalenyu:
> Passing the snapshot parameter to nbdkit avoids the error and allows delta copies to be applied to a multi-stage data volume.

Could you elaborate a little what used to happen prior to passing the snapshot parameter? Was it looping through and looking for them?

/test pull-cdi-apidocs
unrelated - this is now fixed

mrnold:
> Could you elaborate a little what used to happen prior to passing the snapshot parameter? Was it looping through and looking for them?

I'm not sure. As far as I can tell nbdkit just passes this parameter on to [VixDiskLib_ConnectEx](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/8f96698a-0e7b-4d67-bb6c-d18a1d101540/ef536a47-27cd-481a-90ef-76b38e75353c/doc/GUID-A0F8043B-BA97-4296-A81C-D6A6B981FD7F.html) in the VDDK, which says it's required but doesn't really say why.

rwmjones:
> I'm not sure. As far as I can tell nbdkit just passes this parameter on to [VixDiskLib_ConnectEx](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/8f96698a-0e7b-4d67-bb6c-d18a1d101540/ef536a47-27cd-481a-90ef-76b38e75353c/doc/GUID-A0F8043B-BA97-4296-A81C-D6A6B981FD7F.html) in the VDDK, which says it's required but doesn't really say why.

The VMware docs are wrong, the snapshot parameter is only needed if you want to open a snapshot.  In this case since you're trying to open a snapshot, it seems you do need this parameter.  But inside virt-v2v we never use the snapshot parameter and it's not a problem.

PR #3388, opened 2024-08-15
fix(import-populator): Make copying of annotations more robust

0xFelix:

Similar to 2fa5f27, make copying of annotations from a prime PVC to a target PVC more robust, by moving it before rebinding the PV from prime to target. This way we can ensure the annotations are already present once the PVC becomes ready.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/controller/populators/import-populator.go
+++ b/pkg/controller/populators/import-populator.go
@@ -160,7 +160,10 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		}
 
 		if cc.IsPVCComplete(pvcPrime) && cc.IsUnbound(pvc) {
-			// Once the import is succeeded, we copy labels and rebind the PV from PVC to target PVC
+			// Once the import is succeeded, we copy annotations and labels and rebind the PV from PVC to target PVC
+			if pvcCopy, err = r.updatePVCWithPVCPrimeAnnotations(pvcCopy, pvcPrime, r.updateImportAnnotations); err != nil {
+				return reconcile.Result{}, err
+			}
 			if pvcCopy, err = r.updatePVCWithPVCPrimeLabels(pvcCopy, pvcPrime.GetLabels()); err != nil {
 				return reconcile.Result{}, err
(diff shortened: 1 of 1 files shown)


0xFelix:
@akalenyu PTAL, not sure this is 100% required.

0xFelix:
@akalenyu Do you think this is still needed?

akalenyu:
> @akalenyu Do you think this is still needed?

I think so, yeah
It's completely possible that the annotations only make it after the target is rebound

0xFelix:
@akalenyu It looks better now, wdyt?

mhenriks:
looks good I think but I'll let @alromeros be the judge

pkg/controller/populators/import-populator.go
```diff
@@ -160,7 +160,10 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 		}
 
 		if cc.IsPVCComplete(pvcPrime) && cc.IsUnbound(pvc) {
-			// Once the import is succeeded, we copy labels and rebind the PV from PVC to target PVC
+			// Once the import is succeeded, we copy annotations and labels and rebind the PV from PVC to target PVC
```
alromeros, line 163: We will update the annotations again after the rebind, are we sure we need that? Having three `updatePVCWithPVCPrimeAnnotations` calls in the same function makes this harder to maintain. It's not this PR's fault though. Maybe we could move the two updateAnnotations and updateLabels just above the switch?

alromeros: Nevermind, `UpdatesMultistageImportSucceeded` updates the PVC prime annotations and we would need to call it afterwards

alromeros: What about having one at the top before entering the switch and another one after checking `UpdatesMultistageImportSucceeded`? That would leave us with two calls. Not a must though, I'll lgtm and we can consider improving this code a bit in the near future.

alromeros:
Thanks!
/lgtm

PR #3391, opened 2024-08-16
fix: mismatched type (stat.Bsize to int64)

nestoracunablanco:
The linter find this error in some archs. Paradoxically it works on x86. This was introduced in the commit
3c06e26bacac7d0de43a8cf20884574c95237423

**What this PR does / why we need it**: s390x arch enablement

Fixes #

**Release note**:

```release-note NONE

```

### DIFF
--- a/pkg/util/file.go
+++ b/pkg/util/file.go
@@ -113,7 +113,8 @@ func GetAvailableSpace(path string) (int64, error) {
 	if err != nil {
 		return int64(-1), err
 	}
-	return int64(stat.Bavail) * stat.Bsize, nil
+	//nolint:unconvert
+	return int64(stat.Bavail) * int64(stat.Bsize), nil
 }
 
 // GetAvailableSpaceBlock gets the amount of available space at the block device path specified.


awels:
Hmmm odd I was under the impression that if the first element of an operation `*` in this case it automatically converts the other elements to the correct type.
/test all

Davo911:
Linter still catches this, should a [exclusion rule](https://golangci-lint.run/usage/false-positives/) be added?

awels:
I am thinking maybe a [no lint directive](https://golangci-lint.run/usage/false-positives/#nolint-directive) would be appropriate for this. Unless you have this popping up in many locations?

nestoracunablanco:
@awels @Davo911, I used the `//nolint:all` directive, both at the end of the line and at the beginning of the method, but it did not work. Are there any specific rules that cannot be excluded?

@awels, in response to your question: the issue occurs in only one location, but it is critical for enabling the s390x architecture, as we cannot run `make all` without either modifying the code or disabling the linter.

awels:
Weird, I just tried this locally, and it passed the linter
```go
	//nolint:unconvert
	return int64(stat.Bavail) * int64(stat.Bsize), nil
```
Not sure why it is failing for you.

nestoracunablanco:
@awels, I apologize for the confusion. I initially checked only the first failing test and noticed that there were timeouts, which prompted me to call retest. After reviewing the second failed test case, I understand your point.

To summarize:
- On s390x, the compiler raises an issue because `stat.Bsize` is a uint32, leading to a type mismatch.
- On amd64, the linter flags this as an unnecessary conversion.

I have implemented your suggestion. Let's wait for the CI's verdict.

nestoracunablanco:
FYI: @cfilleke

nestoracunablanco:
@awels, do you think the failing tests are a result of this change, or could it simply be a hiccup in the test suite?

awels:
There is a lot of load on the CI right now, I don't think these issues are related to the PR. which really doesn't change anything.

nestoracunablanco:
@awels Thank you for the information. Would it be possible to re-trigger the tests once the CI load decreases? I can contact you in a couple of days.

PR #3406, opened 2024-08-22
feat: Set default host architecture in makefile

nestoracunablanco:
When building natively on ARM or s390x, the default architecture is set to x86_64. This can lead to issues, as Bazel may not be able to find the corresponding toolchain, resulting in a failed build process.

**What this PR does / why we need it**: multiarch

Fixes #

**Release note**:

```release-note NONE
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ else
 	DO_BAZ=eval
 endif
 # x86_64 aarch64 crossbuild-aarch64 s390x crossbuild-s390x
-BUILD_ARCH?=x86_64
+BUILD_ARCH?=$(shell uname -m)
 
 ##@ General
 .DEFAULT_GOAL := help


Makefile
```diff
@@ -23,7 +23,7 @@ else
 	DO_BAZ=eval
 endif
 # x86_64 aarch64 crossbuild-aarch64
-BUILD_ARCH?=x86_64
+BUILD_ARCH?=$(shell uname -m)
```
akalenyu, line 26: Hmm wouldn't this be an issue if you're on a mac and doing x86 testing? @mhenriks IIRC that's your setup?

nestoracunablanco:
Hi @awels, in one of the failed tests, I found the following log entry: 19075 http-datasource.go:353] http: expected status code 200, got 500. Do you think this error could be related to the recent change?

awels:
This means the http server returned an internal error (500). There can be many reasons for that. Is this in the failed hpp-previous lane or somewhere locally for you?

nestoracunablanco:
It appears that the pull-cdi-unit-test throws this error the CI, but the command `make test-unit` runs successfully for me on both locally tested architectures: amd64 and s390x.

awels:
Oh the unit tests simulate all kinds of failures on the server side to ensure we handle them properly the unit test here failed due to a pr limit which sometimes flakes.
/test all

PR #3410, opened 2024-08-27
Remove git diff, refine plattform check for multiarch builder

Davo911:
There was an issue building the builder image with emulation.

The git diff check seemed to prevent the build to run, which should also be done in prow.

The Plattform check is made more readable as well as use the rpm list instead of a specific location.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/bazel-build-builder.sh
+++ b/hack/build/bazel-build-builder.sh
@@ -20,24 +20,20 @@ source "${script_dir}"/common.sh
 source "${script_dir}"/config.sh
 
 if [ "${CDI_CONTAINER_BUILDCMD}" = "buildah" ]; then
-    if [ -e /proc/sys/fs/binfmt_misc/qemu-aarch64 ]; then
-        BUILDAH_PLATFORM_FLAG="--platform linux/s390x,linux/amd64,linux/arm64"
-    else
-        echo "No qemu-user-static on host machine, building only native container"
-        BUILDAH_PLATFORM_FLAG=""
+    BUILDAH_PLATFORM_FLAG="--platform linux/amd64"
+    
+    # Check for qemu-user-static for arm64
+    if rpm -qa | grep -q qemu-user-static-aarch64; then
+        BUILDAH_PLATFORM_FLAG="${BUILDAH_PLATFORM_FLAG},linux/arm64"
     fi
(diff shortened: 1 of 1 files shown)


Davo911:
cc: @brianmcarey

brianmcarey:
The `git diff` check should not be required when the postsubmit is updated with `run_if_changed` - https://github.com/kubevirt/project-infra/pull/3604/

hack/build/bazel-build-builder.sh
```diff
@@ -20,25 +20,21 @@ source "${script_dir}"/common.sh
...
 fi
 
-# When this runs during the post-submit job, the PR will have squashed into a single
-# commit and we can use HEAD~1 to compare. The exit code of the git diff will therefore
-# be 0 in this case, so negating it indicates that yes this is a post-submit job and
-# we should re-build the builder. Separating out this logic from the test for clarity
-
-set +e
-git diff-index --quiet HEAD~1 hack/build/docker
-diffret=$?
```
brianmcarey, line 39: I think the diff return is checked here - https://github.com/kubevirt/containerized-data-importer/pull/3410/files#diff-e17c215b0d874f885d9e2dd93002cc694859b349a99ffa0975b9f3a64f9a759aL50 - so there are some updates needed there.

hack/build/bazel-build-builder.sh
```diff
@@ -47,7 +41,7 @@ set -e
 # to use DOCKER_PREFIX as it is set in config.sh and used elsewhere in
 # the build system, to introduce a little more consistency
 
-if [ $diffret -ne 0 ] || [ x"${ADHOC_BUILDER}" != "x" ]; then
+if [ x"${ADHOC_BUILDER}" != "x" ]; then
```
akalenyu, line 44: I think we can also drop `ADHOC_BUILDER`, no? it was only added since the git diff condition was blocking the publish

Davo911: makes sense to me
I also don't find any other reference to this

akalenyu: https://github.com/kubevirt/containerized-data-importer/pull/3251

brianmcarey (commented):
/lgtm 

The lane has been updated to only run when there are changes under hack/build/docker/builder/

https://github.com/kubevirt/project-infra/commit/17598ad4d17709be18c09431ba75ec3c7a72d117

Issue #3417, opened 2024-09-01
Deprecate and delete Object transfer


ShellyKa13:
This feature is no longer in use and not being supported. It should be marked as deprecated in beta and deleted when v1.

PR #3418, opened 2024-09-01
Deprecate ObjectTransfer feature

ShellyKa13:
Added warnings and comments to ObjectTransfer crd and webhook.
Reference to relevant github issue:
https://github.com/kubevirt/containerized-data-importer/issues/3417

**Release note**:

```release-note
Deprecating ObjectTransfer feature
```


akalenyu:
So there's this reference for a deprecation that occurred in kubevirt:
https://github.com/kubevirt/kubevirt/pull/8069
Which seems to also specify the deprecation field on the CRD. Maybe we want this too?

akalenyu:
```
ERROR: checking pvc target-dv phase failed: etcdserver: request timed out
```
This etcd failure stuff is super weird. It's been making our polling funcs return early due to an error
which historically would never happen.
I will try to run some dummy PRs against older CDI branches to scratch out the possibility that we have some
insane hotlooping that overwhelms etcd.

PR #3431, opened 2024-09-10
Run unit-tests without bazel

Davo911:

As we don't really need bazel for the unit tests I'd run them using `in-docker.sh` instead of `bazel-docker.sh`
This would also solve some issues on `s390x` where bazel is currently not available.

Fixes https://github.com/kubevirt/project-infra/pull/3552

**Release note**:

```release-note
NONE
```


Davo911:
cc @brianmcarey

PR #3433, opened 2024-09-12
3026: Add support for s3 data importer inheriting service account credentials

### LINKED ISSUE #3429: CDI importer requires static credentials for S3 and GCS (renewed)
This is a reopening of https://github.com/kubevirt/containerized-data-importer/issues/3026 -- It was marked stale, but will be worked on this week.

**Is your feature request related to a problem? Please describe:**
Workloads should not have access to long-lived credentials. In cloud environments in particular, there are mechanisms for distributing short-lived credentials and SDKs support fetching and utilizing such creds. CDI prevents using the security best practice of using short-lived credentials for image retrieval from S3 and GCS sources.

**Describe the solution you'd like:**
CDI's operator should provide a method for specifying the image importer pod's ServiceAccount. This would enable access to existing credential distribution mechanisms: AWS's IRSA credentials as well as GKE Workload Identity. Additionally, the importer will need to be updated to rely on "ambient" credentials (retrieved by the cloud provider SDK), instead of [the importer's current hard-coded reliance on static credentials](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/s3-datasource.go#L174). This could be accomplished by extending the S3DataSource and GCSDataSource structs with a serviceAccountName member, updating the relevant function signatures, and then branching when creating the client.

**Describe alternatives you've considered:**
Currently have created a specific IAM user, issued an access key ID and secret key, and using a set of hard-coded creds in a Kubernetes secret referenced by the importer. It "works", at the cost of violating existing security approaches for workload authentication.

russellcain:
What type of PR is this?
- /kind feature

What this PR does / why we need it:
- Allows for consistent use of credentials configured in the service account, for S3 data volume sources.

Which issue(s) this PR fixes:
- Fixes #3429

Special notes for your reviewer:
- Hi -- I'll add in some tests shortly, just wanted to get eyes on the core of the change to make sure I'm not barking up the wrong trunk. If we could start the conversation in the interim, I'd appreciate that.

Does this PR introduce a user-facing change?:
- It does allow a new value in the DataVolumeSourceS3 configuration. The needed code-gen files are included in this pull request. 

**Release note**:

```release-note
Related to https://github.com/kubevirt/containerized-data-importer/issues/3429

This change stems from the desire for a spun-up S3 reader to utilize the aws credentials configured at the Service Account level. The belief is that this would promote consistency in access / minimize any duplication of aws credential keys in the configuration of this S3 CDI. For this PR, we just provide the user with the ability to specify the `ServiceAccountName` instead of an `AccessKey`/`SecretKey` combo. No core/default functionality will change for users which wish to stick with that key combo approach.  

At the importer level, this update utilizes AWS' IRSA credential chain approach. This is a stable approach that has go sdk approach. Additionally, this PR added the recommended `CredentialsChainVerboseErrors` to the aws session configuration. If reviewers worry that this will be too noisy, I am fine with it being removed.

```

Signed-off-by: _____


russellcain:
(recent force push to include signoff)
@aglitke  and @ShellyKa13  thanks for taking the time to look this over! any and all pointers on how you'd like tests to be done are appreciated

russellcain:
@mhenriks would you mind kicking off the tests for me, please?

cmd/cdi-importer/importer.go
```diff
@@ -255,6 +255,7 @@ func newDataSource(source string, contentType string, volumeMode v1.PersistentVo
 	ep, _ := util.ParseEnvVar(common.ImporterEndpoint, false)
 	acc, _ := util.ParseEnvVar(common.ImporterAccessKeyID, false)
 	sec, _ := util.ParseEnvVar(common.ImporterSecretKey, false)
+	serviceAccount, _ := util.ParseEnvVar(common.ImporterServiceAccountName, false)
```
mhenriks, line 258: I don't see that this env var is getting set on the import pod.  Here are some reference points:

https://github.com/russellcain/containerized-data-importer/blob/a87f77d4f010ee3fb2ca4b30111b7d70596c2eb2/pkg/controller/import-controller.go#L558
https://github.com/russellcain/containerized-data-importer/blob/a87f77d4f010ee3fb2ca4b30111b7d70596c2eb2/pkg/controller/import-controller.go#L1176

cmd/cdi-importer/importer.go
```diff
@@ -271,34 +272,45 @@ func newDataSource(source string, contentType string, volumeMode v1.PersistentVo
...
 	case cc.SourceRegistry:
 		ds := importer.NewRegistryDataSource(ep, acc, sec, certDir, insecureTLS)
 		return ds
 	case cc.SourceS3:
-		ds, err := importer.NewS3DataSource(ep, acc, sec, certDir)
+		var (
+			ds  *importer.S3DataSource
+			err error
+		)
+		if serviceAccount != "" {
```
mhenriks, line 292: Is the value of `serviceAccount` ever used in the aws API? Or does the aws lib assume that the pod is running as the privileged service account?

russellcain: The latter! hopefully i was too far off base, but the approach was that this would be used as a "flag" given the service account should have been properly set up by the time it is invoked here? (or fail on referencing an un-instantiated SA)

cmd/cdi-importer/importer.go
russellcain: ah rats! thanks for closing this gap in my understanding -- let me take a stab at that right now

russellcain: okay so fingers crossed that I understood how this flows in [04d29c3](https://github.com/kubevirt/containerized-data-importer/pull/3433/commits/04d29c38a690926df87a6e2c51794cebb1d7210a)

I dont need to add anything here, right? 
https://github.com/russellcain/containerized-data-importer/blob/41ba02d03ce0205b38fdfc9cbae4a1dc17b3abff/tools/cdi-source-update-poller/main.go#L48
looks like more so this parses env vars (which the SA doesnt fall into) and args passed at invocation (also not applicable)?

should i be adding a check about the source [here](https://github.com/russellcain/containerized-data-importer/blob/a12446110b32c2f3fe1888dd3aed888111a075f6/pkg/controller/import-controller.go#L1290)? and then set it to any value if the PodSpec has a SA set?

pkg/controller/import-controller.go
```diff
@@ -1263,6 +1264,10 @@ func makeImportEnv(podEnvVar *importPodEnvVar, uid types.UID) []corev1.EnvVar {
 			Name:  common.CacheMode,
 			Value: podEnvVar.cacheMode,
 		},
+		{
+			Name:  common.ImporterServiceAccountName,
+			Value: podEnvVar.serviceAccountName,
```
mhenriks, line 1269: Couple things 

1) if the worker pod does not need to know the SA name and this param is just a flag for setting `sessionConfig.CredentialsChainVerboseErrors = aws.Bool(true)`  then why not make this a boolean value with key something like `S3_CREDENTIALS_CHAIN_AUTH`

2) You have to set the service account name in the importer pod spec somewhere in here: https://github.com/russellcain/containerized-data-importer/blob/3026/IRSA-integration-for-s3-service-accounts/pkg/controller/import-controller.go#L884-L949

russellcain:
will wait back on a response to [this conversation](https://github.com/kubevirt/containerized-data-importer/pull/3433#discussion_r1775183308) before i push up my changes. thanks again, @mhenriks

pkg/controller/import-controller.go
mhenriks: - `S3_CREDENTIALS_CHAIN_AUTH` would just be the name of the environment variable in the importer pod spec. This should not be visible to users in any way. Importer pods are created/managed by cdi-deployment controller. The user will still specify the service account name in the API as you have already declared

- The pod `spec.serviceAccountName` should be set to the value the user set in the datavolume.  See here: https://github.com/russellcain/containerized-data-importer/blob/3026/IRSA-integration-for-s3-service-accounts/pkg/controller/import-controller.go#L915-L925

```golang
Spec: corev1.PodSpec{
			Containers:        makeImporterContainerSpec(args),
			InitContainers:    makeImporterInitContainersSpec(args),
			Volumes:           makeImporterVolumeSpec(args),
			RestartPolicy:     corev1.RestartPolicyOnFailure,
			NodeSelector:      args.workloadNodePlacement.NodeSelector,
			Tolerations:       args.workloadNodePlacement.Tolerations,
			Affinity:          args.workloadNodePlacement.Affinity,
			PriorityClassName: args.priorityClassName,
			ImagePullSecrets:  args.imagePullSecrets,
-------------> ServiceAccountName: args.serviceAccountName
		},
```

chomatdam: some documentation [here](https://github.com/aws/amazon-eks-pod-identity-webhook?tab=readme-ov-file#eks-walkthrough) explaining what happens when the `serviceAccountName` is added to the pod, what the `pod-identity-webhook` is mutating in the pod spec

russellcain:
thanks for the added context, @chomatdam  -- can you let me know if my changes in [69501ee](https://github.com/kubevirt/containerized-data-importer/pull/3433/commits/69501ee9fdb1844470ceb9d67e5349a09f3b4a1f) seem to line up with the approach you described, please @mhenriks? thanks again for your guidance on this

russellcain:
@mhenriks hi - would you mind kicking off the tests for this / taking a quick look over [69501ee](https://github.com/kubevirt/containerized-data-importer/pull/3433/commits/69501ee9fdb1844470ceb9d67e5349a09f3b4a1f) when you have a chance, please? thank you!

cmd/cdi-importer/importer.go
```diff
@@ -255,6 +255,7 @@ func newDataSource(source string, contentType string, volumeMode v1.PersistentVo
 	ep, _ := util.ParseEnvVar(common.ImporterEndpoint, false)
 	acc, _ := util.ParseEnvVar(common.ImporterAccessKeyID, false)
 	sec, _ := util.ParseEnvVar(common.ImporterSecretKey, false)
+	use_s3_credential_chain_auth, _ := strconv.ParseBool(os.Getenv(common.UseS3CredentialsChainAuth))
```
mhenriks, line 258: `use_s3_credential_chain_auth` is a little verbose compared to other variable names maybe just `s3ChainAuth`?

pkg/controller/import-controller.go
```diff
@@ -76,32 +76,33 @@ type ImportReconciler struct {
...
+	finalCheckpoint              string
+	preallocation                bool
+	httpProxy                    string
+	httpsProxy                   string
+	noProxy                      string
+	certConfigMapProxy           string
+	extraHeaders                 []string
+	secretExtraHeaders           []string
+	cacheMode                    string
+	use_s3_credential_chain_auth bool
```
mhenriks, line 105: how about `s3ChainAuth` as variable name?

pkg/controller/import-controller.go
```diff
@@ -117,6 +118,7 @@ type importerPodArgs struct {
 	workloadNodePlacement   *sdkapi.NodePlacement
 	vddkImageName           *string
 	priorityClassName       string
+	serviceAccountName      string
```
mhenriks, line 121: Where is this getting assigned?

mhenriks (commented):
getting there!

pkg/controller/import-controller.go
russellcain: i was thinking it should be set here? https://github.com/kubevirt/containerized-data-importer/pull/3433/files#diff-868af4b730094e9180f78af4cc069ae6cecf7077a46c186553f331dccbd32993R954

russellcain:
@mhenriks may I have another review / test run, please?

awels:
Looks like there is a compiler error here 
```
# kubevirt.io/containerized-data-importer/pkg/controller [kubevirt.io/containerized-data-importer/pkg/controller.test]
pkg/controller/import-controller_test.go:966:4: unknown field serviceAccountName in struct literal of type importPodEnvVar
pkg/controller/import-controller_test.go:1250:18: undefined: common.ImporterServiceAccountName
pkg/controller/import-controller_test.go:1251:21: podEnvVar.serviceAccountName undefined (type *importPodEnvVar has no field or method serviceAccountName)
FAIL	kubevirt.io/containerized-data-importer/pkg/controller [build failed] 
```

pkg/controller/import-controller.go
akalenyu: went back to this PR following a similar change https://github.com/kubevirt/containerized-data-importer/pull/4056, I agree that this wasn't being assigned anywhere like @mhenriks suggested. But really this is just a ping so we know the new PR is a step in the right direction

PR #3434, opened 2024-09-13
Handle PunchHole/AppendZeroWithTruncate failures

mhenriks:

Turns out gcp standard-csi provisioner does not support PunchHole for block devices We should handle these failures and simply write zeros

Fixes #

**Release note**:

```release-note
Handle PunchHole/AppendZeroWithTruncate failures
```

### DIFF
--- a/pkg/util/file.go
+++ b/pkg/util/file.go
@@ -168,6 +168,9 @@ func PunchHole(outFile *os.File, start, length int64) error {
 	return err
 }
 
+// unit test support to simulate failure and fallback to zero writes
+var appendZeroWithTruncateFunc = AppendZeroWithTruncate
+
 // AppendZeroWithTruncate resizes the file to append zeroes, meant only for newly-created (empty and zero-length) regular files.
 func AppendZeroWithTruncate(outFile *os.File, start, length int64) error {
 	klog.V(4).Infof("Truncating %d-bytes from offset %d", length, start)
@@ -228,7 +231,7 @@ func StreamDataToFile(r io.Reader, fileName string, preallocate bool) (int64, in
 
 	if !preallocate {
 		var isDevice bool
-		zeroWriter := AppendZeroWithTruncate
+		zeroWriter := appendZeroWithTruncateFunc
 		isDevice, err = IsDevice(fileName)
 		if err != nil {
 			return 0, 0, err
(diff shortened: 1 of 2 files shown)


mhenriks:
/hold
Will validate on gcp tomorrow

mhenriks:
/unhold

Validated on gcp. Interesting observation is that PunchHole would work fine initially then stop for all subsequent calls.  I assumed that it would fail all the time

pkg/util/file.go
```diff
@@ -261,28 +264,50 @@ func StreamDataToFile(r io.Reader, fileName string, preallocate bool) (int64, in
 
 type zeroWriterFunc func(*os.File, int64, int64) error
 
+func zeroWriterWithFallback(zwf zeroWriterFunc) func(dst *os.File, start, length int64) (int64, error) {
+	return func(dst *os.File, start, length int64) (int64, error) {
+		err := zwf(dst, start, length)
+		if err != nil {
```
akalenyu, line 270: should we be more precise in here? something like https://github.com/containers/storage/blob/10981abafe3e3ece4e0cc872761f7624cd1bef0e/drivers/copy/copy_linux.go#L111

mhenriks: Figured best to do exactly what we are doing here: https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/imageio-datasource.go#L309. I don't think we can count on all provisioners returning the "right" error

I doubt that transient errors happen frequently and since "retrying over transient errors" is starting the entire clone operation over from the beginning I think this solution is fine

pkg/util/file.go
```diff
@@ -261,28 +264,50 @@ func StreamDataToFile(r io.Reader, fileName string, preallocate bool) (int64, in
...
-			if bytes.Equal(writeBuf[0:nr], zeroBuf[0:nr]) {
+			var zbw int64
+			if checkZeros && bytes.Equal(writeBuf[0:nr], zeroBuf[0:nr]) {
 				bytesRead += int64(nr)
 			} else {
 				if bytesRead > writeOffset {
-					// zeroWriter func should seek to bytesRead before returning
-					ew = zeroWriter(dst, writeOffset, bytesRead-writeOffset)
+					// func should seek to bytesRead before returning
+					zbw, ew = zeroWriterFunc(dst, writeOffset, bytesRead-writeOffset)
```
alromeros, line 302: I think using abbreviations like these (nw, we, er, etc.) makes the code hard to maintain, but the logic looks really good. Not a must so lgtm.

alromeros (commented):
/lgtm

PR #3447, opened 2024-09-23
[release-v1.59] Bump tar rpm to fix concurrent host-assisted clone failure

arnongilboa:

Manual backport of #3347

```
```

```
```

jira-ticket: https://issues.redhat.com/browse/CNV-43105

**Release note**:
```release-note
```


akalenyu (commented):
/approve
/lgtm

PR #3448, opened 2024-09-23
[release-v1.60] Bump tar rpm to fix concurrent host-assisted clone failure

arnongilboa:

Manual backport of #3347

```
```

```
```

jira-ticket: https://issues.redhat.com/browse/CNV-48778

**Release note**:
```release-note
```


WORKSPACE
```diff
@@ -3345,20 +3345,20 @@ rpm(
...
 rpm(
-    name = "tar-2__1.34-6.el9.aarch64",
-    sha256 = "98a9ca5a25c6aa73b5183b3333abad062a8f82d8b9390d2b2fbdc1eea5b4fb9b",
+    name = "tar-2__1.34-7.el9.aarch64",
+    sha256 = "e3ee12a44a68c84627e43c2512ad8904a4778a82b274d0e8147ca46645f4a1fb",
     urls = [
-        "http://mirror.stream.centos.org/9-stream/BaseOS/aarch64/os/Packages/tar-1.34-6.el9.aarch64.rpm",
-        "https://storage.googleapis.com/builddeps/98a9ca5a25c6aa73b5183b3333abad062a8f82d8b9390d2b2fbdc1eea5b4fb9b",
+        "http://mirror.stream.centos.org/9-stream/BaseOS/aarch64/os/Packages/tar-1.34-7.el9.aarch64.rpm",
+        "https://storage.googleapis.com/builddeps/e3ee12a44a68c84627e43c2512ad8904a4778a82b274d0e8147ca46645f4a1fb",
```
akalenyu, line 3352: wondering how you got the mirroring to work. AFAIK this is something that gets added by a bot in such PRs:
https://github.com/kubevirt/containerized-data-importer/pull/3420

awels: Its probably pointing to the mirroring from the main PR.

arnongilboa: right

akalenyu: I actually looked at main, the mirroring was set up in the original PR :O

arnongilboa: The mirroring would have been automated to the same value by the following `make rpm-deps` anyway. No idea why I kept it as two commits.

akalenyu: The mirroring isn't done as part of the deps target, it's done by the bot https://github.com/kubevirt/containerized-data-importer/pull/3420
I guess you just predicted the mirror address, and the bot populated it
/approve
/lgtm

Issue #3159, opened 2024-04-01
Virtual image size is larger than the reported available storage


g7nguym:
**What happened**:
I tried to upload .iso image via CDI to a PVC. The iso image is nearly 5GB. It always throws an error saying the virtual image is larger than the reported available storage. I have tried multiple PVC sizes: 12Gi, 64Gi, 120Gi but it is not working.

**What you expected to happen**:
The iso image should be successfully uploaded to PVC

**How to reproduce it (as minimally and precisely as possible)**:
Following command: 
kubectl virt image-upload pvc win-2022-std-iso --size=120Gi --image-path=win22-std.iso --storage-class=datavg-thin-pool --uploadproxy-url=https://10.49.172.185:31876 --insecure --wait-secs=60 --access-mode=ReadWriteOnce --force-bind
PVC default/win-2022-std-iso not found
PersistentVolumeClaim default/win-2022-std-iso created
Waiting for PVC win-2022-std-iso upload pod to be ready...
Pod now ready
Uploading data to https://10.49.172.185:31876

 4.67 GiB / 4.67 GiB [==========================================================================================================================================================] 100.00% 2m13s

unexpected return value 400, Saving stream failed: Virtual image size 12886302720 is larger than the reported available storage 12884901888. A larger PVC is required.

**Additional context**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 1.58.3
- Kubernetes version (use `kubectl version`): 1.29.0
- DV specification: N/A
- Cloud provider or hardware configuration: kind
- OS (e.g. from /etc/os-release): Rocky Linux 9.3
- Kernel (e.g. `uname -a`): 5.14.0-362.24.1.el9_3.x86_64
- Install tools: kubectl virt
- Others: N/A

akalenyu:
Definitely weird - the error indicates that only 12Gi is available in the volume (even though you asked for 120)
Is it possible that the storage provisioner (`--storage-class=datavg-thin-pool`) is providing a volume smaller than the request?

You could test this by creating a 120Gi PVC and a pod mounting it;
Then run something like (%a - available (what we care about), %f - total free)
```bash
bash-5.1# stat /pvcmountpath/ -fc %a
907
bash-5.1# stat /pvcmountpath/ -fc %f
974
```
To get total size you would multiply by the block size (`stat /pvcmountpath/ -fc %s`)

g7nguym:
Hi @akalenyu 
[...]

PR #3458, opened 2024-10-11
Fix: checking available image size

kvaps:
Signed-off-by: Andrei Kvapil <kvapss@gmail.com>

This PR fixes the issue of checking size for the end device.

Currently we check if actual block-volume is larger than requested size in PVC.
However we should check if actual block-volume is smaller than requested size in PVC, because this might blocking the image upload.

Fixes https://github.com/kubevirt/containerized-data-importer/issues/3159

**Release note**:

```release-note
Fix: checking available image size
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -242,8 +242,8 @@ func checkIfURLIsValid(info *ImgInfo, availableSize int64, image string) error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


kvaps:
Closed in favor https://github.com/kubevirt/containerized-data-importer/pull/3461

PR #3463, opened 2024-10-15
Exclude Unsupported Func-Test Images for s390x

Davo911:
`vddk` and `imagio` in not available on s390x and won't be in the near future.
Because of this I propose a exclusion of the corresponding func-test images, when calling `cluster-sync`

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-sync/sync.sh
+++ b/cluster-sync/sync.sh
@@ -181,8 +181,8 @@ if [ "${CDI_SYNC}" == "test-infra" ]; then
   _kubectl apply -f "./_out/manifests/sample-populator.yaml"
   _kubectl apply -f "./_out/manifests/uploadproxy-nodeport.yaml"
 
-  # Check if CPU architecture is s390x
-  if [ "$(uname -m)" != "s390x" ]; then
+  # Disable unsupported functest images for s390x
+  if [ "${ARCHITECTURE}" != "s390x" ]; then
     # Imageio test service:
     _kubectl apply -f "./_out/manifests/imageio.yaml"
     # vCenter (VDDK) test service:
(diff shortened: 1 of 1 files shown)


akalenyu:
This is just changing the sync script, which basically deploys the local CDI into the dev envirnment cluster. Are you sure this is what you're after?

Davo911:
Hey @akalenyu thanks for looking into this,

my Issue is, that for the functional tests `make cluster-sync` deploys the func-test images onto the cluster and as expected fails with two related to `vddk` and `imageio`. As these never become ready [the func-tests won't run](https://github.com/kubevirt/containerized-data-importer/blob/75f00e2f149ab6ac64510c70fc081d76f5309a08/hack/build/run-functional-tests.sh#L77-L94), at least for s390x

cluster-sync/sync.sh
```diff
@@ -180,10 +180,15 @@ if [ "${CDI_SYNC}" == "test-infra" ]; then
   _kubectl apply -f "./_out/manifests/test-proxy.yaml"
   _kubectl apply -f "./_out/manifests/sample-populator.yaml"
   _kubectl apply -f "./_out/manifests/uploadproxy-nodeport.yaml"
-  # Imageio test service:
-  _kubectl apply -f "./_out/manifests/imageio.yaml"
-  # vCenter (VDDK) test service:
-  _kubectl apply -f "./_out/manifests/vcenter.yaml"
+
+  # Check if CPU architecture is s390x
+  if [ "$(uname -m)" != "s390x" ]; then
```
akalenyu, line 185: So this would work if the host running the development env setup script is on s390x, does that cut it?
What if the developer is on x86 deploying custom CDI to a remote cluster? Maybe there's a better env var to use here for the s390x target cluster check?

akalenyu:
> Hey @akalenyu thanks for looking into this,
> 
> my Issue is, that for the functional tests `make cluster-sync` deploys the func-test images onto the cluster and as expected fails with two related to `vddk` and `imageio`. As these never become ready [the func-tests won't run](https://github.com/kubevirt/containerized-data-importer/blob/75f00e2f149ab6ac64510c70fc081d76f5309a08/hack/build/run-functional-tests.sh#L77-L94), at least for s390x

Gotcha, one more question, just poking at this a bit before we merge

PR #3465, opened 2024-10-17
Add default containerPort for cdi-apiserver deployment

scoquelin:

We have a use case where we run `kubevirt/cdi` with `HostNetwork=true`

This PR is just adding `containerPort` explicitly when creating the deployment for `cdi-apiserver` so that Kubernetes can pro-actively detect if a pod needs to be assigned to other nodes in case the port is already taken.

**Release note**:
```release-note
Explicitly declaring containerPort in cdi-apiserver deployment
```


mhenriks:
I'm fine with this

/lgtm

scoquelin:
Thank you all! 🥳

PR #3479, opened 2024-10-25
Handle potential nil pointer in handleSnapshot

awels:
If the dataimportcron status has an empty currentImports, the returned PVC will be nil and
handleSnapshot will be called. handleSnapshot should check for nil PVC before proceeding.

This can happen when installing CDI and KubeVirt together with ceph on a cloud platform. The cloud platform will have a default storage class, which causes the imports to start. Then ceph becomes ready and switches the default virt class to itself. This then causes the imports to switch and somehow the currentImports is empty which causes the PVC to be nil and that causes a 
nil pointer exception in handleSnapshots.

Fixes #

**Release note**:

```release-note
BugFix: nil pointer in handleSnapshot
```


ShellyKa13 (commented):
/lgtm

PR #3489, opened 2024-10-31
cdi-operator updating cdi-apiserver all the time

mhenriks:
Make sure TerminationMessagePolicy/Path are set when creating cdi-apiserver

Fixing regression introduced in #3465

Fixes #

**Release note**:

```release-note
Make sure TerminationMessagePolicy/Path are set when creating cdi-apiserver
```


akalenyu (commented):
/approve

Nice catch, how does the error look like? I imagine lots of these messages? https://github.com/kubevirt/containerized-data-importer/blob/ff403fd5adfbb73e6624d03ba8fd0e3e1e214599/vendor/kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/util.go#L156 Should we try to write a functest for this? (could be outside this PR)

PR #3500, opened 2024-11-04
Populators: Add to PVC' exclude velero backup annotation

ShellyKa13:
When backing up with Velero if a PVC is terminating then the backup is marked as PartiallyFailed, while the backup is still valid to use and its more of a warning, PVC' should be transparent to the user so have the backup partiallyFailed because of PVC' is a bad user experience. To fix that added to all our PVC' an annotation so it will be excluded from the Velero backup.

```release-note
NONE
```


awels (commented):
/approve

PR #3520, opened 2024-11-07
Change clone strategy to avoid PowerStore snapshot limits

coulof:
PowerStore has a limit of 32 to 64 writable snapshots which prevents to create more VMs from a base OS.

Fixes #

**Release note**:

```release-note
* Dell PowerStore uses `cloneStrategy: copy` over `csi-clone` to avoid array snapshot limits
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -139,7 +139,7 @@ var CloneStrategyByProvisionerKey = map[string]cdiv1.CDICloneStrategy{
 	"csi-vxflexos.dellemc.com":                 cdiv1.CloneStrategyCsiClone,
 	"csi-isilon.dellemc.com":                   cdiv1.CloneStrategyCsiClone,
 	"csi-powermax.dellemc.com":                 cdiv1.CloneStrategyCsiClone,
-	"csi-powerstore.dellemc.com":               cdiv1.CloneStrategyCsiClone,
+	"csi-powerstore.dellemc.com":               cdiv1.CloneStrategyHostAssisted,
 	"hspc.csi.hitachi.com":                     cdiv1.CloneStrategyCsiClone,
 	"csi.hpe.com":                              cdiv1.CloneStrategyCsiClone,
 	"spectrumscale.csi.ibm.com":                cdiv1.CloneStrategyCsiClone,


akalenyu:
So CSI/snapshot clones normally create ephemeral snapshots, which should keep us within limits.
Could you please attach some PowerStore source? I am probably misunderstanding the issue.

I want to ensure I poke at this a bit, since doing host assisted clones all over is going to be a major hit for anyone who uses this solution.

coulof:
> So CSI/snapshot clones normally create ephemeral snapshots, which should keep us within limits. Could you please attach some PowerStore source? I am probably misunderstanding the issue.
> 
> I want to ensure I poke at this a bit, since doing host assisted clones all over is going to be a major hit for anyone who uses this solution.

In PowerStore, a PVC restore from a Snapshot is actually a writable Snapshot. There are limited number of writable snapshot per LUN therefore we have to switch to host copy.

akalenyu:
> > So CSI/snapshot clones normally create ephemeral snapshots, which should keep us within limits. Could you please attach some PowerStore source? I am probably misunderstanding the issue.
> > I want to ensure I poke at this a bit, since doing host assisted clones all over is going to be a major hit for anyone who uses this solution.
> 
> In PowerStore, a PVC restore from a Snapshot is actually a writable Snapshot. There are limited number of writable snapshot per LUN therefore we have to switch to host copy.

In that case, you also want to not use snapshots as golden sources (many VMs will restore from them), right https://github.com/kubevirt/containerized-data-importer/blob/1360dfd9a6f10e15c014a80ad510188cbe171b23/doc/onboarding-storage-provisioners.md?plain=1#L18-L19

But my worries remain, host-assisted is a big hammer,
But other things could break too, for example, you're saying the number of VMSnapshot/VMRestores one can make is also limited?

coulof:
@akalenyu, per documentation [here](https://elabnavigator.dell.com/vault/pdf/PowerStore_ESSM.pdf) & [there](https://www.delltechnologies.com/asset/en-us/products/storage/industry-market/h18156-dell-powerstore-snapshots-and-thin-clones.pdf)

> A thin clone is a read-write copy of a volume, volume group, or snapshot that shares blocks with the parent resource.... Thin clones are not full copies of the original source.

> There is limit of 64 Max Clones per Family (A Family is a volume, volume group, or base storage container and all its derivative thin clones and snapshots).

As you can imagine you can quickly reach 64 VMs starting from a Golden image.

My understanding was that with the updated `StorageProfile`, the Golden image won't be restored from a snapshot but from a host copy.

We are investigating to add full-blown detached copy to PowerStore. But in the meantime it is probably wise to not rely on PVC Clone.

akalenyu:
> @akalenyu, per documentation [here](https://elabnavigator.dell.com/vault/pdf/PowerStore_ESSM.pdf) & [there](https://www.delltechnologies.com/asset/en-us/products/storage/industry-market/h18156-dell-powerstore-snapshots-and-thin-clones.pdf)
> 
> > A thin clone is a read-write copy of a volume, volume group, or snapshot that shares blocks with the parent resource.... Thin clones are not full copies of the original source.
> 
> > There is limit of 64 Max Clones per Family (A Family is a volume, volume group, or base storage container and all its derivative thin clones and snapshots).
> 
> As you can imagine you can quickly reach 64 VMs starting from a Golden image.
> 
> My understanding was that with the updated `StorageProfile`, the Golden image won't be restored from a snapshot but from a host copy.
> 
> We are investigating to add full-blown detached copy to PowerStore. But in the meantime it is probably wise to not rely on PVC Clone.

Hey, I totally get the reasoning, I am just saying that the same problem could meet you further down the road
(vmsnapshot/vmrestore https://kubevirt.io/user-guide/storage/snapshot_restore_api/)

Anyhow, this decision is entirely up to you, but, for visibility I would appreciate an lgtm from
someone else here too
/approve
/cc @aglitke @peterclauterbach

PR #3534, opened 2024-11-20
CNV-51521 | Fix Incomplete transfer of change blocks leads to data corruption

mnecas:
**Fix**: Query the changed areas by every 2000 block chunks, combine the chunks together and request the next QueryChangedDiskAreas to start from the end of the last block. Once the QueryChangedDiskAreas returns less than 2000 blocks we know that it was the last chunk and can continue with the disk import.

**Issue**: When the user has a large disk with over 2000 changed blocks the QueryChangedDiskAreas returns only first 2000 changed blocks. The disk is not getting all the changes from the VMware Change Block Tracking and this can cause disk corruption.

Fixes https://issues.redhat.com/browse/CNV-51521

**Release note**:

```release-note
VDDK Fix Incomplete transfer of change blocks leads to data corruption
```

### DIFF
--- a/pkg/importer/vddk-datasource_amd64.go
+++ b/pkg/importer/vddk-datasource_amd64.go
@@ -855,15 +855,17 @@ func createVddkDataSource(endpoint string, accessKey string, secKey string, thum
 	// then get the list of changed blocks from VMware for a delta copy.
 	var changed *types.DiskChangeInfo
 	if currentSnapshot != nil && previousCheckpoint != "" {
-		// The QueryChangedDiskAreas returns maximally 2000 changed blocks. If the disk has more than 2000 blocks
-		// we need to query the next chunk of the blocks starting from previous
-		var lastChangedSize = 0
+		// QueryChangedDiskAreas needs to be called multiple times to get all possible disk changes.
+		// Experimentation shows it returns maximally 2000 changed blocks. If the disk has more than
+		// 2000 changed blocks we need to query the next chunk of the blocks starting from previous.
+		// Loop until QueryChangedDiskAreas starts returning zero-length block lists.
 		changed = &types.DiskChangeInfo{}
 		// Check if this is a snapshot or a change ID, and query disk areas as appropriate.
 		// Change IDs look like: 52 de c0 d9 b9 43 9d 10-61 d5 4c 1b e9 7b 65 63/81
 		changeIDPattern := `([0-9a-fA-F]{2}\s?)*-([0-9a-fA-F]{2}\s?)*\/([0-9a-fA-F]*)`
 		if matched, _ := regexp.MatchString(changeIDPattern, previousCheckpoint); matched {
-			for changed.Length == 0 || lastChangedSize != 0 {
+			for {
+				klog.Infof("Querying changed disk areas at offset %d", changed.Length)
 				request := types.QueryChangedDiskAreas{
 					ChangeId:    previousCheckpoint,
 					DeviceKey:   backingFileObject.Key,
@@ -873,9 +875,13 @@ func createVddkDataSource(endpoint string, accessKey string, secKey string, thum
 				}
 				response, err := QueryChangedDiskAreas(vmware.context, vmware.vm.Client(), &request)
 				if err != nil {
+					klog.Errorf("Failed to query changed areas: %s", err)
 					return nil, err
 				}
-				lastChangedSize = len(response.Returnval.ChangedArea)
+				klog.Infof("%d changed areas reported at offset %d", len(response.Returnval.ChangedArea), changed.Length)
+				if len(response.Returnval.ChangedArea) == 0 { // No more changes
+					break
+				}
 				changed.ChangedArea = append(changed.ChangedArea, response.Returnval.ChangedArea...)
 				changed.Length += response.Returnval.Length
 			}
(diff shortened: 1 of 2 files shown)


akalenyu:
Some boring linter errors
```
+ golangci-lint run ./...
pkg/importer/vddk-datasource_amd64.go:861:27: var-declaration: should drop = 0 from declaration of var startOffset; it is the zero value (revive)
		var startOffset int64 = 0
		                        ^
pkg/importer/vddk-datasource_test.go:799:4: increment-decrement: should replace counter += 1 with counter++ (revive)
			counter += 1
			^
pkg/importer/vddk-datasource_amd64.go:868:21: SA6000: calling regexp.MatchString in a loop has poor performance, consider using regexp.Compile (staticcheck)
			if matched, _ := regexp.MatchString(changeIDPattern, previousCheckpoint); matched {
			                 ^
make: *** [Makefile:199: test-lint] Error 1
```

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -855,35 +855,47 @@ func createVddkDataSource(endpoint string, accessKey string, secKey string, thum
...
-			}
-			if previousSnapshot != nil {
-				changedAreas, err := vmware.vm.QueryChangedDiskAreas(vmware.context, previousSnapshot, currentSnapshot, backingFileObject, 0)
+		// The QueryChangedDiskAreas returns maximally 2000 changed blocks. If the disk has more than 2000 blocks
+		// we need to query the next chunk of the blocks starting from previous
+		const MaximumQueryResponse = 2000
+		var startOffset int64 = 0
+		var lastChangedSize = 0
+		changed = &types.DiskChangeInfo{}
+		for changed.Length == 0 || lastChangedSize == MaximumQueryResponse {
```
mrnold, line 864: Can you make this two smaller loops instead? Like just around the actual QueryChangedDiskAreas calls. That will clean up one of the lint errors about MatchString and avoid repeated calls to FindSnapshot, which can be expensive.

pkg/importer/vddk-datasource_test.go
```diff
@@ -724,6 +724,86 @@ var _ = Describe("VDDK get block status", func() {
 			}
 		}
 	})
+	It("VDDK return more than 2000 changed block area", func() {
```
akalenyu, line 727: You probably want to move this test under the other describe (`Describe("VDDK data source", func() {`), otherwise, you don't have the client creation mocked https://github.com/kubevirt/containerized-data-importer/blob/d24d618956524959f444d23ddf61796491454f97/pkg/importer/vddk-datasource_test.go#L53-L64

mnecas: Thanks fixed!

mrnold (approved):

mrnold (approved):

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -855,35 +855,47 @@ func createVddkDataSource(endpoint string, accessKey string, secKey string, thum
...
+					Snapshot:    currentSnapshot,
+					StartOffset: changed.Length,
+					This:        vmware.vm.Reference(),
+				}
+				response, err := QueryChangedDiskAreas(vmware.context, vmware.vm.Client(), &request)
+				if err != nil {
+					return nil, err
+				}
+				lastChangedSize = len(response.Returnval.ChangedArea)
+				changed.ChangedArea = append(changed.ChangedArea, response.Returnval.ChangedArea...)
```
mhenriks, line 879: Are we not concerned about fetching all this data upfront and storing it in memory? What is the granularity of the change intervals?

mrnold: I think it's a concern, but not enough of one to rework this whole thing for this bug. I don't know exactly what the granularity is, but I doubt it is finer than the disk sector size. The structures in this list should be two int64s for starting offset and entry count, with a sub-list of change structures with two more int64s for starting offset and length. I'm sure there's a few more ints worth of overhead, but it's not huge amounts per entry.

We saw a single QueryChangedDiskAreas call hit 2000 entries in one request, so I would spitball something like 1MB per call? And it should only take a few calls to get all the changes. I'm sure we can find a pathological worst case to blow the number up, but probably most of the time more changes than this get rolled up into one bigger block.

mhenriks: Would it be possible to get more data on this? Do a migration that was previously failing and log how many records and maybe average block size?

mrnold: Sure, we can probably even get the actual memory usage.

mhenriks: I don't think it would be wise to let this grow unbounded, importer pods don't have a lot of memory

mhenriks:
Memory issue to be addressed in subsequent PR
/lgtm

akalenyu:
/test pull-cdi-unit-test
unrelated 
```
Process Limits exec [It] command success with real limits
```
https://github.com/kubevirt/containerized-data-importer/issues/3341#issuecomment-2316174754

akalenyu:
/hold
fails on real env, we're looking into it

akalenyu:
ignore, some fluke in the generate script. we don't see this in the others

PR #3544, opened 2024-11-27
Fix PVC webhook rendering segfault

arnongilboa:
When there is no default `StorageClass` and no virt default `StorageClass`, creating a `PVC` with no `StorageClass` specified will cause a segfault in the webhook:
```
Error from server (InternalError): error when creating
"pvc_render.yaml": Internal error occurred: failed calling webhook
"pvc-mutate.cdi.kubevirt.io": failed to call webhook: Post
"[https://cdi-api.cdi.svc:443/pvc-mutate?timeout=10s](https://cdi-api.cdi.svc/pvc-mutate?timeout=10s)": EOF
```
In cdi-apiserver log:
```
panic({0x1b49360?, 0x3295550?})
        GOROOT/src/runtime/panic.go:770 +0x132
kubevirt.io/containerized-data-importer/pkg/controller/datavolume.
renderPvcSpecVolumeModeAndAccessModesAndStorageClass({0x2164d00,
0xc000156090}, {0x0, 0x0}, 0x0?, 0x0, 0xc0002e86a8, {0x1e09ed8?,
0xc0002e8530?})
        pkg/controller/datavolume/util.go:162 +0xb37
```
`renderPvcSpecVolumeModeAndAccessModesAndStorageClass()` is used from from both DV controller and PVC webhook rendering. In the PVC mutating webhook (`cdi-apiserver`) we don't have `logr.Logger` (unlike in `cdi-deployment`), but `klog` which has different api, and no reason to log there as we return the failure anyway to the user when she creates the PVC. We also don't have a DV there, and no event recorder.

jira-ticket: https://issues.redhat.com/browse/CNV-51910

**Release note**:
```release-note
Fix PVC webhook rendering segfault when there is no default default storage class and no virtualization default storage class
```

### DIFF
--- a/pkg/apiserver/webhooks/BUILD.bazel
+++ b/pkg/apiserver/webhooks/BUILD.bazel
@@ -71,6 +71,7 @@ go_test(
         "//vendor/k8s.io/api/admission/v1:go_default_library",
         "//vendor/k8s.io/api/authorization/v1:go_default_library",
         "//vendor/k8s.io/api/core/v1:go_default_library",
+        "//vendor/k8s.io/api/storage/v1:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/api/resource:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/runtime:go_default_library",
@@ -78,6 +79,7 @@ go_test(
         "//vendor/k8s.io/client-go/testing:go_default_library",
         "//vendor/k8s.io/utils/ptr:go_default_library",
         "//vendor/kubevirt.io/controller-lifecycle-operator-sdk/api:go_default_library",
+        "//vendor/sigs.k8s.io/controller-runtime/pkg/client:go_default_library",
         "//vendor/sigs.k8s.io/controller-runtime/pkg/client/fake:go_default_library",
     ],
 )
--- a/pkg/apiserver/webhooks/BUILD.bazel
+++ b/pkg/apiserver/webhooks/BUILD.bazel
@@ -55,6 +55,7 @@ go_test(
         "datavolume-mutate_test.go",
         "datavolume-validate_test.go",
         "populators-validate_test.go",
+        "pvc-mutate_test.go",
         "transfer-validate_test.go",
         "webhook_suite_test.go",
     ],
--- a/pkg/apiserver/webhooks/pvc-mutate_test.go
+++ b/pkg/apiserver/webhooks/pvc-mutate_test.go
@@ -1,0 +1,185 @@
+/*
+ * This file is part of the CDI project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


pkg/controller/datavolume/util.go
```diff
@@ -129,21 +128,22 @@ func pvcFromStorage(client client.Client, recorder record.EventRecorder, log log
 func renderPvcSpecVolumeModeAndAccessModesAndStorageClass(client client.Client, recorder record.EventRecorder, log *logr.Logger,
 	dv *cdiv1.DataVolume, pvcSpec *v1.PersistentVolumeClaimSpec, dvContentType cdiv1.DataVolumeContentType) error {
 	logInfo := func(msg string, keysAndValues ...interface{}) {
-		if log != nil {
-			log.V(1).Info(msg, keysAndValues...)
+		if log != nil && dv != nil {
```
ShellyKa13, line 131: not sure I get this changes.. In some places you are not logging anything if dv doesnt exists and in others you are removing any information related to dv which can be useful? seems odd

arnongilboa: In the PVC webhook (`cdi-apiserver`) we don't have `logr.Logger` (unlike in `cdi-deployment`), but `klog` which has different api, and I see no reason to log there as we return the failure anyway to the user when she creates the PVC. Clearly, we don't have a DV there (it's a PVC webhook...), and no event recorder. However I think we should keep using `renderPvcSpecVolumeModeAndAccessModesAndStorageClass()` (don't complain about the name, it's Michael's afair) func from both DV controller and PVC webhook rendering. wdyt?

arnongilboa: btw, it's still wip as I need to better explain the issue and add utests.

ShellyKa13: Oh I missed the fact that all this is inside functions so the dv data is kept. So ok looks good :) If you would have explained it in the commit message it would have been clearer

pkg/apiserver/webhooks/pvc-mutate_test.go
```diff
@@ -0,0 +1,182 @@
...
+			defaultStorageClass = storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "sc",
+					Annotations: map[string]string{
+						"storageclass.kubernetes.io/is-default-class": "true",
+					},
+				},
+			}
+			storageProfile = cdiv1.StorageProfile{
+				ObjectMeta: metav1.ObjectMeta{Name: "sc"},
```
ShellyKa13, line 70: nit : use a const for storage class name "sc"

pkg/apiserver/webhooks/BUILD.bazel
```diff
@@ -71,13 +71,15 @@ go_test(
         "//vendor/k8s.io/api/admission/v1:go_default_library",
```
ShellyKa13, line 72: Will appreciate the description of the PR as a commit message here :)

ShellyKa13 (commented):
Looks good, two small nits

pkg/controller/datavolume/util.go
```diff
@@ -129,21 +128,22 @@ func pvcFromStorage(client client.Client, recorder record.EventRecorder, log log
...
+			keysAndValuesWithDv := append(keysAndValues, "namespace", dv.Namespace, "name", dv.Name)
+			log.V(1).Info(msg, keysAndValuesWithDv...)
 		}
 	}
 
-	recordEventf := func(object runtime.Object, eventtype, reason, messageFmt string, args ...interface{}) {
-		if recorder != nil {
-			recorder.Eventf(object, eventtype, reason, messageFmt, args...)
+	recordEventf := func(eventtype, reason, messageFmt string, args ...interface{}) {
+		if recorder != nil && dv != nil {
```
awels, line 138: Would it make sense here to write this event to the PVC instead in the case there is no DV? Right now the event is not associated with any resource and it ends up in the general events list.

PR #3552, opened 2024-12-02
Deprecate DataVolume garbage collection

arnongilboa:
After several releases with `DataVolume` garbage collection disabled by default, we decided to deprecate it, as unfortunately it violates fundamental principle of Kubernetes. CR should not be auto-deleted when it completes its role (`Job` with `TTLSecondsAfterFinished` is an exception), and once CR was created we can assume it is there until explicitly deleted. In addition, CR should keep idempotency, so the same CR manifest can be applied multiple times, as long as it is a valid update (e.g. `DataVolume` validation webhook does not allow updating the spec).

jira-ticket: https://issues.redhat.com/browse/CNV-30806

**Release note**:
```release-note
Deprecate DataVolume garbage collection
```


akalenyu:
@fabiand @mnecas @mrnold do you know if any forklift related project is using dv garbage collection?

mnecas:
Forklfit uses the annotation but we always disable the GC so we should be good
https://github.com/kubev2v/forklift/blob/87e3925da8cea784c7698e82afe729dbdf7e29eb/pkg/controller/plan/kubevirt.go#L1241

mnecas:
The forklfit disables the GC [1]. 
I don't want to remove it from Forklift yet for backwards compatibility.
And having additional annotation even tho it's not used should not harm the users
[1] https://github.com/kubev2v/forklift/blob/87e3925da8cea784c7698e82afe729dbdf7e29eb/pkg/controller/plan/kubevirt.go#L1241

arnongilboa:
@ksimon1 @0xFelix does the deprecation require any fix in kubevirt-tekton-tasks? I can't forget this tiny [adjustment PR](https://github.com/kubevirt/kubevirt-tekton-tasks/pull/204#event-8353528036).

0xFelix:
@arnongilboa kubevirt-tekton-tasks do not assume that garbage collection is active by default AFAIK. I created an issue to keep track of it.

https://github.com/kubevirt/kubevirt-tekton-tasks/issues/566

PR #3553, opened 2024-12-03
Move size detection pod deletion to cleanup

akalenyu:
Apparently, it's possible that the size detection pod
sticks around even if the dv is succeeded/deleted.
I think this belongs in the cleanup loop anyway

Fixes #
https://issues.redhat.com/browse/CNV-52272

**Release note**:

```release-note
BugFix: size-detection pod is not always cleaned after it completes
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -114,6 +114,9 @@ make cluster-sync
 
 kubectl version
 
+echo "Nil check --PRE-- test run"
+kubectl get pods -n $CDI_NAMESPACE -o'custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace,RESTARTS:status.containerStatuses[*].restartCount' --no-headers
+
 ginko_params="--test-args=--ginkgo.no-color --ginkgo.junit-report=${ARTIFACTS_PATH}/junit.functest.xml"
 
 if [[ -n "$CDI_DV_GC" ]]; then
@@ -122,3 +125,7 @@ fi
 
 # Run functional tests
 TEST_ARGS=$ginko_params make test-functional
+
+echo "Nil check --POST-- test run"
+kubectl get pods -n $CDI_NAMESPACE -o'custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace,RESTARTS:status.containerStatuses[*].restartCount' --no-headers
+kubectl logs -n $CDI_NAMESPACE $(kubectl get pod -n $CDI_NAMESPACE -l=cdi.kubevirt.io=cdi-deployment --output=jsonpath='{$.items[0].metadata.name}') --previous || echo "this is fine"
--- a/pkg/controller/datavolume/clone-controller-base.go
+++ b/pkg/controller/datavolume/clone-controller-base.go
@@ -583,3 +583,50 @@ func addCloneWithoutSourceWatch(mgr manager.Manager, datavolumeController contro
 
 	return nil
 }
+
+func addDataSourceWatch(mgr manager.Manager, c controller.Controller, indexingKey string, op dataVolumeOp) error {
+	getKey := func(namespace, name string) string {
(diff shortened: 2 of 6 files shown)


automation/test.sh
```diff
@@ -122,3 +125,7 @@ fi
 
 # Run functional tests
 TEST_ARGS=$ginko_params make test-functional
+
+echo "Nil check --POST-- test run"
```
alromeros, line 129: Nice :+1:

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -132,6 +132,10 @@ func (r *SnapshotCloneReconciler) addDataVolumeSnapshotCloneControllerWatches(mg
 		return err
 	}
 
+	if err := addDataSourceWatch(mgr, datavolumeController, "snapshotdatasource", dataVolumeSnapshotClone); err != nil {
```
alromeros, line 135: We should add some test coverage and ensure the detection pod works with the snapshot source. There might be something missing in this flow as the size-detection logic was implemented before this.

pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -632,8 +655,8 @@ func (r *PvcCloneReconciler) getSizeFromPod(targetPvc, sourcePvc *corev1.Persist
 	if err := r.updateClonePVCAnnotations(sourcePvc, termMsg); err != nil {
 		return imgSize, err
 	}
-	// Finally, detelete the pod
-	if cc.ShouldDeletePod(sourcePvc) {
+	// Finally, delete the pod
+	if targetPvc != nil && cc.ShouldDeletePod(targetPvc) {
```
alromeros, line 659: Nice catch

pkg/controller/datavolume/snapshot-clone-controller.go
alromeros: Aw, cool. How come this worked when it was only implemented in the pvc source section?

akalenyu: This watch is about data source changes, so not about size detection. Size detection for snap sources has been a thing for ages now

alromeros:
Good work!
/lgtm

pkg/controller/datavolume/clone-controller-base.go
```diff
@@ -583,3 +583,50 @@ func addCloneWithoutSourceWatch(mgr manager.Manager, datavolumeController contro
...
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	mapToDataVolume := func(ctx context.Context, obj *cdiv1.DataSource) []reconcile.Request {
+		var dvs cdiv1.DataVolumeList
+		matchingFields := client.MatchingFields{indexingKey: getKey(obj.GetNamespace(), obj.GetName())}
```
awels, line 610: Maybe instead of the getKey function use this as in other places in this file:

```go
namespacedName := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
matchingFields := client.MatchingFields{indexingKey: namespacedName.String()}
```

I just don't like defining an extra getKey function if it already exists. But not a huge deal since you just moved the entire function as it existed already. Up to you if you want to ignore.

awels (approved):

pkg/controller/datavolume/clone-controller-base.go
akalenyu: Yeah maybe not for this PR, there's some occurrences of this and we can tackle them all later

awels: Yeah not a big deal, and since you are not making a net new one, I am fine with this.

akalenyu:
looks unrelated but lets monitor

PR #3564, opened 2024-12-11
CWE-79 small tightening up

akalenyu:
The uploadReady func **could** be returning an error that's somehow dependent on user input.
It's somehow unlikely and an edge case but better safe than sorry

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/uploadproxy/uploadproxy.go
+++ b/pkg/uploadproxy/uploadproxy.go
@@ -217,7 +217,7 @@ func (app *uploadProxyApp) handleUploadRequest(w http.ResponseWriter, r *http.Re
 		klog.Error(err)
 		w.WriteHeader(http.StatusServiceUnavailable)
 		// Return the error to the caller in the body.
-		_, err = fmt.Fprint(w, err.Error())
+		_, err = fmt.Fprint(w, html.EscapeString(err.Error()))
 		if err != nil {
 			klog.Errorf("handleUploadRequest: failed to send error response: %v", err)
 		}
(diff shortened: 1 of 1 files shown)


awels:
looks like this change is affecting a particular XSS test?

pkg/uploadproxy/uploadproxy.go
```diff
@@ -252,8 +252,8 @@ func (app *uploadProxyApp) resolveUploadPath(pvc *v1.PersistentVolumeClaim, pvcN
 			path = common.UploadArchivePath
 		}
 	default:
-		// Escaping user-controlled strings to avoid cross-site scripting (XSS) attacks
-		return "", fmt.Errorf("rejecting upload request for PVC %s - upload content-type %s is invalid", html.EscapeString(pvcName), html.EscapeString(contentType))
+		// Caller is escaping user-controlled strings to avoid cross-site scripting (XSS) attacks
```
alromeros, line 255: Nit: Since we are keeping this comment it may make more sense to move it where we are escaping the strings.

akalenyu: I changed it a bit with the intention to keep it here, but I don't mind. Did you catch the wording change?

alromeros: Waiting for the tests, I'll lgtm once they pass

akalenyu: Yeah this time I ran them locally :stuck_out_tongue:

PR #3565, opened 2024-12-12
Add vddk label to a func test 

Davo911:
For the functional tests under s390x we filter out tests using unsupported tools. 
For this to work correctly, that tool needs to be mentioned in the description of the test or have a corresponding label assigned.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -157,7 +157,7 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		Expect(importer.DeletionTimestamp).To(BeNil())
 	})
 
-	It("[test_id:6688] [VDDK] Should retain all multi-stage importer pods after completion with dv annotation cdi.kubevirt.io/storage.pod.retainAfterCompletion=true", func() {
+	It("[test_id:6688] Should retain all multi-stage importer pods after completion with dv annotation cdi.kubevirt.io/storage.pod.retainAfterCompletion=true", Label("VDDK"), func() {
 		vcenterURL := fmt.Sprintf(utils.VcenterURL, f.CdiInstallNs)
 		dataVolume := f.CreateVddkWarmImportDataVolume("import-pod-retain-test", "100Mi", vcenterURL)
 
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -157,7 +157,7 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		Expect(importer.DeletionTimestamp).To(BeNil())
 	})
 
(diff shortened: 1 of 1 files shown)


tests/import_test.go
```diff
@@ -157,7 +157,7 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 		Expect(importer.DeletionTimestamp).To(BeNil())
 	})
 
-	It("[test_id:6688] Should retain all multi-stage importer pods after completion with dv annotation cdi.kubevirt.io/storage.pod.retainAfterCompletion=true", func() {
+	It("[test_id:6688] Should retain all multi-stage importer pods after completion with vddk dv annotation cdi.kubevirt.io/storage.pod.retainAfterCompletion=true", func() {
```
akalenyu, line 160: I think something like labels would work a lot better for your use case, and less likely to break:
https://github.com/kubevirt/kubevirt/blob/main/tests/decorators/decorators.go#L21

Davo911: Something we also thought about, but as it only affected a single test, this seemed like a less cluttery solution.
A label like this would be fine?
```suggestion
	It("[test_id:6688] [VDDK] Should retain all multi-stage importer pods after completion with dv annotation cdi.kubevirt.io/storage.pod.retainAfterCompletion=true", func() {
```

akalenyu: No, I meant like a real ginkgo label, if you look up the one I linked through the kubevirt test project, you'll see what I mean

Davo911: Ah missed your link, sorry.
Ok so you'd like to introduce labeling mechanism in the same fashion for cdi, as it is not used yet here?

cfilleke: I vote for real ginkgo labels & documented, ultimately.  e.g. [arch:s390x], [arch:arm64] with the default being understood to be [arm:amd64] so it's clearer to people maintaining this code what is going on.  

It'd be a more verbose PR later, which would not conflict with this change at all.

akalenyu: These cleartext additions have been a pain in kubevirt, they ended up being duplicated in the test name and labels. I would prefer something like Label("VDDK") (don't have to go further than that)

Davo911: Alright
I'll enable filtering in a separate PR then

akalenyu:
> @Davo911: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-cdi-verify-go-mod	[96ba51a](https://github.com/kubevirt/containerized-data-importer/commit/96ba51a50206e11269295e605aa51c771aa38264)	[link](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3565/pull-cdi-verify-go-mod/1868299441077751808)	true	`/test pull-cdi-verify-go-mod`

@brianmcarey hey, I am seeing this kind of failure quite a bit lately. a simple dep update lane times out after an hour. Are you familiar with something like this?

PR #3566, opened 2024-12-13
Add filter label parameter for functional tests

Davo911:
This change adds a label-filter parameter to the functional test script.

This allows users to specify a label filter via the `CDI_LABEL_FILTER` environment variable, enabling better test selection without relying on hardcoded labels in the test descriptions. 
Discussed in https://github.com/kubevirt/containerized-data-importer/pull/3565 and makes the introduced label usable. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/run-functional-tests.sh
+++ b/hack/build/run-functional-tests.sh
@@ -95,14 +95,10 @@ fi
 
 (
     export TESTS_WORKDIR=${CDI_DIR}/tests
-    ginkgo_args="--trace --timeout=8h --v"
-
-    if [[ -n "$CDI_E2E_SKIP" ]]; then
-        ginkgo_args="${ginkgo_args} --skip=${CDI_E2E_SKIP}"
-    fi
+    ginkgo_args="--trace --timeout=8h --v --skip=${CDI_E2E_SKIP} --label-filter=${CDI_LABEL_FILTER}"
 
     if [[ -n "$CDI_E2E_FOCUS" || -n "$CDI_E2E_SKIP" ]]; then
-            ginkgo_args="${ginkgo_args} --nodes=6"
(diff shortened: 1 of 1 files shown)


hack/build/run-functional-tests.sh
```diff
@@ -95,14 +95,10 @@ fi
 
 (
     export TESTS_WORKDIR=${CDI_DIR}/tests
-    ginkgo_args="--trace --timeout=8h --v"
-
-    if [[ -n "$CDI_E2E_SKIP" ]]; then
-        ginkgo_args="${ginkgo_args} --skip=${CDI_E2E_SKIP}"
-    fi
+    ginkgo_args="--trace --timeout=8h --v --skip=${CDI_E2E_SKIP} --label-filter=${CDI_LABEL_FILTER}"
```
akalenyu, line 98: Are we sure that an empty skip value will not mess things up?

akalenyu (commented):
Looks good! just worried this could mess up the existing test set in the non destructive lane

hack/build/run-functional-tests.sh
Davo911: I tried it on my side without issues, but decided for this out of pure aesthetic reasons.
On a second thought being explicit may be the better choice

akalenyu: TBH I took a look at the test set that was exercised, everything looks the same, so it's fine

akalenyu: Actually I think adding these variables conditionally is cleaner (skip/filter)

Davo911: agree!

akalenyu:
/lgtm
/approve
/hold

I will unhold once I the CI is finished, want to ensure the test set remains the same

PR #3572, opened 2024-12-16
CNV-52722: Pass through extra VDDK configuration options to importer pod.

mrnold:
This pull request adds a new annotation "cdi.kubevirt.io/storage.pod.vddk.extraargs", referencing a ConfigMap that contains extra parameters to pass directly to the VDDK library. The use case is to allow tuning of asynchronous buffer counts for MTV as requested in [CNV-52722](https://issues.redhat.com/browse/CNV-52722). Testing has shown good results for cold migrations with:
```
VixDiskLib.nfcAio.Session.BufSizeIn64KB=16
VixDiskLib.nfcAio.Session.BufCount=4
```

These parameters are stored in a file whose path is passed to the VDDK via the nbdkit "config=" option. The file contents come from the referenced ConfigMap, and the ConfigMap is mounted to the importer pod as a volume.

Fixes [CNV-52722](https://issues.redhat.com/browse/CNV-52722)

As far as I can tell, a ConfigMap volume mount must be in the same namespace as the importer pod. So MTV will need to create or duplicate the ConfigMap to the same namespace as the DataVolume it creates.

**Release note**:
```release-note
Allow extra VDDK configuration parameters to be passed to VDDK importer pods.
```

### DIFF
--- a/doc/datavolumes.md
+++ b/doc/datavolumes.md
@@ -349,6 +349,13 @@ spec:
 [Get VDDK ConfigMap example](../manifests/example/vddk-configmap.yaml)
 [Ways to find thumbprint](https://libguestfs.org/nbdkit-vddk-plugin.1.html#THUMBPRINTS)
 
+#### Extra VDDK Configuration Options
+
+The VDDK library itself looks in a configuration file (such as `/etc/vmware/config`) for extra options to fine tune data transfers. To pass these options through to the VDDK, store the configuration file contents in a ConfigMap and add a `cdi.kubevirt.io/storage.pod.vddk.extraargs` annotation to the DataVolume specification. The ConfigMap will be mounted to the importer pod as a volume, and the first file in the mounted directory will be passed to the VDDK. This means that the ConfigMap must be placed in the same namespace as the DataVolume, and the ConfigMap should only have one file entry.
+
+[Example annotation](../manifests/example/vddk-args-annotation.yaml)
+[Example ConfigMap](../manifests/example/vddk-args-configmap.yaml)
+
 ## Multi-stage Import
  In a multi-stage import, multiple pods are started in succession to copy different parts of the source to an existing base disk image. Currently only the [ImageIO](#multi-stage-imageio-import) and [VDDK](#multi-stage-vddk-import) data sources support multi-stage imports.
 
--- a/doc/datavolumes.md
+++ b/doc/datavolumes.md
@@ -351,7 +351,7 @@ spec:
 
 #### Extra VDDK Configuration Options
 
-The VDDK library itself looks in a configuration file (such as `/etc/vmware/config`) for extra options to fine tune data transfers. To pass these options through to the VDDK, store the configuration file contents in a ConfigMap and add a `cdi.kubevirt.io/storage.pod.vddk.extraargs` annotation to the DataVolume specification. The ConfigMap will be mounted to the importer pod as a volume, and the first file in the mounted directory will be passed to the VDDK. This means that the ConfigMap must be placed in the same namespace as the DataVolume, and the ConfigMap should only have one file entry.
+The VDDK library itself looks in a configuration file (such as `/etc/vmware/config`) for extra options to fine tune data transfers. To pass these options through to the VDDK, store the configuration file contents in a ConfigMap with the key `vddk-config-file` and add a `cdi.kubevirt.io/storage.pod.vddk.extraargs` annotation to the DataVolume specification. The ConfigMap will be mounted to the importer pod as a volume, and the mount directory will have a file named `vddk-config-file` with the contents of the file. This means that the ConfigMap must be placed in the same namespace as the DataVolume, and the ConfigMap should only have one file entry, `vddk-config-file`.
 
 [Example annotation](../manifests/example/vddk-args-annotation.yaml)
 [Example ConfigMap](../manifests/example/vddk-args-configmap.yaml)
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -242,7 +242,7 @@ func getVddkPluginPath() NbdkitPlugin {
 func getVddkConfig() (string, error) {
 	withHidden, err := os.ReadDir(common.VddkArgsDir)
 	if err != nil {
-		if os.IsNotExist(err) {
+		if os.IsNotExist(err) { // No extra arguments ConfigMap specified, so mount directory does not exist
 			return "", nil
 		}
 		return "", err
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -240,23 +239,15 @@ func getVddkPluginPath() NbdkitPlugin {
 // importer pod. Just look for the first file in the mounted directory, and
 // pass that through nbdkit via the "config=" option.
 func getVddkConfig() (string, error) {
-	withHidden, err := os.ReadDir(common.VddkArgsDir)
+	path := filepath.Join(common.VddkArgsDir, common.VddkArgsKeyName)
+	_, err := os.Stat(path)
 	if err != nil {
-		if os.IsNotExist(err) { // No extra arguments ConfigMap specified, so mount directory does not exist
+		if os.IsNotExist(err) { // No configuration file found, so no extra arguments to give to VDDK
 			return "", nil
 		}
 		return "", err
 	}
-	files := []fs.DirEntry{}
-	for _, file := range withHidden { // Ignore hidden files
-		if !strings.HasPrefix(file.Name(), ".") {
-			files = append(files, file)
-		}
-	}
-	if len(files) < 1 {
-		return "", fmt.Errorf("no VDDK configuration files found under %s", common.VddkArgsDir)
-	}
-	path := filepath.Join(common.VddkArgsDir, files[0].Name())
+
 	return path, nil
 }
 
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -168,6 +169,13 @@ func NewNbdkitVddk(nbdkitPidFile, socket string, args NbdKitVddkPluginArgs) (Nbd
 	pluginArgs = append(pluginArgs, "-D", "nbdkit.backend.datapath=0")
 	pluginArgs = append(pluginArgs, "-D", "vddk.datapath=0")
 	pluginArgs = append(pluginArgs, "-D", "vddk.stats=1")
+	config, err := getVddkConfig()
+	if err != nil {
+		return nil, err
+	}
+	if config != "" {
+		pluginArgs = append(pluginArgs, "config="+config)
+	}
 	p := getVddkPluginPath()
 	n := &Nbdkit{
 		NbdPidFile: nbdkitPidFile,
@@ -226,6 +234,30 @@ func getVddkPluginPath() NbdkitPlugin {
 		return NbdkitVddkMockPlugin
 	}
 	return NbdkitVddkPlugin
+}
+
+// Extra VDDK configuration options are stored in a ConfigMap mounted to the
+// importer pod. Just look for the first file in the mounted directory, and
+// pass that through nbdkit via the "config=" option.
+func getVddkConfig() (string, error) {
+	withHidden, err := os.ReadDir(common.VddkArgsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", nil
+		}
+		return "", err
+	}
+	files := []fs.DirEntry{}
+	for _, file := range withHidden { // Ignore hidden files
+		if !strings.HasPrefix(file.Name(), ".") {
+			files = append(files, file)
+		}
+	}
+	if len(files) < 1 {
+		return "", fmt.Errorf("no VDDK configuration files found under %s", common.VddkArgsDir)
+	}
+	path := filepath.Join(common.VddkArgsDir, files[0].Name())
+	return path, nil
 }
 
 func (n *Nbdkit) getSourceArg(s string) string {
--- a/pkg/image/nbdkit.go
+++ b/pkg/image/nbdkit.go
@@ -236,7 +236,7 @@ func getVddkPluginPath() NbdkitPlugin {
 }
 
 // Extra VDDK configuration options are stored in a ConfigMap mounted to the
-// importer pod. Just look for the first file in the mounted directory, and
+// importer pod. Make sure a "vddk-config-file" exists in that directory and
 // pass that through nbdkit via the "config=" option.
 func getVddkConfig() (string, error) {
 	path := filepath.Join(common.VddkArgsDir, common.VddkArgsKeyName)
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -1199,10 +1199,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 			if !k8serrors.IsAlreadyExists(err) {
 				Expect(err).ToNot(HaveOccurred())
 			}
-			if dv.Annotations == nil {
-				dv.Annotations = make(map[string]string)
-			}
-			dv.Annotations[controller.AnnVddkExtraArgs] = extraArguments.Name
+			controller.AddAnnotation(dv, controller.AnnVddkExtraArgs, extraArguments.Name)
 			return dv
 		}
 
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -1192,7 +1192,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 					Name:      "vddk-extra-args-map",
 				},
 				Data: map[string]string{ // Must match vddk-test-plugin
-					"vddk-config-file": "VixDiskLib.nfcAio.Session.BufSizeIn64KB=16",
+					common.VddkArgsKeyName: "VixDiskLib.nfcAio.Session.BufSizeIn64KB=16",
 				},
 			}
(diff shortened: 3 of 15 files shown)


mrnold:
@mnecas Any concerns about the need to copy the ConfigMap to the importer namespace? I wasn't sure if that would make things awkward to use from the forklift side.

pkg/image/nbdkit.go
```diff
@@ -228,6 +236,30 @@ func getVddkPluginPath() NbdkitPlugin {
...
 }
 
+// Extra VDDK configuration options are stored in a ConfigMap mounted to the
+// importer pod. Just look for the first file in the mounted directory, and
+// pass that through nbdkit via the "config=" option.
+func getVddkConfig() (string, error) {
+	withHidden, err := os.ReadDir(common.VddkArgsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", nil
```
mnecas, line 246: NP: Please add a comment that the user did not specify the vddk additional config

pkg/image/nbdkit.go
```diff
@@ -228,6 +236,30 @@ func getVddkPluginPath() NbdkitPlugin {
 	return NbdkitVddkPlugin
 }
 
+// Extra VDDK configuration options are stored in a ConfigMap mounted to the
+// importer pod. Just look for the first file in the mounted directory, and
```
mnecas, line 240: `Just look for the first file` just note we need to be sure to document this so user does not chain the configs to separate configmaps.

mnecas (approved):
Generally LGTM, I have added some a note and NP but nothing on my side

tests/datavolume_test.go
```diff
@@ -1254,6 +1276,30 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 					Message: "Import Complete",
 					Reason:  "Completed",
 				}}),
+			Entry("[test_id:5083]succeed importing VDDK data volume with extra arguments ConfigMap set", dataVolumeTestArguments{
```
akalenyu, line 1279: Do you want to assert something in the final step of this test? like checking the config was applied?

tools/vddk-test/vddk-test-plugin.c
```diff
@@ -32,6 +33,26 @@ int fakevddk_config(const char *key, const char *value) {
     if (strcmp(key, "snapshot") == 0) {
         expected_arg_count = 9; // Expect one for 'snapshot' and one for 'transports'
     }
+    if (strcmp(key, "config") == 0) {
```
akalenyu, line 36: Are there plans to backport this? just making sure I have a maintenance path crafted.

pkg/image/nbdkit.go
```diff
@@ -228,6 +236,30 @@ func getVddkPluginPath() NbdkitPlugin {
...
+func getVddkConfig() (string, error) {
+	withHidden, err := os.ReadDir(common.VddkArgsDir)
+	if err != nil {
+		if os.IsNotExist(err) { // No extra arguments ConfigMap specified, so mount directory does not exist
+			return "", nil
+		}
+		return "", err
+	}
+	files := []fs.DirEntry{}
+	for _, file := range withHidden { // Ignore hidden files
```
akalenyu, line 251: hmm, how come mounting a config map generates hidden files?

tests/datavolume_test.go
```diff
@@ -1184,6 +1184,28 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+					"vddk-config-file": "VixDiskLib.nfcAio.Session.BufSizeIn64KB=16",
+				},
+			}
+			_, err := f.K8sClient.CoreV1().ConfigMaps(f.Namespace.Name).Create(context.TODO(), &extraArguments, metav1.CreateOptions{})
+			if !k8serrors.IsAlreadyExists(err) {
+				Expect(err).ToNot(HaveOccurred())
+			}
+			if dv.Annotations == nil {
+				dv.Annotations = make(map[string]string)
+			}
```
akalenyu, line 1204: could use controller.AddAnnotation() here

doc/datavolumes.md
```diff
@@ -349,6 +349,13 @@ spec:
 [Get VDDK ConfigMap example](../manifests/example/vddk-configmap.yaml)
 [Ways to find thumbprint](https://libguestfs.org/nbdkit-vddk-plugin.1.html#THUMBPRINTS)
 
+#### Extra VDDK Configuration Options
+
+The VDDK library itself looks in a configuration file (such as `/etc/vmware/config`) for extra options to fine tune data transfers. To pass these options through to the VDDK, store the configuration file contents in a ConfigMap and add a `cdi.kubevirt.io/storage.pod.vddk.extraargs` annotation to the DataVolume specification. The ConfigMap will be mounted to the importer pod as a volume, and the first file in the mounted directory will be passed to the VDDK. This means that the ConfigMap must be placed in the same namespace as the DataVolume, and the ConfigMap should only have one file entry.
```
akalenyu, line 354: I guess you considered making this an API on the DataVolume, but, since you need a backport, you prefer the annotation?

akalenyu (commented):
Looks good!

tests/datavolume_test.go
mrnold: For this test I wanted to make sure that the contents of the ConfigMap are present in the file, so the assertion happens in the vddk-test-plugin (the fgets/strcmp). Is there a better way to check the result from the importer? Like can this test read the pod logs or termination message?

tools/vddk-test/vddk-test-plugin.c
mrnold: Yes, I asked around and I guess the customer looking for this is currently on 4.15. So we will definitely be asking for another barrage of backports.

pkg/image/nbdkit.go
mrnold: The volume mount sets up the directory like this:
```
drwxr-sr-x. 2 root  107 4096 Dec 18 15:55 ..2024_12_18_15_55_18.1193095340
lrwxrwxrwx. 1 root  107   32 Dec 18 15:55 ..data -> ..2024_12_18_15_55_18.1193095340
lrwxrwxrwx. 1 root  107   18 Dec 18 15:55 vddk-key -> ..data/vddk-key
```
I don't really know why, I didn't look into the implementation details. The documented way to use it is to open the file with the name of the key from the ConfigMap, so I was just skipping the entries with leading dots.

I guess it would be nicer to pass down the name of the key somehow, but what I was really trying to do was get the ConfigMap contents passed down in a way that CDI doesn't have to parse any of it.

tests/datavolume_test.go
mrnold: Much tidier, thanks. Done.

doc/datavolumes.md
mrnold: Yes, it didn't seem worth changing the CRDs and all the generated stuff for just for this uncommon fine-tuning configuration option. I can certainly change the API if that would be better.

akalenyu: @mhenriks wdyt? since this is backporting I am leaning to the annotation as well, but I am not sure.. usually any  annotation becomes an API that we forget about

pkg/image/nbdkit.go
akalenyu: You could do something like this https://github.com/kubevirt/containerized-data-importer/blob/625a9e944e3b75e514ca67e394c3e9931bf9634a/pkg/operator/resources/namespaced/controller.go#L342-L355

tests/datavolume_test.go
akalenyu: oh okay I see https://github.com/kubevirt/containerized-data-importer/blob/625a9e944e3b75e514ca67e394c3e9931bf9634a/tools/vddk-test/vddk-test-plugin.c#L50 Yeah I think we can come up with a cleaner way,
You can either read pod logs or make this a part of the termination message struct https://github.com/kubevirt/containerized-data-importer/blob/625a9e944e3b75e514ca67e394c3e9931bf9634a/pkg/importer/vddk-datasource_amd64.go#L1011-L1019

tools/vddk-test/vddk-test-plugin.c
akalenyu: Thanks for the heads up. Is there any way to avoid the test plugin change? I guess not. I am just asking since it could spare making new test images for all releases.

pkg/image/nbdkit.go
mrnold: This would be nice, but I was trying to not force a specific key name. With KeyToPath I would need to open the ConfigMap and look for the first key. Maybe it makes sense to just require a specific key name and avoid this issue entirely?

tools/vddk-test/vddk-test-plugin.c
mrnold: I didn't realize this was an issue, the test is definitely not set up to help avoid generating images. I'm not sure how else to do it though, short of hiding the plugin in the importer image or something.

akalenyu: That's fine

pkg/image/nbdkit.go
akalenyu: Yeah sounds cleaner.. whatever you prefer

tests/datavolume_test.go
akalenyu: WDYT?

rwmjones:
For interest only, here's an alternative to maintaining the fake VDDK plugin:

In nbdkit itself we are able to test the real nbdkit-vddk-plugin without VDDK, using a fake VDDK:

https://gitlab.com/nbdkit/nbdkit/-/blob/master/tests/dummy-vddk.c?ref_type=heads

which is compiled to `libvixDiskLib.so.6`:

https://gitlab.com/nbdkit/nbdkit/-/blob/9ed65418c57128d4bf372f39b9f98bf6ecbe470a/tests/Makefile.am

then you just have to point libdir= to the directory containing this file:

https://gitlab.com/nbdkit/nbdkit/-/blob/9ed65418c57128d4bf372f39b9f98bf6ecbe470a/tests/test-vddk.c#L61

tests/datavolume_test.go
mrnold: I have an in-progress implementation to check the pod logs, but before I commit to that I am also going to try getting the result into the termination message to see if it's any cleaner. Either way it's definitely better to check the result from the test code than from the VDDK test plugin itself, that should get rid of a source of silent failures in CI.

mrnold:
> For interest only, here's an alternative to maintaining the fake VDDK plugin:
> 
> In nbdkit itself we are able to test the real nbdkit-vddk-plugin without VDDK, using a fake VDDK:
> 
> https://gitlab.com/nbdkit/nbdkit/-/blob/master/tests/dummy-vddk.c?ref_type=heads
> 
> which is compiled to `libvixDiskLib.so.6`:
> 
> https://gitlab.com/nbdkit/nbdkit/-/blob/9ed65418c57128d4bf372f39b9f98bf6ecbe470a/tests/Makefile.am
> 
> then you just have to point libdir= to the directory containing this file:
> 
> https://gitlab.com/nbdkit/nbdkit/-/blob/9ed65418c57128d4bf372f39b9f98bf6ecbe470a/tests/test-vddk.c#L61

This is cool, I will look into this. I think it would still require maintaining and updating an equivalent image to hold libvixDiskLib.so.6, but it would let me get rid of [this](https://github.com/kubevirt/containerized-data-importer/blob/625a9e944e3b75e514ca67e394c3e9931bf9634a/pkg/image/nbdkit.go#L231) test plugin check.

tests/datavolume_test.go
akalenyu: The termination message approach should be cleaner, you would be asserting on a common.TerminationMessage struct that you got from the importer's API termination message

mrnold: I tried this and putting the result in the termination message is only cleaner from this angle, the final assert. The rest of the implementation needed to have the importer scanning for test-only output from the nbdkit log just to save an arbitrary string value into the termination struct, which is fairly limited space that could probably be put to better use.

I think it ended up being nicer to just scan the log from the test, so the actual importer doesn't need to do extra stuff to accommodate the test.

tests/import_test.go
```diff
@@ -189,6 +189,63 @@ var _ = Describe("[rfe_id:1115][crit:high][vendor:cnv-qe@redhat.com][level:compo
 			Expect(importer.DeletionTimestamp).To(BeNil())
 		}
 	})
+
+	It("[test_id:6689]succeed importing VDDK data volume with extra arguments ConfigMap set", Label("VDDK"), func() {
```
akalenyu, line 193: Would you mind putting this under a separate describe so it doesn't inherit the "Serial" part?
(I think this test can run in parallel to others just fine)

You can also change `[test_id:6689]` to `[test_id:XXXX]`; I don't think it works well if this entry is missing from the ID database

tests/framework/vddk.go
```diff
@@ -12,6 +12,38 @@ import (
 	"kubevirt.io/containerized-data-importer/tests/utils"
 )
 
+// CreateVddkDataVolume returns a VDDK data volume
+func (f *Framework) CreateVddkDataVolume(dataVolumeName, size, url string) *cdiv1.DataVolume {
```
akalenyu, line 16: How come something like this didn't already exist?

tests/datavolume_test.go
akalenyu: Fair, thanks for trying :pray:

tests/framework/vddk.go
mrnold: It does, but only in datavolume_test. I moved this test to import_test for some reason, I guess because I couldn't shoehorn the log scanning nicely into the big testDataVolume table. I moved it back to get rid of this change, it seems like it doesn't really belong in import_test any more than datavolume_test.

tests/datavolume_test.go
```diff
@@ -3463,6 +3463,65 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			By("Find importer pod " + pvcName)
+			importer, err := utils.FindPodByPrefixOnce(f.K8sClient, dataVolume.Namespace, common.ImporterPodName, common.CDILabelSelector)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(importer.DeletionTimestamp).To(BeNil())
+
+			logs, err := f.RunKubectlCommand("logs", "-n", dataVolume.Namespace, importer.Name)
+			Expect(err).ToNot(HaveOccurred())
+			for _, option := range vddkConfigOptions {
+				By(fmt.Sprintf("Check for configuration value %s in nbdkit logs", option))
+				Expect(strings.Contains(logs, option)).To(BeTrue())
```
akalenyu, line 3521: You can use the pod logs API, check this out: https://github.com/kubevirt/containerized-data-importer/pull/3184
And if you use `.Should(ContainSubstring` I am pretty sure it'll print the logs so we can properly understand why CI fails

mrnold: Done, the end of the log gets truncated though. I am going to try showing the whole log temporarily and see if it offers any insights.

akalenyu: So, it passes locally?

mrnold: Yes, it passes locally. It also correctly fails if I add another ContainSubstring line, and I can see the target strings in the log output.

mrnold: Interesting, the config argument is not getting passed to nbdkit:
```
I0109 19:07:46.993477       1 nbdkit.go:311] Start nbdkit with: ['--foreground' '--readonly' '--exit-with-parent' '-U' '/tmp/nbd.sock' '--pidfile' '/tmp/nbd.pid' '--filter=retry' '--filter=cacheextents' '/opt/testing/libvddk-test-plugin.so' 'libdir=/opt/vmware-vix-disklib-distrib' 'server=vcenter.cdi-ukbxj:8989' 'user=user' 'password=+/tmp/password2397077821' 'thumbprint=testprint' 'vm=moref=vm-21' '--verbose' '-D' 'nbdkit.backend.datapath=0' '-D' 'vddk.datapath=0' '-D' 'vddk.stats=1' 'file=[teststore] testvm/testdisk.vmdk']
```

I must have a mistake further up somewhere.

mrnold: I see, I didn't get the annotations copied over correctly through the import populator stuff.  The importer prime doesn't get the annotations, so I guess the pod doesn't get the ConfigMap mounted. I can reproduce this locally now.

akalenyu: I see, yeah usually the hot path is CSI storage (and thus populators). I assume you were not seeing any KUBEVIRT_STORAGE env var, and thus falling back to legacy population with the `local` storage class.

mrnold: Yes, that was it. The tests are looking better now.

akalenyu (commented):
/approve

thank you!

mrnold:
I guess the last outstanding question is [this one](https://github.com/kubevirt/containerized-data-importer/pull/3572#discussion_r1890522039) for @mhenriks, is this fine as an annotation or should it be moved to a CRD change?

akalenyu:
> I guess the last outstanding question is [this one](https://github.com/kubevirt/containerized-data-importer/pull/3572#discussion_r1890522039) for @mhenriks, is this fine as an annotation or should it be moved to a CRD change?

Maybe it can go in first with the annotation, and that will also go into backports. And then we can make this an API in main

mrnold:
> > I guess the last outstanding question is [this one](https://github.com/kubevirt/containerized-data-importer/pull/3572#discussion_r1890522039) for @mhenriks, is this fine as an annotation or should it be moved to a CRD change?
> 
> Maybe it can go in first with the annotation, and that will also go into backports. And then we can make this an API in main

Sure, that's fine with me.

mhenriks:
> Maybe it can go in first with the annotation, and that will also go into backports. And then we can make this an API in main

@akalenyu @mrnold this plan works for me

akalenyu:
> > Maybe it can go in first with the annotation, and that will also go into backports. And then we can make this an API in main
> 
> @akalenyu @mrnold this plan works for me

going to lgtm myself according to this, don't want to wait longer for this to get in
/lgtm

mrnold:
Thanks!

mrnold:
It looks like this applies cleanly to 1.61, 1.60, and 1.59, but 1.58 and older will need some work.

/cherry-pick release-v1.61

Issue #3557, opened 2024-12-04
Virtctl upload memory requirments


mw-0:
**What happened**:
When using the import feature  " virtctl image-upload dv ......"  i get an error when it hits the ram limit.  When i try and upload a larger 5GB qcow image i have to set the podresourcerequirements in the cdi spec to a higher amount or memory than the download size.  

I have tested a few different images and found the same issue when matching the limit lower.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 1.61.0
- Kubernetes version (use `kubectl version`): 1.31.2
- OS (e.g. from /etc/os-release): Debian12
- Kernel (e.g. `uname -a`): 6.10.11

akalenyu:
So you are seeing OOMs on the upload pod? which storage class are you using?
We've had similar situations with some kernel bug, was very common with nfs based storage classes:
https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/4523

mw-0:
i am using NFS 4.1 PVC with Netapp Trident.  I can see the pvc are attached.  Is the patch upstream into kernel or specific for RHEL there?

akalenyu:
It made it intos centos stream. But I don't know where it stands in other distros. If you have the debian sources should be easy enough to check if the revert made it in.

PR #3574, opened 2024-12-19
Solve the problem of pod OOM caused by increased buffer size due to inconsistent read and write speeds

mengkai514:
Description: The storage is not writing fast enough in most cases and the writes get backed up in memory, which caused pod OOM killed.

When I use DV to clone across storage classes (the image is very large, the storage is 2000Gi), after the upload is completed, kubectl get dv sees restart twice. And according to the log of cdi-upload pod, I found that if the image is raw, it will enter the ProcessingPhaseTransferDataFile(ProcessingPhaseConvert will not be executed) func to perform io.Copy, The problem should be that the cdi-upload pod OOMKilled during io.Copy. It would be the storage is too slow and data gets cached which causes memory usage to increase above the pod resource limit, which then causes an OOM on the server pod.

So there should be a method to write a certain amount of cache to disk.

Fixes #

**Release note**:

```release-note
Modify write method during io.Copy to prevent pod OOM.
```

### DIFF
--- a/pkg/importer/util.go
+++ b/pkg/importer/util.go
@@ -21,6 +21,40 @@ const (
 	kubevirtLabelPrefix = "kubevirt.io/"
 )
 
+// FsyncWriter is a custom writer that wraps an os.File
+// and calls fsync after writing a certain amount of data.
+type FsyncWriter struct {
+    file      *os.File
+    bytesWritten int64 // Keeps track of how many bytes have been written
+    threshold    int64 // Threshold for calling fsync
+}
+
+// NewFsyncWriter initializes a new FsyncWriter
+func NewFsyncWriter(file *os.File, threshold int64) *FsyncWriter {
+    return &FsyncWriter{
+        file:      file,
+        threshold: threshold,
+    }
+}
+
+// Write writes data to the underlying file and calls fsync
+// after the threshold is exceeded.
+func (w *FsyncWriter) Write(p []byte) (n int, err error) {
+    n, err = w.file.Write(p)
+    if err != nil {
+        return n, err
+    }
+    w.bytesWritten += int64(n)
+    if w.bytesWritten >= w.threshold {
+        err = w.file.Sync() // Call fsync after threshold is reached
+        if err != nil {
+            return n, err
+        }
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
Thank you for posting the PR & issue :pray: 

This is actually a very opinionated topic we've discussed many times before, most recently due to a kernel bug:
https://github.com/kubevirt/containerized-data-importer/issues/3557#issuecomment-2518662605
And you can see some of back and forth through the years using this PR query:
https://github.com/kubevirt/containerized-data-importer/pulls?q=is%3Apr+writeback+is%3Aclosed
Basically, the conclusion of these is that we would prefer to leave this up to the OS, and avoid O_DIRECT.

BTW, it should not be needed with cgroupsv2 (which is the standard now);
Under cgroupsv2, the cache should get forced to disk much faster, 0.3 * 600M
in the case of CDI importer (dirty_ratio considered against container limit, not host)

I would like to poke more at your setup instead, specifically that kernel bug.
If we do come to the conclusion that the storage is "just slow" (this would make many fsyncs perform badly), maybe, in this case, the right thing would be to just configure higher memory limits for the CDI pods.

mengkai514:
@akalenyu The PR you listed seems to be a little different from my problem, because mine does not involve qemu-img convert. My problem occurs in the Reader and Writer during the Copy phase, so the fsync method you suggested may be able to solve it. However, adjusting higher limits memory can only solve the problem temporarily. If you want a long-term solution, I think it should be fsync and cgroupsv2, but now there are probably some that do not support cgroupsv2.

akalenyu:
> @akalenyu The PR you listed seems to be a little different from my problem, because mine does not involve qemu-img convert. My problem occurs in the Reader and Writer during the Copy phase, so the fsync method you suggested may be able to solve it. However, adjusting higher limits memory can only solve the problem temporarily. If you want a long-term solution, I think it should be fsync and cgroupsv2, but now there are probably some that do not support cgroupsv2.

cgroupsv1 is no longer supported. I would really prefer that we don't have custom flushing logic in the project and that we inherit whatever the OS knobs (dirty_ratio) give us

PR #3578, opened 2025-01-05
Fix flaky DataImportCron volume snapshot test

arnongilboa:
As commented by @akalenyu : the condition check flakes because the switch to snapshot boot sources is almost instant. The `DataImportCron` never becomes "not ready" following the switch. We also moved the DIC "ready" check after the snapshot "ready" check, otherwise we may check it before it's "not ready", which checks nothing.

jira-ticket: https://issues.redhat.com/browse/CNV-40016

**Release note**:
```release-note
NONE
```


akalenyu:
/approve
/lgtm

:pray:

PR #3585, opened 2025-01-08
fix: Use storage class from annotation for host-assisted PVC operation

### LINKED ISSUE #3530: Random Storage Class Selection During Host-Assisted Snapshot Restore
**What happened**:
During host-assisted snapshot restore operations, CDI creates intermediate PVCs with randomly selected storage class (matching only the CSI driver), ignoring both the DataVolume's specified storage class and the source PVC's storage class. This leads to restore failures when an incompatible storage class is selected.

Error seen:

in the host-assisted pvc 
```
ProvisioningFailed: failed to provision volume with StorageClass "ms-ext4-3-replicas-local": 
rpc error: code = InvalidArgument desc = error in response: status code '400 Bad Request', content: 'RestJsonError { details: "", message: "SvcError :: ClonedSnapshotVolumeThin: Cloned snapshot volumes must be thin provisioned", kind: InvalidArgument }'
```

CDI should follow a predictable storage class selection priority during host-assisted snapshot restore:

- Use the storage class specified in the DataVolume
- Fall back to source PVC's storage class
- Use default storage class with matching driver
- Only as last resort, use any matching storage class

Create a snapshot from existing PVC:

```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: blank-datavolume-snapshot
spec:
  source:
    persistentVolumeClaimName: blank-datavolume
```

Create DataVolume to restore from snapshot:

```
kind: DataVolume
metadata:
  name: new-datavolume-from-snapshot
spec:
  source:
    snapshot:
      name: blank-datavolume-snapshot
      namespace: default
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 2Gi
    storageClassName: ms-xfs-snapshot-restore
```

Observe intermediate PVC creation:

```
kubectl get pvc <uuid>-host-assisted-source-pvc -o yaml
```

Shows storage class ms-ext4-3-replicas-local instead of the specified ms-xfs-snapshot-restore

**Additional context**:

The current implementation in getStorageClassCorrespondingToSnapClass only matches on CSI driver and takes the first match:

```
func (r *SnapshotCloneReconciler) getStorageClassCorrespondingToSnapClass(driver string) (string, error) {
    matches := []storagev1.StorageClass{}
    // Lists all storage classes
    storageClasses := &storagev1.StorageClassList{}
    if err := r.client.List(context.TODO(), storageClasses); err != nil {
        return "", errors.New("unable to retrieve storage classes")
    }
    // Just matches on driver and takes first match
    for _, storageClass := range storageClasses.Items {
        if storageClass.Provisioner == driver {
            matches = append(matches, matches)
        }
    }
    if len(matches) > 1 {
        return matches[0].Name, nil
    }
    //...
}
```

Suggested Changes:

Prioritized Storage Class Selection:

1. First: Use DV's specified storage class (if matches driver)
2. Second: Try source PVC's storage class (from snapshot)
3. Third: Try default storage class (with matching driver)
4. Last Resort: Any matching storage class

or 

Have a way for the user to configure behaviour for intermediate pvc

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`):v1.60.3
Server: v1.22.17
Client: v1.31.0
- DV specification: 
```
kind: DataVolume
metadata:
  name: new-datavolume-from-snapshot
  namespace: default
spec:
  source:
    snapshot:
      name: blank-datavolume-snapshot
      namespace: default
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 2Gi
    storageClassName: ms-xfs-snapshot-restore
```
    
- Install tools: 
- Others: 
   Multiple storage classes using same CSI driver `io.openebs.csi-mayastor`:
  - ms-ext4-3-replicas-local
  - ms-xfs-snapshot-restore
  - VolumeSnapshot feature enabled and CRDs installed

TheRealSibasishBehera:
modify the `getStorageClassCorrespondingToSnapClass` function to use the storage class from the annotation for host-assisted PVC operations.

Fixes #3530

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume/snapshot-clone-controller.go
+++ b/pkg/controller/datavolume/snapshot-clone-controller.go
@@ -418,7 +418,7 @@ func (r *SnapshotCloneReconciler) createTempHostAssistedSourcePvc(dv *cdiv1.Data
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: *snapshot.Spec.VolumeSnapshotClassName}, vsc); err != nil {
 		return err
 	}
-	sc, err := r.getStorageClassCorrespondingToSnapClass(vsc.Driver)
+	sc, err := r.getStorageClassCorrespondingToSnapClass(dv, vsc.Driver)
 	if err != nil {
 		return err
 	}
@@ -450,7 +450,12 @@ func (r *SnapshotCloneReconciler) createTempHostAssistedSourcePvc(dv *cdiv1.Data
 	return nil
 }
 
-func (r *SnapshotCloneReconciler) getStorageClassCorrespondingToSnapClass(driver string) (string, error) {
+func (r *SnapshotCloneReconciler) getStorageClassCorrespondingToSnapClass(dv *cdiv1.DataVolume, driver string) (string, error) {
+	if sc, found := dv.Annotations[cc.AnnSnapshotRestoreStorageClass]; found {
+		r.log.V(3).Info("Using storage class from annotation for host-assisted source PVC operation", "storageClass.Name", sc)
+		return sc, nil
+	}
+
 	matches := []storagev1.StorageClass{}
 
 	storageClasses := &storagev1.StorageClassList{}
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -225,6 +225,9 @@ const (
 
 	// AnnSourceVolumeMode is the volume mode of the source PVC specified as an annotation on snapshots
 	AnnSourceVolumeMode = AnnAPIGroup + "/storage.import.sourceVolumeMode"
+	// AnnSnapshotRestoreStorageClass indicates the storage class to use for restoring snapshots.
+	// This annotation should be added to DataVolumes for the temporary host-assisted source PVC.
+	AnnSnapshotRestoreStorageClass = AnnAPIGroup + "/snapshot.restore.storageClass"
 
 	// AnnOpenShiftImageLookup is the annotation for OpenShift image stream lookup
 	AnnOpenShiftImageLookup = "alpha.image.policy.openshift.io/resolve-names"
--- a/pkg/controller/datavolume/snapshot-clone-controller_test.go
+++ b/pkg/controller/datavolume/snapshot-clone-controller_test.go
@@ -168,6 +168,47 @@ var _ = Describe("All DataVolume Tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("Should use the storage class used in annotation AnnSnapshotRestoreStorageClass if present", func() {
+			dv := newCloneFromSnapshotDataVolume("test-dv")
+			scName := "testsc"
+			expectedSnapshotClass := "snap-class"
+			sc := CreateStorageClassWithProvisioner(scName, map[string]string{
+				AnnDefaultStorageClass: "true",
+			}, map[string]string{}, "csi-plugin")
+			targetScName := "targetsc"
+			scSameProvisioner := sc.DeepCopy()
+			scSameProvisioner.Name = "same-provisioner-as-source-sc"
+			scUsedInAnnotation := sc.DeepCopy()
+			scUsedInAnnotation.Name = "sc-used-in-annotation"
+			dv.Annotations[AnnSnapshotRestoreStorageClass] = scUsedInAnnotation.Name
+
+			tsc := CreateStorageClassWithProvisioner(targetScName, map[string]string{}, map[string]string{}, "another-csi-plugin")
+			sp := createStorageProfile(scName, []corev1.PersistentVolumeAccessMode{corev1.ReadOnlyMany}, BlockMode)
+			sp2 := createStorageProfile(targetScName, []corev1.PersistentVolumeAccessMode{corev1.ReadOnlyMany}, BlockMode)
+			sp3 := createStorageProfile(scSameProvisioner.Name, []corev1.PersistentVolumeAccessMode{corev1.ReadOnlyMany}, BlockMode)
+			sp4 := createStorageProfile(scUsedInAnnotation.Name, []corev1.PersistentVolumeAccessMode{corev1.ReadOnlyMany}, BlockMode)
+			dv.Spec.PVC.StorageClassName = &targetScName
+			snapshot := createSnapshotInVolumeSnapshotClass("test-snap", metav1.NamespaceDefault, &expectedSnapshotClass, nil, nil, true)
+			snapClass := createSnapshotClass(expectedSnapshotClass, nil, "csi-plugin")
+			reconciler = createSnapshotCloneReconciler(sc, scSameProvisioner, tsc, sp, sp2, sp3, sp4, dv, snapshot, snapClass, createDefaultVolumeSnapshotContent(), createVolumeSnapshotContentCrd(), createVolumeSnapshotClassCrd(), createVolumeSnapshotCrd())
+			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}})
+			Expect(err).ToNot(HaveOccurred())
+			By("Verifying that temp host assisted source PVC is being created")
+			pvc := &corev1.PersistentVolumeClaim{}
+			tempPvcName := getTempHostAssistedSourcePvcName(dv)
+			err = reconciler.client.Get(context.TODO(), types.NamespacedName{Namespace: snapshot.Namespace, Name: tempPvcName}, pvc)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pvc.Labels[common.CDIComponentLabel]).To(Equal("cdi-clone-from-snapshot-source-host-assisted-fallback-pvc"))
+			Expect(pvc.Labels[common.AppKubernetesPartOfLabel]).To(Equal("testing"))
+			Expect(*pvc.Spec.StorageClassName).To(Equal(scUsedInAnnotation.Name))
+			By("Verifying that target host assisted PVC is being created")
+			err = reconciler.client.Get(context.TODO(), types.NamespacedName{Namespace: dv.Namespace, Name: dv.Name}, pvc)
+			Expect(err).ToNot(HaveOccurred())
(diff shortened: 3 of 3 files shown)


TheRealSibasishBehera:
cc: @akalenyu

akalenyu:
Thank you for the PR! :pray: 

What do you think about capturing the mapping between the volumesnapshotclass->storageclass
on the `volumesnapshotclass` annotations instead?
This way, the admin configures it once on a cluster level and subsequent restores should work seamlessly

pkg/controller/datavolume/snapshot-clone-controller.go
```diff
@@ -450,7 +450,12 @@ func (r *SnapshotCloneReconciler) createTempHostAssistedSourcePvc(dv *cdiv1.Data
 	return nil
 }
 
-func (r *SnapshotCloneReconciler) getStorageClassCorrespondingToSnapClass(driver string) (string, error) {
+func (r *SnapshotCloneReconciler) getStorageClassCorrespondingToSnapClass(dv *cdiv1.DataVolume, driver string) (string, error) {
```
akalenyu, line 453: So this only covers the non-CSI path (`initLegacyClone`), I assume you're more interested in the CSI one?

TheRealSibasishBehera:
> Thank you for the PR! 🙏
> 
> What do you think about capturing the mapping between the volumesnapshotclass->storageclass on the `volumesnapshotclass` annotations instead? This way, the admin configures it once on a cluster level and subsequent restores should work seamlessly for any DV

Yes it sounds good that the configuration can be set once .

If I understand correctly, we would refer to the VolumeSnapshotClass from the source snapshot during restores. However, I am wondering about the scenario where the VolumeSnapshotClass is deleted after the snapshot is created. In such case we cant refer to the appropriate storageclass used in VolumeSnapshotClass annotation

pkg/controller/datavolume/snapshot-clone-controller.go
TheRealSibasishBehera: The changes I made are for cases where TempHostAssistedSourcePvc is created, i.e., the csi driver is not present,  or the csi driver for the source and target is mismatched. IIRC both are triggered by  initLegacyClone . the first one is my use case

happy to work on any similar cases. Are you referring to the remaining case, which is the csi driver for target pvc and snapshot match? It tries for a direct snapshot to target pvc restore , but it fails because of a csi incompatibility issue ?

akalenyu:
> > Thank you for the PR! 🙏
> > What do you think about capturing the mapping between the volumesnapshotclass->storageclass on the `volumesnapshotclass` annotations instead? This way, the admin configures it once on a cluster level and subsequent restores should work seamlessly for any DV
> 
> Yes it sounds good that the configuration can be set once .
> 
> If I understand correctly, we would refer to the VolumeSnapshotClass from the source snapshot during restores. However, I am wondering about the scenario where the VolumeSnapshotClass is deleted after the snapshot is created. In such case we cant refer to the appropriate storageclass used in VolumeSnapshotClass annotation

I think that is either unlikely or impossible when a snapshot that provisioned by it still exists

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: Oh right, I forgot the issue details. Let me stress again that what you're describing is very unusual today
> the csi driver is not present

> It tries for a direct snapshot to target pvc restore , but it fails because of a csi incompatibility issue ?

Yes, here's a ref to this path https://github.com/kubevirt/containerized-data-importer/blob/dcf7cfba8a0616cd4931881768057a84fdea960b/pkg/controller/clone/planner.go#L835

TheRealSibasishBehera: Thanks! I will work on this

TheRealSibasishBehera:
> I think that is either unlikely or impossible when a snapshot that was provisioned by it still exists (that snapshot cannot be deleted forever in these cases)

Got it, I’ll update the implementation 

Also, do you think this ability needs to be documented anywhere?

akalenyu:
> > I think that is either unlikely or impossible when a snapshot that was provisioned by it still exists (that snapshot cannot be deleted forever in these cases)
> 
> Got it, I’ll update the implementation
> 
> Also, do you think this ability needs to be documented anywhere?

Yes, feel free to drop a hint at https://github.com/kubevirt/containerized-data-importer/blob/dcf7cfba8a0616cd4931881768057a84fdea960b/doc/clone-from-volumesnapshot-source.md?plain=1#L1-L3

pkg/controller/datavolume/snapshot-clone-controller.go
akalenyu: So just to poke at the setup a little bit again, the provisioner has snapshots (which I thought implies it's CSI) but there's no CSIDriver object?

TheRealSibasishBehera: Yes, that's the case. The provisioner is [openebs mayastor](https://github.com/openebs/mayastor) 
The deployment manifest does not contain a `CSIDriver` packaged within

TheRealSibasishBehera: wdyt think about also using [pv.spec.CSI](https://github.com/kubernetes/api/blob/v0.32.0/core/v1/types.go#L348) since it is directly populated in when using a csi driver 

 I found this logic being used by [velero](https://github.com/vmware-tanzu/velero/blob/main/pkg/util/kube/utils.go#L238-L241) in a way

akalenyu: > Yes, that's the case. The provisioner is [openebs mayastor](https://github.com/openebs/mayastor) The deployment manifest does not contain a `CSIDriver` packaged within

That sounds unintended, how would you then configure things like `fsGroupPolicy`?
https://kubernetes-csi.github.io/docs/support-fsgroup.html

TheRealSibasishBehera: > > Yes, that's the case. The provisioner is [openebs mayastor](https://github.com/openebs/mayastor) The deployment manifest does not contain a `CSIDriver` packaged within
> 
> That sounds unintended, how would you then configure things like `fsGroupPolicy`? https://kubernetes-csi.github.io/docs/support-fsgroup.html

These are filled in as parameters for the storage class. I have confirmed with their community that they do not have a CSI driver

TheRealSibasishBehera: > > wdyt think about also using [pv.spec.CSI](https://github.com/kubernetes/api/blob/v0.32.0/core/v1/types.go#L348) since it is directly populated in when using a csi driver
> 
> I don't think it's possible since a PV does not exist at the time CDI is interested in this

so its not evaluated on fly for each dv/pvc ?

akalenyu: I don't understand the motivation to avoid a CSIDriver object.. AFAIK it's a requirement https://kubernetes-csi.github.io/docs/csi-driver-object.html

akalenyu:
> Thank you for the PR! 🙏
> 
> What do you think about capturing the mapping between the volumesnapshotclass->storageclass on the `volumesnapshotclass` annotations instead? This way, the admin configures it once on a cluster level and subsequent restores should work seamlessly for any DV

Just realized we already have a mechanism to tie storageclass->snapclass via https://github.com/kubevirt/containerized-data-importer/pull/2898 so we could just use this instead of an annotation?
(Loop over storageprofiles with vsc == sourceSnap.vsc)
Not completely sure what happens if 2 storage classes point at the same snapshot class though

pkg/controller/datavolume/snapshot-clone-controller.go
TheRealSibasishBehera: > I don't understand the motivation to avoid a CSIDriver object.. AFAIK it's a requirement https://kubernetes-csi.github.io/docs/csi-driver-object.html

Yeah I agree it’s a standard and should be supported :disappointed: 
This seems like a case of delayed adoption to me, as it used to work without it

TheRealSibasishBehera:
> > Thank you for the PR! 🙏
> > What do you think about capturing the mapping between the volumesnapshotclass->storageclass on the `volumesnapshotclass` annotations instead? This way, the admin configures it once on a cluster level and subsequent restores should work seamlessly for any DV
> 
> Just realized we already have a mechanism to tie storageclass->snapclass via #2898 so we could just use this instead of an annotation? (Loop over storageprofiles with vsc == sourceSnap.vsc) Not completely sure what happens if 2 storage classes point at the same snapshot class though

This would not work if we check the status. If multiple StorageProfiles have the same snapshot class (e.g., because it is set as the default or is the first in a sorted list), we might end up picking a storage class(say the first one in loop over storageprofiles with vsc == sourceSnap.vsc ) which is not intentionally set by the admin and is incompatible

we can check the spec for this, as it is user-defined. This way, we can trust that the storage class and snapshot class are compatible as a pair, and the selection is not arbitrary

akalenyu:
> > > Thank you for the PR! 🙏
> > > What do you think about capturing the mapping between the volumesnapshotclass->storageclass on the `volumesnapshotclass` annotations instead? This way, the admin configures it once on a cluster level and subsequent restores should work seamlessly for any DV
> > 
> > 
> > Just realized we already have a mechanism to tie storageclass->snapclass via #2898 so we could just use this instead of an annotation? (Loop over storageprofiles with vsc == sourceSnap.vsc) Not completely sure what happens if 2 storage classes point at the same snapshot class though
> 
> This would not work if we check the status. If multiple StorageProfiles have the same snapshot class (e.g., because it is set as the default or is the first in a sorted list), we might end up picking a storage class(say the first one in loop over storageprofiles with vsc == sourceSnap.vsc ) which is not intentionally set by the admin and is incompatible
> 
> we can check the spec for this, as it is user-defined. This way, we can trust that the storage class and snapshot class are compatible as a pair, and the selection is not arbitrary

I'm afraid that would suffer from the same issue where two profiles point to the same volumesnapshotclass

adamcharnock:
This is currently getting us too. We have different Mayastor volume types which are not really compatible. Either some are not thin provisioned, or some are configuring to retain volumes post-deletion (which is also very not good for temporary volumes).

I'd love to see this moved forward if at all possible ❤️

PR #3591, opened 2025-01-14
Label VDDK and ImageIO tests

Davo911:

The preferred way of filtering tests using unsupported tools is to use ginkgo built-in label filters. 
Here I'm labeling every test using VDDK and ImageIO, so that in the example of `s390x`, one can run the tests with `CDI_LABEL_FILTER='!VDDK && !ImageIO'` to skip those.

**Release note**:

```release-note
NONE
```


tests/api_validation_test.go
```diff
@@ -111,10 +111,10 @@ var _ = Describe("[rfe_id:1130][crit:medium][posneg:negative][vendor:cnv-qe@redh
 				Entry("[test_id:1325]fail with empty PVC source namespace", "pvc", "", "test-pvc"),
 				Entry("[test_id:1326]fail with empty PVC source name", "pvc", "test", ""),
 				Entry("[test_id:3917]fail with source PVC doesn't exist", "pvc", "test", "test-pvc"),
-				Entry("[test_id:3918]fail with empty Imageio source diskId", "imageio", validURL, "secret", "tls-cert", ""),
-				Entry("[test_id:3926]fail with empty VDDK source UUID", "vddk", validURL, "secret", "", "backingfile", "thumbprint"),
-				Entry("[test_id:3927]fail with empty VDDK source backing file", "vddk", validURL, "secret", "uuid", "", "thumbprint"),
-				Entry("[test_id:3928]fail with empty VDDK source thumbprint", "vddk", validURL, "secret", "uuid", "backingfile", ""),
+				Entry("[test_id:3918]fail with empty Imageio source diskId", "imageio", validURL, "secret", "tls-cert", "", Label("ImageIO")),
```
akalenyu, line 114: I am pretty sure labels should follow test entry name
/hold
feel free to unhold if this is wrong but at the very least it's what I've seen mostly

awels: Oh yeah, you are right. The labels should be after the name of the test, the rest are arguments to table function.

Davo911: At least in a initial test it got filtered out, so my understanding was, that the positions doesn't matter, but I'm happy to move it for conformance

Davo911:
I also added a slight change to the `run-functional-test.sh`, as I noticed, that if operators are used without escaped quotes the filter only applies to the first expression. 
In the case of my example `CDI_LABEL_FILTER='!VDDK && !ImageIO'` only VDDK was filtered out.

akalenyu:
> I also added a slight change to the `run-functional-test.sh`, as I noticed, that if operators are used without escaped quotes the filter only applies to the first expression. In the case of my example `CDI_LABEL_FILTER='!VDDK && !ImageIO'` only VDDK was filtered out.

Interesting, can you attach a comparison of the arguments between a working/broken run?

Davo911:
> Interesting, can you attach a comparison of the arguments between a working/broken run?

This one is created, when running without escaped quotes
```shell
+ containerized-data-importer/_out/tests/ginkgo --trace --timeout=8h --v '--label-filter=!VDDK' '&&' '!ImageIO' containerized-data-importer/_out/tests/tests.test -- -cdi-namespace=cdi -kubeconfig=/root/.kube/config -kubectl-path=/usr/local/bin/kubectl -oc-path=/usr/local/bin/kubectl -gocli-path=containerized-data-importer/cluster-up/cli.sh -snapshot-sc=rook-ceph-block -block-sc=rook-ceph-block -csiclone-sc=rook-ceph-block -docker-prefix=quay.io/kubevirt -docker-tag=20241119_448fe9d56-s390x
``` 

With escaped quotes:
```shell
+ containerized-data-importer/_out/tests/ginkgo --trace --timeout=8h --v '--label-filter="!VDDK' '&&' '!ImageIO"' containerized-data-importer/_out/tests/tests.test -- -cdi-namespace=cdi -kubeconfig=/root/.kube/config -kubectl-path=/usr/local/bin/kubectl -oc-path=/usr/local/bin/kubectl -gocli-path=containerized-data-importer/cluster-up/cli.sh -snapshot-sc=rook-ceph-block -block-sc=rook-ceph-block -csiclone-sc=rook-ceph-block -docker-prefix=quay.io/kubevirt -docker-tag=20241119_448fe9d56-s390x
``` 
But I just noticed this also only works when adding a `eval` to the final ginkgo command

```shell
eval ${TESTS_OUT_DIR}/ginkgo ${ginkgo_args} ${TESTS_OUT_DIR}/tests.test -- ${test_args}
``` 
This could be problematic, what do you say?

awels:
/hold
wait for #3600 to merge, then we can re-run this.

Davo911:
@awels Could we unhold this?

akalenyu (approved):

PR #3600, opened 2025-01-15
switch to bash array to construct ginkgo cmd

Davo911:
While using ginkgos `--label-filter` through `$CDI_LABEL_FILTER` [with operators](https://onsi.github.io/ginkgo/#spec-labels) the whitespaces and special characters result in a faulty final command.
For example in the current state when having `CDI_LABEL_FILTER='!VDDK && !ImageIO'` we get:
```shell
ginkgo '--label-filter=!VDDK' '&&' '!ImageIO' 
```
This only filters out VDDK and ignores ImageIO.

To overcome this I propose to use a bash array to construct the final command, enabling the use of operators without having to use escaping or `eval`

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/run-functional-tests.sh
+++ b/hack/build/run-functional-tests.sh
@@ -95,25 +95,26 @@ fi
 
 (
     export TESTS_WORKDIR=${CDI_DIR}/tests
-    ginkgo_args="--trace --timeout=8h --v"
+    declare -a ginkgo_args
+    ginkgo_args+=(--trace --timeout=8h --v)
 
     if [[ -n "$CDI_LABEL_FILTER" ]]; then
-        ginkgo_args="${ginkgo_args} --label-filter=${CDI_LABEL_FILTER}"
+        ginkgo_args+=(--label-filter="${CDI_LABEL_FILTER}")
     fi
 
     if [[ -n "$CDI_E2E_SKIP" ]]; then
-        ginkgo_args="${ginkgo_args} --skip=${CDI_E2E_SKIP}"
+        ginkgo_args+=(--skip="${CDI_E2E_SKIP}")
     fi
 
     if [[ "$CDI_E2E_FOCUS" =~ /.+\.go/ ]]; then
-        ginkgo_args="${ginkgo_args} --focus-file=${CDI_E2E_FOCUS}"
+        ginkgo_args+=(--focus-file="${CDI_E2E_FOCUS}")
     elif [[ -n "$CDI_E2E_FOCUS" ]]; then
-        ginkgo_args="${ginkgo_args} --focus=${CDI_E2E_FOCUS}"
+        ginkgo_args+=(--focus="${CDI_E2E_FOCUS}")
     fi
 
     if [[ -n "$CDI_E2E_FOCUS" || -n "$CDI_E2E_SKIP" ]]; then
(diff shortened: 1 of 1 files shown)


hack/build/run-functional-tests.sh
```diff
@@ -95,25 +95,26 @@ fi
 
 (
     export TESTS_WORKDIR=${CDI_DIR}/tests
-    ginkgo_args="--trace --timeout=8h --v"
+    declare -a ginkgo_args
+    ginkgo_args+=(--trace --timeout=8h --v)
 
     if [[ -n "$CDI_LABEL_FILTER" ]]; then
-        ginkgo_args="${ginkgo_args} --label-filter=${CDI_LABEL_FILTER}"
+        ginkgo_args+=(--label-filter="${CDI_LABEL_FILTER}")
```
akalenyu, line 102: Wouldn't this alone solve the issue? I mean, quoting should solve bash breaking apart the single argument to many?
`--label-filter="${CDI_LABEL_FILTER}"`

Davo911: simply adding quoting here results in this in the final command : 
```shell
ginkgo --trace --timeout=8h --v '--label-filter=!ImageIO' '&&' '!VDDK'
```
escaping the quotes goes in the right direction but fails at the special characters requiring to add an `eval` 
```shell
ginkgo --trace --timeout=8h --v '--label-filter="!ImageIO' '&&' '!VDDK"'

Ginkgo detected configuration issues:
Syntax Error Parsing Label Filter
  "!ImageIO
  Invalid token '!'.
```

akalenyu:
/test all
/approve
/hold
will remove hold once I make sure the test set is identical. most likely is, just double checking

hack/build/run-functional-tests.sh
```diff
@@ -95,25 +95,26 @@ fi
...
     fi
 
     if [[ -n "$CDI_E2E_FOCUS" || -n "$CDI_E2E_SKIP" ]]; then
-        ginkgo_args="${ginkgo_args} --nodes=6"
+        ginkgo_args+=(--nodes=6)
     fi
 
-    ${TESTS_OUT_DIR}/ginkgo ${ginkgo_args} ${TESTS_OUT_DIR}/tests.test -- ${test_args}
+    ${TESTS_OUT_DIR}/ginkgo "${ginkgo_args[@]}" ${TESTS_OUT_DIR}/tests.test -- ${test_args}
 )
```
akalenyu, line 120: Are you deleting the newline at the end here?

Davo911: At the end of the file?

akalenyu: yeah

Davo911: Maybe I'm wrong but there was no newline before or at least I don't see it 👀

akalenyu: Right yup got confused by github showing me that red error sign

ShellyKa13:
/lgtm
Thanks for the PR!
I want to ask for next time (hopefully there will be one :) ) that the explanation you did in the PR description to also have in the commit message to explain the reason for the change. Not blocking the merge cause I dont want all the tests to run again just for this

Davo911:
Thanks a lot , will do @ShellyKa13 !

PR #3617, opened 2025-01-26
Add PVC source support for DataImportCron

arnongilboa:
A `PVC` from any namespace can now be the source for a `DataImportCron`. The source digest is based on the `PVC` `UID`, which is polled by the schedule similarly to `ImageStream`, so when a new `PVC` is detected it will be imported.

jira-ticket: https://issues.redhat.com/browse/CNV-55496

**Release note**:
```release-note
Add PVC source support for DataImportCron
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -100,7 +100,8 @@ const (
 	AnnStorageClass = cc.AnnAPIGroup + "/storage.import.storageClass"
 
 	dataImportControllerName    = "dataimportcron-controller"
-	digestPrefix                = "sha256:"
+	digestSha256Prefix          = "sha256:"
+	digestUIDPrefix             = "uid:"
 	digestDvNameSuffixLength    = 12
 	cronJobUIDSuffixLength      = 8
 	defaultImportsToKeepPerCron = 3
@@ -161,7 +162,7 @@ func (r *DataImportCronReconciler) initCron(ctx context.Context, dataImportCron 
 	if dataImportCron.Spec.Schedule == "" {
 		return nil
 	}
-	if isImageStreamSource(dataImportCron) {
+	if isControllerPolledSource(dataImportCron) {
 		if dataImportCron.Annotations[AnnNextCronTime] == "" {
 			cc.AddAnnotation(dataImportCron, AnnNextCronTime, time.Now().Format(time.RFC3339))
 		}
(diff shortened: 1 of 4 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -248,15 +249,21 @@ func splitImageStreamName(imageStreamName string) (string, string, error) {
 	return "", "", errors.Errorf("Illegal ImageStream name %s", imageStreamName)
 }
 
-func (r *DataImportCronReconciler) pollImageStreamDigest(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
+func (r *DataImportCronReconciler) pollSourceDigest(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
 	if nextTimeStr := dataImportCron.Annotations[AnnNextCronTime]; nextTimeStr != "" {
```
ShellyKa13, line 253: nit: return here and remove all function indentation

pkg/controller/dataimportcron-controller.go
```diff
@@ -1508,13 +1560,21 @@ func (r *DataImportCronReconciler) newDataSource(cron *cdiv1.DataImportCron) *cd
 	return dataSource
 }
 
-// Create DataVolume name based on the DataSource name + prefix of the digest sha256
+// Create DataVolume name based on the DataSource name + prefix of the digest
 func createDvName(prefix, digest string) (string, error) {
-	fromIdx := len(digestPrefix)
-	toIdx := fromIdx + digestDvNameSuffixLength
-	if !strings.HasPrefix(digest, digestPrefix) {
```
ShellyKa13, line 1515: isnt it better to just do 3 `if` one per suffix since currently we only have 2 prefix options?
```
if strings.HasPrefix(digest, digestSha256Prefix) {
    digestPrefix = digestSha256Prefix
} else if strings.HasPrefix(digest, digestUIDPrefix) {
    digestPrefix = digestUIDPrefix
} else {
    return "", errors.Errorf("Digest has no supported prefix")
}
```

pkg/controller/dataimportcron-controller.go
```diff
@@ -1434,16 +1484,18 @@ func (r *DataImportCronReconciler) setJobCommon(cron *cdiv1.DataImportCron, obj
...
 	dv := cron.Spec.Template.DeepCopy()
-	if isURLSource(cron) {
-		digestedURL = untagDigestedDockerURL(*dv.Spec.Source.Registry.URL + "@" + cron.Annotations[AnnSourceDesiredDigest])
-	} else if isImageStreamSource(cron) {
-		// No way to import image stream by name when we want specific digest, so we use its docker reference
-		digestedURL = "docker://" + cron.Annotations[AnnImageStreamDockerRef]
-		dv.Spec.Source.Registry.ImageStream = nil
-	}
-	dv.Spec.Source.Registry.URL = &digestedURL
+	if isCronRegistrySource(cron) {
```
ShellyKa13, line 1488: not familiar enough with the code but, no need to do anything here with pvc source?

ShellyKa13 (commented):
2 nits and a question, which are up to you :)
/lgtm

pkg/controller/dataimportcron-controller.go
arnongilboa: I see you are in code quality mood. This is dusty code:)

pkg/controller/dataimportcron-controller.go
arnongilboa: Agree:)

pkg/controller/dataimportcron-controller.go
arnongilboa: we use it as is.

pkg/controller/dataimportcron-controller.go
ShellyKa13: I know but if you are in the neighborhood you can beautify the houses lol

ShellyKa13 (commented):
Thanks!
/lgtm

awels (approved):
/approve

PR #3619, opened 2025-01-28
Allow configuration of import pod restart policy

paullanum:
We had some DataVolumes failing because the storage request size was accidentally too small for the size of our ISO sources. Our DVs retried continuously until we noticed the issue, stuck in a download -> fail loop that caused a spike in our cloud costs due to the increased network usage.  By being able to specify the RestartPolicy of the importer pods to `Never`, we could prevent this scenario in the future."

Fixes #

**Release note**:

```release-note
Allow configuring Restart Policy for importer pods
```


akalenyu (commented):
Thanks for the PR and sharing a clear motivation!

My worry is introducing another API knob, one that we would have to work really hard to advertise and contain the complexity of
(from my experience, folks would just overlook it and hit those issues)

What if we could instead introduce a framework for detecting a terminal state like the one described here, and respond by deleting the pod via the controller? WDYT?

paullanum:
I definitely understand the concern about adding another knob for users, though in this case I don't believe it would need to be heavily advertised since this is optional configuration and the currently hardcoded value (`v1.RestartPolicyOnFailure`) is still the default. The configuration is only for those who want or need it and existing users who overlook the change should not be affected.

Regarding complexity, I have to respectfully disagree that adding logic to detect failure states would be less complex than leveraging the out-of-the-box RestartPolicy behavior in Kubernetes.

alromeros:
Might be wrong but I remember testing the `Never` restart policy with importer pods and it was pretty flaky, I think especially for flows needing scratch space. Transient HTTP errors are also common... generally using never is discouraged as CDI  has been tested with OnFailure and we make assumptions based on this logic. I see why this feature could be useful, but I understand @akalenyu concerns. I'm not fully against it but it would be great if we could consider alternatives.

seanmorton:
@alromeros those assumptions are good to know about. How much of a lift would it be to convert the importer pod to a `Job`? If that direction was taken we could use `.spec.backoffLimit` to limit the number of retries.

PR #3625, opened 2025-01-30
update k8s & related libraries to 1.31

akalenyu:

Fixes #

**Release note**:

```release-note
Cleanup: k8s 1.31 dependencies
```


akalenyu:
@mhenriks looks like with this library-go change https://github.com/openshift/library-go/pull/1772 there's additional logic to call the CREATE.
not sure if this messes up functionality (should requeue on already exists err) but it definitely hurts the unit tests

I tried shifting the order a bit but it's clunky. Calling Sync() an extra time also doesn't help, seems like the lister is never updated and thus the library-go mechanism keeps trying to CREATE (which I guess makes sense since it's all fake)

pkg/operator/resources/cert/factory.go
```diff
@@ -211,12 +211,17 @@ func createCertificateDefinitions() []CertificateDefinition {
...
+func createTLSSecret(name string) *corev1.Secret {
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   name,
 			Labels: utils.ResourceBuilder.WithCommonLabels(nil),
 		},
+		Type: corev1.SecretTypeTLS,
+		Data: map[string][]byte{
+			"tls.key": []byte(""),
+			"tls.crt": []byte(""),
```
akalenyu, line 223: @mhenriks this feels so wrong but but not sure if we have alternatives as we must be the ones creating them?

pkg/operator/controller/certrotation_test.go
```diff
@@ -116,6 +147,8 @@ var _ = Describe("Cert rotation tests", func() {
 
 			checkCerts(client, namespace, false)
 
+			populateClientAndStoreWithSecrets(client, cm.(*certManager), namespace)
+			syncStoreWithLocalUpdateCalls(client, cm.(*certManager), namespace)
```
akalenyu, line 151: @mhenriks looks like with this library-go change https://github.com/openshift/library-go/pull/1772 there's additional logic to call the CREATE.
not sure if this messes up functionality (should requeue on already exists err) but it definitely hurts the unit tests

I tried shifting the order a bit but it's clunky. Calling Sync() an extra time also doesn't help, seems like the lister is never updated and thus the library-go mechanism keeps trying to CREATE (which I guess makes sense since it's all fake)

akalenyu: I got further here but can't help feeling like we're doing something fundamentally wrong because I don't ever recall being in this problem space in kubevirt
(client updates/patches needing to be reflected on the objects in the `cache.Store`)

mhenriks: KubeVirt doesn't use this library

pkg/operator/resources/cert/factory.go
mhenriks: This doesn't offend me very much

PR #3628, opened 2025-02-05
Enable Tinycore Functest image multiarch

### LINKED ISSUE #3627: Exclude tinycore tests from s390x e2e test suite
When running s390x e2e tests, uses tinycore image which is not supported on s390x

label tinycore tests so we can exclude them explicitly from s390x e2e test suite

currently we use a work-around which is to use a regex in the manual ginkgo test call to exclude tinycore image; it's preferable to have it explicitly labelled; considered building tinycore image for s390x but that would be a bigger and external project with more permissions required.

**Additional context**:
only applies to s390x architecture.

Davo911:
Changeup some bazel code to be able to build the tinycore image needed for some functional tests for arm and s390x

Fixes #3627 

**Release note**:

```release-note
NONE
```


Davo911:
cc: @cfilleke

awels:
Just going to run through the tests before approving, looks fine to me.

Issue #2531, opened 2023-01-07
Support VolumeSnapshot as a volume source


mattlqx:
**Is your feature request related to a problem? Please describe**:
I'm looking at ways to minimize launch time for large images. I have images pre-imported into AWS EBS but using them as a source via PVC can be slow as it can take around 5-10 minutes to snapshot a 200GB volume.

**Describe the solution you'd like**:
Since the image volumes themselves don't change and restoring an EBS snapshot is quick, but taking one is slow. I'd like to just pre-create VolumeSnapshots and have KubeVirt create a new DataVolume from that. This should significantly reduce creation times.

**Describe alternatives you've considered**:
I've experimented around a lot with ZFS but snapshots have reference issues, detached snapshots are just about as slow and then you'd have to deal with locality issues of which node has which data and how to migrate between hosts. Using EBS largely solves the data location issues and scales well, but has attachment limits as a trade-off.

**Additional context**:
I'm not sure if this is appropriate here for CDI or if this is more of a kubevirt proper thing.

akalenyu:
Actively being worked on at https://github.com/kubevirt/containerized-data-importer/pull/2522,
feel free to review

alromeros:
@akalenyu @mattlqx  I think we can close this issue since https://github.com/kubevirt/containerized-data-importer/pull/2522 is already merged. WDYT?

awels:
Closing due to issue being solved

PR #3639, opened 2025-02-17
Clone populator: Add clone source watches

alromeros:

We lack proper clone source watches in clone populator, which can cause unwanted behaviors when we depend on source updates to trigger a reconcile.

This PR adds two watches for each clone source type.

Fixes # https://issues.redhat.com/browse/CNV-56518

**Release note**:

```release-note
Add clone source watches in clone populator
```


alromeros:
/cc @akalenyu @arnongilboa 
Sadly the test passes also without the fix, so it's useless as it is. Any idea to ensure this fixes the issue with gcp?

mhenriks:
@alromeros I'm not saying we shouldn't do this but we could do a cheap fix like this when the snapshot source does not exist:

https://github.com/kubevirt/containerized-data-importer/blob/249b004d13422b2ee52f800efe77ef9086a58321/pkg/controller/clone/csi-clone.go#L67

If you go with the watches, make sure to change the line above

akalenyu:
You probably have a second watch in that dynamic snap one https://github.com/kubevirt/containerized-data-importer/blob/249b004d13422b2ee52f800efe77ef9086a58321/pkg/controller/clone/planner.go#L179-L183 Should be able to remove this

akalenyu:
> /cc @akalenyu @arnongilboa Sadly the test passes also without the fix, so it's useless as it is. Any idea to ensure this fixes the issue with gcp?

Yeah testing watches is not something we do, I think you would use https://book.kubebuilder.io/reference/envtest.html but clearly it's not integrated at all in CDI today.
You could unit test the watch func or something like that, make sure it doesn't `return nil` when the snapshot source isn't CDI-owned.

akalenyu:
> @alromeros I'm not saying we shouldn't do this but we could do a cheap fix like this when the snapshot source does not exist:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/249b004d13422b2ee52f800efe77ef9086a58321/pkg/controller/clone/csi-clone.go#L67
> 
> If you go with the watches, make sure to change the line above

Interesting, yeah, you could do that in https://github.com/kubevirt/containerized-data-importer/blob/249b004d13422b2ee52f800efe77ef9086a58321/pkg/controller/clone/snap-clone.go#L62-L64 but it looks like the original intention was to use watches anyway with that TODO.
Should we be concerned about the hit we take with that new watch?

alromeros:
Removed both the func test and the requeue when source PVC is not ready.

arnongilboa:
Since the only thing we are interested in is the` VolumeSnapshot` update to `Ready` a few secs after it's created, why not go with @mhenriks RequeueAfter?

akalenyu:
> @alromeros since the only thing we are interested in is the` VolumeSnapshot` update to `Ready` a few secs after it's created, why not go with @mhenriks RequeueAfter?

You'd be spamming requeues for no reason, especially for provisioners where snap taking is long https://github.com/kubevirt/containerized-data-importer/issues/2531

akalenyu:
> > > @alromeros since the only thing we are interested in is the` VolumeSnapshot` update to `Ready` a few secs after it's created, why not go with @mhenriks RequeueAfter?
> > 
> > 
> > You'd be spamming requeues for no reason, especially for provisioners where snap taking is long #2531
> 
> Isn't watching for any update also potentially expansive?

It's not free but the alternative is occupying an entire thread on potentially useless requeues

> why not add a ReadyToUse == true predicate, so we won't reconcile when not needed?

I think it's a good idea, if that is the only thing we care about, we should be doing that

alromeros:
I prefer to keep requeue logic within a watch and handle all potential issues there.

> > > > @alromeros since the only thing we are interested in is the` VolumeSnapshot` update to `Ready` a few secs after it's created, why not go with @mhenriks RequeueAfter?
> > > 
> > > 
> > > You'd be spamming requeues for no reason, especially for provisioners where snap taking is long #2531
> > 
> > 
> > Isn't watching for any update also potentially expansive?
> 
> It's not free but the alternative is occupying an entire thread on potentially useless requeues
> 
> > why not add a ReadyToUse == true predicate, so we won't reconcile when not needed?
> 
> I think it's a good idea, if that is the only thing we care about, we should be doing that

By that logic wouldn't we might as well drop most of our watches and rely on requeues when necessary? I prefer to keep requeue logic within a watch and handle all potential issues there, but I'm not an expert in the potential performance implications. Just that having watches for sources it's what we do in the DV controller and makes sense to keep it here.

arnongilboa:
> > It's not free but the alternative is occupying an entire thread on potentially useless requeues
> > > why not add a ReadyToUse == true predicate, so we won't reconcile when not needed?
> > 
> > 
> > I think it's a good idea, if that is the only thing we care about, we should be doing that
> 
> By that logic wouldn't we might as well drop most of our watches and rely on requeues when necessary? I prefer to keep requeue logic within a watch and handle all potential issues there, but I'm not an expert in the potential performance implications. Just that having watches for sources it's what we do in the DV controller and makes sense to keep it here.

I meant adding the mentioned predicate in the watch like we do with Pending DVs.

pkg/controller/populators/clone-populator.go
```diff
@@ -81,6 +88,90 @@ type ClonePopulatorReconciler struct {
...
+	}
+	return false
+}
+
+func addCloneSourceWatches(mgr manager.Manager, c controller.Controller, log logr.Logger) error {
+	getKey := func(kind, namespace, name string) string {
+		return kind + "/" + namespace + "/" + name
+	}
+
+	if err := mgr.GetClient().List(context.TODO(), &snapshotv1.VolumeSnapshotList{}); err != nil {
```
akalenyu, line 111: So I don't know if this is a real world issue, but, if there is not snapshot support in the cluster, you'd back out and not set up a PVC watch

pkg/controller/populators/clone-populator.go
```diff
@@ -81,6 +88,90 @@ type ClonePopulatorReconciler struct {
...
+	}
+
+	// Register watches for all supported clone source types
+	for kind, obj := range supportedCloneSources {
+		if err := c.Watch(source.Kind(mgr.GetCache(), obj,
+			handler.EnqueueRequestsFromMapFunc(genericSourceMapper(kind)),
+			predicate.Funcs{
+				CreateFunc: func(e event.CreateEvent) bool { return true },
+				DeleteFunc: func(e event.DeleteEvent) bool { return false },
+				UpdateFunc: func(e event.UpdateEvent) bool { return isSourceReady(obj) },
```
akalenyu, line 165: I think this will requeue for any readyToUse snapshot. Instead, what you're after is a single case where readyToUse switches from false to true, so something like https://github.com/kubevirt/containerized-data-importer/blob/ce1e83c3e509d559cb5e61168860ac28d741a37c/pkg/controller/datasource-controller.go#L307-L309

akalenyu:
I think watchOwned could also use a predicate, I believe what happens today is that any snap goes through it's `EnqueueRequestsFromMapFunc`

pkg/controller/populators/clone-populator.go
alromeros: I assume that we still want to requeue any volume snapshot update even after it becomes ready? This would only allow for one single update requeue.

pkg/controller/populators/clone-populator.go
alromeros: I considered it but I don't know how much I should worry about this case... Don't mind handling it if necessary but code might become more complex.

pkg/controller/populators/clone-populator.go
akalenyu: The controller requeues the request if there's an error/some other reason. You only have to catch the transition

pkg/controller/populators/clone-populator.go
arnongilboa: Not so complex:
https://github.com/kubevirt/containerized-data-importer/blob/f4ddcc1748f8b2eeb14eb46e94b658a47e66445b/pkg/controller/dataimportcron-controller.go#L1199-L1207

alromeros: By complex I mean that it requires duplicating the watches and mappers instead of having a list of supported sources, but yeah I'll change it if we prefer to handle this error.

pkg/controller/populators/clone-populator.go
alromeros: ack

pkg/controller/populators/clone-populator.go
alromeros: ack, found a way to do this without duplicating the code, hope it's ok

alromeros:
> I think watchOwned could also use a predicate, I believe what happens today is that any snap goes through it's `EnqueueRequestsFromMapFunc`

I've been wondering about this, watchOwned cares also about PVCs and Pods and might also depend on updates other than snapshot becoming ready or PVC becoming bound. It's also unrelated to the bug/issue I'm trying to solve here so I'd prefer to leave it outside this PR unless necessary.

pkg/controller/populators/clone-populator.go
alromeros: @akalenyu @arnongilboa gentle ping

akalenyu (commented):
/approve

arnongilboa:
/lgtm

Great work!

PR #3642, opened 2025-02-24
Move cleartext destructive labels to ginkgo label filters

Davo911:
Now as Ginkgo label filters are enabled the cleartext bracket labels marking destructive tests should move too.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cdiconfig_test.go
+++ b/tests/cdiconfig_test.go
@@ -528,30 +528,28 @@ var _ = Describe("CDI route config tests", Serial, func() {
 	})
 })
 
-var _ = Describe("CDIConfig instance management", Serial, func() {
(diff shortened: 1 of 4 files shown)


Davo911:
Forgot to commit the changes in the test scripts to pick up the labels 
@akalenyu

Davo911:
Could we kick off the /test all again here @akalenyu

akalenyu:
sure!
/test all

akalenyu (commented):
Sorry I missed checking this out.
So I think we have outsiders relying on skipping the destructive test set using ginkgo.skip. Maybe we could go with duplicates for a while until those are fixed?

Davo911:
You are right, for compatibility, we should keep them both.
As they don't work together at the same time a check is neccesary though.

PR #3645, opened 2025-02-26
DataImportCron tests: add waitForConditions

jpeimer:
The test changes the DataImportCron source format and causes the re-import of the boot source images. If we do not wait for the import to finish - the next tests may flake and the cluster may look unhealthy. 

Fixes #

**Release note**:

```release-note

```


tests/dataimportcron_test.go
```diff
@@ -784,6 +784,7 @@ var _ = Describe("DataImportCron", Serial, func() {
 				},
 			}
 			Expect(dataSource.Spec.Source).To(Equal(expectedSource))
+			waitForConditions(corev1.ConditionFalse, corev1.ConditionTrue)
```
akalenyu, line 787: This just waits for the dataimportcron in the test to converge https://github.com/kubevirt/containerized-data-importer/blob/898e8946cc3221e927ce4b56783da756b254bcd6/tests/dataimportcron_test.go#L115-L119 Are you sure this is what you were after?

akalenyu (commented):
Thanks for submitting the PR!

tests/dataimportcron_test.go
jpeimer: I want to make sure that **all** dataimportcrons converged at the very end of the run. But honestly not sure if that's what my PR does :) will it work this way?

akalenyu: In that case you want to do that in the AfterEach
something like LIST all dataimportcrons, iterate over each one and ensure `waitForConditions(corev1.ConditionFalse, corev1.ConditionTrue)`

jpeimer: Okay, thanks! Giving it another try

tests/dataimportcron_test.go
```diff
@@ -95,6 +95,21 @@ var _ = Describe("DataImportCron", Serial, func() {
 				Satisfy(meta.IsNoMatchError),
 			))
 		}
+
+		// Wait for all DataImportCrons to converge
+		By("[AfterEach] Wait for DataImportCrons UpToDate")
+		dataImportCrons := &cdiv1.DataImportCronList{}
+		err = f.CrClient.List(context.TODO(), dataImportCrons, &client.ListOptions{Namespace: metav1.NamespaceAll, LabelSelector: l})
```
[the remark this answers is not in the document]
jpeimer: Oh right, DataImportCron itself doesn't have this label, will remove now

akalenyu:
The tests are failing since there is no explicit cron deletion in the after each
(rather, the global namespace deletion takes care of that)
So, negative tests leave a bad cron that will never converge

If you add
```go
err = f.CdiClient.CdiV1beta1().DataImportCrons(ns).Delete(context.TODO(), cronName, metav1.DeleteOptions{})
```
Before your check it should be fine

jpeimer:
@akalenyu I see... but this will not wait for the DataImportCron to be gone, so we may still pick it in the list. And in some tests, it may already be deleted?
Can I do something like
```
Eventually(func() bool {
	_ = f.CdiClient.CdiV1beta1().DataImportCrons(ns).Delete(context.TODO(), cronName, metav1.DeleteOptions{})
	_, err := f.CdiClient.CdiV1beta1().DataImportCrons(ns).Get(context.TODO(), cronName, metav1.GetOptions{})
	return errors.IsNotFound(err)
}, dataImportCronTimeout, pollingInterval).Should(BeTrue(), "DataImportCron was not deleted")
```

jpeimer:
> It's fine for it to get picked up in the list, just make sure the list is refreshed in the `Eventually` so it passes once it's gone
> (BTW I missed it completely that the current revision will keep looping over the same stale list)

So I can check in `Eventually` that if Get() returns `IsNotFound` then we're fine, but then `Eventually` starts to look too tangled... and we still can't expect error `ToNot(HaveOccurred()` upon deletion (because some tests could have delete it before us)

So maybe `delete -> wait deleted -> grab all -> wait for conditions` would be better?

tests/dataimportcron_test.go
```diff
@@ -95,6 +95,29 @@ var _ = Describe("DataImportCron", Serial, func() {
...
+		}, dataImportCronTimeout, pollingInterval).Should(BeTrue(), "DataImportCron was not deleted")
+
+		By("[AfterEach] Wait for all DataImportCrons UpToDate")
+		// Wait for all DataImportCrons to converge
+		dataImportCrons := &cdiv1.DataImportCronList{}
+		err = f.CrClient.List(context.TODO(), dataImportCrons, &client.ListOptions{Namespace: metav1.NamespaceAll})
+		Expect(err).ToNot(HaveOccurred())
+		for i := range dataImportCrons.Items {
+			cronItem := &dataImportCrons.Items[i]
+			Eventually(func() bool {
```
akalenyu, line 114: I meant this eventually loop isn't doing any polling since it always loops over the same `dataImportCrons.Items` list

jpeimer: So I delete the DataImportCron cronName (created by the tests), then Eventually waits for it to be deleted

Then I get the list of all, loop over it, get each by `cronItem := &dataImportCrons.Items[i]`, and then Eventually waits for the `cronItem`'s conditions (at least that's what I want)

akalenyu: yeah but you only fetch the list from the cluster once. so if any cron takes more time to converge, the eventually loop will not catch that, and just time out on a stale entry

jpeimer: Oh I seeee! `cronItem` is not really a reference, so we are not really polling the live object... let me try a new patch :) 

Should we also increase a timeout? Maybe 10 minutes? I'm afraid that 4 minutes may not be enough when we have a slow cluster and 6 concurrent imports.

Also changed a loop, using `for _, cronItem := range dataImportCrons.Items`

tests/dataimportcron_test.go
```diff
@@ -95,6 +96,30 @@ var _ = Describe("DataImportCron", Serial, func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		for _, cronItem := range dataImportCrons.Items {
+			Eventually(func() bool {
+				cron, err = f.CdiClient.CdiV1beta1().DataImportCrons(cronItem.Namespace).Get(context.TODO(), cronItem.Name, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				condProgressing := controller.FindDataImportCronConditionByType(cron, cdiv1.DataImportCronProgressing)
+				condUpToDate := controller.FindDataImportCronConditionByType(cron, cdiv1.DataImportCronUpToDate)
+				return condProgressing != nil && condProgressing.Status == corev1.ConditionFalse &&
+					condUpToDate != nil && condUpToDate.Status == corev1.ConditionTrue
+			}, dataImportCronConvergeTimeout, pollingInterval).Should(BeTrue(), fmt.Sprintf("Timeout waiting for DataImportCron conditions %q", cronItem.Namespace+"/"+cronItem.Name))
```
akalenyu, line 121: You can do
```go
.Should(BeTrue(), "Timeout waiting for DataImportCron conditions %q", cronItem.Namespace+"/"+cronItem.Name)
```
Since assertions will call fmt.Sprintf later on your behalf https://github.com/kubevirt/containerized-data-importer/blob/436b808dd4215d644040bb33fdd2b4fd59dd455a/vendor/github.com/onsi/gomega/gomega_dsl.go#L497-L501

jpeimer: Thanks! Done

akalenyu (approved):
/test all

PR #3651, opened 2025-03-02
Adding an if statement for the case that the certDir is not configured

omerpresler:
This PR introduces an additional if statement to ensure that proxy certificates remain available even when the certConfigMap field is not present.

Currently, if certConfigMap is missing while pulling from a registry, the proxy certificates are not utilized, which can lead to trust issues. This change ensures that proxy certificates are always accessible when configured, regardless of if certConfigMap field is present or not in the DataVolumeSourceRegistry section.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/importer/registry-datasource.go
+++ b/pkg/importer/registry-datasource.go
@@ -191,6 +191,12 @@ func CreateCertificateDir(registryCertDir string) (string, error) {
 	if err := collectCerts(common.ImporterProxyCertDir, allCerts, "proxy-"); err != nil {
 		return allCerts, err
 	}
+
+	if registryCertDir == "" {
+		klog.Info("Registry certs directory not configured")
+		return allCerts, nil
+	}
+
 	klog.Info("Copying registry certs")
 	if err := collectCerts(registryCertDir, allCerts, ""); err != nil {
 		return allCerts, err


omerpresler:
Looking at it again i see that it will also happen when the data source is http or imageIO

http: 
pkg > importer > http-datasource.go > createHTTPClient(...
will skip adding the proxy certs if certDir is empty 
ImageIO
calls pkg > importer > http-datasource.go > createCertPool
which will fail to open an empty string and fail to add the proxy certs too.

should i make a new pr with fixes http and imageIO, create a new pr with fixes to http, imageIO and registry and close this one or maybe i misunderstood and that's the intended behavior.

omerpresler:
Also in tools > cdi-source-update-poller > main.go > main
it looks like there is an assumption that if importer.CreateCertificateDir(certDir) failed then we can ignore the common cert path and assume the registry certificates path is okay

omerpresler:
> So the problem you are trying to solve that if the certDir is blank, it is right now returning an error since it can't copy from there?

Yes, so right now if you don't put the certConfigMap field inside the dataVolume yaml the code will return an error and wont copy/use the proxy certs. 
A workaround to that issue that i use right now is putting the field certConfigMap with the value "kube-root-ca.crt" even when i don't need that ca just so the importer pod will be able to use the proxy CA.

awels:
Sorry been on PTO and extremely busy, I don't see any issues with the code itself. Lets see what CI thinks.

PR #3663, opened 2025-03-06
Add GCP hyperdisk capabilities

arnongilboa:
- Add RWX Block to supported hyperdisk types
- Use snapshot for clone and for DataImportCron created boot sources

**Release note**:
```release-note
Add GCP hyperdisk capabilities
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -80,8 +80,9 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"kubernetes.io/azure-file": {{rwx, file}},
 	"file.csi.azure.com":       {{rwx, file}},
 	// GCE Persistent Disk
-	"kubernetes.io/gce-pd":  {{rwo, block}},
-	"pd.csi.storage.gke.io": {{rwo, block}},
+	"kubernetes.io/gce-pd":            {{rwo, block}},
+	"pd.csi.storage.gke.io":           {{rwo, block}},
+	"pd.csi.storage.gke.io/hyperdisk": {{rwx, block}, {rwo, block}, {rwo, file}},
 	// Hitachi
 	"hspc.csi.hitachi.com": {{rwx, block}, {rwo, block}, {rwo, file}},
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
Looks good, would love some background around why they don't want CSI clones
and why golden snapshots are preferred

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -368,6 +371,13 @@ var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageCl
 		}
 		return "ebs.csi.aws.com"
 	},
+	"pd.csi.storage.gke.io": func(sc *storagev1.StorageClass) string {
+		val := sc.Parameters["type"]
+		if val == "hyperdisk-balanced" {
```
akalenyu, line 376: Got bitten before by a similar parameter, are you sure exact match is a must here? Maybe something more forgiving fits better like strings.Contains or?

akalenyu: For instance if `hyperdisk-nonbalanced` is another option, you would fail here and fall back to non RWX.

akalenyu: Something like this is what worried me:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/d7c5a23550f4fb4d44ec6f38cf7abf69492b6898/pkg/common/utils.go#L736
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/d7c5a23550f4fb4d44ec6f38cf7abf69492b6898/pkg/common/parameters.go#L80

karkunpavan (approved):
LGTM

pkg/storagecapabilities/storagecapabilities.go
arnongilboa: Per discussion with GCP folks and according to their [ref](https://cloud.google.com/compute/docs/disks/modify-hyperdisks#limits), I prefer adding RWX explicitly to the specific types. We will sure extend the coverage in the future if needed.

Issue #3621, opened 2025-01-29
Smart Clone not possible with Windows disk


nick123pig:
**What happened**:

When attempting a CSI snapshot clone with a windows disk, I am unable to clone a virtual machine. Host-assisted clone still works fine. The mount command is what fails.

```
Events:
  Type     Reason       Age                  From     Message
  ----     ------       ----                 ----     -------
  Warning  FailedMount  71s (x538 over 18h)  kubelet  MountVolume.MountDevice failed for volume "pvc-5cc66ec2-9c7d-438e-a2f1-fd753b1c8427" : rpc error: code = Internal desc = mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t ext4 -o defaults /dev/longhorn/pvc-5cc66ec2-9c7d-438e-a2f1-fd753b1c8427 /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/f98632dd5d308fa99779cd6b9f3055766314ae829e84fc801c0c9493b1d62eab/globalmount
Output: mount: /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/f98632dd5d308fa99779cd6b9f3055766314ae829e84fc801c0c9493b1d62eab/globalmount: wrong fs type, bad option, bad superblock on /dev/longhorn/pvc-5cc66ec2-9c7d-438e-a2f1-fd753b1c8427, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
```

**What you expected to happen**:

I would expect it to be able to smart clone the disk

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a windows VM 
2. Clone it using a basic volume snapshot smart clone
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: windows-vm-test-snap
  namespace: csi-snapshot-test
spec:
  volumeSnapshotClassName: longhorn-snapshot-vsc
  source:
    persistentVolumeClaimName: windows-vm-test
```
3. Create a new vm with the snapshot source
```
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: windows-vm-test-snap
spec:
  running: false
  template:
    metadata:
      labels:
        kubevirt.io/domain: windows-vm-test-snap
    spec:
      domain:
        firmware:
          bootloader:
            efi: 
              secureBoot: false
        features:
          acpi: {}
          apic: {}
[...]

PR #3664, opened 2025-03-09
Clone: fallback to host-assisted when snapshot volume mode differs (#3621)

noamasu:
When creating a DataVolume from a snapshot with a specified target PVC, configuration mismatches can occur. For example, if the snapshot is configured with block mode and the target PVC is set to filesystem mode, the VolumeSnapshotContent's SourceVolumeMode will differ from the target PVC's volume mode. In this scenario, the cloning process should fall back to the host-assisted strategy. This ensures that cloning is performed by the CDI code rather than by a third-party driver, which may not correctly handle the conversion and could lead to mounting failures as seen in this bug.

Fixes https://github.com/kubevirt/containerized-data-importer/issues/3621

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/clone/common.go
+++ b/pkg/controller/clone/common.go
@@ -10,11 +10,13 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
@@ -126,6 +128,18 @@ func GetGlobalCloneStrategyOverride(ctx context.Context, c client.Client) (*cdiv
 	}
 
 	return cr.Spec.CloneStrategyOverride, nil
+}
+
+// GetSnapshotContentFromSnapshot returns the VolumeSnapshotContent of a given VolumeSnapshot
+func GetSnapshotContentFromSnapshot(ctx context.Context, c client.Client, snapshot *snapshotv1.VolumeSnapshot) (*snapshotv1.VolumeSnapshotContent, error) {
+	if snapshot.Status == nil || snapshot.Status.BoundVolumeSnapshotContentName == nil {
+		return nil, fmt.Errorf("volumeSnapshotContent name not found")
+	}
+	vsc := &snapshotv1.VolumeSnapshotContent{}
+	if err := c.Get(ctx, types.NamespacedName{Name: *snapshot.Status.BoundVolumeSnapshotContentName}, vsc); err != nil {
+		return nil, err
+	}
+	return vsc, nil
 }
 
 // GetStorageClassForClaim returns the storageclass for a PVC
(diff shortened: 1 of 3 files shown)


akalenyu (commented):
Looks good, thank you!
Couple of comments

pkg/controller/clone/common.go
```diff
@@ -280,6 +294,20 @@ func SameVolumeMode(pvc1 *corev1.PersistentVolumeClaim, others ...*corev1.Persis
...
 }
 
+// SameVolumeModeSnapshot returns true if all pvcs have the same volume mode as the source snapshot
+func SameVolumeModeSnapshot(ctx context.Context, client client.Client, vsc *snapshotv1.VolumeSnapshotContent, others ...*corev1.PersistentVolumeClaim) bool {
+	if vsc.Spec.SourceVolumeMode == nil {
+		return false
+	}
+
+	for _, pvc := range others {
+		if *pvc.Spec.VolumeMode != *vsc.Spec.SourceVolumeMode {
```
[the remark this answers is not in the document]
noamasu: `ResolveVolumeMode` will return `v1.PersistentVolumeFilesystem` if the volumeMode is nil.
in case `pvc.Spec.VolumeMode:=nil`  and `vsc.Spec.SourceVolumeMode:=v1.PersistentVolumeFilesystem` (they are not the same)  then  `SameVolumeModeSnapshot` will return true, is that desired?

pkg/controller/clone/common.go
```diff
@@ -128,6 +130,18 @@ func GetGlobalCloneStrategyOverride(ctx context.Context, c client.Client) (*cdiv
 	return cr.Spec.CloneStrategyOverride, nil
 }
 
+// GetSnapshotContentFromSnapshot returns the VolumeSnapshotContent of a given VolumeSnapshot
+func GetSnapshotContentFromSnapshot(ctx context.Context, c client.Client, snapshot *snapshotv1.VolumeSnapshot) (*snapshotv1.VolumeSnapshotContent, error) {
```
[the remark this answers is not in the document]
noamasu: I agree, will do.

pkg/controller/common/util.go
```diff
@@ -1806,19 +1806,12 @@ func ValidateSnapshotCloneSize(snapshot *snapshotv1.VolumeSnapshot, pvcSpec *cor
 }
 
 // ValidateSnapshotCloneProvisioners validates the target PVC storage class against the snapshot class provisioner
-func ValidateSnapshotCloneProvisioners(ctx context.Context, c client.Client, snapshot *snapshotv1.VolumeSnapshot, storageClass *storagev1.StorageClass) (bool, error) {
+func ValidateSnapshotCloneProvisioners(ctx context.Context, c client.Client, snapshot *snapshotv1.VolumeSnapshot, vsc *snapshotv1.VolumeSnapshotContent, storageClass *storagev1.StorageClass) (bool, error) {
```
akalenyu, line 1809: do you still need the snapshot passed here, post changes?

pkg/controller/clone/common.go
```diff
@@ -128,6 +130,18 @@ func GetGlobalCloneStrategyOverride(ctx context.Context, c client.Client) (*cdiv
...
+	if snapshot.Status == nil || snapshot.Status.BoundVolumeSnapshotContentName == nil {
+		return nil, fmt.Errorf("volumeSnapshotContent name not found")
+	}
+	vsc := &snapshotv1.VolumeSnapshotContent{}
+	if err := c.Get(ctx, types.NamespacedName{Name: *snapshot.Status.BoundVolumeSnapshotContentName}, vsc); err != nil {
```
ShellyKa13, line 139: dont you need to add here the namespace?

pkg/controller/clone/planner.go
```diff
@@ -414,6 +418,14 @@ func (p *Planner) computeStrategyForSourceSnapshot(ctx context.Context, args *Ch
 		p.fallbackToHostAssisted(args.TargetClaim, res, NoVolumeExpansion, MessageNoVolumeExpansion)
 		return res, nil
 	}
+
+	// Lastly, do volume mode validation to determine whether to use dumb or smart cloning
+	if !SameVolumeMode(vsc.Spec.SourceVolumeMode, args.TargetClaim) {
+		p.fallbackToHostAssisted(args.TargetClaim, res, IncompatibleVolumeModes, MessageIncompatibleVolumeModes)
+		args.Log.V(3).Info("Volume modes not compatible for advanced clone - Snapshot")
```
ShellyKa13, line 425: Maybe message should be similar to the provisioners print? something like:
"source and target volume modes differs, need to fall back to host assisted"

pkg/controller/common/util.go
ShellyKa13: +1

ShellyKa13 (commented):
looks good! question and small nits

pkg/controller/common/util.go
noamasu: right!
i think `ctx context.Context, c client.Client` are also not needed. removing.

pkg/controller/clone/common.go
noamasu: I think that since we are getting a VolumeSnapshotContent which is a cluster-wide resource, no need for namespace

ShellyKa13: right right, my bad

akalenyu:
/test pull-cdi-unit-test
not related https://github.com/kubevirt/containerized-data-importer/pull/3673

akalenyu (commented):
great work thx!
/approve

PR #3665, opened 2025-03-10
WIP Add SupportLevel to StorageProfile status

arnongilboa:
The field indicates CDI support level per storage class, which can be either supported, unsupported or unknown. It can be used by UI for filtering the available storage classes.

**Release note**:
```release-note
Add SupportLevel to StorageProfile status
```

### DIFF
--- a/pkg/apis/core/v1beta1/openapi_generated.go
+++ b/pkg/apis/core/v1beta1/openapi_generated.go
@@ -19234,6 +19234,13 @@ func schema_pkg_apis_core_v1beta1_StorageProfileStatus(ref common.ReferenceCallb
 							Format:      "",
 						},
 					},
+					"supportLevel": {
+						SchemaProps: spec.SchemaProps{
+							Description: "SupportLevel indicates CDI support level for the storage class",
+							Type:        []string{"string"},
+							Format:      "",
+						},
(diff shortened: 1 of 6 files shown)


awels:
So why can't we simply say, if the status has an empty `claimPropertySet` it is unsupported since it can't be used by CDI, regardless of it is because it is an unknown provider or if it is a provider that is known to not work with CDI. The end result is the same.

arnongilboa:
> So why can't we simply say, if the status has an empty `claimPropertySet` it is unsupported since it can't be used by CDI, regardless of it is because it is an unknown provider or if it is a provider that is known to not work with CDI. The end result is the same.

Referring the status `claimPropertySet` was my initial direction, but note it won't allow UI to differentiate between officially-covered `StorageProfiles` and those manually defined (e.g. green/yellow/red bullet). If we are ok with that, I agree this no-api-change is the best. Another alternative is adding a label instead of the `SupportLevel` field, which will allow UI to filter via `List()` label selector instead of iterating the `StorageProfiles`. wdyt?

awels:
I think we are okay with not differentiating between blank storage profiles and unsupported ones. Effectively they are the same from the user perspective, neither is usable.

PR #3673, opened 2025-03-12
don't start informer in cert rotation unit tests

akalenyu:
one is not supposed to run informers on a fake client
https://github.com/kubernetes/kubernetes/pull/95897
https://github.com/kubernetes-csi/external-attacher/blob/e32b64308dcb147c6f5a714d1f7e71e0b91a2319/pkg/controller/framework_test.go#L160-L162

Hopefully this is finally fixing the cert rotation unit tests

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/operator/controller/certrotation_test.go
+++ b/pkg/operator/controller/certrotation_test.go
@@ -32,6 +32,25 @@ var secretNames = []string{
 	"cdi-uploadserver-signer", "cdi-uploadserver-client-signer", "cdi-uploadserver-client-cert",
 }
 
+type noInformerStartCertManager struct {
+	*certManager
+}
+
+func (cm *noInformerStartCertManager) Start(ctx context.Context) error {
+	for _, ns := range cm.namespaces {
+		if cm.listerMap == nil {
+			cm.listerMap = make(map[string]*certListers)
+		}
+
+		cm.listerMap[ns] = &certListers{
+			secretLister:    cm.informers.InformersFor(ns).Core().V1().Secrets().Lister(),
+			configMapLister: cm.informers.InformersFor(ns).Core().V1().ConfigMaps().Lister(),
+		}
+	}
+
+	return nil
+}
+
 type fakeCertManager struct {
 	client    client.Client
 	namespace string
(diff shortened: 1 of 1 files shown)


akalenyu:
/cc @mhenriks 
:melting_face:

pkg/operator/controller/certrotation_test.go
```diff
@@ -32,6 +32,25 @@ var secretNames = []string{
 	"cdi-uploadserver-signer", "cdi-uploadserver-client-signer", "cdi-uploadserver-client-cert",
```
ShellyKa13, line 32: do you mind explaining in the commit message why? better the needing to get inside the link and also whats linked does explain exactly the issue.

akalenyu: it's a rabbit hole, have to follow the readouts. but mostly its the fake client not supporting resourceVersion

pkg/operator/controller/certrotation_test.go
```diff
@@ -81,7 +103,10 @@ func getCertConfigAnno(client kubernetes.Interface, namespace, name string) stri
 	s, err := client.CoreV1().Secrets(namespace).Get(context.TODO(), name, metav1.GetOptions{})
 	Expect(err).ToNot(HaveOccurred())
 	val, ok := s.Annotations["operator.cdi.kubevirt.io/certConfig"]
-	Expect(ok).To(BeTrue())
+	Expect(ok).To(BeTrue(), func() string {
+		secretJSON, _ := json.MarshalIndent(s, "", "    ")
+		return string(secretJSON)
```
mhenriks, line 108: pardon my ignorance but what is the side effect of this function here? What happens to the result?

akalenyu: that's fine it's not common in our projects, it's evaluated in case of failure to allow for a more descriptive error msg https://github.com/kubevirt/containerized-data-importer/blob/673e15a77170d1be9ba9bc554b032f4368b67e1a/vendor/github.com/onsi/gomega/gomega_dsl.go#L497-L508

pkg/operator/controller/certrotation_test.go
```diff
@@ -32,6 +32,25 @@ var secretNames = []string{
 	"cdi-uploadserver-signer", "cdi-uploadserver-client-signer", "cdi-uploadserver-client-cert",
 }
 
+type noInformerStartCertManager struct {
+	*certManager
+}
+
+func (cm *noInformerStartCertManager) Start(ctx context.Context) error {
+	for _, ns := range cm.namespaces {
+		if cm.listerMap == nil {
```
ShellyKa13, line 41: I would have put the if outside of the for

ShellyKa13 (commented):
One nit but otherwise
/lgtm

PR #3692, opened 2025-04-08
Fix cdi-deployment Deployment selector

Acedus:
The cdi-deployment selector has a matchLabels value of {"app": "containerized-data-importer"} which is a common label that appears on many CDI related Pods (e.g., cdi-apiserver, cdi-uploadproxy, importer, etc.).

This causes certain functionalities such as the log subresource to misbehave as it targets the first Pod of the Pod list according to the ByLogging sorting conditions[1].

This commit changes the cdi-deployment selector to the common "cdi.kubevirt.io" with the value "cdi-deployment".

Since deployment selectors are immutable[2], it also adds a reconcile callback to delete the cdi-deployment Deployment in the event of detecting a selector that differs from the operator's desired one.

[1] https://github.com/kubernetes/kubectl/blob/fec9d5b3d5b1c3addf8cebd3c2437257ea5e4332/pkg/util/podutils/podutils.go#L96
[2] https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#selector

Partially fixes [CNV-58338](https://issues.redhat.com/browse/CNV-58338)
 
Note that the reason why rollout and general functionality is retained despite the selector is mismatch is because of the ReplicaSet's `pod-template-hash`. 

**Release note**:

```release-note
cdi-deployment Deployment selector changed to "cdi.kubevirt.io":"cdi-deployment"
```


pkg/operator/controller/callbacks.go
```diff
@@ -89,6 +115,14 @@ func reconcileDeleteControllerDeployment(args *callbacks.ReconcileCallbackArgs)
 		return nil
 	}
 
+	if err := deleteControllerDeployment(deployment, args); err != nil {
```
akalenyu, line 118: wouldn't `deleteControllerDeployment` whack unrelated worker pods related to CDI? I don't think you care about those?

Acedus: You're absolutely right, fixing.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/01fbb541ec95fd143c6aef58a61f608db92be7da..ccb611d42fffe78d25baaff505cc2e549b2e57bf) - Addressed @akalenyu comment.

pkg/operator/controller/callbacks.go
```diff
@@ -68,6 +69,31 @@ func isControllerDeployment(d *appsv1.Deployment) bool {
...
+	}
+
+	deployment := args.DesiredObject.(*appsv1.Deployment)
+
+	if !isControllerDeployment(deployment) {
+		return nil
+	}
+
+	if !reflect.DeepEqual(args.CurrentObject.(*appsv1.Deployment).Spec.Selector.MatchLabels, deployment.Spec.Selector.MatchLabels) {
+		args.Logger.Info("Mismatching selector detected for CDI deployment, cleaning up and requeuing")
```
akalenyu, line 88: do you actually requeue here? if I'm reading correctly an error only gets returned if the deletion fails

Acedus: Yeah you're right, technically it would eventually reconcile due to the missing deployment (when it should exist) but in the wrong reconcile state and for the wrong reason. I'll change it.

pkg/operator/controller/callbacks.go
```diff
@@ -68,6 +69,32 @@ func isControllerDeployment(d *appsv1.Deployment) bool {
...
+		if err := deleteControllerDeployment(deployment, args); err != nil {
+			return err
+		}
+		return fmt.Errorf("requeuing CDI deployment selector update")
```
mhenriks, line 92: Why is delete required? The deployment can't be updated?

Assuming delete is required I don't think it is necessary to force requeue by returning an error since the operator is watching resources owned by CDI such as this deployment

Acedus: A deployment's selector is an immutable field, ergo the deletion and recreation.

You're right that the resource will eventually reconcile anyway post-deletion but requeuing prevents additional error logs from ramping up and should be generally more correct IMO. WDYT?

mhenriks: What additional errors? Do you just mean more noise in the operator log? Because I don't think returning the error here is going to reduce that much.

A problem with returning an error is that the reconcile will happen again right away. And it is very possible that the deployment is still int he system and not gone yet

At the end of the day it doesn't matter much but I don't see how returning the error is more correct

Acedus: Ack, reverting.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/1c30f6ab7d734a1fd9de2f45ab6b3cae08e05d6a..80f4bba37c3acef8d8ca2775e24fe50edbbf8065) - Addressed @mhenriks comments.

akalenyu:
pvname appearing late flake

pkg/operator/resources/namespaced/controller.go
```diff
@@ -175,7 +175,7 @@ func createControllerServiceAccount() *corev1.ServiceAccount {
 func createControllerDeployment(controllerImage, importerImage, clonerImage, ovirtPopulatorImage, uploadServerImage, verbosity, pullPolicy string, imagePullSecrets []corev1.LocalObjectReference, priorityClassName string, infraNodePlacement *sdkapi.NodePlacement, replicas int32) *appsv1.Deployment {
 	defaultMode := corev1.ConfigMapVolumeSourceDefaultMode
 	// The match selector is immutable. that's why we should always use the same labels.
-	deployment := utils.CreateDeployment(common.CDIControllerResourceName, common.CDILabelKey, common.CDILabelValue, common.ControllerServiceAccountName, imagePullSecrets, int32(1), infraNodePlacement)
+	deployment := utils.CreateDeployment(common.CDIControllerResourceName, common.CDIComponentLabel, common.CDIControllerResourceName, common.ControllerServiceAccountName, imagePullSecrets, int32(1), infraNodePlacement)
```
mhenriks, line 178: Can we keep the existing label in the Deployment `spec.metadata.`, but add the new label and update the matchLabel in `spec.selector` to be the new one?

Acedus: Perhaps I misunderstood but that should already be the case.
The `spec.metadata` labels are already populated with both the existing `app: containerized-data-importer` and `cdi.kubevirt.io: cdi-deployment`. It's just the selector that matches the incorrect label of the former instead of the latter.

mhenriks: okay, got it

PR #3694, opened 2025-04-09
Bump kvci to avoid loopback NFS

akalenyu:
Current NFS CSI setup runs a high risk of ending up with loopback nfs mounts (share mounted on same host exporting it)
which is known to cause deadlocks https://www.suse.com/support/kb/doc/?id=000018709

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
@brianmcarey do you happen to know the block size on the original volume source for the CI's `/var/lib/containers` mount?
This would help me understand the discrepancy that causes NFS actual size to be much larger than content size

```
/dev/sda4 on /var/lib/containers type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,sunit=512,swidth=512,prjquota)
```

tests/upload_test.go
```diff
@@ -388,8 +388,6 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 					same, err := f.VerifyTargetPVCContentMD5(f.Namespace, archivePVC, pathInPvc, expectedMd5)
 					Expect(err).ToNot(HaveOccurred())
 					Expect(same).To(BeTrue())
-					By("Verifying the image is sparse")
```
Acedus, line 391: Can you elaborate on where it's getting stuck?
Did we try to forcibly free cache and then running verification to see if it reproduces?

tests/import_test.go
```diff
@@ -1700,8 +1700,6 @@ var _ = Describe("Import populator", func() {
 		md5, err := f.GetMD5(f.Namespace, pvc, utils.DefaultImagePath, utils.MD5PrefixSize)
 		Expect(err).ToNot(HaveOccurred())
 		Expect(md5).To(Equal(utils.TinyCoreMD5))
-		By("Verifying the image is sparse")
```
Acedus, line 1719: nit: Consider squashing this commit with the previous one.

Acedus (commented):
Thanks.
I'm a bit confused as to how this PR avoids loopback NFS, can you clarify?

tests/import_test.go
akalenyu: I've kept them separate on purpose since the latter is about preallocation

tests/upload_test.go
akalenyu: so I don't think we can force that speculative space trimming on demand
(if it ends up being that, I am not sure it is)

> Can you elaborate on where it's getting stuck?

yeah, what happens is that the file is not sparse (check evaluates to false, content<size on disk)
unless you loop over the check for 3 minutes.
I tested all sorts of combinations like --sparse on the original uploaded archive and it's still the same

But the gist of all of this is really about whether we should care about the resulting extracted archive files not being sparse? I don't think we can guarantee how folks tar their files (--sparse or not)

akalenyu:
> Thanks. I'm a bit confused as to how this PR avoids loopback NFS, can you clarify?

Totally right I attached the context to the PR description. We started using a container on the host for the NFS server instead of a pod inside the k8s node.

tests/upload_test.go
Acedus: Fair ebough.

Acedus (approved):

akalenyu:

I **think** the `ENOENT` comes from https://github.com/kubevirt/containerized-data-importer/blob/3df4efdfc20f9446b694c45360986a94563a281d/pkg/image/qemu.go#L140-L142 We used to be okay with notexist in the device checks https://github.com/kubevirt/containerized-data-importer/blob/3df4efdfc20f9446b694c45360986a94563a281d/pkg/importer/file.go#L95-L97 But I don't understand why this doesn't fail more often

akalenyu:
So seems like this path only gets exercised because OOMs are occurring often 
```
2025-05-29T19:44:22Z	DEBUG	controller.import-controller	Pod died of an OOM, deleting pod, and restarting with qemu cache mode=none if storage supports it	{"PVC": {"name":"prime-5ec5bd58-8e68-4d3f-9213-78b0311d6a2e","namespace":"cdi-e2e-tests-importer-m8gfj"}, "pod.Name": "importer-prime-5ec5bd58-8e68-4d3f-9213-78b0311d6a2e"}
```
/hold

akalenyu:
discussed with @arnongilboa, feel like we have an idea where it's coming from
/test pull-containerized-data-importer-e2e-istio
spamming this a bit to ensure this PR doesn't make things worse

akalenyu:
unrelated - https://github.com/kubevirt/containerized-data-importer/pull/3694#issuecomment-2935565009

PR #3695, opened 2025-04-09
Bugfix/qemu img convert lose sparseness

### LINKED ISSUE #3614: Importing a containerdisk onto a block volume loses sparseness
**What happened**:
Importing a containerdisk onto a block volume loses sparseness. When I imported the centos-stream:9 containerdisk, which only uses 2 GB of non-zero data onto an empty 10 GB block volume, all 10 GB were written by CDI. Preallocation was not enabled.

Only the non-zero data should be written to the block volume. This saves space on the underlying storage.

Create a DataVolume from the YAML below and observe the amount of storage allocated. I used KubeSAN as the CSI driver, so the LVM `lvs` command can be used to see the thin provisioned storage usage. If you don't have thin provisioned storage you could use I/O stats or tracing to determine how much data is being written.

**Additional context**:
I discussed this with @aglitke and we looked at the qemu-img command that is invoked:
```
Running qemu-img with args: [convert -t writeback -p -O raw /scratch/disk/disk.img /dev/cdi-block-volume]
```

Adding the `--target-is-zero` option should avoid writing every block in the target block volume.

If there are concerns that some new block volumes come uninitialized (blocks not zeroed), then it should be possible to run `blkdiscard --zeroout /path/to/block/device` before invoking qemu-img with `--target-is-zero`. I have not tested this, but blkdiscard should zero the device efficiently and fall back to writing zero buffers on old hardware. On modern devices this would still be faster and preserve sparseness compared to writing all zeroes. On old devices it would be slower, depending on how many non-zero the input disk image has.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): 4.17.3
- Kubernetes version (use `kubectl version`): v1.30.5
- DV specification:
```
kind: DataVolume
metadata:
  annotations:
    cdi.kubevirt.io/storage.import.lastUseTime: "2025-01-22T20:19:34.821435785Z"
    cdi.kubevirt.io/storage.usePopulator: "true"
  creationTimestamp: "2025-01-22T20:19:34Z"
  generation: 1
  labels:
    app.kubernetes.io/part-of: hyperconverged-cluster
    app.kubernetes.io/version: 4.17.3
    cdi.kubevirt.io/dataImportCron: centos-stream-9-test-import-cron-3vs0wc
    instancetype.kubevirt.io/default-instancetype: u1.medium
    instancetype.kubevirt.io/default-preference: centos.stream9
  name: centos-stream-9-test-9515270a7f3d
  namespace: default
  resourceVersion: "217802868"
  uid: d8d56487-5686-47d0-93d0-79de02a8c2c3
spec:
  source:
    registry:
      url: docker://quay.io/containerdisks/centos-stream@sha256:9515270a7f3d3fd053732c15232071cb544d847e56aa2005f27002014b5becaa
  storage:
    resources:
      requests:
        storage: 10Gi
    storageClassName: kubesan
```
- Install tools: N/A
- Others: N/A

noamasu:
This PR updates the `qemu-img convert` command by conditionally adding the `--target-is-zero `and `-n` flags when preallocation is not requested. Since we assume that the target block device is already zeroed, the flag allows qemu-img to skip writing zero blocks—reducing unnecessary I/O and speeding up image conversion. The `-n` flag (which is used to skip target image creation) is required together with `--target-is-zero`, so it is automatically added when preallocation is disabled.

Fixes #3614

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -107,9 +107,11 @@ func convertToRaw(src, dest string, preallocate bool, cacheMode string) error {
 	}
 
 	args := []string{"convert", "-t", cacheMode, "-p"}
-	// We assume that the destination block is already zeroed (empty),
-	// so we include the '-n' flag with '--target-is-zero' to skip unnecessary writes.
-	args = append(args, "-n", "--target-is-zero")
+	if !preallocate {
+		// We assume that the destination block is already zeroed (empty),
+		// so we include the '-n' flag with '--target-is-zero' to skip unnecessary writes.
+		args = append(args, "-n", "--target-is-zero")
+	}
 	args = append(args, "-O", "raw", src, dest)
 
 	if preallocate {
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -105,7 +105,12 @@ func convertToRaw(src, dest string, preallocate bool, cacheMode string) error {
 	if err != nil {
 		return err
 	}
-	args := []string{"convert", "-t", cacheMode, "-p", "-O", "raw", src, dest}
+
+	args := []string{"convert", "-t", cacheMode, "-p"}
+	// We assume that the destination block is already zeroed (empty),
+	// so we include the '-n' flag with '--target-is-zero' to skip unnecessary writes.
+	args = append(args, "-n", "--target-is-zero")
+	args = append(args, "-O", "raw", src, dest)
 
 	if preallocate {
 		err = addPreallocation(args, convertPreallocationMethods, func(args []string) ([]byte, error) {
--- a/pkg/image/qemu_test.go
+++ b/pkg/image/qemu_test.go
@@ -213,7 +213,7 @@ var _ = Describe("Convert to Raw", func() {
 		})
 	})
 
-	It("should add preallocation if requested", func() {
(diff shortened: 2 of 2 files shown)


akalenyu:
@noamasu following offline discussions, should we hold this for now?
I guess we're not in the wrong to not use this flag but rather device mapper devices return ENOSUPP on the appropriate ioctl, where other storage providers don't?

noamasu:
Hey @akalenyu,
I guess we should hold this for now, at least until we know whats going on with the ENOSUPP when using thin lv, not sure if this is the only one with ENOSUPP.

but worth mentioning that
using --target-is zero will always reduce I/O and amount of sectors written to the block device. in fact, it will write the minimum amount of I/O and sectors possible which is always what we want. on an ideal scenario, if we knew 100% that the target device is zeroed out, it is worth adding this flag.

but yes, looks like most devices actually do preserve sparseness even when not using --target-is-zero (except that thin lv ENOSUPP), in that case not adding the flag will just mean more I/O  and sectors written but at least the target block device conversion result is reliable (results in the same sha256sum disk signature even if the disk is not zeroed).

PR #3699, opened 2025-04-10
Fix rpm deps generation

akalenyu:
Couple of things wrong with current rpm-deps target. See commits for explanation

Fixes #

**Release note**:

```release-note
NONE
```


repo.yaml
```diff
@@ -7,10 +7,6 @@ repositories:
   baseurl: http://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/
   name: centos/stream9-appstream-x86_64
   gpgkey: https://www.stream.centos.org/keys/RPM-GPG-KEY-CentOS-Official
-- arch: x86_64
-  baseurl: https://download.copr.fedorainfracloud.org/results/nsoffer/ovirt-imageio-preview/centos-stream-9-x86_64/
-  name: ovirt-imageio-preview
-  gpgkey: https://download.copr.fedorainfracloud.org/results/nsoffer/ovirt-imageio-preview/pubkey.gpg
 - arch: x86_64
   baseurl: https://download.copr.fedorainfracloud.org/results/ovirt/ovirt-master-snapshot/centos-stream-9-x86_64/
```
Acedus, line 11: Why not change this to the equivalent `https://mirror.stream.centos.org/SIGs/9-stream/virt/x86_64/ovirt-45/` as was done for aarch64?

Acedus (approved):
LGTM, just had a small question.

repo.yaml
akalenyu: good question, I let it be since I understood it's going to remain in place (unlike arm). I don't mind switching @bennyz WDYT?

bennyz: I don't know what differences between the centos repo and the copr currently are, if the centos repo has the same `ovirt-img` package, then it probably makes more sense to use the centos one

akalenyu: I prefer we keep this one for the time being, we can always switch

akalenyu:
/approve cancel
this is my own lol

/cc @awels

PR #3702, opened 2025-04-15
change terminationMessagePolicy to FallbackToLogsOnError for infra Pods 

Acedus:

The only Pods which explicitly write to /dev/termination-log (and therefore make use of the default File terminationMessagePolicy) are cdi-importer, cdi-upload-server and cdi-cloner. This means that for infra deployments such as cdi-deployment and cdi-apiserver, there is no termination message[1].

In order for the latter Pods to report information upon termination, the `terminationMessagePolicy` has been changed to `FallbackToLogsOnError`.

[1] https://github.com/kubernetes/kubernetes/blob/30469e180361d7da07b0fee6d47c776fa2cf3e86/pkg/kubelet/kuberuntime/kuberuntime_container.go#L596

Fixes:
- https://issues.redhat.com/browse/CNV-58253

**Release note**:

```release-note
Termination message now gets reported correctly for CDI infra Pods.
```


akalenyu (commented):
> By setting the terminationMessagePolicy to "FallbackToLogsOnError", you can tell Kubernetes to use the last chunk of container log output if the termination message file is empty and the container exited with an error

According to this, wouldn't the importer work well with this policy too, since (hopefully) the termination message file is never empty?

Acedus:
> > By setting the terminationMessagePolicy to "FallbackToLogsOnError", you can tell Kubernetes to use the last chunk of container log output if the termination message file is empty and the container exited with an error
> 
> According to this, wouldn't the importer work well with this policy too, since (hopefully) the termination message file is never empty?
> 
> https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message

That's fair enough, I'll change it for all other Pods as well.

pkg/operator/resources/utils/common.go
```diff
@@ -54,7 +54,7 @@ var ResourceBuilder = utils.NewResourceBuilder(commonLabels, operatorLabels)
 // CreateContainer creates container
 func CreateContainer(name, image, verbosity, pullPolicy string) corev1.Container {
 	container := ResourceBuilder.CreateContainer(name, image, pullPolicy)
-	container.TerminationMessagePolicy = corev1.TerminationMessageReadFile
+	container.TerminationMessagePolicy = corev1.TerminationMessageFallbackToLogsOnError
```
akalenyu, line 57: Might want to unit tests these (could probably do that in existing tests) since we're now "complying" with the platform and we don't want to accidentally lose that status

Acedus: I’m okay with adding the test, but it feels like it sets an unusual precedent. In the same notion, why don’t we also have a test ensuring readiness probes are properly defined, since that’s also part of the compliance manifest?

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/a34b94bafe5501d5284c3b194d9409572cf5ec5f..bf5fb1b6c826e99539bdb9199248051d5108dd87) - Set the `terminationMessagePolicy` for all containers including ones that write to `/dev/termination-log`.

pkg/operator/resources/utils/common.go
akalenyu: > I’m okay with adding the test, but it feels like it sets an unusual precedent

It's not that unusual - https://github.com/kubevirt/containerized-data-importer/pull/3646

> In the same notion, why don’t we also have a test ensuring readiness probes are properly defined, since that’s also part of the compliance manifest?

We'd probably benefit from having that unit tested as well though I get it's out of scope for this change

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/bf5fb1b6c826e99539bdb9199248051d5108dd87..db9c5e9c3edfb3facfed8cd0546c76e9ad19e928) - Addressed @akalenyu comments.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/db9c5e9c3edfb3facfed8cd0546c76e9ad19e928..ea067ce7a0c99b2c90cd51af2a31a1150f6a2640) - Addressed @akalenyu comment.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/b1e0fe77ac6636ca22ae7e1416d98a84ca1f6fc4..35880c6a104046f5511e5fb0772e29272e15e5f7) - Rebase.

PR #3711, opened 2025-04-23
Allow increasing PVC size to the minimum supported by its storage class

arnongilboa:
Supports both `DataVolume` PVC rendering and PVC mutating webhook rendering, so it can cover the kubevirt-side created `TPM` small PVC and any other PVC created by `DataVolume` or independently. For independent PVCs they should be labeled with `cdi.kubevirt.io/applyStorageProfile: "true"` and CDI featureGate `WebhookPvcRendering` should be enabled. A `StorageProfile` can be annotated with its minimum supported volume size with e.g.:
`cdi.kubevirt.io/minimumSupportedPvcSize: 4Gi`

Fixes #

**Release note**:
```release-note
Allow increasing PVC size to the minimum supported by its storage class
```

### DIFF
--- a/pkg/controller/datavolume/util.go
+++ b/pkg/controller/datavolume/util.go
@@ -78,7 +78,7 @@ func RenderPvc(ctx context.Context, client client.Client, pvc *v1.PersistentVolu
 	}
 
 	isClone := pvc.Annotations[cc.AnnCloneType] != ""
-	return renderPvcSpecVolumeSize(client, &pvc.Spec, isClone)
+	return renderPvcSpecVolumeSize(client, &pvc.Spec, isClone, nil)
 }
 
 // renderPvcSpec creates a new PVC Spec based on either the dv.spec.pvc or dv.spec.storage section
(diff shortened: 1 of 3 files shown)


pkg/controller/datavolume/util.go
```diff
@@ -297,6 +300,21 @@ func renderPvcSpecVolumeSize(client client.Client, pvcSpec *v1.PersistentVolumeC
 	if err != nil {
 		return err
 	}
+
+	if scName := pvcSpec.StorageClassName; scName != nil {
+		storageClass := &storagev1.StorageClass{}
+		if err := client.Get(context.TODO(), types.NamespacedName{Name: *scName}, storageClass); err == nil {
+			if val, exists := storageClass.Annotations[AnnMinimumSupportedPVCSize]; exists {
+				minSize, err := resource.ParseQuantity(val)
+				if err == nil && requestedSize.Cmp(minSize) == -1 {
```
alromeros, line 309: I think it would be helpful to log or return an error if parsing the annotation fails, similar to what was done in the [kubevirt PR](https://github.com/kubevirt/kubevirt/pull/13800/files#diff-2013a813861b9274b3272f633523779e9b73b6db8b21ea2806c8de05d847ac26R441). This way users can be notified if the annotation is set incorrectly.

awels (commented):
looks good, just the one comment by Alvaro.

awels (approved):
/approve

PR #3723, opened 2025-05-04
Allow more restricted volume plugins in our scc

akalenyu:
Our SCC is a bit tighter when it comes to volume plugins,
even compared to the restricted-v2 consensus SCC.
We should be in the clear to allow exactly the same as it does.

This also prevents CDI from breaking in an event that
something injects csi inline volumes into it's pods.

Fixes #

**Release note**:

```release-note
BugFix: CDI pods rejected due to it's SCC not allowing some consensus volume plugins
```

### DIFF
--- a/pkg/operator/controller/controller_test.go
+++ b/pkg/operator/controller/controller_test.go
@@ -243,15 +243,19 @@ var _ = Describe("Controller", func() {
 				Expect(scc.Labels[common.AppKubernetesPartOfLabel]).To(Equal("testing"))
 				Expect(scc.Priority).To(BeNil())
 
-				for _, eu := range []string{"system:serviceaccount:cdi:cdi-sa"} {
-					found := false
-					for _, au := range scc.Users {
-						if eu == au {
-							found = true
-						}
-					}
-					Expect(found).To(BeTrue())
-				}
+				Expect(scc.Users).To(ContainElement("system:serviceaccount:cdi:cdi-sa"))
+
+				Expect(scc.Volumes).To(ConsistOf(
+					secv1.FSTypeConfigMap,
+					secv1.FSTypeDownwardAPI,
+					secv1.FSTypeEmptyDir,
+					secv1.FSTypePersistentVolumeClaim,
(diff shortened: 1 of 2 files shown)


akalenyu:
restricted-v2 SCC definition:
```yaml
$ oc get scc restricted-v2 -o yaml
allowHostDirVolumePlugin: false
allowHostIPC: false
allowHostNetwork: false
allowHostPID: false
allowHostPorts: false
allowPrivilegeEscalation: false
allowPrivilegedContainer: false
allowedCapabilities:
- NET_BIND_SERVICE
apiVersion: security.openshift.io/v1
defaultAddCapabilities: null
fsGroup:
  type: MustRunAs
groups: []
kind: SecurityContextConstraints
metadata:
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
    kubernetes.io/description: restricted-v2 denies access to all host features and
      requires pods to be run with a UID, and SELinux context that are allocated to
      the namespace. This is the most restrictive SCC and it is used by default for
      authenticated users. On top of the legacy 'restricted' SCC, it also requires
      to drop ALL capabilities and does not allow privilege escalation binaries. It
      will also default the seccomp profile to runtime/default if unset, otherwise
      this seccomp profile is required.
  generation: 1
  name: restricted-v2
priority: null
readOnlyRootFilesystem: false
requiredDropCapabilities:
- ALL
runAsUser:
  type: MustRunAsRange
seLinuxContext:
  type: MustRunAs
seccompProfiles:
- runtime/default
supplementalGroups:
  type: RunAsAny
users: []
volumes:
- configMap
- csi
- downwardAPI
- emptyDir
- ephemeral
- persistentVolumeClaim
- projected
- secret
```

akalenyu:
> /lgtm
> 
> @akalenyu can you please add reference to the `restricted-v2 consensus SCC` and to the issue it fixes.

done

PR #3750, opened 2025-05-16
add build functest shebang

Davo911:
I've been running recently into
```bash
eval ./hack/build/build-functest.sh
./hack/build/build-functest.sh: 15: set: Illegal option -o pipefail
```
Adding a shebang solved that, so I'd include this for compatibility.

**Release note**:

```release-note
NONE
```


akalenyu (commented):
/approve
/lgtm

PR #3753, opened 2025-05-18
DNM: Support architecture specific image import for registry datasource

Acedus:
This PR implements CDI support for architecture specific image import using the registry datasource as outlined in VEP https://github.com/kubevirt/enhancements/pull/49.

Partially implements https://github.com/kubevirt/enhancements/pull/49
Addresses https://issues.redhat.com/browse/CNV-61527

- This PR depends on the merge of https://github.com/kubevirt/containerized-data-importer/pull/3745 for functional tests to run properly.
- `DataSource` adjustments as outlined in the aforementioned VEP are out-of-scope for this PR.

**Release note**:

```release-note
Introduce support for architecture specific image import with registry datasource
```


Acedus:
/cc @akalenyu 

fyi: @nunnatsa

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/a64c0a057cca6138430709fc0f6da8a548ea2576..261b902c73d47b770520ad698b3163fb5c78c5d7) - Update DV condition in the event that the importer Pod is unschedulable.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -199,6 +199,15 @@ type DataVolumeSourceRegistry struct {
...
 	CertConfigMap *string `json:"certConfigMap,omitempty"`
+	//Platform describes the minimum runtime requirements of the image
+	// +optional
+	Platform *PlatformOptions `json:"platform,omitempty"`
+}
+
+type PlatformOptions struct {
+	//Architecture specifies the image target CPU architecture
+	// +optional
+	Architecture string `json:"architecture,omitempty"`
```
akalenyu, line 210: I guess we're going to start off light, only exposing arch? My main worry is that variant is going to be needed but I guess we can extend for that
https://github.com/opencontainers/image-spec/blob/c05acf7eb327dae4704a4efe01253a0e60af6b34/specs-go/v1/descriptor.go#L53

tests/import_test.go
```diff
@@ -2153,13 +2154,123 @@ var _ = Describe("pull image failure", func() {
...
+		err = utils.WaitForDataVolumePhase(f, f.Namespace.Name, phase, dv.Name)
+		Expect(err).ToNot(HaveOccurred())
+	},
+		Entry("multi-arch image matching non-host architecture with pull method pod",
+			cdiv1.PlatformOptions{Architecture: archNotHost()},
+			tinyCoreMultiarchRegistry,
+			cdiv1.RegistryPullPod,
+		),
+		Entry("multi-arch image matching host architecture with pull method pod",
+			cdiv1.PlatformOptions{Architecture: runtime.GOARCH},
```
akalenyu, line 2225: So using runtime.GOARCH will couple the test executing machine to the target cluster.
While initially that should be fine, it's going to hit us with cryptic errors as soon as someone decides to start saving $ on the test executing binary, so, arm64 executors running against x86 for example
(from my experience, this is already happening outside of kvci)

Maybe passing it as a test flag to the test suite will be more robust? default to amd64?

akalenyu (commented):
really high level overview since I am short on time but generally looks good!

tests/import_test.go
Acedus: So this is kind of an odd scenario where `runtime.GOARCH` might make the most sense for *some* of the scenarios here.

Ideally we'd just specify "amd64" and call it a day, we're testing for behavior, after all. But specifically for something like single-arch image failure scenarios we also assert the error message which can vary according to the `populate-registry` tarball that's layered in based on the architecture where it was built. 

This generally means that the architecture of the single-arch image will always be equal to `runtime.GOARCH`.
I do agree that for all other cases we should stick with "amd64" and not introduce any additional logic (such as that `archMap`) to avoid potential quirks.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
Acedus: I disagree but for a different reason from my previous one.
Kubernetes currently doesn't support variant level scheduling as it labels nodes with the `kubernetes.io/arch` label according to `runtime.GOARCH`[1] and it doesn't use any `cpuinfo` shenanigans like some of its delegated CRIs do[2] to determine the variant. Since Kubernetes doesn't offer any labels for distinguishing node processor architecture variants, we probably shouldn't support it (yet) either.

[1] https://github.com/kubernetes/kubernetes/blob/82db38a23c7820b1924d89f458fd368023f3980c/pkg/kubelet/kubelet_node_status.go#L308
[2] https://github.com/containers/image/blob/3152ce9a40792572a5536accef3aaf4ca90a44f0/internal/pkg/platform/platform_matcher.go#L171

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/901573bd512204e830faf5977e598c66e1985b6f..5896320d2792017176e4e5f2ce43a9853db15084) - Rebase.

tests/import_test.go
```diff
@@ -2153,13 +2154,120 @@ var _ = Describe("pull image failure", func() {
...
+	)
+	var (
+		f                         = framework.NewFramework(namespacePrefix)
+		tinyCoreRegistry          = func() string { return fmt.Sprintf(tinyCoreSingleArchRegistryURL, f.CdiInstallNs) }
+		tinyCoreMultiarchRegistry = func() string { return fmt.Sprintf(utils.TinyCoreIsoRegistryURL, f.CdiInstallNs) }
+		trustedRegistryURL        = func() string { return fmt.Sprintf(utils.TrustedRegistryURL, f.DockerPrefix) }
+	)
+
+	DescribeTable("Should succeed to pull image with platform options", func(platform cdiv1.PlatformOptions, urlFn func() string, pullMethod cdiv1.RegistryPullMethod) {
+		dv := utils.NewDataVolumeWithRegistryImport("multi-arch-pull", "10Gi", urlFn())
```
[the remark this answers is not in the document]
Acedus: Yep, reducing to match the majority of other tests, thanks.

tests/import_test.go
Acedus: We could infer the architecture and name of a node in the cluster which would be our testing dummy so the test would always be deterministic, this could be flaky with heterogeneous clusters, though, but I doubt it'd be an issue considering other processes (such as the registry image generation) would rely on homogeneous clusters as well.

It will come at the cost of:
1. Querying the API for a node.
2. Updating CDI nodeSelector temporarily.

We already do both of these as part of `import_test` so it shouldn't be too significant, WDYT?

akalenyu: > I think the main issue is with the `single-arch image matching architecture with pull method Pod`

Ah, I see... yeah I overlooked that one a bit. I won't feel too bad dropping it

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/5896320d2792017176e4e5f2ce43a9853db15084..415d002b106d78bb14e62ec4b0171572226b399d) - Addressed @akalenyu's comments.

tests/import_test.go
```diff
@@ -2199,13 +2199,110 @@ var _ = Describe("pull image failure", func() {
...
+			dv.Annotations = make(map[string]string)
+		}
+		pullMethod := cdiv1.RegistryPullNode
+		dv.Annotations[controller.AnnImmediateBinding] = "true"
+		dv.Spec.Source.Registry.PullMethod = &pullMethod
+		dv.Spec.Source.Registry.Platform = &cdiv1.PlatformOptions{Architecture: "absent"}
+
+		cm, err := utils.CopyRegistryCertConfigMap(f.K8sClient, f.Namespace.Name, f.CdiInstallNs)
+		Expect(err).ToNot(HaveOccurred())
+		dv.Spec.Source.Registry.CertConfigMap = &cm
```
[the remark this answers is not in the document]
Acedus: You're right, oversight from copying the stanza, removing.

Acedus:
> Looks really good and probably the implementation won't change, but maybe it makes sense to wait until the vep merges before lgtming this? i guess [kubevirt/enhancements#49](https://github.com/kubevirt/enhancements/pull/49) just needs one final pass before being approved.

I generally agree and don't mind holding this PR until the VEP goes through. With that said, all remaining unresolved comments are on none-CDI issues so the final design for this section in particular shouldn't change.

mhenriks:
/hold

Waiting for [VEP](https://github.com/kubevirt/enhancements/pull/49) to merge

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/pull/3753/commits/b7b67f64ce51186b9feadacecad7a913efd91440) - Added docs on multi-platform image pull support.

Acedus:
VEP has been merged, unholding.
/unhold

alromeros:
Great job!
/lgtm
/test pull-cdi-generate-verify

akalenyu:
I think it jsut takes a while for pvc capacity to populate

PR #3760, opened 2025-05-22
VEP48: Introduce DataSource Source DataSource (DataSource Pointers)

Acedus:
This PR is part of the implementation of https://github.com/kubevirt/enhancements/issues/48 for multi-architecture golden image import support.
In this PR, a new DataSource is added in addition to the existing `Snapshot` and `PVC` sources aptly named `DataSource`, in functionality one could think of it as a "DataSource Pointer".

Pointers (`Source.DataSource`) allow one DataSource to point to the source of another DataSource with some limitations:
1. The DataSource may not self-reference, this is validated at the API level.
2. The DataSource pointer reference chain may not exceed a maxium depth of 1.

jira-ticket: https://issues.redhat.com/browse/CNV-61527

**Release note**:

```release-note
Allow DataSources to reference the source of another DataSource using new Source.DataSource option
```


pkg/apiserver/apiserver.go
```diff
@@ -219,6 +221,11 @@ func NewCdiAPIServer(bindAddress string,
 		return nil, errors.Errorf("failed to create Populator validating webhook: %s", err)
 	}
 
+	err = app.createDataSourceValidatingWebhook()
```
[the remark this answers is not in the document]
Acedus: Sure, since we don't validate anything else I don't mind adding it to the resolution conditions.
I do think shifting everything to admission time should prove to be a better user-experience, mutable or not, as currently we allow users to accidentally make a configuration mistake.

pkg/controller/common/util.go
```diff
@@ -2074,3 +2074,31 @@ func AllowClaimAdoption(c client.Client, pvc *corev1.PersistentVolumeClaim, dv *
 	}
 	return featuregates.NewFeatureGates(c).ClaimAdoptionEnabled()
 }
+
+// ResolveDataSourceChain resolves a DataSource reference until there's either
```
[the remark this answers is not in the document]
Acedus: Fair enough, done.

pkg/controller/common/util.go
```diff
@@ -2074,3 +2074,31 @@ func AllowClaimAdoption(c client.Client, pvc *corev1.PersistentVolumeClaim, dv *
...
+
+		next := &cdiv1.DataSource{}
+		if err := client.Get(ctx, types.NamespacedName{Name: name, Namespace: ns}, next); err != nil {
+			return current, err
+		}
+		current = next
+	}
+
+	if current.Spec.Source.DataSource != nil {
+		return current, fmt.Errorf("DataSource reference chain exceeds maximum depth of %d", maxDepth)
```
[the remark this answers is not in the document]
Acedus: Done.

pkg/controller/datasource-controller.go
```diff
@@ -75,20 +76,30 @@ func (r *DataSourceReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
+	if resolved != nil && !reflect.DeepEqual(dataSource.Status.Source, resolved.Spec.Source) {
+		resolved.Spec.Source.DeepCopyInto(&dataSource.Status.Source)
 		dataSource.Status.Conditions = nil
 	}
+
 	dataSourceCopy := dataSource.DeepCopy()
-	if sourcePVC := dataSource.Spec.Source.PVC; sourcePVC != nil {
-		if err := r.handlePvcSource(ctx, sourcePVC, dataSource); err != nil {
+	switch {
+	case resolved.Spec.Source.DataSource != nil && err != nil:
```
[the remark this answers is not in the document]
Acedus: Done.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/62d1579004786b522bb92863a23622decc40d882..611311d808a0a665de8c6bafb1716683dad9bb41) - Addressed @arnongilboa's comments.

pkg/apiserver/apiserver.go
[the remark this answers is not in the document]
awels: The way I understand it, if you have to look up the state of the cluster at validation time it is usually better to push the check into the controller. But any static validation that does not require the state of the cluster it is better to do it in the webhook since it will give the user earlier feedback. But if you look up the cluster state, by the time you have reported the error back to the user, the state could have changed where it is no longer an issue. 

For example if someone applies a bunch of resources all at once, and a webhook fails intermittently because sometimes the resources are applied in the wrong order.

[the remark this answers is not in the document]
arnongilboa: we don't validate much simpler cases here - e.g. empty/illegal name for the source DataSource/PVC/VolumeSnapshot, so why validate specifically this esoteric case?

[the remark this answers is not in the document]
awels: I am just explaining my understanding on when to validate in the webhook vs doing it in the controller. I missed the actual code so I can't really tell.

pkg/controller/datasource-controller.go
```diff
@@ -75,20 +78,41 @@ func (r *DataSourceReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
+		switch {
+		case errors.Is(err, cc.ErrDataSourceMaxDepthReached):
+			reason = maxReferenceDepthReached
+		case errors.Is(err, cc.ErrDataSourceSelfReference):
+			reason = selfReference
+		case k8serrors.IsNotFound(err):
+			reason = cc.NotFound
+		default:
+			return err
+		}
```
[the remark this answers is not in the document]
Acedus: I considered that, looks a bit awkward IMO but I don't mind doing it anyhow.

Acedus:
/hold

Waiting for [VEP](https://github.com/kubevirt/enhancements/pull/49) to merge.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/7dca7608d0a249e563f115d8e51f48ae4a928368..2a254adce50afe46f987315cb931a53c44a6e999) - Addressed @arnongilboa's comment.

Acedus:
VEP has been merged, unholding.
/unhold

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/2a254adce50afe46f987315cb931a53c44a6e999..12708c5ad0bc67eadb206f5d80fd4f2dee925fc9) - Rebase

akalenyu:
Failure in https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3760/pull-containerized-data-importer-e2e-nfs/1937039108996075520 looks related

Acedus:
Looks like an odd flake, going to try something.
/hold

Acedus:
> Failure in https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3760/pull-containerized-data-importer-e2e-nfs/1937039108996075520 looks related

Kind of suspecting a temporary CI issue failed that run but let's see.
/retest-required

akalenyu:
> > Failure in https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3760/pull-containerized-data-importer-e2e-nfs/1937039108996075520 looks related
> 
> Kind of suspecting a temporary CI issue failed that run but let's see. /retest-required

Happened for both nfs/ceph lanes so at the very least it's going to be hard to get through CI

pkg/controller/import-controller.go
```diff
@@ -548,6 +548,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim)
 	pod, err := createImporterPod(context.TODO(), r.log, r.client, podArgs, r.installerLabels)
 	// Check if pod has failed and, in that case, record an event with the error
 	if podErr := cc.HandleFailedPod(err, pvc.Annotations[cc.AnnImportPod], pvc, r.recorder, r.client); podErr != nil {
+		r.log.V(1).Error(err, "failed creating pod", "pod.Name", pod.Name)
```
akalenyu: ugh seems like log for cdi-controller was rotated. got bit by that too https://github.com/kubevirt/containerized-data-importer/pull/3795/commits/8d6167b833e4d48c06576c000774359395e37d8b

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/56c4520b2a3f731815a10efa873ce01d97980776..eaf7a0070631d4f8d565b309387243ed5cfd602e) - 90% sure last PR accidentally lets `AnnRunningCondition` take precedence over `AnnPodSchedulable`. If this fixes it I'll submit a follow-up PR to address that race.

Acedus:
/hold
Until https://github.com/kubevirt/containerized-data-importer/pull/3799 gets merged.

pkg/controller/datasource-controller.go
```diff
@@ -75,20 +78,33 @@ func (r *DataSourceReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
+	dataSourceCopy := dataSource.DeepCopy()
+	resolved, err := cc.ResolveDataSourceChain(ctx, r.client, dataSource)
+	if err != nil {
+		log := r.log.WithValues("datasource", dataSource.Name, "namespace", dataSource.Namespace)
+		log.Info(err.Error())
+		if err := handleDataSourceRefError(dataSource, err); err != nil {
+			return err
+		}
 		dataSource.Spec.Source.DeepCopyInto(&dataSource.Status.Source)
+		resolved = dataSource
```
arnongilboa: why are we handling it as a non-error? if `Status.Source` was a pointer wouldn't you expect it to be `nil` in this case?

pkg/controller/datasource-controller.go
```diff
@@ -75,20 +78,33 @@ func (r *DataSourceReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
+	} else if !reflect.DeepEqual(dataSource.Status.Source, resolved.Spec.Source) {
```
arnongilboa: why waste time on the `DeepEqual()` check? we usually don't check the `Status` in `update()`.

Acedus: We want to update the the datasource's status to reflect the resolved datasource instead of the one that it has in its spec, this is one way to achieve that.

EDIT: Emphasizing that this is to avoid repeating `DeepCopy` in the event that `resolved` is equal to `dataSource`.

pkg/controller/datasource-controller.go
Acedus: I'm a bit confused as to what you mean. We're handling the error by updating the relevant condition if it's a non-fatal one and proceed to set the datasource status to its spec as we could not successfully resolve the chain. 

If you believe that the status should be empty if we encounter an error that's also reasonable, I don't mind either.

arnongilboa: Of course the conditions must be updated in case of such error, but not sure about updating `Status.Source`. @akalenyu wdyt?

akalenyu: yeah I think propagating the spec to status is a bit misleading in case of errors

Acedus: Let me clarify that this is existing behavior. We populate the status according to the spec even before this change regardless of success or failure.

akalenyu: do we reach the status update call today when there's an err?

Acedus: Upon DS reconciliation we copy the spec to the status if they're not equal:
https://github.com/kubevirt/containerized-data-importer/blob/e40a2e544c5455a9fa5bb0e13103356c9ebf8473/pkg/controller/datasource-controller.go#L77-L81

pkg/controller/datasource-controller.go
arnongilboa: Afaik since `DeepEqual()` uses reflection it's more expansive than the `DeepCopy()` call here, so removing the `if` will make it overall cheaper.

pkg/controller/datasource-controller.go
arnongilboa: So maybe it's time to fix it (I'm the one to blame for the original lines) :)

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/8fdac3cf153ab37bdcebc9335b0c55082249ed71..cbad774175000038f4b93443a042689a32b1d984) - Addressed @arnongilboa's comments. I'm still not sure how I feel about not setting the status.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -505,6 +513,8 @@ type DataSourceSource struct {
 	PVC *DataVolumeSourcePVC `json:"pvc,omitempty"`
 	// +optional
 	Snapshot *DataVolumeSourceSnapshot `json:"snapshot,omitempty"`
+	// +optional
+	DataSource *DataSourceRefSourceDataSource `json:"dataSource,omitempty"`
```
akalenyu, line 520: Could be nice to add a few words about max depth/self reference here so they show up in `kubectl explain`

pkg/controller/datavolume/controller-base.go
```diff
@@ -403,11 +403,16 @@ func getSourceRefOp(ctx context.Context, log logr.Logger, dv *cdiv1.DataVolume,
 		log.Error(err, "Unable to get DataSource", "namespacedName", nn)
 		return dataVolumeNop
 	}
+	resolved, err := cc.ResolveDataSourceChain(ctx, client, dataSource)
+	if err != nil {
+		log.Error(err, "Unable to get DataSource", "namespacedName", nn)
```
akalenyu: Unable to get DataSource->Unable to resolve DataSource chain?

pkg/controller/datasource-controller.go
akalenyu: Ah, I see what you mean now, even prior to this PR we'd reach the UPDATE call for some errors.
> I'm still not sure how I feel about not setting the status.

Same here, not sure which is the right one "philosophically".
But as long as the conditions tell the consumer it can't be trusted, we should be covered.

akalenyu (commented):
:+1:

pkg/controller/datavolume/controller-base.go
Acedus: Good catch.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/cbad774175000038f4b93443a042689a32b1d984..54d2c24) - Addressed @akalenyu's comments.

Acedus:
Looks related to previous [webhook rendering failures ](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3760/pull-containerized-data-importer-e2e-upg/1939641430142095360) @arnongilboa, seems like something around https://github.com/kubevirt/containerized-data-importer/blob/e40a2e544c5455a9fa5bb0e13103356c9ebf8473/pkg/controller/datavolume/util.go#L201 is flaking.

/retest-required

akalenyu:
for what it's worth I'm looking at the NFS failures in https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3760/pull-containerized-data-importer-e2e-nfs/1939641429701693440
at a separate PR https://github.com/kubevirt/containerized-data-importer/pull/3795. I suspect directly using the CI storage with no abstractions (NFS CSI setup we currently have) is just prone to errors.

PR #3764, opened 2025-05-28
Populate DV with events from PVC Prime

Dsanatar:
Add new functionality to watch for any new events that come from a prime PVC, these events will be re-emitted to the regular PVC along with a prefix to indicate the prime PVC's name. Additionally while the DV's bound condition is set to pending, the prime PVC name as well as the latest event message from the prime PVC will be appended to the message which will allow for easier debugging if a provisioning problem occurs.

```release-note
DV will now contain events from PVC Prime and DV bound condition message will be updated to contain latest PVC Prime event message while bound condition is pending"
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -351,6 +351,9 @@ const (
 
 	// AnnCreatedForDataVolume stores the UID of the datavolume that the PVC was created for
 	AnnCreatedForDataVolume = AnnAPIGroup + "/createdForDataVolume"
+
+	// AnnPVCPrimeName annotation is the name of the PVC' that is added to the target PVC
+	AnnPVCPrimeName = AnnAPIGroup + "/storage.populator.pvcPrime"
 )
 
 // Size-detection pod error codes
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -352,7 +352,7 @@ const (
 	// AnnCreatedForDataVolume stores the UID of the datavolume that the PVC was created for
 	AnnCreatedForDataVolume = AnnAPIGroup + "/createdForDataVolume"
 
-	// AnnPVCPrimeName annotation is the name of the PVC' that is added to the target PVC
+	// AnnPVCPrimeName annotation is the name of the PVC' that is used to populate the PV which is then rebound to the target PVC
 	AnnPVCPrimeName = AnnAPIGroup + "/storage.populator.pvcPrime"
 )
 
@@ -2116,4 +2116,107 @@ func ResolveDataSourceChain(ctx context.Context, client client.Client, dataSourc
 	}
 
 	return resolved, nil
+}
+
+// UpdatePVCBoundContionFromEvents updates the bound condition annotations on the PVC based on recent events
+// This function can be used by both controller and populator packages to update PVC bound condition information
+func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client.Client, log logr.Logger) error {
+	currentPvcCopy := pvc.DeepCopyObject()
+
+	anno := pvc.GetAnnotations()
+	if anno == nil {
+		return nil
+	}
+
+	if IsBound(pvc) {
+		anno := pvc.GetAnnotations()
+		delete(anno, AnnBoundCondition)
+		delete(anno, AnnBoundConditionReason)
+		delete(anno, AnnBoundConditionMessage)
+
+		if !reflect.DeepEqual(currentPvcCopy, pvc) {
+			if err := c.Update(context.TODO(), pvc); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}
+
+	if pvc.Status.Phase != corev1.ClaimPending {
+		return nil
+	}
+
+	// set bound condition by getting the latest event
+	events := &corev1.EventList{}
+
+	err := c.List(context.TODO(), events,
+		client.InNamespace(pvc.GetNamespace()),
+		client.MatchingFields{"involvedObject.name": pvc.GetName(),
+			"involvedObject.uid": string(pvc.GetUID())},
+	)
+
+	if err != nil {
+		// Log the error but don't fail the reconciliation
+		log.Error(err, "Unable to list events for PVC bound condition update", "pvc", pvc.Name)
+		return nil
+	}
+
+	if len(events.Items) == 0 {
+		return nil
+	}
+
+	pvcPrime, exists := anno[AnnPVCPrimeName]
+
+	// Sort event lists by containing primeName substring and most recent timestamp
+	sort.Slice(events.Items, func(i, j int) bool {
+		if exists {
+			firstContainsPrime := strings.Contains(events.Items[i].Message, pvcPrime)
+			secondContainsPrime := strings.Contains(events.Items[j].Message, pvcPrime)
+
+			if firstContainsPrime && !secondContainsPrime {
+				return true
+			}
+			if !firstContainsPrime && secondContainsPrime {
+				return false
+			}
+		}
+
+		// if the timestamps are the same just make sure our sorting is deterministic
+		if events.Items[i].LastTimestamp.Time.Equal(events.Items[j].LastTimestamp.Time) {
+			return events.Items[i].Message < events.Items[j].Message
+		}
+
+		// if both contains primeName substring or neither, just sort on timestamp
+		return events.Items[i].LastTimestamp.Time.After(events.Items[j].LastTimestamp.Time)
+	})
+
+	boundMessage := ""
+
+	// check if prime name annotation exists
+	if exists {
+		// if we are using populators get the latest event from prime pvc
+		pvcPrime = fmt.Sprintf("[%s] : ", pvcPrime)
+
+		// TODO get index after pvcPrime to get message
+
+		// split so we can remove prime name prefix from event message
+		res := strings.Split(events.Items[0].Message, pvcPrime)
+		boundMessage = res[len(res)-1]
+
+		if boundMessage == "" {
+			log.V(1).Info("No bound message found, skipping bound condition update")
+			return nil
+		}
+	} else {
+		// if not using populators just get the latest event
+		boundMessage = events.Items[0].Message
+	}
+
+	// since we checked status of phase above, we know this is pending
+	anno[AnnBoundCondition] = "false"
+	anno[AnnBoundConditionReason] = "Pending"
+	anno[AnnBoundConditionMessage] = boundMessage
+
+	return nil
 }
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -2135,7 +2135,8 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
 		delete(anno, AnnBoundConditionMessage)
 
 		if !reflect.DeepEqual(currentPvcCopy, pvc) {
-			if err := c.Update(context.TODO(), pvc); err != nil {
+			patch := client.MergeFrom(currentPvcCopy.(*corev1.PersistentVolumeClaim))
+			if err := c.Patch(context.TODO(), pvc, patch); err != nil {
 				return err
 			}
 		}
@@ -2214,6 +2215,11 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
 	anno[AnnBoundCondition] = "false"
 	anno[AnnBoundConditionReason] = "Pending"
 	anno[AnnBoundConditionMessage] = boundMessage
+
+	patch := client.MergeFrom(currentPvcCopy.(*corev1.PersistentVolumeClaim))
+	if err := c.Patch(context.TODO(), pvc, patch); err != nil {
+		return err
+	}
 
 	return nil
 }
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -2198,16 +2198,13 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
 		// if we are using populators get the latest event from prime pvc
 		pvcPrime = fmt.Sprintf("[%s] : ", pvcPrime)
 
-		// TODO get index after pvcPrime to get message
-
-		// split so we can remove prime name prefix from event message
-		res := strings.Split(events.Items[0].Message, pvcPrime)
-		boundMessage = res[len(res)-1]
-
-		if boundMessage == "" {
-			log.V(1).Info("No bound message found, skipping bound condition update")
+		// if the first event does not contain a prime message, none will so return
+		primeIdx := strings.Index(events.Items[0].Message, pvcPrime)
+		if primeIdx == -1 {
+			log.V(1).Info("No bound message found, skipping bound condition update", "pvc", pvc.Name)
 			return nil
 		}
+		boundMessage = events.Items[0].Message[primeIdx+len(pvcPrime):]
 	} else {
 		// if not using populators just get the latest event
 		boundMessage = events.Items[0].Message
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -2118,6 +2118,31 @@ func ResolveDataSourceChain(ctx context.Context, client client.Client, dataSourc
 	return resolved, nil
 }
 
+func sortEvents(events *corev1.EventList, usingPopulator bool, pvcPrimeName string) {
+	// Sort event lists by containing primeName substring and most recent timestamp
+	sort.Slice(events.Items, func(i, j int) bool {
+		if usingPopulator {
+			firstContainsPrime := strings.Contains(events.Items[i].Message, pvcPrimeName)
+			secondContainsPrime := strings.Contains(events.Items[j].Message, pvcPrimeName)
+
+			if firstContainsPrime && !secondContainsPrime {
+				return true
+			}
+			if !firstContainsPrime && secondContainsPrime {
+				return false
+			}
+		}
+
+		// if the timestamps are the same, prioritze longer messages to make sure our sorting is deterministic
+		if events.Items[i].LastTimestamp.Time.Equal(events.Items[j].LastTimestamp.Time) {
+			return len(events.Items[i].Message) > len(events.Items[j].Message)
+		}
+
+		// if both contains primeName substring or neither, just sort on timestamp
+		return events.Items[i].LastTimestamp.Time.After(events.Items[j].LastTimestamp.Time)
+	})
+}
+
 // UpdatePVCBoundContionFromEvents updates the bound condition annotations on the PVC based on recent events
 // This function can be used by both controller and populator packages to update PVC bound condition information
 func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client.Client, log logr.Logger) error {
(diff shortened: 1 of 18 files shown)


Dsanatar:
`/retest`

Dsanatar:
`/test pull-cdi-generate-verify`

pkg/controller/datavolume/conditions.go
```diff
@@ -125,6 +125,11 @@ func UpdateReadyCondition(conditions []cdiv1.DataVolumeCondition, status corev1.
 func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.PersistentVolumeClaim, message, reason string) []cdiv1.DataVolumeCondition {
 	if pvc != nil {
 		pvcCondition := getPVCCondition(pvc.GetAnnotations())
+		pvcPrimeMessage := ""
+		val, exists := pvc.GetAnnotations()[cc.AnnAPIGroup+"/storage.populator.pvcPrime"]
```
awels: Can we make this a const, we have a whole list of annotation consts already, you can add it there.

Also since we only care about this inside the case below, maybe move it there.
This might also be a nice candidate for a helper function, something like this
```go
func getPrimeMessage(pvc *k8sv1.PersistentVolumeClaim) string {
  val, exists := pvc.GetAnnotations()[cc.AnnAPIGroup+"/storage.populator.pvcPrime"]
  if exists {
    return value
  }
  return ""
}
```
Then below you can something like this:
```go
  pvcPrimeMessage := getPrimeMessage(pvc)
  if pvcPrimeMessage != "" {
    conditions = updateCondition(conditions, cdiv1.DataVolumeBound, corev1.ConditionFalse, fmt.Sprintf("PVC %s Pending%s", pvc.Name, pvcPrimeMessage), pvcPending)
  }
```

pkg/controller/populators/import-populator.go
```diff
@@ -97,9 +100,29 @@ func NewImportPopulator(
...
+		return nil, err
+	}
+
 	return importPopulator, nil
 }
 
+func addEventWatcher(mgr manager.Manager, c controller.Controller) error {
+	if err := c.Watch(source.Kind(mgr.GetCache(), &corev1.Event{}, handler.TypedEnqueueRequestsFromMapFunc[*corev1.Event](
+		func(_ context.Context, e *corev1.Event) []reconcile.Request {
+			if e.InvolvedObject.Kind == "PersistentVolumeClaim" && strings.Contains(e.InvolvedObject.Name, "prime") {
```
awels: Maybe also check that the `APIGroup = ""` so that if someone else decides they want their own API group with persistent volume claims kind they can do so without messing with us.

pkg/controller/populators/import-populator.go
```diff
@@ -97,9 +100,29 @@ func NewImportPopulator(
...
+				return []reconcile.Request{{
+					NamespacedName: types.NamespacedName{Name: e.Namespace},
```
awels: This looks weird, I was expecting you to create a key consisting of the Name and Namespace of the involved object. This looks like your name is the namespace of the event. Something like this should work
```go
NamespacedName: types.NamespacedName{Name: e.InvolvedObject.Name, Namespace: e.InvolvedObject.Namespace}
``

File: pkg/controller/populators/import-populator.go [hunk 4]
```diff
@@ -183,6 +214,68 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
...
+
+	if err != nil {
+		log.Error(err, "Could not retrieve srcObj list of Events")
+	}
+
+	var emitEvent bool
+	for idx, newEvent := range newEvents.Items {
+		emitEvent = true
+		currTime := newEvent.FirstTimestamp.Unix()
+		for _, currEvent := range currEvents.Items {
```
awels: Maybe instead of looping over the currEvents, can we put them in a map, and key it on the timestamp. Then if the currEvents already exists, we don't have to emit it. Not sure. Doing 2 loops doesn't seem like the optimal way to go about it. Even if the number of events is pretty low.

pkg/controller/populators/populator-base.go
```diff
@@ -101,6 +104,22 @@ func getIndexArgs() []indexArgs {
 				return []string{string(obj.GetUID())}
 			},
 		},
+		{
```
awels, line 108: I am not sure I understand why this index is here. Can you explain it to me?

pkg/controller/populators/upload-populator.go
```diff
@@ -121,13 +121,13 @@ func (r *UploadPopulatorReconciler) updatePVCForPopulation(pvc *corev1.Persisten
 	pvc.Annotations[cc.AnnPreallocationRequested] = strconv.FormatBool(cc.GetPreallocation(context.TODO(), r.client, uploadSource.Spec.Preallocation))
 }
 
-func (r *UploadPopulatorReconciler) updatePVCPrimeNameAnnotation(pvc *corev1.PersistentVolumeClaim, pvcPrimeName string) (bool, error) {
+func updatePVCPrimeNameAnnotation(pvc *corev1.PersistentVolumeClaim, pvcPrimeName string, client client.Client) (bool, error) {
```
awels: Maybe move this into the base controller, so all other controller have access to it. That way you don't have to pass in the client.

awels (commented):
Did a quick review, have some questions and suggestions.

pkg/controller/populators/populator-base.go
dsanatar: I believe I need these two new indexes so that I can filter the events by their InvolvedObect's Name/UID when calling `List()` , see  [here](https://github.com/kubevirt/containerized-data-importer/pull/3764/files/2ce865ff936834308e2f2f62cc3697a4565bcd3a#diff-bc696f76dbd372ce4c99e349fccc22a7f5bb0a9311f11557c97c74affa659f41R223-R227)

pkg/controller/populators/import-populator.go
dsanatar: Can you clarify this? I see that there is an APIGroup annotation that get set for PVCs but I don't think I can access that data from an Event object

pkg/controller/common/util.go
```diff
@@ -347,6 +347,9 @@ const (
 
 	// AnnCreatedForDataVolume stores the UID of the datavolume that the PVC was created for
 	AnnCreatedForDataVolume = AnnAPIGroup + "/createdForDataVolume"
+
+	// AnnPVCPrimeName annotation is the name of the PVC' that is added to the target PVC
```
awels: Not sure what you mean by this comment, do you mean that it is the name of the PVC' that is used to populate the PV that is then rebound to the target PVC?

pkg/controller/datavolume/conditions.go
```diff
@@ -177,3 +178,12 @@ func getPVCCondition(anno map[string]string) *cdiv1.DataVolumeCondition {
 	}
 	return nil
 }
+
+func getPrimeMessage(pvc *corev1.PersistentVolumeClaim) string {
+	val, exists := pvc.GetAnnotations()[cc.AnnPVCPrimeName]
+	if exists {
+		pvcPrimeMessage := fmt.Sprintf(" [prime PVC %s]", val)
```
awels: nit: why put the space here? This is forcing anyone that is calling this function to take the space into account. Right now not a big deal though. Now of course if you remove the space here, you have to add it to the callers.

pkg/controller/import-controller.go
```diff
@@ -803,6 +806,58 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+	// set bound condition by getting the latest event
+	events := &corev1.EventList{}
+
+	err := r.client.List(context.TODO(), events,
+		client.InNamespace(pvc.GetNamespace()),
+		client.MatchingFields{"involvedObject.name": pvc.GetName(),
+			"involvedObject.uid": string(pvc.GetUID())},
+	)
+
+	if err != nil || len(events.Items) == 0 {
```
awels: We should not eat the error here. I think it makes sense for this function to return an error. So something like:
```go
if err != nil {
  return err
}
```
You can sort and iterator over empty lists, so no need to check if events.Items is empty.

pkg/controller/import-controller.go
```diff
@@ -803,6 +806,58 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+		return
+	}
+
+	// Sort event lists by most recent
+	sort.Slice(events.Items, func(i, j int) bool {
+		return events.Items[i].FirstTimestamp.Time.After(events.Items[j].FirstTimestamp.Time)
```
awels: Would it make sense to sort on both event.Message containing and the time? Then if you sort it so the newest event with pvcPrime in the message is the top element, you don't have to loop anymore. You can just get the first item.

pkg/controller/import-controller.go
```diff
@@ -803,6 +806,58 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+			return
+		}
+	} else {
+		// if not using populators just get the latest event
+		boundMessage = events.Items[0].Message
+	}
+
+	anno := pvc.GetAnnotations()
+
+	if pvc.Status.Phase == corev1.ClaimBound {
```
awels: Is the other code not setting the bound condition annotation right? I know you are adding the message here, but the other two fields are already IIRC.

pkg/controller/populators/import-populator.go
awels: So if I remember correctly, the InvolvedObject will have at least the following fields: `Kind`, `APIGroup`, `Name`. Those will uniquely identify a resource in the cluster. So I am asking for `e.InvolvedObject.APIGroup` to be blank as well.

awels (commented):
Have not finished reviewing, but some feedback.

pkg/controller/import-controller.go
dsanatar: currently the only places I can see that update these boundCondition annotations are for scratch and clone pvcs.

pkg/controller/import-controller.go
dsanatar: I updated this to log the error instead. There are existing unit tests that do not fully setup the controllers and their associated field indexers which is required for this `List()` function call to work. Because the tests don't do that, `List()` throws an error which causes the tests to see unexpected errors in the Reconcile loop.

pkg/controller/datavolume/conditions.go
```diff
@@ -134,11 +134,16 @@ func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.Pe
...
 			}
 		case corev1.ClaimPending:
+			pvcPrimeName := getPrimeMessage(pvc)
+			if pvcPrimeName != "" {
+				// add space for readability
+				pvcPrimeName = " " + pvcPrimeName
+			}
 			if pvcCondition == nil || pvcCondition.Status == corev1.ConditionTrue {
-				conditions = updateCondition(conditions, cdiv1.DataVolumeBound, corev1.ConditionFalse, fmt.Sprintf("PVC %s Pending", pvc.Name), pvcPending)
+				conditions = updateCondition(conditions, cdiv1.DataVolumeBound, corev1.ConditionFalse, fmt.Sprintf("PVC %s Pending%s", pvc.Name, pvcPrimeName), pvcPending)
```
alromeros: Wouldn't it be cleaner to include the space inside the string "PVC %s Pending %s" instead of adding it separately? It feels less hacky imo.

pkg/controller/datavolume/conditions.go
```diff
@@ -134,11 +134,16 @@ func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.Pe
 				conditions = UpdateReadyCondition(conditions, corev1.ConditionFalse, "", "")
 			}
 		case corev1.ClaimPending:
+			pvcPrimeName := getPrimeMessage(pvc)
```
alromeros: Is this function returning the prime name or "prime message"? what's exactly prime message? I think we can improve the naming a bit here.

pkg/controller/common/util.go
```diff
@@ -347,6 +347,9 @@ const (
 
 	// AnnCreatedForDataVolume stores the UID of the datavolume that the PVC was created for
 	AnnCreatedForDataVolume = AnnAPIGroup + "/createdForDataVolume"
+
+	// AnnPVCPrimeName annotation is the name of the PVC' that is used to populate the PV which is then rebound to the target PVC
+	AnnPVCPrimeName = AnnAPIGroup + "/storage.populator.pvcPrime"
```
alromeros, line 356: We should've added this long ago, thanks!

pkg/controller/import-controller.go
```diff
@@ -803,6 +807,76 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+	if len(events.Items) == 0 {
+		return nil
+	}
+
+	pvcPrime, exists := pvc.GetAnnotations()[cc.AnnPVCPrimeName]
+
+	// Sort event lists by containing primeName substring and most recent timestamp
+	sort.Slice(events.Items, func(i, j int) bool {
+		if exists {
+			firstConatinsPrime := strings.Contains(events.Items[i].Message, pvcPrime)
```
alromeros: typo: Conatins

pkg/controller/import-controller.go
```diff
@@ -803,6 +807,76 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+			secondConatinsPrime := strings.Contains(events.Items[j].Message, pvcPrime)
```
alromeros: typo: Conatins

alromeros (commented):
Looking good! Populators have historically been a bit racy so we need to be extra sure none of these changes are causing the test errors, but I'm not seeing anything wrong with the changes. Will probably need to do a second review later.

pkg/controller/datavolume/conditions.go
dsanatar: yea maybe I can find a better way to do this, we need to preserve this PVC Pending string because there are a few unit tests that check for this exact string and adding a space here causes them to fail

pkg/controller/import-controller.go
```diff
@@ -803,6 +807,76 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+
+		if boundMessage == "" {
+			return nil
+		}
+	} else {
+		// if not using populators just get the latest event
+		boundMessage = events.Items[0].Message
+	}
+
+	anno := pvc.GetAnnotations()
```
alromeros: Seems there's a panic in this function (check [this log](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3764/pull-containerized-data-importer-e2e-istio/1938001112250454016/artifacts/1/pods/1_cdi-zdxo7_cdi-deployment-7c9cfc486d-n5hhm-cdi-deployment.log)). We should probably nil check these annotations in case they are nil.

pkg/controller/import-controller.go
```diff
@@ -803,6 +807,76 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+		// Log the error but don't fail the reconciliation
+		r.log.Error(err, "Unable to list events for PVC bound condition update", "pvc", pvc.Name)
+		return nil
+	}
+
+	if len(events.Items) == 0 {
+		return nil
+	}
+
+	pvcPrime, exists := pvc.GetAnnotations()[cc.AnnPVCPrimeName]
```
alromeros: Same, let's nil check these annotations before dereferencing.

pkg/controller/datavolume/conditions.go
```diff
@@ -134,11 +134,12 @@ func updateBoundCondition(conditions []cdiv1.DataVolumeCondition, pvc *corev1.Pe
 				conditions = UpdateReadyCondition(conditions, corev1.ConditionFalse, "", "")
 			}
 		case corev1.ClaimPending:
+			pvcPrimeName := getPrimeName(pvc)
 			if pvcCondition == nil || pvcCondition.Status == corev1.ConditionTrue {
 				conditions = updateCondition(conditions, cdiv1.DataVolumeBound, corev1.ConditionFalse, fmt.Sprintf("PVC %s Pending", pvc.Name), pvcPending)
```
alromeros, line 140: we don't want the prime name here anymore? In that case may be worth included inside the `else`.

pkg/controller/import-controller.go
alromeros: Though maybe here it's not necessary.

pkg/controller/import-controller.go
```diff
@@ -803,6 +807,76 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
...
+	anno := pvc.GetAnnotations()
+
+	if pvc.Status.Phase == corev1.ClaimBound {
+		anno[cc.AnnBoundCondition] = "true"
+		anno[cc.AnnBoundConditionReason] = "Bound"
+	} else {
+		anno[cc.AnnBoundCondition] = "false"
+		anno[cc.AnnBoundConditionReason] = "Pending"
+	}
+	anno[cc.AnnBoundConditionMessage] = boundMessage
```
alromeros: I'm seeing [here](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3764/pull-containerized-data-importer-e2e-ceph-wffc/1937952423154814976/artifacts/1/serial_2_pvcs.log) a PVC with a bound message that doesn't match what's shown in  `condition.bound": "true"`. Are we sure we want to assign the message of the latest event even if it doesnt match its `AnnBoundCondition` and  `AnnBoundConditionReason`? This logic seems a bit buggy imo.

pkg/controller/import-controller.go
alromeros: I'm seeing odd logs in general so it's possible there are more issues other than the nil address, but let's first address this and see how tests behave.

dsanatar: ah nice thanks good catch

pkg/controller/import-controller.go
dsanatar: hmm yea that definitely doesn't look good. I was never too confident with this logic, I just needed the `AnnBoundCondition` and `AnnBoundConditionReason` to be set before the DV updates it's bound condition because it first [checks](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume/conditions.go#L166-L178) to see these fields exist before the update occurs. maybe we can further discuss how these should be set here.

pkg/controller/common/util.go
```diff
@@ -2114,3 +2117,109 @@ func ResolveDataSourceChain(ctx context.Context, client client.Client, dataSourc
 
 	return resolved, nil
 }
+
+// UpdatePVCBoundContionFromEvents updates the bound condition annotations on the PVC based on recent events
+// This function can be used by both controller and populator packages to update PVC bound condition information
+func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client.Client, log logr.Logger) error {
+	currentPvcCopy := pvc.DeepCopyObject()
```
awels: looks like you only use this in the part where you remove the bound condition annotations. So you can move it inside the if. And once you do that, you can probably extract the code in that if statement into its own function called `clearBoundCondition' or something like that.

pkg/controller/common/util.go
```diff
@@ -2114,3 +2117,109 @@ func ResolveDataSourceChain(ctx context.Context, client client.Client, dataSourc
...
+
+	if len(events.Items) == 0 {
+		return nil
+	}
+
+	pvcPrime, exists := anno[AnnPVCPrimeName]
+
+	// Sort event lists by containing primeName substring and most recent timestamp
+	sort.Slice(events.Items, func(i, j int) bool {
+		if exists {
```
awels: I think it makes sense to extra this inline func into its own function, and to write some unit tests for it. That way we can be sure the behavior is what we want. Especially the weird edge cases where the timestamp is the same.

pkg/controller/populators/clone-populator.go
```diff
@@ -264,12 +264,16 @@ func (r *ClonePopulatorReconciler) Reconcile(ctx context.Context, req reconcile.
...
+		return reconcile.Result{}, err
+	}
+
 	hasFinalizer := cc.HasFinalizer(pvc, cloneFinalizer)
 	isBound := cc.IsBound(pvc)
 	isDeleted := !pvc.DeletionTimestamp.IsZero()
 	isSucceeded := isClonePhaseSucceeded(pvc)
 
-	log.V(3).Info("pvc state", "hasFinalizer", hasFinalizer,
+	log.V(1).Info("pvc state", "hasFinalizer", hasFinalizer,
```
awels: debugging leftover?

pkg/controller/populators/clone-populator.go
```diff
@@ -389,7 +393,7 @@ func (r *ClonePopulatorReconciler) planAndExecute(ctx context.Context, log logr.
 		}
 	}
 
-	log.V(3).Info("executed all phases, setting phase to Succeeded")
+	log.V(1).Info("executed all phases, setting phase to Succeeded")
```
awels: debugging leftover?

pkg/controller/populators/import-populator.go
```diff
@@ -183,6 +219,56 @@ func (r *ImportPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.Per
 	return reconcile.Result{}, nil
 }
 
+// CopyEvents gets primePVC events and re-emits them on the target PVC with the prime name prefix
+func CopyEvents(primePVC, targetPVC client.Object, c client.Client, log logr.Logger, recorder record.EventRecorder) {
```
awels: Does it make sense to make this part of the populator-base file? And maybe even make it a member function there, so you don't have to pass in the client/logger/recorder?

pkg/controller/common/util.go
dsanatar: If the pvc is not bound, I use this DeepCopy to call `Patch` at the end of this function. I can move this code block into it's own function but the call to `Patch` requires a `client.Client` object that i'm not sure we want to pass into another func. I'll defer to you on this one.

awels: Oh I didn't see that you use `DeepCopyObject` any reason you aren't using `DeepCopy` instead, that way you don't have to cast later since it will be the right type.

awels (approved):
/approve

PR #3779, opened 2025-06-05
Bugfix: Fix math in overhead size calculations 

alromeros:

Both `GetRequiredSpace` and `GetUsableSpace` functions had incorrect math for handling filesystem overhead:

1. **GetRequiredSpace**: Used `size / (1 - overhead)` which applies the percentage wrongly. For example 100GB with 70% overhead resulted in 333GB allocation instead of expected 170GB

2. **GetUsableSpace**: Used `availableSpace * (1 - overhead)` which cannot correctly reverse the overhead calculation. For example If 110GB was allocated with 10% overhead, it returned 99GB usable instead of correct 100GB.

These changes caused some variations in size expectations in tests, specifically in ones involving scratch space PVC sizing. I've tried minimizing the impact by adding an additional `RoundUp` before setting the size to account for both the `math.Floor` and `RoundDown` performed in the aforementioned functions. 

Fixes # https://issues.redhat.com/browse/CNV-61057

Need to be extra careful not breaking any size validation with `RoundUp`, `RoundDown`, `math.Ceil` and `math.Floor`. Let's see how tests behave.

**Release note**:

```release-note
Bugfix: Fix math in overhead size calculations
```


pkg/util/util.go
```diff
@@ -221,11 +221,11 @@ func Md5sum(filePath string) (string, error) {
 
 // GetUsableSpace calculates usable space to use taking file system overhead into account
 func GetUsableSpace(filesystemOverhead float64, availableSpace int64) int64 {
-	// +1 always rounds up.
-	spaceWithOverhead := int64(math.Ceil((1 - filesystemOverhead) * float64(availableSpace)))
+	// Reverse the overhead calculation
+	spaceWithoutOverhead := int64(math.Ceil(float64(availableSpace) / (1 + filesystemOverhead)))
```
akalenyu, line 225: So I suppose this was less about undoing the inflating but rather the original filesystem overhead mechanism which just took away a certain percentage of the user request
(Back then there was no .storage API, only .pvc, so you get a PVC with same size as your request)

akalenyu (commented):
Really great find! definitely a PR to ponder about a bit before we merge,
But logically I agree with the statements in the description

I guess my first pass worry is whether the new `GetUsableSpace` calculation makes sense for PVC-API-created volumes, and if so, should we care?

pkg/util/util.go
alromeros: ah okay that makes sense, thanks for the clarification. Still, seems that we use it in several places to undo the filesystem inflation so I think this current implementation makes more sense.

alromeros:
As suspected there's an issue with scratch PVC inflation in https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3779/pull-containerized-data-importer-e2e-ceph/1930656117604487168. Need to take a better look and see how to handle the round up (we used to apply the round up after overhead inflation when creating the scratch PVC).

pkg/util/util.go
akalenyu: It does for most cases. But it's probably wrong for .pvc API-created-volumes
(volume not inflated)

akalenyu:
> As suspected there's an issue with scratch PVC inflation in https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3779/pull-containerized-data-importer-e2e-ceph/1930656117604487168. Need to take a better look and see how to handle the round up (we used to apply the round up after overhead inflation when creating the scratch PVC).

I see the mismatch in scratch/target size you're referring to, just note the importer pod seemed to fail over writing to target, not scratch

alromeros:
So the test is failing because the 5.5% overhead increase is no longer enough to cover the ceph FS overhead. It used to work before because the previous math effectively gave us an overhead closer to 5.8%, which was sufficient.

```sh
 # stat -f /data/
  File: "/data/"
    ID: be1ec1f940252570 Namelen: 255     Type: ext2/ext3
Block size: 4096      
Blocks: Total: 213681     Free: 213675     Available: 209579
Inodes: Total: 55440      Free: 55429

# Available Space is less than the requested size before applying overhead inflation
209579 (Bavail) * 4096 (Bsize) = 858435584 bytes
```
The test passes when using a 5.8% overhead, which gives us an inflation similar to what we had with the previous behavior.

A couple of options we could consider:

- Increase the default overhead to 7%, as we've discussed before. This would also help prevent similar ceph issues going forward, considering that the new behavior gives smaller overhead inflation overall and this will be easily replicable when creating DVs using the same size request as the image size.
- Only increase the overhead in this specific test for now, so it passes and we can have more time to decide whether to update the default overhead.

@akalenyu @mhenriks @awels

akalenyu:
> So the test is failing because the 5.5% overhead increase is no longer enough to cover the ceph FS overhead. It used to work before because the previous math effectively gave us an overhead closer to 5.8%, which was sufficient.
> 
> ```shell
>  # stat -f /data/
>   File: "/data/"
>     ID: be1ec1f940252570 Namelen: 255     Type: ext2/ext3
> Block size: 4096      
> Blocks: Total: 213681     Free: 213675     Available: 209579
> Inodes: Total: 55440      Free: 55429
> 
> # Available Space is less than the requested size before applying overhead inflation
> 209579 (Bavail) * 4096 (Bsize) = 858435584 bytes
> ```
> 
> The test passes when using a 5.8% overhead, which gives us an inflation similar to what we had with the previous behavior.
> 
> A couple of options we could consider:
> 
> * Increase the default overhead to 7%, as we've discussed before. This would also help prevent similar ceph issues going forward, considering that the new behavior gives smaller overhead inflation overall and this will be easily replicable when creating DVs using the same size request as the image size.
> * Only increase the overhead in this specific test for now, so it passes and we can have more time to decide whether to update the default overhead.
> 
> @akalenyu @mhenriks @awels

I think this test has always passed due to the calculation being wrong on the inflating side.
We fail because physically available on fs pvc < virtual size (overhead not even considered)

That said I think we should still make sure the scenario works (`Upload an image exactly the same size as DV request`)
I just think we grabbed the wrong size input instead of looking up the virtual size from this specifically purposed image (fs-overhead.qcow2)
```bash
$ qemu-img info tests/images/fs-overhead.qcow2 --output json | jq '."virtual-size"'
858993664
```

Which is different from the size input in the test
(but this may not help make it pass, see below)

I think there's either:
- Still a bug in the math inflating (less likely)
- Ceph rbd filesystem PVC just doesn't satisfy the size request to a point that it can't even contain the virtual size

alromeros:
> I think this test has always passed due to the calculation being wrong on the inflating side.

Yeah, that’s what I meant, the previous calculation was incorrect and ended up giving us something closer to 5.8% inflation instead of the intended 5.5%.

> We fail because physically available on fs pvc < virtual size (overhead not even considered)

That’s also correct, but I believe the calculations are now accurate. We request 858993459 bytes, which gets rounded up to 858993664 and then inflated by 5.5%. The issue is that the actual available space (858435584) is even lower than that, likely because some space is already used in the PVC by ceph's overhead. So the 5.5% overhead is no longer enough to account for that.

>  I just think we grabbed the wrong size input instead of looking up the virtual size from this specifically purposed image (fs-overhead.qcow2)

The size used by the test is wrong indeed but afaiu this is done intentionally, or that's what I understood from the comment in the test. 858993459 is chosen but then rounded up in CDI to the closest MiB aproximation which is 858993664, which equals to the virtual image of the disk.

pkg/util/util.go
```diff
@@ -221,16 +221,23 @@ func Md5sum(filePath string) (string, error) {
...
 }
 
-func CalculateOverheadSpace(filesystemOverhead float64, availableSpace int64) int64 {
-	spaceWithOverhead := int64(math.Ceil(float64(availableSpace) / (1 - filesystemOverhead)))
-	return RoundUp(spaceWithOverhead, DefaultAlignBlockSize)
+// GetRequiredSpace calculates space required taking file system overhead into account
+func GetRequiredSpace(filesystemOverhead float64, requestedSpace int64) int64 {
+	// the `image` has to be aligned correctly, so the space requested has to be aligned to
+	// next value that is a multiple of a block size
+	alignedSize := RoundUp(requestedSpace, DefaultAlignBlockSize)
```
mhenriks, line 235: shouldn't the alignment happen after the calculation below? Don't think we can guarantee alignment after the calculation

akalenyu:
> > We fail because physically available on fs pvc < virtual size (overhead not even considered)
> 
> That’s also correct, but I believe the calculations are now accurate. We request 858993459 bytes, which gets rounded up to 858993664 and then inflated by 5.5%. The issue is that the actual available space (858435584) is even lower than that, likely because some space is already used in the PVC by ceph's overhead. So the 5.5% overhead is no longer enough to account for that.
> 
> >

I **think** that this scenario (inflated by 5.5% still lower than DV size request) is only possible under 1Gi so sort of a non issue. Maybe we can pick another size value and recreate that dummy image?

pkg/util/util.go
alromeros: I was torn between keeping this behavior (alignment before inflation) or following the one used in `CalculateOverheadSpace` (alignment after inflation). I ended up choosing this because we're aligning the size of the virtual image itself (which is what we care about) and then inflating it to account for the overhead, which I assume doesn't need to be aligned.

mhenriks: Say you have a 1G disk, so it is already aligned. If fs overhead is 0.03, the output of the calculation will not be aligned

alromeros: Isn’t that actually what we want?  I mean, to ensure that the image itself fits fully within an aligned space and then account for filesystem overhead on top of that. That’s also what the current implementation does, and the comment in the function kinda reflects that reasoning (aligning the image size itself). I'm ok with changing it but need to know if we have valid reasons for it.

awels: The PVCs size doesn't need to be aligned, the disk.img size needs to be aligned so qemu doesn't start complaining.

mhenriks: okay

mhenriks:
@akalenyu @alromeros am I correct in assuming that 5.5% will most likely only be an issue with small disks (under 1G), if so how about we simply bump the fsoverhead for functest env? Will have to do it for kubevirt/kubevirt as well.

I am keen to get this change in and backported. Hesitant to change the default overhead value

alromeros:
> @akalenyu @alromeros am I correct in assuming that 5.5% will most likely only be an issue with small disks (under 1G), if so how about we simply bump the fsoverhead for functest env? Will have to do it for kubevirt/kubevirt as well.
> 
> I am keen to get this change in and backported. Hesitant to change the default overhead value

Yeah you're right, I’m pretty sure this change will trigger test failures in both Kubevirt and CDI. I'm fine with handling it just in our functional tests for now. That said, I worry this may also affect users with provisioners that behave similar to ceph (which occupies more overhead space).

mhenriks:
> That said, I worry this may also affect users with provisioners that behave similar to ceph (which occupies more overhead space).

To me this is not really a provisioner thing, rather a filesystem thing that should behave similarly across different provisioners for the same fs type

akalenyu:
> I am keen to get this change in and backported. Hesitant to change the default overhead value

Yeah we should get this in but backport is playing with fire IMO since at this point this faulty behavior
could be someones desired (https://www.hyrumslaw.com/ and friends).

> if so how about we simply bump the fsoverhead for functest env?

I'm not sure this is a good idea for our functest environment to diverge from the user's,
But I do think we could refactor this test to work on a > 1Gi image, and if that doesn't work,
we may have a problem as the scenario is legit
Basically worried we may be sweeping a real issue under the rug

> To me this is not really a provisioner thing, rather a filesystem thing that should behave similarly across different provisioners for the same fs type

Yeah it's a thing for all block providers that also allow creating filesystems on top of their block devices.


@alromeros also note the discussion in https://github.com/kubevirt/containerized-data-importer/pull/3779#discussion_r2134499123

alromeros:
> But I do think we could refactor this test to work on a > 1Gi image, and if that doesn't work,
> we may have a problem as the scenario is legit
> Basically worried we may be sweeping a real issue under the rug

The issue with using a >1Gi disk image for the test is that the test is meant to specifically reproduce a bug tied to sub-1Gi images. Using a larger image wouldn't validate the same condition. Notice the:

https://github.com/kubevirt/containerized-data-importer/blob/b6b746878b67e927d17ba2c90a3d5f251b3de56f/tests/upload_test.go#L1369-L1373

We can use a different image there but the test wouldn't be replicating and testing the bug case that motivated its inclusion. I also worry we might end up triggering other errors in Kubevirt tests where using a larger image size might not be as straightforward.

alromeros:
I'll run the test locally with a larger image so we can at least discard if the issue happens with +1Gi images.

akalenyu:
> > But I do think we could refactor this test to work on a > 1Gi image, and if that doesn't work,
> > we may have a problem as the scenario is legit
> > Basically worried we may be sweeping a real issue under the rug
> 
> The issue with using a >1Gi disk image for the test is that the test is meant to specifically reproduce a bug tied to sub-1Gi images. Using a larger image wouldn't validate the same condition. Notice the:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/b6b746878b67e927d17ba2c90a3d5f251b3de56f/tests/upload_test.go#L1369-L1373
> 
> We can use a different image there but the test wouldn't be replicating and testing the bug case that motivated its inclusion. I also worry we might end up triggering other errors in Kubevirt tests where using a larger image size might not be as straightforward.

AFAIU the experimental sub 1Gi test was just a convenient reproducer, the original bugzilla has a rhel data volume

akalenyu (commented):
For me the only thing missing here is handling of .PVC APIs (hate to be that guy)
through an env var or something along those lines, that would keep the existing internal overhead behavior. 

BTW here's some graphs to visualize this
(green=good calc, blue=bad calc)
image
image
So, smaller PVCs but bigger disk images

akalenyu:
Talked to @mhenriks, we can skip the .pvc API change, just make sure to mention that in the release note
/approve

alromeros:
> Talked to @mhenriks, we can skip the .pvc API change, just make sure to mention that in the release note /approve

Just to clarify, the main issue I see with the old behavior is that it seems like an absolute arbitrary choice to use `availableSpace * (1 - overhead)` instead of the new `availableSpace / (1 + filesystemOverhead)` so I don't see a reason to worry about preserving it. I get that this changes the current behavior slightly when users might expect otherwise, but with typical overhead values it shouldn't be a significant change in size. In the worst case, we provide more usable space which is probably better anyway. So IMO this is an acceptable change, and adding complexity just to preserve the old behavior seemed unnecessary.

I'll update the release note, let me know if it looks good. Thanks!

mhenriks:
> Talked to @mhenriks, we can skip the .pvc API change, just make sure to mention that in the release note

Reason being, with `.pvc` API we guarantee a specific PVC size and we still do that. With `.storage` API, we guarantee a specific disk image size (and we still do that).

alromeros:
> > Talked to @mhenriks, we can skip the .pvc API change, just make sure to mention that in the release note
> 
> Reason being, with `.pvc` API we guarantee a specific PVC size and we still do that. With `.storage` API, we guarantee a specific disk image size (and we still do that).

Yeah great way to put it.

akalenyu:
> When using the PVC API, usable disk space is now calculated more accurately with "allocatedSize / (1 + overhead)". This may result in changes in usable size compared to the previous behavior.

With .pvc API you're now undoing inflating that doesn't exist i.e. `available_space / 105.5%` so the user's disk is effectively smaller. I would directly mention that in the release note
```
When using the PVC API, usable disk space is now lower according to the correct new calculation. Users still relying on this API should adjust for manual inflation to keep old VM disk sizes.
```

alromeros:
> > When using the PVC API, usable disk space is now calculated more accurately with "allocatedSize / (1 + overhead)". This may result in changes in usable size compared to the previous behavior.
> 
> With .pvc API you're now undoing inflating that doesn't exist i.e. `available_space / 105.5%` so the user's disk is effectively smaller. I would directly mention that in the release note
> 
> ```
> When using the PVC API, usable disk space is now lower according to the correct new calculation. Users still relying on this API should adjust for manual inflation to keep old VM disk sizes.
> ```

Maybe I'm missing something but the previous calculation (`availableSpace * (1 - overhead)`) actually provided smaller "usable spaces" than the current one (`availableSpace / (1 + overhead)`).

akalenyu:
> > > When using the PVC API, usable disk space is now calculated more accurately with "allocatedSize / (1 + overhead)". This may result in changes in usable size compared to the previous behavior.
> > 
> > 
> > With .pvc API you're now undoing inflating that doesn't exist i.e. `available_space / 105.5%` so the user's disk is effectively smaller. I would directly mention that in the release note
> > ```
> > When using the PVC API, usable disk space is now lower according to the correct new calculation. Users still relying on this API should adjust for manual inflation to keep old VM disk sizes.
> > ```
> 
> Maybe I'm missing something but the previous calculation (`availableSpace * (1 - overhead)`) actually provided smaller "usable spaces" than the current one (`availableSpace / (1 + overhead)`).

~~For inflated volumes, yes. But PVC API sizes just get propagated as they are to the PVC~~

~~For example, an 8Gi DV results in a 8Gi PVC and you're undoing a 5.5% off of that now.~~

EDIT: I totally confused everyone. We were doing that even before this. Totally right, if anything, this "correct" calculation makes the problem a little less bad.

akalenyu:
seems stuck

alromeros:
@mhenriks @akalenyu @awels thanks for the reviews, glad to have this finally merged. The bug was initially opened for 4.17. I know there's been some concerns about backporting this, are we ok with backporting this far? 

/cherrypick release-v1.62
/cherrypick release-v1.61
/cherrypick release-v1.60

akalenyu:
> @mhenriks @akalenyu @awels thanks for the reviews, glad to have this finally merged. The bug was initially opened for 4.17. I know there's been some concerns about backporting this, are we ok with backporting this far?
> 
> /cherrypick release-v1.62 /cherrypick release-v1.61 /cherrypick release-v1.60

I think any backport is risky. Somebody could be relying on this faulty behavior by now.
Usually in these cases you see a feature gate that comes in alongside the change that allows you to revert to the faulty behavior:
https://github.com/kubernetes/kubernetes/blob/v1.33.2/pkg/features/kube_features.go#L974

But I'm open to what others think

PR #3782, opened 2025-06-10
Fix empty StorageProfile test event check

arnongilboa:
Also drop redundant tests which are covered by following tests.

**Release note**:
```release-note
NONE
```


arnongilboa:
@jpeimer @alromeros instead of looking for an empty `StorageProfile` (we won't find one on most envs), let's add a temporary `StorageClass` with unknown provisioner and get an empty `StorageProfile`. Check the top commit. wdyt?

jpeimer:
> @jpeimer @alromeros instead of looking for an empty `StorageProfile` (we won't find one on most envs), let's add a temporary `StorageClass` with unknown provisioner and get an empty `StorageProfile`. Check the top commit. wdyt?

Yeah, sounds good to me 
/lgtm

alromeros:
Looks good!
/lgtm

PR #3787, opened 2025-06-15
add HCO network-policy labels to required components

Acedus:
HCO includes built-in network policies to allow essential traffic for operators which it deploys. These policies permit:
1. Traffic to Cluster API and DNS
2. Traffic from Prometheus to allow metric scraping

This commit adds the necessary labels which are used as the selectors of the network policies to CDI deployments which require them.

Fixes [CNV-63386](https://issues.redhat.com/browse/CNV-63386)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -312,6 +312,11 @@ const (
 
 	// The restricted SCC and particularly v2 is considered best practice for workloads that can manage without extended privileges
 	RestrictedSCCName = "restricted-v2"
+
+	// Label key that is used by HCO managed network policies that allows access of matched Pods to cluster services such as DNS and API
+	HCOAllowAccessClusterServices = "hco.kubevirt.io/allow-access-cluster-services"
+	// Label key that is used by HCO managed network policies that allows Prometheus access to matched Pods
+	HCOAllowAccessPrometheus = "hco.kubevirt.io/allow-prometheus-access"
 )
 
 // ProxyPaths are all supported paths
--- a/pkg/operator/resources/namespaced/apiserver.go
+++ b/pkg/operator/resources/namespaced/apiserver.go
@@ -31,6 +31,7 @@ import (
 
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	utils "kubevirt.io/containerized-data-importer/pkg/operator/resources/utils"
+	"kubevirt.io/containerized-data-importer/pkg/util"
 	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/api"
 )
 
@@ -100,6 +101,12 @@ func createAPIServerDeployment(image, verbosity, pullPolicy string, imagePullSec
 	if replicas > 1 {
 		deployment.Spec.Replicas = &replicas
 	}
+	labels := util.MergeLabels(deployment.Spec.Template.GetLabels(), map[string]string{
+		common.HCOAllowAccessClusterServices: "",
+	})
+	deployment.SetLabels(labels)
+	deployment.Spec.Template.SetLabels(labels)
+
 	container := utils.CreatePortsContainer(common.CDIApiServerResourceName, image, pullPolicy, createAPIServerPorts(common.CDIApiServerResourceName))
 	container.Args = []string{"-v=" + verbosity}
 	container.Env = []corev1.EnvVar{
--- a/pkg/operator/resources/namespaced/controller.go
+++ b/pkg/operator/resources/namespaced/controller.go
@@ -191,7 +191,11 @@ func createControllerDeployment(controllerImage, importerImage, clonerImage, ovi
 			Protocol:      "TCP",
 		},
 	}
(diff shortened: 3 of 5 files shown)


Acedus:
/cc @akalenyu 
fyi @nunnatsa 

This should get us started on the CDI side of things. I'm personally still a bit conflicted on whether operands should inherit the labels as well. While convenient, I lean towards the only enabling the operator itself. Any additional NPs required by operands, even if seemingly redundant in the same namespace (e.g., when CDI is deployed alongside KubeVirt), should be managed by the operator. Your input will be very valuable as it will affect this and future pull requests.

pkg/common/common.go
```diff
@@ -312,6 +312,11 @@ const (
 
 	// The restricted SCC and particularly v2 is considered best practice for workloads that can manage without extended privileges
 	RestrictedSCCName = "restricted-v2"
+
+	// Label key that is used by HCO managed network policies that allows access of matched Pods to cluster services such as DNS and API
+	HCOAllowAccessClusterServices = "hco.kubevirt.io/allow-access-cluster-services"
+	// Label key that is used by HCO managed network policies that allows Prometheus access to matched Pods
+	HCOAllowAccessPrometheus = "hco.kubevirt.io/allow-prometheus-access"
 )
```
akalenyu, line 320: I wonder if we could generalize these outside of HCO by having the uber operator managing CDI (HCO) put something on our CR?

Or, if not, maybe this could belong in https://github.com/kubevirt/controller-lifecycle-operator-sdk
under a knob? Really not sure about this one just a raising this so others could benefit
(https://github.com/kubevirt/application-aware-quota)

nunnatsa: @akalenyu - I think the comment is misleading, and HCO is not involved here.

So the plan is to add three NetworkPolicies to the bundle image:
1. Allow egress to the API Server, from a pod with the `hco.kubevirt.io/allow-access-cluster-services` label.
2. Allow egress to the DNS, from a pod with the `hco.kubevirt.io/allow-access-cluster-services` label.
3. Allow ingress to the `metrics` port, in a pod with the `hco.kubevirt.io/allow-prometheus-access` label.

Once the bundle image includes these NetworkPolicies, OLM will deploy them, prior to the deployment of the operator pods. HCO can't do that because if a cluster with deny-all network policy, it won't work either. It must be done before the deployment of the operator pods.

Acedus: HCO is the one adding the NPs to the bundle to be deployed by the OLM.
I could remove the `HCO` naming to avoid confusion. It was suggested in a discussion on the matter which confused me too as I expected it to be something along the lines of `kubeivrt.io/allow-access-cluster-services` without the `hco` prefix.

nunnatsa: HCO does not do any of this. This all done in the bundle image build time.

nunnatsa: WDYT about something like this?

```go
// Label key that is used to bind a pod to NetworkPolicies that allows egress to the DNS and the API-Server
HCOAllowAccessClusterServices = "hco.kubevirt.io/allow-access-cluster-services"
// Label key that is used bind a pod to the NetworkPolicy that allows ingress from Prometheus
HCOAllowAccessPrometheus = "hco.kubevirt.io/allow-prometheus-access"
```

akalenyu: I'm just thinking about it from a non-HCO angle. Some other operator managing CDI and friends, that wants to go ahead with this feature.

Why don't we get an environment variable on the operator that tells CDI which annotation key to use for each  type of access?

Acedus: If I understand correctly, you're suggesting setting it as part of the manifest-generator and propagating it downwards after cdi-operator is deployed. I don't have an issue with that, seems more elegant actually. Is that what you meant @akalenyu?

akalenyu: I meant that we'd respect a new env var on the operator, through which HCO would tell us which key/value to use

ormergi: The mentioned labels were meant for the project operator pods only, when managed by HCO, allowing them to operate as soon they are installed in an env that features network-restrictions (deny-all network-policy).

It seems they are used for the pods CDI operator installed, although the appropriate way is to have the operator install the required NPs for its components as any other required object. Although this can work.

Originally the agreed approach was to extend the projects csv-generator to dump the relevant network policies, providing the same results w/o touching production code. And allow provide better solution with appropriate APIs in followup releases.
The same can be done here with the mentioned labels, consider changing csv-gen to label the relevant components with the mentioned HCO labels and get done with it :slightly_smiling_face:

Acedus: Yeah this PR is already quite dated, a lot has changed regarding the approach we took with NPs since and I'll probably close it in favor of an org-wide approach that will be agreed upon in a VEP. 

Thanks for taking the time to suggest this, though.

PR #3793, opened 2025-06-17
Fix error message in datavolume-import-controller

jing2uo:

This PR corrects an error message in the datavolume-import-controller to improve clarity and aid troubleshooting. The original error message incorrectly referenced Spec when reporting missing accessMode and volumeMode in the StorageProfile. The actual validation logic in the getDefaultVolumeAndAccessMode function checks the Status.ClaimPropertySets field of the StorageProfile. This discrepancy caused confusion during debugging, as the error message misled users to inspect the Spec instead of the Status.

The fix updates the error message to accurately reflect that the Status field is missing accessMode and volumeMode, aligning with the code's behavior and improving the debugging experience for users.

Fixes #

During troubleshooting, the following error logs were observed:
```text
Normal   ErrClaimNotValid  6s                datavolume-import-controller  no accessMode specified in StorageProfile cephfs
Warning  ErrClaimNotValid  4s (x12 over 6s)  datavolume-import-controller  Spec is missing accessMode and volumeMode, cannot get access mode from StorageProfile cephfs

```

**Release note**:

```release-note
NONE
```


pkg/controller/datavolume/util.go
```diff
@@ -173,7 +173,7 @@ func renderPvcSpecVolumeModeAndAccessModesAndStorageClass(client client.Client,
 		if err != nil {
 			logInfo("Cannot set accessMode and volumeMode for new pvc", "Error", err)
 			recordEventf(v1.EventTypeWarning, cc.ErrClaimNotValid,
-				fmt.Sprintf("Spec is missing accessMode and volumeMode, cannot get access mode from StorageProfile %s", getName(storageClass)))
+				fmt.Sprintf("Status is missing accessMode and volumeMode, cannot get access mode from StorageProfile %s", getName(storageClass)))
```
ShellyKa13: The error msg is accurate since the Spec that is refernced here is the pvcSpec which is missing both accessMode and volumeMode. And then the rest of the sentence explains that it tried to get the access mode from the relevant storageprofile and failed..
So the message is accurate but maybe not informative enough..
Maybe we can make the message clearer like:
`fmt.Sprintf("Data volume spec is missing accessMode and volumeMode, failed to get accessMode from StorageProfile %s, Error: %s", getName(storageClass), err))`
This way you will have the relevant error also in the Event. (though you should already have it in the log)

ShellyKa13: Now that I look at the PR description and there are 2 events which the first already adds the error that I mentioned.
So maybe you want to update the error that is returned from getDefaultVolumeAndAccessMode
instead of:
`return nil, nil, errors.Errorf("no accessMode specified in StorageProfile %s", storageProfile.Name)`
do something like:
`return nil, nil, errors.Errorf("no accessMode specified in ClaimPropertySets of StorageProfile %s", storageProfile.Name)`
and the message here just add the Datavolume before the Spec to clarify which spec is missing this properties,
WDYT?

jing2uo: I understand they correspond to two events. Thanks!

jing2uo: Please review the code again; I added a check for when storageProfile.Status.ClaimPropertySets is undefined or empty. Thank you!

jing2uo:
> /ok-to-test

Should use /test to trigger the test?

PR #3795, opened 2025-06-18
log error when O_DIRECT not supported on storage

akalenyu:

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu:
The problem ended up being that we don't support O_DIRECT writing to scratch. That's (I think) where it OOMs again even with TRYNONE CACHE_MODE.

Issue #3796, opened 2025-06-18
PVC cloning not working properly


bigbenaugust:
**What happened**:
When creating kubevirt VMs using CDI to clone the disk from a PVC, the VM is unbootable. We're not sure what it's doing, but it isn't cloning the image correctly. 

**What you expected to happen**:
A functional VM with a bootable disk. 

**How to reproduce it (as minimally and precisely as possible)**:
kubevirt VM manifest attached. [dev-image-02.yaml.txt](https://github.com/user-attachments/files/20796507/dev-image-02.yaml.txt)

**Additional context**:
We have three bare metal 22.04 clusters with kubevirt
- one was on k8s 1.28 and kubevirt 1.2
- one was on k8s 1.27 and kubevirt 0.59
- one was on k8s 1.26 and kubevirt 0.59

All were on CDI 1.56.
All were working before and had for ~2 years with the same images. 
All clusters were stepwise updated to k8s 1.32.5 with kubevirt 1.5.1 and CDI 1.62.0

The issues started after the upgrades. 
Existing VMs are unaffected. 
Image downloads from URL (qcow2 and ISO) work as expected. 
We were able to verify that the issue affects multiple images (we only have a few VMs in kubevirt)

From the PVC via a debug pod:
```
root@debug-pod-2:/mnt# md5sum disk.img 
246321f7b17b17beff16a200db3d2e40  disk.img
root@debug-pod-2:/mnt# qemu-img info disk.img 
image: disk.img
file format: raw
virtual size: 55 GiB (59055800320 bytes)
disk size: 55 GiB
root@debug-pod-2:/mnt# ls -la disk.img 
-rw-rw---- 1 107 107 59055800320 Apr 16  2024 disk.img
```

From the unbootable VM:
```
bash-5.1$ md5sum /var/run/kubevirt-private/vmi-disks/image-02-boot/disk.img
7ff62dcb4027a539260690ca1b6a5ed5  /var/run/kubevirt-private/vmi-disks/image-02-boot/disk.img
bash-5.1$ qemu-img info /var/run/kubevirt-private/vmi-disks/image-02-boot/disk.img
image: /var/run/kubevirt-private/vmi-disks/image-02-boot/disk.img
file format: raw
virtual size: 56.7 GiB (60880322560 bytes)
disk size: 56.7 GiB
Child node '/file':
    filename: /var/run/kubevirt-private/vmi-disks/image-02-boot/disk.img
    protocol type: file
[...]

PR #3797, opened 2025-06-23
Add proper cephfs provisioner string in capabilities

akalenyu:
The common occurence upstream is rook-ceph.rbd.csi.ceph.com,
I think cephfs.csi.ceph.com is a legacy pattern but keeping
both for compatibility

Fixes #

**Release note**:

```release-note
BugFix: Wrong clone strategy for upstream cephfs
```


akalenyu:
Does not fix the root cause (cephfs copy clone somehow results in not being able to boot) but helps https://github.com/kubevirt/containerized-data-importer/issues/3796

PR #3798, opened 2025-06-24
Support DIC DV node pullMethod for update polling

arnongilboa:
It allows the `DataImportCron` controller to fetch a registry `containerImage` `imageID` digest by a do-nothing image-puller container, without depending on `OpenShift` `ImageStream` or `cronJobs`. k8s node caching makes it also quite cheap. If the image was not updated, it is already cached. If the image is a new one, k8s will pull and cache it, so in the following import it will be retrieved from the cache, assuming both pods are scheduled on the same node.

**Release note**:
```release-note
Support DataImportCron DataVolume node pullMethod for update polling
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -262,15 +262,23 @@ func (r *DataImportCronReconciler) pollSourceDigest(ctx context.Context, dataImp
 	if nextTime.After(time.Now()) {
 		return r.setNextCronTime(dataImportCron)
 	}
-	if isImageStreamSource(dataImportCron) {
+	switch {
+	case isImageStreamSource(dataImportCron):
 		if err := r.updateImageStreamDesiredDigest(ctx, dataImportCron); err != nil {
 			return reconcile.Result{}, err
 		}
-	} else if isPvcSource(dataImportCron) {
+	case isPvcSource(dataImportCron):
 		if err := r.updatePvcDesiredDigest(ctx, dataImportCron); err != nil {
 			return reconcile.Result{}, err
 		}
+	case isNodePull(dataImportCron):
+		if done, err := r.updateContainerImageDesiredDigest(ctx, dataImportCron); !done {
+			return reconcile.Result{RequeueAfter: 3 * time.Second}, err
+		} else if err != nil {
+			return reconcile.Result{}, err
+		}
 	}
+
 	return r.setNextCronTime(dataImportCron)
 }
 
@@ -282,7 +290,7 @@ func (r *DataImportCronReconciler) setNextCronTime(dataImportCron *cdiv1.DataImp
 	}
 	nextTime := expr.Next(now)
 	requeueAfter := nextTime.Sub(now)
-	res := reconcile.Result{Requeue: true, RequeueAfter: requeueAfter}
+	res := reconcile.Result{RequeueAfter: requeueAfter}
 	cc.AddAnnotation(dataImportCron, AnnNextCronTime, nextTime.Format(time.RFC3339))
 	return res, err
 }
@@ -295,6 +303,12 @@ func isImageStreamSource(dataImportCron *cdiv1.DataImportCron) bool {
 func isURLSource(dataImportCron *cdiv1.DataImportCron) bool {
 	regSource, err := getCronRegistrySource(dataImportCron)
 	return err == nil && regSource.URL != nil
+}
+
+func isNodePull(cron *cdiv1.DataImportCron) bool {
+	regSource, err := getCronRegistrySource(cron)
+	return err == nil && regSource != nil && regSource.PullMethod != nil &&
+		*regSource.PullMethod == cdiv1.RegistryPullNode
 }
 
 func getCronRegistrySource(cron *cdiv1.DataImportCron) (*cdiv1.DataVolumeSourceRegistry, error) {
@@ -322,7 +336,7 @@ func isPvcSource(cron *cdiv1.DataImportCron) bool {
 }
 
 func isControllerPolledSource(cron *cdiv1.DataImportCron) bool {
-	return isImageStreamSource(cron) || isPvcSource(cron)
+	return isImageStreamSource(cron) || isPvcSource(cron) || isNodePull(cron)
 }
 
 func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *cdiv1.DataImportCron) (reconcile.Result, error) {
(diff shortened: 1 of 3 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -617,6 +631,126 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
...
+	}
+
+	workloadNodePlacement, err := cc.GetWorkloadNodePlacement(ctx, r.client)
+	if err != nil {
+		return false, err
+	}
+
+	containerImage := strings.TrimPrefix(*cron.Spec.Template.Spec.Source.Registry.URL, "docker://")
+
+	pod = &corev1.Pod{
```
Acedus, line 663: Why do we need the `shared-server` here? We only need the digest, why not just create a Pod with a single container of the target image?

pkg/controller/dataimportcron-controller.go
```diff
@@ -617,6 +631,126 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
...
+					VolumeSource: corev1.VolumeSource{
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
+					},
+				},
+			},
+			InitContainers: []corev1.Container{
+				{
+					Name:                     "init",
+					Image:                    r.image,
+					ImagePullPolicy:          corev1.PullPolicy(r.pullPolicy),
```
Acedus, line 696: We should probably use `Always` as we're polling for a new digest and kubelet identifies images by their name and tag.

pkg/controller/dataimportcron-controller.go
```diff
@@ -617,6 +631,126 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
...
+		digest, err := r.fetchContainerImageDigest(pod)
+		if err != nil || digest == "" {
+			return false, err
+		}
+		cc.AddAnnotation(cron, AnnLastCronTime, time.Now().Format(time.RFC3339))
+		if cron.Annotations[AnnSourceDesiredDigest] != digest {
+			log.Info("Updating DataImportCron", "digest", digest)
+			cc.AddAnnotation(cron, AnnSourceDesiredDigest, digest)
+		}
+		return true, r.client.Delete(ctx, pod)
```
Acedus, line 651: If deletion fails here we return that `done` is true so it re-queues after 3 seconds and disregards the `else if`, is that intended?

pkg/controller/dataimportcron-controller.go
arnongilboa: We don't "need" `shared-server` here, but that's the cheapest way I know to provide the container with a  runnable do-nothing `command`, otherwise pod will fail execution without providing us the imageID.

pkg/controller/dataimportcron-controller.go
arnongilboa: when `done` is `true` it goes directly to the `else if err != nil`

pkg/controller/dataimportcron-controller.go
arnongilboa: `IfNotPresent` does not care about digest change but only about tags? so we cannot use a cached image if there is one with the same digest?

arnongilboa:
> Very quick first pass, left some initial questions. Also, can we add a functional test to exercise this new logic?

We cover this with some of the existing tests, but I'll consider adding f/utests.

pkg/controller/dataimportcron-controller.go
Acedus: How about using `sleep`?

pkg/controller/dataimportcron-controller.go
Acedus: AFAIK it doesn't care about digest, need to check source code to fully confirm, though.

pkg/controller/dataimportcron-controller.go
Acedus: Right, disregard this.

pkg/controller/dataimportcron-controller.go
Acedus: Technically any entrypoint that can simply keep the Pod alive should be fine. In an image was built with `FROM: scratch` we should be able to still use it even if the Pod reaches a `Failed` state as we're only looking for the container image digest anyway.

I understand if you'd want to avoid that, though.

Acedus:
So reviewing this I think the solution is fine, I don't see any particular issues with it.
With that said, I wonder if we can leverage the external poller to achieve a more elegant solution where we add an [ephemeral container](url) that is the target image to the `cdi-source-update-poller` Pod, that way we don't care about its execution and we can simply let the Pod retrieve its ephemeral container image ID (given that the pull was successful) if the `pullMethod` was set to node (can probably be set via an environment variable). WDYT?

arnongilboa:
> So reviewing this I think the solution is fine, I don't see any particular issues with it. With that said, I wonder if we can leverage the external poller to achieve a more elegant solution where we add an [ephemeral container](url) that is the target image to the `cdi-source-update-poller` Pod, that way we don't care about its execution and we can simply let the Pod retrieve its ephemeral container image ID (given that the pull was successful) if the `pullMethod` was set to node (can probably be set via an environment variable). WDYT?

Not sure I understand how you create that `ephemeral container` and how the poller access the image. Can you please elaborate?

Acedus:
> Not sure I understand how you create that ephemeral container and how the poller access the image. Can you please elaborate?

So theoretically I imagine the following:
1. Poller starts, sees that the `pullMethod` is set to node.
2. Adds an ephemeral container with the image to itself via the `ephemeralcontainers` subresource.
3. Queries its status for the ephemeral container's image UID and proceeds to update the DIC digest as it normally would.

It's a bit of a roundabout way of achieving things so I don't mind either solution, but just wanted to hear your thoughts.

pkg/controller/dataimportcron-controller.go
arnongilboa: > How about using `sleep`?

You have no binaries at all (`sleep`, `true` etc.) when your container image is a random containerDisk, which is probably an OS boot image but it doesn't help.

arnongilboa: > Technically any entrypoint that can simply keep the Pod alive should be fine. In an image was built with `FROM: scratch` we should be able to still use it even if the Pod reaches a `Failed` state as we're only looking for the container image digest anyway.
> 
> I understand if you'd want to avoid that, though.

Find me a cheaper entrypoint, which must be a statically-linked binary (don't say `busybox:uclibc`) and I'll use it. We already have `cdi-importer` cached, and that's why I choose the dummy `cdi-containerimage-server -h` as you get it for free...

arnongilboa:
> > Not sure I understand how you create that ephemeral container and how the poller access the image. Can you please elaborate?
> 
> So theoretically I imagine the following:
> 
> 1. Poller starts, sees that the `pullMethod` is set to node.
> 2. Adds an ephemeral container with the image to itself via the `ephemeralcontainers` subresource.
> 3. Queries its status for the ephemeral container's image UID and proceeds to update the DIC digest as it normally would.
> 
> It's a bit of a roundabout way of achieving things so I don't mind either solution, but just wanted to hear your thoughts.

The whole original problem which led us to add support for `OpenShift ImageStream`  was that the cronjob-created pod cannot pull `containerImages` from registry such as `registry.redhat.io` that requires authentication, and since the poller has no way to get the credentials, it will get something like: `Failed to get image digest: failed to pull image: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication`

pkg/controller/dataimportcron-controller.go
```diff
@@ -617,6 +631,126 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
...
+		}
+		return "", nil
+	}
+
+	if status.State.Terminated == nil {
+		return "", nil
+	}
+
+	imageID := status.ImageID
+	if len(imageID) == 0 {
```
Acedus: nit:
```suggestion
	if imageID == ""
```

pkg/controller/dataimportcron-controller.go
```diff
@@ -617,6 +631,126 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
...
+		return "", errors.Errorf("Container has no imageID")
+	}
+	idx := strings.Index(imageID, digestSha256Prefix)
+	if idx <= 0 {
```
Acedus: nit: While technically not possible in this context, `idx == 0` is a valid possibility for a match so `idx < 0` would be more correct.

Acedus:
> > > Not sure I understand how you create that ephemeral container and how the poller access the image. Can you please elaborate?
> > 
> > 
> > So theoretically I imagine the following:
> > 
> > 1. Poller starts, sees that the `pullMethod` is set to node.
> > 2. Adds an ephemeral container with the image to itself via the `ephemeralcontainers` subresource.
> > 3. Queries its status for the ephemeral container's image UID and proceeds to update the DIC digest as it normally would.
> > 
> > It's a bit of a roundabout way of achieving things so I don't mind either solution, but just wanted to hear your thoughts.
> 
> The whole original problem which led us to add support for `OpenShift ImageStream` was that the cronjob-created pod cannot pull `containerImages` from registry such as `registry.redhat.io` that requires authentication, and since the poller has no way to get the credentials, it will get something like: `Failed to get image digest: failed to pull image: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication`

The Poller wouldn't attempt to directly infer the image digest from the registry, instead it will send an API call to the `ephemeralcontainers` subresource to add an ephemeral container to itself with the target image, that way the added container pull is delegated to kubelet which is akin to what your existing solution achieves.

In any event, the code is fine as is so this is just something to consider.

Acedus:
/lgtm
@arnongilboa thanks!

pkg/controller/dataimportcron-controller.go
```diff
@@ -617,6 +631,126 @@ func (r *DataImportCronReconciler) updateImageStreamDesiredDigest(ctx context.Co
...
+
+	cc.SetRestrictedSecurityContext(&pod.Spec)
+	if pod.Spec.SecurityContext != nil {
+		pod.Spec.SecurityContext.FSGroup = nil
+	}
+
+	return false, r.client.Create(ctx, pod)
+}
+
+func (r *DataImportCronReconciler) fetchContainerImageDigest(pod *corev1.Pod) (string, error) {
```
akalenyu: you don't have to create this func as a method (`r *DataImportCronReconciler`)

pkg/controller/dataimportcron-controller.go
```diff
@@ -262,15 +262,23 @@ func (r *DataImportCronReconciler) pollSourceDigest(ctx context.Context, dataImp
...
+	case isPvcSource(dataImportCron):
 		if err := r.updatePvcDesiredDigest(ctx, dataImportCron); err != nil {
 			return reconcile.Result{}, err
 		}
+	case isNodePull(dataImportCron):
+		if done, err := r.updateContainerImageDesiredDigest(ctx, dataImportCron); !done {
+			return reconcile.Result{RequeueAfter: 3 * time.Second}, err
+		} else if err != nil {
+			return reconcile.Result{}, err
+		}
```
akalenyu, line 279: maybe worth a unit test?

akalenyu (commented):
Looks good, I guess worst case scenario is if poller ends up on different node but the image should get GCed eventually

> Also, can we add a functional test to exercise this new logic?

Maybe a unit test is fine

pkg/controller/dataimportcron-controller.go
arnongilboa: Right, added.

pkg/controller/dataimportcron-controller.go
arnongilboa: Fixed.

Acedus:
Thanks @arnongilboa!
/lgtm

pkg/controller/dataimportcron-controller.go
awels: According to Google, IfNotPresent doesn't download if the hash changes, only if the tag changes.

awels: So I think we should do Always, otherwise this might not work.

awels (approved):
/approve

PR #3799, opened 2025-06-25
dv, conditions: AnnRunningCondition should not take precedence over AnnPodSchedulable

Acedus:
With https://github.com/kubevirt/containerized-data-importer/pull/3753 a new annotation AnnPodSchedulable was introduced to allow updating the running condition of an importer Pod that was unschedulable.

This PR makes it so the unschedulable condition takes precedence over AnnRunningCondition otherwise a race can occur where initial Pod creation failure sets the running condition to an erroneous one indefinitely as the Pod never finishes scheduling and thus never updates.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume/conditions.go
+++ b/pkg/controller/datavolume/conditions.go
@@ -66,7 +66,9 @@ func updateCondition(conditions []cdiv1.DataVolumeCondition, conditionType cdiv1
 }
 
 func updateRunningCondition(conditions []cdiv1.DataVolumeCondition, anno map[string]string) []cdiv1.DataVolumeCondition {
-	if val, ok := anno[cc.AnnRunningCondition]; ok {
+	if schedulable, ok := anno[cc.AnnPodSchedulable]; ok && schedulable == "false" {
+		conditions = updateCondition(conditions, cdiv1.DataVolumeRunning, corev1.ConditionFalse, "Importer pod cannot be scheduled", "Unschedulable")
+	} else if val, ok := anno[cc.AnnRunningCondition]; ok {
 		switch strings.ToLower(val) {
 		case "true":
 			conditions = updateWithTargetRunning(conditions, anno)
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
> where initial Pod creation failure

how likely is it to fail on pod creation? does the api create just error? bit worrying

Acedus:
> > where initial Pod creation failure
> 
> how likely is it to fail on pod creation? does the api create just error? bit worrying

So an example scenario (which was failing a test):
1. Importer Pod is attempting creation and fails for some reason, could be that the namespace it was attempting creation in was being terminated at the time of the request.
2. The PVC now has `AnnRunningCondition` set to `ErrStartingPod`.
3. Subsequent attempt to create the Importer Pod will succeed however the scheduler will prevent it from being scheduled due to the mismatched selector, this will in turn set the `AnnPodSchedulable` annotation on the PVC to false.
4. Because `AnnRunningCondition` takes precedence right now, we'll reach a deadlock where the condition doesn't change as the Pod never actually reaches a running state nor does the DV report that it is, in fact, unschedulable.

akalenyu:
> > > where initial Pod creation failure
> > 
> > 
> > how likely is it to fail on pod creation? does the api create just error? bit worrying
> 
> So an example scenario (which was failing a test):
> 
> 1. Importer Pod is attempting creation and fails for some reason, could be that the namespace it was attempting creation in was being terminated at the time of the request.
> 2. The PVC now has `AnnRunningCondition` set to `ErrStartingPod`.
> 3. Subsequent attempt to create the Importer Pod will succeed however the scheduler will prevent it from being scheduled due to the mismatched selector, this will in turn set the `AnnPodSchedulable` annotation on the PVC to false.
> 4. Because `AnnRunningCondition` takes precedence right now, we'll reach a deadlock where the condition doesn't change as the Pod never actually reaches a running state nor does the DV report that it is, in fact, unschedulable.

I totally understand the scenario and it makes sense to me to fix this. I'm just more worried about the initial creation failures. It's very unlikely that
> the namespace it was attempting creation in was being terminated at the time of the request.

Since your test was being executed at that point. I would get quota failures causing this but AFAIK that wasn't the case (each test runs in it's own ns)

/approve

akalenyu:
@arnongilboa [fyi](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3799/pull-containerized-data-importer-e2e-ceph/1937767208914522112)
/test pull-containerized-data-importer-e2e-hpp-previous
fails to run tests

PR #3802, opened 2025-06-26
Tests: Fix flaky variable assignments during Update

alromeros:

This PR fixes a flaky variable assignment in tests during Update by using Patch. When a PVC or DV Update call fails, then the errored PVC assignment may be used in subsequent retries and fail. Patch is atomic so these operations shouldn't fail normally.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -23,6 +23,8 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	openapicommon "k8s.io/kube-openapi/pkg/common"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/common"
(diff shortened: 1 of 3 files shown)


ShellyKa13 (commented):
If you are changing this, maybe its better to do patch instead of update? then it shouldnt fail and no need for the eventually..

alromeros:
> If you are changing this, maybe its better to do patch instead of update? then it shouldnt fail and no need for the eventually..

I was torn about this because I plan to backport it to 1.60 to avoid some D/S failures and this may introduce more complexity than I wanted. I ended up updating both tests and another import one to use patch instead of update. Hope this looks good.

tests/upload_test.go
```diff
@@ -105,19 +106,19 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		}
 	})
 
+	// TODO: Move this to a common package once we adopt Patch in all the codebase
+	escapeJSONPointer := func(ptr string) string {
```
ShellyKa13: I believe we can copy kubevirts https://github.com/kubevirt/kubevirt/blob/main/pkg/apimachinery/patch/patch.go
but as mentioned it can be done in another PR. I wont block this one for it..

akalenyu: check openapicommon.EscapeJsonPointer

alromeros: ack

alromeros:
The import test is still flaky in older versions, doesn't hurt to backport this way back
/cherrypick release-v1.59

PR #3803, opened 2025-06-26
Support putting external images into the local registry of the kubevirtci cluster (#3801)

### LINKED ISSUE #3801: Support putting external images into the local registry of the kubevirtci cluster
When starting up a dev cluster using `make cluster-up` and `make cluster-sync`, some resources are deployed. Some of the resources use images outside the local registry such as from `quay.io` etc. If a developer/contributor works in an isolated environment that doesn't have access to those `external` registries the deployment will fail because it cannot pull images from those remote registries.

To solve the described issue above, one way should be to have a script to manually pull a list of
such external images into the local registry so that those resources can referencing the images from the local registry.

gaohoward:
User can define an environment variable EXTRA_IMAGES that contains a list of external image urls separated by spaces.

At the end of 'make cluster-up' target the EXTRA_IMAGES are processed by a script ./cluster-up/extra-images.sh so that for each image url, it pulls the image and then push the image into the local registry.

Then when 'make cluster-sync' is invoked, it searches all external image urls in the generated yamls in _out/manifests/ and replaces them with corresponding urls in the local registry (registry:5000)

Added functionality for cluster-up target to allow adding external images to the internal local docker registry

Fixes #3801

**Release note**:

```release-note
Normal users should not affected by this feature. Only users with isolated environments can use this feature to add
any external images to the local registry. Make sure your local machine have access to the external repositories in order to pull them and push to the local registry.
```


hack/extra-images.sh
```diff
@@ -0,0 +1,22 @@
...
+script_dir="$(cd "$(dirname "$0")" && pwd -P)"
+source "${script_dir}"/build/common.sh
+
+cri_bin=$1
+if [ -z $cri_bin ]; then
+    cri_bin=$CDI_CRI
+fi
+
+if [ -n "${EXTRA_IMAGES}" ]; then
+    registry_port=$(./cluster-up/cli.sh ports registry | xargs)
```
awels: if you 
```bash
source ./cluster-up/cluster/${KUBEVIRT_PROVIDER}/provider.sh
```
then you can call `_port registry` to get the port.

cluster-sync/sync.sh
```diff
@@ -64,6 +64,17 @@ fi
 DOCKER_PREFIX=$MANIFEST_REGISTRY PULL_POLICY=$PULL_POLICY make manifests
 DOCKER_PREFIX=$DOCKER_PREFIX make push
 
+# Update manifests to replace extra images with internal registry ones
+if [ -n "${EXTRA_IMAGES}" ]; then
```
awels: does it make sense to do this in the script that is called with `make manifests`?

hack/extra-images.sh
```diff
@@ -0,0 +1,22 @@
...
+
+script_dir="$(cd "$(dirname "$0")" && pwd -P)"
+source "${script_dir}"/build/common.sh
+
+cri_bin=$1
+if [ -z $cri_bin ]; then
+    cri_bin=$CDI_CRI
+fi
+
+if [ -n "${EXTRA_IMAGES}" ]; then
```
awels, line 15: I am thinking maybe we should call these `EXTERNAL_IMAGE_MINIO` and `EXTERNAL_IMAGE_FAKEOVER` and default them to the existing images with something like:
```
EXTERNAL_IMAGE_MINIO=${EXTERNAL_IMAGE_MINIO:-quay.io/kubevirt/minio:RELEASE.2024-10-02T17-50-41Z}
EXTERNAL_IMAGE_FAKEOVIRT=${EXTERNAL_IMAGE_FAKEOVIRT:-quay.io/kubevirt/fakeovirt:v1.38.0}
```

Then we can modify `hack/build/resource-generator.sh` and add some new variables for these two variables, and put them directly in the manifest templates in `manifests/templates/file-host.yaml.in` and `manifests/templates/imageio.yaml.in`

Then we can override each one individually.

gaohoward:
@awels Hi Alex, just updated the branch. Pls take a look and see if it looks better. Thanks!

PR #3807, opened 2025-07-03
Add missing RBAC to forklift volume populator CRDs.

mrnold:
This continues pull request #3320 to address feedback on the forklift populator policy rules.

Hopefully fixes [CNV-64424](https://issues.redhat.com/browse/CNV-64424), and a bunch of older clones

This passes the previously failing unit tests that check for wildcards in the verb lists, and gets `system:cluster-readers` to show up in the output of the `oc adm policy who-can get` commands shown in [CNV-64424](https://issues.redhat.com/browse/CNV-64424). I don't have a deep enough understanding of RBAC to tell if anything else is missing though.

**Release note**:
```release-note
Add missing RBAC for ovirt and openstack volume populator CRDs
```

### DIFF
--- a/pkg/operator/resources/cluster/rbac.go
+++ b/pkg/operator/resources/cluster/rbac.go
@@ -89,14 +89,7 @@ func getAdminPolicyRules() []rbacv1.PolicyRule {
 				"openstackvolumepopulators",
 			},
 			Verbs: []string{
-				"create",
-				"delete",
-				"deletecollection",
-				"get",
-				"list",
-				"patch",
-				"update",
-				"watch",
+				"*",
 			},
 		},
 	}
--- a/pkg/operator/resources/cluster/rbac.go
+++ b/pkg/operator/resources/cluster/rbac.go
@@ -168,20 +168,6 @@ func createConfigReaderClusterRole(name string) *rbacv1.ClusterRole {
 				"watch",
 			},
 		},
-		{
-			APIGroups: []string{
-				"forklift.cdi.kubevirt.io",
-			},
-			Resources: []string{
-				"ovirtvolumepopulators",
-				"openstackvolumepopulators",
-			},
-			Verbs: []string{
-				"get",
-				"list",
-				"watch",
-			},
-		},
 	}
 
 	return utils.ResourceBuilder.CreateClusterRole(name, rules)
--- a/pkg/operator/resources/cluster/rbac.go
+++ b/pkg/operator/resources/cluster/rbac.go
@@ -30,13 +30,8 @@ func createAggregateClusterRoles(_ *FactoryArgs) []client.Object {
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:admin", "admin", getAdminPolicyRules()),
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:edit", "edit", getEditPolicyRules()),
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:view", "view", getViewPolicyRules()),
-		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:admin", "admin", getAdminPolicyRules()),
-		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:edit", "edit", getEditPolicyRules()),
-		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:view", "view", getViewPolicyRules()),
 		createConfigReaderClusterRole("cdi.kubevirt.io:config-reader"),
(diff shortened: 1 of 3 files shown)


pkg/operator/resources/cluster/rbac.go
```diff
@@ -75,6 +80,25 @@ func getAdminPolicyRules() []rbacv1.PolicyRule {
...
+			},
+			Verbs: []string{
+				"create",
+				"delete",
+				"deletecollection",
+				"get",
+				"list",
+				"patch",
+				"update",
+				"watch",
```
akalenyu: So this is completely fine, but you could have also extended the exception of * to for forklift.cdi.kubevirt.io if you preferred that. The idea behind that exception is that "cdi can do anything on cdi resources" which I believe the forklift resources fall under as well.

pkg/operator/resources/cluster/rbac.go
```diff
@@ -30,8 +30,13 @@ func createAggregateClusterRoles(_ *FactoryArgs) []client.Object {
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:admin", "admin", getAdminPolicyRules()),
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:edit", "edit", getEditPolicyRules()),
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:view", "view", getViewPolicyRules()),
+		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:admin", "admin", getAdminPolicyRules()),
+		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:edit", "edit", getEditPolicyRules()),
+		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:view", "view", getViewPolicyRules()),
 		createConfigReaderClusterRole("cdi.kubevirt.io:config-reader"),
+		createConfigReaderClusterRole("forklift.cdi.kubevirt.io:config-reader"),
```
akalenyu: I'm confused, does forklift need it's own config reader role?

pkg/operator/resources/cluster/rbac.go
mrnold: Okay, I changed this back to * and reworked the unit test to allow it. Is [this](https://github.com/kubevirt/containerized-data-importer/pull/3807/commits/8a32c954f7db4c57d5f635dccb70ed69cff72569) what you had in mind?

pkg/operator/resources/cluster/rbac.go
mrnold: I am not entirely sure. In cmd/ovirt-populator and cmd/openstack-populator, I see references to a `secretName` that looks like it was meant to get credentials from a secret. I guess this would need a config-reader role, but I don't see anything that actually reads that secret. Maybe this was planned for later.

akalenyu: This role exists so anyone can read cdi config https://github.com/kubevirt/containerized-data-importer/blob/115d66a61c8a9d33b2fbdc27e3f36dff8d89f9b3/pkg/operator/resources/cluster/rbac.go#L160-L170

pkg/operator/resources/cluster/rbac.go
```diff
@@ -30,8 +30,13 @@ func createAggregateClusterRoles(_ *FactoryArgs) []client.Object {
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:admin", "admin", getAdminPolicyRules()),
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:edit", "edit", getEditPolicyRules()),
 		utils.ResourceBuilder.CreateAggregateClusterRole("cdi.kubevirt.io:view", "view", getViewPolicyRules()),
+		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:admin", "admin", getAdminPolicyRules()),
+		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:edit", "edit", getEditPolicyRules()),
+		utils.ResourceBuilder.CreateAggregateClusterRole("forklift.cdi.kubevirt.io:view", "view", getViewPolicyRules()),
```
akalenyu: sure you need a separate role? can't we just use CDIs?

pkg/operator/resources/cluster/rbac.go
akalenyu: Yes

pkg/operator/resources/cluster/rbac.go
mrnold: Okay, I removed the forklift role and still see passing tests. I haven't tried to actually run the populator yet, but I don't see a reason it can't just use CDI's roles.

akalenyu (commented):
/approve
/lgtm

PR #3812, opened 2025-07-07
Increase default filesystem overhead to 6%

alromeros:

The previous default of 5.5% was originally chosen to account for filesystem overhead. Due to inaccuracies in our overhead inflation calculations, the actual effective overhead was closer to 5.8%.
    
Now that the calculation functions have been fixed in https://github.com/kubevirt/containerized-data-importer/pull/3779, we're increasing the default to 6% to ensure sufficient inflation for typical filesystem overhead and avoid under-provisioning.

Feel free to propose a different value. 7% was considered in the past (I think around https://github.com/kubevirt/kubevirt/pull/13326 was merged) but 6% is closer to the original and apparently solves the issue we had with smaller PVCs. 

**Release note**:

```release-note
Increase default filesystem overhead to 6%
```

### DIFF
--- a/tests/import_test.go
+++ b/tests/import_test.go
@@ -1295,8 +1295,6 @@ var _ = Describe("Preallocation", func() {
 	)
 
 	It("Filesystem overhead is honored with blank volume", Serial, func() {
-		tests.SetFilesystemOverhead(f, "0.055", "0.055")
-
 		dv := utils.NewDataVolumeForBlankRawImage("import-dv", "100Mi")
 		preallocation := true
 		dv.Spec.Preallocation = &preallocation
--- a/tests/upload_test.go
+++ b/tests/upload_test.go
@@ -30,7 +30,6 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/controller/populators"
 	"kubevirt.io/containerized-data-importer/pkg/image"
 	"kubevirt.io/containerized-data-importer/pkg/util"
-	"kubevirt.io/containerized-data-importer/tests"
 	"kubevirt.io/containerized-data-importer/tests/framework"
 	"kubevirt.io/containerized-data-importer/tests/utils"
 )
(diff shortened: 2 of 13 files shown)


mhenriks:
how is upgrade handled? If the user never specified their own overhead will the new default be applied?

alromeros:
> how is upgrade handled? If the user never specified their own overhead will the new default be applied?

The controller reconciles the overhead by setting the default value if none is specified in the spec https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/config-controller.go#L504 so yeah, if the user didn't set any overhead the new default should be applied on upgrade.

alromeros:
@awels @akalenyu @mhenriks Hi, would like to unblock this PR. Assuming the upgrade is handled in case the user didn't set a default overhead, are we all ok with this increase?

tests/upload_test.go
```diff
@@ -1371,7 +1371,7 @@ var _ = Describe("[rfe_id:138][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		// is insufficient to account for the fs overhead in small images.
 		// This issue is not seen with larger images where the overhead is sufficient.
 		size := "2147483648"
-		fsOverhead := "0.055" // The default value
+		fsOverhead := "0.06" // The default value
 		tests.SetFilesystemOverhead(f, fsOverhead, fsOverhead)
```
akalenyu: I'm not sure why this test sets the default? should already be there

tests/import_test.go
```diff
@@ -1295,7 +1295,7 @@ var _ = Describe("Preallocation", func() {
 	)
 
 	It("Filesystem overhead is honored with blank volume", Serial, func() {
-		tests.SetFilesystemOverhead(f, "0.055", "0.055")
+		tests.SetFilesystemOverhead(f, "0.06", "0.06")
```
akalenyu: Same point about "setting" it to the default

akalenyu (commented):
looks good, really minor comment

tests/upload_test.go
alromeros: I assumed we wanted to make sure we are using the default here in case we chose a different value in cdiConfig, but just might be some leftover from legacy behavior. I can remove it if you want.

tests/import_test.go
alromeros: same as above

tests/upload_test.go
akalenyu: It seems off to me, if other tests change it they probably:
- label themselves Serial
- Revert back in AfterEach

Regarding the "input cluster" having a different overhead, I'd say we want to test it as is, but I'm not aware of such testing setup existing

alromeros: ack

PR #3809, opened 2025-07-06
Targeted fsync in resource generating

akalenyu:
**What this PR does / why we need it**:
These syncs exist as a bandaid on a CI issue;
They recently started causing the entire job to fail over the sync itself.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3806/pull-containerized-data-importer-e2e-hpp-latest/1940758388946243584

**Release note**:

```release-note
NONE
```

PR #3831, opened 2025-07-08
fix: add DataSource reference handling to authorize utils

Acedus:
A previous PR[1] introduced the concept of DataSource references as a way for a DataSource to reference another DataSource.

A DV that has its sourceRef set to a DataSource reference that resides in a different namespace would fail as authorize utils did not include the necessary changes to the clone source handler to accomodate DataSource references 

This PR fixes that and introduces a functional test for it.

[1] https://github.com/kubevirt/containerized-data-importer/pull/3760

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -393,6 +393,7 @@ var (
 
 	ErrDataSourceMaxDepthReached = errors.New("DataSource reference chain exceeds maximum depth of 1")
 	ErrDataSourceSelfReference   = errors.New("DataSource cannot self-reference")
+	ErrDataSourceCrossNamespace  = errors.New("DataSource cannot reference a DataSource in another namespace")
 )
 
 // FakeValidator is a fake token validator
@@ -2087,9 +2088,13 @@ func ResolveDataSourceChain(ctx context.Context, client client.Client, dataSourc
 
 	ref := dataSource.Spec.Source.DataSource
 	refNs := GetNamespace(ref.Namespace, dataSource.Namespace)
+	if dataSource.Namespace != refNs {
+		return dataSource, ErrDataSourceCrossNamespace
+	}
 	if ref.Name == dataSource.Name && refNs == dataSource.Namespace {
 		return nil, ErrDataSourceSelfReference
 	}
+
 	resolved := &cdiv1.DataSource{}
 	if err := client.Get(ctx, types.NamespacedName{Name: ref.Name, Namespace: refNs}, resolved); err != nil {
 		return nil, err
--- a/pkg/controller/datasource-controller.go
+++ b/pkg/controller/datasource-controller.go
@@ -60,6 +60,7 @@ const (
 	dataSourceControllerName = "datasource-controller"
 	maxReferenceDepthReached = "MaxReferenceDepthReached"
 	selfReference            = "SelfReference"
+	crossNamespaceReference  = "CrossNamespaceReference"
 )
 
 // Reconcile loop for DataSourceReconciler
@@ -181,6 +182,8 @@ func handleDataSourceRefError(dataSource *cdiv1.DataSource, err error) error {
 		reason = maxReferenceDepthReached
 	case errors.Is(err, cc.ErrDataSourceSelfReference):
 		reason = selfReference
+	case errors.Is(err, cc.ErrDataSourceCrossNamespace):
+		reason = crossNamespaceReference
 	case k8serrors.IsNotFound(err):
 		reason = cc.NotFound
 	default:
--- a/pkg/controller/datasource-controller_test.go
+++ b/pkg/controller/datasource-controller_test.go
@@ -243,6 +243,16 @@ var _ = Describe("All DataSource Tests", func() {
 				Expect(dsPointer.Spec.Source.DataSource.Name).To(Equal(dsPointer.Name))
 				Expect(dsPointer.Status.Source.DataSource).To(BeNil())
 			})
+			It("DataSource pointer should fail to resolve cross-namespace DataSource", func() {
+				dsPointer := createDataSource(dsName + "-pointer")
+				dsPointer.Spec.Source = cdiv1.DataSourceSource{DataSource: &cdiv1.DataSourceRefSourceDataSource{Namespace: "differentNamespace", Name: dsPointer.Name}}
+				reconciler := createDataSourceReconciler(dsPointer)
+				dsPointerKey := types.NamespacedName{Name: dsPointer.Name, Namespace: metav1.NamespaceDefault}
+				verifyConditions("DataSource cross-namespace reference", false, crossNamespaceReference, dsPointer, reconciler)
+				Expect(reconciler.client.Get(context.TODO(), dsPointerKey, dsPointer)).To(Succeed())
+				Expect(dsPointer.Spec.Source.DataSource.Name).To(Equal(dsPointer.Name))
+				Expect(dsPointer.Status.Source.DataSource).To(BeNil())
+			})
 		})
 
 	})


staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize_utils.go
```diff
@@ -45,11 +45,12 @@ func newCloneSourceHandler(dataVolume *DataVolume, dsGet dsGetFunc) (CloneSource
...
-		if dataSource.Spec.Source.PVC != nil {
-			pvcSource = dataSource.Spec.Source.PVC
-		} else if dataSource.Spec.Source.Snapshot != nil {
-			snapshotSource = dataSource.Spec.Source.Snapshot
-		}
+		pvcSource = dataSource.Spec.Source.PVC
+		snapshotSource = dataSource.Spec.Source.Snapshot
+		if dataSource.Spec.Source.DataSource != nil {
+			pvcSource = dataSource.Status.Source.PVC
+			snapshotSource = dataSource.Status.Source.Snapshot
```
arnongilboa, line 52: Since you are referring the Status here, can you be sure it's populated?

Acedus: I can't, but the switch case after that makes it safe.

tests/datasource_test.go
```diff
@@ -358,6 +367,31 @@ var _ = Describe("DataSource", func() {
 			_ = waitForReadyCondition(ds2, corev1.ConditionFalse, "NotFound")
 		})
 	})
+
+	Context("DataSource references", func() {
```
akalenyu: I am a bit split on this, on the one hand we could probably reduce this to a unit test in the webhook, on the other, we could skip the datasource layer and have a dataimportcron func test to make sure this critical cross ns path is covered. WDYT?

Acedus: I don't mind, I think both will be good.

akalenyu: It would be ideal if you could add a non-serial describe block in dataimportcron_test.go so we minimize the cost, but I am not 100% sure it's possible since there's some manipulation of cluster state in there.

akalenyu:
The CI's host wide storage is notorious for triggering OOMs (slow flush?). It dies when writing to scratch which we've never implemented O_DIRECT writing for.

tests/datasource_test.go
Acedus: TBF I'm not sure how much I prefer the func test to be part of the dataimportcron section, this current func test exercises the only scenario where using DataSource references is even possible to begin with.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/bd735ca973eee0cd2e7e63afe9d3ed237737f4fb..cd1dfa7685b777c10d79a791f980a765e8fa37c9) - Added unit test as per @akalenyu's comment. LMK if you think something should be done differently.

tests/datasource_test.go
akalenyu: Cancel that comment, I got confused thinking cron will manage a datasource that's referencing another datasource

akalenyu (commented):
/approve

akalenyu:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3831/pull-containerized-data-importer-e2e-upg/1942865990899470336 could be related to the changes in this PR (dangling cross ns finalizer)

```
{NamespaceContentRemaining True 2025-07-09 09:19:23 +0000 UTC SomeResourcesRemain Some resources are remaining: datavolumes.cdi.kubevirt.io has 1 resource instances} {NamespaceFinalizersRemaining True 2025-07-09 09:19:23 +0000 UTC SomeFinalizersRemain Some content in the namespace has finalizers remaining: cdi.kubevirt.io/dataVolumeFinalizer in 1 resource instances}]}}]
```

mhenriks:
/hold
@Acedus I apologize for being unaware of this PR and its predecessor. But my reading is that target could be namespace A, DataSource in namespace B, and source namespace C

If so, we validate that user can access namespace C but not B.

When datasource was introduced the intention was only to support jumping one namespace so:
1. DataSource in target namespace, source in another
2. DataSource and source in same namespace

Do these conditions still hold true?

Acedus:
> @Acedus I apologize for being unaware of this PR and its predecessor. But my reading is that target could be namespace A, DataSource in namespace B, and source namespace C
> 
> If so, we validate that user can access namespace C but not B.
> 
> When datasource was introduced the intention was only to support jumping one namespace so:
> 
> 1. DataSource in target namespace, source in another
> 2. DataSource and source in same namespace
> 
> Do these conditions still hold true?

That's a really good point, you're right. This scenario requires more finesse, I'll adjust the PR to match that.
EDIT: For the sake of public record, this vulnerability has existed for all DataSources since https://github.com/kubevirt/containerized-data-importer/pull/1851, this new version is simply a different manifestation of it.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/cd1dfa7685b777c10d79a791f980a765e8fa37c9..7a868ffce6e3997fd5d4ee3e37ed7058e57f24d6) - Addressed @mhenriks comment and added FT for it. Dropped previously added FT as logic is being tested properly as part of the new UT and FT.

Note that the previously mentioned issue of permission verification a single datasource (no intermediate) is still unresolved and will be addressed in a follow-up PR.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize.go
```diff
@@ -72,6 +72,7 @@ func (dv *DataVolume) AuthorizeUser(requestNamespace, requestName string, proxy
 	if sourceNamespace == "" {
 		sourceNamespace = targetNamespace
 	}
+	intermediateDSName, intermediateDSNamespace := cloneSourceHandler.IntermediateDSName, cloneSourceHandler.IntermediateDSNamespace
```
akalenyu: do you want defaulting for intermediateDSNamespace in case its ""?

tests/webhook_test.go
```diff
@@ -350,5 +369,115 @@ var _ = Describe("Clone Auth Webhook tests", func() {
...
+				}, 60*time.Second, 2*time.Second).ShouldNot(HaveOccurred())
+
+				// let's do another manual check as well
+				response, err = targetDV.AuthorizeSA(targetNamespace.Name, targetDV.Name, proxy, targetNamespace.Name, serviceAccountName)
+				Expect(response.Allowed).To(BeTrue())
+				Expect(response.Reason).To(BeEmpty())
+				Expect(err).ToNot(HaveOccurred())
+			},
+				Entry("when using explicit CDI permissions", explicitRole, serviceAccountName, "", false),
+				Entry("when using explicit CDI permissions without permissions on intermediate datasource", explicitRole, serviceAccountName, "", true),
```
akalenyu: I would pass the role that makes this succeed as part of entry args instead of the !fail bit

akalenyu (commented):
Looks good

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/authorize.go
Acedus: It's handled in the case that it's "".

akalenyu: we don't mind if `handler.SourceNamespace` gets passed as `""` to `proxy.GetNamespace(sourceNamespace)` for example?

Acedus: We do, that's why we set it to `targetNamespace` in the case that it is `""`.
`sourceNamespace` must be a valid namespace, but not all clones require an intermediate namespace.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/617eae20b63dcdb253e91cf20fbcf63107ff09df..19b5bdff184a310de02cdd94a65b00eeb5d1926a) - Addressed @akalenyu's comments.

tests/webhook_test.go
```diff
@@ -350,5 +350,133 @@ var _ = Describe("Clone Auth Webhook tests", func() {
...
+				}, 60*time.Second, 2*time.Second).ShouldNot(HaveOccurred())
+
+				// let's do another manual check as well
+				response, err = targetDV.AuthorizeSA(targetNamespace.Name, targetDV.Name, proxy, targetNamespace.Name, serviceAccountName)
+				Expect(response.Allowed).To(BeTrue())
+				Expect(response.Reason).To(BeEmpty())
+				Expect(err).ToNot(HaveOccurred())
+			},
+				Entry("when using explicit CDI permissions", explicitRole, explicitRoleDataSource, serviceAccountName, ""),
+				Entry("when using explicit CDI permissions without permissions on intermediate datasource", explicitRole, nil, serviceAccountName, ""),
```
akalenyu: If you want you can also accept `...*rbacv1.Role` but that's up to you

akalenyu (commented):
My approval exists on the PR so going to give mhenriks a chance to take a look too

tests/webhook_test.go
Acedus: I need to apply the roles and bind them in different namespaces so the iteration on the variadic parameter to parse where I need to create them makes this redundant.

akalenyu:
> @Acedus: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-containerized-data-importer-e2e-istio	[19b5bdf](https://github.com/kubevirt/containerized-data-importer/commit/19b5bdff184a310de02cdd94a65b00eeb5d1926a)	[link](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3831/pull-containerized-data-importer-e2e-istio/1945007582640869376)	true	`/test pull-containerized-data-importer-e2e-istio`

I think something is actually wrong here. poller is being recreated with different uid https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3831/pull-containerized-data-importer-e2e-istio/1945007582640869376/artifacts/1/serial_1_events.log

/cc @arnongilboa

arnongilboa:
@Acedus why not simply force the intermediate DataSource namespace to be the same as the pointing DataSource namespace. I see no reason to support namespace-crossing between DataSources.

Acedus:
> @Acedus why not simply force the intermediate DataSource namespace to be the same as the pointing DataSource namespace. I see no reason to support namespace-crossing between DataSources.

Sure, I have no objections.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/19b5bdff184a310de02cdd94a65b00eeb5d1926a..522269d1b179c60f07c85601c4c30a39e2077e6e) - Addressed @arnongilboa's comment. Dropped previously introduced commit for DataSource intermediate namespaces.

Acedus:
fyi @akalenyu, would appreciate a re-approval.

Acedus:
/hold

Waiting for re-approval of @akalenyu

akalenyu:
> > @Acedus why not simply force the intermediate DataSource namespace to be the same as the pointing DataSource namespace. I see no reason to support namespace-crossing between DataSources.
> 
> Sure, I have no objections.

~~Me too, definitely makes things a lot easier, but I am little worried we're taking away something that previously worked~~
Missed that it's only done for datasource chain resolving
/hold cancel
/approve

akalenyu:
/hold cancel
/approve

https://github.com/kubevirt/containerized-data-importer/pull/3809

pkg/controller/common/util.go
```diff
@@ -393,6 +393,7 @@ var (
 
 	ErrDataSourceMaxDepthReached = errors.New("DataSource reference chain exceeds maximum depth of 1")
 	ErrDataSourceSelfReference   = errors.New("DataSource cannot self-reference")
+	ErrDataSourceCrossNamespace  = errors.New("DataSource cannot reference a DataSource in another namespace")
```
mhenriks, line 396: love it

Acedus:
/retest-required

Failures seem to be the same as the previous ones, unrelated to this PR.

PR #3837, opened 2025-07-14
metrics-server: use SecureServing option

Acedus:

As part of an effort to standardize metric exposure across KubeVirt components on port 8443, we are transitioning to HTTPS with TLS encryption for the metrics-server.

To facilitate this, we leverage the controller-runtime's SecureServing option, which creates a self-signed certificate and configures it as the server certificate for the metrics endpoint when no external certificate is provided[1].

Subsequent PRs will replace this self-signed certificate with a CDI generated one to enable a fully trusted and secure connection between the Prometheus instance and the target metrics endpoints as specified by the CDI ServiceMonitor. Until that integration is complete, the ServiceMonitor will be configured with insecureSkipVerify to allow scraping despite the untrusted certificate.

[1] https://github.com/kubernetes-sigs/controller-runtime/pull/2407

Fixes #

**Release note**:

```release-note
Metrics port for cdi-prometheus-metrics service changed from 8080 to 8443.
```

### DIFF
--- a/cmd/cdi-operator/operator.go
+++ b/cmd/cdi-operator/operator.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"os"
@@ -38,6 +39,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"kubevirt.io/containerized-data-importer/pkg/operator/controller"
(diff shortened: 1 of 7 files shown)


Acedus:
/cc @akalenyu 

FYI: @nunnatsa

cmd/cdi-operator/operator.go
```diff
@@ -96,6 +98,15 @@ func main() {
 		LeaderElectionNamespace:    namespace,
 		LeaderElectionID:           "cdi-operator-leader-election-helper",
 		LeaderElectionResourceLock: "leases",
+		Metrics: server.Options{
```
akalenyu, line 101: Do we want to use https://github.com/openshift/dpu-operator/blob/2b792e8e2fc144b396cd28b87ff369ab11fa97a6/cmd/main.go#L69?

akalenyu: Would be a nice extra touch I think, pretty sure it's defaulted in new projects based on ctrl runtime

Acedus: Yeah I see what they're doing. Looks better, let's do that.

Acedus: So I looked into it and the filters package has an indirect dependency of `grpc-gateway` which gazelle overrides some of its internal dependencies as that package uses bazel-gazelle as well. This in turn makes it so it the `grpc-gateway` build files has a dependency for `//utilities` which would make sense if that was the actual root package but because it isn't it search for `//utilities` in CDI's root directory and obviously fails.

There are a few solutions like upgrading gazelle to allow us to utilize some newer `update-repos` flags to restrict instantiation of `go_repository` to just the things we need (i.e., `grpc-gateway`) and then override gazelle's override so it could resolve the `utilities` dependency correct. Another option is to not update gazelle and just declare a static `go_repository` directive similarly to how `kubevirt/kubevirt` does it but that would mean we need to actively maintain that declaration and ensure that we update the version when we run `deps-update` so it matches the version specified in our go.mod. I'd like to avoid both and just not use this filter for the time being, maybe we could re-examine this later on.

akalenyu: that's painful

akalenyu (approved):
/approve

PR #3852, opened 2025-07-28
Support storageProfile minimumSupportedPVCSize in clone

arnongilboa:
When the target `DataVolume` storage requests a size smaller than the source PVC. For target without size it already worked correctly.

jira-ticket: https://issues.redhat.com/browse/CNV-64264

**Release note**:
```release-note
Support storageProfile minimumSupportedPVCSize in clone
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -1101,6 +1101,51 @@ func AddImportVolumeMounts() []corev1.VolumeMount {
 	return volumeMounts
 }
 
+// GetEffectiveStorageResources returns the maximum of the passed storageResources and the storageProfile minimumSupportedPVCSize.
+// If the passed storageResources has no size, it is returned as-is.
+func GetEffectiveStorageResources(ctx context.Context, client client.Client, storageResources corev1.VolumeResourceRequirements,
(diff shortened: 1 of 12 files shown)


pkg/controller/common/util.go
```diff
@@ -1099,6 +1099,42 @@ func AddImportVolumeMounts() []corev1.VolumeMount {
...
+	if err := client.Get(ctx, types.NamespacedName{Name: sc.Name}, storageProfile); err != nil {
+		return storageResources, IgnoreNotFound(err)
+	}
+
+	requestedSize, hasSize := storageResources.Requests[corev1.ResourceStorage]
+	if !hasSize {
+		return storageResources, nil
+	}
+
+	if val, exists := storageProfile.Annotations[AnnMinimumSupportedPVCSize]; exists {
```
alromeros, line 1136: This logic is reimplemented in several places in the code, does it make sense to have a single helper and reuse it each time?

arnongilboa: The only other place with similar logic is this one, so not sure adding a common helper for this simple will improve the code.
https://github.com/kubevirt/containerized-data-importer/blob/6d6517ae7ac980cb0de36d0db74009438596f529/pkg/controller/datavolume/util.go#L302-L308

arnongilboa: Sure, fixing that. Thanks!

PR #3862, opened 2025-08-01
[release-v1.58] Tests: Fix Get assignment in multi-stage import test

alromeros:

This test has been flaky for a long time: https://issues.redhat.com/browse/CNV-40015.

The issue seems to be caused by an inline DV declaration in the Eventually block, which leads to outdated data in the final check. This PR ensures we use the updated DV when comparing with the VolumeImportSource.

**Release note**:

```release-note
NONE
```


alromeros:
This has been fixed in recent versions but the test is still flaky in v1.59 and v1.58. Doesn't hurt to fix the variable assignment.

PR #3863, opened 2025-08-12
Add status subresource to StorageProfile

noamasu:
We have been observing flaky failures in the StorageProfile functional tests.

The root cause appears to be a race condition. Each test resets the spec upon completion, but the controller is likely still processing the previous state when the next test begins. Without a status field, the test has no way to confirm that the controller has finished reconciling the previous state.

This change introduces a status subresource to the StorageProfile. This allows tests and clients to observe the resource's actual state and wait for the status field to be populated - signaling that reconciliation is complete before proceeding.

The implementation includes:

1. Adding the +kubebuilder:subresource:status annotation to the StorageProfile type.
2. Updating the controller to handle status updates separately from the spec.
3. Regenerating the CRD to support the status subresource.

Fixes https://issues.redhat.com/browse/CNV-48282

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/storageprofile-controller.go
+++ b/pkg/controller/storageprofile-controller.go
@@ -14,6 +14,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -129,15 +130,47 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
 }
 
 func (r *StorageProfileReconciler) updateStorageProfile(prevStorageProfile runtime.Object, storageProfile *cdiv1.StorageProfile, log logr.Logger) error {
-	if prevStorageProfile == nil {
+	var prevSP *cdiv1.StorageProfile
+	if p, ok := prevStorageProfile.(*cdiv1.StorageProfile); ok {
+		prevSP = p
+	}
+
+	if prevSP == nil {
 		return r.client.Create(context.TODO(), storageProfile)
-	} else if !reflect.DeepEqual(prevStorageProfile, storageProfile) {
-		// Updates have happened, update StorageProfile.
+	}
+
+	if storageProfileSpecMetaChanged(prevSP, storageProfile) {
 		log.Info("Updating StorageProfile", "StorageProfile.Name", storageProfile.Name, "storageProfile", storageProfile)
-		return r.client.Update(context.TODO(), storageProfile)
+		if err := r.client.Update(context.TODO(), storageProfile); err != nil {
+			return err
+		}
+	}
+
+	if !reflect.DeepEqual(prevSP.Status, storageProfile.Status) {
+		log.Info("Updating StorageProfile Status", "StorageProfile.Name", storageProfile.Name, "storageProfile", storageProfile)
+		if err := r.client.Status().Update(context.TODO(), storageProfile); err != nil {
+			return err
+		}
 	}
 
 	return nil
+}
+
+// storageProfileSpecMetaChanged returns true if Spec, Labels, or Annotations differ
+func storageProfileSpecMetaChanged(previous, desired *cdiv1.StorageProfile) bool {
+	if previous == nil || desired == nil {
+		return previous != desired
+	}
+	if !apiequality.Semantic.DeepEqual(previous.Spec, desired.Spec) {
+		return true
+	}
+	if !apiequality.Semantic.DeepEqual(previous.GetLabels(), desired.GetLabels()) {
+		return true
+	}
+	if !apiequality.Semantic.DeepEqual(previous.GetAnnotations(), desired.GetAnnotations()) {
+		return true
+	}
+	return false
 }
 
 func (r *StorageProfileReconciler) getStorageProfile(sc *storagev1.StorageClass) (*cdiv1.StorageProfile, runtime.Object, error) {
--- a/pkg/controller/BUILD.bazel
+++ b/pkg/controller/BUILD.bazel
@@ -44,6 +44,7 @@ go_library(
         "//vendor/k8s.io/api/core/v1:go_default_library",
         "//vendor/k8s.io/api/networking/v1:go_default_library",
         "//vendor/k8s.io/api/storage/v1:go_default_library",
+        "//vendor/k8s.io/apimachinery/pkg/api/equality:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/api/errors:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/api/meta:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/api/resource:go_default_library",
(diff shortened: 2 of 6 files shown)


pkg/controller/storageprofile-controller.go
```diff
@@ -130,11 +130,26 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
...
+	}
+
+	prevSP := prevStorageProfile.(*cdiv1.StorageProfile)
+
+	if !reflect.DeepEqual(prevSP.ObjectMeta, storageProfile.ObjectMeta) {
+		log.Info("Updating StorageProfile spec/metadata", "StorageProfile.Name", storageProfile.Name)
+		if err := r.client.Update(context.TODO(), storageProfile); err != nil {
+			return err
+		}
+	}
```
akalenyu: I don't think we ever update the spec, do we?

pkg/controller/storageprofile-controller.go
```diff
@@ -130,11 +130,26 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
...
-	} else if !reflect.DeepEqual(prevStorageProfile, storageProfile) {
-		// Updates have happened, update StorageProfile.
-		log.Info("Updating StorageProfile", "StorageProfile.Name", storageProfile.Name, "storageProfile", storageProfile)
-		return r.client.Update(context.TODO(), storageProfile)
+		log.Info("Creating StorageProfile", "StorageProfile.Name", storageProfile.Name)
+		if err := r.client.Create(context.TODO(), storageProfile); err != nil {
+			return err
+		}
+		log.Info("Creating StorageProfile status", "StorageProfile.Name", storageProfile.Name)
+		return r.client.Status().Update(context.TODO(), storageProfile)
```
akalenyu: I'm confused, why do we need to update the status right after creation?

akalenyu (commented):
Looks good, some questions, and my main concern is mostly about upgrade bugs,
In kubevirt, this was a big deal with the "status updater" package:
https://github.com/kubevirt/kubevirt/pull/12601/commits/aa10116d3ef513c04f4846841a4ec368586a63cb

@mhenriks Do you recall how the issue is surfaced in the kubevirt case? I am not sure if we need something like this in CDI

pkg/controller/storageprofile-controller.go
akalenyu: Objectmeta, my bad

pkg/controller/storageprofile-controller.go
noamasu: according to k8s docs, since we now use a status subresource, the .Create() will discard all configured `.status` changes applied earlier in `reconcileStorageProfile`. because of that there was no status and most tests failed.

the question is, do i want to update(add) the status right after creation? or is it ok not to have status for that small period of time between the creation and the first reconcile?

pkg/controller/storageprofile-controller.go
noamasu: in `reconcileStorageProfile` we just update status and metadata labels, so instead of deepequal the enitre object, i assume we can compare just metadata and status only. 

on a general note, since status subresource is enabled, we have to call 
`r.client.Status().Update` separately  so comparing the entire object wont work anymore.

pkg/controller/storageprofile-controller.go
akalenyu: The status can get added once the creation gets queued up, this is totally fine

pkg/controller/storageprofile-controller.go
akalenyu: yeah ofc, what surprised me is that we would ever need to update anything but the status. but I guess we do update labels/annotations

akalenyu: maybe make this more targeted? so we don't update over random metadata changes?

noamasu: outside of the added status update, im explicitly checking changes in spec/labels/annotations before updating in storageProfileSpecMetaChanged (a suggestion for a better name for this function would be great :D)

akalenyu: I'm not sure we want to allow spec changes in this controller

pkg/controller/storageprofile-controller.go
```diff
@@ -129,17 +130,50 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
...
-		return r.client.Update(context.TODO(), storageProfile)
+		if err := r.client.Update(context.TODO(), storageProfile); err != nil {
+			return err
+		}
+	}
+
+	if !reflect.DeepEqual(prevSP.Status, storageProfile.Status) {
+		log.Info("Updating StorageProfile Status", "StorageProfile.Name", storageProfile.Name, "storageProfile", storageProfile)
+		patch := client.MergeFrom(prevSP.DeepCopy())
+		if err := r.client.Status().Patch(context.TODO(), storageProfile, patch); err != nil {
```
akalenyu: I think you can use Status Update() in here, why did you change?

noamasu: the patch updates the diffs only and not the entire object which is good for when using multiple controllers that want to update different areas in the status, but i guess we only have 1 deployment pod anyway, so yeah this is not really needed... i will bring it back :)

mhenriks:
> @mhenriks Do you recall how the issue is surfaced in the kubevirt case? I am not sure if we need something like this in CDI

I don't think we have to worry about this too much for StorageProfile but the issue is that when upgrading it is possible to silently lose an update()

1. old cdi-deployment running
2. Upgrade started
3. StorageProfile CRD updated
4. Old cdi-deployment calls update() and status field ignored

akalenyu:
> > @mhenriks Do you recall how the issue is surfaced in the kubevirt case? I am not sure if we need something like this in CDI
> 
> I don't think we have to worry about this too much for StorageProfile but the issue is that when upgrading it is possible to silently lose an update()
> 
> 1. old cdi-deployment running
> 2. Upgrade started
> 3. StorageProfile CRD updated
> 4. Old cdi-deployment calls update() and status field ignored

I see. At some point new cdi-deployment boots and all profiles get queued, so should try proper update

akalenyu:
these don't seem related, though https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3863/pull-containerized-data-importer-e2e-ceph-wffc/1966103419387121664 is interesting

pkg/controller/storageprofile-controller.go
```diff
@@ -129,17 +130,49 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
...
+			return err
+		}
 	}
 
 	return nil
 }
 
+// storageProfileSpecMetaChanged returns true if Spec, Labels, or Annotations differ
+func storageProfileSpecMetaChanged(previous, desired *cdiv1.StorageProfile) bool {
+	if previous == nil || desired == nil {
```
alromeros, line 161: Nit: These shouldn’t ever be nil, don't think the check if necessary.

pkg/controller/storageprofile-controller.go
```diff
@@ -129,17 +130,49 @@ func (r *StorageProfileReconciler) reconcileStorageProfile(sc *storagev1.Storage
 }
 
 func (r *StorageProfileReconciler) updateStorageProfile(prevStorageProfile runtime.Object, storageProfile *cdiv1.StorageProfile, log logr.Logger) error {
-	if prevStorageProfile == nil {
+	var prevSP *cdiv1.StorageProfile
```
alromeros, line 133: Nit: Don't love the variable naming here, seems a bit confusing. Not a big deal though

akalenyu:
/test pull-cdi-goveralls
unrelated, nfs lane is failing over something I think would only occur in our unique CI setup

PR #3867, opened 2025-08-17
[WIP]Fix bug with CDIDataImportCronOutdated alert

sradco:
Updated CDIDataImportCronOutdated to fire
only if the issue is related to the
Pre-defined golden images.

Add a CDIUserDefinedDataImportCronOutdated alert
for user defined DIC that will not impact the
operator health.

Updated the namspabe label name fro ns to
namespace, since each alert should report a
namespace.

Signed-off-by: Shirly Radco <sradco@redhat.com>

Fixes #https://issues.redhat.com/browse/CNV-67060

**Release note**:

```release-note
Added CDIUserDefinedDataImportCronOutdated alert and updated CDIDataImportCronOutdated condition so it will fire only for pre-defined golden images.
```

### DIFF
--- a/pkg/monitoring/rules/alerts/operator.go
+++ b/pkg/monitoring/rules/alerts/operator.go
@@ -58,14 +58,26 @@ var operatorAlerts = []promv1.Rule{
 	},
 	{
 		Alert: "CDIDataImportCronOutdated",
-		Expr:  intstr.FromString(`sum by(ns,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false"}) > 0`),
+		Expr:  intstr.FromString(`sum by(namespace,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false", namespace=~"openshift-virtualization-os-images|kubevirt-os-images"}) > 0`),
 		For:   (*promv1.Duration)(ptr.To("15m")),
 		Annotations: map[string]string{
 			"summary": "DataImportCron (recurring polling of VM templates disk image sources, also known as golden images) PVCs are not being updated on the defined schedule",
 		},
 		Labels: map[string]string{
-			severityAlertLabelKey:        "info",
+			severityAlertLabelKey:        "warning",
 			operatorHealthImpactLabelKey: "warning",
+		},
+	},
+	{
+		Alert: "CDIUserDefinedDataImportCronOutdated",
+		Expr:  intstr.FromString(`sum by(namespace,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false", namespace!~"openshift-virtualization-os-images|kubevirt-os-images"}) > 0`),
+		For:   (*promv1.Duration)(ptr.To("15m")),
+		Annotations: map[string]string{
+			"summary": "DataImportCron (recurring polling of VM templates disk image sources, also known as golden images) PVCs are not being updated on the defined schedule",
+		},
+		Labels: map[string]string{
+			severityAlertLabelKey:        "warning",
+			operatorHealthImpactLabelKey: "none",
 		},
 	},
 	{
--- a/pkg/monitoring/metrics/cdi-controller/dataimportcron.go
+++ b/pkg/monitoring/metrics/cdi-controller/dataimportcron.go
@@ -8,7 +8,7 @@ import (
 
 const (
 	// PrometheusCronNsLabel labels the DataImportCron namespace
-	PrometheusCronNsLabel = "ns"
+	PrometheusCronNsLabel = "namespace"
 	// PrometheusCronNameLabel labels the DataImportCron name
 	PrometheusCronNameLabel = "cron_name"
 	// PrometheusCronPendingLabel labels whether the DataImportCron import DataVolume is pending for default storage class


pkg/monitoring/rules/alerts/operator.go
```diff
@@ -58,16 +58,28 @@ var operatorAlerts = []promv1.Rule{
 	},
 	{
 		Alert: "CDIDataImportCronOutdated",
-		Expr:  intstr.FromString(`sum by(ns,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false"}) > 0`),
+		Expr:  intstr.FromString(`sum by(namespace,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false", namespace="openshift-virtualization-os-images"}) > 0`),
```
Acedus: SSP controls the name of that namespace, it isn't limited specifically to OpenShift.

pkg/monitoring/rules/alerts/operator.go
```diff
@@ -58,16 +58,28 @@ var operatorAlerts = []promv1.Rule{
...
 		},
 		Labels: map[string]string{
-			severityAlertLabelKey:        "info",
+			severityAlertLabelKey:        "warning",
 			operatorHealthImpactLabelKey: "warning",
 		},
 	},
+	{
+		Alert: "CDIUserDefinedDataImportCronOutdated",
+		Expr:  intstr.FromString(`sum by(namespace,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false", namespace!="openshift-virtualization-os-images}) > 0`),
```
Acedus: Same here.

Acedus (commented):
Thanks, please see inline comments.

pkg/monitoring/rules/alerts/operator.go
sradco: Not sure what you mean. namespace should reprt the cron namespace. The current metric has a bug.

sradco: Prometheus will overwrite the default namespace with the one we supply.

Acedus (approved):
/cc @akalenyu

pkg/monitoring/rules/alerts/operator.go
```diff
@@ -58,16 +58,28 @@ var operatorAlerts = []promv1.Rule{
 	},
 	{
 		Alert: "CDIDataImportCronOutdated",
-		Expr:  intstr.FromString(`sum by(ns,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false"}) > 0`),
+		Expr:  intstr.FromString(`sum by(namespace,cron_name) (kubevirt_cdi_dataimportcron_outdated{pending="false", namespace=~"openshift-virtualization-os-images|kubevirt-os-images"}) > 0`),
```
akalenyu, line 61: So I am struggling a bit with CDI having to "know" about "golden image namespaces" or "special dataimportcrons".
I think we should move this definition to other operators that encapsulate this knowledge.

Alternatively, we could also label the metric with this information though I don't know if anything gives it away easily.

akalenyu (commented):
Great catch, my concern is about making CDI aware of those namespaces

pkg/monitoring/rules/alerts/operator.go
sradco: From what I was told the images in openshift-virtualization-os-images|kubevirt-os-images are the golden images.

sradco: @nunnatsa am I correct?

machadovilaca:
release note has a typo

looks good to me
but i don't think the PR title is appropriate, what was the bug? the namespace label? but that doesn't justify the addition of a new alert in this PR

sradco:
> release note has a typo
Fixed
>looks good to me
but i don't think the PR title is appropriate, what was the bug? the namespace label? but that doesn't justify the addition of a new alert in this PR

We had to split the alert, but it is still valuable for all namespaces.
The difference is the audience for the alerts.
One alert impact the operator health and the other doesnt

sradco:
Hi @Acedus, I will appreciate a review of this PR.

Acedus:
I had some more time to think about it and I'm a bit split on whether the alert in its current form should even exist in CDI to begin with. As @akalenyu correctly stated, golden images is an SSP concept, it isn't directly part of CDI. 

With that said, the other alert which targets ANY namespace besides the golden images one could belong to CDI.

I'm leaning more towards moving these alerts to SSP, seeing as they have the relevant context.

sradco:
@arnongilboa @nunnatsa  please review. What do you think the comment that @Acedus made about moving these alerts to SSP?

nunnatsa:
@sradco - the whole concept is very hard to implement, because the relevant information is spread across three different components. 

First, the namespace is not the right way to distinguish between pre-prepared and user-defined images, for two reasons: 
1. By default, user-defined images will be created on the *-os-images namespace as well.
2. The images namespace is configurable (in the HyperConverged CR) and may be changed by the user.

The only component that "knows" if an image is pre-defined or user-defined is the HCO. HCO adds the images (the DataImportCronTemplate objects) to SSP, with no indication if they are pre-defined or user defined, or a modified pre-defined. BTW, this information **is** reflected in the HyperConverged status.

However, HCO has no information about the DataImportCron CRs, the DataSources or the VolumeSnapshots. HCO does not know them, nor watch or read them.

Acedus:
I think this may be solvable by adding a label to golden image DICs and adding that label to the timeseries used to make the alerts. That way we allow to shift the alerts into SSP while also allowing users to further discern between default images and user-created ones.

sradco:
@Acedus @arnongilboa @machadovilaca  Since this has a high impact on the operator health indicator, I created a simpler PR https://github.com/kubevirt/containerized-data-importer/pull/3885 to address minor changes.

I will keep this one open and create an RFE for the suggested changes.

akalenyu:
> I think this may be solvable by adding a label to golden image DICs and adding that label to the timeseries used to make the alerts. That way we allow to shift the alerts into SSP while also allowing users to further discern between default images and user-created ones.

+1
Maybe we'd have to iron out some details like whether the label is excluding them from the series or the other way around but overall this is a solid approach imo

nunnatsa:
> > I think this may be solvable by adding a label to golden image DICs and adding that label to the timeseries used to make the alerts. That way we allow to shift the alerts into SSP while also allowing users to further discern between default images and user-created ones.
> 
> +1 Maybe we'd have to iron out some details like whether the label is excluding them from the series or the other way around but overall this is a solid approach imo

Agree. Since the source of all the pre-defined (common) DICs is a static file in the HCO image, we can do it w/o code change in HCO/SSP, IIUC.

We can use either label or an annotation.

PR #3868, opened 2025-08-17
Add Prometheus rules unit tests

sradco:
This PR add unit tests for existing alerts.
The Prom unit test are important for making sure the alert behaves as expected.

- Code was copied from the kubevirt repository and updated to fit here.
- All other sub operators already include this code.
- Assisted by: Cursor

Fixes #https://issues.redhat.com/browse/CNV-67430

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ build-functest: ## Build the functional tests (content of tests/ subdirectory)
 test: test-unit test-functional test-lint ## execute all tests (_NOTE:_ 'WHAT' is expected to match the go cli pattern for paths e.g. './pkg/...'.  This differs slightly from rest of the 'make' targets)
 
 test-unit: WHAT = ./pkg/... ./cmd/...
-test-unit: ## Run unit tests.
+test-unit: prom-rules-verify ## Run unit tests.
 	${DO} "ACK_GINKGO_DEPRECATIONS=${ACK_GINKGO_DEPRECATIONS} ./hack/build/run-unit-tests.sh ${WHAT}"
 
 test-functional: WHAT = ./tests/...
@@ -136,6 +136,20 @@ builder-push: ## Build and push the builder container image, declared in docker/
 
 openshift-ci-image-push: ## Build and push the OpenShift CI build+test container image, declared in hack/ci/Dockerfile.ci
 	./hack/build/osci-image-builder.sh
+
+rule-spec-dumper-executable := "rule-spec-dumper"
+
+build-prom-spec-dumper:
+	${DO} "go build -o ./hack/${rule-spec-dumper-executable} ./hack/prom-rule-ci/rule-spec-dumper.go"
+
+clean-prom-spec-dumper:
+	rm -f ./hack/${rule-spec-dumper-executable}
+
+prom-rules-verify: build-prom-spec-dumper
+	./hack/prom-rule-ci/verify-rules.sh \
+		"./hack/${rule-spec-dumper-executable}" \
+		"./hack/prom-rule-ci/prom-rules-tests.yaml"
+	rm ./hack/${rule-spec-dumper-executable}
 
 ##@ Local cluster management
 cluster-up: ## Start a default Kubernetes or Open Shift cluster. set KUBEVIRT_PROVIDER environment variable to either 'k8s-1.18' or 'os-3.11.0' to select the type of cluster. set KUBEVIRT_NUM_NODES to something higher than 1 to have more than one node.
--- a/hack/prom-rule-ci/BUILD.bazel
+++ b/hack/prom-rule-ci/BUILD.bazel
@@ -1,0 +1,15 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["rule-spec-dumper.go"],
+    importpath = "kubevirt.io/kubevirt/hack/prom-rule-ci",
+    visibility = ["//visibility:private"],
+    deps = ["//pkg/monitoring/rules:go_default_library"],
+)
+
+go_binary(
+    name = "prom-rule-ci",
+    embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+)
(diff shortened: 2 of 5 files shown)


hack/dockerized
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
Acedus: We use `DO` instead which is an alias for `hack/build/in-docker.sh`, there's no need to introduce this.

sradco:
@Acedus please review this pr

hack/dockerized
sradco: Done! Thanks

sradco:
@Acedus @akalenyu please review. I updated the pr with the fix.

automation/postsubmit-main.sh
```diff
@@ -9,6 +9,7 @@ export GIT_ASKPASS
 export DOCKER_TAG=""
 
 make apidocs
+make prom-rules-verify
```
akalenyu: I don't think we have to run this on postsubmit/release. Having it under the unit test make target (test-unit) should suffice since we strictly use a pull request flow

Makefile
```diff
@@ -137,6 +137,21 @@ builder-push: ## Build and push the builder container image, declared in docker/
...
+build-prom-spec-dumper:
+	${DO} "go build -o ${rule-spec-dumper-executable} ./hack/prom-rule-ci/rule-spec-dumper.go"
+
+clean-prom-spec-dumper:
+	rm -f ${rule-spec-dumper-executable}
+
+prom-rules-verify: build-prom-spec-dumper
+	./hack/prom-rule-ci/verify-rules.sh \
+		"${current-dir}/${rule-spec-dumper-executable}" \
+		"${current-dir}/hack/prom-rule-ci/prom-rules-tests.yaml"
```
akalenyu: I think we can align with other targets in this file and use `./hack/...`

akalenyu (commented):
Looks good though I think in the current state this does not run on PR presubmits?

automation/postsubmit-main.sh
sradco: This is how it is done also in the other repositories and its required

akalenyu: It sounds like an oversight to me; there's little value catching the violation post merge/release. I also see that kubevirt has a separate job `pull-kubevirt-prom-rules-verify` that runs on every PR, which is probably sufficient on it's own.

sradco: removed

Makefile
sradco: updated

sradco:
Hi @Acedus @akalenyu Thank you for the review. I updated the PR based on your comments.
Please review again.

automation/postsubmit-main.sh
akalenyu: > Having it under the unit test make target (test-unit) should suffice

Unless I'm missing something the current revision still doesn't run the added tests on the PR (only on release).

Makefile
```diff
@@ -137,6 +137,20 @@ builder-push: ## Build and push the builder container image, declared in docker/
...
+	${DO} "go build -o ./hack/${rule-spec-dumper-executable} ./hack/prom-rule-ci/rule-spec-dumper.go"
+
+clean-prom-spec-dumper:
+	rm -f ./hack/${rule-spec-dumper-executable}
+
+prom-rules-verify: build-prom-spec-dumper
+	./hack/prom-rule-ci/verify-rules.sh \
+		"./hack/${rule-spec-dumper-executable}" \
+		"./hack/prom-rule-ci/prom-rules-tests.yaml"
+	rm ./hack/${rule-spec-dumper-executable}
```
machadovilaca, line 152: can we find a way not to repeat all these scripts across repositories, maybe they can be in kubevirt/monitoring

sradco: Not sure, But even if we decide to its not in the scope of this effort

automation/postsubmit-main.sh
sradco: Can you please share where I should run this?

akalenyu: I'm fine with having it in the unit test target at first and possibly later it's own lane similar to kubevirt

sradco: Added it now to the test-unit.
I added it before the Go tests, since this this is very short and better to fail fast.

sradco:
@Acedus @akalenyu PR ready to be reviewed again. Thank you

avlitman:
/lgtm
looks good to me

Makefile
```diff
@@ -106,6 +106,7 @@ build-functest: ## Build the functional tests (content of tests/ subdirectory)
 test: test-unit test-functional test-lint ## execute all tests (_NOTE:_ 'WHAT' is expected to match the go cli pattern for paths e.g. './pkg/...'.  This differs slightly from rest of the 'make' targets)
 
 test-unit: WHAT = ./pkg/... ./cmd/...
+test-unit: prom-rules-verify
```
akalenyu: :+1:

automation/release.sh
```diff
@@ -32,6 +32,7 @@ function build_release_artifacts() {
     make apidocs
     make manifests
     make build-functest
+    make prom-rules-verify
```
akalenyu: I believe this is not needed (every PR runs the tests, not possible to release with bugs)

sradco: Updated

Makefile
```diff
@@ -106,6 +106,7 @@ build-functest: ## Build the functional tests (content of tests/ subdirectory)
...
 test-unit: ## Run unit tests.
 	${DO} "ACK_GINKGO_DEPRECATIONS=${ACK_GINKGO_DEPRECATIONS} ./hack/build/run-unit-tests.sh ${WHAT}"
```
machadovilaca:
```suggestion
test-unit: WHAT = ./pkg/... ./cmd/...
test-unit: prom-rules-verify ## Run unit tests.
	${DO} "ACK_GINKGO_DEPRECATIONS=${ACK_GINKGO_DEPRECATIONS} ./hack/build/run-unit-tests.sh ${WHAT}"

```

PR #3870, opened 2025-08-18
Bugfix: Reconcile pointer datasource on reference status update

Acedus:
Previously when a datasource reference status would change pointers to
said reference wouldn't reconcile to reflect the latest status.

This commit takes care to update them accordingly.

Fixes #

With this change we now reconcile the datasource keys and other datasources which reference it as source for either a spec change or a condition reason/message change specifically to avoid redundant requeues (i.e., comparing conditions directly would be problematic due to timestamps). `status.Source` can be evaluated instead.

**Release note**:

```release-note
bugfix: DataSource reference status updates are now correctly propagated to DataSources that have them as sources.
```

### DIFF
--- a/pkg/controller/datasource-controller.go
+++ b/pkg/controller/datasource-controller.go
@@ -392,6 +392,9 @@ func sameSourceSpec(objOld, objNew client.Object) bool {
 	if dsOld.Spec.Source.Snapshot != nil {
 		return reflect.DeepEqual(dsOld.Spec.Source.Snapshot, dsNew.Spec.Source.Snapshot)
 	}
+	if dsOld.Spec.Source.DataSource != nil {
+		return reflect.DeepEqual(dsOld.Spec.Source.DataSource, dsNew.Spec.Source.DataSource)
+	}
 
 	return false
 }
--- a/pkg/controller/datasource-controller.go
+++ b/pkg/controller/datasource-controller.go
@@ -237,20 +237,9 @@ func NewDataSourceController(mgr manager.Manager, log logr.Logger, installerLabe
 }
 
 func addDataSourceControllerWatches(mgr manager.Manager, c controller.Controller, log logr.Logger) error {
-	if err := c.Watch(source.Kind(mgr.GetCache(), &cdiv1.DataSource{}, &handler.TypedEnqueueRequestForObject[*cdiv1.DataSource]{},
-		predicate.TypedFuncs[*cdiv1.DataSource]{
-			CreateFunc: func(e event.TypedCreateEvent[*cdiv1.DataSource]) bool { return true },
-			DeleteFunc: func(e event.TypedDeleteEvent[*cdiv1.DataSource]) bool { return true },
-			UpdateFunc: func(e event.TypedUpdateEvent[*cdiv1.DataSource]) bool {
-				return !sameSourceSpec(e.ObjectOld, e.ObjectNew)
-			},
-		},
-	)); err != nil {
-		return err
-	}
-
 	const dataSourcePvcField = "spec.source.pvc"
 	const dataSourceSnapshotField = "spec.source.snapshot"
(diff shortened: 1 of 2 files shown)


Acedus:
That's a mouthful.

/cc @akalenyu @arnongilboa 

fyi @nunnatsa

pkg/controller/datasource-controller.go
```diff
@@ -367,6 +392,40 @@ func sameSourceSpec(objOld, objNew client.Object) bool {
...
+
+	for _, c := range newConditions {
+		if oldC, ok := condMap[c.Type]; !ok ||
+			oldC.Reason != c.Reason ||
+			oldC.Message != c.Message {
+			return false
+		}
+	}
+
+	return true
```
nunnatsa: I wonder if we can use the `slices.CompareFunc()` function from the standard library (assuming both conditions slices are sorted); e.g.
```go
return slices.CompareFunc(oldConditions, newConditions, func(oldC, newC cdiv1.DataSourceCondition) int{
	if oldC.Type == newC.Type &&
		oldC.Reason == newC.Reason && 
		oldC.Message == newC.Message {

		return 0
	}

	return -1
}) == 0
```

Acedus: I thought about that, but this is order sensitive. In theory, datasources shouldn't have more than a single condition, but I prefer to keep it this way.

pkg/controller/datasource-controller.go
```diff
@@ -237,20 +237,9 @@ func NewDataSourceController(mgr manager.Manager, log logr.Logger, installerLabe
 }
 
 func addDataSourceControllerWatches(mgr manager.Manager, c controller.Controller, log logr.Logger) error {
```
alromeros, line 239: This function is getting a bit messy and overly large, IMO would benefit from a small refactor (at least grouping watches and indexers together). Not a must for this PR, your changes look great.

alromeros (approved):
PR looks great! adding just a small comment about `addDataSourceControllerWatches` getting too complex, feel free to ignore.

pkg/controller/datasource-controller.go
Acedus: Yeah I definitely agree. I do have plans to do it as part of a follow-up PR.

pkg/controller/datasource-controller.go
```diff
@@ -367,6 +392,40 @@ func sameSourceSpec(objOld, objNew client.Object) bool {
...
+		condMap[c.Type] = c
+	}
+
+	for _, c := range newConditions {
+		if oldC, ok := condMap[c.Type]; !ok ||
+			oldC.Reason != c.Reason ||
+			oldC.Message != c.Message {
+			return false
+		}
+	}
```
akalenyu: Do you really want to queue over reason/message changes? I would assume you care more about the `cond.Status` changing for a certain `cond.Type`

Acedus: It might be seen as slightly redundant but the message also includes stuff like import progress which can be seen as valuable and something I'd like to reflect.

akalenyu: Don't you need cond.Status comparison here too?

Acedus: I imagine reason and message is granular enough already, but shouldn't harm to add it as well.

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/7c2f7d6bd53e0dc5d0b720b25a4f83656b85fcf6..4a47cc2e8605e2c89c1d814b49a9e460a5258d2d) - Address @akalenyu's comment.

pkg/controller/datasource-controller.go
akalenyu: I meant in a practical sense, that some updates won't get queued for processing
(especially important ones where false->true)

PR #3876, opened 2025-08-21
Add Readiness/Liveness Probes to Operator Pod

Dsanatar:

In order to comply with Red Hat operator best practices, all containers must have a Readiness and Liveness Probes defined for them. See here: https://github.com/redhat-best-practices-for-k8s/certsuite/blob/main/CATALOG.md#lifecycle-readiness-probe 

Fixes #

**Release note**:

```release-note
Readiness/Liveness Probes for Operator
```

### DIFF
--- a/cmd/cdi-operator/operator.go
+++ b/cmd/cdi-operator/operator.go
@@ -35,6 +35,7 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -98,6 +99,7 @@ func main() {
 		LeaderElectionNamespace:    namespace,
 		LeaderElectionID:           "cdi-operator-leader-election-helper",
 		LeaderElectionResourceLock: "leases",
+		HealthProbeBindAddress:     ":8444",
 		Metrics: server.Options{
 			BindAddress:   ":8443",
 			SecureServing: true,
@@ -156,6 +158,15 @@ func main() {
 	// Setup the controller
 	if err := controller.Add(mgr); err != nil {
 		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		log.Error(err, "unable to set up health check")
+		os.Exit(1)
+	}
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		log.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/operator/resources/operator/operator.go
```diff
@@ -433,6 +434,34 @@ func createOperatorDeployment(operatorVersion, namespace, deployClusterResources
...
+	}
+	container.ReadinessProbe = &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Scheme: corev1.URISchemeHTTPS,
+				Port: intstr.IntOrString{
+					Type:   intstr.Int,
+					IntVal: 8443,
+				},
+				Path: "/metrics",
```
alromeros: Looks good, just curious why choosing metrics here

dsanatar: decided to use `/metrics` only because I saw this endpoint was initialized and used in other places. However, after checking again there also is a preexisting `/healthz` endpoint that probably would be more suitable here.

cmd/cdi-operator/operator.go
```diff
@@ -98,6 +99,7 @@ func main() {
 		LeaderElectionNamespace:    namespace,
 		LeaderElectionID:           "cdi-operator-leader-election-helper",
 		LeaderElectionResourceLock: "leases",
+		HealthProbeBindAddress:     ":8444",
```
akalenyu, line 102: not a problem just curious, what's the issue with the default?

akalenyu (commented):
/approve

pkg/operator/resources/operator/operator.go
```diff
@@ -445,6 +474,11 @@ func createPrometheusPorts() []corev1.ContainerPort {
 			ContainerPort: 8443,
 			Protocol:      "TCP",
 		},
+		{
+			Name:          "health",
+			ContainerPort: 8444,
+			Protocol:      "TCP",
+		},
```
akalenyu, line 481: just noticed this is under createPrometheusPorts, which is no longer true

akalenyu:
/hold
chance to address comment

cmd/cdi-operator/operator.go
dsanatar: I don't believe there is a default bind address for the health probe. When the controller manager attempts to create the listener, it first checks to see that this bind address option is set.

akalenyu: ah okay my bad. newly created operator-sdk projects set this to 8081 on the cli flag level, which apparently is problematic https://github.com/openshift/dpu-operator/commit/acc0afd41ea524790a214114fd7bec922e16a304

pkg/operator/resources/operator/operator.go
akalenyu: hmm are you sure you want to eliminate this?

dsanatar: ah sorry I must have gotten confused, I thought you meant the port shouldn't be exposed here. Even without it _explicitly_ defined here, I can still reach the new /heathz and /readyz endpoints from within the container, I believe this is because of the way the controller manager sets up the health server. Can you explain what you meant by your previous comment? It's probably better that I revert back to the original state where we have the port explicitly exposed, but I wasn't sure what you meant about the createPrometheusPorts not longer being true. Is it the fact that the function name is referring to Prometheus but is no relevant anymore?

akalenyu: ah I see, I only meant the name for this function became confusing once you added a non prometheus port (createPrometheusPorts)

dsanatar: ahh ok gotcha, i'll address that. thanks

PR #3890, opened 2025-09-01
Allow mutating only PVC requested size to the minimum supported

arnongilboa:
Allow mutating only the PVC requested size to the minimum supported by the relevant `StorageProfile`, but without applying the `StorageProfile` `accessModes`, `volumeMode`, or `filesystemOverhead`.

**Release note**:
```release-note
Allow mutating only the PVC requested size to the minimum supported
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -67,6 +67,8 @@ const (
 
 	// PvcApplyStorageProfileLabel tells whether the PVC should be rendered by the mutating webhook based on StorageProfiles
 	PvcApplyStorageProfileLabel = CDIComponentLabel + "/applyStorageProfile"
+	// PvcApplyStorageProfileMinimumSize tells whether the PVC mutating webhook rendering should apply only the StorageProfile minimum volume size if exists
+	PvcApplyStorageProfileMinimumSize = "minimumSize"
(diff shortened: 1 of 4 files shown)


pkg/common/common.go
```diff
@@ -67,6 +67,8 @@ const (
 
 	// PvcApplyStorageProfileLabel tells whether the PVC should be rendered by the mutating webhook based on StorageProfiles
 	PvcApplyStorageProfileLabel = CDIComponentLabel + "/applyStorageProfile"
+	// PvcApplyStorageProfileMinimumSize tells whether the PVC mutating webhook rendering should apply only the StorageProfile minimum volume size if exists
+	PvcApplyStorageProfileMinimumSize = "minimumSize"
```
alromeros, line 71: Do we plan to add more `applyStorageProfile` modes? (accesmode, volume mode, etc)

alromeros: Even if it's just this one we should update the repository's docs with info about the possible options.

arnongilboa: This PR is just an option for consideration. I'm not sure it's the best one, and that's why it's on hold :)

Issue #3900, opened 2025-09-10
Host assisted clone should adjust requested storage when cloning from block to filesystem volume mode


gaohoward:
**What happened**:
When cloning from a source, the host cloner should check the requested size for
the target tmp pvc if the tmp source pvc is of **block** volume mode and the
target pvc is of **filesystem** volume mode because it should take filesystem
overhead into consideration and enlarge the original request size accordingly
before creating the tmp pvc.

Because currently the check is missing the cloning will fail on validation.

**What you expected to happen**:
The cloning should be successful.

**How to reproduce it (as minimally and precisely as possible)**:
Steps to reproduce the behavior.

1. create a volume snapshot and make sure the VolumeSnapshotContent spec.sourceVolumeMode is Block.
2. create a DV that clone from the snapshot, for example 

apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  annotations:
    cdi.kubevirt.io/storage.bind.immediate.requested: "true"
  name: cloned-datavolume
  namespace: default
spec:
  source:
    snapshot:
      namespace: golden-ns
      name: golden-volumesnapshot
  storage:
    storageClassName: hostpath-csi
    resources: {}

3. Observe that the clone never finishes.

**Additional context**:Add any other context about the problem here.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): N/A
- Kubernetes version (use `kubectl version`): N/A
- DV specification: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

PR #3901, opened 2025-09-10
Host assisted clone should adjust requested storage when cloning from block to filesystem volume mode (#3900)

gaohoward:
When cloning from a source, the host cloner should check the requested size for the target tmp pvc if the tmp source pvc is of block volume mode and the target pvc is of filesystem volume mode because it should take filesystem overhead into consideration and enlarge the original request size accordingly
before creating the tmp pvc.

Because currently the check is missing the cloning will fail on validation.

Fixes #

https://issues.redhat.com/browse/CNV-44140

**Release note**:

```release-note
Fix: Host assisted clone should adjust requested storage when cloning from block to filesystem volume mode
```

### DIFF
--- a/pkg/controller/clone/host-clone.go
+++ b/pkg/controller/clone/host-clone.go
@@ -201,12 +201,73 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
 	}
 	cc.AddLabel(claim, cc.LabelExcludeFromVeleroBackup, "true")
 
+	if err := p.MakeSureTargetPVCHasSufficientSpace(ctx, claim); err != nil {
+		return nil, err
+	}
+
 	if err := p.Client.Create(ctx, claim); err != nil {
 		checkQuotaExceeded(p.Recorder, p.Owner, err)
 		return nil, err
 	}
 
 	return claim, nil
+}
+
+// The purpose of this check is to make sure the target PVC to have sufficient space
+// before the cloning actually happens.
+func (p *HostClonePhase) MakeSureTargetPVCHasSufficientSpace(ctx context.Context, targetPvc *corev1.PersistentVolumeClaim) error {
+	sourcePvc := &corev1.PersistentVolumeClaim{}
+	sourcePvcKey := client.ObjectKey{Namespace: p.Namespace, Name: p.SourceName}
+
+	if err := p.Client.Get(ctx, sourcePvcKey, sourcePvc); err != nil {
+		return err
+	}
+
+	if targetPvc.Spec.Resources.Requests == nil {
+		return fmt.Errorf("no target resource request specified")
+	}
+	targetSize, ok := targetPvc.Spec.Resources.Requests[corev1.ResourceStorage]
+
+	if !ok {
+		return fmt.Errorf("no target size specified")
+	}
+
+	unInflatedSourceSize := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
+	sourceVolumeMode := cc.GetVolumeMode(sourcePvc)
+	targetVolumeMode := cc.GetVolumeMode(targetPvc)
+	var err error
+
+	// For filesystem source PVCs, we need to get the original DV size to account for
+	if sourceVolumeMode == corev1.PersistentVolumeFilesystem {
+		original, err := cc.GetHostCloneOriginalSourceDVSize(ctx, p.Client, sourcePvc)
+		if err != nil {
+			return err
+		}
+		if !original.IsZero() {
+			unInflatedSourceSize = original
+		}
+	}
+
+	targetSizeUpdated := false
+	if unInflatedSourceSize.Cmp(targetSize) >= 0 {
+		targetSizeUpdated = true
+		targetSize = unInflatedSourceSize
+	}
+
+	// if target is filesystem, inflate the original size if target
+	if targetVolumeMode == corev1.PersistentVolumeFilesystem && targetSizeUpdated {
+		// when only sourc pvc size is available (snapshot case) we have no
+		// way to trace back to its original DV size (because the DV probably
+		// doesn't exist anymore), so we use source pvc size to just inflate
+		// the target size with overhead
+		targetSize, err = cc.InflateSizeWithOverhead(ctx, p.Client, unInflatedSourceSize.Value(), &targetPvc.Spec)
+		if err != nil {
+			return err
+		}
+	}
+
+	targetPvc.Spec.Resources.Requests[corev1.ResourceStorage] = targetSize
+	return nil
 }
 
 func (p *HostClonePhase) hostCloneComplete(pvc *corev1.PersistentVolumeClaim) bool {
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -2325,3 +2325,58 @@ func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder recor
 		recorder.Event(targetPVC, newEvent.Type, newEvent.Reason, formattedMsg)
 	}
 }
+
+func GetPVCDataSourceKind(pvc *corev1.PersistentVolumeClaim) string {
+	pvcDataSource := pvc.Spec.DataSource
+	pvcDataSourceRef := pvc.Spec.DataSourceRef
+	if pvcDataSourceRef != nil {
+		return pvcDataSourceRef.Kind
+	}
+	if pvcDataSource != nil {
+		return pvcDataSource.Kind
+	}
+	return ""
+}
+
+// In host clone the source PVC comes from 2 sources
+// 1. snaphot clone - in this case there is no DV associated with the sourcePVC
+// 2. direct PVC clone - in this case there may be a DV associated with the sourcePVC
+func GetDVFromHostCloneSourcePVC(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (*cdiv1.DataVolume, error) {
+	pvcDataSourceKind := GetPVCDataSourceKind(pvc)
+
+	if pvcDataSourceKind != "PersistentVolumeClaim" {
+		return nil, nil
+	}
+
+	dvName := ""
+	for _, ownerRef := range pvc.GetOwnerReferences() {
+		if ownerRef.Kind == "DataVolume" && ownerRef.APIVersion == "cdi.kubevirt.io/v1beta1" {
+			dvName = ownerRef.Name
+			break
+		}
+	}
+	if dvName != "" {
+		dv := &cdiv1.DataVolume{}
+		if err := c.Get(ctx, types.NamespacedName{Namespace: pvc.Namespace, Name: dvName}, dv); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return nil, nil
+			}
+			return nil, err
+		}
+		return dv, nil
+	}
+	return nil, nil
+}
+
+func GetHostCloneOriginalSourceDVSize(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (resource.Quantity, error) {
+	// Get the datavolume associate with the source
+	dv, err := GetDVFromHostCloneSourcePVC(ctx, c, pvc)
+	if err != nil {
+		return resource.Quantity{}, err
+	}
+	size := resource.Quantity{}
+	if dv != nil {
+		size = dv.Spec.PVC.Resources.Requests[corev1.ResourceStorage]
+	}
+	return size, nil
+}
--- a/pkg/controller/clone/BUILD.bazel
+++ b/pkg/controller/clone/BUILD.bazel
@@ -55,6 +55,7 @@ go_test(
     ],
     embed = [":go_default_library"],
     deps = [
+        "//pkg/common:go_default_library",
         "//pkg/controller/common:go_default_library",
         "//staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1:go_default_library",
         "//vendor/github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1:go_default_library",
--- a/pkg/controller/clone/host-clone_test.go
+++ b/pkg/controller/clone/host-clone_test.go
@@ -23,6 +23,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -33,17 +34,36 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	"kubevirt.io/containerized-data-importer/pkg/common"
 	cc "kubevirt.io/containerized-data-importer/pkg/controller/common"
 )
 
 var _ = Describe("HostClonePhase test", func() {
 	log := logf.Log.WithName("host-clone-phase-test")
 
-	creatHostClonePhase := func(objects ...runtime.Object) *HostClonePhase {
+	type ResourceModifier struct {
+		modifySourcePvc  func(pvcSpec *corev1.PersistentVolumeClaim)
+		modifyDesiredPvc func(pvc *corev1.PersistentVolumeClaim)
+	}
+
+	creatHostClonePhase := func(modifier *ResourceModifier, objects ...runtime.Object) *HostClonePhase {
 		s := scheme.Scheme
 		_ = cdiv1.AddToScheme(s)
 
 		objects = append(objects, cc.MakeEmptyCDICR())
+
+		source := &corev1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "ns",
+				Name:      "source",
+			},
+		}
+
+		if modifier != nil && modifier.modifySourcePvc != nil {
+			modifier.modifySourcePvc(source)
+		}
+
+		objects = append(objects, source)
 
 		// Create a fake client to mock API calls.
 		builder := fake.NewClientBuilder().
@@ -67,13 +87,24 @@ var _ = Describe("HostClonePhase test", func() {
 				Namespace: "ns",
 				Name:      "desired",
 			},
+			Spec: corev1.PersistentVolumeClaimSpec{
+				Resources: corev1.VolumeResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceStorage: resource.MustParse("1Gi"),
+					},
+				},
+			},
+		}
+
+		if modifier != nil && modifier.modifyDesiredPvc != nil {
+			modifier.modifyDesiredPvc(desired)
 		}
 
 		return &HostClonePhase{
 			Owner:          owner,
 			OwnershipLabel: "label",
 			Namespace:      "ns",
-			SourceName:     "source",
+			SourceName:     source.Name,
 			DesiredClaim:   desired,
 			ImmediateBind:  true,
 			Preallocation:  false,
@@ -91,7 +122,7 @@ var _ = Describe("HostClonePhase test", func() {
 	}
 
 	It("should create pvc", func() {
-		p := creatHostClonePhase()
+		p := creatHostClonePhase(nil)
 
 		result, err := p.Reconcile(context.Background())
 		Expect(err).ToNot(HaveOccurred())
@@ -114,7 +145,7 @@ var _ = Describe("HostClonePhase test", func() {
 	})
 
 	It("should create pvc with priorityclass", func() {
-		p := creatHostClonePhase()
+		p := creatHostClonePhase(nil)
 		p.PriorityClassName = "priority"
 
 		result, err := p.Reconcile(context.Background())
@@ -125,6 +156,100 @@ var _ = Describe("HostClonePhase test", func() {
 
 		pvc := getDesiredClaim(p)
 		Expect(pvc.Annotations[cc.AnnPriorityClassName]).To(Equal("priority"))
+	})
+
+	It("should get error if target pvc does not have size", func() {
+		makePvcSizeMissing := func(pvc *corev1.PersistentVolumeClaim, _ corev1.PersistentVolumeMode) {
+			pvc.Spec = corev1.PersistentVolumeClaimSpec{}
+		}
+		cdiConfig := cc.MakeEmptyCDIConfigSpec(common.ConfigName)
+		cdiConfig.Status.FilesystemOverhead = &cdiv1.FilesystemOverhead{
+			Global: common.DefaultGlobalOverhead,
+		}
+
+		p := creatHostClonePhase(&ResourceModifier{
+			modifyDesiredPvc: func(pvc *corev1.PersistentVolumeClaim) {
+				makePvcSizeMissing(pvc, corev1.PersistentVolumeFilesystem)
+			},
+		}, cdiConfig)
+
+		_, err := p.Reconcile(context.Background())
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("no target resource request specified"))
+	})
+
+	It("should adjust requested size for block to filesystem volume mode clone", func() {
+		setPvcAttributes := func(pvcSpec *corev1.PersistentVolumeClaimSpec, volumeMode corev1.PersistentVolumeMode, storage string) {
+			pvcSpec.VolumeMode = &volumeMode
+			if pvcSpec.Resources.Requests == nil {
+				pvcSpec.Resources.Requests = corev1.ResourceList{}
+			}
+			pvcSpec.Resources.Requests[corev1.ResourceStorage] = resource.MustParse(storage)
+		}
+		cdiConfig := cc.MakeEmptyCDIConfigSpec(common.ConfigName)
+		cdiConfig.Status.FilesystemOverhead = &cdiv1.FilesystemOverhead{
+			Global: common.DefaultGlobalOverhead,
+		}
+
+		p := creatHostClonePhase(&ResourceModifier{
+			modifySourcePvc: func(pvc *corev1.PersistentVolumeClaim) {
+				setPvcAttributes(&pvc.Spec, corev1.PersistentVolumeBlock, "8Gi")
+			},
+			modifyDesiredPvc: func(pvc *corev1.PersistentVolumeClaim) {
+				setPvcAttributes(&pvc.Spec, corev1.PersistentVolumeFilesystem, "8Gi")
+			},
+		}, cdiConfig)
+
+		result, err := p.Reconcile(context.Background())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(result).ToNot(BeNil())
+		Expect(result.Requeue).To(BeFalse())
+		Expect(result.RequeueAfter).ToNot(BeZero())
+
+		pvc := getDesiredClaim(p)
+
+		Expect(*pvc.Spec.VolumeMode).To(Equal(corev1.PersistentVolumeFilesystem))
+		actualSize := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+		originalRequested := resource.MustParse("8Gi")
+
+		Expect(actualSize.Cmp(originalRequested)).To(BeNumerically(">", 0), "The actual should be greater than the requested", "actual", actualSize, "requested", originalRequested)
+	})
+
+	It("should inflate the target size for filesystem to filesystem volume mode clone", func() {
+		setPvcAttributes := func(pvcSpec *corev1.PersistentVolumeClaimSpec, volumeMode corev1.PersistentVolumeMode, storage string) {
+			pvcSpec.VolumeMode = &volumeMode
+			if pvcSpec.Resources.Requests == nil {
+				pvcSpec.Resources.Requests = corev1.ResourceList{}
+			}
+			pvcSpec.Resources.Requests[corev1.ResourceStorage] = resource.MustParse(storage)
+		}
+		cdiConfig := cc.MakeEmptyCDIConfigSpec(common.ConfigName)
+		cdiConfig.Status.FilesystemOverhead = &cdiv1.FilesystemOverhead{
+			Global: common.DefaultGlobalOverhead,
+		}
+
+		p := creatHostClonePhase(&ResourceModifier{
+			modifySourcePvc: func(pvc *corev1.PersistentVolumeClaim) {
+				setPvcAttributes(&pvc.Spec, corev1.PersistentVolumeFilesystem, "8Gi")
+			},
+			modifyDesiredPvc: func(pvc *corev1.PersistentVolumeClaim) {
+				setPvcAttributes(&pvc.Spec, corev1.PersistentVolumeFilesystem, "8Gi")
+			},
+		}, cdiConfig)
+
+		result, err := p.Reconcile(context.Background())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(result).ToNot(BeNil())
+		Expect(result.Requeue).To(BeFalse())
+		Expect(result.RequeueAfter).ToNot(BeZero())
+
+		pvc := getDesiredClaim(p)
+
+		Expect(*pvc.Spec.VolumeMode).To(Equal(corev1.PersistentVolumeFilesystem))
+		actualSize := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+		originalRequested := resource.MustParse("8Gi")
+
+		Expect(actualSize.Cmp(originalRequested)).To(BeNumerically(">", 0), "The actual should be greater than the requested", "actual", actualSize, "requested", originalRequested)
 	})
 
 	Context("with desired claim created", func() {
@@ -141,7 +266,7 @@ var _ = Describe("HostClonePhase test", func() {
 		It("should wait for clone to succeed", func() {
 			desired := getCliam()
 			desired.Annotations[cc.AnnPodPhase] = "Running"
-			p := creatHostClonePhase(desired)
+			p := creatHostClonePhase(nil, desired)
 
 			result, err := p.Reconcile(context.Background())
 			Expect(err).ToNot(HaveOccurred())
@@ -153,7 +278,7 @@ var _ = Describe("HostClonePhase test", func() {
 		It("should wait for clone to succeed with preallocation", func() {
 			desired := getCliam()
 			desired.Annotations[cc.AnnPodPhase] = "Succeeded"
-			p := creatHostClonePhase(desired)
+			p := creatHostClonePhase(nil, desired)
 			p.Preallocation = true
 
 			result, err := p.Reconcile(context.Background())
@@ -166,7 +291,7 @@ var _ = Describe("HostClonePhase test", func() {
 		It("should succeed", func() {
 			desired := getCliam()
 			desired.Annotations[cc.AnnPodPhase] = "Succeeded"
-			p := creatHostClonePhase(desired)
+			p := creatHostClonePhase(nil, desired)
 
 			result, err := p.Reconcile(context.Background())
 			Expect(err).ToNot(HaveOccurred())
@@ -177,7 +302,7 @@ var _ = Describe("HostClonePhase test", func() {
 			desired := getCliam()
 			desired.Annotations[cc.AnnPodPhase] = "Succeeded"
 			desired.Annotations[cc.AnnPreallocationApplied] = "true"
-			p := creatHostClonePhase(desired)
+			p := creatHostClonePhase(nil, desired)
 			p.Preallocation = true
 
 			result, err := p.Reconcile(context.Background())


awels:
I am probably missing something, but didn't https://github.com/kubevirt/containerized-data-importer/pull/3384 fix this particular issue? Or did I miss a particular case with that PR?

gaohoward:
> I am probably missing something, but didn't #3384 fix this particular issue? Or did I miss a particular case with that PR?

Yeah I think it goes a different path. In this case it falls back to host-clone from smart-clone (supported by provisioner), and in that case it didn't take the fs overhead into account.

awels:
Gotcha, okay, let me know if you need a review on the PR.

gaohoward:
> Gotcha, okay, let me know if you need a review on the PR.

Yes pleas review. Thanks @awels !

pkg/controller/clone/host-clone.go
```diff
@@ -195,6 +195,31 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+		// It is possible when the source pvc has VolumMode 'block'
+		// and the claim has 'filesystem' in which case the filesystem overhead need to be considered
+		sourcePvc := &corev1.PersistentVolumeClaim{}
+		sourcePvcKey := client.ObjectKey{Namespace: p.Namespace, Name: p.SourceName}
+
+		if err := p.Client.Get(ctx, sourcePvcKey, sourcePvc); err != nil {
+			return nil, err
+		}
+
+		realSourcePvcSizeRequest := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
```
awels: This is the only thing I have some concern about, this works fine if the source volume mode is block, since we can just inflate the size of the source PVC to account for the overhead. But if the source volume mode is filesystem, it is likely the source is already inflated, and we don't want to inflate it again. I think in that case we should check if there is a datavolume associated with the PVC, and get the size from there.

gaohoward: Thanks @awel. I think I see the point. I'll try correct it.

gaohoward: @awels  I've made some changes to improve it. Pls take a look. Thanks!

pkg/controller/clone/host-clone.go
```diff
@@ -195,6 +195,41 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+		if err := p.Client.Get(ctx, sourcePvcKey, sourcePvc); err != nil {
+			return nil, err
+		}
+
+		if claim.Spec.Resources.Requests != nil {
+			realSourcePvcSizeRequest := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
+
+			if sourceVolumeMode := cc.GetVolumeMode(sourcePvc); sourceVolumeMode == corev1.PersistentVolumeFilesystem {
+				// If the source PVC is filesystem, just directly compare
+				if realSourcePvcSizeRequest.Cmp(claim.Spec.Resources.Requests[corev1.ResourceStorage]) > 0 {
```
awels: I would implement the logic something like this pseudo code, in particular look up the DV size for a filesystem source. This is because the source overhead might not equal the target overhead (you can have different overheads for different storage classes). This means it is better to look up the size of the DV (if there) and use that and then do the inflation.

```go
size := realSourcePvcSizeRequest 
inflate := true
if sourceVolume == filesystem {
  // check if there is a datavolume associated with the source, and get the size from the datavolume.
  dv := getDataVolume(sourcePVC) // There should be functions that do this already.
  if dv != nil {
    size = getDvSize(dv) // There should be functions that do this already.
  } else {
    // Can't determine if any overhead was used for source, so assume it is correct and
    // don't inflate
    inflate = false
  }
}
if inflate {
  newSize = InflateSizeWithOverhead(size)
}
claim.Spec.Resources.Requests[corev1.ResourceStorage] = newSize
```

gaohoward: @awels that makes sense. I'll update the code. Thanks!

gaohoward:
@awels Hi Alexander, I finally get the tests pass, can you please take a look again?

Thanks

pkg/controller/common/util.go
```diff
@@ -2224,3 +2224,57 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
...
+	dv := &cdiv1.DataVolume{}
+	if err := c.Get(ctx, types.NamespacedName{Namespace: pvc.Namespace, Name: pvc.Name}, dv); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+	}
+	return dv, nil
+}
+
+func GetDVCloneSize(ctx context.Context, c client.Client, dv *cdiv1.DataVolume) (*resource.Quantity, error) {
```
awels: I was thinking maybe you can re-use the function we have here https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume/pvc-clone-controller.go#L533-L606 as I don't understand what you are doing here.

pkg/controller/clone/host-clone.go
```diff
@@ -195,6 +195,60 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+					} else {
+						// can't determine the overhead, assuming size is correct
+						inflate = false
+					}
+				} else {
+					inflate = false
+				}
+			} else {
+				// If the source PVC is block, we need to account for the overhead
+				if usableSpace, err := cc.GetUsableSpace(ctx, p.Client, claim); err == nil {
```
awels: I would simplify this to:
```go
inflate = false
if usableSpace, err := cc.GetUsableSpace(ctx, p.Client, claim); err != nil {
  return nil, err
}
```
I don't understand the extra check for if the usableSpace against the requested size. Also you are eating any errors from `GetUsableSpace` right now.

pkg/controller/clone/host-clone.go
```diff
@@ -195,6 +195,60 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+		if err := p.Client.Get(ctx, sourcePvcKey, sourcePvc); err != nil {
+			return nil, err
+		}
+
+		if claim.Spec.Resources.Requests != nil {
+			size := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
+			inflate := true
+			if sourceVolumeMode := cc.GetVolumeMode(sourcePvc); sourceVolumeMode == corev1.PersistentVolumeFilesystem {
+				// Get the datavolume associate with the source
+				if dv, err := cc.GetDVFromPVC(ctx, p.Client, sourcePvc); err == nil {
```
awels: The canonical way to do these types of comparisons is the first one checks for the error so something like this
```go
if dv, err := cc.GetDVFromPVC(ctx, p.Client, sourcePvc); err != nil {
  return nil, err
} else if dv != nil {
  if sourceSize, err := cc.GetDVCloneSize(ctx, p.Client, dv); err == nil {
    // If the source PVC is filesystem, just directly compare
    targetSize := claim.Spec.Resources.Requests[corev1.ResourceStorage]
    if targetSize.Cmp(*sourceSize) >= 0 {
      // the target size has enough space, not to inflate
      inflate = false
    }
  }
}
```
Right now you are not returning an error if `GetDVFromPVC` returns an error.

pkg/controller/clone/host-clone.go
```diff
@@ -195,6 +195,60 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
 	}
 	cc.AddLabel(claim, cc.LabelExcludeFromVeleroBackup, "true")
 
+	if targetVolumeMode := cc.GetVolumeMode(claim); targetVolumeMode == corev1.PersistentVolumeFilesystem {
```
awels: I am not sure I follow this logic. Here we have determined that the target is a filesystem volume. We have the following combinations we need to take into consideration:
1. If source is block, take source size and inflate
2. If source if filesystem, we need to figure out what the original source is as the current size might be inflated already.
3. To get the original size, check if there is a DV associated with the source PVC. We need to check for 2 things
  a. Does the PVC have an ownerRef with Kind: DataVolume and APIVersion v1beta
  b. Does the DV exist.
4. If you find the DV, take the size of the DV and use that and inflate
5. If you don't find the DV, then take the size of the PVC and use that. You don't need to inflate.

pkg/controller/clone/host-clone.go
gaohoward: right, the error is ignored. The check is to make sure the target volume has enough size for the source.
will update. Thanks!

pkg/controller/clone/host-clone.go
gaohoward: yeah, the error is ignored here again. Will correct. Thx!

pkg/controller/common/util.go
gaohoward: This method just gets the source size of the pvc/snapshot. The code you mentioned actually deals with the sizing problem in pvc clone case only. The issue happens when the clone goes through snapshot clone controller.

pkg/controller/clone/host-clone.go
gaohoward: @awels  I think my code is a bit confusing. The issue we want to fix is on the target pvc in host-assisted clone.
Here we decide whether or not to inflate the target PVC as follows:

1. If the source is block and target is filesystem, we compare the sizes of them and inflate if needed (the original bug happened in this case, i.e. clone from a snapshot source, not a pvc source which works fine)
2. If the source is filesystem too, we compare them and and inflate if needed (this never happened as far as I can see, but we check anyway for sure)
3. if the target is block (whether the source is filesystem or block), we don't need to worry as the controllers handle them well. 

So No.1 is the actual fix. No.2 is just make sure (not likely to happen).

Does that makes sense?

awels: So no.2 is possible and fairly likely to happen. There are some permutations here (which I tried to explain in my original question).
- The source filesystem has the same inflation as the target. This is the easiest scenario, you can just pick the same as the source. But you have to verify the inflation of the source.
- The source filesystem has a different inflation as the target. To calculate the size you need, you first need to find the original un-inflated size (from the DV). And then inflate that size.

For both scenarios it is simplest to find the source DV (if it exists). And get the size from there. Then do the normal inflation calculation against that size. To me the logic should look something like this:
1. Figure out the un-inflated source size
2. If the target is filesystem, inflate it because we know at this point the source is the un-inflated size.

```go
// If block this is correct
uninflatedSourceSize := sourcePvc.Spec.Resources.Requests[corev1.ResourceStorage]
if sourceVolumeMode := cc.GetVolumeMode(sourcePvc); sourceVolumeMode == corev1.PersistentVolumeFilesystem {
  dv, err := cc.GetDVFromPVC(ctx, p.Client, sourcePvc)
  if err != nil {
    return nil, err
  }
  if dv != nil {
    // Found the DV, set the uninflated size to its size.
    uninflatedSourceSize, err = cc.GetDVCloneSize(ctx, p.Client, dv)
  }
}
targetSize := uninflatedSourceSize
// Here we know the uninflatedSize is always correct regardless of source volume mode
if targetVolumeMode := cc.GetVolumeMode(targetPvc); targetVolumeMode == corev1.PersistentVolumeFilesystem {
  targetSize, err := cc.InflateSizeWithOverhead(ctx, p.Client, uninflatedSourceSize.Value(), &claim.Spec)
}
// Return the targetSize
```

gaohoward: Hi @awels , I updated the PR and now it's ok for review again. Please take a look. 
I've move the code block into a new function to improve the readability. I adjust the logic as you suggested here. Thanks!

pkg/controller/datavolume/pvc-clone-controller.go
```diff
@@ -548,6 +548,9 @@ func (r *PvcCloneReconciler) detectCloneSize(syncState *dvSyncState) (bool, erro
 		return true, nil
 	}
 
+	// size required
```
awels: left over? Not sure what this means.

pkg/controller/common/util.go
```diff
@@ -2319,3 +2319,74 @@ func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder recor
...
+	if pvcDataSource != nil {
+		return pvcDataSource.Kind
+	}
+	return ""
+}
+
+// In host clone the source PVC comes from 2 sources
+// 1. snaphot clone - in this case there is no DV associated with the sourcePVC
+// 2. direct PVC clone - in this case there may be a DV associated with the sourcePVC
+func GetDVFromHostCloneSourcePVC(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) *cdiv1.DataVolume {
```
awels: Probably want to return (*cdiv1.DataVolume, error) since it is possible the call to List can fail and we want to propagate that.

pkg/controller/common/util.go
```diff
@@ -2319,3 +2319,74 @@ func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder recor
...
+	pvcDataSourceKind := GetPVCDataSourceKind(pvc)
+
+	if pvcDataSourceKind != "PersistentVolumeClaim" {
+		return nil
+	}
+
+	// if the pvc is from a DV, it will have the following annotation
+	// cdi.kubevirt.io/createdForDataVolume
+	// and the value is the DV's UID
+	createdForDVUID, ok := pvc.Annotations[AnnCreatedForDataVolume]
```
awels: I would look at the ownerRefs, the owner references will have the name of the owning DataVolume. Then instead of listing all the DVs and looping over them, you can just do a Get to get the DV. So something like this:

```go
dvName := ""
for _, ownerRef := range pvc.GetOwnerReferences() {
  if ownerRef.Kind == "DataVolume" && ownerRef.APIVersion == "cdi/v1beta1" {
    dvName := ownerRef.Name
    break
  }
}
if name != "" {
  dv := &cdiv1.DataVolume{}
  if err := c.Get(ctx, dv, types.NamespacedName{Namespace: pvc.namespace, Name: dvName}); err != nil {
	return nil, err
  }
  return dv, nil
}
return nil, nil
```

pkg/controller/common/util.go
```diff
@@ -2319,3 +2319,74 @@ func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder recor
...
+		if string(dv.UID) == createdForDVUID {
+			return &dv
+		}
+	}
+	return nil
+}
+
+func GetHostCloneOriginalSourceDVSize(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) resource.Quantity {
+	// Get the datavolume associate with the source
+	dv := GetDVFromHostCloneSourcePVC(ctx, c, pvc)
```
awels: If we return an error from this function, will need to handle it or return it as well.

pkg/controller/clone/host-clone.go
```diff
@@ -209,6 +214,67 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+// The purpose of this check is to make sure the target PVC to have sufficient space
+// before the cloning actually happens.
+func (p *HostClonePhase) MakeSureTargetPVCHasSufficientSpace(ctx context.Context, targetPvc *corev1.PersistentVolumeClaim) error {
+	sourcePvc := &corev1.PersistentVolumeClaim{}
+	sourcePvcKey := client.ObjectKey{Namespace: p.Namespace, Name: p.SourceName}
+
+	if err := p.Client.Get(ctx, sourcePvcKey, sourcePvc); err != nil {
+		return err
+	}
+	if targetPvc.Spec.Resources.Requests != nil {
```
awels: If there is no target PVC size, that is probably a problem and we should return an error. You can do something like this
```go
if targetPvc.Spec.Resources.Requests == nil {
  return fmt.Errorf("no target resource request specified")
}
targetSize, ok := targetPvc.Spec.Resources.Requests[corev1.ResourceStorage]
if !ok {
  return fmt.Errorf("no target size specified")
}
```
Then you can remove the reverse if statement and save one indentation.

pkg/controller/clone/host-clone.go
```diff
@@ -209,6 +214,67 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+
+		targetSize := targetPvc.Spec.Resources.Requests[corev1.ResourceStorage]
+
+		if unInflatedSourceSize.Cmp(targetSize) > 0 {
+			targetSize = unInflatedSourceSize
+			// if both target and source have same volume mode, take a original size if target is smaller
+			// no need to inflate
+
+			// if filesystem to block, inflate original size if block size is smaller
+			if sourceVolumeMode == corev1.PersistentVolumeFilesystem && targetVolumeMode == corev1.PersistentVolumeBlock {
```
awels: I am not sure I understand this check. If the target volume is block then we don't have to worry about inflating it with the filesystem overhead since there is no filesystem on it. You can just set the target directly to the un-inflated source and return.

pkg/controller/clone/host-clone.go
```diff
@@ -209,6 +214,67 @@ func (p *HostClonePhase) createClaim(ctx context.Context) (*corev1.PersistentVol
...
+				inflate = true
+			}
+		}
+
+		// if block to filesystem, inflate the original size if target is smaller
+		// need to take the filesystem overhead into account when comparing sizes
+		if sourceVolumeMode == corev1.PersistentVolumeBlock && targetVolumeMode == corev1.PersistentVolumeFilesystem {
```
awels: At this point the uninflated size is one of two options:
1. The source is block, and thus it is just the source size.
2. The source is filesystem and we looked up the original size of the DV.

Either way, the source is now correct, we only need to inflate it if the target is filesystem. So no need for `sourceVolumeMode == corev1.PersistentVolumeBlock`

pkg/controller/common/util.go
```diff
@@ -2319,3 +2319,74 @@ func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder recor
 		recorder.Event(targetPVC, newEvent.Type, newEvent.Reason, formattedMsg)
 	}
 }
+
+func GetUsableSpace(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) (resource.Quantity, error) {
```
awels: This seems to be a copy of `InflateSizeWithOverhead` any reason this is needed?

pkg/controller/datavolume/pvc-clone-controller.go
gaohoward: yeah not sure when I put them there. I'll removed them. :)

pkg/controller/common/util.go
gaohoward: +1

pkg/controller/common/util.go
gaohoward: ok.

pkg/controller/common/util.go
gaohoward: +1

pkg/controller/clone/host-clone.go
gaohoward: I remember I added this check because some tests make the target resource requests to be nil and expect it to be handled somewhere else. Maybe now its no longer valid as we have changed a lot logic here. I'll try remove the check and see if it breaks anything.

pkg/controller/clone/host-clone.go
gaohoward: yeah it makes sense. +1

pkg/controller/clone/host-clone.go
gaohoward: +1

pkg/controller/common/util.go
gaohoward: It is used to calculate the usable size of a volume. Now it seems not necessary. I'll remove it.

akalenyu (commented):
Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.

I am a little confused that we have to inflate in the populator layer;
usually we do this in the mutating pvc webhook (on by default)/dv controllers

gaohoward:
> Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> 
> I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers

Here you are: 
https://issues.redhat.com/browse/CNV-44140 

Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.

akalenyu:
> > Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> > I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers
> 
> Here you are: https://issues.redhat.com/browse/CNV-44140
> 
> Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.

I see, though I am still confused by inflating this far into the process. Doesn't the target PVC get inflated anyhow by the webhook/DV controller (and thus so does the temp)?

akalenyu:
Also note that bug describes a separate issue that may still exist
```
$ oc create  -f dv-hpp-cloned-from-datasource.yaml
Error from server: error when creating "dv-hpp-cloned-from-datasource.yaml": admission webhook "datavolume-validate.cdi.kubevirt.io" denied the request:  Storage size is missing
```

gaohoward:
> > > Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> > > I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers
> > 
> > 
> > Here you are: https://issues.redhat.com/browse/CNV-44140
> > Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.
> 
> I see, though I am still confused by inflating this far into the process. Doesn't the target PVC get inflated anyhow by the webhook/DV controller (and thus so does the temp)?

No the target PVC doesn't get inflated when doing snapshot clone. When the controller plans the clone it finds out the storage class is different from the volume snapshot's one and then it fall back to host assisted clone. When doing host clone phase it create the tmp target PVC, if we don't check here then it will report the failure because the target size is not enough (for this to happen the tmp target PVC volume mode must be filesystem).

gaohoward:
> dv-hpp-cloned-from-datasource.yaml

It doesn't happen for me on upstream. Can you share your yaml file?

akalenyu:
> > dv-hpp-cloned-from-datasource.yaml
> 
> It doesn't happen for me on upstream. Can you share your yaml file?

Yeah I don't see it either, may have to do with an older version



> > > > Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> > > > I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers
> > > 
> > > 
> > > Here you are: https://issues.redhat.com/browse/CNV-44140
> > > Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.
> > 
> > 
> > I see, though I am still confused by inflating this far into the process. Doesn't the target PVC get inflated anyhow by the webhook/DV controller (and thus so does the temp)?
> 
> No the target PVC doesn't get inflated when doing snapshot clone. When the controller plans the clone it finds out the storage class is different from the volume snapshot's one and then it fall back to host assisted clone. When doing host clone phase it create the tmp target PVC, if we don't check here then it will report the failure because the target size is not enough (for this to happen the tmp target PVC volume mode must be filesystem).

I was pretty sure we inflate filesystem target PVCs here - https://github.com/kubevirt/containerized-data-importer/blob/e147bbdf17252d9bd3b29738cf12c0734c9e6884/pkg/controller/datavolume/util.go#L314 Sorry for poking I just think we have a problem elsewhere and that inflating the temp PVC in the populator will just cover that up

gaohoward:
> > > dv-hpp-cloned-from-datasource.yaml
> > 
> > 
> > It doesn't happen for me on upstream. Can you share your yaml file?
> 
> Yeah I don't see it either, may have to do with an older version
> 
> > > > > Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> > > > > I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers
> > > > 
> > > > 
> > > > Here you are: https://issues.redhat.com/browse/CNV-44140
> > > > Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.
> > > 
> > > 
> > > I see, though I am still confused by inflating this far into the process. Doesn't the target PVC get inflated anyhow by the webhook/DV controller (and thus so does the temp)?
> > 
> > 
> > No the target PVC doesn't get inflated when doing snapshot clone. When the controller plans the clone it finds out the storage class is different from the volume snapshot's one and then it fall back to host assisted clone. When doing host clone phase it create the tmp target PVC, if we don't check here then it will report the failure because the target size is not enough (for this to happen the tmp target PVC volume mode must be filesystem).
> 
> I was pretty sure we inflate filesystem target PVCs here -
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e147bbdf17252d9bd3b29738cf12c0734c9e6884/pkg/controller/datavolume/util.go#L314
> Sorry for poking I just think we have a problem elsewhere and that inflating the target in the populator will just cover that up

This method is never called in this case. Actually the webhook rendering is not enabled. For snapshot cloning it basically fallback to host-assisted during the planning phase, and when the host-clone phase is handling the temp target pvc creation, it has a chance to make sure the target size is right (inflate if necessary).

akalenyu:
> > > > dv-hpp-cloned-from-datasource.yaml
> > > 
> > > 
> > > It doesn't happen for me on upstream. Can you share your yaml file?
> > 
> > 
> > Yeah I don't see it either, may have to do with an older version
> > > > > > Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> > > > > > I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers
> > > > > 
> > > > > 
> > > > > Here you are: https://issues.redhat.com/browse/CNV-44140
> > > > > Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.
> > > > 
> > > > 
> > > > I see, though I am still confused by inflating this far into the process. Doesn't the target PVC get inflated anyhow by the webhook/DV controller (and thus so does the temp)?
> > > 
> > > 
> > > No the target PVC doesn't get inflated when doing snapshot clone. When the controller plans the clone it finds out the storage class is different from the volume snapshot's one and then it fall back to host assisted clone. When doing host clone phase it create the tmp target PVC, if we don't check here then it will report the failure because the target size is not enough (for this to happen the tmp target PVC volume mode must be filesystem).
> > 
> > 
> > I was pretty sure we inflate filesystem target PVCs here -
> > https://github.com/kubevirt/containerized-data-importer/blob/e147bbdf17252d9bd3b29738cf12c0734c9e6884/pkg/controller/datavolume/util.go#L314
> > 
> > Sorry for poking I just think we have a problem elsewhere and that inflating the target in the populator will just cover that up
> 
> This method is never called in this case. Actually the webhook rendering is not enabled. For snapshot cloning it basically fallback to host-assisted during the planning phase, and when the host-clone phase is handling the temp target pvc creation, it has a chance to make sure the target size is right (inflate if necessary).

Webhook rendering is enabled since a few releases now, but anyway renderPvcSpecVolumeSize existed prior to it for the DV storage API. How come the target doesn't get inflated?

Acedus:
I think the real issue here is that we don't inflate the PVC during mutation if it has a data source:
https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L65-L88

https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L246-L255

We simply set the restore size if it's available and leave it as is otherwise.

gaohoward:
> I think the real issue here is that we don't inflate the PVC during mutation if it has a data source:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L65-L88
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L246-L255
> 
> We simply set the restore size if it's available and leave it as is otherwise.

I rerun the test again and yes the webhook rendering is enabled. However the issue is that if the snapshot clone going as normal (doesn't fall back to host assisted) the target will get inflated as you said. But when it falls back to host clone, (clone strategy is copy), when the host-clone phase prepares the target tmp PVC, it didn't check the size
and do inflate if necessary. Here we added it so that the host clone won't get size problem. 

https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/populators/clone-populator.go#L315

After this point the host clone takes over. If it doesn't fall back to host clone, the methods you mentioned get called and works. But if it go through host clone phase, we need this fix to take care of the size issue.

Acedus:
> > I think the real issue here is that we don't inflate the PVC during mutation if it has a data source:
> > https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L65-L88
> > 
> > https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L246-L255
> > 
> > We simply set the restore size if it's available and leave it as is otherwise.
> 
> I rerun the test again and yes the webhook rendering is enabled. However the issue is that if the snapshot clone going as normal (doesn't fall back to host assisted) the target will get inflated as you said. But when it falls back to host clone, (clone strategy is copy), when the host-clone phase prepares the target tmp PVC, it didn't check the size and do inflate if necessary. Here we added it so that the host clone won't get size problem.
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/populators/clone-populator.go#L315
> 
> After this point the host clone takes over. If it doesn't fall back to host clone, the methods you mentioned get called and works. But if it go through host clone phase, we need this fix to take care of the size issue.

Even if it does fall back to host clone the target PVC is created so it should be mutated according to the logic I wrote above, shouldn't we fix that logic instead?

gaohoward:
> > I think the real issue here is that we don't inflate the PVC during mutation if it has a data source:
> > https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L65-L88
> > 
> > https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L246-L255
> > 
> > We simply set the restore size if it's available and leave it as is otherwise.
> 
> I rerun the test again and yes the webhook rendering is enabled. However the issue is that if the snapshot clone going as normal (doesn't fall back to host assisted) the target will get inflated as you said. But when it falls back to host clone, (clone strategy is copy), when the host-clone phase prepares the target tmp PVC, it didn't check the size and do inflate if necessary. Here we added it so that the host clone won't get size problem.
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/populators/clone-populator.go#L315
> 
> After this point the host clone takes over. If it doesn't fall back to host clone, the methods you mentioned get called and works. But if it go through host clone phase, we need this fix to take care of the size issue.

Not sure that's another issue or not but this doesn't help the host assisted clone. The host clone will create a temp target PVC based on its clone source (in this case a volume snapshot). The temp targetPVC must satisfy the size requirement (infalted if necessary)

gaohoward:
> > > > > dv-hpp-cloned-from-datasource.yaml
> > > > 
> > > > 
> > > > It doesn't happen for me on upstream. Can you share your yaml file?
> > > 
> > > 
> > > Yeah I don't see it either, may have to do with an older version
> > > > > > > Thank you for tackling this! I would appreciate a link to the bug so I can do some testing and see the origin of the issue.
> > > > > > > I am a little confused that we have to inflate in the populator layer; usually we do this in the mutating pvc webhook (on by default)/dv controllers
> > > > > > 
> > > > > > 
> > > > > > Here you are: https://issues.redhat.com/browse/CNV-44140
> > > > > > Yes the issue happens at the very late stage where the snapshot clone falls back to host assisted clone.
> > > > > 
> > > > > 
> > > > > I see, though I am still confused by inflating this far into the process. Doesn't the target PVC get inflated anyhow by the webhook/DV controller (and thus so does the temp)?
> > > > 
> > > > 
> > > > No the target PVC doesn't get inflated when doing snapshot clone. When the controller plans the clone it finds out the storage class is different from the volume snapshot's one and then it fall back to host assisted clone. When doing host clone phase it create the tmp target PVC, if we don't check here then it will report the failure because the target size is not enough (for this to happen the tmp target PVC volume mode must be filesystem).
> > > 
> > > 
> > > I was pretty sure we inflate filesystem target PVCs here -
> > > https://github.com/kubevirt/containerized-data-importer/blob/e147bbdf17252d9bd3b29738cf12c0734c9e6884/pkg/controller/datavolume/util.go#L314
> > > 
> > > Sorry for poking I just think we have a problem elsewhere and that inflating the target in the populator will just cover that up
> > 
> > 
> > This method is never called in this case. Actually the webhook rendering is not enabled. For snapshot cloning it basically fallback to host-assisted during the planning phase, and when the host-clone phase is handling the temp target pvc creation, it has a chance to make sure the target size is right (inflate if necessary).
> 
> Webhook rendering is enabled since a few releases now, but anyway renderPvcSpecVolumeSize existed prior to it for the DV storage API. How come the target doesn't get inflated?

Yes webhook rendering is enabled. Please see my reply to @Acedus for explanation why it is happening

akalenyu:
> I think the real issue here is that we don't inflate the PVC during mutation if it has a data source:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L65-L88
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L246-L255
> 
> We simply set the restore size if it's available and leave it as is otherwise.

This path is not exercised though if a size exists https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L225-L228

Acedus:
> > I think the real issue here is that we don't inflate the PVC during mutation if it has a data source:
> > https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L65-L88
> > 
> > https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L246-L255
> > 
> > We simply set the restore size if it's available and leave it as is otherwise.
> 
> This path is not exercised though if a size exists
> 
> https://github.com/kubevirt/containerized-data-importer/blob/e1bfde033dc8c83ecdbfea0e555f35f42704c03a/pkg/controller/datavolume/util.go#L225-L228

~~Result is the same, if we would hit the latter path (`renderPvcSpecVolumeSize`) we'd correctly inflate the size.~~

EDIT: The target tmp PVC doesn't have a dataSource so in theory it should just hit the standard path and inflate properly.

awels (approved):
Logic looks solid to me. Just to make sure I understand `createClaim` is used to create the 'source' PVC from the snapshot, after which it can be host assisted clone right?

akalenyu:
/hold
I think we've agreed to pivot entirely from this approach? we've had some discussion on this PR and offline

awels:
I must have missed some discussion then. I don't understand what other options we have.

awels:
just checking if this is a consistent failure in this PR or not.

akalenyu:
> /test pull-containerized-data-importer-e2e-nfs just checking if this is a consistent failure in this PR or not.

The CI environment for nfs is really bad, once in a while it suffers from a run filled with OOMs
```
2026-03-16T14:29:19Z	DEBUG	controller.import-controller	Pod died of an OOM
```

gaohoward:
@awels @akalenyu So far IMO this is the proper place to put this fix. If you guys have a better alternative I'd like to change.

gaohoward:
> snapshot, after which it can be host assisted clone right?

Yes the method is used to create the "source pvc". But at this point its already in the 'host assisted clone' phase.
The 'fallback' happens earlier in 
[planner.ChooseStrategy(ctx, args)](https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/clone/planner.go#L157)

where the planner choose the strategy based on the source kind
   in the VolumeCloneSource spec.source.kind (pvc or volumeSnapshot)
   our case is snapshot so it calls [computeStrategyForSourceSnapshot()](https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/clone/planner.go#L370)

where after some checks, it fetches the VolumeSnapshot
 it then gets the snapshotcontent and pvc's storage class
and does a series of checks, one of the checks matches our
 case, [cc.ValidateSnapshotCloneProvisioners()](https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/clone/planner.go#L407):
 it finds the target storage provisioner (hostpath-csi, for example)
 is different from the source (in volumeSnapshotContent.spec.Driver)
which is rook-ceph.rbd.csi.ceph.com, for example.

Because of this, the planner decides to fall back to HostAssisted
and plans as such.

@awels @akalenyu ^^

akalenyu:
I think the issue you were solving is really that sizeless clones were not being inflated, which is an issue outside the populator, and thus I don't think we should be tackling it in that component. I believe the fault sits in PVC spec rendering (whether through webhook/otherwise)

awels:
@akalenyu I think this particular flow should be in the populator flow, since we create the 'source PVC' to do the host assisted clone. The bug being fixed is that the created source PVC was of the incorrect size. I don't understand what you are saying when you say it needs to be fixed elsewhere.

akalenyu:
> @akalenyu I think this particular flow should be in the populator flow, since we create the 'source PVC' to do the host assisted clone. The bug being fixed is that the created source PVC was of the incorrect size. I don't understand what you are saying when you say it needs to be fixed elsewhere.

The target size is wrong as well, that is what I'm trying to say. The temp PVC being of the wrong size is just one symptom.

awels:
> > @akalenyu I think this particular flow should be in the populator flow, since we create the 'source PVC' to do the host assisted clone. The bug being fixed is that the created source PVC was of the incorrect size. I don't understand what you are saying when you say it needs to be fixed elsewhere.
> 
> The target size is wrong as well, that is what I'm trying to say. The temp PVC being of the wrong size is just one symptom.

That would be a huge issue everywhere not just this particular case

gaohoward:
> > @akalenyu I think this particular flow should be in the populator flow, since we create the 'source PVC' to do the host assisted clone. The bug being fixed is that the created source PVC was of the incorrect size. I don't understand what you are saying when you say it needs to be fixed elsewhere.
> 
> The target size is wrong as well, that is what I'm trying to say. The temp PVC being of the wrong size is just one symptom.

For 'snapshot' clone the sourcePVC is not a problem because cdi is just use its 'restoreSize' from the VolumeSnapshot. The target PVC however comes from the user. If user specifies a smaller size cdi needs to
adjust and enlarge it. If user doesn't specify request size at all (as reported in the jira) cdi copies the source
snapshot's restore size and just don't bother to inflate (because the snapshot clone requires same size of source
and target). If this doesn't fall back CDI would just handed the clone to the csi-driver. However in our case
when CDI plans the clone the result is fallback to Host-assisted clone, which can be seen as a total different clone. The host-assisted clone should have its own size check to make sure the source/targetPvc size is valid.

I'm a bit lost as to where to move the code exactly. A bit more advice on it would be appreciated. :)

Acedus:
Hi @howardgao and @awels. I must admit that I'm a bit confused by the choice to do the inflation as part of the cloning process, after all, `tmp-pvc` is created as a copy of the target PVC so if our size calculation for it is wrong then we should address the latter and not the former. 

Correct me if I'm wrong but the flow as I understand it is:
target DV is created without a size -> we reach initial PVC volume size rendering from the DV base controller and we set its size to 0 because it's a clone and we defer that handling to the relevant clone controller: https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L310-L319

In our case, it's the snapshot controller, so we hit the following path:
https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/snapshot-clone-controller.go#L281-L301

Because the request size is 0, we set the PVC size to that of the snapshot restore size, but in doing so we completely ignore the need for potential inflation. 

My question thus becomes why are we not handling inflation at any of those points instead?

gaohoward:
> Hi @howardgao and @awels. I must admit that I'm a bit confused by the choice to do the inflation as part of the cloning process, after all, `tmp-pvc` is created as a copy of the target PVC so if our size calculation for it is wrong then we should address the latter and not the former.
> 
> Correct me if I'm wrong but the flow as I understand it is: target DV is created without a size -> we reach initial PVC volume size rendering from the DV base controller and we set its size to 0 because it's a clone and we defer that handling to the relevant clone controller:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L310-L319
> 
> In our case, it's the snapshot controller, so we hit the following path:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/snapshot-clone-controller.go#L281-L301
> 
> Because the request size is 0, we set the PVC size to that of the snapshot restore size, but in doing so we completely ignore the need for potential inflation.
> 
> My question thus becomes why are we not handling inflation at any of those points instead?

Hi @Acedus I believe for snapshot clone (without falling back) this is correct process, i.e. don't inflate the source size and target size have to be equal. I got this hint [Here](https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L80)

Acedus:
> Hi @Acedus I believe for snapshot clone (without falling back) this is correct process, i.e. don't inflate the source size and target size have to be equal. I got this hint [Here](https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L80)

The thing I find problematic with that is that for the scenario where we do specify a size for the DV, we will inflate the PVC. We currently never hit the path you've referenced since by default the `cdi.kubevirt.io/applyStorageProfile` label isn't added to DVs by default, so it won't get propagated to the PVC and thus it won't satisfy the mutating webhook object selector:
```yaml
  namespaceSelector: {}
  objectSelector:
    matchLabels:
      cdi.kubevirt.io/applyStorageProfile: "true"
  reinvocationPolicy: Never
  rules:
  - apiGroups:
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    resources:
    - persistentvolumeclaims
    scope: '*'
```

So instead it will hit `pvcFromStorage and because the rendering webhook is enabled by default, it will inflate: https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L119-L148
https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L309-L340

It could be argued that `shouldRender` or the logic that follows it is therefore wrong and doesn't take into account the comment you've referenced regarding clones (maybe we should take that into account and debate that), but right now I believe it should be hitting the same paths for sizeless as it does when one specifies a size.

akalenyu:
> In our case, it's the snapshot controller, so we hit the following path:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/snapshot-clone-controller.go#L281-L301
> 
> Because the request size is 0, we set the PVC size to that of the snapshot restore size, but in doing so we completely ignore the need for potential inflation.

Good spot, if we just inflate here we should be fine. WDYT @howardgao?

I do agree that we have to take a look at other points but this is out of scope for this PR, any non DV path can be looked at later

gaohoward:
> > In our case, it's the snapshot controller, so we hit the following path:
> > https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/snapshot-clone-controller.go#L281-L301
> > 
> > Because the request size is 0, we set the PVC size to that of the snapshot restore size, but in doing so we completely ignore the need for potential inflation.
> 
> Good spot, if we just inflate here we should be fine. WDYT @howardgao?
> 
> I do agree that we have to take a look at other points but this is out of scope for this PR, any non DV path can be looked at later

I don't think it's fine to inflate at this point because it is still 'snapshot-clone' processing which means if it goes on as snapshot clone, all the rest will be handled by csi driver -- no host assist clone. The inflation is a 'host assisted clone' issue and should stay in host assist clone process.

gaohoward:
> > Hi @Acedus I believe for snapshot clone (without falling back) this is correct process, i.e. don't inflate the source size and target size have to be equal. I got this hint [Here](https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L80)
> 
> The thing I find problematic with that is that for the scenario where we do specify a size for the DV, we will inflate the PVC. We currently never hit the path you've referenced since by default the `cdi.kubevirt.io/applyStorageProfile` label isn't added to DVs by default, so it won't get propagated to the PVC and thus it won't satisfy the mutating webhook object selector:
> 
> ```yaml
>   namespaceSelector: {}
>   objectSelector:
>     matchLabels:
>       cdi.kubevirt.io/applyStorageProfile: "true"
>   reinvocationPolicy: Never
>   rules:
>   - apiGroups:
>     - ""
>     apiVersions:
>     - v1
>     operations:
>     - CREATE
>     resources:
>     - persistentvolumeclaims
>     scope: '*'
> ```
> 
> So instead it will hit `pvcFromStorage and because the rendering webhook is enabled by default, it will inflate:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L119-L148
> 
> https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/util.go#L309-L340
> 
> It could be argued that `shouldRender` or the logic that follows it is therefore wrong and doesn't take into account the comment you've referenced regarding clones (maybe we should take that into account and debate that), but right now I believe it should be hitting the same paths for sizeless as it does when one specifies a size.

It deserves more investigation and itself can be treated as a separate issue.

akalenyu:
> > > In our case, it's the snapshot controller, so we hit the following path:
> > > https://github.com/kubevirt/containerized-data-importer/blob/5591d897e06ddf51e60800fbf7a626215a75ea6f/pkg/controller/datavolume/snapshot-clone-controller.go#L281-L301
> > > 
> > > Because the request size is 0, we set the PVC size to that of the snapshot restore size, but in doing so we completely ignore the need for potential inflation.
> > 
> > 
> > Good spot, if we just inflate here we should be fine. WDYT @howardgao?
> > I do agree that we have to take a look at other points but this is out of scope for this PR, any non DV path can be looked at later
> 
> I don't think it's fine to inflate at this point because it is still 'snapshot-clone' processing which means if it goes on as snapshot clone, all the rest will be handled by csi driver -- no host assist clone. The inflation is a 'host assisted clone' issue and should stay in host assist clone process.

I think the intention was to inflate for all cases, similarly to the PVC sizeless handling. But in any case, you can check if it's a block->fs clone in here too and guarantee it's host assisted. I think that is the sweet spot. I really hate to block this PR but I truly believe inflating through tmp/populator path isn't the right choice here

dsanatar:
@akalenyu it seems this PR has been superseded by the implementation of #4107 , is it safe to assume we can close this one out?

akalenyu:
> @akalenyu it seems this PR has been superseded by the implementation of #4107 , is it safe to assume we can close this one out?

yup
/close

PR #2750, opened 2023-06-13
DataVolume Controller uses VolumeCloneSource Populator

mhenriks:
**What this PR does / why we need it**:

The DataVolume controller should use our populators internally for CSI storage provisioners.  This PR builds off of #2722 to include `VolumeCloneSource` populators.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

I think everything is here but I bet there will be more whackamole on func tests

**Release note**:

```release-note
DataVolume Controller uses VolumeCloneSource Populator
```

PR #3902, opened 2025-09-10
Enable CSI cloning functionality when CDI populators are disabled

noamasu:
This draft introduces an attempt to allow DataVolumes to use CSI volume cloning (instead of falling back to host-assisted clone strategy) even when the usePopulator feature is set to false, providing direct cloning for storage classes that support CSI clone operations.

added csi-clone support for both pvc and snapshot sources - withing the cloning process, PVC with a dataSource/dataSourceRef pointing to the pvc/volumesnapshot will be created.

Following our discussion about Portworx’s set of the userPopulator field to false, we decided to reintroduce the ability to perform a CSI-based clone even without CDI populators. This avoids falling back to host-assisted cloning, which significantly increases clone completion time. 

**Which issue(s) this PR fixes** *
Jira Ticket: https://issues.redhat.com/browse/CNV-66958
for additional context:
The PR that removed csi-cloning after introducing CDI populators: https://github.com/kubevirt/containerized-data-importer/pull/2750/commits/2285ceb46045bc281264591b4d0497a57f9bd527#r1235350852

@alromeros  
@akalenyu 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/datavolume/BUILD.bazel
+++ b/pkg/controller/datavolume/BUILD.bazel
@@ -24,6 +24,7 @@ go_library(
         "//pkg/monitoring/metrics/cdi-cloner:go_default_library",
         "//pkg/monitoring/metrics/cdi-controller:go_default_library",
         "//pkg/monitoring/metrics/cdi-importer:go_default_library",
+        "//pkg/storagecapabilities:go_default_library",
         "//pkg/token:go_default_library",
         "//pkg/util:go_default_library",
         "//staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1:go_default_library",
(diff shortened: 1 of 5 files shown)


akalenyu:
I would really prefer that we instead help Portworx overcome their issue with using populators -
We don't test the non populator path nearly enough to have confidence in it working.
We have bug fixes only going in the populator path for a while now too, so that's also risky.

/cc @mhenriks

mhenriks:
> I would really prefer that we instead help Portworx overcome their issue with using populators - We don't test the non populator path nearly enough to have confidence in it working. We have bug fixes only going in the populator path for a while now too, so that's also risky.
> 
> /cc @mhenriks

This will definitely be a maintenance burden but also don't forget that without populators we cannot support `WaitForFirstConsumer` snapshot/csi cloning so the benefits are severely limited

PR #3912, opened 2025-09-17
DV Recreated with Wrong Status

Dsanatar:

When a DV that is apart of a VM's DVTemplate gets deleted occasionally during the recreation, the new DV will be seen to immediately have status of Succeeded with Progress at 100% and then will eventually switch back to Pending. This is due to a race condition where the associated PVC is not deleted/cleaned up in time for when the new DV goes to update its status. Since the new DV will see the old PVC is bound, it will immediately update it's first status to Succeeded.

Added a new check during `updateStatus()` to prevent DVs from using PVCs that have a `DeletionTimeStamp`

**Release note**:

```release-note
Check for PVC DeletionTimeStamp before updating DV status
```

### DIFF
--- a/pkg/controller/datavolume/import-controller_test.go
+++ b/pkg/controller/datavolume/import-controller_test.go
@@ -1466,6 +1466,44 @@ var _ = Describe("All DataVolume Tests", func() {
 			Expect(readyCondition.Message).To(Equal(""))
 		})
 
+		It("Should not use PVC to update status if PVC is marked for deletion", func() {
+			dv := NewImportDataVolume("test-dv")
+			annotations := map[string]string{
+				AnnPopulatedFor: "test-dv",
+				AnnPodPhase:     string(corev1.PodSucceeded),
+			}
+			pvc := CreatePvc("test-dv", metav1.NamespaceDefault, annotations, nil)
+			// Set finalizer so PVC is not removed when Delete is called
+			pvc.Finalizers = append(pvc.Finalizers, "test-finalizer")
+
+			reconciler = createImportReconciler(dv, pvc)
+			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}})
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Marking PVC for deletion")
+			err = reconciler.client.Delete(context.TODO(), pvc)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Resetting DV to Pending phase")
+			pendingStatus := cdiv1.Pending
+			dv = &cdiv1.DataVolume{}
+			err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, dv)
+			Expect(err).ToNot(HaveOccurred())
+			dv.Status.Phase = pendingStatus
+			err = reconciler.client.Status().Update(context.TODO(), dv)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Calling updateStatus")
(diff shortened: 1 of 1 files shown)


akalenyu:
> Good catch! Testing this would be great but seems tricky to do. /lgtm

I think we should be able to supply the erroneous state to a fake reconciler in a unit test, tbh, that would also make me feel a little better about this change as I don't completely understand the scenario

alromeros:
> > Good catch! Testing this would be great but seems tricky to do. /lgtm
> 
> I think we should be able to supply the erroneous state to a fake reconciler in a unit test, tbh, that would also make me feel a little better about this change as I don't completely understand the scenario

Agree but isolating this logic might not be straightforward, and also there're no utests for this file. The behavior is racy but maybe it's worth having a functest that deletes the pvc and ensures the DV completes nicely to ensure there're no weird behaviors in between?

alromeros:
/hold
Let's discuss first the testing options

akalenyu:
> > > Good catch! Testing this would be great but seems tricky to do. /lgtm
> > 
> > 
> > I think we should be able to supply the erroneous state to a fake reconciler in a unit test, tbh, that would also make me feel a little better about this change as I don't completely understand the scenario
> 
> Agree but isolating this logic might not be straightforward, and also there're no utests for this file. The behavior is racy but maybe it's worth having a functest that deletes the pvc and ensures the DV completes nicely to ensure there're no weird behaviors in between?

That's prone to flakes, I'd unit test it with a pinpointed state identified to be wrong
We can use any of the unit test suites for controller that piggyback on the controller base, so import for example

alromeros:
/unhold
Looks great!
/lgtm

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -1466,6 +1466,64 @@ var _ = Describe("All DataVolume Tests", func() {
...
+
+			// Mark PVC for deletion
+			err = reconciler.client.Delete(context.TODO(), pvc)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Reset the DV status phase
+			newPhase := cdiv1.Pending
+			dv.Status.Phase = newPhase
+			err = reconciler.client.Status().Update(context.TODO(), dv)
+			Expect(err).ToNot(HaveOccurred())
```
akalenyu: Which entity would be causing this in a "real" setup?

dsanatar: The problem spawns from when a deleted DV from a VM's DVTemplate is recreated. Here I am trying to emulate that behavior of the "new" DV by just resetting the status and making sure it doesn't inherit a new status from it's associated PVC that is in the process of deletion.

akalenyu: I see, maybe we can reduce the test to that then? IOW are all the setup steps really needed?

dsanatar: Yea good point, I can give that a go and see.

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -1466,6 +1466,49 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			pvc.GetAnnotations()[AnnPodPhase] = string(corev1.PodSucceeded)
+
+			// Set finalizer so PVC is not removed when Delete is called
+			pvc.Finalizers = append(pvc.Finalizers, "test-finalizer")
+			err = reconciler.client.Update(context.TODO(), pvc)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Mark PVC for deletion
+			err = reconciler.client.Delete(context.TODO(), pvc)
+			Expect(err).ToNot(HaveOccurred())
```
akalenyu, line 1485: You don't have to use the client to setup the state, you can just pass the objects in the target state to `createImportReconciler`

PR #3919, opened 2025-09-19
dataimportcron-controller: fix nil map assignment

Acedus:
The code path assumed that workloadNodePlacement's NodeSelector map was already initialized. This commit ensures that it is initialized before assignment to avoid a panic when the map is nil.

Fixes #

**Release note**:

```release-note
NONE
```


awels:
/lgtm
/approve
Thanks for finding this.

PR #3927, opened 2025-09-28
Add volume-data-source-validator support

Acedus:
The volume-data-source-validator project[1] introduces the VolumePopulator CRD which is used to register data sources which serve as volume populators and validate them. This PR extends CDI to detect whether the VolumePopulator CRD exists and if so, create VolumePopulators that match the ones which CDI provides.

[1] https://github.com/kubernetes-csi/volume-data-source-validator

Fixes https://issues.redhat.com/browse/CNV-69838

**Release note**:

```release-note
CDI's volume populators are now compatible with volume-data-source-validator
```


pkg/operator/controller/callbacks.go
```diff
@@ -63,6 +63,9 @@ func addReconcileCallbacks(r *ReconcileCDI) {
 	if r.haveRoutes {
 		r.reconciler.AddCallback(&appsv1.Deployment{}, reconcileRoute)
 	}
+	if r.haveVolumeDataSourceValidator {
```
alromeros, line 66: Maybe we should just care about creating and handling our VolumePopulators without checking whether the validator is installed? For routes it makes sense as they are openshift-exclusive, but if having VolumePopulators is the k8s standard maybe we should be creating them anyway.

Acedus: I agree, but `VolumePopulator` is a validator API: https://github.com/kubernetes-csi/volume-data-source-validator/blob/90ffb8220766aba39ba7fea5e806952714090f31/client/config/crd/populator.storage.k8s.io_volumepopulators.yaml

pkg/operator/controller/volumepopulator.go
```diff
@@ -0,0 +1,122 @@
...
+
+		if err = controllerutil.SetControllerReference(cdi, desiredPop, scheme); err != nil {
+			return false, err
+		}
+
+		if err := c.Create(ctx, desiredPop); err != nil {
+			return false, err
+		}
+	}
+	return true, nil
```
akalenyu: PR looks great only thing I was worried about is event spam over updated==true,
which turned out to be the case:
```bash
$ k get events --all-namespaces -o json | jq -r '.items[] | select(.message == "Successfully ensured volume populators exist") | .count'
8
```
(grows over time, this gets caught in conformance tests in openshift for example)

Here's a reference for a similar issue:
https://github.com/kubevirt/containerized-data-importer/pull/3490

akalenyu (commented):
I think the approach makes sense and so does the PR itself. Only worry is event spam

Here's the original populator blog strongly recommending installing those alongside the populator:
https://kubernetes.io/blog/2022/05/16/volume-populators-beta/
> Cluster administrations are strongly encouraged to install the volume-data-source-validator controller and associated VolumePopulator CRD before installing any populators so that users can get feedback about invalid PVC data sources.

akalenyu:
/test pull-cdi-unit-test
unrelated; https://github.com/kubevirt/containerized-data-importer/issues/3341#issuecomment-2316174754
we could get rid of the prlimit tbh and just rely on cgroups

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/597e0c26a1214e0562221cf60ac4c347aba3a9f1..57a96965cb402993367c28e52b7aad1b6c65a681) - Addressed @akalenyu's comment and improved VolumePopulator's creation/update logic.

pkg/operator/controller/volumepopulator.go
```diff
@@ -0,0 +1,107 @@
...
+			util.SetRecommendedLabels(desiredPop, installerLabels, "cdi-operator")
+			desiredPop.SourceKind = metav1.GroupKind(gvk.GroupKind())
+			return controllerutil.SetControllerReference(cdi, desiredPop, scheme)
+		})
+		if err != nil {
+			return false, fmt.Errorf("failed to create/update VolumePopulator %s: %w", popName, err)
+		}
+
+		if result == controllerutil.OperationResultCreated {
+			created = true
```
akalenyu: So I think that we normally fire events for both update/creates
(so anything that required our intervention, basically)

alromeros:
Looks great, thanks!
/lgtm

PR #3933, opened 2025-10-06
Copy Events from tmp PVCs during Clone

Dsanatar:

Extension of #3764

If a clone gets stuck for whatever reason it is helpful to be able to see the original error from the PVC/DV that initiated the clone. 

Leveraging the functionality from the above PR to copy events from tmp-pvcs during a clone to the their owner pvc.

Fixes #

**Release note**:

```release-note
Events from Clone PVCs
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -2219,3 +2219,53 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
 
 	return nil
 }
+
+// CopyEvents gets srcPvc events and re-emits them on the target PVC with the src name prefix
+func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder record.EventRecorder) {
+	srcPrefixMsg := fmt.Sprintf("[%s] : ", srcPVC.GetName())
+
+	newEvents := &corev1.EventList{}
+	err := c.List(context.TODO(), newEvents,
+		client.InNamespace(srcPVC.GetNamespace()),
+		client.MatchingFields{"involvedObject.name": srcPVC.GetName(),
+			"involvedObject.uid": string(srcPVC.GetUID())},
+	)
+
+	if err != nil {
+		klog.Error(err, "Could not retrieve srcPVC list of Events")
+	}
+
+	currEvents := &corev1.EventList{}
+	err = c.List(context.TODO(), currEvents,
+		client.InNamespace(targetPVC.GetNamespace()),
+		client.MatchingFields{"involvedObject.name": targetPVC.GetName(),
+			"involvedObject.uid": string(targetPVC.GetUID())},
+	)
+
+	if err != nil {
+		klog.Error(err, "Could not retrieve targetPVC list of Events")
+	}
+
+	// use this to hash each message for quick lookup, value is unused
+	eventMap := make(map[string]bool)
+
+	for _, event := range currEvents.Items {
+		eventMap[event.Message] = true
+	}
+
+	for _, newEvent := range newEvents.Items {
+		msg := newEvent.Message
+
+		// check if target PVC already has this equivalent event
+		if _, exists := eventMap[msg]; exists {
+			continue
+		}
+
+		formattedMsg := srcPrefixMsg + msg
+		// check if we already emitted this event with the src prefix
+		if _, exists := eventMap[formattedMsg]; exists {
+			continue
+		}
+		recorder.Event(targetPVC, newEvent.Type, newEvent.Reason, formattedMsg)
+	}
+}
--- a/pkg/controller/clone/csi-clone.go
+++ b/pkg/controller/clone/csi-clone.go
@@ -73,6 +73,12 @@ func (p *CSIClonePhase) Reconcile(ctx context.Context) (*reconcile.Result, error
 		}
 	}
 
+	targetPvc, err := cc.GetAnnotatedEventSource(ctx, p.Client, pvc)
+	if err != nil {
+		return nil, err
+	}
+	cc.CopyEvents(pvc, targetPvc, p.Client, p.Recorder)
+
 	done, err := isClaimBoundOrWFFC(ctx, p.Client, pvc)
 	if err != nil {
 		return nil, err
--- a/pkg/controller/clone/host-clone.go
+++ b/pkg/controller/clone/host-clone.go
@@ -165,6 +165,12 @@ func (p *HostClonePhase) Reconcile(ctx context.Context) (*reconcile.Result, erro
 		}
 	}
 
+	targetPvc, err := cc.GetAnnotatedEventSource(ctx, p.Client, actualClaim)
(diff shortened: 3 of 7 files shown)


pkg/controller/common/util.go
```diff
@@ -2219,3 +2219,53 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
...
+		eventMap[event.Message] = true
+	}
+
+	for _, newEvent := range newEvents.Items {
+		msg := newEvent.Message
+
+		// check if target PVC already has this equivalent event
+		if _, exists := eventMap[msg]; exists {
+			continue
+		}
```
akalenyu, line 2262: How do we feel about using the message as a sign of existence?
I can totally imagine some part of the message being dynamic

akalenyu (commented):
Just a quick first pass :+1: 

Do I understand correctly that the host assisted/upload controllers are already okay? It's just the csi/snap clone that's broken?

BTW, we should be able to unit test this
(we already use record.NewFakeRecorder in some unit test suites)

pkg/controller/common/util.go
akalenyu: Depending on your answer (maybe this is somehow a non issue) we can come up with other ideas like a combo of `reason` and something else

pkg/controller/common/util.go
```diff
@@ -2219,3 +2219,53 @@ func UpdatePVCBoundContionFromEvents(pvc *corev1.PersistentVolumeClaim, c client
...
+		client.MatchingFields{"involvedObject.name": targetPVC.GetName(),
+			"involvedObject.uid": string(targetPVC.GetUID())},
+	)
+
+	if err != nil {
+		klog.Error(err, "Could not retrieve targetPVC list of Events")
+	}
+
+	// use this to hash each message for quick lookup, value is unused
+	eventMap := make(map[string]bool)
```
akalenyu: not a huge deal but normally in go you'd use an empty struct instead of the bool in this kind of maps, to spare the allocation:
https://stackoverflow.com/a/22770744

dsanatar: ah ok good to know thanks!

pkg/controller/common/util.go
dsanatar: It can't hurt to prepend the `reason` for a more precise comparison, will address

dsanatar:
> Just a quick first pass 👍
> 
> Do I understand correctly that the host assisted/upload controllers are already okay? It's just the csi/snap clone that's broken?
> 
> BTW, we should be able to unit test this (we already use record.NewFakeRecorder in some unit test suites)

The `CopyEvents` func was originally used to get events from a prime PVC to it's associated "parent" PVC/DV in case something got stuck during the prime PVC reconciliation. A similar issue was reported for clones (specially host clone) where the newly created tmp PVC could encounter an error but no indication of this ever makes it back to it's owner PVC/DV. I figured since each of these clone methods use a tmp PVC that could potentially get stuck during reconcile, we should bubble up these events for each clone type.

pkg/controller/common/util.go
akalenyu: Actually I think there's a `types.UID` in the event object

akalenyu: At first I wasn't sure we want to propagate duplicates but as long as our fired events will qualify for aggregation (.count increase instead of lots of CREATEs for every event) we should be good

dsanatar: Correct, any duplicate events that may get re-emitted just increment the event counter.

akalenyu:
> > Just a quick first pass 👍
> > Do I understand correctly that the host assisted/upload controllers are already okay? It's just the csi/snap clone that's broken?
> > BTW, we should be able to unit test this (we already use record.NewFakeRecorder in some unit test suites)
> 
> The `CopyEvents` func was originally used to get events from a prime PVC to it's associated "parent" PVC/DV in case something got stuck during the prime PVC reconciliation. A similar issue was reported for clones (specially host clone) where the newly created tmp PVC could encounter an error but no indication of this ever makes it back to it's owner PVC/DV. I figured since each of these clone methods use a tmp PVC that could potentially get stuck during reconcile, we should bubble up these events for each clone type.

Interesting, does this not achieve that for host clone? https://github.com/kubevirt/containerized-data-importer/blob/239edacb4e54670ec1246423aaded1265cb0c693/pkg/controller/clone/host-clone.go#L186

pkg/controller/common/util.go
akalenyu: Hmm yeah as long as we don't do anything funky (reason/event source/etc are all the same) I guess we should be fine. If the message has a dynamic element, the coalescing should break, but, that's not on us - we just propagate existing events.

dsanatar:
> > > Just a quick first pass 👍
> > > Do I understand correctly that the host assisted/upload controllers are already okay? It's just the csi/snap clone that's broken?
> > > BTW, we should be able to unit test this (we already use record.NewFakeRecorder in some unit test suites)
> > 
> > 
> > The `CopyEvents` func was originally used to get events from a prime PVC to it's associated "parent" PVC/DV in case something got stuck during the prime PVC reconciliation. A similar issue was reported for clones (specially host clone) where the newly created tmp PVC could encounter an error but no indication of this ever makes it back to it's owner PVC/DV. I figured since each of these clone methods use a tmp PVC that could potentially get stuck during reconcile, we should bubble up these events for each clone type.
> 
> Interesting, does this not achieve that for host clone?
> 
> https://github.com/kubevirt/containerized-data-importer/blob/239edacb4e54670ec1246423aaded1265cb0c693/pkg/controller/clone/host-clone.go#L186

I believe this annotation is only used to report specific events from within the controller. The issue seen in the [bug report](https://issues.redhat.com/browse/CNV-32268) that causes these tmp PVCs to fail is an event failure from the provisioner. I am however using this annotation to get the owner PVC object, so I will need to add this Ann to the snap/csi clone as well.

dsanatar:
> Just a quick first pass 👍
> 
> Do I understand correctly that the host assisted/upload controllers are already okay? It's just the csi/snap clone that's broken?
> 
> BTW, we should be able to unit test this (we already use record.NewFakeRecorder in some unit test suites)

After further investigating, it is actually slightly harder to unit test this using the FakeRecorders since the events we are looking for don't get emitted from the normal Reconcile loop, but instead from `CopyEvents`. This makes it complicated since the function calls client.List()` to get and compare events, which requires the client to be setup with the proper Indexers for the list function to work. I think I ran into this issue when I originally implemented this and opted for a functional test instead https://github.com/kubevirt/containerized-data-importer/blob/4c3dc65a772e0cd66fdbabf75c26b688f4b4afb9/tests/datavolume_test.go#L3380

awels:
/approve
I am fine with the functional test.

awels:
We are on different versions from KubeVirt, I think our release branch is v1.63

PR #3938, opened 2025-10-21
Add network policies to CDI

Acedus:
This PR extends CDI to allow an optional generation and deployment of network policies that generally allow only the traffic patterns required by CDI's internal components, those being:
- cdi-deployment
- cdi-apiserver
- cdi-operator
- cdi-uploadproxy
- DataImportCron's poller pod

It introduces a new environment variable `CDI_DEPLOY_NP` which allows controlling the deployment of said network policies during cluster-sync.

A separate static network policy that allows all traffic patterns for CDI testing pods is created if the sync is ran with `test-infra`.

Fixes #

**Release note**:

```release-note
Added example network policies as part of the manifest-generator and csv-generator tools.
```

### DIFF
--- a/tools/csv-generator/BUILD.bazel
+++ b/tools/csv-generator/BUILD.bazel
@@ -6,6 +6,7 @@ go_library(
     importpath = "kubevirt.io/containerized-data-importer/tools/csv-generator",
     visibility = ["//visibility:private"],
     deps = [
+        "//pkg/operator/resources/namespaced:go_default_library",
         "//pkg/operator/resources/operator:go_default_library",
         "//tools/util:go_default_library",
     ],
--- a/tools/csv-generator/csv-generator.go
+++ b/tools/csv-generator/csv-generator.go
@@ -16,6 +16,7 @@ import (
 	"flag"
 	"os"
 
+	cdinamespaced "kubevirt.io/containerized-data-importer/pkg/operator/resources/namespaced"
 	cdioperator "kubevirt.io/containerized-data-importer/pkg/operator/resources/operator"
 	"kubevirt.io/containerized-data-importer/tools/util"
 )
@@ -40,6 +41,7 @@ var (
 	uploadServerImage   = flag.String("uploadserver-image", "", "")
 	ovirtPopulatorImage = flag.String("ovirt-populator-image", "", "")
 	dumpCRDs            = flag.Bool("dump-crds", false, "optional - dumps cdi-operator related crd manifests to stdout")
+	dumpNetworkPolicies = flag.Bool("dump-network-policies", false, "optional - dumps cdi related network policies")
 )
 
 func main() {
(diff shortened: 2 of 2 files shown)


Acedus:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3938/pull-containerized-data-importer-e2e-nfs/1981272575073849344
Failure seems unrelated, fyi @arnongilboa.

tests/dataimportcron_test.go
```diff
@@ -472,10 +474,12 @@ var _ = Describe("DataImportCron", Serial, func() {
 		importerImage := f.GetEnvVarValue("IMPORTER_IMAGE")
 		Expect(importerImage).ToNot(BeEmpty())
 
-		pod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: pollerPodName}}
-		err = controller.InitPollerPodSpec(f.CrClient, cron, &pod.Spec, importerImage, corev1.PullIfNotPresent, log)
+		podTemplateSpec := &corev1.PodTemplateSpec{ObjectMeta: metav1.ObjectMeta{Name: pollerPodName}}
+		err = controller.InitPollerPod(f.CrClient, cron, podTemplateSpec, importerImage, corev1.PullIfNotPresent, log)
 		Expect(err).ToNot(HaveOccurred())
 
+		pod := &corev1.Pod{ObjectMeta: podTemplateSpec.ObjectMeta, Spec: podTemplateSpec.Spec}
```
awels, line 481: Do we want to check the label is on this pod? I guess we would need to check if the env variable is set, and if so we would expect the label on the pod, otherwise we would expect it to not be there.

Acedus: So the label should always be on any poller Pod in general, irrespective of whether NetworkPolicies exist so that it is identifiable by a label selector. I don't mind adding a check that the label exists on the Pod as part of a UT, is that what you meant?

awels: Yeah, or just add a check here, after the InitPollerPod is called, it should have the label right?

Acedus:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/19a03742d9270beb5fd6498748290d950c540548..31de2f6721d05d883f143afd042df1b6067b7544) - Dropped commit that enables NetworkPolicy deployment (further testing will be done in a separate periodic job which will be introduced in a subsequent PR). Addressed @awels comment.

cluster-sync/networkpolicies/cdi-testing-np.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: networking.k8s.io/v1
```
akalenyu, line 1: Do we want to ship this to external entities running our func test suite, similarly to the SCC? Totally not a must for this PR, just something to consider.

cluster-sync/networkpolicies/cdi-testing-np.yaml
```diff
@@ -0,0 +1,17 @@
...
+  podSelector:
+    matchLabels:
+      cdi.kubevirt.io/testing: ""
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress: 
+  - {}
+  egress: 
+  - {}
```
akalenyu, line 17: Just curious, what's this mostly about? http/registry test servers?

akalenyu:
/test pull-cdi-goveralls
/test pull-cdi-unit-test

looks like timeouts

cluster-sync/networkpolicies/cdi-testing-np.yaml
Acedus: Our testing pods all have unique network patterns. For the sake of brevity (and because these aren't used in any production environment), I chose to create an allow all NP for such pods.

cluster-sync/networkpolicies/cdi-testing-np.yaml
Acedus: I think that we probably should, otherwise I expect such tests which rely on these pods to fail. Not entirely sure how we should go about it without polluting the CDI namespace needlessly, though. I'm open to suggestions.

PR #3941, opened 2025-10-29
Fix VolumeImportSource DataVolume status tracking

DiMalovanyy:

  This PR fixes a bug where DataVolumes using `VolumeImportSource` via `spec.storage.dataSourceRef` were immediately marked as `Succeeded` without tracking the
  actual import progress.

  When using CDI's own volume populators (VolumeImportSource, VolumeUploadSource, VolumeCloneSource) through the dataSourceRef field, the
  external-population-controller was treating them like external populators and immediately setting them to Succeeded status, instead of tracking their actual
  phase transitions (ImportScheduled → ImportInProgress → Succeeded) and progress percentages.
  
  #### After the fix
  ```
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportScheduled    N/A                   38s
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportScheduled    N/A                   38s
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportScheduled    N/A                   48s
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportInProgress   N/A                   64s
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportInProgress   0.17%                 68s
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportInProgress   0.65%                 70s
...
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportInProgress   10.03%                85s
...
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportInProgress   19.65%                100s
  ...
kube-nfv    focal-server-cloudimg-amd64-test-block   ImportInProgress   99.46%                8m9s
kube-nfv    focal-server-cloudimg-amd64-test-block   Succeeded          100.0%                8m12s
kube-nfv    focal-server-cloudimg-amd64-test-block   Succeeded          100.0%                8m13s
kube-nfv    focal-server-cloudimg-amd64-test-block   Succeeded          100.0%                8m14s
  ```

  **Changes:**
    - Enable progress tracking in external-population-controller (set `shouldUpdateProgress: true`)
    - Add `updateCDIPopulatorStatus()` to read PVC annotations set by populator controllers and update DataVolume phases accordingly
    - Add helper functions `isCDIVolumePopulator()` and `getCDIVolumePopulatorKind()` to distinguish CDI's own populators from external ones
    - Add comprehensive unit tests for all CDI populator types (Import, Upload, Clone) covering phase transitions, progress tracking, and failure scenarios

  The fix keeps CDI volume populators routed to the external-population-controller (as designed) but adds special handling to read the PVC annotations
  (`cdi.kubevirt.io/storage.pod.phase` and `cdi.kubevirt.io/storage.populator.progress`) that are set by the respective populator controllers (import-populator,
  upload-populator, clone-populator). This mirrors the status update logic from import-controller but adapted for the populator pattern.

  All existing tests pass (302/302), including 8 new tests that verify proper phase transitions and progress tracking for CDI volume populators.

  **Testing**: A pre-built image with this fix is available at `ghcr.io/kube-nfv/cdi-controller:v1.63-dv-rec-fix` for testing purposes.

### DIFF
--- a/pkg/controller/datavolume/external-population-controller.go
+++ b/pkg/controller/datavolume/external-population-controller.go
@@ -68,7 +68,7 @@ func NewPopulatorController(ctx context.Context, mgr manager.Manager, log logr.L
 			recorder:             mgr.GetEventRecorderFor(populatorControllerName),
 			featureGates:         featuregates.NewFeatureGates(client),
 			installerLabels:      installerLabels,
-			shouldUpdateProgress: false,
+			shouldUpdateProgress: true,
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


awels:
Hi, thanks for the PR, I will review shortly. To get the `dco` check to pass, all you have to do is `git commit --amend -s` which will 'sign' the commit.

akalenyu:
Thanks for the PR! would love to hear more about the use case, we generally only create the target PVCs with our own dataSourceRefs, not the DVs themselves.

/cc @alromeros

DiMalovanyy:
Hi @akalenyu, 
I'm using volumeimportsources as an image definition: "Let's say image ID". Later multiple image instances (eg. with different storageClasses might be instantiated from it), and finally the kubevirt VM will creates the bootable volume by cloning from it (with increased size). Also, I'd like to receive the status updates from DV, like progress, status, phases, etc.).
So, I'm just using the DataSourceRef for DV, I saw it should be supported in documentation https://github.com/kubevirt/containerized-data-importer/blob/main/doc/cdi-populators.md#using-populators-with-datavolumes. 
Thanks, Dmytro

akalenyu:
> Hi @akalenyu, I'm using volumeimportsources as an image definition: "Let's say image ID". Later multiple image instances (eg. with different storageClasses might be instantiated from it), and finally the kubevirt VM will creates the bootable volume by cloning from it (with increased size). Also, I'd like to receive the status updates from DV, like progress, status, phases, etc.). So, I'm just using the DataSourceRef for DV, I saw it should be supported in documentation https://github.com/kubevirt/containerized-data-importer/blob/main/doc/cdi-populators.md#using-populators-with-datavolumes. Thanks, Dmytro

I see, you don't want to rely on us creating the volumeImportSource on your behalf?

DiMalovanyy:
> > Hi @akalenyu, I'm using volumeimportsources as an image definition: "Let's say image ID". Later multiple image instances (eg. with different storageClasses might be instantiated from it), and finally the kubevirt VM will creates the bootable volume by cloning from it (with increased size). Also, I'd like to receive the status updates from DV, like progress, status, phases, etc.). So, I'm just using the DataSourceRef for DV, I saw it should be supported in documentation https://github.com/kubevirt/containerized-data-importer/blob/main/doc/cdi-populators.md#using-populators-with-datavolumes. Thanks, Dmytro
> 
> I see, you don't want to rely on us creating the volumeImportSource on your behalf?

In my system, software image creation and image instantiation is two different processes. Also I want to have multiple instances (DataVolumes) that references the same VolumeImportSource.

pkg/controller/datavolume/external-population-controller.go
```diff
@@ -217,6 +311,10 @@ func (r *PopulatorReconciler) syncExternalPopulation(log logr.Logger, req reconc
 }
 
 func (r *PopulatorReconciler) updateStatusPhase(pvc *corev1.PersistentVolumeClaim, dataVolumeCopy *cdiv1.DataVolume, event *Event) error {
+	if isCDIVolumePopulator(dataVolumeCopy) {
```
alromeros, line 314: The `checkPopulationRequirements` function might be a bit rusty (we are still checking if the feature gate is enabled though it's been true by default since v1.24), but imo it makes more sense to include this check nested in the `else if supported` condition.

alromeros (commented):
Thanks @DiMalovanyy  for the PR! The overall logic looks fine, but I have a couple of concerns.

In principle I’m not keen to add CDI populator logic into the external-population controller. That controller is intended as a generic way to support external populators, and including CDI-populator specific behavior introduces a level of knowledge that'd be better to avoid considering we have a native way to use those populators. If the use case justifies the change I’m ok with merging, but in my opinion creating a regular dv with an import source is probably just as simple at scale as the approach you are currently using. That said I'd also like to hear what other maintainers think.

Aside from that, I second @awels’ comment about signing the commits. It also looks like there is a leftover commit that should be removed:
https://github.com/kubevirt/containerized-data-importer/pull/3941/commits/fd295deff7f74f7702a36566f323a46336d3ddbb

DiMalovanyy:
> Thanks @DiMalovanyy for the PR! The overall logic looks fine, but I have a couple of concerns.
> 
> In principle I’m not keen to add CDI populator logic into the external-population controller. That controller is intended as a generic way to support external populators, and including CDI-populator specific behavior introduces a level of knowledge that'd be better to avoid considering we have a native way to use those populators. If the use case justifies the change I’m ok with merging, but in my opinion creating a regular dv with an import source is probably just as simple at scale as the approach you are currently using. That said I'd also like to hear what other maintainers think.
> 
> Aside from that, I second @awels’ comment about signing the commits. It also looks like there is a leftover commit that should be removed: [fd295de](https://github.com/kubevirt/containerized-data-importer/commit/fd295deff7f74f7702a36566f323a46336d3ddbb)

Thanks for your comment.
CDI already supports creating a DataVolume with a dataSourceRef pointing to CDI-managed resources such as VolumeImportSource. The import-populator controller correctly handles PVC creation, annotation patching, and other related operations.

However, the external-population-controller does not handle these annotations properly, even though it could. This might be confusing for users who try to use VolumeImportSource or other CDI resources as a dataSourceRef.

If you decide not to merge these changes, I think it should at least be documented that using a DataVolume with a dataSourceRef to CDI-managed resources may not work as expected.

akalenyu:
> > Thanks @DiMalovanyy for the PR! The overall logic looks fine, but I have a couple of concerns.
> > In principle I’m not keen to add CDI populator logic into the external-population controller. That controller is intended as a generic way to support external populators, and including CDI-populator specific behavior introduces a level of knowledge that'd be better to avoid considering we have a native way to use those populators. If the use case justifies the change I’m ok with merging, but in my opinion creating a regular dv with an import source is probably just as simple at scale as the approach you are currently using. That said I'd also like to hear what other maintainers think.
> > Aside from that, I second @awels’ comment about signing the commits. It also looks like there is a leftover commit that should be removed: [fd295de](https://github.com/kubevirt/containerized-data-importer/commit/fd295deff7f74f7702a36566f323a46336d3ddbb)
> 
> Thanks for your comment. CDI already supports creating a DataVolume with a dataSourceRef pointing to CDI-managed resources such as VolumeImportSource. The import-populator controller correctly handles PVC creation, annotation patching, and other related operations.
> 
> However, the external-population-controller does not handle these annotations properly, even though it could. This might be confusing for users who try to use VolumeImportSource or other CDI resources as a dataSourceRef.
> 
> If you decide not to merge these changes, I think it should at least be documented that using a DataVolume with a dataSourceRef to CDI-managed resources may not work as expected.

Was that the use case? I imagined you had your own non-CDI-managed sources

DiMalovanyy:
Hi @akalenyu. Use case is to separate image definition (VolumeImportSource) from the image instance (DataVolume) without the losing the flexibility of the DataVolume (eg. monitoring, progress, status, etc.). I wrote about my use case in comment above https://github.com/kubevirt/containerized-data-importer/pull/3941#pullrequestreview-3441483677. 

I don't use my own CRD's for external populator, I'm trying to use VolumeImportSource as an "image definition".

PR #3943, opened 2025-10-31
Tests: Replace functional tests for virt-default storage class selection with unit tests

alromeros:

This PR removes several functional tests that verify whether the virt-default or the k8s default storage class is selected when creating a PVC/dv.  These tests have historically been flaky and do not cover any functionality that cannot be tested through unit tests.  

They are now replaced with equivalent unit tests.

Fixes #

**Release note**:

```release-note
NONE
```


pkg/apiserver/webhooks/pvc-mutate_test.go
```diff
@@ -137,19 +159,20 @@ var _ = Describe("Mutating PVC Webhook", func() {
...
 			Entry("fail with no default storage classes", false,
-				"PVC spec is missing accessMode and no storageClass to choose profile", &storageClass, &storageProfile),
+				"PVC spec is missing accessMode and no storageClass to choose profile", "", &storageClass, &storageProfile),
 			Entry("fail with default storage classes but with partial storage profile", false,
-				fmt.Sprintf("no accessMode specified in StorageProfile %s", testStorageClassName), &defaultStorageClass, &partialStorageProfile),
-			Entry("succeed with default storage classes and complete storage profile", true, "", &defaultStorageClass, &storageProfile),
+				fmt.Sprintf("no accessMode specified in StorageProfile %s", testStorageClassName), testStorageClassName, &defaultStorageClass, &partialStorageProfile),
+			Entry("succeed with default storage classes and complete storage profile", true, "", testStorageClassName, &defaultStorageClass, &storageProfile),
+			Entry("choose virt default storage class over default", true, "", virtDefaultStorageClassName, &defaultStorageClass, &virtDefaultStorageClass, &storageProfile, &virtDefaultStorageProfile),
```
alromeros, line 175: @arnongilboa just noticed the func test behavior was actually testing the opposite (k8s default is used with PVC api). However I'm not seeing that logic anywhere when checking the code and the mutating webhook actually chooses virt-default. Am I missing something?

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -2047,6 +2075,32 @@ func newImportDataVolumeWithPvc(name string, pvc *corev1.PersistentVolumeClaimSp
 	}
 }
 
+func newImportDataVolumeWithStorage(name string, storage *cdiv1.StorageSpec) *cdiv1.DataVolume {
```
Acedus: I believe we can use the existing https://github.com/kubevirt/containerized-data-importer/blob/6de420bb507f2105f2522bce56b23ed8f672ad44/pkg/controller/datavolume/util_test.go#L450

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -1918,7 +1938,11 @@ func dvPhaseTest(reconciler ReconcilerBase, dvc dvController, testDv runtime.Obj
 }
 
 func createStorageSpec() *cdiv1.StorageSpec {
```
Acedus, line 1940: To me this looks like it should be the other way around (i.e., `createStorageSpecWithStoragecClass` should invoke `createStorageSpec` and alter it accordingly).

pkg/controller/datavolume/upload-controller_test.go
```diff
@@ -483,3 +499,20 @@ func createUploadReconcilerWithoutConfig(objects ...client.Object) *UploadReconc
 	}
 	return r
 }
+
+func newUploadDataVolumeWithStorage(name string) *cdiv1.DataVolume {
```
Acedus: Same here.

pkg/controller/datavolume/import-controller_test.go
alromeros: agree, ack!

pkg/controller/datavolume/upload-controller_test.go
alromeros: ack

pkg/apiserver/webhooks/pvc-mutate_test.go
arnongilboa: RenderPvc() calls the common renderPvcSpecVolumeModeAndAccessModesAndStorageClass() which calls GetStorageClassByNameWithVirtFallback(), so I guess it should prefer virt-default.

alromeros:
/unhold
Unholding as all the conversations have been resolved.

PR #3944, opened 2025-11-05
Add InsecureSkipVerify support for ImageIO data source

Hazanel:

This PR adds support for skipping TLS certificate verification when using ImageIO as a data source for DataVolumes. This enables warm migrations from oVirt/RHV providers when "Skip certificate validation" is enabled.

The implementation includes:
- Added InsecureSkipVerify field to DataVolumeSourceImageIO API spec
- Added AnnInsecureSkipVerify annotation for passing the flag to importer pods
- Updated UpdateImageIOAnnotations to set the annotation when enabled
- Modified NewImageioDataSource to accept and use the insecureSkipVerify parameter
- Updated getOvirtClient to use tls.Insecure() when flag is true
- Modified createHTTPClient to configure TLSClientConfig.InsecureSkipVerify
- Updated all test cases to pass the new parameter

**After the fix**:

Warm migrations from oVirt/RHV providers now work correctly when certificate validation is skipped. The CDI ImageIO importer can connect to oVirt ImageIO services without requiring a valid CA certificate in a ConfigMap, as long as the InsecureSkipVerify flag is set to true in the DataVolume spec.

Example DataVolume usage:
```yaml
spec:
  source:
    imageio:
      url: "https://ovirt-engine.example.com/ovirt-engine/api"
      diskId: "disk-uuid"
      secretRef: "ovirt-credentials"
      insecureSkipVerify: true  # Skip TLS certificate validation
```

Jira Ticket: https://issues.redhat.com/browse/CNV-71978

**Release note**:
 DataVolumes can now skip TLS certificate verification when importing from oVirt/RHV ImageIO sources
    by setting `spec.source.imageio.insecureSkipVerify: true`.
```release-note
Allow insecureSkipVerify on imageio importer
```


pkg/importer/http-datasource.go
```diff
@@ -291,26 +291,36 @@ func createCertPool(certDir string) (*x509.CertPool, error) {
 	return certPool, nil
 }
 
-func createHTTPClient(certDir string) (*http.Client, error) {
+func createHTTPClient(certDir string, insecureSkipVerify ...bool) (*http.Client, error) {
```
arnongilboa: You need only a single bool here, so why use variadic param? just to keep the few calls as is?

pkg/importer/imageio-datasource_test.go
```diff
@@ -142,34 +142,78 @@ var _ = Describe("Imageio data source", func() {
...
+	It("NewImageioDataSource should create datasource with InsecureSkipVerify enabled", func() {
+		dp, err := NewImageioDataSource(ts.URL, "", "", "", diskID, "", "", true)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(dp).ToNot(BeNil())
+	})
+
+	It("NewImageioDataSource should fail without cert when InsecureSkipVerify is disabled", func() {
+		newOvirtClientFunc = getOvirtClient
+		dp, err := NewImageioDataSource(ts.URL, "", "", "", diskID, "", "", false)
+		Expect(err).To(HaveOccurred())
```
arnongilboa, line 162: Maybe check for specific error?

arnongilboa:
@Hazanel looks good w/minor comments. Please add release note.

pkg/controller/import-controller.go
```diff
@@ -671,6 +671,11 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 }
 
 func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim, cdiConfig *cdiv1.CDIConfig) (bool, error) {
+	// Check if insecureSkipVerify annotation is set (for ImageIO sources)
+	if insecureSkipVerify, ok := pvc.Annotations[cc.AnnInsecureSkipVerify]; ok && insecureSkipVerify == "true" {
```
awels: Can we check that this is imageio here as well? The skip verify only applies to imageio endpoints. We explicitly set the skip to false for http/s3 endspoints in other parts of this PR.

PR #3946, opened 2025-11-09
Authorize DataImportCron PVC clone based on creator UserInfo

arnongilboa:
Add the `DataImportCron` spec a `CreatedBy` field with JSON-marshaled `UserInfo` of the user who created the `DataImportCron`. The field is set by the mutating webhook and cannot be set by users.

In case of `DataImportCron` with PVC source, the controller checks the creator `ServiceAccount`/`User` is authorized to clone the source PVC.

Assisted by Cursor AI.

**Release note**:
```release-note
Authorize DataImportCron PVC clone based on creator UserInfo
```

### DIFF
--- a/pkg/apiserver/webhooks/dataimportcron-mutate.go
+++ b/pkg/apiserver/webhooks/dataimportcron-mutate.go
@@ -1,0 +1,63 @@
+/*
+ * This file is part of the CDI project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2025 Red Hat, Inc.
+ *
+ */
+
+package webhooks
+
+import (
+	"encoding/json"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	"k8s.io/klog/v2"
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+)
+
+type dataImportCronMutatingWebhook struct{}
+
+func (wh *dataImportCronMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	klog.V(3).Infof("Got AdmissionReview %+v", ar)
+
+	if ar.Request.Operation != admissionv1.Create {
+		return allowedAdmissionResponse()
+	}
+
+	cron := &cdiv1.DataImportCron{}
+	if err := json.Unmarshal(ar.Request.Object.Raw, cron); err != nil {
+		return toAdmissionResponseError(err)
+	}
+
+	modifiedCron := cron.DeepCopy()
+	userInfoStr, err := getUserInfoString(&ar.Request.UserInfo)
+	if err != nil {
+		return toAdmissionResponseError(err)
+	}
+	modifiedCron.Spec.CreatedBy = userInfoStr
+
+	return toPatchResponse(cron, modifiedCron)
+}
+
+func getUserInfoString(ui *authenticationv1.UserInfo) (*string, error) {
+	bs, err := json.Marshal(ui)
+	if err != nil {
+		return nil, err
+	}
+	userInfoStr := string(bs)
+	return &userInfoStr, nil
+}
--- a/pkg/apiserver/webhooks/dataimportcron-validate.go
+++ b/pkg/apiserver/webhooks/dataimportcron-validate.go
@@ -161,5 +161,25 @@ func (wh *dataImportCronValidatingWebhook) validateDataImportCronSpec(request *a
 		return causes
 	}
 
+	if request.Operation == admissionv1.Create {
+		userInfoStr, err := getUserInfoString(&request.UserInfo)
+		if err != nil {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeInternal,
+				Message: fmt.Sprintf("Cannot marshal UserInfo to string: %v", err),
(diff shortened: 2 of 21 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -889,15 +892,73 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+
+func (r *DataImportCronReconciler) CreateSar(sar *authorizationv1.SubjectAccessReview) (*authorizationv1.SubjectAccessReview, error) {
+	if err := r.client.Create(context.TODO(), sar); err != nil {
+		return nil, err
+	}
+	return sar, nil
+}
+
+func (r *DataImportCronReconciler) GetNamespace(name string) (*corev1.Namespace, error) {
+	return nil, nil
```
noamasu: wondering if a reason not to check if namespace exists and return it?

pkg/controller/dataimportcron-controller.go
```diff
@@ -889,15 +892,73 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
+}
+
+func (r *DataImportCronReconciler) GetDataSource(namespace, name string) (*cdiv1.DataSource, error) {
+	return nil, nil
```
noamasu: same question here, do we want to return datasource?

pkg/apiserver/webhooks/dataimportcron-mutate.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/klog/v2"
+
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+)
+
+type dataImportCronMutatingWebhook struct{}
+
+func (wh *dataImportCronMutatingWebhook) Admit(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	klog.V(3).Infof("Got AdmissionReview %+v", ar)
+
```
noamasu, line 36: not sure if this is a must but t might be good to add resource type validation before processing, similar to how `validateDataVolumeResource` is used in the DataVolume mutating webhook's `Admit` function

pkg/apiserver/webhooks/dataimportcron-validate.go
```diff
@@ -161,5 +161,25 @@ func (wh *dataImportCronValidatingWebhook) validateDataImportCronSpec(request *a
...
+		userInfoStr, err := getUserInfoString(&request.UserInfo)
+		if err != nil {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeInternal,
+				Message: fmt.Sprintf("Cannot marshal UserInfo to string: %v", err),
+				Field:   field.Child("CreatedBy").String(),
+			})
+			return causes
+		}
+		if spec.CreatedBy != nil && *userInfoStr != *spec.CreatedBy {
```
noamasu, line 174: nit: wondering if direct string comparison for userinfo is ok here or should we consider comparing unmarshaled structs using reflect.DeepEqual instead?

noamasu (commented):
waiting for others to have a look, but looks good to me! added some nits and questions :)

tests/dataimportcron_test.go
```diff
@@ -831,6 +833,65 @@ var _ = Describe("DataImportCron", Serial, func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			f.AddNamespaceToDelete(sourceNamespace)
+
+			sourceDv = utils.NewDataVolumeWithHTTPImport("source-pvc", "1Gi", fmt.Sprintf(utils.TinyCoreIsoURL, f.CdiInstallNs))
+			sourceDv, err = utils.CreateDataVolumeFromDefinition(f.CdiClient, sourceNamespace.Name, sourceDv)
+			Expect(err).ToNot(HaveOccurred())
+			f.ForceBindPvcIfDvIsWaitForFirstConsumer(sourceDv)
+		})
+
+		It("[test_id:XXXX] should create DataVolume when kubernetes-admin created the DataImportCron", func() {
```
awels: Do we have test_ids for this, or are we waiting on something to generate them, and these are just placeholders?

pkg/controller/dataimportcron-controller.go
arnongilboa: We don't use dataSources in this flow which handles clone from pvc, but maybe we should return (nil, some-error), to prevent future usage. wdyt? @mhenriks

pkg/controller/dataimportcron-controller.go
arnongilboa: `CreateSar()`, `GetNamespace()` & `GetDataSource()` need to be implemented (we pass the  `AuthorizationHelperProxy` interface to `AuthorizeUser()`/`AuthorizeSA()`). Not sure there is a reason to check if the source namespace exists before we create the `SubjectAccessReview` for the source, but if we decide to check it we need to add rbacs for `namespaces` `get`/`list`. wdyt @mhenriks ?

pkg/apiserver/webhooks/dataimportcron-mutate.go
arnongilboa: Can be added although I have no idea how non-DataImportCron CR can get there.

pkg/apiserver/webhooks/dataimportcron-validate.go
arnongilboa: We use `getUserInfoString()` when mutating the `DataImportCron` `CreatedBy`, and validating with the same func that `CreatedBy` actually has the expected value, so I think it's ok.

tests/dataimportcron_test.go
arnongilboa: placeholders

pkg/controller/dataimportcron-controller.go
mhenriks: In certain cases it is okay for the source namespace to not exist (prepopulated or static pv). Not sure they apply here but should probably still implement this function

pkg/controller/dataimportcron-controller.go
mhenriks: You can return an error here if that is not supported but I would hope the webhook catches this case

pkg/apiserver/webhooks/dataimportcron-validate.go
noamasu: got it, thanks!

tests/dataimportcron_test.go
arnongilboa: filled

PR #3970, opened 2025-12-03
Simplify DataImportCron ServiceAccount authorization

arnongilboa:
**What this PR does / why we need it**:
Add `ServiceAccountName` to `DataImportCron` spec, replacing `CreatedBy` which was added in #3946.

In case of `DataImportCron` with PVC source, the controller checks the `ServiceAccount` is authorized to clone the source PVC.

**Release note**:
```release-note
Simplify DataImportCron ServiceAccount authorization
```

PR #3971, opened 2025-12-03
[release-v1.63] Authorize default SA on DataImportCron PVC source

arnongilboa:
In case of `DataImportCron` with PVC source, the controller checks the namespace default `ServiceAccount` is authorized to clone the source PVC.

This is a minimally partial backport of #3970.

**Release note**:
```release-note
Authorize default SA on DataImportCron PVC source
```


pkg/controller/dataimportcron-controller.go
```diff
@@ -889,15 +890,59 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
...
 	}
-
 	dv := r.newSourceDataVolume(dataImportCron, dvName)
+	if allowed, err := r.authorizeCloneDataVolume(dataImportCron, dv); err != nil {
+		return err
+	} else if !allowed {
+		updateDataImportCronCondition(dataImportCron, cdiv1.DataImportCronProgressing, corev1.ConditionFalse,
+			"Not authorized to create DataVolume", notAuthorized)
+		return nil
+	}
```
akalenyu, line 900: I think we're missing a small addition in the handling, otherwise this looks identical to the kv invocation
https://github.com/kubevirt/kubevirt/blob/03f0b4a02e4457cd0f3b7de8c145547d5214ee12/pkg/virt-controller/watch/vm/vm.go#L498-L505

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -1601,22 +1661,43 @@ func createDataImportCronReconcilerWithoutConfig(objects ...runtime.Object) *Dat
...
+	reconciler.client = client
+	reconciler.uncachedClient = client
+}
+
+type fakeSarClient struct {
+	client.Client
+	allowed bool
+}
+
+func (s *fakeSarClient) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
```
akalenyu, line 1675: nice, last I checked this was the way to use reactors with ctrl runtime

pkg/controller/dataimportcron-controller.go
arnongilboa: `cdiv1.ErrNoTokenOkay` is irrelevant for our use case of cloning from existing PVC.

PR #3973, opened 2025-12-08
Fix PVC size inflation during snapshot restore

grandeit:

When restoring a PVC from a `VolumeSnapshot`, CDI was incorrectly adding filesystem overhead to the requested storage size. This caused strict CSI drivers (e.g., NetApp Trident) to reject the restore because the inflated size exceeds what the snapshot can provide.

Snapshot restores are constrained by `VolumeSnapshot.Status.RestoreSize`, and the caller (e.g., KubeVirt snapshot restore) already sets the PVC size appropriately. CDI should not modify it.

This change updates `RenderPvc` to skip size inflation when the PVC has a `VolumeSnapshot` as its `DataSource` or `DataSourceRef`.

This fix is intentionally scoped to the `RenderPvc` webhook path only. The controller path (`pvcFromStorage`) also calls the sizing logic, but standard DataVolumes don't carry the `cdi.kubevirt.io/applyStorageProfile` label during snapshot restore, so they won't trigger the webhook rendering.

The affected use case is KubeVirt VM snapshot restores for PVCs like TPM/EFI volumes, which are created directly (not via `DataVolume`) and do carry the label. KubeVirt already sets the PVC size correctly based on `VolumeSnapshot.Status.RestoreSize`, so CDI should not inflate it.

**Release note**:

```release-note
Fix PVC size inflation during VolumeSnapshot restore that caused strict CSI drivers (e.g., NetApp Trident) to reject the operation.
Note: CDI PVC mutating webhook will no longer wrongly inflate direct restores from volumesnapshot src
```

### DIFF
--- a/pkg/controller/datavolume/util.go
+++ b/pkg/controller/datavolume/util.go
@@ -77,7 +77,32 @@ func RenderPvc(ctx context.Context, client client.Client, pvc *v1.PersistentVolu
 		return renderClonePvcVolumeSizeFromSource(ctx, client, pvc)
 	}
 
+	// Skip size inflation for PVCs restoring from VolumeSnapshots.
+	// The PVC size must not exceed what the snapshot can provide; inflating it
+	// causes strict CSI drivers (e.g., NetApp Trident) to reject the restore.
+	if hasVolumeSnapshotDataSource(pvc) {
+		return nil
+	}
+
 	return renderPvcSpecVolumeSize(client, &pvc.Spec, false, nil)
+}
+
+// hasVolumeSnapshotDataSource returns true if the PVC's DataSource or DataSourceRef
+// references a VolumeSnapshot.
+func hasVolumeSnapshotDataSource(pvc *v1.PersistentVolumeClaim) bool {
+	if pvc.Spec.DataSourceRef != nil &&
+		pvc.Spec.DataSourceRef.APIGroup != nil &&
+		*pvc.Spec.DataSourceRef.APIGroup == snapshotv1.GroupName &&
+		pvc.Spec.DataSourceRef.Kind == "VolumeSnapshot" {
+		return true
+	}
+	if pvc.Spec.DataSource != nil &&
+		pvc.Spec.DataSource.APIGroup != nil &&
+		*pvc.Spec.DataSource.APIGroup == snapshotv1.GroupName &&
+		pvc.Spec.DataSource.Kind == "VolumeSnapshot" {
+		return true
+	}
+	return false
 }
 
 // renderPvcSpec creates a new PVC Spec based on either the dv.spec.pvc or dv.spec.storage section
--- a/pkg/controller/datavolume/util_test.go
+++ b/pkg/controller/datavolume/util_test.go
@@ -145,6 +145,74 @@ var _ = Describe("renderPvcSpecVolumeSize", func() {
 	)
 })
 
+var _ = Describe("RenderPvc", func() {
+	var (
+		scName       = "test"
+		originalSize = resource.MustParse("1Gi")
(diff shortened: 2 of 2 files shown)


pkg/controller/datavolume/util.go
```diff
@@ -77,9 +77,34 @@ func RenderPvc(ctx context.Context, client client.Client, pvc *v1.PersistentVolu
 		return renderClonePvcVolumeSizeFromSource(ctx, client, pvc)
 	}
 
+	// Skip size inflation for PVCs restoring from VolumeSnapshots.
+	// The PVC size must not exceed what the snapshot can provide; inflating it
+	// causes strict CSI drivers (e.g., NetApp Trident) to reject the restore.
+	if hasVolumeSnapshotDataSource(pvc) {
```
akalenyu, line 83: I think this is going to be an issue with PVC sources as well, should we tackle those too while we're at it?

grandeit: Hey.
Thanks for the fast review!
You are right, it will also be a problem for strict CSI drivers if one clones a PVC like this:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: direct-clone
  namespace: default
  labels:
    cdi.kubevirt.io/applyStorageProfile: "true"
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: storage-class
  dataSource:
    kind: PersistentVolumeClaim
    name: source-pvc
```
Wasn't sure about including a fix for this case since I haven't found a scenario where PVCs like this are created automatically/without user intervention in KubeVirt.

But I can easily add a fix for this case as well if you want me to. :)
What do you think?

akalenyu: Actually I think we can leave this for a different PR

akalenyu:
/hold cancel

I originally got the intention wrong thinking this is also an issue for our in house populators, turns out that path
keeps working fine restoring to restoreSize->expanding.

I think there is no use case for inflating a direct volumesnap restore in our webhook
(all the cross volume mode funky stuff has to go through our populators)
so while it is somewhat of an API change we should be fine

If you could just mention it in the release note, that would be great
Something like `CDI PVC mutating webhook will no longer wrongly inflate direct restores from volumesnapshot src`

grandeit:
> If you could just mention it in the release note, that would be great Something like `CDI PVC mutating webhook will no longer wrongly inflate direct restores from volumesnapshot src`

Do you need me to change the PR description for that?

akalenyu:
> > If you could just mention it in the release note, that would be great Something like `CDI PVC mutating webhook will no longer wrongly inflate direct restores from volumesnapshot src`
> 
> Do you need me to change the PR description for that?

I did that for you actually, forgot I could do that

dsanatar:
Thanks for this!
/lgtm

akalenyu:
> @akalenyu: #3973 failed to apply on top of branch "release-v1.62":
> 
> ```
> Applying: Fix PVC size inflation during snapshot restore
> Using index info to reconstruct a base tree...
> M	pkg/controller/datavolume/util.go
> Falling back to patching base and 3-way merge...
> Auto-merging pkg/controller/datavolume/util.go
> CONFLICT (content): Merge conflict in pkg/controller/datavolume/util.go
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> hint: When you have resolved this problem, run "git am --continue".
> hint: If you prefer to skip this patch, run "git am --skip" instead.
> hint: To restore the original branch and stop patching, run "git am --abort".
> hint: Disable this message with "git config advice.mergeConflict false"
> Patch failed at 0001 Fix PVC size inflation during snapshot restore
> ```

looks like applying fails on 1.62 but not 1.63/1.61 :O

PR #3980, opened 2025-12-11
Fix metric names linter to handle recording rules separately

avlitman:
#### Before this PR:
recording rules and metrics sent as metrics to metrics name linter which handled them same
#### After this PR:
recording rules and metrics sent as 2 list (in the same file as before) to metrics name linter which handle them separately
  
Fixes #
jira-ticket: https://issues.redhat.com/browse/CNV-73602

### Why we need it and why it was done in this way
The plan is to add duplicate recording rules with correct name and deprecate the old recording rules after a few versions.
This is what we were asked to do by the openshift observability team that need to be able to distinguish recording rules from metrics.
This linter updat will make sure no new wrong naming recording rules will be added. 

#### The following tradeoffs were made:
To avoid breaking existing rules, the linter (in the monitoring repo) will include an operator-scoped allowlist for current names, any new rule that isn’t on that list will fail CI. this keeps all policy and exceptions in the monitoring repo (no future kubevirt changes), and in a couple of versions, once old names are gone, we just remove the allowlist from monitoring and bump the linter version in kubevirt.
 
#### The following alternatives were considered:
keep testing the rules as metrics like we do now, and fix the linter only when wrong names is deprecated.
since the linter fails also with this solution and to enforces correct naming for all new rules immediately we decided to implement the new linter now and allowlist the existing wrong rules name that will fail the new linter. 

### Release note

```release-note
none
```

### DIFF
--- a/tools/prom-metrics-collector/metrics_json_generator.go
+++ b/tools/prom-metrics-collector/metrics_json_generator.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/kubevirt/monitoring/pkg/metrics/parser"
+	dto "github.com/prometheus/client_model/go"
 	om "github.com/rhobs/operator-observability-toolkit/pkg/operatormetrics"
 
 	cdiClonerMetrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-cloner"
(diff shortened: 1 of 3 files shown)


avlitman:
cc @akalenyu @aglitke

tools/prom-metrics-collector/metrics_json_generator.go
```diff
@@ -57,31 +69,50 @@ func main() {
...
 	rulesList := rules.ListRecordingRules()
 	for _, r := range rulesList {
-		if _, isExcludedMetric := excludedMetrics[r.GetOpts().Name]; !isExcludedMetric {
-			metricFamilies = append(metricFamilies, parser.Metric{
-				Name: r.GetOpts().Name,
-				Help: r.GetOpts().Help,
-				Type: strings.ToUpper(string(r.GetType())),
-			})
+		name := r.GetOpts().Name
+		if _, isExcludedMetric := excludedMetrics[name]; isExcludedMetric {
```
dsanatar, line 91: maybe i'm missing something, but does this map ever get populated?

dsanatar:
Looks pretty good! just had a quick question

tools/prom-metrics-collector/metrics_json_generator.go
avlitman: it can be, we decided to leave it for all operators in case you must ignore some metric for good reason and to have this logic enabled in case it's needed.

akalenyu (commented):
/approve
:pray:

PR #3991, opened 2025-12-23
Add provisioner-aware VolumeSnapshotClass selection and RWO access mode for DataImportCron

noamasu:
Add support for provisioner-specific requirements when creating snapshots and PVCs for DataImportCron. Some provisioners have specific needs:

- GKE Persistent Disk requires snapshot-type: images parameter in VSC for DataImportCron snapshots
- GKE Persistent Disk requires RWO access mode for DataImportCron PVCs

Standard GCP snapshots (using pd.csi.storage.gke.io) are limited to 6 restores per hour per snapshot. Using a VolumeSnapshotClass with `snapshot-type: images` enables unlimited restores, but these snapshots cannot be created from ReadWriteMany (RWX) PVCs. More details can be found [here](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/backup-pd-volume-snapshots#restore-snapshot)

This PR adds provisioner-aware DataImportCron configuration via StorageProfile annotations:

- **StorageProfile Controller**: Automatically detects provisioner requirements and sets:
  - `cdi.kubevirt.io/useReadWriteOnceForDataImportCron`: Signals RWO access mode for DataImportCron PVCs when not explicitly configured.
  Automatically populated for `pd.csi.storage.gke.io/*` driver.
  - `cdi.kubevirt.io/snapshotClassForDataImportCron`: Specifies the VolumeSnapshotClass name (auto-discovers matching VSC with required parameters)

- **DataImportCron Controller**: 
  - Applies RWO access mode to DataVolume PVCs when annotation is set and no access modes are configured (preserves existing configurations)
  - Uses the specified VolumeSnapshotClass from annotation when creating snapshots
  - **edit:** The Controller monitors these annotations for changes and triggers re-creation automatically to ensure the correct VSC is used.

This ensures snapshot creation succeeds without restore rate limits for GKE, while still allowing the final volume to be restored with the desired access mode. The approach is extensible - new provisioner requirements can be added by updating the storage capabilities configuration.

Jira: https://issues.redhat.com/browse/CNV-73302

**Release note**:

```release-note
Add provisioner-aware VolumeSnapshotClass selection and RWO access mode for DataImportCron
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -423,6 +423,16 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			r.log.Info("Snapshot is around even though format switched to PVC, requeueing")
 			return reconcile.Result{RequeueAfter: time.Second}, nil
 		}
+		// Check if snapshot class changed
+		if desiredStorageClass != nil {
+			changed, err := r.handleSnapshotClassChange(ctx, snapshot, desiredStorageClass.Name)
+			if err != nil {
+				return res, err
+			}
+			if changed {
+				return reconcile.Result{RequeueAfter: time.Second}, nil
+			}
+		}
 		// Below k8s 1.29 there's no way to know the source volume mode
 		// Let's at least expose this info on our own snapshots
 		if _, ok := snapshot.Annotations[cc.AnnSourceVolumeMode]; !ok {
@@ -489,7 +499,7 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 			}
 		}
 		if importSucceeded || len(imports) == 0 {
-			if err := r.createImportDataVolume(ctx, dataImportCron); err != nil {
+			if err := r.createImportDataVolume(ctx, dataImportCron, desiredStorageClass); err != nil {
 				return res, err
 			}
 		}
@@ -864,7 +874,7 @@ func updateLastExecutionTimestamp(cron *cdiv1.DataImportCron) error {
 	return nil
 }
 
-func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
+func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, dataImportCron *cdiv1.DataImportCron, desiredStorageClass *storagev1.StorageClass) error {
 	dataSourceName := dataImportCron.Spec.ManagedDataSource
 	digest := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	if digest == "" {
@@ -891,7 +901,13 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
 		}
 	}
 
-	dv := r.newSourceDataVolume(dataImportCron, dvName)
+	storageProfile := &cdiv1.StorageProfile{}
+	if desiredStorageClass != nil {
+		if err := r.client.Get(ctx, types.NamespacedName{Name: desiredStorageClass.Name}, storageProfile); err != nil {
+			return err
+		}
+	}
+	dv := r.newSourceDataVolume(dataImportCron, dvName, storageProfile)
 	if allowed, err := r.authorizeCloneDataVolume(dataImportCron, dv); err != nil {
 		return err
 	} else if !allowed {
@@ -1007,10 +1023,7 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 	if err := r.client.Get(ctx, types.NamespacedName{Name: desiredStorageClass.Name}, storageProfile); err != nil {
 		return err
 	}
-	className, err := cc.GetSnapshotClassForSmartClone(pvc, &desiredStorageClass.Name, storageProfile.Status.SnapshotClass, r.log, r.client, r.recorder)
-	if err != nil {
-		return err
-	}
+
 	desiredSnapshot := &snapshotv1.VolumeSnapshot{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      pvc.Name,
@@ -1024,8 +1037,15 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 			Source: snapshotv1.VolumeSnapshotSource{
 				PersistentVolumeClaimName: &pvc.Name,
 			},
-			VolumeSnapshotClassName: &className,
 		},
+	}
+	// Select VolumeSnapshotClass for boot source snapshot
+	snapshotClassName, err := r.getSnapshotClassForDataImportCron(pvc, storageProfile)
+	if err != nil {
+		return err
+	}
+	if snapshotClassName != "" {
+		desiredSnapshot.Spec.VolumeSnapshotClassName = &snapshotClassName
 	}
 	r.setDataImportCronResourceLabels(dataImportCron, desiredSnapshot)
 	cc.CopyAllowedLabels(pvc.GetLabels(), desiredSnapshot, false)
@@ -1053,6 +1073,39 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 	}
 
 	return nil
+}
+
+func (r *DataImportCronReconciler) handleSnapshotClassChange(ctx context.Context, snapshot *snapshotv1.VolumeSnapshot, storageClassName string) (bool, error) {
+	sp := &cdiv1.StorageProfile{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: storageClassName}, sp); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+
+	desiredVSC, err := r.getSnapshotClassForDataImportCron(nil, sp)
+	if err != nil {
+		return false, err
+	}
+	actualVSC := ""
+	if snapshot.Spec.VolumeSnapshotClassName != nil {
+		actualVSC = *snapshot.Spec.VolumeSnapshotClassName
+	}
+	if desiredVSC == "" || actualVSC == desiredVSC {
+		return false, nil
+	}
+
+	r.log.Info("Snapshot class changed, deleting", "name", snapshot.Name, "from", actualVSC, "to", desiredVSC)
+	if err := r.client.Delete(ctx, snapshot); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+	return true, nil
+}
+
+// getSnapshotClassForDataImportCron returns the VolumeSnapshotClass name to use for DataImportCron snapshots.
+func (r *DataImportCronReconciler) getSnapshotClassForDataImportCron(pvc *corev1.PersistentVolumeClaim, storageProfile *cdiv1.StorageProfile) (string, error) {
+	if vscName := storageProfile.Annotations[cc.AnnSnapshotClassForDataImportCron]; vscName != "" {
+		return vscName, nil
+	}
+	return cc.GetSnapshotClassForSmartClone(pvc, &storageProfile.Name, storageProfile.Status.SnapshotClass, r.log, r.client, r.recorder)
 }
 
 func (r *DataImportCronReconciler) updateDataImportCronSuccessCondition(dataImportCron *cdiv1.DataImportCron, format cdiv1.DataImportCronSourceFormat, snapshot *snapshotv1.VolumeSnapshot) error {
@@ -1367,7 +1420,7 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 			CreateFunc: func(event.TypedCreateEvent[*cdiv1.StorageProfile]) bool { return true },
 			DeleteFunc: func(event.TypedDeleteEvent[*cdiv1.StorageProfile]) bool { return false },
 			UpdateFunc: func(e event.TypedUpdateEvent[*cdiv1.StorageProfile]) bool {
-				return e.ObjectOld.Status.DataImportCronSourceFormat != e.ObjectNew.Status.DataImportCronSourceFormat
+				return dicRelevantFieldsChanged(e.ObjectOld, e.ObjectNew)
 			},
 		},
 	)); err != nil {
@@ -1412,6 +1465,13 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 	}
 
 	return nil
+}
+
+func dicRelevantFieldsChanged(oldSp, newSp *cdiv1.StorageProfile) bool {
+	sourceFormatChanged := oldSp.Status.DataImportCronSourceFormat != newSp.Status.DataImportCronSourceFormat
+	rwoAnnotationChanged := oldSp.Annotations[cc.AnnUseReadWriteOnceForDataImportCron] != newSp.Annotations[cc.AnnUseReadWriteOnceForDataImportCron]
+	snapshotClassChanged := oldSp.Annotations[cc.AnnSnapshotClassForDataImportCron] != newSp.Annotations[cc.AnnSnapshotClassForDataImportCron]
+	return sourceFormatChanged || rwoAnnotationChanged || snapshotClassChanged
 }
 
 // addDefaultStorageClassUpdateWatch watches for default/virt default storage class updates
@@ -1697,7 +1757,7 @@ func (r *DataImportCronReconciler) setJobCommon(cron *cdiv1.DataImportCron, obj 
 	return nil
 }
 
-func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCron, dataVolumeName string) *cdiv1.DataVolume {
+func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCron, dataVolumeName string, storageProfile *cdiv1.StorageProfile) *cdiv1.DataVolume {
 	dv := cron.Spec.Template.DeepCopy()
 	if isCronRegistrySource(cron) {
 		var digestedURL string
@@ -1722,6 +1782,14 @@ func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCro
 	}
 
 	passCronLabelToDv(cron, dv, cc.LabelDynamicCredentialSupport)
+
+	// Apply RWO access mode as default for DataImportCron (from StorageProfile annotation)
+	// Only applies if the DV doesn't already have AccessModes configured
+	if storageProfile != nil && storageProfile.Annotations[cc.AnnUseReadWriteOnceForDataImportCron] == "true" {
+		if dv.Spec.Storage != nil && len(dv.Spec.Storage.AccessModes) == 0 {
+			dv.Spec.Storage.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}
+		}
+	}
 
 	return dv
 }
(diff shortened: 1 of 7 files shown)


pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -207,6 +207,25 @@ var MinimumSupportedPVCSizeByProvisionerKey = map[string]string{
...
+// Some provisioners require specific parameters in the VolumeSnapshotClass for DataImportCron snapshots.
+var DataImportCronSnapshotClassParametersByProvisionerKey = map[string]map[string]string{
+	// https://docs.cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/backup-pd-volume-snapshots#restore-snapshot
+	"pd.csi.storage.gke.io": {
+		"snapshot-type": "images",
+	},
+	"pd.csi.storage.gke.io/hyperdisk": {
+		"snapshot-type": "images",
+	},
+}
```
akalenyu: To me, the fact that this degree of handling is required in CDI screams
that for that provider, there is a need for a separate storage class for cron purposes.

That provider would then represent itself internally as
```go
"pd.csi.storage.gke.io/hyperdisk-crons": {{rwo, block}},
```
somewhat similar to https://github.com/kubevirt/containerized-data-importer/blob/c186a42925b4de4ab582c5bbd913be0ab25441c4/pkg/storagecapabilities/storagecapabilities.go#L412 but I think this can remain vague like "rwo" since that is really the only capability of their golden image storage class

And we would be using the "correct" volumesnapshotclass via `storageProfile.Status.SnapshotClass`
(the special storage class would hint us which one to choose via `snapclass` parameter or something)

akalenyu:
/cc @awels @arnongilboa @Acedus 
I am totally ready to get challenged on this :pray:

pkg/storagecapabilities/storagecapabilities.go
noamasu: Thanks Alex, I completely agree with that approach.. tbh it would eliminate the need for most of these changes.
Trying to avoid adding an extra SC by forcing two VSCs to work with a single SC is too problematic.

we need to see how to reflect that need to create that dedicated SC for golden images.

regardless, working on your suggested changes asap :)

akalenyu: > we need to see how to reflect that need to create that dedicated SC for golden images.

yeah that's the tough bit. I am not against adding convenience APIs under HCO for example if that helps them
(hco.spec field instead of going over each cron and adding a storage class name)

akalenyu:
If we absolutely MUST go this path then I think we should start discussing an API extension for StorageProfiles;
Either we introduce a new overload for `dataImportCronSourceFormat` or a new field entirely, something along the lines of
`dataImportCronSourceSpec`

Acedus:
> If we absolutely MUST go this path then I think we should start discussing an API extension for StorageProfiles; Either we introduce a new overload for `dataImportCronSourceFormat` or a new field entirely, something along the lines of `dataImportCronSourceSpec`

+1, CDI should remain as oblivious as possible to quirks of the storage providers it has to work with, this is no exception.

pkg/controller/dataimportcron-controller.go
```diff
@@ -1697,7 +1725,7 @@ func (r *DataImportCronReconciler) setJobCommon(cron *cdiv1.DataImportCron, obj
 	return nil
 }
 
-func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCron, dataVolumeName string) *cdiv1.DataVolume {
+func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCron, dataVolumeName string, desiredStorageClass *storagev1.StorageClass, storageProfile *cdiv1.StorageProfile) *cdiv1.DataVolume {
```
arnongilboa: why do you need `desiredStorageClass` here?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1063,26 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
...
+func (r *DataImportCronReconciler) getSnapshotClassForDataImportCron(pvc *corev1.PersistentVolumeClaim, storageProfile *cdiv1.StorageProfile, desiredStorageClass *storagev1.StorageClass) (*string, error) {
+	if storageProfile.Annotations != nil {
+		if vscName := storageProfile.Annotations[cc.AnnSnapshotClassForDataImportCron]; vscName != "" {
+			return &vscName, nil
+		}
+	}
+	if storageProfile.Status.SnapshotClass != nil {
+		return storageProfile.Status.SnapshotClass, nil
+	}
+	className, err := cc.GetSnapshotClassForSmartClone(pvc, &desiredStorageClass.Name, nil, r.log, r.client, r.recorder)
```
arnongilboa: You have `desiredStorageClass.Name` in `storageProfile`, so maybe drop `desiredStorageClass` param?

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -207,6 +207,22 @@ var MinimumSupportedPVCSizeByProvisionerKey = map[string]string{
...
+var UseReadWriteOnceForDataImportCronByProvisionerKey = map[string]struct{}{
+	"pd.csi.storage.gke.io":           {},
+	"pd.csi.storage.gke.io/hyperdisk": {},
+}
+
+// SnapshotClassParametersForDataImportCronByProvisionerKey defines required VolumeSnapshotClass parameters for DataImportCron snapshots
+var SnapshotClassParametersForDataImportCronByProvisionerKey = map[string]map[string]string{
+	"pd.csi.storage.gke.io": {
+		"snapshot-type": "images",
+	},
```
arnongilboa: why do we want to cover it for non-hyperdisk?

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -207,6 +207,22 @@ var MinimumSupportedPVCSizeByProvisionerKey = map[string]string{
...
 }
 
+// UseReadWriteOnceForDataImportCronByProvisionerKey is a hash of provisioners which require RWO access mode for DataImportCron PVCs
+var UseReadWriteOnceForDataImportCronByProvisionerKey = map[string]struct{}{
+	"pd.csi.storage.gke.io":           {},
+	"pd.csi.storage.gke.io/hyperdisk": {},
+}
+
+// SnapshotClassParametersForDataImportCronByProvisionerKey defines required VolumeSnapshotClass parameters for DataImportCron snapshots
+var SnapshotClassParametersForDataImportCronByProvisionerKey = map[string]map[string]string{
```
arnongilboa: Maybe generalize it like other maps here with `map[string]func(sc *storagev1.StorageClass) bool`, so it can check not only `parameters`?

pkg/storagecapabilities/storagecapabilities.go
noamasu: according to [Snapshot frequency limits doc](https://docs.cloud.google.com/compute/docs/disks/snapshot-best-practices#freq-create-snaps) Snapshot restore limitation (6 per hour) is for all types, not just hyperdisk.
and this [Limitations doc](https://docs.cloud.google.com/compute/docs/disks/sharing-disks-between-vms#mw-limitations) suggest that its not possible to snapshot RWX pvc if its an image snapshot

pkg/storagecapabilities/storagecapabilities.go
noamasu: it is possible, i want to make sure I undersood what you mean.
you want to make the comparison and return bool?
for that,I also need the VSC to compare, something like:
func(sc *storagev1.StorageClass, vsc *snapshotv1.VolumeSnapshotClass) bool?

akalenyu:
unrelated failure cluster
/retest

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1063,26 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
...
+// getSnapshotClassForDataImportCron returns the VolumeSnapshotClass name to use for DataImportCron snapshots.
+func (r *DataImportCronReconciler) getSnapshotClassForDataImportCron(pvc *corev1.PersistentVolumeClaim, storageProfile *cdiv1.StorageProfile) (*string, error) {
+	if storageProfile.Annotations != nil {
+		if vscName := storageProfile.Annotations[cc.AnnSnapshotClassForDataImportCron]; vscName != "" {
+			return &vscName, nil
+		}
+	}
+	if storageProfile.Status.SnapshotClass != nil {
+		return storageProfile.Status.SnapshotClass, nil
+	}
```
akalenyu: You could pass `storageProfile.Status.SnapshotClass` to `GetSnapshotClassForSmartClone` as we've been doing before, or am I missing something?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1723,6 +1751,16 @@ func (r *DataImportCronReconciler) newSourceDataVolume(cron *cdiv1.DataImportCro
 
 	passCronLabelToDv(cron, dv, cc.LabelDynamicCredentialSupport)
 
+	// Apply RWO access mode as default for DataImportCron (from StorageProfile annotation)
+	// Only applies if the DV doesn't already have AccessModes configured
+	if storageProfile != nil && storageProfile.Annotations != nil && storageProfile.Annotations[cc.AnnUseReadWriteOnceForDataImportCron] == "true" {
```
akalenyu: reading from a nil map is safe (`storageProfile.Annotations != nil`)

pkg/controller/dataimportcron-controller.go
```diff
@@ -891,7 +891,13 @@ func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, d
 		}
 	}
 
-	dv := r.newSourceDataVolume(dataImportCron, dvName)
+	storageProfile := &cdiv1.StorageProfile{}
+	if desiredStorageClass != nil {
+		if err := r.client.Get(ctx, types.NamespacedName{Name: desiredStorageClass.Name}, storageProfile); err != nil {
+			storageProfile = nil
+		}
```
akalenyu, line 908: should probably just requeue on err, doesn't make sense for there to not be a profile

pkg/controller/dataimportcron-controller.go
```diff
@@ -864,7 +864,7 @@ func updateLastExecutionTimestamp(cron *cdiv1.DataImportCron) error {
 	return nil
 }
 
-func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, dataImportCron *cdiv1.DataImportCron) error {
+func (r *DataImportCronReconciler) createImportDataVolume(ctx context.Context, dataImportCron *cdiv1.DataImportCron, desiredStorageClass *storagev1.StorageClass) error {
```
akalenyu, line 877: picked a random line to ask this on. do you care about existing installs or just new ones? i.e. do you want to seamlessly handle someone upgrading with said type of storage and getting all dvs/snaps converted?

it's fine if not, but if you do, that would require some special care

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1063,26 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 	return nil
 }
 
+// getSnapshotClassForDataImportCron returns the VolumeSnapshotClass name to use for DataImportCron snapshots.
+func (r *DataImportCronReconciler) getSnapshotClassForDataImportCron(pvc *corev1.PersistentVolumeClaim, storageProfile *cdiv1.StorageProfile) (*string, error) {
+	if storageProfile.Annotations != nil {
+		if vscName := storageProfile.Annotations[cc.AnnSnapshotClassForDataImportCron]; vscName != "" {
+			return &vscName, nil
+		}
```
akalenyu: probably want to unit test this, especially the ann taking precedence over storageProfile.Status.SnapshotClass

akalenyu (commented):
Looks good!

akalenyu:
@noamasu note linter failure is real

akalenyu:
/retest
both unrelated

pkg/controller/dataimportcron-controller.go
```diff
@@ -382,6 +382,15 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
 	}
 
+	// Handle snapshot class change: delete snapshot and DV/PVC if VSC changed, then requeue
+	if snapshot != nil && format == cdiv1.DataImportCronSourceFormatSnapshot && desiredStorageClass != nil {
+		if shouldRequeue, err := r.handleSnapshotClassChange(ctx, snapshot, desiredStorageClass.Name); err != nil {
+			return res, err
+		} else if shouldRequeue {
+			return reconcile.Result{RequeueAfter: time.Second}, nil
+		}
+	}
```
akalenyu: I believe we can tackle this in the switch statement branches

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1072,60 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
 	return nil
 }
 
+func (r *DataImportCronReconciler) handleSnapshotClassChange(ctx context.Context, snapshot *snapshotv1.VolumeSnapshot, storageClassName string) (bool, error) {
```
akalenyu, line 1078: this is somewhat confusing, can't we just compare actual/desired spec fields?
so, build the dv/snap you're about to create, and compare it to existing in cluster resource
(just for certain fields)

will that make things more complex? this is the usual path k8s controllers take when they reconcile resources
(deepequal on the spec and if needed invoke an UPDATE)

pkg/controller/storageprofile-controller.go
```diff
@@ -285,6 +290,72 @@ func (r *StorageProfileReconciler) reconcileMinimumSupportedPVCSize(sc *storagev
...
+	}
+
+	if desiredClass == "" {
+		delete(sp.Annotations, cc.AnnSnapshotClassForDataImportCron)
+		return
+	}
+
+	if _, exists := sp.Annotations[cc.AnnSnapshotClassForDataImportCron]; exists {
+		return
+	}
```
akalenyu: I missed this on the first revision, won't it cause issues if we're after changing to a different AnnSnapshotClassForDataImportCron? `class-a -> class-b`?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1367,7 +1438,9 @@ func addDataImportCronControllerWatches(mgr manager.Manager, c controller.Contro
 			CreateFunc: func(event.TypedCreateEvent[*cdiv1.StorageProfile]) bool { return true },
 			DeleteFunc: func(event.TypedDeleteEvent[*cdiv1.StorageProfile]) bool { return false },
 			UpdateFunc: func(e event.TypedUpdateEvent[*cdiv1.StorageProfile]) bool {
-				return e.ObjectOld.Status.DataImportCronSourceFormat != e.ObjectNew.Status.DataImportCronSourceFormat
+				return e.ObjectOld.Status.DataImportCronSourceFormat != e.ObjectNew.Status.DataImportCronSourceFormat ||
+					e.ObjectOld.Annotations[cc.AnnUseReadWriteOnceForDataImportCron] != e.ObjectNew.Annotations[cc.AnnUseReadWriteOnceForDataImportCron] ||
+					e.ObjectOld.Annotations[cc.AnnSnapshotClassForDataImportCron] != e.ObjectNew.Annotations[cc.AnnSnapshotClassForDataImportCron]
```
akalenyu: can we extract this out? will be much easier to reason about with a function name and a couple of variables

akalenyu:
/retest
unrelated. nfs lane really a pain. ooms in CI

akalenyu:
@arnongilboa please lgtm if you're happy :pray:

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1073,45 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
...
+	if err := r.client.Get(ctx, types.NamespacedName{Name: storageClassName}, sp); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+
+	desiredVSC, err := r.getSnapshotClassForDataImportCron(nil, sp)
+	if err != nil {
+		return false, err
+	}
+	actualVSC := snapshot.Spec.VolumeSnapshotClassName
+	if desiredVSC == nil || ptr.Equal(actualVSC, desiredVSC) {
```
arnongilboa: why `ptr.Equal()` and not `reflect.DeepEqual()`?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1073,45 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
...
+	r.log.Info("Snapshot class changed, deleting", "name", snapshot.Name, "from", *actualVSC, "to", *desiredVSC)
+	if err := r.client.Delete(ctx, snapshot); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+	return true, nil
+}
+
+// getSnapshotClassForDataImportCron returns the VolumeSnapshotClass name to use for DataImportCron snapshots.
+func (r *DataImportCronReconciler) getSnapshotClassForDataImportCron(pvc *corev1.PersistentVolumeClaim, storageProfile *cdiv1.StorageProfile) (*string, error) {
+	if storageProfile.Annotations != nil {
```
arnongilboa: don't need to check != nil

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -1627,6 +1627,225 @@ var _ = Describe("All DataImportCron Tests", func() {
...
+				dvName := imports[0].DataVolumeName
+
+				dv := &cdiv1.DataVolume{}
+				err = reconciler.client.Get(context.TODO(), dvKey(dvName), dv)
+				Expect(err).ToNot(HaveOccurred())
+				// Verify AccessModes is still empty (not set to RWO)
+				Expect(dv.Spec.Storage.AccessModes).To(BeEmpty())
+			})
+
+			It("Should apply RWO access mode to PVC spec when using PVC instead of Storage", func() {
```
arnongilboa: Missed that point. Are you sure we want this behavior? afaik we shouldn't allow PVC without accessMode and should also not override it when it's set.

pkg/controller/dataimportcron-controller.go
```diff
@@ -1055,6 +1075,39 @@ func (r *DataImportCronReconciler) handleSnapshot(ctx context.Context, dataImpor
...
+
+	desiredVSC, err := r.getSnapshotClassForDataImportCron(nil, sp)
+	if err != nil {
+		return false, err
+	}
+	actualVSC := ""
+	if snapshot.Spec.VolumeSnapshotClassName != nil {
+		actualVSC = *snapshot.Spec.VolumeSnapshotClassName
+	}
+	if desiredVSC == "" || actualVSC == desiredVSC {
```
arnongilboa, line 1092: shouldn't return true when `actualVSC != "" && desiredVSC == ""` ?

noamasu: yes, but why delete a working snapshot when no VSC is available?

arnongilboa: When VSC is not available is says you can't create a snapshot, so not sure about keeping an existing one. @akalenyu ?

akalenyu: yeah it probably won't be able to restore from the volumesnap, but we don't handle this case today prior to this PR. I don't think it's much of a concern if everything converges eventually when it becomes available

pkg/controller/dataimportcron-controller.go
akalenyu: apparently comparing spec vs spec is going to be a little involved so this is fine

akalenyu:
also unrelated

I believe we're waiting for @arnongilboa lgtm

noamasu:
```  [FAIL] Clone Populator tests Clone from Snapshot Fallback to host assisted [It] should finish the clone after creating the source snapshot```
this looks like a cluster-level infrastructure problems..
```[It] [test_id:10033] with snapshot sources```
looks like this one is historically flaky. also setDataImportCronResourceLabels and CopyAllowedLabels calls remain unchanged, i dont see how it is related to my change.

PR #3994, opened 2025-12-24
keystore_test: fix flaky RSA key comparison by replacing DeepEqual

Acedus:
The previous UT relied on reflect.DeepEqual to compare rsa.PrivateKey structs which can be flaky since it compares internal memory state rather than cryptographic equality which is what the test should be doing. This PR replaces DeepEqual with the PrivateKey Equal method which checks whether a key and another key have an equivalent private exponent and primes.

Fixes #

Related failure:  https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3993/pull-cdi-unit-test/2003749757226323968#1:build-log.txt%3A1413

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/keys/keystore_test.go
+++ b/pkg/keys/keystore_test.go
@@ -162,8 +162,6 @@ var _ = Describe("Create Private Key", func() {
 
 		checkActions(actions, client.Actions())
 
-		if !reflect.DeepEqual(privateKey, returnedPrivateKey) {
-			Fail("Keys do not match")
-		}
+		Expect(privateKey.Equal(returnedPrivateKey)).To(BeTrue(), "Keys should match")
 	})
 })


pkg/keys/keystore_test.go
```diff
@@ -162,8 +162,6 @@ var _ = Describe("Create Private Key", func() {
 
 		checkActions(actions, client.Actions())
 
-		if !reflect.DeepEqual(privateKey, returnedPrivateKey) {
-			Fail("Keys do not match")
-		}
+		Expect(privateKey.Equal(returnedPrivateKey)).To(BeTrue())
```
akalenyu: nit: you could keep the error description string 
```suggestion
		Expect(privateKey.Equal(returnedPrivateKey)).To(BeTrue(), "Keys do not match")
```

akalenyu (commented):
/approve
/lgtm
I think you're right

akalenyu (approved):

PR #4004, opened 2026-01-07
use golang auto release notes tool

akalenyu:
no bash, no shell commands, just go + added auto release note generation

Fixes #

**Release note**:

```release-note
Maintenance: release-note tool refactored bash->go with added functionality
```

### DIFF
--- a/hack/ci/language.sh
+++ b/hack/ci/language.sh
@@ -11,6 +11,7 @@ PHRASES='master|slave|whitelist|blacklist'
 
 VIOLATIONS=$(git grep -iI -E $PHRASES -- \
     ':!vendor' \
+    ':!**/vendor/**' \
     ':!cluster-up' \
     ':!cluster-sync' \
     ':!*generated*' \


akalenyu:
/hold
go.work stuff bit problematic

awels (approved):

tools/release-notes/main.go
```diff
@@ -0,0 +1,440 @@
...
+
+	var formatted []string
+	for contributor, count := range contributorCounts {
+		formatted = append(formatted, fmt.Sprintf("    %d\t%s", count, contributor))
+	}
+
+	return strings.Join(formatted, "\n")
+}
+
+func (r *ReleaseData) getDiffStats() string {
```
Acedus, line 417: The old script also featured a short stat, also this currently doesn't include the first commit parent so we don't actually include the first commit in the diff, let's try something like:

```go
func (r *ReleaseData) getDiffStats() string {
	commits, err := r.getCommitsInRange()
	if err != nil {
		log.Printf("Warning: Failed to get diff stats: %v", err)
		return "changes"
	}

	if len(commits) == 0 {
		return "no changes"
	}

	// Get the first and last commit for stats
	firstCommit := commits[len(commits)-1] // Oldest commit in range
	lastCommit := commits[0]               // Newest commit in range

	parentCommit, err := firstCommit.Parent(0)
	if err != nil {
		return "changes"
	}

	firstTree, err := parentCommit.Tree()
	if err != nil {
		return "changes"
	}

	lastTree, err := lastCommit.Tree()
	if err != nil {
		return "changes"
	}

	changes, err := firstTree.Diff(lastTree)
	if err != nil {
		return "changes"
	}

	patch, err := changes.Patch()
	if err != nil {
		return "changes"
	}

	var insertions, deletions int
	for _, s := range patch.Stats() {
		insertions += s.Addition
		deletions += s.Deletion
	}

	filesChanged := len(changes)
	return fmt.Sprintf("%d files changed, %d insertions(+), %d deletions(-)", filesChanged, insertions, deletions)
}
```

tools/release-notes/main.go
```diff
@@ -0,0 +1,440 @@
...
+		log.Printf("Warning: Failed to get contributor list: %v", err)
+		return ""
+	}
+
+	contributorCounts := make(map[string]int)
+	for _, commit := range commits {
+		contributor := fmt.Sprintf("%s <%s>", commit.Author.Name, commit.Author.Email)
+		contributorCounts[contributor]++
+	}
+
```
Acedus, line 399: Contributors used to be listed in descending order based on their contribution count, consider adding something like:

```go
	keys := make([]string, 0, len(contributorCounts))
	for k := range contributorCounts {
		keys = append(keys, k)
	}

	sort.Slice(keys, func(i, j int) bool {
		if contributorCounts[keys[i]] != contributorCounts[keys[j]] {
			return contributorCounts[keys[i]] > contributorCounts[keys[j]]
		}
		return keys[i] < keys[j]
	})
```

akalenyu: yeah I noticed that and I opted to keep it like that (quantity vs quality and what not)

tools/release-notes/main.go
akalenyu: this actually works as intended according to my tests (we don't miss any PR) with or without this change, maybe the old script was including the last commit of prev?

tools/release-notes/main.go
akalenyu: doing alphabetical now for order but prefer to not involve count in sorting
(kubevirt bot or just quantity over quality considerations)

akalenyu:
/hold cancel
:pray:

PR #4010, opened 2026-01-13
feat: Add checksum validation for HTTP/HTTPS DataVolume sources

### LINKED ISSUE #1520: Add expected_hash to DataVolume
/kind enhancement

qemu images may be garbled en route from an http server to CDI's target. To identify this condition, many servers (e.g. [Fedora](http://mirror.isoc.org.il/pub/fedora/releases/32/Cloud/x86_64/images/Fedora-Cloud-32-1.6-x86_64-CHECKSUM) provide the expected checksum of the image.

I would like to state the expected hash in the DataVolume spec, and have the import fail if any of the provided hashes does not match the imported data.
```
kind: DataVolume
metadata:
  name: "example-import-dv"
spec:
  source:
      http:
         url: "http://mirror.isoc.org.il/pub/fedora/releases/33/Cloud/x86_64/images/Fedora-Cloud-Base-33-1.2.x86_64.raw.xz"
      hashes:
         sha256: 35fa778f5d4830b58f7baf121fff6bd2b52500411c9abc46761b29a690415c3f
         length: 203308980
```

halfcrazy:
Introduces cryptographic hash validation for HTTP/HTTPS import sources to prevent data tampering during download.

**Usage Example:**

```yaml
kind: DataVolume
metadata:
  name: fedora-dv
spec:
  source:
    http:
      url: "https://download.fedoraproject.org/pub/fedora/linux/releases/39/Cloud/x86_64/images/Fedora-Cloud-Base-39-1.5.x86_64.qcow2"
      checksum: "sha256:c5b50f90383e7a1e23386a917eb7c6c82ca6f9c3e39b3c5d3b7c7bb9a4c5e4f3"
  pvc:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 10Gi
```

Introduces cryptographic hash validation for HTTP/HTTPS import sources to prevent data tampering during download.
Fixes #1520

**Release note**:

```release-note
Add hash validation for DataVolume HTTP/HTTPS import sources.
```


akalenyu:
> /retest

Seems like CDI cannot deploy properly, should be able to reproduce locally but let us know if not!
Thanks for the PR! will take a deeper look

halfcrazy:
> > /retest
> 
> Seems like CDI cannot deploy properly, should be able to reproduce locally but let us know if not! Thanks for the PR! will take a deeper look

Yes, reproduce in local cluster via `make cluster-sync`. A bit confusing why shared librariy libnbd broken.
edit: reuse importer pkg introduces libnbd denpendency.

```
[root@iZj6c3ducr707d9q5qyf0eZ containerized-data-importer]# KUBECONFIG=/root/containerized-data-importer/_ci-configs/k8s-1.31/.kubeconfig /root/containerized-data-importer/_ci-configs/k8s-1.31/.kubectl get po -A
NAMESPACE     NAME                                       READY   STATUS             RESTARTS      AGE
cdi           cdi-operator-5486bcc4ff-5xqmc              0/1     CrashLoopBackOff   2 (16s ago)   37s
default       local-volume-provisioner-lm9d8             1/1     Running            0             24m
kube-system   calico-kube-controllers-854b7ff8d6-jr4v8   1/1     Running            0             24m
kube-system   calico-node-xlccc                          1/1     Running            0             24m
kube-system   coredns-5469dfb46b-dbqdl                   1/1     Running            0             24m
kube-system   coredns-5469dfb46b-f86gv                   1/1     Running            0             24m
kube-system   etcd-node01                                1/1     Running            1             24m
kube-system   kube-apiserver-node01                      1/1     Running            1             24m
kube-system   kube-controller-manager-node01             1/1     Running            1             24m
kube-system   kube-proxy-txk85                           1/1     Running            0             24m
kube-system   kube-scheduler-node01                      1/1     Running            1             24m
[root@iZj6c3ducr707d9q5qyf0eZ containerized-data-importer]# KUBECONFIG=/root/containerized-data-importer/_ci-configs/k8s-1.31/.kubeconfig /root/containerized-data-importer/_ci-configs/k8s-1.31/.kubectl -n cdi logs cdi-operator-5486bcc4ff-5xqmc
/usr/bin/cdi-operator: error while loading shared libraries: libnbd.so.0: cannot open shared object file: No such file or directory
```

akalenyu:
> > > /retest
> > 
> > 
> > Seems like CDI cannot deploy properly, should be able to reproduce locally but let us know if not! Thanks for the PR! will take a deeper look
> 
> Yes, reproduce in local cluster via `make cluster-sync`. A bit confusing why shared librariy libnbd broken. edit: reuse importer pkg introduces libnbd denpendency.
> 
> ```
> [root@iZj6c3ducr707d9q5qyf0eZ containerized-data-importer]# KUBECONFIG=/root/containerized-data-importer/_ci-configs/k8s-1.31/.kubeconfig /root/containerized-data-importer/_ci-configs/k8s-1.31/.kubectl get po -A
> NAMESPACE     NAME                                       READY   STATUS             RESTARTS      AGE
> cdi           cdi-operator-5486bcc4ff-5xqmc              0/1     CrashLoopBackOff   2 (16s ago)   37s
> default       local-volume-provisioner-lm9d8             1/1     Running            0             24m
> kube-system   calico-kube-controllers-854b7ff8d6-jr4v8   1/1     Running            0             24m
> kube-system   calico-node-xlccc                          1/1     Running            0             24m
> kube-system   coredns-5469dfb46b-dbqdl                   1/1     Running            0             24m
> kube-system   coredns-5469dfb46b-f86gv                   1/1     Running            0             24m
> kube-system   etcd-node01                                1/1     Running            1             24m
> kube-system   kube-apiserver-node01                      1/1     Running            1             24m
> kube-system   kube-controller-manager-node01             1/1     Running            1             24m
> kube-system   kube-proxy-txk85                           1/1     Running            0             24m
> kube-system   kube-scheduler-node01                      1/1     Running            1             24m
> [root@iZj6c3ducr707d9q5qyf0eZ containerized-data-importer]# KUBECONFIG=/root/containerized-data-importer/_ci-configs/k8s-1.31/.kubeconfig /root/containerized-data-importer/_ci-configs/k8s-1.31/.kubectl -n cdi logs cdi-operator-5486bcc4ff-5xqmc
> /usr/bin/cdi-operator: error while loading shared libraries: libnbd.so.0: cannot open shared object file: No such file or directory
> ```

nice catch!
/test all

pkg/importer/checksum.go
```diff
@@ -0,0 +1,117 @@
...
+			calculatedChecksum,
+		)
+	}
+
+	klog.Infof("Checksum verification passed: %s:%s", cv.algorithm, calculatedChecksum)
+	return nil
+}
+
+// GetAlgorithm returns the hash algorithm being used
+func (cv *ChecksumValidator) GetAlgorithm() string {
```
Acedus: nit: Getters are generally considered not idiomatic in Go: https://go.dev/doc/effective_go#Getters

pkg/importer/checksum_test.go
```diff
@@ -0,0 +1,228 @@
+/*
```
Acedus, line 1: Please conform to the rest of the repo and use Ginkgo to write the tests.

pkg/util/checksum/checksum.go
```diff
@@ -0,0 +1,106 @@
...
+	// AlgorithmSHA256 represents SHA-256 hash algorithm
+	AlgorithmSHA256 = "sha256"
+	// AlgorithmSHA512 represents SHA-512 hash algorithm
+	AlgorithmSHA512 = "sha512"
+)
+
+// ValidateFormat validates the format of a checksum string
+// Format: "algorithm:hash", e.g., "sha256:abc123..." or "md5:def456..."
+// Returns nil if valid or empty, error otherwise
+func ValidateFormat(checksumStr string) error {
```
Acedus: This function is redundant as you already check whether the provided `checksumStr` is empty before calling it.

pkg/controller/common/util.go
```diff
@@ -87,6 +87,8 @@ const (
 	AnnPodRestarts = AnnAPIGroup + "/storage.pod.restarts"
 	// AnnPodSchedulable is a PVC annotation that tells if the Pod is schedulable or not
 	AnnPodSchedulable = AnnAPIGroup + "/storage.pod.schedulable"
+	// AnnImportFatalError is a PVC annotation that indicates a fatal import error that should not be retried
```
Acedus, line 90: This sets an odd precedence IMO, we already have several failure paths for an import, none of them completely halts the import retry. I don't think it's necessarily a bad thing to have something like this, but I think it's out of scope for this PR.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -253,6 +253,11 @@ type DataVolumeSourceHTTP struct {
 	// SecretExtraHeaders is a list of Secret references, each containing an extra HTTP header that may include sensitive information
 	// +optional
 	SecretExtraHeaders []string `json:"secretExtraHeaders,omitempty"`
+	// Checksum is the expected checksum of the file. Format: "algorithm:hash", e.g., "sha256:1234abcd..." or "md5:5678efgh..."
+	// Supported algorithms: md5, sha1, sha256, sha512
+	// If specified, the importer will verify the downloaded content matches this checksum
+	// +optional
+	Checksum *string `json:"checksum,omitempty"`
```
Acedus: Why is this a string pointer and not just a string?

pkg/importer/http-datasource.go
```diff
@@ -131,7 +144,12 @@ func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType
 // Info is called to get initial information about the data.
 func (hs *HTTPDataSource) Info() (ProcessingPhase, error) {
 	var err error
-	hs.readers, err = NewFormatReaders(hs.httpReader, hs.contentLength)
+	// Wrap the reader with checksum validator if needed
+	reader := io.ReadCloser(hs.httpReader)
+	if hs.checksumValidator != nil {
+		reader = io.NopCloser(hs.checksumValidator.GetReader(hs.httpReader))
```
Acedus: Cool! Consider using `hs.readers.appendReader` instead (it already takes care of promoting to a `ReadCloser`, too).

pkg/importer/checksum.go
```diff
@@ -0,0 +1,122 @@
...
```
Acedus: nit: getters are not go idiomatic, see https://go.dev/doc/effective_go#Getters

pkg/importer/checksum.go
```diff
@@ -0,0 +1,122 @@
...
+func (cv *ChecksumValidator) GetReader(r io.Reader) io.Reader {
+	if cv == nil || cv.hasher == nil {
+		return r
+	}
+	return io.TeeReader(r, cv.hasher)
+}
+
+// Validate checks if the calculated checksum matches the expected checksum
+func (cv *ChecksumValidator) Validate() error {
+	if cv == nil {
```
Acedus: So this while technically correct doesn't really make much sense considering *you've already called* the method, so if the method's receiver is nil, then we wouldn't be able to call it to begin with. Consider removing this occurrence and the rest of them.

pkg/importer/http-datasource_test.go
```diff
@@ -31,8 +34,8 @@ var (
 	diskimageTarFileName    = "cirros.tar"
 	tinyCoreGz              = "tinyCore.iso.gz"
 	tinyCoreXz              = "tinyCore.iso.xz"
-	cirrosData, _           = readFile(cirrosFilePath)
-	diskimageArchiveData, _ = readFile(diskimageTarFileName)
+	cirrosData, _           = readFile(filepath.Join(imageDir, cirrosFileName))
```
Acedus: Why was this changed?

tests/datavolume_test.go
```diff
@@ -454,6 +454,23 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			Expect(err).ToNot(HaveOccurred())
 		}
 
+		It("[rfe_id:XXXX][crit:high][posneg:negative][test_id:XXXX]should fail creating import dv due to invalid checksum format", func() {
```
Acedus: I think the UT is already sufficient to cover this.

Acedus (changes requested):
Thanks @halfcrazy, see inline comments.

pkg/importer/http-datasource.go
halfcrazy: Thanks for the suggestion! However, the checksum validator needs to be at the bottom of the reader stack (wrapping the original httpReader) to calculate the hash of the entire stream. Since appendReader adds to the top of the stack, we need to wrap it before passing to NewFormatReaders.

The checksum provided by users is typically for the **original file** as downloaded from the HTTP server, which may be compressed (e.g., `.gz`, `.xz`). If we use `appendReader` to add the checksum validator after `FormatReaders` processes the stream, it would validate the **decompressed data** instead of the original compressed data, causing checksum mismatches for compressed files."

If you have a better approach (e.g., a way to prepend to the reader stack or modify NewFormatReaders to accept an optional checksum reader), I'd be happy to refactor it!

pkg/importer/http-datasource_test.go
halfcrazy: previous `diskimageTarFileName    = "cirros.tar"` does not contains dir path result `diskimageArchiveData` to empty.

`cirrosData` this line can be reverted i modify just for consistency

tests/datavolume_test.go
halfcrazy: Yes, but the test cases in E2E serve an additional purpose: they can verify that the YAML configurations for the webhook are correct and that the webhook callbacks are properly installed.

pkg/controller/common/util.go
halfcrazy: HTTP checksum verification errors are a special case here. Retrying continuously is meaningless if neither the remote file nor the HTTP checksum in the spec is updated. That said, this does highlight a potential issue: when users update the HTTP checksum, they should delete the corresponding annotation to manually trigger a retry mechanism.

pkg/importer/http-datasource.go
Acedus: Well then why not change the format-readers to include checksum validation support? We have full control over the stack.

pkg/importer/checksum_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"errors"
+	"io"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("ChecksumValidator", func() {
+	DescribeTable("NewChecksumValidator should", func(checksumStr string, wantErr bool, wantAlgo string) {
```
Acedus: Can we consolidate this test and the next one into a single DT?

pkg/importer/format-readers_test.go
```diff
@@ -45,7 +45,7 @@ var _ = Describe("Format Readers", func() {
 		Expect(err).ToNot(HaveOccurred())
```
Acedus, line 47: Can we add some tests with checksumValidator reader?

pkg/util/checksum/checksum_test.go
```diff
@@ -0,0 +1,81 @@
+/*
```
Acedus, line 1: Package uses standard go test instead of ginkgo, please amend.

Acedus (commented):
Looks much better! Please see additional inline comments.

Acedus:
@halfcrazy also please squash your commits; we aim for intentful commits.

tests/datavolume_test.go
Acedus, line 1: I wonder if we should also add another functional test for the actual import, I think it would be cheap enough to make it worthwhile.

Acedus (approved):
Small suggestions but otherwise looks great! Thank you @halfcrazy!

Acedus:
@halfcrazy Linter failure is related. I'm contemplating whether we should support md5 and sha1 TBH, WDYT @akalenyu?

halfcrazy:
> @halfcrazy Linter failure is related. I'm contemplating whether we should support md5 and sha1 TBH, WDYT @akalenyu?

Regarding the linter failure, I suggest at least retaining MD5 due to its widespread use. What’s your thought?

tests/datavolume_test.go
halfcrazy: Could you be more specific? Do you mean `VolumeImport`?

Acedus:
> > @halfcrazy Linter failure is related. I'm contemplating whether we should support md5 and sha1 TBH, WDYT @akalenyu?
> 
> Regarding the linter failure, I suggest at least retaining MD5 due to its widespread use. What’s your thought?

Yes, I'm fine with this.

tests/datavolume_test.go
akalenyu: I'm with @Acedus on this one; UT covers this and we have separate tests that make sure webhooks are functional so we can drop this

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -253,6 +253,11 @@ type DataVolumeSourceHTTP struct {
 	// SecretExtraHeaders is a list of Secret references, each containing an extra HTTP header that may include sensitive information
 	// +optional
 	SecretExtraHeaders []string `json:"secretExtraHeaders,omitempty"`
+	// Checksum is the expected checksum of the file. Format: "algorithm:hash", e.g., "sha256:1234abcd..." or "md5:5678efgh..."
+	// Supported algorithms: md5, sha1, sha256, sha512
+	// If specified, the importer will verify the downloaded content matches this checksum
+	// +optional
+	Checksum string `json:"checksum,omitempty"`
```
akalenyu, line 260: should we use `*string`? I know technically the API conventions allow this
(semantically equivalent unset and "")
but the small cost of nil checks is worth the clarity and defensive futureproof design.
It's the more "correct" k8s convention for optional primitives

pkg/controller/datavolume/import-controller.go
```diff
@@ -288,6 +288,11 @@ func (r *ImportReconciler) updateStatusPhase(pvc *corev1.PersistentVolumeClaim,
 		event.reason = ImportInProgress
 		event.message = fmt.Sprintf(MessageImportInProgress, pvc.Name)
 	case string(corev1.PodFailed):
+		// Only mark as Failed if it's a fatal error (e.g., checksum mismatch).
+		// Otherwise, keep the current phase to allow retries.
+		if pvc.Annotations[cc.AnnImportFatalError] == "true" {
+			dataVolumeCopy.Status.Phase = cdiv1.Failed
```
akalenyu, line 294: Just a cute thought, may delete later; should we propagate the actual checksum to the status and let the controller deduce whether the operation failed? I am going to try to come up with pros/cons

cmd/cdi-importer/importer.go
```diff
@@ -188,8 +188,25 @@ func handleImport(
 	err := processor.ProcessData()
 
 	scratchSpaceRequired := errors.Is(err, importer.ErrRequiresScratchSpace)
+	checksumMismatch := errors.Is(err, importer.ErrChecksumMismatch)
 	if err != nil && !scratchSpaceRequired {
 		klog.Errorf("%+v", err)
+		// Special-case checksum validation failures: exit 0 to avoid kubelet restart loops
+		// and let the controller read termination message to mark fatal.
+		if checksumMismatch {
```
akalenyu, line 197: Do we want to remove the disk image from the PVC in those cases?

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
halfcrazy: I'm fine with either approach. Using a string type keeps it consistent with other optional fields in this struct, while using a string pointer is more aligned with standard practice. I originally used a string pointer, but Adi asked why I didn't just use a regular string instead.

pkg/controller/datavolume/import-controller.go
halfcrazy: Current status, are you referring to an extra Checksum/Validation type in conditions?
```
    - lastHeartbeatTime: "2026-01-21T04:08:42Z"
      lastTransitionTime: "2026-01-21T04:08:42Z"
      message: 'Unable to process data: Unable to transfer source data to scratch
        space: checksum validation failed: checksum mismatch: expected sha256:0000000000000000000000000000000000000000000000000000000000000000,
        calculated sha256:11d74aa12309da7240f171c140394729bb9b407e8fa3cb52c6dcbf7009352fab'
      reason: ChecksumError
      status: "False"
      type: Running
```

cmd/cdi-importer/importer.go
halfcrazy: I'm not sure about this. Typically, users will wait until the DataVolume (DV) is ready before using it, so the contents within the PVC don't actually take effect until then. If the DV is deleted, the corresponding PVC will be deleted as well.

Currently, if the validation fails, the contents are not deleted.

akalenyu: Yeah I'm suggesting we delete the potentially malicious data

pkg/controller/datavolume/import-controller.go
akalenyu: No like an explicit `dv.status.checksum`. And if that != `dv.spec.checksum`, go to `cdiv1.Failed`. But I guess if the hash exists in the conditions that is good.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
akalenyu: okay let's keep it

cmd/cdi-importer/importer.go
halfcrazy: How about keep pvc and deleting the PV? This is the simplest cleanup method I can think of. PVCs provisioned by DV may retain some state data, and directly deleting them could cause issues.

akalenyu: I don't think we have to do that, I'm only suggesting we whack the file we downloaded as cleanup

halfcrazy:
ping

akalenyu:
/override pull-containerized-data-importer-e2e-istio
/override pull-containerized-data-importer-non-csi-hpp
something off with ci

PR #1993, opened 2021-10-14
Update kubevirtci, add newer cluster-sync provider logic

maya-r:
**Release note**:

```release-note
NONE
```

PR #4012, opened 2026-01-13
tests: stop using /dev/pvc hack

akalenyu:
we used to have issues with mounting it on /pvc
but that is probably all gone now:
https://github.com/kubevirt/containerized-data-importer/pull/1993/commits/24989ea413a86894174dc64d42173564de099c9d

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/pvc.go
+++ b/tests/framework/pvc.go
@@ -76,7 +76,7 @@ func (f *Framework) FindPVC(pvcName string) (*k8sv1.PersistentVolumeClaim, error
 	return utils.FindPVC(f.K8sClient, f.Namespace.Name, pvcName)
 }
 
-// ForceBindPvcIfDvIsWaitForFirstConsumer creates a Pod with the PVC for passed in DV mounted under /dev/pvc, which forces the PVC to be scheduled and bound.
+// ForceBindPvcIfDvIsWaitForFirstConsumer creates a Pod with the PVC for passed in DV mounted under /pvc, which forces the PVC to be scheduled and bound.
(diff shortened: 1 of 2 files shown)


awels (commented):
I think the /dev/pvc vs /pvc is block vs filesystem. At least that is my recollection of why we did /dev

akalenyu:
> I think the /dev/pvc vs /pvc is block vs filesystem. At least that is my recollection of why we did /dev

Yeah you're right I only wanted to get rid of /dev for filesystem volumes. Whoops
/hold

akalenyu:
Looks like we reuse one path for both block/filesystem, I'd like to keep it that way, and /pvc is more legitimate for both isn't it?

Acedus:
> Looks like we reuse one path for both block/filesystem, I'd like to keep it that way, and /pvc is more legitimate for both isn't it?

+1

awels:
That seems weird that we don't care, in one case (block) it will be a device we are writing to and in the other case it is directory where we need to create a disk.img. It seems kind of confusing me to that both types are the same path.

akalenyu:
> That seems weird that we don't care, in one case (block) it will be a device we are writing to and in the other case it is directory where we need to create a disk.img. It seems kind of confusing me to that both types are the same path.

Yeah, I took a stab at different paths but it's going to make it harder to stay agnostic

awels:
Okay, not super bothered by it, I just find it a bit weird.

awels (approved):

PR #4016, opened 2026-01-21
Use random storageclass names in datavolume_test.go (#4015)

### LINKED ISSUE #4015: Use random storageclass names in datavolume_test.go
**What happened**:
Currently the datavolume_test.go uses the same name to create storage classes
in multiple tests. Sometimes it is seen test failures due to storage class not 
being cleaned up (perhaps the test cluster in bad shape).

The tests shouldn't be affected by such situations. Using a unique (random)
name for each storageclass for each test would improve the stability of
the tests because each test wouldn't be affected by a dangling storageclass
from the last test.

The issue happens if the testing cluster is having problem with
storage class deletion. 

**Additional context**:

**Environment**:
- DV specification: N/A
- Install tools: N/A
- Others: N/A

gaohoward:
Currently the datavolume_test.go uses the same name to create storage classes in multiple tests. Sometimes it is seen test failures due to storage class not being cleaned up (perhaps the test cluster in bad shape).

The tests shouldn't be affected by such situations. Using a unique (random) name for each storageclass for each test would improve the stability of the tests because each test wouldn't be affected by a dangling storageclass
from the last test.

This PR changes the test in datavolume_test.go to use a random/unique storageclass name
for each concerned test. 

Fixes #4015

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -31,6 +31,7 @@ import (
 	dvc "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
 	"kubevirt.io/containerized-data-importer/pkg/controller/populators"
 	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
+	"kubevirt.io/containerized-data-importer/pkg/util"
 	"kubevirt.io/containerized-data-importer/pkg/util/naming"
 	"kubevirt.io/containerized-data-importer/tests/framework"
 	"kubevirt.io/containerized-data-importer/tests/utils"
@@ -2308,12 +2309,13 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 
 	Describe("Verify that when the required storage class is missing", Serial, func() {
 		var (
-			testSc *storagev1.StorageClass
-			pvName string
+			testSc     *storagev1.StorageClass
+			testScName string
+			pvName     string
 		)
 		const (
-			testScName  = "test-sc"
-			blankDVName = "blank-source"
+			testScNamePrefix = "test-sc"
+			blankDVName      = "blank-source"
 		)
 
 		updatePV := func(updateFunc func(*v1.PersistentVolume)) {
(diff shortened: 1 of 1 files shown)


tests/datavolume_test.go
```diff
@@ -2308,12 +2309,13 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 
 	Describe("Verify that when the required storage class is missing", Serial, func() {
```
akalenyu, line 2310: So these tests are serial anyway, the storage classes shouldn't be clashing. Is that not what you're seeing?

awels: We don't wait for the cleanup of a namespace to complete before starting the next test. So since the SC is cluster scoped it could conflict if the cleanup of a previous test hasn't gotten to removing the SC yet.

akalenyu: This subsuite has an AfterEach deleting the storage class from the API server

awels: Right but we don't wait to make sure the resources are really gone. At least that is what we did last time I looked.

akalenyu: With storageclasses I am pretty sure it is gone from the API at that point

Anyways I just think you're facing a different issue, possibly one me and @Acedus have faced before
where OCS will just randomly recreate/patch storageclasses it doesn't own.

gaohoward: I'm not sure about what exactly has happened. I saw the storage class get cleaned up with no error in previous test, but then I saw the late test failed at expecting the storageclass to be absent.

`[2026-01-16T23:20:47.706Z]   STEP: verifying no storage class test-sc @ 01/16/26 23:15:55.151`
`[2026-01-16T23:20:47.706Z]   [FAILED] in [It] - /home/jenkins/agent/workspace/build-tier1-tests-artifacts/containerized-data-importer/tests/datavolume_test.go:2439 @ 01/16/26 23:15:55.155`
...
`[2026-01-16T23:20:47.706Z]   STEP: Destroying namespace "cdi-e2e-tests-dv-func-test-rrnwk" for this suite. @ 01/16/26 23:20:47.473`
...
`[2026-01-16T23:20:47.706Z]   [FAILED] Expected an error, got nil`
`[2026-01-16T23:20:47.706Z]   In [It] at: /home/jenkins/agent/workspace/build-tier1-tests-artifacts/containerized-data-importer/tests/datavolume_test.go:2439 @ 01/16/26 23:15:55.155`

as if the sotrage class been recreated itself like @akalenyu said. Something wired happened in the cluster.

PR #4022, opened 2026-01-29
Expose ProcessingPhase in DataVolume Status

### LINKED ISSUE #4021: Expose ProcessingPhase in DataVolume Status

Currently, the `ProcessingPhase` information (defined in the [importer/data-processor.go](https://github.com/kubevirt/containerized-data-importer/blob/6d28b9a11da1171c7bdd7bbf72ed82fb61c6b707/pkg/importer/data-processor.go#L36-L64)) is only available internally within the importer/uploadserver pod and is not exposed to users. This creates a visibility gap, especially during long-running operations like image conversion.

A common scenario is when downloading a QCOW2 image via HTTP:
1. The progress reaches ~99% (download complete) and last for a while
1. Behind the scenes, QEMU conversion is happening, which can take considerable time (especially for large images)
1. Users have no visibility into this conversion phase unless they check the logs inside the importer pod

This lack of visibility makes it difficult for users to understand what's happening with their DataVolume and whether the system is still working or stuck.

Add a new field `populatorPhase` to the DataVolume status that reflects the current processing phase of the importer/uploadserver pod.

The proposed implementation includes:
1. Adding a metric to track the current processing phase in the importer/uploadserver pod
2. Updating an annotation on the PVC with the current phase information
3. Exposing this phase information to `datavolume.status.populatorPhase`

This would allow users to see exactly what phase their import/upload is in.

Example use case:
```
$ kubectl get datavolume my-dv -o jsonpath='{.status.populatorPhase}'
Convert
```

- Get processing phase from the importer/uploadserver pod logs, but this requires additional kubectl commands and is not user-friendly.
- Emitting Kubernetes events when entering a new processing phase. Events could be emitted from the importer/uploadserver pod during phase transitions. However, there is currently no convention for emitting events from importer/uploadserver pods in the codebase. While events would provide more precise timing information compared to polling-based metrics, the proposed solution uses the Prometheus metrics endpoint, which is polled every 2 seconds, meaning short-lived phases may not be captured between polling intervals. Despite this limitation, the metrics-based approach is more consistent with the existing architecture.

**Additional context**:
This issue was raised in #3537, which mentions that while conversion typically doesn't take as long as the download/upload process, it can still take a considerable amount of time for large images. Exposing this phase information would greatly improve user experience and observability.

brandboat:

This PR adds the `populatorPhase` field to the DataVolume status to expose the current processing phase from importer/uploadserver pods, improving observability for users.

Fixes #4021

**Release note**:

```release-note

```

### DIFF
--- a/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
+++ b/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
@@ -308,8 +308,9 @@ type DataVolumeStatus struct {
 	// ClaimName is the name of the underlying PVC used by the DataVolume.
 	ClaimName string `json:"claimName,omitempty"`
 	//Phase is the current phase of the data volume
-	Phase    DataVolumePhase    `json:"phase,omitempty"`
-	Progress DataVolumeProgress `json:"progress,omitempty"`
+	Phase          DataVolumePhase    `json:"phase,omitempty"`
+	Progress       DataVolumeProgress `json:"progress,omitempty"`
+	PopulatorPhase string             `json:"populatorPhase,omitempty"`
 	// RestartCount is the number of times the pod populating the DataVolume has restarted
 	RestartCount int32                 `json:"restartCount,omitempty"`
 	Conditions   []DataVolumeCondition `json:"conditions,omitempty" optional:"true"`
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -5060,6 +5060,9 @@
       "description": "Phase is the current phase of the data volume",
       "type": "string"
      },
+     "populatorPhase": {
+      "type": "string"
+     },
      "progress": {
       "type": "string"
      },
--- a/doc/metrics.md
+++ b/doc/metrics.md
@@ -6,6 +6,7 @@
 | kubevirt_cdi_cr_ready | Metric | Gauge | CDI install ready |
 | kubevirt_cdi_dataimportcron_outdated | Metric | Gauge | DataImportCron has an outdated import |
 | kubevirt_cdi_datavolume_pending | Metric | Gauge | Number of DataVolumes pending for default storage class to be configured |
+| kubevirt_cdi_import_phase_info | Metric | Gauge | The current processing phase of the import |
 | kubevirt_cdi_import_progress_total | Metric | Counter | The import progress in percentage |
 | kubevirt_cdi_openstack_populator_progress_total | Metric | Counter | Progress of volume population |
 | kubevirt_cdi_ovirt_progress_total | Metric | Counter | Progress of volume population |
--- a/pkg/apis/core/v1beta1/openapi_generated.go
+++ b/pkg/apis/core/v1beta1/openapi_generated.go
@@ -18585,6 +18585,12 @@ func schema_pkg_apis_core_v1beta1_DataVolumeStatus(ref common.ReferenceCallback)
 							Format: "",
 						},
 					},
+					"populatorPhase": {
+						SchemaProps: spec.SchemaProps{
+							Type:   []string{"string"},
+							Format: "",
+						},
+					},
 					"restartCount": {
 						SchemaProps: spec.SchemaProps{
 							Description: "RestartCount is the number of times the pod populating the DataVolume has restarted",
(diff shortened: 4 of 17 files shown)


awels:
Okay, went over the PR, and I think I understand what you did. You are using the prometheus metrics to get a current stage which then goes through the normal annotation on pvc -> phase in DV process.

I am not opposed to this, but I think if we are going to mess around in this code it might make sense to do a little more. In particular you mentioned that the conversion step can take some time especially with large images. Would it make sense to modify the metrics a bit to be phase/progress pairs. In particular right now progress is 'data tranfer' progress, but we also can get conversion progress information from `qemu-img convert` by using the `-p` flag when converting. So maybe we can do some tupels of stage and progress. For some stages progress doesn't make sense (like init). But for others like the data-transfer, conversion, resize it does make sense. Then we can modify the progress reporting to report both the stage and the progress of that stage.

brandboat:
@awels, thanks for the comments!

So this direction is OK? As I saw the comment you left here: https://github.com/kubevirt/containerized-data-importer/issues/4021#issuecomment-3824223072

> So maybe we can do some tupels of stage and progress

That was my first thought too! My only worry is the PR getting massive, and redefining progress is a big jump from what it used to be, and I don't want to break anything downstream unexpectedly. If you're okay with the scope creep, I’m happy to implement it.

akalenyu:
> Okay, went over the PR, and I think I understand what you did. You are using the prometheus metrics to get a current stage which then goes through the normal annotation on pvc -> phase in DV process.
> 
> I am not opposed to this, but I think if we are going to mess around in this code it might make sense to do a little more. In particular you mentioned that the conversion step can take some time especially with large images. Would it make sense to modify the metrics a bit to be phase/progress pairs. In particular right now progress is 'data tranfer' progress, but we also can get conversion progress information from `qemu-img convert` by using the `-p` flag when converting. So maybe we can do some tupels of stage and progress. For some stages progress doesn't make sense (like init). But for others like the data-transfer, conversion, resize it does make sense. Then we can modify the progress reporting to report both the stage and the progress of that stage.

Yup agree

I did want to poke more at the motivation for this. Is your scratch space storage class != target storage class?
That would explain extra latency in the conversion

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -308,8 +308,9 @@ type DataVolumeStatus struct {
 	// ClaimName is the name of the underlying PVC used by the DataVolume.
 	ClaimName string `json:"claimName,omitempty"`
 	//Phase is the current phase of the data volume
-	Phase    DataVolumePhase    `json:"phase,omitempty"`
-	Progress DataVolumeProgress `json:"progress,omitempty"`
+	Phase          DataVolumePhase    `json:"phase,omitempty"`
+	Progress       DataVolumeProgress `json:"progress,omitempty"`
+	PopulatorPhase string             `json:"populatorPhase,omitempty"`
```
akalenyu, line 313: Do we want a `printcolumn` for this?

brandboat:
> I did want to poke more at the motivation for this. Is your scratch space storage class != target storage class?
That would explain extra latency in the conversion

@akalenyu, thanks for the comment. Yes, that could be the case, but even for same storage class, for large images the conversion still might take lots of time. Though I know compared to the download/upload process it might seem not that long, user might feel confused about what's going on during this period.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
brandboat: Good idea, thanks!

awels:
> @awels, thanks for the comments!
> 
> So this direction is OK? As I saw the comment you left here: [#4021 (comment)](https://github.com/kubevirt/containerized-data-importer/issues/4021#issuecomment-3824223072)
> 
> > So maybe we can do some tupels of stage and progress
> 
> That was my first thought too! My only worry is the PR getting massive, and redefining progress is a big jump from what it used to be, and I don't want to break anything downstream unexpectedly. If you're okay with the scope creep, I’m happy to implement it.

I think if we are going to do it, we should probably do it right. I don't think it will be massive change. Especially since the number of stages that have actual progress is pretty limited (data transfer, conversion) are probably the two main ones I can think of.

I don't see any issues with the direction you took. My only issue is having two phase fields. Maybe we can combine the two, or if we make a new one that is a tupel, we can deprecate the original one. We will have to see which one makes more sense.

brandboat:
> I don't see any issues with the direction you took. My only issue is having two phase fields. Maybe we can combine the two, or if we make a new one that is a tupel, we can deprecate the original one. We will have to see which one makes more sense.

I personally lean to introduce new fields, e.g.

```
...
status:
  progress:  "99%" # the original one
  stage: # new field
    name: "Convert"
    progress: "10%"
...
```

That won't break anything because we merely add new fields here. 
I'll update the PR later, thanks for your inputs!

awels:
Right, and once we get that in, we can start to deprecate the original field(s). Please feel free to go ahead and make those modifications and we will review. Thanks!

PR #4032, opened 2026-02-06
RWX Capability for HPP Overlay

Dsanatar:
To enable new HPP Overlay support for backend storage PVCs (https://github.com/kubevirt/hostpath-provisioner/pull/483 
https://github.com/kubevirt/hostpath-provisioner-operator/pull/653),  we need to allow for HPP to provision volumes with RWX access mode. When creating storageprofiles for HPP, add new check for param to determine if HPP Overlay is enabled

Fixes #

**Release note**:

```release-note
RWX Capability for HPP Overlay 
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -36,9 +36,10 @@ const (
 // CapabilitiesByProvisionerKey defines default capabilities for different storage classes
 var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	// hostpath-provisioner
-	"kubevirt.io.hostpath-provisioner": {{rwo, file}},
-	"kubevirt.io/hostpath-provisioner": {{rwo, file}},
-	"k8s.io/minikube-hostpath":         {{rwo, file}},
+	ProvisionerHPPCSI:          {{rwo, file}},
+	ProvisionerHPPOverlayCSI:   {{rwx, file}},
+	ProvisionerHPPLegacy:       {{rwo, file}},
+	"k8s.io/minikube-hostpath": {{rwo, file}},
 	// nfs-csi
 	"nfs.csi.k8s.io": {{rwx, file}},
 	"k8s-sigs.io/nfs-subdir-external-provisioner": {{rwx, file}},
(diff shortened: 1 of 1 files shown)


pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -351,6 +303,12 @@ func storageProvisionerKey(sc *storagev1.StorageClass) string {
 }
 
 var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageClass) string{
+	"kubevirt.io.hostpath-provisioner": func(sc *storagev1.StorageClass) string {
```
akalenyu: both `kubevirt.io.hostpath-provisioner` and `kubevirt.io/hostpath-provisioner` existing is a hazard 😆  especially when you used the latter for the overlay provisioner string

maybe extract to variables and use those? HppNonCSI vs HPPCSI? this would minimize the chance for a mixup to like 0.001

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -351,6 +303,12 @@ func storageProvisionerKey(sc *storagev1.StorageClass) string {
...
+		if _, exists := sc.Parameters["overlayCSI"]; exists {
+			return "kubevirt.io/hostpath-provisioner/overlay"
+		}
+		return "kubevirt.io/hostpath-provisioner"
```
akalenyu: same pretty sure you want to return `kubevirt.io.hostpath-provisioner` here so already 1 mixup lol

pkg/storagecapabilities/storagecapabilities.go
dsanatar: ah geez great catch, my eyes must be failing me... just fixed this to only parse the new param from the CSI variant. I don't think we need to account for legacy hpp here

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -38,9 +36,10 @@ const (
 // CapabilitiesByProvisionerKey defines default capabilities for different storage classes
 var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	// hostpath-provisioner
-	"kubevirt.io.hostpath-provisioner": {{rwo, file}},
-	"kubevirt.io/hostpath-provisioner": {{rwo, file}},
-	"k8s.io/minikube-hostpath":         {{rwo, file}},
+	"kubevirt.io.hostpath-provisioner":         {{rwo, file}},
```
akalenyu: should be able to swap a few strings like this one with ProvisionerHPPCSI

PR #4033, opened 2026-02-09
metrics-server: inherit TLS options from CDI TLSSecurityProfile

Acedus:
Previously the metrics-server would be initiated with default TLS configuration. This change makes it so the TLS configuration is updated during runtime for every request according to the CDI TLSSecurityProfile.

Partially addresses https://issues.redhat.com/browse/CNV-78858

**Release note**:

```release-note
fix: cdi-operator and cdi-deployment metrics-server now correctly inherit TLS options from the CDI TLSSecrutiyProfile. 
```

### DIFF
--- a/cmd/cdi-controller/controller.go
+++ b/cmd/cdi-controller/controller.go
@@ -41,6 +41,7 @@ import (
 
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	forklift "kubevirt.io/containerized-data-importer-api/pkg/apis/forklift/v1beta1"
+	cdiclient "kubevirt.io/containerized-data-importer/pkg/client/clientset/versioned"
 	"kubevirt.io/containerized-data-importer/pkg/common"
 	"kubevirt.io/containerized-data-importer/pkg/controller"
 	dvc "kubevirt.io/containerized-data-importer/pkg/controller/datavolume"
@@ -51,6 +52,7 @@ import (
 	"kubevirt.io/containerized-data-importer/pkg/util/cert"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/fetcher"
 	"kubevirt.io/containerized-data-importer/pkg/util/cert/generator"
+	cryptowatch "kubevirt.io/containerized-data-importer/pkg/util/tls-crypto-watch"
 )
 
 const (
@@ -193,6 +195,8 @@ func start() {
 		klog.Fatalf("Unable to get uncached client: %v\n", errors.WithStack(err))
 	}
 
+	managedTLSWatcher := cryptowatch.NewManagedTLSWatcher(cdiclient.NewForConfigOrDie(cfg))
+
 	opts := manager.Options{
 		LeaderElection:             true,
 		LeaderElectionNamespace:    namespace,
@@ -207,6 +211,15 @@ func start() {
 			// See CVE-2023-44487, CVE-2023-39325
 			TLSOpts: []func(*tls.Config){func(c *tls.Config) {
 				c.NextProtos = []string{"http/1.1"}
+				c.GetConfigForClient = func(t *tls.ClientHelloInfo) (*tls.Config, error) {
+					config := c.Clone()
+					if w := managedTLSWatcher.Watcher(); w != nil {
+						cryptoConfig := w.GetCdiTLSConfig()
+						config.CipherSuites = cryptoConfig.CipherSuites
+						config.MinVersion = cryptoConfig.MinVersion
+					}
+					return config, nil
+				}
 			}},
 		},
 	}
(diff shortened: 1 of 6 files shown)


Acedus:
fyi @Barakmor1

pkg/util/tls-crypto-watch/tls-crypto-watch.go
```diff
@@ -57,7 +57,7 @@ type cdiConfigTLSWatcher struct {
 }
 
 // NewCdiConfigTLSWatcher crates a new cdiConfigTLSWatcher
-func NewCdiConfigTLSWatcher(ctx context.Context, cdiClient cdiclient.Interface) (CdiConfigTLSWatcher, error) {
+func NewCdiConfigTLSWatcher(ctx context.Context, cdiClient cdiclient.Interface) (cache.SharedIndexInformer, CdiConfigTLSWatcher, error) {
```
Barakmor1: Out of curiosity, why is the additional informer output needed?

pkg/util/tls-crypto-watch/tls-crypto-watch.go
```diff
@@ -89,16 +89,12 @@ func NewCdiConfigTLSWatcher(ctx context.Context, cdiClient cdiclient.Interface)
...
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	go cdiInformerFactory.Start(ctx.Done())
 
-	klog.V(3).Infoln("Waiting for cache sync")
-	cache.WaitForCacheSync(ctx.Done(), cdiConfigInformer.HasSynced)
-	klog.V(3).Infoln("Cache sync complete")
```
Barakmor1, line 99: +1 for the logs removal

pkg/util/tls-crypto-watch/tls-crypto-watch.go
Acedus: Certificate secrets creation for the apiserver and upload-proxy are dependent on the existence of the CDI CR, previously during initialization both called `NewCdiConfigTLSWatcher` before `NewCertWatcher` which provided the cache sync, having it removed (as neither cdi-operator nor cdi-deployment can make use of `NewCdiConfigTLSWatcher` with it) makes it so the apiserver would fail initialization once and eventually succeed as the secrets get created and populated. This change just ensures that we keep logic as close as possible to the original.

cmd/cdi-controller/controller.go
```diff
@@ -205,9 +214,17 @@ func start() {
 			SecureServing: true,
 			// Disable HTTP/2 to prevent rapid reset vulnerability
 			// See CVE-2023-44487, CVE-2023-39325
-			TLSOpts: []func(*tls.Config){func(c *tls.Config) {
-				c.NextProtos = []string{"http/1.1"}
-			}},
+			TLSOpts: []func(*tls.Config){
+				func(c *tls.Config) {
```
nunnatsa: I think we also need to set the server's `MinVersion` and `CipherSuites`, not only the client

Acedus: Doesn't `GetConfigForClient` take precedence anyway?

Acedus: If you're referring to the lack of default, we can definitely add that (our TLSProfile watcher provides a default during initialization so we can use that).

Barakmor1: I agree with @Acedus see:
https://github.com/golang/go/blob/go1.24.0/src/crypto/tls/handshake_server.go#L160-L168

nunnatsa: OK. It's clear now. Thanks for the explanations @Acedus and @Barakmor1

cmd/cdi-operator/operator.go
```diff
@@ -91,6 +94,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	cdiClient := cdiclient.NewForConfigOrDie(cfg)
+	_, cdiConfigTLSWatcher, err := cryptowatch.NewCdiConfigTLSWatcher(context.Background(), cdiClient)
```
akalenyu: I am a little worried about regular informers/CDIConfig CRD not being present. While it converges I think you could accumulate a nasty backoff over the CRD not being registered.

I know that controller-runtime has this https://github.com/kubevirt/containerized-data-importer/blob/064785f284c5736be202abc1d58242d1d10e9a07/vendor/sigs.k8s.io/controller-runtime/pkg/internal/source/kind.go#L62-L88 and I wonder if we can somehow reuse that?

Acedus: Let's add the watcher as a manager runnable.

akalenyu:
sonar is unhappy about the duplication but I don't think it makes sense to extract the callback

awels:
> sonar is unhappy about the duplication but I don't think it makes sense to extract the callback

I excluded cmd from the duplication checks in sonar. Good thing it is not a voting lane, but next time it won't complain.

akalenyu:
linter error is real fyi

akalenyu:
/retest
CI issues for both

PR #4035, opened 2026-02-10
Enable go multi-module workspace

awels:
Similar to KubeVirt https://github.com/kubevirt/kubevirt/pull/14485 set up go workspace so we don't have the symlink to staging anymore. This will ensure hermeto no longer complains about changes to the modules.txt

Fixes https://issues.redhat.com/browse/CNV-60075

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/config.sh
+++ b/hack/build/config.sh
@@ -29,7 +29,7 @@ FUNC_TEST_PROXY="cdi-func-test-proxy"
 FUNC_TEST_POPULATOR="cdi-func-test-sample-populator"
 
 # update this whenever new builder tag is created
-BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder:2512251744-a38f16d34}
+BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder:2603091904-cadcae0a1}
 
 BINARIES="cmd/${OPERATOR} cmd/${CONTROLLER} cmd/${IMPORTER} cmd/${CLONER} cmd/${APISERVER} cmd/${UPLOADPROXY} cmd/${UPLOADSERVER} cmd/${OPERATOR} tools/${FUNC_TEST_INIT} tools/${FUNC_TEST_REGISTRY_INIT} tools/${FUNC_TEST_BAD_WEBSERVER} tools/${FUNC_TEST_PROXY} tools/${FUNC_TEST_POPULATOR}"
 CDI_PKGS="cmd/ pkg/ test/"
--- a/pkg/operator/resources/crds_generated.go
+++ b/pkg/operator/resources/crds_generated.go
@@ -7,7 +7,7 @@ var CDICRDs map[string]string = map[string]string{
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.17.2
   creationTimestamp: null
   name: cdis.cdi.kubevirt.io
 spec:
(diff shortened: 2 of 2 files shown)


hack/update-codegen.sh
```diff
@@ -144,6 +144,9 @@ openapi-gen \
 ${SCRIPT_ROOT}/bin/openapi-spec-generator >${SCRIPT_ROOT}/api/openapi-spec/swagger.json
 
 echo "************* running controller-gen to generate schema yaml ********************"
+# controller-gen fails on vendored k8s.io/api/core/v1/doc.go due to empty +groupName= marker (see https://github.com/kubernetes/kubernetes/issues)
```
akalenyu: this was supposed to be a link to a particular issue or just the landing page?

go.work
```diff
@@ -0,0 +1,7 @@
+go 1.24.0
```
akalenyu, line 1: I picked this line randomly to start the discussion. I noticed we are missing 2 things from the kubevirt referenced PR:
- bazel file changes (vendor->staging)
- symlink https://github.com/kubevirt/kubevirt/pull/14485/changes/b2bbf1caddca686dba17b2d3674e94b41822265f#r2068133554 (mockgen)

I just want to make sure this is fine

hack/update-codegen.sh
akalenyu: in any case how come we hit this here? api version bump? why not bump controller-gen in that case?

akalenyu:
@Acedus I think you had a PR to kubevirt saving it from some go work consequence?

Acedus:
> @Acedus I think you had a PR to kubevirt saving it from some go work consequence?

Yeah it was a bug in go cmd that I fixed at the time: https://github.com/golang/go/commit/1ebebf1cc1c91faa6e7c32bc987dd42f4f3da2e5 but I see that it only made it in 1.26, might be worth to see if my workaround applies here: https://github.com/kubevirt/kubevirt/pull/14951

hack/update-codegen.sh
awels: yeah you are right, I updated codegen and the marker issue went away.

go.work
awels: So the bazel file changes are related to client-go which we don't use, we use the runtime controller library instead which doesn't have the issue that is being solved. As for the symlink, I think that is fixing a specific kubevirt issue, that I don't think CDI has, so it didn't make sense to do that.

awels:
I just tried make clean with this PR and it worked without issues, maybe I am missing something, but everything appears to work as expected.

Makefile
```diff
@@ -93,7 +93,10 @@ apidocs: ## Generate client-go code (same as 'make generate') and swagger docs
 ##@ Dependency management
 deps-update: gomod-update bazel-generate ## Runs 'go mod tidy' and 'go mod vendor'
 
-deps-verify: deps-update ## Verify dependencies are up to date
+revert-deps-replace:
+	git checkout -- ${PWD}/vendor/k8s.io/api/core/v1/doc.go
```
akalenyu: we don't need this anymore right?

awels: right

vendor/k8s.io/api/core/v1/doc.go
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 // +k8s:deepcopy-gen=package
 // +k8s:protobuf-gen=package
 // +k8s:prerelease-lifecycle-gen=true
-// +groupName=
+// +groupName=core
```
akalenyu: guess you also want this reverted

awels: yes....

akalenyu:
are we missing the controller-gen bump in the builder? https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/4035/pull-cdi-apidocs/2030977819466534912

awels:
Maybe, I guess I am making a new PR that bumps to controller gen version in the builder.

awels:
Create a PR to update the builder https://github.com/kubevirt/containerized-data-importer/pull/4051

akalenyu:
I think were missing a make generate and otherwise should be good to go

awels:
yeah, ran the make generate, should be good now.

PR #4039, opened 2026-02-17
Fix certificate collection

Dsanatar:

It has been seen that when creating a ConfigMap with a certificate with a `.pem` extension, it gets ignored during cert collection which causes issues when trying to pull images from registries. There's two pieces to this, one is how the `containers` library is setting up certificates, it seems to always expect them to have `.crt` extension https://github.com/containers/image/blob/main/pkg/tlsclientconfig/tlsclientconfig.go#L36 The second part is during cert collection within CDI, we follow the same filtering logic. I don't believe this filtering is necessary since the library is already expecting these certs to be in PEM format which is supported by both `.crt` and `.pem` file types. (and in our example 
https://github.com/kubevirt/containerized-data-importer/blob/6f17d206e91bc16d792a9f97aaf24dc9a2b320b8/manifests/example/cert-configmap.yaml#L1-L9
it uses `.pem` extension, so I believe it should be supported)

Modified the logic during cert collection to accept `.pem` files and change the symlink name to have `.crt` extension so it can be processed by the library.

Finally,  there seems to be a bug when `CreateCertificateDir()` is called from the cdi-source-update-poller. It first attempts to collect certs from the proxy cert directory however if the dir doesn't exist, the function fails and we don't run `collectCerts()` for the registry cert dir. I believe this wasn't caught by tests since we always fallback to using the `/cert` directory if the function fails. This seems like strange logic, so added a check to see proxy cert dir exists prior to `collectCerts()` so the function can continue on normally.

Fixes #

**Release note**:

```release-note
Fix cert collection to include pem certs
```

### DIFF
--- a/doc/image-from-registry.md
+++ b/doc/image-from-registry.md
@@ -126,7 +126,7 @@ Create a `ConfigMap`  in the same namespace as the DataVolume containing all cer
 kubectl create configmap my-registry-certs --from-file=my-registry.crt
 ```
 
-The `ConfigMap` may contain multiple entries if necessary.  Key name is irrelevant but should have suffix `.crt`.
+The `ConfigMap` may contain multiple entries if necessary.  Key name is irrelevant but should have suffix `.crt` or `.pem`
 
 Add `certConfigMap` to `DataVolume` spec.
 
--- a/tests/transport_test.go
+++ b/tests/transport_test.go
@@ -80,8 +80,15 @@ var _ = Describe("Transport Tests", func() {
 		}
 
 		if certConfigMap != "" {
-			n, err := utils.CopyConfigMap(f.K8sClient, f.CdiInstallNs, certConfigMap, ns, "", "")
-			Expect(err).ToNot(HaveOccurred())
+			var n string
+			// Special handling for testing .pem certificate files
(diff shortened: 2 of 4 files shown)


pkg/importer/registry-datasource.go
```diff
@@ -215,10 +217,21 @@ func collectCerts(certDir, targetDir, targetPrefix string) error {
...
+		if !strings.HasSuffix(obj.Name(), ".crt") && !strings.HasSuffix(obj.Name(), ".pem") {
+			klog.Warningf("Unable to collect cert: %s Must have file extension .crt or .pem", obj.Name())
 			continue
 		}
-		if err := LinkFile(filepath.Join(certDir, obj.Name()), filepath.Join(targetDir, targetPrefix+obj.Name())); err != nil {
+
+		targetName := targetPrefix + obj.Name()
+		if strings.HasSuffix(obj.Name(), ".pem") {
+			// the containers library is currently filtering out any certs that don't have .crt file extension
+			// https://github.com/containers/image/blob/main/pkg/tlsclientconfig/tlsclientconfig.go#L36
```
akalenyu: might want to link a specific blob since main is going to move

dsanatar: ack

PR #4040, opened 2026-02-19
Use patch instead of update for DataVolume status to reduce conflicts

### LINKED ISSUE #4019: CDI is unable to schedule an import because the target PVC is acted upon by multiple controllers, causing it to transition directly to Bound instead of ImportScheduled
**What happened**:
A DataVolume created by DataVolume template in vm.yaml spec has transitioned quickly into Bound State instead of ImportScheduled state when multiple controllers act simultaneously on the PVC created by DataVolume.  

CDI deployment operator must retry and reconcile so that PVC/DataVolume moves into ImportSchedule state even when multiple controllers act on the PVC managed by CDI. 

1. create a vm.yaml with DataVolumeTemplate 
2. This would created a DataVolume backed by PVC
3. As soon as the PVC created another custom operator modifies/adds a custom annotiation to the PVC
4. Now CDI logs the below error message
   "error":"Operation cannot be fulfilled on datavolumes.cdi.kubevirt.io ... the object has been modified; please apply your changes to the latest version and try again"
5. Finally PVC quickly transitioned into Bound state instead of ImportScheduled state.

**Additional context**:

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): v1.63.1
- Kubernetes version (use `kubectl version`): 1.3.0
- DV specification: 

Image

- Install tools: N/A
- Others: N/A

aglitke:

This PR switches from `Status().Update()` to `Status().Patch()` for DataVolume status updates to reduce conflict errors when multiple controllers act on the same PVC concurrently.

When multiple controllers modify the same PVC, the DataVolume is reconciled multiple times in parallel. This causes status updates to fail with conflict errors because the resourceVersion becomes stale between read and update operations.

The patch-based approach using `MergeFrom` is more resilient to concurrent updates as it applies changes without requiring an exact resourceVersion match, significantly reducing conflict errors.

Fixes #4019

- This change follows the same pattern already used in other parts of the codebase (see `pkg/controller/common/util.go:2211`)
- The fix is minimal and focused - only changing the status update mechanism in `emitEvent()`
- All 307 existing datavolume controller tests pass
- The original `dataVolume` object is used directly (not deep copied) since it's never modified and there's a check to ensure metadata changes aren't allowed

**Release note**:
```release-note
Reduce DataVolume status update conflicts when multiple controllers act on the same resource by using patch instead of update
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -1081,7 +1081,10 @@ func (r *ReconcilerBase) emitEvent(dataVolume *cdiv1.DataVolume, dataVolumeCopy 
 	}
 	// Update status subresource only if changed
 	if !reflect.DeepEqual(dataVolume.Status, dataVolumeCopy.Status) {
-		if err := r.client.Status().Update(context.TODO(), dataVolumeCopy); err != nil {
+		// Use patch instead of update to reduce conflicts when multiple controllers
+		// are acting on the same resource
+		patch := client.MergeFrom(dataVolume)
+		if err := r.client.Status().Patch(context.TODO(), dataVolumeCopy, patch); err != nil {
 			r.log.Error(err, "unable to update datavolume status", "name", dataVolumeCopy.Name)
 			return err
 		}


pkg/controller/datavolume/controller-base.go
```diff
@@ -1081,7 +1081,10 @@ func (r *ReconcilerBase) emitEvent(dataVolume *cdiv1.DataVolume, dataVolumeCopy
 	}
 	// Update status subresource only if changed
 	if !reflect.DeepEqual(dataVolume.Status, dataVolumeCopy.Status) {
-		if err := r.client.Status().Update(context.TODO(), dataVolumeCopy); err != nil {
```
akalenyu, line 1084: I have no issue with changing to patch but I am not fully understanding why update would be failing often here.

We only have one reconciler type handling each DV (according to the `.source`), and the workqueue impl. should somewhat guard against concurrent reconciles of the same object.

My worry is that our issue is elsewhere; maybe re queuing on status updates, and/or the deepequal branch entering when it shouldn't.

rporanki: As per my understanding, in my work environment, completely a different custom operator is also reconciling on this PVC resource. It's mainly adding a label to the PVC object. Due to the other operator reconcile,  reconcile on CDI operator is failing as the object has been modified. This is my understanding. So, I beleive reconcile on the modified object should address this issue

akalenyu: ~~This status subresource update call will not conflict over anything other than other status updates.
So objectmeta changes should not cause conflicts here~~

EDIT: I did not realize the resourceVersion is shared between the entire object

> completely a different custom operator is also reconciling on this PVC resource

So you are saying that an external component is operating on **DataVolume** labels in your environment?

rporanki: True. Although  I am not quite sure, just a thought on this. Probably label change or by some means could cause resource version update on the object and hence the status update failure.

akalenyu: You could double check this by looking up the managed fields, something like
`kubectl get dv <DV_NAME> -o yaml --show-managed-fields`

I just want to make sure there's an external component at play here since having this happen with vanilla CDI essentially means we have a hotloop somewhere or something along those lines

rporanki: Below is the output 
 managedFields:
  - apiVersion: cdi.kubevirt.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:cdi.kubevirt.io/allowClaimAdoption: {}
          f:cdi.kubevirt.io/storage.bind.immediate.requested: {}
          f:cdi.kubevirt.io/storage.usePopulator: {}
          f:storage.dac.nokia.com/max-size-limit: {}
        f:labels:
          .: {}
          f:kubevirt.io/created-by: {}
        f:ownerReferences:
          .: {}
          k:{"uid":"31881214-541f-49a7-9578-1f98aa28254b"}: {}
      f:spec:
        .: {}
        f:pvc:
          .: {}
          f:accessModes: {}
          f:resources:
            .: {}
            f:requests:
              .: {}
              f:storage: {}
          f:storageClassName: {}
        f:source:
          .: {}
          f:registry:
            .: {}
            f:certConfigMap: {}
            f:secretRef: {}
            f:url: {}
    manager: virt-controller
    operation: Update
    time: "2026-02-23T12:21:27Z"
  - apiVersion: cdi.kubevirt.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:claimName: {}
        f:conditions: {}
        f:phase: {}
        f:progress: {}
    manager: cdi-controller
    operation: Update
    subresource: status
    time: "2026-02-23T12:21:28Z"

I only see virt-controller and cdi-controller here. Not sure if this would help

akalenyu: I am not sure that it's a stock virt-controller applying `storage.dac.nokia.com/max-size-limit`.
This seems like the suspect, no? Do you have any idea where this is coming from?

rporanki: That annotation label was present in vm.yaml. I have removed the annotation now and sharing the output again

  managedFields:
  - apiVersion: cdi.kubevirt.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:cdi.kubevirt.io/allowClaimAdoption: {}
          f:cdi.kubevirt.io/storage.bind.immediate.requested: {}
          f:cdi.kubevirt.io/storage.usePopulator: {}
        f:labels:
          .: {}
          f:kubevirt.io/created-by: {}
        f:ownerReferences:
          .: {}
          k:{"uid":"e4aaf1ee-b5c2-4ee2-b648-4dada16e0842"}: {}
      f:spec:
        .: {}
        f:pvc:
          .: {}
          f:accessModes: {}
          f:resources:
            .: {}
            f:requests:
              .: {}
              f:storage: {}
          f:storageClassName: {}
        f:source:
          .: {}
          f:registry:
            .: {}
            f:certConfigMap: {}
            f:secretRef: {}
            f:url: {}
    manager: virt-controller
    operation: Update
    time: "2026-02-23T12:48:19Z"
  - apiVersion: cdi.kubevirt.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:claimName: {}
        f:conditions: {}
        f:phase: {}
        f:progress: {}
    manager: cdi-controller
    operation: Update
    subresource: status
    time: "2026-02-23T12:48:20Z"

akalenyu: according to this output, there is no external entity adding labels. if you want to jump on slack (k8s) I go by the same handle there so we can chat there

rporanki: sure.

PR #4043, opened 2026-02-23
Add filesystem SEEK_HOLE detection to avoid tar performance issues

alromeros:

This PR detects fs support for SEEK_HOLE by creating a test file so we can avoid performance issues in filesystems that don't support it.

Fixes # https://issues.redhat.com/browse/CNV-45701

**Release note**:

```release-note
Add filesystem SEEK_HOLE detection to avoid tar -S performance issues
```

### DIFF
--- a/cmd/cdi-cloner/clone-source.go
+++ b/cmd/cdi-cloner/clone-source.go
@@ -10,9 +10,11 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 
 	"github.com/golang/snappy"
+	"golang.org/x/sys/unix"
 
 	"k8s.io/klog/v2"
 
@@ -144,6 +146,55 @@ func validateMount() {
 	}
 }
 
+// seekHoleSupported probes whether SEEK_HOLE works on the given file.
+// Filesystems without real SEEK_HOLE support (e.g. NFS < 4.2) use the
+// generic VFS fallback which returns file-size, making tar -S think
+// there are no holes. We replicate the same heuristic GNU tar uses:
+// if SEEK_DATA returns 0 and SEEK_HOLE returns file-size on the first
+// call, SEEK_HOLE is not efficiently implemented.
+func seekHoleSupported(path string) bool {
+	f, err := os.Open(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// No disk.img — likely an archive of small files where
+			// raw hole-scanning overhead is negligible.
+			klog.Infof("No %s found, skipping SEEK_HOLE probe", path)
+			return true
+		}
+		klog.Fatalf("Failed to open %s for SEEK_HOLE probe: %v", path, err)
+	}
+	defer f.Close()
+
+	fi, err := f.Stat()
+	if err != nil {
+		klog.Fatalf("Failed to stat %s: %v", path, err)
+	}
+	if fi.Size() == 0 {
+		return true
+	}
+
+	dataOff, dataErr := unix.Seek(int(f.Fd()), 0, unix.SEEK_DATA)
+	if errors.Is(dataErr, unix.ENXIO) {
+		// ENXIO from SEEK_DATA means the file is entirely sparse (no data).
+		// Only a real SEEK_HOLE implementation returns this.
+		return true
+	}
+
+	holeOff, holeErr := unix.Seek(int(f.Fd()), 0, unix.SEEK_HOLE)
+
+	if dataErr != nil || holeErr != nil {
+		klog.Warningf("SEEK_HOLE probe failed on %s: SEEK_DATA=%v, SEEK_HOLE=%v", path, dataErr, holeErr)
+		return false
+	}
+
+	if dataOff == 0 && holeOff == fi.Size() {
+		klog.Infof("SEEK_HOLE not supported on %s (VFS fallback detected), omitting tar -S", path)
+		return false
+	}
+
+	return true
+}
+
 func newTarReader(preallocation bool) (io.ReadCloser, error) {
 	excludeMap := map[string]struct{}{
 		"lost+found": {},
@@ -151,8 +202,9 @@ func newTarReader(preallocation bool) (io.ReadCloser, error) {
 
 	const path = "/usr/bin/tar"
 	args := []string{"cv"}
-	if !preallocation {
-		// -S is used to handle sparse files. It can only be used when preallocation is not requested
+	if !preallocation && seekHoleSupported(filepath.Join(mountPoint, common.DiskImageName)) {
+		// -S is used to handle sparse files. It can only be used when preallocation
+		// is not requested and the filesystem supports efficient SEEK_HOLE.
 		args = append(args, "-S")
 	}
 
--- a/cmd/cdi-cloner/clone-source.go
+++ b/cmd/cdi-cloner/clone-source.go
@@ -30,6 +30,11 @@ var (
 	uploadBytes uint64
 )
 
+// seekFunc is a function variable for unix.Seek, allowing test injection.
+var seekFunc = func(fd int, offset int64, whence int) (int64, error) {
+	return unix.Seek(fd, offset, whence)
+}
+
 type execReader struct {
 	cmd    *exec.Cmd
 	stdout io.ReadCloser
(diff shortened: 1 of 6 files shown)


alromeros:
There are several ways we could approach this fix, I'm choosing this one because it keeps the changes more localized. As @mhenriks suggested previously, we could also detect when the clone contains only a single disk.img and avoid using tar in that case. I have a local implementation of that approach, but it felt too convoluted for what seems like a corner case. That said I'm happy to revisit if you feel strongly about it.

cmd/cdi-cloner/clone-source_test.go
```diff
@@ -29,6 +29,25 @@ var _ = Describe("Prometheus Endpoint", func() {
 	})
 })
 
+var _ = Describe("SEEK_HOLE Detection", func() {
+	It("Should detect SEEK_HOLE support on tmpfs", func() {
+		tmpDir, err := os.MkdirTemp("", "seekhole_test")
+		Expect(err).NotTo(HaveOccurred())
+		defer os.RemoveAll(tmpDir)
+
+		// This should return true on most modern linux systems with tmpfs
```
akalenyu: You could get the tests to be "real" unit tests by injecting os dep/using a variable
(ref https://github.com/kubevirt/containerized-data-importer/pull/3333)
This way you could also fake refusing SEEK_HOLE and test that scenario, which is missing here

akalenyu (commented):
Super cool! did you get a chance to test this with a few real environments?
/cc @noamasu

alromeros:
I've tested this further and found that the source clone pod always mounts filesystem PVCs as read-only. Thus the temporary file cannot be created and the logic always falls back to the non-sparse path. Don't know if there's a way to make this viable, I'll leave the PR open for a while just in case.

akalenyu:
> I've tested this further and found that the source clone pod always mounts filesystem PVCs as read-only. Thus the temporary file cannot be created and the logic always falls back to the non-sparse path. Don't know if there's a way to make this viable, I'll leave the PR open for a while just in case.

we might be able to walk back on the `readonly: true` in container mount level, similarly to https://github.com/kubevirt/containerized-data-importer/pull/2944
but I wonder if there's a better way to detect this. There's a bunch of tar/filesystem/virt forums that could probably guide us in the correct direction

alromeros:
Just updated the PR to try testing SEEK_HOLE on one of the existing files instead of creating a new one. @akalenyu I also added mocked interfaces to better isolate the tests per your suggestion, but I personally think is a bit convoluted and preferred the previous simpler tests. Let me know what you think.

cmd/cdi-cloner/clone-source.go
```diff
@@ -144,23 +161,58 @@ func validateMount() {
...
+		testFile, err := opener.Open(testPath)
+		if err != nil {
+			klog.V(3).Infof("Failed to open file %s for SEEK_HOLE detection: %v, trying next file", f.Name(), err)
+			continue
+		}
+
+		_, err = testFile.Seek(0, unix.SEEK_HOLE)
+		testFile.Close()
+
+		if err != nil {
```
mhenriks: For 0 sized files, err will supposedly be `ENXIO`. For filesystems that don't support `SEEK_HOLE`, `EINVAL` should be returned. So maybe something like this would be better?

```golang
if err != nil {
      if errors.Is(err, unix.EINVAL) {
          klog.V(3).Infof("Filesystem does not support SEEK_HOLE: %v", err)
          return false
      }
      // ENXIO or other error — could be an empty file, try next one
      klog.V(3).Infof("SEEK_HOLE failed on %s: %v, trying next file", f.Name(), err)
      continue
  }
```

alromeros: Cool, thanks! Updated the PR with your suggestion. That said, I'm now wondering if it's better to also skip -S if we hit other non-EINVAL seek errors.

alromeros: I mean, that extra check would only be useful if the clone only contains empty files and at that point the lack of sparsity during tar wouldn’t even be an issue.

cmd/cdi-cloner/clone-source.go
```diff
@@ -144,23 +161,62 @@ func validateMount() {
...
+			if errors.Is(err, unix.EINVAL) {
```
noamasu: After testing this PR against GCNV and vanilla NFS
It looks like NFS v3/v4.1 mounts won't actually return EINVAL here... instead they seem to use the generic SEEK_HOLE fallback which succeeds but always reports the entire file as data

because of that, the -S flag still exists for NFS filesystems (confirmed in my tests)

I wonder if something like creating a small sparse temp file on the volume and verifying SEEK_HOLE actually finds the hole, rather than just checking for EINVAL?

check out
https://man7.org/linux/man-pages/man2/lseek.2.html

> "a filesystem can support the operations by making SEEK_HOLE always return the offset of the end of the file, and making SEEK_DATA always return offset."

akalenyu: > After testing this PR against GCNV and vanilla NFS
It looks like NFS v3/v4.1 mounts won't actually return EINVAL here

Nice catch.. I feel like checking for this is nontrivial and we might want to engage lower level folks for some guidelines. Don't want to end up with something really fragile

> I wonder if something like creating a small sparse temp file on the volume and verifying SEEK_HOLE actually finds the hole, rather than just checking for EINVAL?

I think you are right, but we will have to start mounting the source RW, which is getting into complex territory..

akalenyu (commented):
I think we may be able to reuse the same heuristic as `tar` - https://github.com/akalenyu/containerized-data-importer/commit/0d0e090675d1ab028c510e8a5b6ac2e075f612c6
(I did not test it in a real cloud provider cluster)

Note: any approach where we drop -S will result in fs->fs host clones losing sparseness. Not sure how common those are, and we can tackle them by avoiding tar and using the sparse aware writer on disk.img.

akalenyu:
We will also need the second commit (though I believe it has some bugs) to preserve sparseness:
https://github.com/akalenyu/containerized-data-importer/commits/untar-sparse/

akalenyu:
unrelated, source pvc deletion issue in snap sources

akalenyu:
unrelated

alromeros:
I’m starting to have some doubts about the complexity of this PR. As @akalenyu mentioned, this might be going a bit out of scope for CDI. Until the last commit things felt reasonable, but after adding the tar handling I’m not fully confident I’m covering all edge cases, specifically in the `extractTarEntry`.

As an alternative we can copy the stream into a temporary file and check if a disk.img exists. If it exists, stream it with StreamDataToFile and ignore the other files. If not, fall back to UnArchiveTar as we did before.

akalenyu:
> I’m starting to have some doubts about the complexity of this PR. As @akalenyu mentioned, this might be going a bit out of scope for CDI. Until the last commit things felt reasonable, but after adding the tar handling I’m not fully confident I’m covering all edge cases, specifically in the `extractTarEntry`.
> 
> As an alternative we can copy the stream into a temporary file and check if a disk.img exists. If it exists, stream it with StreamDataToFile and ignore the other files. If not, fall back to UnArchiveTar as we did before.



> I’m starting to have some doubts about the complexity of this PR. As @akalenyu mentioned, this might be going a bit out of scope for CDI. Until the last commit things felt reasonable, but after adding the tar handling I’m not fully confident I’m covering all edge cases, specifically in the `extractTarEntry`.
> 
> As an alternative we can copy the stream into a temporary file and check if a disk.img exists. If it exists, stream it with StreamDataToFile and ignore the other files. If not, fall back to UnArchiveTar as we did before.



> I’m starting to have some doubts about the complexity of this PR. As @akalenyu mentioned, this might be going a bit out of scope for CDI. Until the last commit things felt reasonable, but after adding the tar handling I’m not fully confident I’m covering all edge cases, specifically in the `extractTarEntry`.
> 
> As an alternative we can copy the stream into a temporary file and check if a disk.img exists. If it exists, stream it with StreamDataToFile and ignore the other files. If not, fall back to UnArchiveTar as we did before.

As you know I thought this heuristic was already a bit too involved and low level for CDI, but if the only thing that bothers you is the last commit, I think you can just add a new `x-cdi-content-type` for bad SEEK_HOLE storage and keep the old behavior for the regular one.

pkg/uploadserver/uploadserver.go
```diff
@@ -496,12 +493,72 @@ func cloneProcessor(stream io.ReadCloser, contentType, dest string, preallocate
 	return false, nil
 }
 
-func fileToFileCloneProcessor(stream io.ReadCloser) (bool, error) {
-	defer stream.Close()
-	if err := util.UnArchiveTar(stream, common.ImporterVolumePath); err != nil {
-		return false, errors.Wrapf(err, "error unarchiving to %s", common.ImporterVolumePath)
+func fileToFileCloneProcessor(stream io.ReadCloser, preallocate bool, volumePath, diskImagePath string) (bool, error) {
```
noamasu, line 496: As Alex suggested,

maybe adding that third contentType might be worth it?something like
`FilesystemCloneRawContentType = "filesystem-clone-raw"`
and use it in case `SEEK_HOLE` is not suppored

then we can keep the old `UnArchiveTar` logic for `"filesystem-clone"` and use the new writer for `"filesystem-clone-raw"`

regardless, I'm also fine with keeping it as is (removing UnArchiveTar)

alromeros: Tried adding the third contentType locally, but I might actually prefer the current implementation instead of maintaining a third flow. Leaving it as is for now, feel free to discuss

PR #4044, opened 2026-02-24
Add StorageProfile "Recognized" status condition

arnongilboa:
Explaining if the provisioner or the storage class parameters are not recognized, so `ClaimPropertySets` are not auto-filled and need to be specified by the user.

jira-ticket: https://issues.redhat.com/browse/CNV-78356

**Release note**:
```release-note
Add StorageProfile "Recognized" status condition, explaining if the provisioner or the storage class parameters are not recognized by CDI
```


pkg/controller/storageprofile-controller.go
```diff
@@ -198,6 +203,24 @@ func (r *StorageProfileReconciler) getStorageProfile(sc *storagev1.StorageClass)
...
 
+func (r *StorageProfileReconciler) reconcileConditions(ctx context.Context, sc *storagev1.StorageClass, sp *cdiv1.StorageProfile) {
+	cond := cdiv1.StorageProfileCondition{
+		Type: cdiv1.StorageProfileRecognized,
+	}
+
+	provisionerRecognized, parametersRecognized := storagecapabilities.IsRecognized(sc)
+	switch {
+	case provisionerRecognized && parametersRecognized:
+		updateConditionState(&cond.ConditionState, v1.ConditionTrue, "Provisioner is recognized", reasonRecognizedProvisioner)
```
awels: probably want to make these constants as well. Since a condition has both a `reason` and `message` field. Just call them xxxReason, and xxxMessage

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -259,6 +259,16 @@ var UnsupportedProvisioners = map[string]struct{}{
 	storagehelpers.NotSupportedProvisioner: {},
 }
 
+// IsRecognized checks if the storage class provisioner and parameters are recognized so capabilities are available
+func IsRecognized(sc *storagev1.StorageClass) (bool, bool) {
```
awels: Maybe just make a type string, that maps onto one of the 3 reasons here. Or define that in the types.go. And have a function that returns the type. Then you can use that in the switch in controller. I don't really like bool, bool return type of this function. It gets really hard to tell what the booleans means.

PR #4046, opened 2026-02-24
Use an uncached client for webhooks, only cache PVs/StorageClasses with transform

akalenyu:
Using a cached client for webhooks was never a great idea,
with stale entries being somewhat dangerous or plain and simply blocking.
For example:
```bash
#!/bin/bash
kubectl create -f source-pvc.yaml
kubectl wait --for=condition=Bound pvc/source-pvc --timeout=60s
# clone target PVC goes through the mutating webhook
# a stale cache may not see it's bound, resulting in a rejection
kubectl create -f clone-target-pvc.yaml
```
(with VMs this is going to generate bad events and noise)
The cost of using an uncached client is low, since the only verb we act on is CREATE,
and we are not amplifying the request (1x GETs for 1 creation).

`PersistentVolumes` are the exception; they can be numerous with large specs,
and renderPvcSpecFromAvailablePv relies on a field index that requires a cache.
A dedicated PV-only cache with a transform function is used to keep only the fields needed by the webhook.

Fixes #
https://issues.redhat.com/browse/CNV-80572

Real example from this occuring in tests https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/3799/pull-containerized-data-importer-e2e-ceph/1937767208914522112

**Release note**:

```release-note
BugFix: Use an uncached client for webhooks, cache only PVs with transform
```


cmd/cdi-apiserver/apiserver.go
```diff
@@ -58,6 +62,55 @@ const (
...
+	case *corev1.PersistentVolumeList, *storagev1.StorageClassList, *cdiv1.StorageProfileList:
+		return c.cache.List(ctx, list, opts...)
+	}
+	return c.Client.List(ctx, list, opts...)
+}
+
+// pvTransform replaces each PersistentVolume in the informer cache with a
+// stripped copy that only contains the fields used by renderPvcSpecFromAvailablePv
+// and CheckVolumeSatisfyClaim, reducing the memory footprint.
+func pvTransform(obj interface{}) (interface{}, error) {
```
arnongilboa, line 91: is memory footprint a real issue?

tests/clone-populator_test.go
```diff
@@ -190,16 +190,11 @@ var _ = Describe("Clone Populator tests", func() {
 		pvc := generateTargetPVCWithStrategy(size, vm, strategy, scName)
 		pvc.Spec.AccessModes = nil
 		cc.AddLabel(pvc, common.PvcApplyStorageProfileLabel, "true")
-		Eventually(func() error {
```
arnongilboa, line 193: so this was hiding the issue?

pkg/controller/datavolume/util.go
```diff
@@ -291,15 +291,10 @@ func renderClonePvcVolumeSizeFromSource(ctx context.Context, client client.Clien
 	// If target has the source volume mode and storage class, it can request the source requested volume size.
 	// Otherwise try using the source capacity.
 	volSize := sourcePvc.Spec.Resources.Requests[v1.ResourceStorage]
-	if sourcePvc.Status.Phase != v1.ClaimBound {
-		return errors.Errorf("source pvc is not bound")
-	}
 	if targetVolumeMode != sourceVolumeMode || targetSC.Name != sourceSC.Name {
-		capacity, exists := sourcePvc.Status.Capacity[v1.ResourceStorage]
-		if !exists {
-			return errors.Errorf("source pvc has no status.capacity[storage]")
```
arnongilboa, line 300: nice to see this one deleted!

tests/clone-populator_test.go
akalenyu: yep

pkg/controller/datavolume/util.go
akalenyu: :+1:

cmd/cdi-apiserver/apiserver.go
akalenyu: I think on a "real" cluster with a basically unbounded number of PVs and (potentially) driver data on them, yes, it's a real waste.

From a somewhat real cluster
```
$ oc get pv --show-managed-fields -A -o yaml | wc -c
3039890
$ oc get pv -A | wc -l
886
```

akalenyu: CSI drivers can stuff arbitrary data into pv.spec.csi.volumeAttributes

akalenyu:
/cc @ormergi 
just a reference for a semi cached client if you ever want to go this path in the ipam hook

PR #4050, opened 2026-03-08
Respect pod retain annotation in cross ns host assisted clone

akalenyu:
May be missing something but an admin might have the relevant RBAC
and might want to look up the retained pod
(for themselves or tenants behalf)

Relevant discussion on original PR:
https://github.com/kubevirt/containerized-data-importer/pull/2873/changes#r1313295871

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/populators/clone-populator.go
+++ b/pkg/controller/populators/clone-populator.go
@@ -599,20 +599,9 @@ func shouldSkipCleanup(pvc *corev1.PersistentVolumeClaim) bool {
 	//   - AnnPodRetainAfterCompletion annotation is set to true. This means that the user explicitly wants
 	//     to keep the pods.
 	//   - Clone is host-assisted, which is the only clone type with transfer pods worth debugging.
-	//   - Clone occurs in a single namespace, so we avoid retaining pods in a namespace we don't have right to access.
 	if pvc.Annotations[cc.AnnCloneType] == string(cdiv1.CloneStrategyHostAssisted) &&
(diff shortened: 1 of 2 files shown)


awels:
/retest-required
What is the use case here? I am not sure I understand

akalenyu:
> /retest-required What is the use case here? I am not sure I understand

https://github.com/kubevirt/containerized-data-importer/blob/1ab94ade466d7f2b83e4399a40f6c40103903271/doc/debug.md?plain=1#L9 doesn't work in cross ns host assisted clones, so you can't really debug those

akalenyu:
> lgtm, can we add a UT for this? I don't think we exercise this path besides e2e and it's cheap.

Yep done

akalenyu:
@alromeros @mhenriks need your blessing for this one, might have missed something. original discussion linked

alromeros:
> @alromeros @mhenriks need your blessing for this one, might have missed something. original discussion linked

I’m ok with the changes. A user could potentially end up flooding the source namespace with clone pods without having access to them, but since these can be cleaned up later and there aren't any security concerns, I don’t think there's a strong reason to prevent this behavior. @mhenriks wdyt?

akalenyu:
> > @alromeros @mhenriks need your blessing for this one, might have missed something. original discussion linked
> 
> I’m ok with the changes. A user could potentially end up flooding the source namespace with clone pods without having access to them, but since these can be cleaned up later and there aren't any security concerns, I don’t think there's a strong reason to prevent this behavior. @mhenriks wdyt?

Yeah, they'd have to use the retain annotation though. Anyway this bit me when I tried to debug the restart count on eventually succeeded host assisted pods.

PR #4047, opened 2026-02-25
Update kubevirtci, 1.35 lanes

akalenyu:
**What this PR does / why we need it**:
See commits for details, this was not a smooth ride! Mainly https://github.com/kubevirt/containerized-data-importer/pull/4052

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #3545, opened 2024-11-27
[WIP] wait for service readiness in upload controller

akalenyu:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
upload controller: wait for service readiness
```

PR #4052, opened 2026-03-09
use headless service for upload server

### LINKED ISSUE #3851: Enable Headless Service for CDI UploadProxy to Improve Image Upload Performance
Currently, the cdi-uploadproxy service uses a standard ClusterIP type. 

Modify the CDI UploadProxy service to operate in headless mode (clusterIP: None) while retaining type: ClusterIP. This would:
1.Allow clients to connect directly to backend pods via DNS discovery
2.Eliminate kube-proxy from the data path

> spec:
  clusterIP: None  # Enable headless mode
  ports:
  >- port: 443
    protocol: TCP
    targetPort: 8443
  selector:
    cdi.kubevirt.io: cdi-uploadproxy
  type: ClusterIP  # Maintain current type

akalenyu:
apparently, there's no k8s API that says
"kube-proxy has synced iptables rules for this service on all nodes"
so, pod readiness/healthz (even endpointslices) could all report success,
but the client would still face a `connection refused`
(this is causing unnecessary restarts in the k8s 1.35 bump [PR](https://github.com/kubevirt/containerized-data-importer/pull/4047))

we faced something similar in virtctl too
https://github.com/kubevirt/containerized-data-importer/pull/3545

arguably, a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) was the right fit for us anyway
and cuts the above overhead out of the equation.

Fixes #3851 

Not sure if there's a backwards compatibility consideration in here,
since upload pods are not usually carried over an upgrade

**Release note**:

```release-note
Enhancement: use headless service for upload server
```

### DIFF
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -592,14 +592,12 @@ func (r *UploadReconciler) makeUploadServiceSpec(name string, pvc *corev1.Persis
 			},
 		},
 		Spec: corev1.ServiceSpec{
+			ClusterIP: corev1.ClusterIPNone,
 			Ports: []corev1.ServicePort{
 				{
-					Protocol: "TCP",
-					Port:     443,
-					TargetPort: intstr.IntOrString{
-						Type:   intstr.Int,
-						IntVal: 8443,
-					},
+					Protocol:   corev1.ProtocolTCP,
+					Port:       8443,
+					TargetPort: intstr.FromInt32(8443),
 				},
 			},
 			Selector: map[string]string{
@@ -765,7 +763,7 @@ func UploadPossibleForPVC(pvc *corev1.PersistentVolumeClaim) error {
 // GetUploadServerURL returns the url the proxy should post to for a particular pvc
 func GetUploadServerURL(namespace, pvc, uploadPath string) string {
 	serviceName := createUploadServiceNameFromPvcName(pvc)
-	return fmt.Sprintf("https://%s.%s.svc%s", serviceName, namespace, uploadPath)
+	return fmt.Sprintf("https://%s.%s.svc:8443%s", serviceName, namespace, uploadPath)
 }
 
 // createUploadServiceName returns the name given to upload service shortened if needed
--- a/pkg/controller/upload-controller_test.go
+++ b/pkg/controller/upload-controller_test.go
@@ -867,14 +867,12 @@ func createUploadService(pvc *corev1.PersistentVolumeClaim) *corev1.Service {
 			},
 		},
 		Spec: corev1.ServiceSpec{
+			ClusterIP: corev1.ClusterIPNone,
 			Ports: []corev1.ServicePort{
 				{
-					Protocol: "TCP",
-					Port:     443,
-					TargetPort: intstr.IntOrString{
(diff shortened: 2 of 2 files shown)


akalenyu:
/cc @Acedus @mhenriks 
would especially appreciate thoughts on 

> **Special notes for your reviewer**:
> Not sure if there's a backwards compatibility consideration in here,
> since upload pods are not usually carried over an upgrade

pkg/controller/upload-controller.go
```diff
@@ -592,14 +592,12 @@ func (r *UploadReconciler) makeUploadServiceSpec(name string, pvc *corev1.Persis
...
 			Ports: []corev1.ServicePort{
 				{
-					Protocol: "TCP",
-					Port:     443,
-					TargetPort: intstr.IntOrString{
-						Type:   intstr.Int,
-						IntVal: 8443,
-					},
+					Protocol:   corev1.ProtocolTCP,
+					Port:       8443,
```
Acedus, line 599: Why do we change the exposed port to 8443 instead of 443?

Acedus:
> /cc @Acedus @mhenriks would especially appreciate thoughts on
> 
> > **Special notes for your reviewer**:
> > Not sure if there's a backwards compatibility consideration in here,
> > since upload pods are not usually carried over an upgrade

I'd argue that it shouldn't be problematic since the upload-server Pod is transient and bound to the lifetime of an upload.

pkg/controller/upload-controller.go
akalenyu: this is a no op for headless, there will be no remapping

Acedus: Right, no port translation... I wonder if some users rely on this somehow (although I doubt it due to the proxy).

Acedus (approved):
I think this is good, thanks!

pkg/controller/upload-controller.go
akalenyu: yeah definitely [hyrums law](https://www.hyrumslaw.com/) territory, thought about this too. if anyone did, they can still use their own services

akalenyu:
/cc @0xFelix 
Hopefully this is the time where we can finally make the virtctl image upload tests parallel lol

0xFelix:
@akalenyu Let's hope so! :grin:

mhenriks:
Fine with me. Getting rid of the cluster IP is fine for this case but endpointslices still have to be created and that may not happen right away either

akalenyu:
> Fine with me. Getting rid of the cluster IP is fine for this case but endpointslices still have to be created and that may not happen right away either

Yeah good point. I haven't seen that play out but I agree it's completely possible.
I think this PR stands regardless of that though

akalenyu:
/cherrypick release-v1.64 release-v1.63 release-v1.62 release-v1.61

RamLavi:
/cherry-pick release-v1.64

@akalenyu FYI following our conversation

RamLavi:
> /cherry-pick release-v1.64
> 
> @akalenyu FYI following our conversation

oh sorry was not aware you already cherry-picked. Thanks!!

PR #4053, opened 2026-03-10
report failure earlier for node pull imports

akalenyu:
previously the imagepullerr reporting would get delayed
until the main container stopped, which was wasteful.

don't report running unless all containers in the pod are running

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -315,8 +315,14 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *corev1.Pod, te
 		anno[cc.AnnPodRestarts] = strconv.Itoa(podRestarts)
 	}
 
-	containerState := pod.Status.ContainerStatuses[0].State
-	if containerState.Running != nil {
+	running := true
+	for _, status := range pod.Status.ContainerStatuses {
+		if status.State.Running == nil {
+			running = false
+			break
+		}
+	}
+	if running {
 		anno[prefix] = "true"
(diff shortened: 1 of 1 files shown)


pkg/controller/util.go
```diff
@@ -315,8 +315,14 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *corev1.Pod, te
 		anno[cc.AnnPodRestarts] = strconv.Itoa(podRestarts)
 	}
 
-	containerState := pod.Status.ContainerStatuses[0].State
-	if containerState.Running != nil {
+	running := true
```
Acedus, line 318: How come we don't evaluate on Pod conditions?

akalenyu: we could, I was just not ready to challenge us using containerstate instead
this should be a more minimal change

akalenyu: we'd have to look at `Ready/ContainersReady` since there is no direct `Running` condition
anyway for now, not sure we want to go there

Acedus: > we'd have to look at Ready/ContainersReady

Isn't that sufficient? 

> we could, I was just not ready to challenge us using containerstate instead
this should be a more minimal change

I'd argue that would be a minimal change, too, but we can postpone if you deem it necessary.

akalenyu: yeah I don't want to challenge it for now

PR #4054, opened 2026-03-10
Inherit Storage Class for Scratch Space PVCs

### LINKED ISSUE #3848: Option to define a DataVolume-specific `scratchSpaceStorageClass`
We installed Kubevirt and CDI in a namespace called `kubevirt`.

We created a DataVolume in a separate namespace `restricted-namespace` that uses a namespace-specific storage class and has no permission to use the default storage class `default`:

```yaml
kind: DataVolume
metadata:
  name: my-image-dv
spec:
  source:
    registry:
      url: docker://myrepo.azurecr.io/my/image:latest
  storage:
    accessModes:
      - ReadWriteMany
    storageClassName: namespace-specific-storage-class
    volumeMode: Filesystem
    resources:
      requests:
        storage: 7Gi
```

Now the pod `importer-prime-4faba107-5228-4956-8378-cd9a34b89e98` creates two PVCs: `prime-4faba107-5228-4956-8378-cd9a34b89e98` that uses the storage class `namespace-specific-storage-class` and `prime-4faba107-5228-4956-8378-cd9a34b89e98-scratch` that uses the `default` storage class. But as the namespace does not allow PVCs to use the default storage class, the importing process fails.

We can set the `scratchSpaceStorageClass` in the CDI CR globally ([API reference](https://kubevirt.io/cdi-api-reference/v1.62.0/definitions.html#_v1beta1_cdiconfigspec)), but as we have multiple restricted namespaces with their individual storage classes, this is not an option. So it would be great to have an option to set the `scratchSpaceStorageClass` for each DataVolume individually and override the global value. An alternative would be to have an option to always set the scratch PVC storage class to the one specified in the DataVolume.

Or is there a workaround we're missing?

Dsanatar:
There are some scenarios where we would like scratch space PVCs to use the same storage class as the associated target PVC instead of falling back to the default storage class. This PR allows scratch space PVCs to inherit the storage class from the target PVC if the `config.Spec.ScratchSpaceStorageClass` is not set. 

All this requires is to set the ScratchSpaceStorageClass to be an empty string and the existing code for `GetScratchPvcStorageClass()` will already inherit the storage class from the owner PVC

https://github.com/kubevirt/containerized-data-importer/blob/c3097623efebbbd1ff3248996359126e0471b86d/pkg/controller/util.go#L237-L243

Fixes #3848 

**Release note**:

```release-note
Inherit Storage Class for Scratch Space PVCs instead of using default sc 
```

### DIFF
--- a/pkg/controller/config-controller_test.go
+++ b/pkg/controller/config-controller_test.go
@@ -357,29 +357,28 @@ var _ = Describe("Controller storage class reconcile loop", func() {
 		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal(""))
 	})
 
-	It("Should set the scratchspaceStorageClass to the default without override", func() {
+	It("Should set the scratchspaceStorageClass to blank if there is default sc", func() {
 		reconciler, cdiConfig := createConfigReconciler(createStorageClassList(
-			*CreateStorageClass("test-default-sc", map[string]string{
-				AnnDefaultStorageClass: "true",
-			},
-			)))
-		err := reconciler.reconcileStorageClass(cdiConfig)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal("test-default-sc"))
-	})
-
-	It("Should set the scratchspaceStorageClass to the default without override and multiple sc", func() {
-		reconciler, cdiConfig := createConfigReconciler(createStorageClassList(
-			*CreateStorageClass("test-sc3", nil),
 			*CreateStorageClass("test-default-sc", map[string]string{
 				AnnDefaultStorageClass: "true",
 			}),
-			*CreateStorageClass("test-sc", nil),
-			*CreateStorageClass("test-sc2", nil),
 		))
 		err := reconciler.reconcileStorageClass(cdiConfig)
 		Expect(err).ToNot(HaveOccurred())
-		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal("test-default-sc"))
+		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal(""))
+	})
+
+	It("Should clear the scratchspaceStorageClass when only the CDIConfig status was previously set", func() {
+		reconciler, cdiConfig := createConfigReconciler(createStorageClassList(
+			*CreateStorageClass("test-default-sc", map[string]string{
+				AnnDefaultStorageClass: "true",
+			}),
+		))
+		// status was leftover, but spec is not set
+		cdiConfig.Status.ScratchSpaceStorageClass = "test"
+		err := reconciler.reconcileStorageClass(cdiConfig)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal(""))
 	})
 
 	It("Should set the scratchspaceStorageClass to the override even with default", func() {
@@ -398,7 +397,7 @@ var _ = Describe("Controller storage class reconcile loop", func() {
 		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal(override))
 	})
 
-	It("Should set the scratchspaceStorageClass to the default with invalid override", func() {
+	It("Should set the scratchspaceStorageClass to empty with invalid override", func() {
 		reconciler, cdiConfig := createConfigReconciler(createStorageClassList(
 			*CreateStorageClass("test-sc3", nil),
 			*CreateStorageClass("test-default-sc", map[string]string{
@@ -411,7 +410,7 @@ var _ = Describe("Controller storage class reconcile loop", func() {
 		cdiConfig.Spec.ScratchSpaceStorageClass = &override
 		err := reconciler.reconcileStorageClass(cdiConfig)
 		Expect(err).ToNot(HaveOccurred())
-		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal("test-default-sc"))
+		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal(""))
 	})
 })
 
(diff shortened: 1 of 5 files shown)


akalenyu:
There should be an issue in the repo with some users asking for precisely this, could link that PR to it so it closes :pray:

pkg/controller/config-controller_test.go
```diff
@@ -357,29 +357,15 @@ var _ = Describe("Controller storage class reconcile loop", func() {
 		Expect(cdiConfig.Status.ScratchSpaceStorageClass).To(Equal(""))
```
akalenyu, line 357: I think you also want a test case for upgrade scenarios, i.e. given a CDIConfig with status.scratchStorageClass set, it gets reconciled to ""

dsanatar:
@akalenyu updated those identical test names and also added a test for the upgrade path.

> Interesting, so there's essentially no use for populating status.scratchStorageClass with the default, since PVC storage class will always be non empty? Not sure if there are edge cases here. Particularly binding via provisioner string. Is there a reason you prefer this impl. to changing GetScratchPvcStorageClass? Not necessarily against just feel like something we have to talk out a bit since there could be rough edges.
>
I considered changing `GetScratchPvcStorageClass` but the existing logic there actually seemed well setup to handle the empty string for the `status.scratchStorageClass`, the function description seemed to almost expect this behavior already. I suppose the main worry here is if the PVC _does_ have their storage class set to an empty string then this function would return "", not sure if that would be unwanted behavior?

Should it instead return the default storage class if the config is not set and the owner PVC has an empty storage class?

akalenyu:
> @akalenyu updated those identical test names and also added a test for the upgrade path.
> 
> > Interesting, so there's essentially no use for populating status.scratchStorageClass with the default, since PVC storage class will always be non empty? Not sure if there are edge cases here. Particularly binding via provisioner string. Is there a reason you prefer this impl. to changing GetScratchPvcStorageClass? Not necessarily against just feel like something we have to talk out a bit since there could be rough edges.
> 
> I considered changing `GetScratchPvcStorageClass` but the existing logic there actually seemed well setup to handle the empty string for the `status.scratchStorageClass`, the function description seemed to almost expect this behavior already. I suppose the main worry here is if the PVC _does_ have their storage class set to an empty string then this function would return "", not sure if that would be unwanted behavior?
> 
> Should it instead return the default storage class if the config is not set and the owner PVC has an empty storage class?

Actually I had it wrong mentioning binding via provisioner name, that should still work post changes. Maybe some edge case in that area but they could always set a scratch storage class.

Regarding your scenario though, yeah, "" is somewhat nuanced https://kubernetes.io/blog/2023/01/05/retroactive-default-storage-class/ so I think returning "" does actually make sense?

akalenyu:
Long story short folks in the "" scName area could experience a regression in that previously they were falling back to a safe default (probably dynamic) provisioner. With the new logic, they might not have a spare matching "" PV.

With that said, this is extremely niche and I don't even know if CDI works with those today, nor wants to support those. Again, should be able to override.

/approve
/cc @awels

tests/datavolume_test.go
```diff
@@ -2381,6 +2451,12 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 		BeforeEach(func() {
 			testScName = fmt.Sprintf("%s-%s", testScNamePrefix, strings.ToLower(util.RandAlphaNum(8)))
 			By(fmt.Sprintf("init test storage class name: %s", testScName))
+
+			By("Setting scratch space storage class to default storage class")
+			err := utils.UpdateCDIConfig(f.CrClient, func(config *cdiv1.CDIConfigSpec) {
+				config.ScratchSpaceStorageClass = &utils.DefaultStorageClass.Name
+			})
+			Expect(err).ToNot(HaveOccurred())
```
akalenyu, line 2389: Apologies totally missed the tests changes in datavolume_test. Why is this needed? Can't it work with the target DVs storage class?

tests/datavolume_test.go
```diff
@@ -1450,6 +1450,76 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
 			Expect(err).ToNot(HaveOccurred())
 			Expect(md5Match).To(BeTrue())
 		})
+
+		It("should have scratch space PVC storage class set to same as the target PVC", func() {
```
akalenyu: Feel like we could piggyback on existing cases or even settle for the unit tests. WDYT?

pkg/controller/config-controller_test.go
```diff
@@ -357,29 +357,28 @@ var _ = Describe("Controller storage class reconcile loop", func() {
...
 		reconciler, cdiConfig := createConfigReconciler(createStorageClassList(
-			*CreateStorageClass("test-sc3", nil),
 			*CreateStorageClass("test-default-sc", map[string]string{
 				AnnDefaultStorageClass: "true",
 			}),
-			*CreateStorageClass("test-sc", nil),
-			*CreateStorageClass("test-sc2", nil),
 		))
+		// status was leftover, but spec is not set
+		cdiConfig.Status.ScratchSpaceStorageClass = "test"
```
akalenyu, line 378: Don't you need to call update after this change? otherwise control loop runs on cdiConfig without your change

tests/datavolume_test.go
dsanatar: will remove since deemed unneeded

tests/datavolume_test.go
dsanatar: we need this now because the below tests create scratch space PVCs but are using fake storage classes for their PVs. Previously, the scratch space PVCs would be able to bind because it would fallback to using the default storage class however, now it inherits this fake storage class and gets stuck pending.

pkg/controller/config-controller_test.go
akalenyu: ah nvm we call reconcileStorageClass and not Reconcile

tests/datavolume_test.go
akalenyu: Ah ofcourse, missed them being under `Describe("Verify that when the required storage class is missing", Serial, func() {`

pkg/controller/config-controller_test.go
dsanatar: I don't believe so? we are just modifying the struct before the reconcile is called. same like how we check that the override in the below test

PR #4056, opened 2026-03-16
Add AnnPodServiceAccount to allow custom ServiceAccount for CDI pods

Hazanel:
Add AnnPodServiceAccount to allow custom ServiceAccount for CDI pods

This PR adds support for specifying a custom ServiceAccount for importer
and uploader pods via a new PVC/DV annotation. This enables organizations
with security policies that require migration workloads to run under a
specific ServiceAccount rather than the namespace default.

The feature can be used in two ways:
1. **DataVolume spec field**: Set `spec.serviceAccountName` on a DataVolume,
   which gets propagated as an annotation onto the underlying PVC.
2. **PVC annotation**: Set `cdi.kubevirt.io/storage.pod.serviceAccountName`
   directly on a PVC.

The implementation mirrors the existing AnnPriorityClassName pattern:
- Added AnnPodServiceAccount constant and GetPodServiceAccount() helper
- Added AnnPodServiceAccount to allowedAnnotations whitelist
- Set pod.Spec.ServiceAccountName in import and upload controllers
- Added ServiceAccountName field to DataVolumeSpec
- Propagate annotation from DataVolume to PVC in controller-base.go
- Updated OpenAPI and CRD generated schemas for the new spec field
- Added unit tests following the AnnPriorityClassName test pattern

Clone flows are intentionally excluded — custom ServiceAccount is only
supported for import and upload. This avoids cross-namespace security
concerns where the user may not have visibility into ServiceAccounts
in the source namespace (e.g., golden images namespace).

Note: prep-claim.go also adds the missing PriorityClassName to the pod spec.
This is a pre-existing bug — CopyAllowedAnnotations copied the annotation to
pod metadata, but PodSpec.PriorityClassName was never set, so Kubernetes
ignored it for scheduling. Fixed here since we're already touching the same
code block.

Also refactored planHostAssistedFromPVC/planHostAssistedFromSnapshot in
planner.go to extract shared helpers (applyCloneSourceSpec, newRebindPhase),
fixing a pre-existing bug where Preallocation was not set on the snapshot
clone path.

**After the fix**:

Users can set the annotation cdi.kubevirt.io/storage.pod.serviceAccountName
on a DataVolume or PVC, and CDI importer/uploader pods will run under that
ServiceAccount. Clone and source-namespace pods are not affected and continue
using the default SA. If the SA does not exist, Kubernetes rejects the pod
with a clear error message.

Example DataVolume usage:
```yaml
kind: DataVolume
metadata:
  annotations:
    cdi.kubevirt.io/storage.pod.serviceAccountName: "my-custom-sa"
spec:
  source:
    http:
      url: "https://example.com/disk.img"
  storage:
    resources:
      requests:
        storage: 1Gi
```

Jira Ticket: https://redhat.atlassian.net/browse/CNV-81806

**Release note**:
```release-note
CDI importer and uploader pods can now use a custom ServiceAccount by
setting the annotation cdi.kubevirt.io/storage.pod.serviceAccountName
on the DataVolume or PVC. Clone flows are not affected.
```

### DIFF
--- a/pkg/controller/clone/prep-claim_test.go
+++ b/pkg/controller/clone/prep-claim_test.go
@@ -244,6 +244,24 @@ var _ = Describe("PrepClaimPhase test", func() {
 			Expect(pod.Spec.NodeName).To(Equal("node1"))
 		})
 
+		It("should create pod with priority class from PVC annotations but not service account", func() {
+			claim := getClaim()
+			cc.AddAnnotation(claim, cc.AnnPriorityClassName, "high-priority")
+			cc.AddAnnotation(claim, cc.AnnPodServiceAccount, "my-sa")
+			claim.Spec.Resources.Requests[corev1.ResourceStorage] = defaultRequestSize
+			claim.Status.Phase = corev1.ClaimPending
+
+			p := createPrepClaimPhase(claim)
+
+			result, err := p.Reconcile(context.Background())
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result).ToNot(BeNil())
+
+			pod := getCreatedPod(p)
+			Expect(pod.Spec.PriorityClassName).To(Equal("high-priority"))
+			Expect(pod.Spec.ServiceAccountName).To(BeEmpty())
+		})
+
 		It("should create pod if desired is bigger", func() {
 			claim := getClaim()
 			claim.Spec.Resources.Requests[corev1.ResourceStorage] = defaultRequestSize
--- a/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
+++ b/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
@@ -56,6 +56,9 @@ type DataVolumeSpec struct {
 	Storage *StorageSpec `json:"storage,omitempty"`
 	//PriorityClassName for Importer, Cloner and Uploader pod
 	PriorityClassName string `json:"priorityClassName,omitempty"`
+	// ServiceAccountName for Importer and Uploader pod
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 	//DataVolumeContentType options: "kubevirt", "archive"
 	// +kubebuilder:validation:Enum="kubevirt";"archive"
 	ContentType DataVolumeContentType `json:"contentType,omitempty"`
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -5027,6 +5027,10 @@
       "description": "PVC is the PVC specification",
       "$ref": "#/definitions/v1.PersistentVolumeClaimSpec"
      },
+     "serviceAccountName": {
+      "description": "ServiceAccountName for Importer and Uploader pod",
+      "type": "string"
+     },
      "source": {
       "description": "Source is the src of the data for the requested DataVolume",
       "$ref": "#/definitions/v1beta1.DataVolumeSource"
--- a/pkg/apis/core/v1beta1/openapi_generated.go
+++ b/pkg/apis/core/v1beta1/openapi_generated.go
@@ -18516,6 +18516,13 @@ func schema_pkg_apis_core_v1beta1_DataVolumeSpec(ref common.ReferenceCallback) c
 							Format:      "",
 						},
 					},
+					"serviceAccountName": {
+						SchemaProps: spec.SchemaProps{
+							Description: "ServiceAccountName for Importer and Uploader pod",
+							Type:        []string{"string"},
+							Format:      "",
+						},
+					},
 					"contentType": {
 						SchemaProps: spec.SchemaProps{
 							Description: "DataVolumeContentType options: \"kubevirt\", \"archive\"",
--- a/pkg/controller/clone/planner.go
+++ b/pkg/controller/clone/planner.go
@@ -533,25 +533,13 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
 		DesiredClaim:   desiredClaim,
 		ImmediateBind:  true,
 		OwnershipLabel: p.OwnershipLabel,
-		Preallocation:  cc.GetPreallocation(ctx, p.Client, args.DataSource.Spec.Preallocation),
 		Client:         p.Client,
 		Log:            args.Log,
 		Recorder:       p.Recorder,
 	}
 
-	if args.DataSource.Spec.PriorityClassName != nil {
-		hcp.PriorityClassName = *args.DataSource.Spec.PriorityClassName
-	}
-
-	rp := &RebindPhase{
-		SourceNamespace: desiredClaim.Namespace,
-		SourceName:      desiredClaim.Name,
-		TargetNamespace: args.TargetClaim.Namespace,
-		TargetName:      args.TargetClaim.Name,
-		Client:          p.Client,
-		Log:             args.Log,
-		Recorder:        p.Recorder,
-	}
+	p.applyCloneSourceSpec(ctx, args, hcp)
+	rp := p.newRebindPhase(args, desiredClaim)
 
 	return []Phase{hcp, rp}, nil
 }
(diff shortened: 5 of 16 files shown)


staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -56,6 +56,9 @@ type DataVolumeSpec struct {
 	Storage *StorageSpec `json:"storage,omitempty"`
 	//PriorityClassName for Importer, Cloner and Uploader pod
 	PriorityClassName string `json:"priorityClassName,omitempty"`
+	// ServiceAccountName for Importer, Cloner and Uploader pod
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
arnongilboa, line 61: In the PR description you mentiond an annotation, but I see it's actually a spec field.

Btw, any reason here it's not `*string` like in `VolumeCloneSourceSpec`?

Hazanel: It's both — the annotation (cdi.kubevirt.io/storage.pod.serviceAccountName) works on PVCs directly, and the spec field (spec.serviceAccountName) works on DataVolumes (which gets propagated as the annotation onto the PVC). I'll update the PR description to clarify both paths.

Regarding string vs *string: this follows the existing PriorityClassName pattern — it's string in DataVolumeSpec and *string in VolumeCloneSourceSpec. That inconsistency is pre-existing from PriorityClassName. Changing it to *string in DataVolumeSpec would be an API-breaking change

Hazanel:
@awels @akalenyu  - Hey, Could you please also take a look please?

Hazanel:
**End-to-End Testing on OpenShift 4.21 Cluster**
**CDI Custom ServiceAccount (upstream feature verification):**
- Patched CDI controller with custom-built image containing `AnnPodServiceAccount` support
- Created DataVolume with `cdi.kubevirt.io/storage.pod.serviceAccountName: cdi-test-sa` annotation
- Verified: importer pod ran with `serviceAccountName: cdi-test-sa`
- Verified: annotation propagated from DataVolume → PVC → importer pod
- Import completed successfully

akalenyu:
> CDI pods (importer, cloner, uploader) can now use a custom ServiceAccount
> by setting the annotation cdi.kubevirt.io/storage.pod.serviceAccountName
> on the DataVolume or PVC.

I believe you went with an API under .spec for DVs?

akalenyu:
> I am a little worried about the security implications of this; CDI bypasses checking that the user can use this specified service account, because the pod is created by a privileged controller, not the user.
> 
> I am not completely sure if this is a problem in most clusters but just something to consider from a first pass.

@mhenriks do you think this is this a real concern? I am pretty sure kubevirt could be used in a similar way to use cloud creds in the VM

Hazanel:
Hey @akalenyu 

This mirrors the existing AnnPriorityClassName pattern where CDI trusts the annotation without additional RBAC checks.
KubeVirt has a similar exposure with VM ServiceAccount specs and in practice, the primary consumer (Forklift/MTV) sets this annotation from a controller, not directly from user input. 
For the raw CDI case, access is gated by the ability to create DVs/PVCs in the namespace, and the SA only affects the pod's identity  not the user's access. 

About the second question: 
Yes, the primary API is spec.serviceAccountName on the DataVolume. Updated the release note to reflect this.

Acedus:
> I am a little worried about the security implications of this; CDI bypasses checking that the user can use this specified service account, because the pod is created by a privileged controller, not the user.
> 
> I am not completely sure if this is a problem in most clusters but just something to consider from a first pass.

I think that's a general issue of k8s? There isn't a specific RBAC one needs to have in order to be able to specify a service account for a pod, all they need is sufficient permissions to create that pod in the first place. Going by that rule, I think it's safe to delegate the assignment of a service account to CDI (if the user has exec permissions but not create that's the only escalation route I can think of off the top of my head). There is the question of whether we should raise in the documentation that a cluster admin who would like to restrict use of certain service accounts for certain entities will need to enforce it at the DV/PVC level as well via a custom policy (since examining the pod will no longer be sufficient).

akalenyu:
> > I am a little worried about the security implications of this; CDI bypasses checking that the user can use this specified service account, because the pod is created by a privileged controller, not the user.
> > I am not completely sure if this is a problem in most clusters but just something to consider from a first pass.
> 
> I think that's a general issue of k8s? There isn't a specific RBAC one needs to have in order to be able to specify a service account for a pod, all they need is sufficient permissions to create that pod in the first place. Going by that rule, I think it's safe to delegate the assignment of a service account to CDI (if the user has exec permissions but not create that's the only escalation route I can think of off the top of my head). There is the question of whether we should raise in the documentation that a cluster admin who would like to restrict use of certain service accounts for certain entities will need to enforce it at the DV/PVC level as well via a custom policy (since examining the pod will no longer be sufficient).

Yeah I was more concerned with custom kyverno policies and things of this nature, in conjunction with cloud credentials. I am also leaning towards this being totally fine (and it's anyway the situation in kubevirt/kubevirt today too as @Hazanel mentioned). @mhenriks If you could just ack this so we're not missing anything :pray:

mhenriks:
> > > I am a little worried about the security implications of this; CDI bypasses checking that the user can use this specified service account, because the pod is created by a privileged controller, not the user.
> > > I am not completely sure if this is a problem in most clusters but just something to consider from a first pass.
> > 
> > 
> > I think that's a general issue of k8s? There isn't a specific RBAC one needs to have in order to be able to specify a service account for a pod, all they need is sufficient permissions to create that pod in the first place. Going by that rule, I think it's safe to delegate the assignment of a service account to CDI (if the user has exec permissions but not create that's the only escalation route I can think of off the top of my head). There is the question of whether we should raise in the documentation that a cluster admin who would like to restrict use of certain service accounts for certain entities will need to enforce it at the DV/PVC level as well via a custom policy (since examining the pod will no longer be sufficient).
> 
> Yeah I was more concerned with custom kyverno policies and things of this nature, in conjunction with cloud credentials. I am also leaning towards this being totally fine (and it's anyway the situation in kubevirt/kubevirt today too as @Hazanel mentioned). @mhenriks If you could just ack this so we're not missing anything 🙏

Not concerned about this from a security perspective in target namespace but I think the API is kind of questionable WRT cross namespace cloning. I see that the serviceaccount name is used for pods created in both source and target namespaces. Can we get away with just setting the serviceaccount on pods created in the target namespace? 

Issue 1: Are we sure that serviceaccounts with the same name will be in both source and target namespaces? Does the API have to be more flexible?

Issue 2: Regular users have no visibility into serviceaccounts in the Openshift "golden images" namespace for example so it could possibly be considered a violation to allow them to specify a specific serviceaccount for pods created there

akalenyu:
> Issue 2: Regular users have no visibility into serviceaccounts in the Openshift "golden images" namespace for example so it could possibly be considered a violation to allow them to specify a specific serviceaccount for pods created there

I believe the original forklift issue (and the community issue prior) are focused around the import flow. Maybe we can start slow with supporting just that?

pkg/controller/clone/prep-claim_test.go
```diff
@@ -244,6 +244,24 @@ var _ = Describe("PrepClaimPhase test", func() {
 			Expect(pod.Spec.NodeName).To(Equal("node1"))
 		})
 
+		It("should create pod with priority class from PVC annotations but not service account", func() {
```
akalenyu, line 247: Nice catch!

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -56,6 +56,9 @@ type DataVolumeSpec struct {
 	Storage *StorageSpec `json:"storage,omitempty"`
 	//PriorityClassName for Importer, Cloner and Uploader pod
 	PriorityClassName string `json:"priorityClassName,omitempty"`
+	// ServiceAccountName for Importer, Cloner and Uploader pod
```
akalenyu: I think I would just leave out clone if it comes down to only doing target ns.
Most are going to be using cross ns, and we'd have to document this super internal impl. specific behavior for serviceAccountName. At that point I think it's simpler to say "only import/upload supported"
@mhenriks @arnongilboa @Acedus WDYT?

Either way we should adjust the API doc here with whatever we choose

Acedus: +1

mhenriks: works for me

PR #4061, opened 2026-03-19
Fix snapshot clone watcher to re-trigger reconciliation when snapshot becomes ready

noamasu:
when cloning from a VolumeSnapshot that is not yet ReadyToUse, the clone populator stops reconciling and never retries. Since no watch re-triggers reconciliation when the snapshot becomes ready, the clone stays stuck in CloneFromSnapshotSourceInProgress permanently

This affects CSI drivers that report RestoreSize before ReadyToUse, creating a window where a clone can be initiated against a snapshot that isn't fully usable yet

this corrects the clone source mapper to resolve VolumeCloneSource -> target PVC and the namespace resolution for cross namespace clones, so the existing watcher properly re-triggers reconciliation when ReadyToUse becomes true.

Jira: https://redhat.atlassian.net/browse/CNV-82089

**Release note**:

```release-note
Snapshot clone no longer gets permanently stuck in CloneFromSnapshotSourceInProgress when the snapshot is not immediately ReadyToUse
```

### DIFF
--- a/pkg/controller/populators/util.go
+++ b/pkg/controller/populators/util.go
@@ -68,9 +68,11 @@ func isDataSourceRefValid(dataSourceRef *corev1.TypedObjectReference) bool {
 
 func getPopulationSourceNamespace(pvc *corev1.PersistentVolumeClaim) string {
 	namespace := pvc.GetNamespace()
-	// The populator CR can be in a different namespace from the target PVC
-	// if the CrossNamespaceVolumeDataSource feature gate is enabled in the
-	// kube-apiserver and the kube-controller-manager.
+	// For cross-namespace clones, CDI stores the source namespace in this annotation
+	// because DataSourceRef.Namespace requires the CrossNamespaceVolumeDataSource feature gate.
+	if dsns, ok := pvc.GetAnnotations()[AnnDataSourceNamespace]; ok && dsns != "" {
+		return dsns
+	}
 	dataSourceRef := pvc.Spec.DataSourceRef
 	if dataSourceRef != nil && dataSourceRef.Namespace != nil && *dataSourceRef.Namespace != "" {
 		namespace = *pvc.Spec.DataSourceRef.Namespace


pkg/controller/populators/populator-base.go
```diff
@@ -153,18 +153,7 @@ func addCommonPopulatorsWatches(mgr manager.Manager, c controller.Controller, lo
 	}
 
 	mapDataSourceRefToPVC := func(ctx context.Context, obj client.Object) (reqs []reconcile.Request) {
```
akalenyu: do we want to reuse for https://github.com/kubevirt/containerized-data-importer/blob/1f3002ea403c16140998868cde20e005a55d9967/pkg/controller/populators/forklift-populator.go#L150-L163

pkg/controller/populators/clone-populator.go
```diff
@@ -162,7 +162,7 @@ func addCloneSourceWatches(mgr manager.Manager, c controller.Controller, log log
 			}
 
 			for _, cs := range volumeCloneSources.Items {
-				reqs = append(reqs, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: cs.Namespace, Name: cs.Name}})
+				reqs = append(reqs, mapSourceToPVCs(ctx, mgr.GetClient(), log, cdiv1.VolumeCloneSourceRef, &cs)...)
```
akalenyu: just thinking out loud, should we be going through this index dance (src snap/pvc->vcs->target pvc)? can't we come up with a different index to bypass and go from src->target directly?

pkg/controller/populators/util.go
```diff
@@ -68,9 +68,11 @@ func isDataSourceRefValid(dataSourceRef *corev1.TypedObjectReference) bool {
 
 func getPopulationSourceNamespace(pvc *corev1.PersistentVolumeClaim) string {
 	namespace := pvc.GetNamespace()
-	// The populator CR can be in a different namespace from the target PVC
-	// if the CrossNamespaceVolumeDataSource feature gate is enabled in the
-	// kube-apiserver and the kube-controller-manager.
+	// For cross-namespace clones, CDI stores the source namespace in this annotation
+	// because DataSourceRef.Namespace requires the CrossNamespaceVolumeDataSource feature gate.
+	if anno, ok := pvc.GetAnnotations()[AnnDataSourceNamespace]; ok && anno != "" {
```
akalenyu: nitzone; anno is a bit of a weird naming choice for the value, I would do `dsns`

akalenyu:
@noamasu might want to edit title/pr desc/release note

pkg/controller/populators/populator-base.go
noamasu: good catch!

pkg/controller/populators/clone-populator.go
noamasu: I was thinking about it. but what would the direct src->target index be keyed on that doesnt require reading another resource? the target pvc doesnt know about the source snapshot directly since datasourceref is pointing to the cs. unless we add some info about the snapshot in the pvc somehow (maybe annotation?)

pkg/controller/populators/util.go
noamasu: gotcha

pkg/controller/populators/clone-populator.go
akalenyu: We could annotate the target PVC through the DV snap/pvc clone controllers, but yeah not sure if it's worth it. Would allow us to get rid of 1 index but it's micro optimization territory. probably keep as is

akalenyu:
/retest
seems like this cron test is failing even prior to 1.35 bump, should get better with https://github.com/kubevirt/containerized-data-importer/pull/4047 but considering whacking it for unit test anyway

pkg/controller/populators/populator-base.go
Acedus: nit: Is there even a need for a separate func declaration now? Let's just call `mapSourceToPVCs` with the relevant parameters.

pkg/controller/populators/forklift-populator.go
```diff
@@ -147,19 +147,8 @@ func addWatchers(mgr manager.Manager, c controller.Controller, log logr.Logger,
...
-		}
-		if err := mgr.GetClient().List(ctx, &pvcs, matchingFields); err != nil {
-			log.Error(err, "Unable to list PVCs", "matchingFields", matchingFields)
-			return reqs
-		}
-		for _, pvc := range pvcs.Items {
-			reqs = append(reqs, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: pvc.Namespace, Name: pvc.Name}})
-		}
-		return reqs
+	mapDataSourceRefToPVC := func(ctx context.Context, obj client.Object) []reconcile.Request {
```
Acedus: Same here.

Acedus (approved):
Thanks @noamasu! 

Small nit, feel free to unhold
/hold

pkg/controller/populators/populator-base.go
noamasu: done.

akalenyu:
@noamasu cancel hold when you are good

akalenyu:
/retest
no test run

akalenyu:
CI filesystem is bad for nfs, lots of oom

akalenyu:
CI filesystem is bad for nfs, lots of oom

PR #4067, opened 2026-03-20
Set EnableServiceLinks=false on importer/cloner/uploader pods

### LINKED ISSUE #4059: Set enableServiceLinks: false on importer/cloner/uploader pods
## Problem

CDI importer pods fail with `exec /usr/bin/sh: argument list too long` in namespaces with a large number of Kubernetes Services.

By default, Kubernetes injects environment variables for every Service in the namespace into each pod (`enableServiceLinks` defaults to `true`). Each Service generates ~7 env vars (`{SVC}_SERVICE_HOST`, `{SVC}_SERVICE_PORT`, `{SVC}_PORT`, etc.). In namespaces with ~2,000+ Services, the combined size of these environment variables exceeds the Linux kernel's `ARG_MAX` limit (~2MB), causing the init container to fail on `execve()`.

CDI importer pods do not use Kubernetes service discovery env vars — they connect to registries via URLs, not service env vars. There is no functional reason for these pods to have `enableServiceLinks: true`.

## Environment

- CDI v1.63.1 (also checked v1.64.0, same behavior)
- ~2,200 Services in the target namespace
- Importer pod init container fails immediately with `exec /usr/bin/sh: argument list too long`

## Expected behavior

CDI importer (and cloner/uploader) pods should start successfully regardless of how many Services exist in the namespace.

## Proposed fix

Set `EnableServiceLinks: ptr.To(false)` in `makeImporterPodSpec` (and equivalent functions for cloner/uploader pods).

This is a one-line change per pod type. KubeVirt itself applied the same fix for `virt-launcher` pods in https://github.com/kubevirt/kubevirt/issues/4391 / https://github.com/kubevirt/kubevirt/pull/4393 back in 2020 for the same reason.

## Workaround

A mutating admission webhook (e.g., Kyverno policy) can patch `enableServiceLinks: false` onto CDI pods at admission time, but this shouldn't be necessary for pods that don't use service discovery env vars.

jholm117:
## Summary

Fixes #4059. CDI pods fail with "argument list too long" in namespaces with 2000+ Services because Kubernetes injects ~7 env vars per Service by default.

- Set `EnableServiceLinks: ptr.To(false)` on importer, cloner, uploader, and forklift populator pod specs
- Updated corresponding test helpers to include the new field
- All 428 existing unit tests pass

Same fix KubeVirt applied to virt-launcher pods in 2020 (kubevirt/kubevirt#4393).

```release-note
Set EnableServiceLinks=false on CDI worker pods to prevent "argument list too long" failures in namespaces with many Services.
```

### DIFF
--- a/pkg/controller/clone-controller.go
+++ b/pkg/controller/clone-controller.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -630,6 +631,10 @@ func MakeCloneSourcePodSpec(sourceVolumeMode corev1.PersistentVolumeMode, image,
 			},
 			ImagePullSecrets: imagePullSecrets,
 			RestartPolicy:    corev1.RestartPolicyOnFailure,
+			// https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables
+			// Disable service environment variable injection to avoid 'argument list too long'
+			// errors in namespaces with many Services (each injects ~7 env vars).
+			EnableServiceLinks: ptr.To(false),
 			Volumes: []corev1.Volume{
 				{
 					Name: cc.DataVolName,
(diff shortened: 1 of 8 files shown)


awels (approved):

awels:
Unfortunately you can only run the PROW commands if you are in the OWNERS file. But I triggered it again. I am trying to find someone else to approve it.

PR #4072, opened 2026-03-24
[release-v1.64] Add AnnPodServiceAccount annotation for custom ServiceAccount in CDI pods

Hazanel:
Backport of #4056 to release-v1.64 with the `spec.serviceAccountName` API
field removed — only the annotation-based approach is included.
This PR adds support for specifying a custom ServiceAccount for importer
and uploader pods via a PVC/DV annotation. This enables organizations
with security policies that require migration workloads to run under a
specific ServiceAccount rather than the namespace default.
Set the annotation `cdi.kubevirt.io/storage.pod.serviceAccountName` on
a DataVolume or PVC metadata. The annotation is propagated from
DataVolume to the underlying PVC via the existing metadata propagation
loop in controller-base.go.
**Key differences from the main branch version (#4056)**:
- No `spec.serviceAccountName` field in DataVolumeSpec (no CRD/API change)
- No spec-to-annotation propagation logic in controller-base.go
- Annotation must be set directly on DataVolume/PVC metadata
**The implementation adds**:
- `AnnPodServiceAccount` constant and `GetPodServiceAccount()` helper
- `AnnPodServiceAccount` to `allowedAnnotations` whitelist
- `ServiceAccountName` on pod specs in import and upload controllers
- Unit tests following the `AnnPriorityClassName` test pattern
Clone flows are intentionally excluded — custom ServiceAccount is only
supported for import and upload. This avoids cross-namespace security
concerns where the user may not have visibility into ServiceAccounts
in the source namespace (e.g., golden images namespace).
Note: prep-claim.go also adds the missing PriorityClassName to the pod spec.
This is a pre-existing bug — `CopyAllowedAnnotations` copied the annotation to
pod metadata, but `PodSpec.PriorityClassName` was never set, so Kubernetes
ignored it for scheduling. Fixed here since we're already touching the same
code block.
Also refactored `planHostAssistedFromPVC`/`planHostAssistedFromSnapshot` in
planner.go to extract shared helpers (`applyCloneSourceSpec`, `newRebindPhase`),
fixing a pre-existing bug where Preallocation was not set on the snapshot
clone path.
**After the fix**:
Users can set the annotation `cdi.kubevirt.io/storage.pod.serviceAccountName`
on a DataVolume or PVC, and CDI importer/uploader pods will run under that
ServiceAccount. Clone and source-namespace pods are not affected and continue
using the default SA. If the SA does not exist, Kubernetes rejects the pod
with a clear error message.
Example DataVolume usage:
```yaml
kind: DataVolume
metadata:
  annotations:
    cdi.kubevirt.io/storage.pod.serviceAccountName: "my-custom-sa"
spec:
  source:
    http:
      url: "https://example.com/disk.img"
  storage:
    resources:
      requests:
        storage: 1Gi
```

Which issue(s) this PR fixes:
Jira Ticket: https://redhat.atlassian.net/browse/CNV-81806

**Release note**:
```release-note
CDI importer and uploader pods can now use a custom ServiceAccount by
setting the annotation cdi.kubevirt.io/storage.pod.serviceAccountName
on the DataVolume or PVC. Clone flows are not affected.
```

### DIFF
--- a/pkg/controller/clone/planner.go
+++ b/pkg/controller/clone/planner.go
@@ -533,25 +533,13 @@ func (p *Planner) planHostAssistedFromPVC(ctx context.Context, args *PlanArgs) (
 		DesiredClaim:   desiredClaim,
 		ImmediateBind:  true,
 		OwnershipLabel: p.OwnershipLabel,
-		Preallocation:  cc.GetPreallocation(ctx, p.Client, args.DataSource.Spec.Preallocation),
 		Client:         p.Client,
 		Log:            args.Log,
 		Recorder:       p.Recorder,
 	}
 
-	if args.DataSource.Spec.PriorityClassName != nil {
-		hcp.PriorityClassName = *args.DataSource.Spec.PriorityClassName
-	}
-
-	rp := &RebindPhase{
-		SourceNamespace: desiredClaim.Namespace,
-		SourceName:      desiredClaim.Name,
-		TargetNamespace: args.TargetClaim.Namespace,
-		TargetName:      args.TargetClaim.Name,
-		Client:          p.Client,
-		Log:             args.Log,
-		Recorder:        p.Recorder,
-	}
+	p.applyCloneSourceSpec(ctx, args, hcp)
+	rp := p.newRebindPhase(args, desiredClaim)
 
 	return []Phase{hcp, rp}, nil
 }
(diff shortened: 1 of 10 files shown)


Acedus:
As discussed in the sig-storage meeting, I think this is fine considering we don't have a concrete backporting policy. With that said, a PR like this falls into a bit of a gray area IMO, so we should strive to create a policy as soon as possible.

/lgtm
/hold

Waiting for @awels' opinion on the matter.

awels:
I think there is an issue with the hpp release, I am investigating.

akalenyu:
/override pull-containerized-data-importer-e2e-hpp-previous
hpp failing to start, unrelated

akalenyu:
@awels 
you can remove the hold if you agree with this backport

awels:
/release-note-edit "CDI importer and uploader pods can now use a custom ServiceAccount by setting the annotation cdi.kubevirt.io/storage.pod.serviceAccountName on the DataVolume or PVC. Clone flows are not affected."

PR #4073, opened 2026-03-24
Improve filesystem cloning by skipping tar for single disk.img files

alromeros:

Different implementation to avoid the issues described in https://github.com/kubevirt/containerized-data-importer/pull/4043.

This PR adds disk-image-clone content type to bypass tar overhead when cloning filesystems containing only disk.img.

This avoids SEEK_HOLE performance issues that can be detrimental for filesystems that don't support it.

Fixes #

https://issues.redhat.com/browse/CNV-45701
**Release note**:

```release-note
Avoid using tar for single file disk.img clones 
```

### DIFF
--- a/cmd/cdi-cloner/BUILD.bazel
+++ b/cmd/cdi-cloner/BUILD.bazel
@@ -60,6 +60,7 @@ go_test(
     ],
     embed = [":go_default_library"],
     deps = [
+        "//pkg/common:go_default_library",
         "//pkg/util/prometheus:go_default_library",
         "//vendor/github.com/onsi/ginkgo/v2:go_default_library",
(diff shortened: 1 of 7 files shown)


akalenyu:
Is tarring really that inconsequential? don't we regress on performance by starting to write holes etc over the wire?

alromeros:
> Is tarring really that inconsequential? don't we regress on performance by starting to write holes etc over the wire?

Removing the sparse handling will definitely impact performance in some cases, yes... we already see how export server in kubevirt is slower in some workflows due to lack of sparse handling (I'm thinking of those oracle 50Gi volumes...). I’ll move the other PR back to draft and keep both open for now while I explore other approaches.

PR #4074, opened 2026-03-25
Recreate DIC import DV when first default SC is set

noamasu:
The default SC change condition only handled changes from one SC to another, but did not handle the first default SC being set.

To get the DIC-created DV with the DIC-special RWO preference, it must be deleted (as DV is immutable) and recreated with the right settings from the new default StorageClass.

Jira: https://redhat.atlassian.net/browse/CNV-81826

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -359,19 +359,18 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		return res, err
 	}
 	if desiredStorageClass != nil {
-		if deleted, err := r.deleteOutdatedPendingPvc(ctx, pvc, desiredStorageClass.Name, dataImportCron.Name); deleted || err != nil {
+		desiredSc := desiredStorageClass.Name
+		outdated, err := r.isImportOutdated(ctx, dataImportCron, dv, pvc, desiredSc)
+		if err != nil {
 			return res, err
 		}
-		currentSc, hasCurrent := dataImportCron.Annotations[AnnStorageClass]
-		desiredSc := desiredStorageClass.Name
-		if hasCurrent && currentSc != desiredSc {
-			r.log.Info("Storage class changed, delete most recent source on the old sc as it's no longer the desired", "currentSc", currentSc, "desiredSc", desiredSc)
-			if err := r.handleStorageClassChange(ctx, dataImportCron, desiredSc); err != nil {
+		if outdated {
+			if err := r.recreateImport(ctx, dataImportCron, desiredSc); err != nil {
 				return res, err
 			}
 			return reconcile.Result{RequeueAfter: time.Second}, nil
 		}
-		cc.AddAnnotation(dataImportCron, AnnStorageClass, desiredStorageClass.Name)
+		cc.AddAnnotation(dataImportCron, AnnStorageClass, desiredSc)
 	}
 	format, err := r.getSourceFormat(ctx, desiredStorageClass)
 	if err != nil {
@@ -974,7 +973,7 @@ func (p *authProxy) GetDataSource(namespace, name string) (*cdiv1.DataSource, er
 	return das, nil
 }
 
-func (r *DataImportCronReconciler) handleStorageClassChange(ctx context.Context, dataImportCron *cdiv1.DataImportCron, desiredStorageClass string) error {
+func (r *DataImportCronReconciler) recreateImport(ctx context.Context, dataImportCron *cdiv1.DataImportCron, desiredStorageClass string) error {
 	digest, ok := dataImportCron.Annotations[AnnSourceDesiredDigest]
 	if !ok {
 		// nothing to delete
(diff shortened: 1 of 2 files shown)


pkg/controller/dataimportcron-controller.go
```diff
@@ -359,19 +359,25 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
...
 		}
 		currentSc, hasCurrent := dataImportCron.Annotations[AnnStorageClass]
-		desiredSc := desiredStorageClass.Name
-		if hasCurrent && currentSc != desiredSc {
-			r.log.Info("Storage class changed, delete most recent source on the old sc as it's no longer the desired", "currentSc", currentSc, "desiredSc", desiredSc)
+		var storageClassChanged bool
+		if hasCurrent {
+			storageClassChanged = currentSc != desiredSc
+		} else {
+			storageClassChanged = len(dataImportCron.Status.CurrentImports) > 0
```
arnongilboa: `len(imports) > 0` ?

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -1844,6 +1844,50 @@ var _ = Describe("All DataImportCron Tests", func() {
 				Expect(dv.Spec.Storage.AccessModes).To(BeEmpty())
 			})
 
+			It("Should delete old DV and requeue when first default StorageClass is set", func() {
```
arnongilboa, line 1847: shouldn't it test the creation of a new dv?

pkg/controller/dataimportcron-controller.go
noamasu: yep, updating

pkg/controller/dataimportcron-controller.go
```diff
@@ -359,19 +359,25 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		return res, err
 	}
 	if desiredStorageClass != nil {
-		if deleted, err := r.deleteOutdatedPendingPvc(ctx, pvc, desiredStorageClass.Name, dataImportCron.Name); deleted || err != nil {
+		desiredSc := desiredStorageClass.Name
+		if deleted, err := r.deleteOutdatedPendingPvc(ctx, pvc, desiredSc, dataImportCron.Name); deleted || err != nil {
```
akalenyu: I'm not sure why the described issue ins't being handled by deleteOutdatedPendingPvc but that was going to solve the no sc->set sc

akalenyu (commented):
Bit confused about the issue

pkg/controller/dataimportcron-controller.go
arnongilboa: Because in the problematic scenario the DV was already created, and will prefer RWX as specified by the default StorageProfile, as it was created before the DIC controller reconciled to prefer RWO, so just deleting the PVC (and not recreating DV) will recreate the PVC again with RWX. I think we better refactor this code and do all the deletion magics in a single call deleting the DV.

akalenyu: I agree but for backportability why don't we just whack the DV in the deleteOutdatedPendingPvc path?

arnongilboa: `deleteOutdatedPendingPvc` returns immediately when pvc status != Pending, and also when the pvc SC is the default/desired one, so afaiu it won't help much in the mentioned scenario. @noamasu ?
https://github.com/kubevirt/containerized-data-importer/blob/33bb1597c9faf35555ae1d4bf670b26aeb74d7b9/pkg/controller/dataimportcron-controller.go#L1536-L1547

noamasu: when SC is nil (and no accessmode specificed in DIC), the DV is stuck with error "`PVC spec is missing accessMode and no storageClass to choose profile`"  and this happens before the DV's PVC is created...
because the PVC is nil, the logic in deleteOutdatedPendingPvc is skipped...

if we want to put the DV deletion in deleteOutdatedPendingPvc, we might need to do some extra changes there (and also change the this func name to reflect DV deletion as well)
I can go to that direction if you think this is worth it 

I tried to keep the fix as contained as possible, but i can refactor if needed

akalenyu: Thinking about it some more, I think we're better off handling a DV spec mismatch at all times, similarly to the snapshot thing (handleSnapshotClassChange). The reasoning is that I don't think this type of issues will be limited to the storage class switching scenario

noamasu: the thing with default SC is that the DV spec are the same for current and desired DV in case a default SC is being set, therefore a DV spec mismatch will probably wont detect that edge case change.. (same for the outdated PVC edge case, you still need to check the PVC's SC name and compare it to desired SC... what we can do is instead of removing the PVC, we can remove the DV instead).

would you prefer to keep that fix contained with what we have right now and later open another PR for a broader mismatch refactor, or would you rather do the refactor in this PR? (which again, will probably be unrelated to this specific nil to default SC case... we will probably still need to track down the `desiredStorageClass` and decide if to delete DV according to its changes).

arnongilboa: The desired DV spec will be created explicitly with RWO, so `reflect.DeepEqual()` should detect the change, right?
https://github.com/kubevirt/containerized-data-importer/blob/4e4aaa33701674618a4ee1ec9783a89c6be6fe60/pkg/controller/dataimportcron-controller.go#L1797-L1803

akalenyu: I think I remember that reconciling over the DV desired != current was trickier to implement but as @arnongilboa said it's definitely doable. Maybe snap has to carry source access mode so you know when to reconcile

pkg/controller/dataimportcron-controller.go
```diff
@@ -1111,6 +1110,37 @@ func (r *DataImportCronReconciler) handleSnapshotClassChange(ctx context.Context
 	return true, nil
 }
 
+func (r *DataImportCronReconciler) importOutdated(ctx context.Context, cron *cdiv1.DataImportCron, currentDV *cdiv1.DataVolume, pvc *corev1.PersistentVolumeClaim, desiredSC *storagev1.StorageClass) (bool, error) {
```
arnongilboa: maybe rename to isImportOutdated() like other bool helpers?

pkg/controller/dataimportcron-controller.go
```diff
@@ -1111,6 +1110,37 @@ func (r *DataImportCronReconciler) handleSnapshotClassChange(ctx context.Context
...
+	currentSc, hasCurrent := cron.Annotations[AnnStorageClass]
+	scChanged := currentSc != desiredSC.Name
+	needsReset := hasCurrent || len(cron.Status.CurrentImports) > 0
```
arnongilboa: reset is a bit unclear, maybe reimportRequired?

pkg/controller/dataimportcron-controller.go
```diff
@@ -359,19 +359,18 @@ func (r *DataImportCronReconciler) update(ctx context.Context, dataImportCron *c
 		return res, err
 	}
 	if desiredStorageClass != nil {
-		if deleted, err := r.deleteOutdatedPendingPvc(ctx, pvc, desiredStorageClass.Name, dataImportCron.Name); deleted || err != nil {
+		desiredSc := desiredStorageClass.Name
+		outdated, err := r.importOutdated(ctx, dataImportCron, dv, pvc, desiredStorageClass)
```
arnongilboa: all importOutdated() internal refs are to the desiredStorageClass.Name, so pass it instead of the SC

pkg/controller/dataimportcron-controller.go
noamasu: i agree, changing. thanks!

pkg/controller/dataimportcron-controller.go
akalenyu: I see we took a hybrid approach instead of strictly focusing on the spec drift, which makes me worry a little since we have to reason about the broader introduced condition.

IMO, the backport-friendly version of this change is to strictly handle spec drifts on the DV and act accordingly, and a follow up to `main` for a bigger refactor.

That said, I fully acknowledge that I am on the cautious side, and if we can get a 2-1 vote on this, I am willing to proceed.

akalenyu (commented):
main worry in https://github.com/kubevirt/containerized-data-importer/pull/4074#discussion_r3512076308 🙏

akalenyu:
I think the destructive lane failure is related https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/4074/pull-containerized-data-importer-e2e-destructive/2072598145111429120

PR #4075, opened 2026-03-26
tests/cloner_test: use patch instead of update to override minimumSupportedPvcSize annotation

Acedus:
`createStorageWithMinimumSupportedPVCSize` was creating a StorageClass and sometimes failing on annotating the associated StorageProfile via an update call due to a conflict. Use merge patch instead to avoid conflicts altogether.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -21,6 +21,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/utils/ptr"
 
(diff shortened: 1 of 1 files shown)


tests/cloner_test.go
```diff
@@ -3127,14 +3128,15 @@ func createStorageWithMinimumSupportedPVCSize(f *framework.Framework, minSize st
...
+		_, err = f.CdiClient.CdiV1beta1().StorageProfiles().Get(context.TODO(), sc.Name, metav1.GetOptions{})
 		return err
 	}, time.Minute, time.Second).Should(Succeed())
 
-	sp.Annotations = map[string]string{controller.AnnMinimumSupportedPVCSize: minSize}
-	_, err = f.CdiClient.CdiV1beta1().StorageProfiles().Update(context.TODO(), sp, metav1.UpdateOptions{})
+	patch := fmt.Sprintf(`{"metadata":{"annotations":{%q:%q}}}`, controller.AnnMinimumSupportedPVCSize, minSize)
+
+	_, err = f.CdiClient.CdiV1beta1().StorageProfiles().Patch(context.TODO(),
+		sc.Name, types.MergePatchType, []byte(patch), metav1.PatchOptions{})
```
akalenyu: I think you want strategic merge patch, but that doesn't work for custom resources, so I would just go for regular JSON patch. I think this will override existing annotations as is

akalenyu:
I think I lied about the merge patch being an issue. We should be able to go with it safely since annotations is a map and not a list. Only lists are a pain point with merge patch

Acedus:
> I think I lied about the merge patch being an issue. We should be able to go with it safely since annotations is a map and not a list. Only lists are a pain point with merge patch

I think I'm just going to use SSA and call it a day (maybe one day we'll even have applyconfig for CDI types).

akalenyu:
/override pull-containerized-data-importer-fossa
/cc @awels 
I think FOSSA is lying about this one

PR #4082, opened 2026-04-02
[release-v1.64] Enable go multi-module workspace (#4035)

awels:
Manual backport of #4035

```release-note
NONE
```


Acedus:
Is the 509c81e2bcd8517593bfcab6f3e89edea3167a1f cherry-pick intentional?

hack/build/config.sh
```diff
@@ -29,7 +29,7 @@ FUNC_TEST_PROXY="cdi-func-test-proxy"
 FUNC_TEST_POPULATOR="cdi-func-test-sample-populator"
 
 # update this whenever new builder tag is created
-BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder:2503021312-67d1f3d4d}
+BUILDER_IMAGE=${BUILDER_IMAGE:-quay.io/kubevirt/kubevirt-cdi-bazel-builder:2603091904-cadcae0a1}
```
akalenyu, line 32: I think we want to cherry pick the builder change, let a 64 builder publish after merge, and then switch to that. Otherwise we could get into weird maintenance territory

awels:
> Is the [509c81e](https://github.com/kubevirt/containerized-data-importer/commit/509c81e2bcd8517593bfcab6f3e89edea3167a1f) cherry-pick intentional?

It is, I was surprised 1.64 didn't have the rules change already, was that intentional? If so then we need to modify some of this.

hack/build/config.sh
awels: We are going to have to backport this stuff a LONG LONG way I didn't want to have to build separate builders for each, but if that is what it takes.

Acedus:
> > Is the [509c81e](https://github.com/kubevirt/containerized-data-importer/commit/509c81e2bcd8517593bfcab6f3e89edea3167a1f) cherry-pick intentional?
> 
> It is, I was surprised 1.64 didn't have the rules change already, was that intentional? If so then we need to modify some of this.

We didn't upgrade Bazel to 6.5.0 in 1.64 so there was no need to make that change.

awels:
> > > Is the [509c81e](https://github.com/kubevirt/containerized-data-importer/commit/509c81e2bcd8517593bfcab6f3e89edea3167a1f) cherry-pick intentional?
> > 
> > 
> > It is, I was surprised 1.64 didn't have the rules change already, was that intentional? If so then we need to modify some of this.
> 
> We didn't upgrade Bazel to 6.5.0 in 1.64 so there was no need to make that change.

It is probably because I didn't make a new builder, and the new builder for 1.65 does use the new bazel... So I guess I am stuck making builders...

hack/build/config.sh
akalenyu: I don't think it's going to be maintainable otherwise, we can't use the main builder for older branches, it'll physically fail. It should all be automated so I don't expect much manual work

akalenyu: Alternatively, can we skip the controller-gen bump (and CRD description changes) on this branch and stick to it's existing builder?
Might make all of this easier, but not sure if it's possible.

PR #4088, opened 2026-04-08
Fix CDI node-placement test for cross-architecture clusters

carterpewpew:

The node-placement test was using runtime.GOARCH to set kubernetes.io/arch in the nodeSelector. This value reflects the architecture of the machine running the test binary, not the architecture of the cluster's nodes. When the test binary ran on an amd64 host while the cluster nodes were s390x, the nodeSelector became incorrect, making CDI infra pods unschedulable. This change removes the dependency on runtime.GOARCH by deriving the nodeSelector from actual cluster nodes using kubernetes.io/hostname, mirroring the existing NodeAffinity configuration.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/framework/nodeplacement.go
+++ b/tests/framework/nodeplacement.go
@@ -2,7 +2,6 @@ package framework
 
 import (
 	"context"
-	"runtime"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


tests/framework/nodeplacement.go
```diff
@@ -12,7 +11,7 @@ import (
 
 var (
 	//NodeSelectorTestValue is nodeSelector value for test
-	NodeSelectorTestValue = map[string]string{"kubernetes.io/arch": runtime.GOARCH}
+	NodeSelectorTestValue = map[string]string{"kubernetes.io/hostname": ""}
```
awels: I hate to be a pain, could you make the bot happy by defining a constant for kubernetes.io/hostname and using that constant in the 3 places in this file? Other than that it looks great.

carterpewpew: Done! Extracted a hostnameLabel constant and updated all 3 references. Thanks for the review!

cfilleke:
This is great to see. Thank you @carterpewpew !!

PR #4089, opened 2026-04-08
rbac: add resourceclaims/driver associated-node verbs for sriov DRA

VaishnavSreekumar:
Adds resourceclaims/driver permissions with associated-node scoped verbs for the SR-IOV DRA driver.

The driver runs as a DaemonSet and performs ResourceClaim status updates, which require additional authorization under the DRA granular status authorization changes introduced in Kubernetes v1.36.

This change adds:
- associated-node:update
- associated-node:patch

to the ClusterRole while keeping existing resourceclaims/status permissions intact.

```release-note
Adds resourceclaims/driver permissions with associated-node scoped verbs for the SR-IOV DRA driver to support Kubernetes v1.36 granular status authorization.
```

### DIFF
--- a/cluster-up/cluster/kind-sriov/sriov-components/manifests/dra/sriov_dra.yaml
+++ b/cluster-up/cluster/kind-sriov/sriov-components/manifests/dra/sriov_dra.yaml
@@ -115,6 +115,11 @@ rules:
 - apiGroups: ["resource.k8s.io"]
   resources: ["resourceclaims/status"]
   verbs: ["get","list","update","patch"]
+- apiGroups: ["resource.k8s.io"]
+  resources: ["resourceclaims/driver"]
+  verbs:
+    - associated-node:update
+    - associated-node:patch
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["get", "list", "watch"]  # Cluster-scoped resource, needs cluster permissions


VaishnavSreekumar:
/remove-lifecycle stale

Gentle ping — this PR updates the SR-IOV DRA RBAC permissions to align with the Kubernetes v1.36 granular status authorization changes for ResourceClaims. Happy to make any adjustments if needed.

awels:
Can you remind me why this would be interesting for CDI?

VaishnavSreekumar:
The motivation here was mainly the Kubernetes v1.36 DRA granular authorization changes around ResourceClaim status updates.

CDI currently vendors/deploys the SR-IOV DRA driver components, and the driver performs driver-scoped ResourceClaim status updates during allocation/lifecycle handling. With the newer authorization model, those operations now require the `resourceclaims/driver` resource with the `associated-node` scoped `update`/`patch` verbs in addition to the existing `resourceclaims/status` permissions.

My understanding was that without these permissions, SR-IOV DRA-backed environments could run into authorization failures when the driver attempts node-associated claim status updates.

So this PR was mainly intended to keep the RBAC aligned with the upstream Kubernetes v1.36 DRA authorization changes.

awels:
Oh I see the confusion now. You are referencing the cluster-up/kind-sriov code. We pull in all the contents from cluster-up from kubevirtci by running `hack/update-kubevirtci.sh` in that script we first update `kubevirtci_release_tag` to be the latest we want, then run the script to replace the contents of cluster-up. This would then automatically pull in anything that is needed from kubevirtci.

In short, first make sure kubevirtci has the right permissions, then run the script, and create a PR out of that. Anything manually added will get overwritten when we run that script.

VaishnavSreekumar:
Thanks for the clarification, @awels. I missed that these manifests were being synced from kubevirtci rather than maintained directly here.

I’ll close this PR for now and look into updating it from the kubevirtci side instead.

PR #3744, opened 2025-05-12
Append Labels From DV to Importer Pod

Dsanatar:
Fixes #3674 

During the creation of the Importer Pod, append any labels from the PVC. This allows for labels to be passed from a Data Volume.

```release-note
Labels on DataVolumes are now copied to the importer pod
```

Issue #3674, opened 2025-03-12
Can we allow for custom labels on the importer pod, or at least add istio.io/dataplane-mode: none to support Ambient Mesh


jbilliau-rcd:
We are running CDI `1.59.0` on Kubevirt `1.3.0`. Things have been going well until we start testing with Ambient Mesh using Istio. CDI already accounts for *sidecar-based* Istio with the annotation of `sidecar.istio.io/inject: 'false'` on the importer-pod by default, which makes the transfer bypass the Istio sidecar:

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cdi.kubevirt.io/storage.createdByController: 'yes'
    linkerd.io/inject: disabled
    sidecar.istio.io/inject: 'false'
  labels:
    app: containerized-data-importer
    app.kubernetes.io/component: storage
    app.kubernetes.io/managed-by: cdi-controller
    cdi.kubevirt.io: importer
    prometheus.cdi.kubevirt.io: 'true'
  name: importer-prime-16bec903-1771-4e03-bad7-fa2029ccd30b
  namespace: test-vm
```

Problem is, Ambient Mesh uses a different mechanism, a **label** of `istio.io/dataplane-mode: none`. Since there is no mechanism to put this on, the importer pod launches, runs for around 2-5 min (it varies), and eventually blows up with this error:

```
I0311 18:47:50.961109       1 importer.go:103] Starting importer
I0311 18:47:50.962691       1 importer.go:173] begin import process
I0311 18:47:50.962752       1 s3-datasource.go:149] Endpoint rcd1kubes3clu1.mi.corp.xxxxx.com
I0311 18:47:50.962775       1 s3-datasource.go:153] bucket templates
I0311 18:47:50.962779       1 s3-datasource.go:154] object win2022-gold.img.gz
I0311 18:47:50.979721       1 http-datasource.go:263] Attempting to get certs from /certs/ca.pem
I0311 18:47:51.054017       1 data-processor.go:358] Calculating available size
I0311 18:47:51.055005       1 data-processor.go:366] Checking out block volume size.
I0311 18:47:51.055024       1 data-processor.go:378] Request image size not empty.
I0311 18:47:51.055038       1 data-processor.go:383] Target size 85Gi.
I0311 18:47:51.055289       1 data-processor.go:257] New phase: TransferDataFile
I0311 18:47:51.056493       1 util.go:188] Writing data...
[...]

PR #4091, opened 2026-04-13
fix: Label Copy for Importer Pods

### LINKED ISSUE #3988: Clarification on copying labels from source PVC to prime PVC
**What happened**:
We noticed that CDI copies labels from the original PVC to the prime PVC, which is a temporary object created during the import process. Could you clarify the reasoning behind this behavior?

Prime PVC should not inherit user-defined labels unless necessary for CDI operations.

Create a PVC with custom labels (e.g., app=myvm).
Trigger an import using CDI (e.g., DataVolume or VolumeImportSource with source: http://quay.io/nirsof/cirros:0.6.2-1).
Observe that the prime PVC created during the process inherits the labels from the original PVC.

**Additional context**:
Prime PVC inherits all labels from the original PVC.

- Is this intentional and required for CDI functionality?
- Could this lead to unintended side effects if external controllers interpret these labels?
- Would it make sense to avoid copying user-defined labels?

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): `image: quay.io/kubevirt/cdi-controller:v1.63.0`
```
# kubectl version
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.34.0
```
- DV specification: N/A
- OS (e.g. from /etc/os-release): RHEL9.6
- Install tools: N/A
- Others: N/A

CC: [nirsof@gmail.com](mailto:nirsof@gmail.com)

Dsanatar:

Issue found with how previous implementation https://github.com/kubevirt/containerized-data-importer/pull/3744 copied labels to Importer Pod by means of the prime pvc. Instead of copying form target pvc -> prime pvc -> pod, skip copying to the prime pvc and instead get the labels from target pvc by using the prime pvc's owner ref.

Fixes #3988

**Release note**:

```release-note
Copy Labels from PVC to Importer Pod
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -910,8 +910,22 @@ func createImporterPod(ctx context.Context, log logr.Logger, client client.Clien
 
 	util.SetRecommendedLabels(pod, installerLabels, "cdi-controller")
 
+	srcLabels := args.pvc.Labels
+	if _, isPopulator := args.pvc.Annotations[cc.AnnPopulatorKind]; isPopulator {
+		ownerRef := metav1.GetControllerOf(args.pvc)
+		if ownerRef == nil || ownerRef.Kind != "PersistentVolumeClaim" {
+			return nil, fmt.Errorf("pvc %s/%s does not have a valid owner reference", args.pvc.Namespace, args.pvc.Name)
+		}
+
+		pvc := &corev1.PersistentVolumeClaim{}
+		if err := client.Get(context.TODO(), types.NamespacedName{Namespace: args.pvc.Namespace, Name: ownerRef.Name}, pvc); err != nil {
+			return nil, err
+		}
+		srcLabels = pvc.GetLabels()
+	}
+
 	// add any labels from pvc to the importer pod
-	util.MergeLabels(args.pvc.Labels, pod.Labels)
+	util.MergeLabels(srcLabels, pod.Labels)
 
 	if err = client.Create(context.TODO(), pod); err != nil {
 		return nil, err
--- a/pkg/controller/import-controller_test.go
+++ b/pkg/controller/import-controller_test.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	kvalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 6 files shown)


Dsanatar:
/cc @akalenyu 
I saw your suggestion about using the `AnnEventSource` logic for this but it seems we can get what we need from the ownerRef of the prime pvc just as easily. thoughts?

Acedus (commented):
Looks good! Can we add a UT for this?

akalenyu (commented):
> /cc @akalenyu I saw your suggestion about using the `AnnEventSource` logic for this but it seems we can get what we need from the ownerRef of the prime pvc just as easily. thoughts?

This looks good tbh, quite simple

I didn't realize we only ever tackled importer pods (upload/clone omitted), is that desired?

BTW, would you mind converting the e2e tests to unit tests https://github.com/kubevirt/containerized-data-importer/blob/2c88a8aaffd6af3b51bc0d4b3a4c564ffa2a5918/tests/import_test.go#L2100 I don't think there's much value over unit in this case

Dsanatar:
@akalenyu  ack, just converted the func test to unit tests

> I didn't realize we only ever tackled importer pods (upload/clone omitted), is that desired?
> 
The original issue was only related to importer pods https://github.com/kubevirt/containerized-data-importer/issues/3674 however if the need is there, it shouldn't be too difficult to implement this for upload and clone as well

akalenyu:
/test pull-cdi-unit-test
/test pull-cdi-goveralls
unrelated/quay down

akalenyu (commented):
/approve

thank you, FYI @nirs

pkg/controller/populators/populator-base.go
```diff
@@ -231,7 +231,6 @@ func (r *ReconcilerBase) createPVCPrime(pvc *corev1.PersistentVolumeClaim, sourc
 	}
 	cc.CopyAllowedAnnotations(pvc, pvcPrime)
 	util.SetRecommendedLabels(pvcPrime, r.installerLabels, "cdi-controller")
-	util.MergeLabels(pvc.Labels, pvcPrime.Labels)
```
nirs, line 234: This merged labels from target pvc into the prime pvc?

pkg/controller/import-controller.go
```diff
@@ -941,8 +910,22 @@ func createImporterPod(ctx context.Context, log logr.Logger, client client.Clien
...
+		pvc := &corev1.PersistentVolumeClaim{}
+		if err := client.Get(context.TODO(), types.NamespacedName{Namespace: args.pvc.Namespace, Name: ownerRef.Name}, pvc); err != nil {
+			return nil, err
+		}
+		srcLabels = pvc.GetLabels()
+	}
+
 	// add any labels from pvc to the importer pod
-	util.MergeLabels(args.pvc.Labels, pod.Labels)
+	util.MergeLabels(srcLabels, pod.Labels)
```
nirs, line 928: And this merges labels from the target pvc into the pod?

So the prime pvc does copy application labels now, right?

nirs (commented):
Can you share example prime pvc before this change and after this change?

pkg/controller/import-controller.go
Dsanatar: yes, that is correct. the original implementation allowed for custom labels to be passed from PVCs to importer pods. As you pointed out, when we are using `DataVolumes` (or populators) we were doing this by copying the labels from the target PVC to the prime pvc so that we could merge them when creating the importer pod.

However, these changes make it so we skip copying the labels to the prime pvc

pkg/controller/import-controller_test.go
```diff
@@ -907,6 +908,64 @@ var _ = Describe("Create Importer Pod", func() {
...
+		}
+
+		reconciler := createImportReconciler(objects...)
+
+		pod, err := createImporterPod(context.TODO(), reconciler.log, reconciler.client, podArgs, map[string]string{})
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Verifying pod has labels from target PVC")
+		Expect(pod.Labels).To(HaveKeyWithValue("custom-label", "value"))
+		if isPopulator {
```
alromeros, line 960: Maybe also add a check to ensure the prime PVC doesn’t have the target PVC labels?

Dsanatar: yea good point. Since that requires reconciling at the populator layer, I added a new UT there

nirs:
@Dsanatar which upstream kubevirt version will include this fix?

Dsanatar:
@nirs the linked issue https://github.com/kubevirt/containerized-data-importer/issues/3988 mentions using CDI v1.63 so this fix will be backported to that version. The equivalent kubevirt release is version 1.6

akalenyu:
@nirs we would have to make a release that includes this commit in all those versions. maybe we could start with just 1.65.1?

nirs:
Including the fix in the latest version would be good enough for us.

Issue #4000, opened 2025-12-30
Feature Request: Support DV-level scheduling constraints (nodeSelector/affinity/tolerations) for importer pods


jimyag:
## Is your feature request related to a problem? Please describe

  Currently, scheduling constraints for CDI workload pods (importer, cloner, uploader) can only be configured globally via `CDI.spec.workload`. There is no way to specify different scheduling requirements for individual DataVolumes.

  When using local storage provisioners, we need to control which node the importer pod runs on at the DV level.

  ## Describe the solution you'd like

  Add a `workloadPlacement` field to DataVolumeSpec, limited to scheduling-related fields only:

```yaml
  apiVersion: cdi.kubevirt.io/v1beta1
  kind: DataVolume
  spec:
    source: ...
    pvc: ...
    # New field - limited to scheduling only
    workloadPlacement:
      nodeSelector: {...}
      affinity: {...}
      tolerations: [...]
```
  This approach:
  1. Only exposes scheduling fields, no env, volumes, or other risky fields
  2. Reuses the same NodePlacement struct pattern as CDI.spec.workload
  3. Optional field; existing DVs continue to use global config

  Precedence: DV-level workloadPlacement overrides CDI.spec.workload for that specific DataVolume. If DV-level is not specified, fall back to global config.

  Describe alternatives you've considered

  | Solution                                        | Limitation                                                                                |
  |-------------------------------------------------|-------------------------------------------------------------------------------------------|
  | CDI.spec.workload                               | Cluster-wide only; cannot vary per DataVolume                                             |
  | WaitForFirstConsumer                            | Waits for consumer pod, but virt-launcher won't start until DV is ready (chicken-and-egg) |
  | PV node affinity                                | Still in alpha; not considered production-ready                                           |
  | Third-party solutions (e.g., mutating webhooks) | Adds external dependencies and complexity                                                 |

  We believe it would be better to leverage the native capabilities of DataVolume and importer pod directly, rather than relying on external workarounds.

  Additional context

  This issue is split from https://github.com/kubevirt/containerized-data-importer/issues/3848 to discuss DV-level scheduling constraints separately from scratch space storage class configuration.

akalenyu:
@jimyag hey, can you elaborate a little about the use case? it sounds like wffc would've worked for your use case

jimyag:
> [@jimyag](https://github.com/jimyag) hey, can you elaborate a little about the use case? it sounds like wffc would've worked for your use case

[...]

PR #4096, opened 2026-04-19
Add VDDK node selector override via annotation

yaacov:
Introduce a new annotation `cdi.kubevirt.io/storage.pod.nodeSelector`
that allows overriding the CDI CR workload node selector for VDDK
importer pods. This enables scheduling VDDK imports on specific nodes
(e.g. ones with network access to the VMware host).

The annotation value is a JSON-encoded map[string]string that, when
present on the PVC, replaces the default NodeSelector from the CDI CR
while preserving tolerations and affinity. The override is applied only
when the import source is VDDK; all other import types and CDI pod types
(clone, upload, populator, cron) are unaffected.

Includes unit tests for annotation override, missing annotation
fallback, non-VDDK source ignoring the annotation, and invalid JSON
error handling.

Signed-off-by: yaacov <yzamir@redhat.com>

Introduce a new annotation `cdi.kubevirt.io/storage.pod.nodeSelector` that allows overriding the CDI CR workload node selector on a per-PVC basis. The annotation value is a JSON-encoded map[string]string that, when present, replaces the default NodeSelector from the CDI CR while preserving tolerations and affinity.

Fixes CNV-84587, MTV-3207

Ref:
https://redhat.atlassian.net/browse/CNV-84587
https://redhat.atlassian.net/browse/MTV-3207

Similar to #4000 

Read PVC annotations, DV annotation is copied into PVC so users can also anotatate the DV 

**Release note**:

```release-note
Add per-PVC node selector override via annotation
```

### DIFF
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"path"
@@ -119,6 +120,7 @@ type importerPodArgs struct {
 	workloadNodePlacement   *sdkapi.NodePlacement
 	vddkImageName           *string
 	vddkExtraArgs           *string
+	vddkNodeSelector        map[string]string
 	priorityClassName       string
 	serviceAccountName      string
 }
@@ -531,6 +533,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 	var scratchPvcName *string
 	var vddkImageName *string
 	var vddkExtraArgs *string
+	var vddkNodeSelector map[string]string
 	var err error
 
 	requiresScratch := r.requiresScratchSpace(pvc)
@@ -563,6 +566,11 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 		if extraArgs, ok := anno[cc.AnnVddkExtraArgs]; ok && extraArgs != "" {
 			r.log.V(1).Info("Mounting extra VDDK args ConfigMap to importer pod", "ConfigMap", extraArgs)
 			vddkExtraArgs = &extraArgs
+
+			vddkNodeSelector, err = r.getVddkNodeSelector(pvc.Namespace, extraArgs)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
@@ -580,6 +588,7 @@ func (r *ImportReconciler) createImporterPod(pvc *corev1.PersistentVolumeClaim) 
 		scratchPvcName:     scratchPvcName,
 		vddkImageName:      vddkImageName,
 		vddkExtraArgs:      vddkExtraArgs,
+		vddkNodeSelector:   vddkNodeSelector,
 		priorityClassName:  cc.GetPriorityClass(pvc),
 		serviceAccountName: cc.GetPodServiceAccount(pvc),
 	}
@@ -862,6 +871,28 @@ func (r *ImportReconciler) getVddkImageName() (*string, error) {
 	return nil, errors.Errorf("found %s ConfigMap in namespace %s, but it does not contain a '%s' entry", common.VddkConfigMap, namespace, common.VddkConfigDataKey)
 }
 
+// getVddkNodeSelector reads the optional node-selector key from the VDDK extra-args
+// ConfigMap and returns the parsed map, or nil if the key is absent.
+func (r *ImportReconciler) getVddkNodeSelector(namespace, cmName string) (map[string]string, error) {
+	cm := &corev1.ConfigMap{}
+	if err := r.uncachedClient.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: namespace}, cm); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	val, ok := cm.Data[common.VddkNodeSelectorKey]
+	if !ok || val == "" {
+		return nil, nil
+	}
+	nodeSelector := make(map[string]string)
+	if err := json.Unmarshal([]byte(val), &nodeSelector); err != nil {
+		return nil, fmt.Errorf("failed to parse %s key in ConfigMap %s/%s: %w",
+			common.VddkNodeSelectorKey, namespace, cmName, err)
+	}
+	return nodeSelector, nil
+}
+
 // returns the import image part of the endpoint string
 func getRegistryImportImage(pvc *corev1.PersistentVolumeClaim) (string, error) {
 	ep, err := cc.GetEndpoint(pvc)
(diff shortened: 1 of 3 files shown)


pkg/controller/populators/forklift-populator.go
```diff
@@ -520,6 +520,14 @@ func (r *ForkliftPopulatorReconciler) createPopulatorPod(pvcPrime, pvc *corev1.P
 		Spec: makePopulatePodSpec(pvcPrime.Name, secretName),
 	}
 
+	workloadNodePlacement, err := cc.GetWorkloadNodePlacement(context.TODO(), r.client, pvc)
```
yaacov: @mrnold hi, i am new to the forklift populator, is this a correct way to set the node placement for this populator ?

akalenyu (commented):
Hey, thanks for the PR! some thoughts;

- I don't think its trivial to allow regular users (non admin) to influence pod scheduling decisions. ~~Some security implications at the very least~~
EDIT: actually, any regular user can place node selector, so I don't think the security argument holds
- Have we considered trying to use WaitForFirstConsumer binding mode/AllowedTopologies setup to achieve this use case? it's kind of similar in that sense, but instead of the piece of storage being topology constrained, the data that goes to it is.
https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies
- ~~Maybe if we must go this path, we could extend the workload placement global API to set the placement setting according to certain labels, instead of a json ann value~~
EDIT: why don't we just expose an API on the DataVolume, similarly to priorityClassName

If you want to join the sig-storage call tomorrow, we could discuss this in length

yaacov:
@akalenyu hi, thank you for your review, I updated the PR based on your feedback, please re-review.

What changed:
The AnnPodNodeSelector annotation constant remains in common/util.go as a shared definition, but the logic that acts on it is isolated to the VDDK importer path in import-controller.go. Non-VDDK imports and all other CDI pod types (clone, upload, populator, cron) are completely unaffected.

> I don't think its trivial to allow regular users (non admin) to influence pod scheduling decisions.

I changed the scope of the change, now the annotation control *only* the vddk importer pod, and does not affect other pods CDI controller may want to spread according to other preferences.

> Have we considered trying to use WaitForFirstConsumer binding mode/AllowedTopologies setup to achieve this use case?

This raises concerns from MTV side, we want to consume CDI and delegate all importing logic, including node placement, to CDI. Changing WaitForFirstConsumer will require implementing a big chunk of data importing logic into MTV.

> why don't we just expose an API on the DataVolume, similarly to priorityClassName

Committing to an API change require touching lots of files, we are also not sure other importers not used by MTV require this option ( I don't know who is the owner #4000 and what are the use case )

pkg/controller/import-controller.go
```diff
@@ -906,6 +907,22 @@ func createImportPodNameFromPvc(pvc *corev1.PersistentVolumeClaim) string {
...
+	val, ok := pvc.GetAnnotations()[cc.AnnPodNodeSelector]
+	if !ok || val == "" {
+		return placement, nil
+	}
+	nodeSelector := make(map[string]string)
+	if err := json.Unmarshal([]byte(val), &nodeSelector); err != nil {
+		return nil, fmt.Errorf("failed to parse %s annotation: %w", cc.AnnPodNodeSelector, err)
+	}
+	out := placement.DeepCopy()
+	out.NodeSelector = nodeSelector
```
akalenyu: I think you want to merge instead of override, otherwise you disregard the global ones set by admin

yaacov: makes sense, I was also concerned about conflicting label selectors.

But, it makes sense to respect the system selector and in case of conflicts to fail, and look into the conflict, instead of silently override system settings.

changed to merge instead of override.

akalenyu:
sig-storage summary; can we maybe use secondary networks to achieve the use case?

trying to see if we can postpone having CDI know about this kind of stuff

pkg/controller/common/util.go
```diff
@@ -97,6 +97,8 @@ const (
 	AnnPriorityClassName = AnnAPIGroup + "/storage.pod.priorityclassname"
 	// AnnPodServiceAccount is a PVC annotation to indicate the service account name for importer and uploader pod
 	AnnPodServiceAccount = AnnAPIGroup + "/storage.pod.serviceAccountName"
+	// AnnPodNodeSelector is a PVC/DV annotation to override the pod node selector (JSON-encoded map[string]string)
+	AnnPodNodeSelector = AnnAPIGroup + "/storage.pod.nodeSelector"
```
akalenyu: @aglitke mentioned this should probably have something about vddk to signal it's the only source that's supporting this

yaacov: moved the vddk node selector information into the vddk extra args config map, no change needed to the DV of PVC created, all configuration changes are now confined to vddk specific config map

yaacov:
@akalenyu 

> trying to see if we can postpone having CDI know about this kind of stuff

I tried to answer this concern, CDI still know about the node selector, but in a very distinct VDDK only flow, 
The information is in a the vddk config map we already use, if the config map exist and we have node selector information, we merge this node selector request for vddk importer pods, does that make better sense now ?

akalenyu:
> @akalenyu
> 
> > trying to see if we can postpone having CDI know about this kind of stuff
> 
> I tried to answer this concern, CDI still know about the node selector, but in a very distinct VDDK only flow, The information is in a the vddk config map we already use, if the config map exist and we have node selector information, we merge this node selector request for vddk importer pods, does that make better sense now ?

Yeah this is better. I still think @awels' direction about secondary networks is the appropriate way to go about this use case though. Do you want to explore that, or would you like to proceed with this (@awels wdyt?)

yaacov:
> Do you want to explore that, or would you like to proceed with this

My task is setting the nodeSelector, the network is a good use case, but does not cover the requested feature.

yaacov:
>  the network is a good use case, but does not cover the requested feature.

@mnecas hi, do you think we can address the request using network settings?

cc: //@awels

mnecas:
Hi @akalenyu @awels 
the main idea is that in some cases there is dedicated NIC on one Node to connect to outside enviroment to perform the disk copy. For this we need to have the nodeselector and secondary netwrok.
- secondary network on the pod: I checked that CDI is copying the annotations `k8s.v1.cni.cncf.io/networks`  and `v1.multus-cni.io/default-network`  from DVs to PVCs and from PVCs to populator pods.
- nodeselector is needed to pin the populator pod to specific node otherwise the pod will fail to start on a node that is not configured with that dedicated NIC 

Alternative to all of this would be purly networking OVNK EgressIP but that has lot of performance issues. https://ovn-kubernetes.io/features/cluster-egress-controls/egress-ip/

Hope it clarifies it, if there are any other questions or suggestions please let me know

akalenyu:
> Hi @akalenyu @awels the main idea is that in some cases there is dedicated NIC on one Node to connect to outside enviroment to perform the disk copy. For this we need to have the nodeselector and secondary netwrok.
> 
> * secondary network on the pod: I checked that CDI is copying the annotations `k8s.v1.cni.cncf.io/networks`  and `v1.multus-cni.io/default-network`  from DVs to PVCs and from PVCs to populator pods.
> * nodeselector is needed to pin the populator pod to specific node otherwise the pod will fail to start on a node that is not configured with that dedicated NIC
> 
> Alternative to all of this would be purly networking OVNK EgressIP but that has lot of performance issues. https://ovn-kubernetes.io/features/cluster-egress-controls/egress-ip/
> 
> Hope it clarifies it, if there are any other questions or suggestions please let me know

Interesting, so the secondary network alone won't influence the scheduling to pick the correct node?

Anyway, I think the current state of the PR is really pinpointed to VDDK so I think we could proceed.
@awels If you think that is okay, could you unhold?
/approve
/hold

awels (approved):
/unhold

akalenyu:
/test pull-cdi-unit-test
unrelated issue, we should remove prlimit

akalenyu:
ci issue

akalenyu:
nfs died

PR #4097, opened 2026-04-20
DNM: tests: deprecate DV spec.pvc, use spec.storage instead

Acedus:
Altering volume size for DVs with an explicitly provided PVC spec is something that's only done by requesting PVC mutation using the `cdi.kubevirt.io/applyStorageProfile` annotation. In order to run tests on clusters with certain storage restrictions that are remediated via the storage profile API, use spec.storage wherever possible.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/datavolume.go
+++ b/tests/utils/datavolume.go
@@ -153,13 +153,9 @@ func CleanupDvPvcNoWait(k8sClient *kubernetes.Clientset, cdiClient *cdiclientset
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 }
 
-func NewCloningDataVolume(dataVolumeName, size string, sourcePvc *k8sv1.PersistentVolumeClaim) *cdiv1.DataVolume {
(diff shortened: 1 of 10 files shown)


arnongilboa:
Looks promising. So `spec.pvc` will be covered only in utests?

tests/utils/datavolume.go
```diff
@@ -468,32 +427,9 @@ func NewDataVolumeCloneToBlockPV(dataVolumeName string, size string, srcNamespac
 	return dataVolume
 }
 
-// NewDataVolumeCloneToBlockPVStorageAPI initializes a DataVolume for block cloning with Storage API
+// Deprecated: Use NewDataVolumeCloneToBlockPV instead which uses the storage API
```
alromeros: I understand keeping deprecated functions temporarily if the old functionality is kept. But this PR updates both the calls and the deprecated functions, which now have identical implementations to the functions they wrap. I believe in this case it would be cleaner to simply remove them.

Acedus: Yeah, wrote the comments before I decided to just remove all existing use, let me do that.

PR #4098, opened 2026-04-20
Fix clone to filesystem PVC sizing when storageProfile minimum applies

arnongilboa:
When cloning to filesystem, apply the storageProfile effective volume size (minimum supported PVC size) before inflating with filesystem overhead. It helps on cloning from block to filesystem where both have minimum supported size but the target also needs filesystem overhead on top of that, otherwise the pre-clone checks fail.

jira-ticket: https://redhat.atlassian.net/browse/CNV-79503

**Release note**:
```release-note
Fix clone to filesystem PVC sizing when storageProfile minimum applies
```

### DIFF
--- a/tests/cloner_test.go
+++ b/tests/cloner_test.go
@@ -2216,7 +2216,7 @@ var _ = Describe("all clone tests", func() {
 
 		})
 
-		DescribeTable("Block volumeMode clone with target smaller than the source, using storgeProfile with minPvcSize annotation", Serial, func(minSize string, shouldSucceed bool) {
+		DescribeTable("Block volumeMode clone with target smaller than the source, using storageProfile with minPvcSize annotation", Serial, func(minSize string, shouldSucceed bool) {
 			if !f.IsBlockVolumeStorageClassAvailable() {
 				Skip("Storage Class for block volume is not available")
 			}
@@ -2249,7 +2249,7 @@ var _ = Describe("all clone tests", func() {
 			Entry("[test_id:12424] empty", "", false),
 		)
 
-		DescribeTable("Filesystem volumeMode clone with target smaller than the source, using storgeProfile with minPvcSize annotation", func(minSize string, shouldSucceed bool) {
+		DescribeTable("Filesystem volumeMode clone with target smaller than the source, using storageProfile with minPvcSize annotation", func(minSize string, shouldSucceed bool) {
 			By(fmt.Sprintf("Create source PVC %s", sourcePVCName))
 			sc := createStorageWithMinimumSupportedPVCSize(f, minSize)
 			sourcePvc = utils.NewPVCDefinition(sourcePVCName, "1Gi", nil, nil)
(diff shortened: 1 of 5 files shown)


Acedus:
fyi @arnongilboa https://github.com/kubevirt/containerized-data-importer/pull/3901

tests/cloner_test.go
```diff
@@ -2279,6 +2279,35 @@ var _ = Describe("all clone tests", func() {
 			Entry("[test_id:12423] empty", "", false),
 		)
 
+		It("[test_id:XXXXX] Block volumeMode clone to filesystem, using storageProfile with minPvcSize annotation", Serial, func() {
```
noamasu: need to add real test_id here

tests/smartclone_test.go
```diff
@@ -223,11 +223,11 @@ func verifyPVC(dataVolume *cdiv1.DataVolume, f *framework.Framework, testPath st
 	Expect(f.VerifyTargetPVCContentMD5(f.Namespace, targetPvc, testPath, md5sum, utils.UploadFileSize)).To(BeTrue())
 }
 
-func verifyPVCRequestedSize(dataVolume *cdiv1.DataVolume, f *framework.Framework, size string) {
+func verifyPVCRequestedSize(dataVolume *cdiv1.DataVolume, f *framework.Framework, minSize, comparator string) {
```
noamasu: maybe worth just adding another function like `verifyPVCRequestedSizeExceeds` for ">" instead of passing comparator?

tests/cloner_test.go
```diff
@@ -2279,6 +2279,35 @@ var _ = Describe("all clone tests", func() {
...
+			if !f.IsBlockVolumeStorageClassAvailable() {
+				Skip("Storage Class for block volume is not available")
+			}
+
+			By(fmt.Sprintf("Create source block DataVolume %s", dataVolumeName))
+			sc := createStorageWithMinimumSupportedPVCSize(f, "4Gi")
```
noamasu, line 2288: what if actual min size of the sc is larger than 4Gi?
for example if actual min size of the storage backend is 10Gi, this test will fail?

noamasu (commented):
thanks! for clone to fs, make sense to to add fsoverhead after effective size.
added some comments

tests/smartclone_test.go
arnongilboa: Agree

tests/smartclone_test.go
```diff
@@ -146,7 +146,7 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]SmartClone tests",
 		waitForDvPhase(cdiv1.Succeeded, dataVolume, f)
 		f.ExpectEvent(dataVolume.Namespace).Should(ContainSubstring(controller.CloneSucceeded))
 
-		verifyPVCRequestedSize(dataVolume, f, testMinPvcSize)
+		verifyPVCRequestedSize(dataVolume, f, testMinPvcSize, "==")
```
Ahmad-Hafe: With the new logic for filesystem clones (minimum applied first, then overhead), shouldn't the resulting PVC size be greater than the minimum rather than equal to it?

arnongilboa: Sure, but this test is for block.

tests/cloner_test.go
arnongilboa: You mean if the storage creates 10Gi in case of a smaller requested size (without using our annotation)? it will fail in the pre-clone check as the target will be too small, but I see no problem with such failure as we don't want to "guess" the backend behavior before the pvc is created. However, in a real use case, explicitly setting the minimum annotation to be aligned with the backup minimum size will solve this issue.

tests/cloner_test.go
arnongilboa: No need, removing.

PR #4100, opened 2026-04-23
Allow CDI_NAMESPACE to be set externally in test.sh

Davo911:
The `CDI_NAMESPACE` in `automation/test.sh` is unconditionally randomized, making it impossible to override from the environment. 
This can be problematic for CI jobs running on shared clusters like in https://github.com/kubevirt/project-infra/pull/4832

This change wraps the assignment with `${CDI_NAMESPACE:-...}` so that an externally set `CDI_NAMESPACE` is honored, while preserving the existing randomization as the default for all other jobs.

/cc @awels @dollierp 
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -24,7 +24,7 @@
 
 set -ex
 
-export CDI_NAMESPACE="cdi-$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 5 | head -n 1)"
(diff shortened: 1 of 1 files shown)


automation/test.sh
awels: Might want to remove this line then. Since it is pretty useless now.

Davo911: true, didn't catch that line

awels (approved):
/approve

PR #4109, opened 2026-04-30
fix: cluster wide certConfigMap for http datasources

dsanatar:
You can define cluster wide certs by modifying the `importProxy.trustedCAProxy` in the CDIConfig. These `certConfigMap` will then be copied to future importer pods so there is no need to define the `datavolume.spec.source.http.certConfigMap` for every DV that needs this cert. 

However, because these cluster wide certs are considered proxyCerts, the current http data source logic would not copy them unless the DV has a defined `certConfigMap` in it's spec. 

This PR will allow proxyCerts (and in turn CA's defined at the global level) to get copied and used for http datasources even when the standard cert directory is not defined.

Fixes https://redhat.atlassian.net/browse/CNV-47012

**Release note**:

```release-note
fix cluster wide certConfigMap for http datasources
```

### DIFF
--- a/tests/import_proxy_test.go
+++ b/tests/import_proxy_test.go
@@ -325,6 +325,48 @@ var _ = Describe("Import Proxy tests", func() {
 				expected:      BeFalse}),
 		)
 
+		It("should import from https endpoint using global trustedCAProxy without per-DV certConfigMap", func() {
+			By("Copying file host CA cert to CDI namespace for use as trustedCAProxy")
+			caConfigMapName, err := utils.CopyConfigMap(
+				f.K8sClient, f.CdiInstallNs, utils.FileHostCertConfigMap,
+				f.CdiInstallNs, "test-trusted-ca", "")
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Setting CDIConfig importProxy.trustedCAProxy")
+			updateCDIConfigProxy(f, "", "", "", caConfigMapName)
+
+			By("Waiting for CDI to reconcile the trustedCAProxy")
+			Eventually(func() string {
+				config, err := f.CdiClient.CdiV1beta1().CDIConfigs().Get(
+					context.TODO(), common.ConfigName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				ca, _ := cont.GetImportProxyConfig(config, common.ImportProxyConfigMapName)
+				return ca
+			}, time.Second*120, time.Second).Should(Equal(caConfigMapName))
+
+			imgURL := fmt.Sprintf("https://%s.%s:%d/%s",
+				fileHostName, f.CdiInstallNs, utils.HTTPSNoAuthPort, tinyCoreIso)
+			dvName = "test-global-trusted-ca"
+
+			By("Creating DataVolume without per-DV certConfigMap")
+			dv := utils.NewDataVolumeWithHTTPImport(dvName, "400Mi", imgURL)
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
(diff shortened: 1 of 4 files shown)


akalenyu (commented):
are we looking to only tackle the proxy path?
what about an internal registry with a custom CA, no proxy, and additional CA set?
the cm exists but CDI has no way to find it because it only looks through the Proxy CR?
(or am I missing a reason that this scenario is purely theoretical)

BTW, on OpenShift, we could simplify a lot of our plumbing with `inject-trusted-cabundle` label/ann
(upstream equivalent cert-manager `Bundle`)

btw, cc @germanovm, I dug out an old thread where you raised the usability issues around this, and the OCP CA bundle integration came up a lot

tests/import_proxy_test.go
```diff
@@ -325,6 +325,47 @@ var _ = Describe("Import Proxy tests", func() {
 				expected:      BeFalse}),
 		)
 
+		It("should import from https endpoint using global trustedCAProxy without per-DV certConfigMap", func() {
+			By("Copying file host CA cert to CDI namespace for use as trustedCAProxy")
+			caConfigMapName, err := utils.CopyConfigMap(
```
akalenyu, line 330: pretty sure i've seen this automated by cdi controller for any dv

dsanatar: so I believe the cdi controller handles copying certs from the `trustedCAProxy` configmap to the ns where the DVs importer pod is created however, to set the trustedCA field in the cdiconfig, the configmap must first exist in the cdi ns

dsanatar:
> are we looking to only tackle the proxy path? what about an internal registry with a custom CA, no proxy, and additional CA set? the cm exists but CDI has no way to find it because it only looks through the Proxy CR? (or am I missing a reason that this scenario is purely theoretical)
> 

the cdiconfig field that handles and stores global CA's is misleadingly named as `trustedCAProxy` (and is likely a candidate to be renamed/refactored in a future effort) but stores both standard and proxy certs. certs in this ConfigMap are copied into the `/proxycerts/` directory when a new importer pod is created, which is why we need the above changes for http data sources to still read certs from the proxy dir even if  `/cert/` dir is empty (populated from DV defined certConfigMap field)

Taking a quick look at the registry-datasource logic, it seems the cert collection already handles this case by collecting certs from `/proxycerts/` first.

tests/import_proxy_test.go
```diff
@@ -325,6 +325,47 @@ var _ = Describe("Import Proxy tests", func() {
...
+				return ca
+			}, time.Second*120, time.Second).Should(Equal(caConfigMapName))
+
+			imgURL := fmt.Sprintf("https://%s.%s:%d/%s",
+				fileHostName, f.CdiInstallNs, utils.HTTPSNoAuthPort, tinyCoreIso)
+			dvName = "test-global-trusted-ca"
+
+			By("Creating DataVolume without per-DV certConfigMap")
+			dv := utils.NewDataVolumeWithHTTPImport(dvName, "400Mi", imgURL)
+			dataVolume, err := utils.CreateDataVolumeFromDefinition(f.CdiClient, f.Namespace.Name, dv)
```
akalenyu, line 353: I think you want to explicitly assert that certConfigMap isn't set on this DV

akalenyu (commented):
/approve

if you could just create an issue to track revisiting 
- proxy being source of truth (downstream could just get cluster wide bundle injected)
- cert copying boilerplate vs certmanager (upstream) API integration

so weird that the source of truth for these additional CAs is the proxy object (proxy.spec.trustedCA), regardless of whether a proxy is used. maybe even an actual issue with install time CAs

dsanatar:
@akalenyu see https://github.com/kubevirt/containerized-data-importer/issues/4152

awels (approved):

PR #4110, opened 2026-05-04
Bump google.golang.org/grpc  to 1.79.3

sbiradar10:

Bump google.golang.org/grpc to 1.79.3

**Release note**:

```release-note
Bump google.golang.org/grpc to 1.79.3 to remediate CVE CVE-2026-33186
```


Acedus (approved):

go.mod
```diff
@@ -137,28 +137,30 @@ require (
...
+	golang.org/x/tools/go/packages/packagestest v0.1.1-deprecated // indirect
 	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
 	google.golang.org/genproto v0.0.0-20240227224415-6ceb2ff114de // indirect
-	google.golang.org/genproto/googleapis/api v0.0.0-20240528184218-531527333157 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240701130421-f6361c86f094 // indirect
-	google.golang.org/grpc v1.65.0 // indirect
-	google.golang.org/protobuf v1.34.2 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20251202230838-ff82c1b0f217 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
+	google.golang.org/grpc v1.79.3 // indirect
```
akalenyu, line 162: don't we need to use a replace directive? otherwise won't future go gets revert this?
cc @Acedus @awels @sbiradar10

Acedus: I don't think so? Pretty sure that just works.

akalenyu: my concern is that it gets overridden in case we update other packages

PR #4116, opened 2026-05-06
Add http endpoint block list

awels:
- Added a block list of known CIDRs that should not have any http end points on them, like internal ip addresses.
- Resolve URLs to IP addresses before checking.
- Added an allowList to override the block list if the block list was too aggressive.
- Added unit tests for both block and allow lists
- 
Fixes #

**Release note**:

```release-note
Prevent user from attempting to connect to internal IP addresses using a datavolume
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3941,6 +3941,14 @@
     "description": "CDIConfigSpec defines specification for user configuration",
     "type": "object",
     "properties": {
+     "allowedSourceURLs": {
+      "description": "AllowedSourceURLs is a list of CIDR blocks or hostnames that are allowed for HTTP/S data sources even if they fall within the default SSRF protection blocklist (private IPs, link-local, etc.). This allows importing from in-cluster services or private endpoints. Format: CIDR (10.96.0.0/12) or hostname (minio.default.svc)",
+      "type": "array",
+      "items": {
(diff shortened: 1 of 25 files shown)


Acedus:
Not sure how I feel about this. I understand adding timeouts but why don't we defer network policy tightening to network policies?

Acedus:
/hold
@awels mind answering my comment?

awels:
> Not sure how I feel about this. I understand adding timeouts but why don't we defer network policy tightening to network policies?

Probably, but we cannot assume everyone is using the appropriate network policies. These are mostly kubernetes internal addresses that we should never attempt to connect to.

awels:
adding a hold, identified that we don't actually have a test for `pullMode: Node` and we use localhost http endpoint in that scenario and that will likely fail right now due to the new block list
/hold

awels:
I need to rebase, and solve the issue where we are blocking 127.0.0.1 in combination with node pulls which connect to a local http server on that address.

akalenyu:
> > Not sure how I feel about this. I understand adding timeouts but why don't we defer network policy tightening to network policies?
> 
> Probably, but we cannot assume everyone is using the appropriate network policies. These are mostly kubernetes internal addresses that we should never attempt to connect to.

Why don't we ship this as part of our policy bundle?

PR #4124, opened 2026-05-12
Add open-default-ports annotation to upload server pod

RamLavi:
`virtctl image-upload` fails with a timeout when the target namespace uses OVN-Kubernetes network segmentation. The upload proxy (on the default network) cannot reach the upload server pod.

Since the upload service became headless (#4052), DNS resolves directly to the upload server pod IP. In OVN-Kubernetes clusters where the target namespace uses network segmentation, the pod's default network interface is locked down. The k8s.ovn.org/open-default-ports annotation opens port 8443 on the default interface, allowing the upload proxy to connect.

The annotation is harmless on non-OVN clusters where it is simply ignored.

Related: CNV-58018

The `k8s.ovn.org/open-default-ports` annotation is ignored by non-OVN CNI plugins — it only takes effect in OVN-Kubernetes managed clusters. See [OVN-K UDN isolation docs](https://ovn-kubernetes.io/features/user-defined-networks/user-defined-networks/) for details.

**Release note**:
```release-note
Add OVN-Kubernetes open-default-ports annotation to upload server pod to fix uploads in namespaces with network segmentation
```

### DIFF
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -596,8 +596,8 @@ func (r *UploadReconciler) makeUploadServiceSpec(name string, pvc *corev1.Persis
 			Ports: []corev1.ServicePort{
 				{
 					Protocol:   corev1.ProtocolTCP,
-					Port:       8443,
-					TargetPort: intstr.FromInt32(8443),
+					Port:       common.UploadServerPort,
+					TargetPort: intstr.FromInt32(common.UploadServerPort),
 				},
 			},
 			Selector: map[string]string{
@@ -763,7 +763,7 @@ func UploadPossibleForPVC(pvc *corev1.PersistentVolumeClaim) error {
 // GetUploadServerURL returns the url the proxy should post to for a particular pvc
 func GetUploadServerURL(namespace, pvc, uploadPath string) string {
 	serviceName := createUploadServiceNameFromPvcName(pvc)
-	return fmt.Sprintf("https://%s.%s.svc:8443%s", serviceName, namespace, uploadPath)
+	return fmt.Sprintf("https://%s.%s.svc:%d%s", serviceName, namespace, common.UploadServerPort, uploadPath)
 }
 
 // createUploadServiceName returns the name given to upload service shortened if needed
@@ -862,11 +862,8 @@ func (r *UploadReconciler) makeUploadPodContainers(args UploadPodArgs, resourceR
 			ReadinessProbe: &corev1.Probe{
 				ProbeHandler: corev1.ProbeHandler{
 					HTTPGet: &corev1.HTTPGetAction{
-						Path: "/healthz",
-						Port: intstr.IntOrString{
-							Type:   intstr.Int,
-							IntVal: 8443,
-						},
+						Path:   "/healthz",
+						Port:   intstr.FromInt32(common.UploadServerPort),
 						Scheme: corev1.URISchemeHTTPS,
 					},
 				},
--- a/pkg/controller/upload-controller.go
+++ b/pkg/controller/upload-controller.go
@@ -777,7 +777,8 @@ func (r *UploadReconciler) makeUploadPodSpec(args UploadPodArgs, resourceRequire
 			Name:      args.Name,
 			Namespace: args.PVC.Namespace,
 			Annotations: map[string]string{
-				annCreatedByUpload: "yes",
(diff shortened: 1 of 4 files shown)


pkg/controller/upload-controller_test.go
```diff
@@ -386,6 +386,7 @@ var _ = Describe("reconcilePVC loop", func() {
 			Expect(uploadPod.GetAnnotations()[cc.AnnPodNetwork]).To(Equal("net1"))
 			Expect(uploadPod.GetAnnotations()[cc.AnnPodSidecarInjectionIstio]).To(Equal(cc.AnnPodSidecarInjectionIstioDefault))
 			Expect(uploadPod.GetAnnotations()[cc.AnnPodSidecarInjectionLinkerd]).To(Equal(cc.AnnPodSidecarInjectionLinkerdDefault))
+			Expect(uploadPod.GetAnnotations()[cc.AnnOpenDefaultPorts]).To(Equal(`[{"protocol":"tcp","port":8443}]`))
```
akalenyu: should we also assert that the server uses 8443? we don't have much testing for udn here so easy to regress if we ever change port

Acedus: I think we should go one step further and add the upload ~~proxy~~ server port as a common const so that if it changes it won't break anything.

akalenyu: (did you mean server?) sounds good

RamLavi: +1, will add a preliminary commit

RamLavi: DONE

RamLavi:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/cfdca5541d7c2bf8f833025728f8ee48e0b2fe4c..46ba61109e45f76a066f05f2097e76c6f3eff348): Address reviews

Acedus (approved):

pkg/controller/upload-controller.go
```diff
@@ -862,11 +863,8 @@ func (r *UploadReconciler) makeUploadPodContainers(args UploadPodArgs, resourceR
 			ReadinessProbe: &corev1.Probe{
 				ProbeHandler: corev1.ProbeHandler{
 					HTTPGet: &corev1.HTTPGetAction{
```
akalenyu, line 865: apparently we don't set container ports in the upload server pod, but not that big of a deal i guess

pkg/controller/upload-controller_test.go
```diff
@@ -386,6 +387,7 @@ var _ = Describe("reconcilePVC loop", func() {
 			Expect(uploadPod.GetAnnotations()[cc.AnnPodNetwork]).To(Equal("net1"))
 			Expect(uploadPod.GetAnnotations()[cc.AnnPodSidecarInjectionIstio]).To(Equal(cc.AnnPodSidecarInjectionIstioDefault))
 			Expect(uploadPod.GetAnnotations()[cc.AnnPodSidecarInjectionLinkerd]).To(Equal(cc.AnnPodSidecarInjectionLinkerdDefault))
+			Expect(uploadPod.GetAnnotations()[cc.AnnOpenDefaultPorts]).To(Equal(fmt.Sprintf(`[{"protocol":"tcp","port":%d}]`, common.UploadServerPort)))
```
akalenyu, line 390: still missing an assertion between the used port and the open port ann value

akalenyu (commented):
just want to make sure we protect against regression and we're good to go

pkg/controller/upload-controller.go
RamLavi: if you think it's worth it I can add it, you call

pkg/controller/upload-controller_test.go
RamLavi: DONE

RamLavi:
[Change](https://github.com/kubevirt/containerized-data-importer/compare/46ba61109e45f76a066f05f2097e76c6f3eff348..1bd7c1e32d05c1045ab3e2509e694c5d8729488b): Added assertion between the used port and the open port ann value

RamLavi:
/cherry-pick release-v1.64

FYI @akalenyu

maiqueb (approved):
tks.

I just wonder if it would make sense to apply these *only* for OpenShift. It's not important, since it's harmless even if present.

PR #4127, opened 2026-05-12
VDDK: use instance UUIDs instead of BIOS UUIDs

mrnold:
This pull request changes the VDDK importer to look up VMs from VMware using the VM's instance UUID field. BIOS UUIDs can be duplicated, resulting in disk transfers from the wrong VM, so forklift is going to start sending instance UUIDs instead. This pull request leaves a fallback to look up the BIOS UUID if the instance UUID can't be found, which should preserve existing behavior in most cases.

Fixes [MTV-5264](https://redhat.atlassian.net/browse/MTV-5264)/[CNV-87549](https://redhat.atlassian.net/browse/CNV-87549)

I have a small VDDK E2E test image change to give this some test coverage, since there's not any existing unit test infrastructure for this part of the code. I would like to put this in a separate pull request to avoid backporting E2E test image changes, which I think cause trouble with builds/CI. If this is not actually a consideration, I can put it back.

**Release note**:
```release-note
VDDK: look up VMs with instance UUIDs instead of BIOS UUIDs
```

### DIFF
--- a/pkg/importer/vddk-datasource_amd64.go
+++ b/pkg/importer/vddk-datasource_amd64.go
@@ -535,7 +535,8 @@ func (vmware *VMwareClient) FindSnapshotDisk(snapshotRef *types.ManagedObjectRef
 	return nil, fmt.Errorf("Could not find disk image with ID %s in snapshot %s", diskID, snapshotRef.Value)
 }
 
-// FindVM takes the UUID of the VM to migrate and finds its MOref
+// FindVM takes the instance UUID of the VM to migrate and finds its MOref,
+// with a fallback to looking for the VM with that UUID as the BIOS UUID.
 func FindVM(context context.Context, conn *govmomi.Client, uuid string) (string, *object.VirtualMachine, error) {
 	// Get the list of datacenters to search for VM UUID
 	finder := find.NewFinder(conn.Client, true)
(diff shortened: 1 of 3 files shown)


mrnold:
This is going to need some backports, maybe it's worth minimizing the diff by putting the datacenter search loop back in? I am not sure how much trouble all the vendor churn will really cause.

mrnold:
> This is going to need some backports, maybe it's worth minimizing the diff by putting the datacenter search loop back in? I am not sure how much trouble all the vendor churn will really cause.

I ended up doing this, I think it will make life easier.

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -544,16 +545,39 @@ func FindVM(context context.Context, conn *govmomi.Client, uuid string) (string,
...
+	}
 
-	// Search for VM matching given UUID, and save the MOref
+	// Search for VM matching given instance UUID, and save the MOref
 	var moref string
 	var instanceUUID bool
 	var vm *object.VirtualMachine
 	searcher := object.NewSearchIndex(conn.Client)
 	for _, datacenter := range datacenters {
+		instanceUUID = true
```
awels: should this be inside the loop? Or does searcher.FindByUUid modify that value? I see it being passed by reference. I checked the code, and it does NOT appear to modify that field. So why not just pass in a pointer to a const `true` here and `false` below?

awels (commented):
Looks fine, just one nit

pkg/importer/vddk-datasource_amd64.go
mrnold: It doesn't have to be in the loop, FindByUuid does not modify it. I left it in purely for diff minimization, because adding ptr here also adds a dependency to the bazel file. If that's not really a problem I can just use ptr.

mrnold: I have taken this out now.

mrnold:
Is there anything else needed here?

awels (approved):
/approve

PR #4123, opened 2026-05-11
Upstream IBM VPC File storage provider

GeraFD14:
**What this PR does / why we need it**:  Need to upstream IBM VPC File Storage Provider into CDI

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: 
Fixes #CNV-84991

**Special notes for your reviewer**: Adding `vpc.file.csi.ibm.io` provider including the dynamic SIZE/IOPS required relation based on https://cloud.ibm.com/docs/vpc?topic=vpc-file-storage-profiles&interface=ui#dp2-profile

**Release note**:

```release-note

```

PR #4138, opened 2026-05-15
Add  IBM VPC File storage capabilities

GeraFD14:
Replaces #4123

**What this PR does / why we need it**:  Need to upstream IBM VPC File Storage Provider into CDI

Fixes https://redhat.atlassian.net/browse/CNV-84991

**Release note**:

```release-note
Add support for IBM VPC File Storage driver (vpc.file.csi.ibm.io)
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -101,6 +101,12 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"block.csi.ibm.com": {{rwx, block}, {rwo, block}, {rwo, file}},
 	// IBM VPC Block CSI
 	"vpc.block.csi.ibm.io": {{rwo, block}, {rwo, file}},
+	// IBM VPC File CSI
+	"vpc.file.csi.ibm.io":         {{rwx, file}, {rwo, file}},
+	"vpc.file.csi.ibm.io/dp2":     {{rwx, file}, {rwo, file}},
+	"vpc.file.csi.ibm.io/dp2-rwo": {{rwo, file}},
+	"vpc.file.csi.ibm.io/rfs":     {{rwx, file}, {rwo, file}},
+	"vpc.file.csi.ibm.io/rfs-rwo": {{rwo, file}},
 	// Portworx in-tree CSI
 	"kubernetes.io/portworx-volume/nfs": {{rwx, file}, {rwo, file}},
 	"kubernetes.io/portworx-volume":     {{rwx, block}, {rwx, file}, {rwo, block}, {rwo, file}},
@@ -159,6 +165,11 @@ var SourceFormatsByProvisionerKey = map[string]cdiv1.DataImportCronSourceFormat{
 	"csi.trident.netapp.io/gcnv-flex":    cdiv1.DataImportCronSourceFormatSnapshot,
 	"pd.csi.storage.gke.io":              cdiv1.DataImportCronSourceFormatSnapshot,
 	"pd.csi.storage.gke.io/hyperdisk":    cdiv1.DataImportCronSourceFormatSnapshot,
+	"vpc.file.csi.ibm.io":                cdiv1.DataImportCronSourceFormatSnapshot,
+	"vpc.file.csi.ibm.io/dp2":            cdiv1.DataImportCronSourceFormatSnapshot,
+	"vpc.file.csi.ibm.io/dp2-rwo":        cdiv1.DataImportCronSourceFormatSnapshot,
+	"vpc.file.csi.ibm.io/rfs":            cdiv1.DataImportCronSourceFormatSnapshot,
+	"vpc.file.csi.ibm.io/rfs-rwo":        cdiv1.DataImportCronSourceFormatSnapshot,
 }
 
 // CloneStrategyByProvisionerKey defines the advised clone strategy for a provisioner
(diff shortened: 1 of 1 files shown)


GeraFD14:
@arnongilboa @awels 
thank you for your comments, unfortunately I made a mistake an the original PR was closed, will continue in this one.

--- 
@arnongilboa from your previous comments:

> 1. Please update about the rounding. Not sure we can approve such vendor-specific code.

Lets continue this discussion on slack.

> 2. What is the vTPM SC provisioner key? shouldn't it default to RWO (without RWX)?

I've added the new provisioner keys for vTPM SC, please help validating last commit and let me know if I need to change anything else

> 3. You also need to align the PR to the repo standards:
>    - Add a description to the commits
>    - Sign the commits
>    - Add a release note to the PR

I'll continue with this and will correct the PR ASAP

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -316,10 +321,51 @@ func GetAdvisedCloneStrategy(sc *storagev1.StorageClass) (cdiv1.CDICloneStrategy
...
+		return getIBMVPCFileDP2MinimumSize(sc)
+	}
 	size, found := MinimumSupportedPVCSizeByProvisionerKey[provisionerKey]
 	return size, found
 }
 
+// getIBMVPCFileDP2MinimumSize calculates the minimum PVC size for IBM VPC File Storage DP2 profile
+// based on the IOPS parameter using a table lookup
+// Reference: https://cloud.ibm.com/docs/vpc?topic=vpc-file-storage-profiles&interface=ui#dp2-profile
+func getIBMVPCFileDP2MinimumSize(sc *storagev1.StorageClass) (string, bool) {
```
arnongilboa: As mentioned before, not sure we want this very vendor-specific logic in CDI. It's also quite impossible to maintain over time. @awels @akalenyu?

noamasu: @gera-fuentes does requesting a pvc size below the minimum result in an error or an auto roundup (to min size)?

GeraFD14: @noamasu requesting a pvc size below the minimum results in an error right now

akalenyu: why don't we have some operator reconcile the storage class with a parameter/annotation that advertises the min size

arnongilboa: @akalenyu they can add an optional "roundup" param in the storage class, so if you set "roundup: true", it will do the roundup logic in the CSI driver. wdyt?

GeraFD14: @arnongilboa @noamasu @akalenyu I have removed the IBM's specific logic for the minimum size annotation; we are going to update the provider to do the size round up instead.

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -539,6 +560,30 @@ var storageClassToProvisionerKeyMapper = map[string]func(sc *storagev1.StorageCl
 			return "UNKNOWN"
 		}
 	},
+	"vpc.file.csi.ibm.io": func(sc *storagev1.StorageClass) string {
+		// https://cloud.ibm.com/docs/vpc?topic=vpc-file-storage-profiles&interface=ui
+		if sc.Parameters == nil {
+			return "vpc.file.csi.ibm.io"
+		}
+		profile := sc.Parameters["profile"]
+		// When vmState is true, use RWO-specific variants
```
akalenyu, line 569: Could you elaborate on the vmstate situation a little? What is wrong with using RWX where the vm state will be truly "live migrated"?

GeraFD14: Our provider has permission issues on the vTPM PVC, it cannot be mounted to the VM's pod so we had to set both `csi.storage.k8s.io/fstype: ext4` and RWO in our SC to make it work

akalenyu: Hmm that sounds familiar and usually points to a bigger problem in that the CSIDriver is not using `fsGroupPolicy: File`. This will bite elsewhere in the kubevirt ecosystem, iirc.

GeraFD14: I'll discuss with the CSIDriver team about the `fsGroupPolicy` parameter. And just wondering, Is this a blocker to merge this code?

akalenyu: Nope, just note general kubevirt GA with this driver may be tricky. Export could break, for example

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -210,6 +226,11 @@ var MinimumSupportedPVCSizeByProvisionerKey = map[string]string{
 	"csi.san.synology.com/smb":   "1Gi",
 	// https://cloud.google.com/netapp/volumes/docs/discover/service-levels
 	"csi.trident.netapp.io/gcnv-flex": "1Gi",
+	// IBM VPC File Storage
+	"vpc.file.csi.ibm.io/rfs":     "1Gi",
+	"vpc.file.csi.ibm.io/rfs-rwo": "1Gi",
+	"vpc.file.csi.ibm.io/dp2":     "10Gi",
+	"vpc.file.csi.ibm.io/dp2-rwo": "10Gi",
```
akalenyu, line 233: This going to be temporary if I understand https://github.com/kubevirt/containerized-data-importer/pull/4138#discussion_r3554175611 correctly, if that's the case, maybe worth a TODO?

akalenyu (commented):
/approve

pkg/storagecapabilities/storagecapabilities.go
GeraFD14: those are our default minimum sizes requirements from our backend, the roundoff logic that will be implemented in our CSIDriver helped removing the custom function to annotate the SP according to the size. so no TODO to remove this for now

akalenyu: oh I thought the driver was going to round up all across the board

GeraFD14: the driver will do the round up for specific SC that has a new parameter, existing SC will keep old functionality

arnongilboa (approved):

PR #4141, opened 2026-05-17
Makefile, hack: Allow compiling ginkgo + tests without Bazel

Acedus:
$title

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build/common.sh
+++ b/hack/build/common.sh
@@ -72,6 +72,10 @@ function format_archname() {
     fi
 }
 
+function go_build() {
+    GOPROXY=${GOPROXY:-""} go build "$@"
+}
+
 CDI_DIR="$(cd $(dirname $0)/../../ && pwd -P)"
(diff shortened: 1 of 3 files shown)


hack/build/common.sh
```diff
@@ -72,6 +72,10 @@ function format_archname() {
     fi
 }
 
+function go_build() {
+    GOPROXY=off go build "$@"
```
akalenyu: I would leave an escape hatch that takes GOPROXY from env similarly to
`export GO111MODULE=${GO111MODULE:-off}`

akalenyu: actually my suggestion was probably incorrect, we want this to be on by default probably, but anyways escape hatch exists

akalenyu (commented):
/lgtm
/approve

akalenyu:
empty pv name race

PR #4162, opened 2026-05-28
fix context leaks in GCS datasource and registry transport

immanuwell:
What this PR does / why we need it:

Three context bugs in pkg/importer that have been quietly leaking resources:

1. NewGCSDataSource discards the cancel func from context.WithTimeout - there's even a nolint:govet todo comment acknowledging it. The context is never cancelled so goroutines/timers leak on every GCS import. Fix: store cancel in the struct, call it in Close() (and on error paths).

2. GetImageDigest creates a timeout ctx via commandTimeoutContext() but then passes context.Background() to src.GetManifest() - so the timeout has no effect on the actual manifest fetch, it can hang forever.

3. validateImagePlatformMatch hardcodes context.Background() even though the caller has a proper timeout ctx. Fixed by threading ctx through as a param.

How to reproduce 1 (GCS leak):
- profile memory while running repeated GCS imports
- or just look at pkg/importer/gcs-datasource.go:61, the nolint comment spells it out

How to reproduce 2 (GetManifest ignores timeout):
- call GetImageDigest against a slow/unresponsive registry with a short timeout - it won't respect the timeout

Release note:
```release-note
NONE
```

### DIFF
--- a/pkg/importer/gcs-datasource.go
+++ b/pkg/importer/gcs-datasource.go
@@ -40,6 +40,8 @@ type GCSDataSource struct {
 	readers *FormatReaders
 	// The image file in scratch space.
 	url *url.URL
+	// cancel cancels the context used for the GCS client and reader
+	cancel context.CancelFunc
 }
 
 // NewGCSDataSource creates a new instance of the GCSDataSource
@@ -58,7 +60,7 @@ func NewGCSDataSource(endpoint, keyFile string) (*GCSDataSource, error) {
 	}
 
 	// Getting Context
-	ctx, _ := context.WithTimeout(context.Background(), time.Second*60) //nolint:govet // todo - solve this: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 
 	if ep.Scheme == "gs" {
 		// Using gs:// endpoint and extracting bucket and object name
@@ -71,8 +73,8 @@ func NewGCSDataSource(endpoint, keyFile string) (*GCSDataSource, error) {
 
 	// Creating GCS Client
 	client, err := getGcsClient(ctx, keyFile, options...)
-
 	if err != nil {
+		cancel()
 		klog.Errorf("GCS Importer: Error creating GCS Client")
 		return nil, err
 	}
@@ -80,6 +82,7 @@ func NewGCSDataSource(endpoint, keyFile string) (*GCSDataSource, error) {
 	// Creating GCS Reader
 	gcsReader, err := newReaderFunc(ctx, client, bucket, object)
 	if err != nil {
+		cancel()
 		klog.Errorf("GCS Importer: Error creating Reader")
 		return nil, err
 	}
@@ -88,6 +91,7 @@ func NewGCSDataSource(endpoint, keyFile string) (*GCSDataSource, error) {
 		ep:        ep,
 		keyFile:   keyFile,
 		gcsReader: gcsReader,
+		cancel:    cancel,
 	}, nil
 }
 
@@ -160,6 +164,9 @@ func (sd *GCSDataSource) GetTerminationMessage() *common.TerminationMessage {
 // Close closes any readers or other open resources.
 func (sd *GCSDataSource) Close() error {
 	var err error
+	if sd.cancel != nil {
+		sd.cancel()
+	}
 	if sd.readers != nil {
 		err = sd.readers.Close()
 	}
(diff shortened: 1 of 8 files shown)


pkg/importer/transport.go
```diff
@@ -287,7 +287,7 @@ func GetImageDigest(url, accessKey, secKey, certDir string, insecureRegistry boo
 	}
 	defer closeImage(src)
 
-	imageManifest, _, err := src.GetManifest(context.Background(), nil)
+	imageManifest, _, err := src.GetManifest(ctx, nil)
```
akalenyu, line 288: this is a correct fix but I don't think there's a real context behind commandTimeoutContext so in practice no deadline

pkg/importer/gcs-datasource.go
```diff
@@ -163,6 +167,9 @@ func (sd *GCSDataSource) Close() error {
 	if sd.readers != nil {
 		err = sd.readers.Close()
 	}
+	if sd.cancel != nil {
+		sd.cancel()
+	}
```
akalenyu, line 169: should we reorder this (cancel first, then close)?

akalenyu (commented):
thank you!

akalenyu:
unrelated flakes

pkg/importer/transport.go
immanuwell: agreed, I updated commandTimeoutContext() to use a real timeout-backed context, so GetManifest now runs with an actual deadline instead of a plain cancel-only context

pkg/importer/gcs-datasource.go
immanuwell: done, Close() now cancels the context before closing the readers, and I added a small test to cover that order

immanuwell:
@akalenyu done, I addressed all your suggestions, PTAL

pkg/importer/transport.go
```diff
@@ -39,13 +40,14 @@ import (
 )
 
 const (
-	whFilePrefix = ".wh."
+	whFilePrefix           = ".wh."
+	registryRequestTimeout = 60 * time.Second
```
akalenyu: 60s for downloading and processing all layers of a container image could be too aggressive

akalenyu: maybe just take a context in importer main app with `signal.NotifyContext` and pass that in struct?

pkg/importer/gcs-datasource_test.go
```diff
@@ -473,8 +473,37 @@ var _ = Describe("Google Cloud Storage data source", func() {
 		Expect(err).To(HaveOccurred())
 		Expect(ProcessingPhaseError).To(Equal(result))
 	})
+
+	It("Close should cancel the context before closing readers", func() {
```
akalenyu, line 477: nice!

pkg/importer/transport.go
immanuwell: done, I removed the hardcoded 60s registry timeout and instead thread a signal-backed root context from the entrypoints down into the registry import/digest paths

immanuwell:
@akalenyu done, PTAL

pkg/importer/transport.go
```diff
@@ -195,11 +191,12 @@ func safeJoinPaths(dir, path string) (v string, err error) {
...
 
-func copyRegistryImage(url, destDir, pathPrefix, accessKey, secKey, imageArchitecture, certDir string, insecureRegistry, stopAtFirst, preallocation bool) (*types.ImageInspectInfo, error) {
+func copyRegistryImage(ctx context.Context, url, destDir, pathPrefix, accessKey, secKey, imageArchitecture, certDir string, insecureRegistry, stopAtFirst, preallocation bool) (*types.ImageInspectInfo, error) {
 	klog.Infof("Downloading image from '%v', copying file from '%v' to '%v'", url, pathPrefix, destDir)
 
-	ctx, cancel := commandTimeoutContext()
-	defer cancel()
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
akalenyu, line 199: don't think we need this?

pkg/importer/transport.go
```diff
@@ -274,11 +271,12 @@ func validateImagePlatformMatch(sys *types.SystemContext, img types.Image) error
...
 // insecureRegistry: boolean if true will allow insecure registries.
-func GetImageDigest(url, accessKey, secKey, certDir string, insecureRegistry bool) (string, error) {
+func GetImageDigest(ctx context.Context, url, accessKey, secKey, certDir string, insecureRegistry bool) (string, error) {
 	klog.Infof("Inspecting image from '%v'", url)
 
-	ctx, cancel := commandTimeoutContext()
-	defer cancel()
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
akalenyu, line 279: same, we own all the calls and can guarantee context is passed

akalenyu (commented):
looks good, just unsure about the extra defensiveness with ctx == nil
/test all

akalenyu:
/retest
ci flake https://github.com/kubevirt/containerized-data-importer/pull/4199

PR #4163, opened 2026-05-28
tests: replace racy cdi_cr_ready e2e test with unit test

akalenyu:
The e2e test that removed and recreated CDI to verify the
kubevirt_cdi_cr_ready metric was racy due to token secret
population timing after CDI reinstall. Replace it with a
deterministic unit test that exercises the metric through
the reconciler lifecycle.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/monitoring/metrics/operator-controller/operator_metrics.go
+++ b/pkg/monitoring/metrics/operator-controller/operator_metrics.go
@@ -1,6 +1,7 @@
 package operatorcontroller
 
 import (
+	ioprometheusclient "github.com/prometheus/client_model/go"
 	"github.com/rhobs/operator-observability-toolkit/pkg/operatormetrics"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/monitoring/metrics/operator-controller/operator_metrics.go
```diff
@@ -31,3 +32,10 @@ func SetNotReady() {
 func SetInit() {
 	readyGauge.Set(-1.0)
 }
+
+// GetReadyValue returns the current value of the readyGauge
+func GetReadyValue() float64 {
+	dto := &ioprometheusclient.Metric{}
+	_ = readyGauge.Write(dto)
```
dsanatar: I recognize for this simple ut this would never error but if this function was to be re-used elsewhere in the future, maybe we would want to be able to consume the possible error?

akalenyu: good point, we seem to be missing a standard on this (some metrics check for this err value, some don't)

akalenyu: done

awels (approved):
/approve

PR #4167, opened 2026-06-09
add insecureSkipVerify to http source type

### LINKED ISSUE #4166: http source data importer can not import from a https source with missing/invalid SAN
**What happened**:
When importing from an HTTPS service that has an invalid or malformed certificate, it is not possible to override the security check with an "insecureSkipVerify" type option.  This can occur when you have a self signed certificate, but the subject name is incorrect, or the SAN names are incorrect. Defining/configuring a `certConfigMap` does not resolve this type of error.

It is not always possible to fix the certificate on the endpoint, so it should be up to the end user to allow the importer to bypass the TLS check. The user should be able to override the check with an "insecureSkipVerify" option on the source, allowing the creation of the volume to complete.

Create a datavolume definition that uses an HTTPS datasource. The certificate on the source should be self signed, but have invalid information for the SAN. You will be unable to import the datasource even if you define a certConfigMap. 

**Additional context**:
This was discussed in the slack thread https://kubernetes.slack.com/archives/C8ED7RKFE/p1780500582897189 . I will be submitting a PR to resolve this issue shortly. 

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): ALL

xphyr:
This PR adds an `insecureSkipVerify` field to `http` DavaVolume imports. This helps to support connections to https datasources that have corrupt, or incomplete TLS certificates but the user still wants to connect anyway. For example, connections to sources that have a self signed certificate, but no SAN defined for the IP address, will not import, even if you have defined a `certConfigMap`. This option allows you to bypass this check on a specific DataVolume. This also simplifies the process for downloading from a source using a self-signed certificate, IF you are OK with completely bypassing the check. 

This leverages the existing code/process for skipping TLS verification that is used by the imageio importer. 

I have updated the documentation, as well as the go unit tests, and functional test. 

**Which issue(s) this PR fixes** *:
Fixes #4166 

This work was discussed on Slack in the following thread: https://kubernetes.slack.com/archives/C8ED7RKFE/p1780500582897189

**Release note**:

```release-note
Added insecureSkipVerify to http source type to support connecting to https endpoints with invalid certificates. 
By default this option is set to false, and must be explicitly set to "true" to use. 
This will not break any existing datavolume definitions.
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -1753,6 +1753,9 @@ func UpdateHTTPAnnotations(annotations map[string]string, http *cdiv1.DataVolume
 	if http.Checksum != "" {
 		annotations[AnnChecksum] = http.Checksum
 	}
+	if http.InsecureSkipVerify != nil && *http.InsecureSkipVerify {
+		annotations[AnnInsecureSkipVerify] = "true"
+	}
 	for index, header := range http.ExtraHeaders {
 		annotations[fmt.Sprintf("%s.%d", AnnExtraHeaders, index)] = header
 	}
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -713,9 +713,9 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 }
 
 func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim, cdiConfig *cdiv1.CDIConfig) (bool, error) {
-	// Check if insecureSkipVerify annotation is set (only applicable for ImageIO sources)
+	// Check if insecureSkipVerify annotation is set (only applicable for ImageIO sources and HTTP sources)
 	source, sourceOk := pvc.Annotations[cc.AnnSource]
-	if sourceOk && source == cc.SourceImageio {
+	if sourceOk && (source == cc.SourceImageio || source == cc.SourceHTTP) {
 		if insecureSkipVerify, ok := pvc.Annotations[cc.AnnInsecureSkipVerify]; ok && insecureSkipVerify == "true" {
 			return true, nil
 		}
--- a/pkg/importer/http-datasource.go
+++ b/pkg/importer/http-datasource.go
@@ -84,14 +84,13 @@ type HTTPDataSource struct {
 	contentLength uint64
 	// checksumValidator validates the checksum of downloaded data
 	checksumValidator *ChecksumValidator
-
-	n image.NbdkitOperation
+	n                 image.NbdkitOperation
 }
 
 var createNbdkitCurl = image.NewNbdkitCurl
 
 // NewHTTPDataSource creates a new instance of the http data provider.
-func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType cdiv1.DataVolumeContentType, checksum string) (*HTTPDataSource, error) {
+func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType cdiv1.DataVolumeContentType, checksum string, insecureSkipVerify bool) (*HTTPDataSource, error) {
 	ep, err := ParseEndpoint(endpoint)
 	if err != nil {
 		return nil, errors.Wrapf(err, "unable to parse endpoint %q", endpoint)
@@ -104,7 +103,7 @@ func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType 
 		return nil, errors.Wrap(err, "Error getting extra headers for HTTP client")
 	}
 
-	httpReader, contentLength, brokenForQemuImg, err := createHTTPReader(ctx, ep, accessKey, secKey, certDir, extraHeaders, secretExtraHeaders, contentType)
+	httpReader, contentLength, brokenForQemuImg, err := createHTTPReader(ctx, ep, accessKey, secKey, certDir, extraHeaders, secretExtraHeaders, contentType, insecureSkipVerify)
 	if err != nil {
 		cancel()
 		return nil, err
@@ -365,9 +364,9 @@ func addExtraheaders(req *http.Request, extraHeaders []string) {
 	req.Header.Add("User-Agent", defaultUserAgent)
 }
 
-func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certDir string, extraHeaders, secretExtraHeaders []string, contentType cdiv1.DataVolumeContentType) (io.ReadCloser, uint64, bool, error) {
+func createHTTPReader(ctx context.Context, ep *url.URL, accessKey, secKey, certDir string, extraHeaders, secretExtraHeaders []string, contentType cdiv1.DataVolumeContentType, insecureSkipVerify bool) (io.ReadCloser, uint64, bool, error) {
 	var brokenForQemuImg bool
-	client, err := createHTTPClient(certDir, false)
+	client, err := createHTTPClient(certDir, insecureSkipVerify)
 	if err != nil {
 		return nil, uint64(0), false, errors.Wrap(err, "Error creating http client")
 	}
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -4794,6 +4794,10 @@
        "default": ""
       }
      },
+     "insecureSkipVerify": {
+      "description": "InsecureSkipVerify is a flag to skip certificate verification for the HTTP endpoint",
+      "type": "boolean"
+     },
      "secretExtraHeaders": {
       "description": "SecretExtraHeaders is a list of Secret references, each containing an extra HTTP header that may include sensitive information",
       "type": "array",
--- a/cmd/cdi-importer/importer.go
+++ b/cmd/cdi-importer/importer.go
@@ -291,7 +291,7 @@ func newDataSource(source string, contentType string, volumeMode v1.PersistentVo
 
 	switch source {
 	case cc.SourceHTTP:
-		ds, err := importer.NewHTTPDataSource(getHTTPEp(ep), acc, sec, certDir, cdiv1.DataVolumeContentType(contentType), checksum)
+		ds, err := importer.NewHTTPDataSource(getHTTPEp(ep), acc, sec, certDir, cdiv1.DataVolumeContentType(contentType), checksum, insecureTLS)
 		if err != nil {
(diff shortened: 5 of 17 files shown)


pkg/controller/common/util.go
```diff
@@ -1753,6 +1753,9 @@ func UpdateHTTPAnnotations(annotations map[string]string, http *cdiv1.DataVolume
 	if http.Checksum != "" {
 		annotations[AnnChecksum] = http.Checksum
 	}
+	if http.InsecureSkipVerify != nil && *http.InsecureSkipVerify {
+		annotations[AnnInsecureSkipVerify] = "true"
+	}
```
akalenyu, line 1758: could we get some unit tests that validate `AnnInsecureSkipVerify` is correctly set when `InsecureSkipVerify` is true, or correctly absent when nil/false?

pkg/controller/import-controller.go
```diff
@@ -713,9 +713,9 @@ func (r *ImportReconciler) createImportEnvVar(pvc *corev1.PersistentVolumeClaim)
 }
 
 func (r *ImportReconciler) isInsecureTLS(pvc *corev1.PersistentVolumeClaim, cdiConfig *cdiv1.CDIConfig) (bool, error) {
-	// Check if insecureSkipVerify annotation is set (only applicable for ImageIO sources)
+	// Check if insecureSkipVerify annotation is set (only applicable for ImageIO sources and HTTP sources)
 	source, sourceOk := pvc.Annotations[cc.AnnSource]
-	if sourceOk && source == cc.SourceImageio {
+	if sourceOk && (source == cc.SourceImageio || source == cc.SourceHTTP) {
 		if insecureSkipVerify, ok := pvc.Annotations[cc.AnnInsecureSkipVerify]; ok && insecureSkipVerify == "true" {
```
akalenyu, line 719: Some unit tests for this would also be great, specifically that no AnnInsecureSkipVerify results in no env var for TLS bypass. Just trying to cover everything that guards the runtime importer app receiving insecure=true

akalenyu (commented):
Thank you!

Overall this makes sense, but I am slightly worried on the security side;
There is no admin level config to say "no one in this cluster may skip TLS on imports". 
This is also a problem for existing ImageIO but arguably those are more controlled environments to begin with. Maybe we could require checksum when using insecure imports?
/cc @Acedus 
Network policies may be reducing the blast radius here

EDIT:
I guess we could also put a insecureHTTPEndpoints next to existing CDI CR insecureRegistries and inherit it from the platform (OpenShift configuration CR, but from a quick search there isn't one).

Acedus:
> Thank you!
> 
> Overall this makes sense, but I am slightly worried on the security side; There is no admin level config to say "no one in this cluster may skip TLS on imports". This is also a problem for existing ImageIO but arguably those are more controlled environments to begin with. Maybe we could require checksum when using insecure imports? /cc @Acedus Network policies may be reducing the blast radius here
> 
> EDIT: I guess we could also put a insecureHTTPEndpoints next to existing CDI CR insecureRegistries and inherit it from the platform (OpenShift configuration CR, but from a quick search there isn't one).

I'd say it would be the admin's responsibility to prevent this using something like a VAP.

xphyr:
> I'd say it would be the admin's responsibility to prevent this using something like a VAP.

This was my thought as well. I feel like an admission policy would be the way to ensure that end user(s) cant use an insecureHTTP endpoint.  I feel like this is a less complicated and more universal way to handle this sort of control. Something like this would handle it base on some quick testing:

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: deny-datavolume-insecure-http
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups: ["cdi.kubevirt.io"]
      apiVersions: ["v1beta1"]
      operations: ["CREATE", "UPDATE"]
      resources: ["datavolumes"]
  validations:
    - expression: >
        !has(object.spec.source) || 
        !has(object.spec.source.http) || 
        !has(object.spec.source.http.insecureSkipVerify) || 
        object.spec.source.http.insecureSkipVerify == false
      message: "Setting insecureSkipVerify to true on an HTTP source is forbidden for security compliance."
---
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: deny-datavolume-insecure-http-binding
spec:
  policyName: deny-datavolume-insecure-http
  validationActions: [Deny]
  matchResources:
    namespaceSelector: {}
```

pkg/controller/common/util.go
xphyr: OK, I think I now have this covered here: https://github.com/kubevirt/containerized-data-importer/pull/4167/changes/8b34ebdb268a09643523ff2aceb04cdbea8bd0ac Please be sure to take a close look, writing unit tests is NOT a strength of mine, so I relied on the help of AI to build this.

pkg/controller/import-controller.go
xphyr: I have also written a test here: https://github.com/kubevirt/containerized-data-importer/pull/4167/changes/2aea8ad69706544fb03db7fafea59c5867aa2b81 with the help of AI. I think that does what you are suggesting.

akalenyu: This is good! I was hoping to assert on the resulting pod env vars though to cover more ground. We should have existing tests that do something along the lines of
```go
_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "testPvc1", Namespace: "default"}})
pod := &corev1.Pod{}
err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "importer-testPvc1", Namespace: "default"}, pod)
<ASSERTION>
```

pkg/controller/common/util.go
akalenyu: This is also a solid unit test, but I think we could take a similar approach to my suggestion https://github.com/kubevirt/containerized-data-importer/pull/4167#discussion_r3395148409 in `pkg/controller/populators/import-populator_test.go`

akalenyu:
Yeah VAP is good but we could follow the admin controlled flow similarly to insecureRegistries (`insecureHTTPEndpoints`?). What do folks think? I know we went with DV spec for ImageIO and I realize that I am effectively challenging that too.
/cc @Acedus @awels

xphyr:
@akalenyu Here are my thoughts on `insecureHTTPEndpoints` and why I think defining the override as part of DataSource definition:
- Primary concern is that this does not allow for dynamic endpoints. Every time someone wants/needs to pull from an insecure endpoint they will need to contact/work with a cluster admin to pull from an insecure site.  This will be the only way to do it. This puts extra burden on the end user and cluster admin in all scenarios. 
- The secondary concern is that I think that this adds extra complexity to the code base (not a lot) needing to check the config and validating the config. 
- I would also argue that the ImageIO code should be updated to follow the same pattern so that the process is the same across multiple providers
- The OpenStack importer also uses "insecureSkipVerify" as part of its process https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/openstack-populator/openstack-populator.go#L49 so to maintain consistency this would probably need to be updated as well
- Finally, it just feels cleaner ... if you want to understand a DataSource everything about it is defined in one place, with this implementation, if I want to know if insecureSkipVerify is required, I need only look at the datasource definition. To understand the same using the `insecureHTTPEndpoints` solution I need to look at the DataSource definition as well as the CDIConfig. As an end user, I wont have permissions to view the CDIConfig to see if my endpoint is even allowed.

Leveraging VAP is an existing solution that requires no extra code to write/test/maintain and is arguably more flexible allowing a Cluster Admin to:
- do nothing, and let the end user "do the right thing"
- disable it entirely with a VAP
- allow specific endpoints as part of the VAP definition if this level of control is required

pkg/controller/import-controller.go
xphyr: I am attempting to re-write the test as you are looking for. Before I move forward with the other unit test, can you review this to see if it is what you were looking for.  [fixed the issue with my first unit test](https://github.com/kubevirt/containerized-data-importer/pull/4167/commits/3d2fd5e5e949efb2a5b0c3a1ad0078d10b3aa46d) If this is the proper direction, I will review the unit test for `util.go`. Please note I have not yet removed the other test. I wanted to ensure that this was the way you wanted to go, before deleting that code. If this is good, I will delete the first test that I wrote.

pkg/controller/common/util.go
xphyr: For this unit test, I feel like we are testing at the proper level... reviewing the rest of the tests for this package none of them appear to be going through the process of checking the environment variables assigned to a pod. While I agree testing for this in the import-controller makes sense, since this is a helper package/function should we not just be testing the results of calling the helper function?  Perhaps I am misunderstanding how you would like me to test this.  Please advise and thanks for the assistance.

pkg/importer/http-datasource.go
```diff
@@ -84,14 +84,13 @@ type HTTPDataSource struct {
 	contentLength uint64
 	// checksumValidator validates the checksum of downloaded data
 	checksumValidator *ChecksumValidator
-
-	n image.NbdkitOperation
+	n                 image.NbdkitOperation
```
akalenyu, line 87: my editor corrects this back, how come you had to pad the spaces?

pkg/controller/import-controller.go
akalenyu: Yeah this was the direction. I was looking for something along those lines:
https://github.com/akalenyu/containerized-data-importer/commit/d6c22d082da74e29eab744ccc4f4b3c978e17deb

akalenyu (commented):
Thanks for sticking through! I went ahead and prompted Claude to help with the unit test direction, something along those lines should be good

pkg/importer/http-datasource.go
xphyr: Its funny, my editor puts all the spaces in. Even when I remove them, it puts them back.  (vscode). If you want I will edit the file outside vscode and commit to remove this change.

xphyr: I dug into this, its `go fmt` that is making this formatting change.  I am using go1.25 on my dev machine ...

pkg/controller/import-controller.go
xphyr: OK, I have integrated that into the most recent change. I think we are all set. Thank you for your continued assistance.

awels (approved):

akalenyu:
pv name flake

PR #4171, opened 2026-06-12
docs: add k3s/RKE2 guidance for device_ownership_from_security_context

wakbijok:

The block CRI ownership doc shows containerd config snippets, but on k3s and RKE2 the containerd config.toml is regenerated on every service start, so following the doc as written silently does nothing there. This adds a section covering:

- the k3s `nonroot-devices` flag (the clean route, confirmed by users in kubevirt/kubevirt#14335 but never documented)
- the containerd config template fallback, including the TOML duplicate-table pitfall: appending the cri runtime section after `{{ template "base" . }}` makes containerd fail to start with `table io.containerd.cri.v1.runtime already exists`, leaving the node NotReady
- the caveat that a full-copy template freezes the base config and should be reviewed after upgrades

Verified end to end on k3s v1.34.6 (containerd 2.2) with rook-ceph RBD block-mode PVCs: importer fails with `blockdev: cannot open /dev/cdi-block-volume: Permission denied` before the change, import and live migration succeed after.

Documentation follow-up to kubevirt/kubevirt#14335 (closed by the lifecycle bot without the fix being documented; a commenter there explicitly asked for docs).

Doc-only change.

```release-note
NONE
```

### DIFF
--- a/doc/block_cri_ownership_config.md
+++ b/doc/block_cri_ownership_config.md
@@ -39,5 +39,24 @@ CRI-O:
 device_ownership_from_security_context = true
 ```
 
+## k3s and RKE2 (managed containerd configuration)
+Distributions such as k3s and RKE2 generate the containerd `config.toml` on every service start, so editing the file directly does not persist: the change is overwritten on the next restart.
+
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
/lgtm
/approve
thank you!

akalenyu:
/retest
pv name empty flakes

PR #4172, opened 2026-06-16
Enable the WebhookPvcRendering feature gate by default

davmarro:
The **WebhookPvcRendering** feature gate was introduced as a temporary solution but it can lead to issues like, if not explicitly set, the users will never get the feature enabled.

So this PR enables the **WebhookPvcRendering** feature gate by default, so that the PVC mutating webhook is always deployed without requiring any feature gate configuration.
It adds also a **DisableWebhookPvcRendering** toggle in CDIConfigSpec to allow disabling the feature when needed.

**Changes:**
- `pkg/feature-gates/feature-gates.go`: `WebhookPvcRenderingEnabled()` returns `true` by default unless `DisableWebhookPvcRendering` is explicitly set
- `pkg/controller/datavolume/controller-base.go`: Remove redundant label assignment
- `staging/.../types.go`: Add `DisableWebhookPvcRendering *bool` to `CDIConfigSpec`
- `manifests/templates/release/cdi-cr.yaml.in`: Remove `WebhookPvcRendering` from default feature gates
- `tests/framework/framework.go`: Remove `WebhookPvcRendering` from test default feature gates
- E2E tests for the disable toggle (DataVolume fallback + standalone PVC behavior)
- Existing e2e tests (`webhook rendering` entries) already cover the default-enabled scenario

**Release note**:

```
Enable WebhookPvcRendering feature gate by default. The PVC mutating webhook is now always active without requiring explicit feature gate configuration. A new `DisableWebhookPvcRendering` field in CDIConfigSpec allows disabling it when needed.
```

### DIFF
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -1167,7 +1167,6 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume, 
 			return nil, err
 		}
 		if isWebhookPvcRenderingEnabled {
-			labels[common.PvcApplyStorageProfileLabel] = "true"
 			if targetPvcSpec.VolumeMode == nil {
 				targetPvcSpec.VolumeMode = ptr.To[corev1.PersistentVolumeMode](cdiv1.PersistentVolumeFromStorageProfile)
 			}
--- a/pkg/controller/datavolume/controller-base.go
+++ b/pkg/controller/datavolume/controller-base.go
@@ -1162,7 +1162,7 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume, 
 	annotations[cc.AnnCreatedForDataVolume] = string(dataVolume.UID)
 
 	if dataVolume.Spec.Storage != nil && labels[common.PvcApplyStorageProfileLabel] == "true" {
-		isWebhookPvcRenderingEnabled, err := featuregates.IsWebhookPvcRenderingEnabled(r.client)
+		isWebhookPvcRenderingEnabled, err := cc.IsWebhookPvcRenderingEnabled(r.client)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/feature-gates/feature-gates.go
+++ b/pkg/feature-gates/feature-gates.go
@@ -21,7 +21,10 @@ const (
 	// it is not an error if PVC of sam name exists before DataVolume is created
 	DataVolumeClaimAdoption = "DataVolumeClaimAdoption"
 
-	// WebhookPvcRendering - if enabled will deploy PVC mutating webhook for PVC rendering instead of the DV controller
+	// WebhookPvcRendering is the legacy feature gate string for PVC webhook rendering
+	// Deprecated: This feature is now always enabled
+	// The string is kept for backward compatibility with existing CDI CRs that list it in featureGates
+	// Use CDIConfigSpec.DisableWebhookPvcRendering to disable if needed
 	WebhookPvcRendering = "WebhookPvcRendering"
 )
 
@@ -48,7 +51,7 @@ func NewFeatureGates(c client.Client) *CDIConfigFeatureGates {
 }
 
 func (f *CDIConfigFeatureGates) isFeatureGateEnabled(featureGate string) (bool, error) {
-	featureGates, err := f.getConfig()
+	featureGates, err := f.getFeatureGates()
 	if err != nil {
 		return false, errors.Wrap(err, "error getting CDIConfig")
 	}
@@ -61,15 +64,6 @@ func (f *CDIConfigFeatureGates) isFeatureGateEnabled(featureGate string) (bool, 
 	return false, nil
 }
 
-func (f *CDIConfigFeatureGates) getConfig() ([]string, error) {
-	config := &cdiv1.CDIConfig{}
-	if err := f.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, config); err != nil {
-		return nil, err
-	}
-
-	return config.Spec.FeatureGates, nil
-}
-
 // HonorWaitForFirstConsumerEnabled - see the HonorWaitForFirstConsumer const
 func (f *CDIConfigFeatureGates) HonorWaitForFirstConsumerEnabled() (bool, error) {
 	return f.isFeatureGateEnabled(HonorWaitForFirstConsumer)
@@ -80,9 +74,34 @@ func (f *CDIConfigFeatureGates) ClaimAdoptionEnabled() (bool, error) {
 	return f.isFeatureGateEnabled(DataVolumeClaimAdoption)
 }
 
+func (f *CDIConfigFeatureGates) getCDIConfig() (*cdiv1.CDIConfig, error) {
+	config := &cdiv1.CDIConfig{}
+	if err := f.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, config); err != nil {
+		return nil, err
+	}
+	return config, nil
+}
+
+func (f *CDIConfigFeatureGates) getFeatureGates() ([]string, error) {
+	config, err := f.getCDIConfig()
+	if err != nil {
+		return nil, err
+	}
+	return config.Spec.FeatureGates, nil
+}
+
 // WebhookPvcRenderingEnabled tells if webhook PVC rendering is enabled
+// The webhook is enabled by default, it is only disabled when
+// CDIConfigSpec.DisableWebhookPvcRendering is set to true
 func (f *CDIConfigFeatureGates) WebhookPvcRenderingEnabled() (bool, error) {
-	return f.isFeatureGateEnabled(WebhookPvcRendering)
+	config, err := f.getCDIConfig()
+	if err != nil {
+		return false, errors.Wrap(err, "error getting CDIConfig")
+	}
+	if config.Spec.DisableWebhookPvcRendering != nil && *config.Spec.DisableWebhookPvcRendering {
+		return false, nil
+	}
+	return true, nil
 }
 
 // IsWebhookPvcRenderingEnabled tells if webhook PVC rendering is enabled
--- a/pkg/feature-gates/feature-gates.go
+++ b/pkg/feature-gates/feature-gates.go
@@ -24,7 +24,7 @@ const (
 	// WebhookPvcRendering is the legacy feature gate string for PVC webhook rendering
 	// Deprecated: This feature is now always enabled
 	// The string is kept for backward compatibility with existing CDI CRs that list it in featureGates
-	// Use CDIConfigSpec.DisableWebhookPvcRendering to disable if needed
+	// Use CDIConfigSpec.WebhookPvcRendering = "Disabled" to disable if needed
 	WebhookPvcRendering = "WebhookPvcRendering"
 )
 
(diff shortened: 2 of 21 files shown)


pkg/feature-gates/feature-gates_test.go
```diff
@@ -75,6 +75,60 @@ var _ = Describe("Feature Gates", func() {
...
 	})
+
+	It("WebhookPvcRendering should be enabled by default", func() {
+		featureGates, _ := createFeatureGatesAndClient()
+		enabled, err := featureGates.WebhookPvcRenderingEnabled()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(enabled).To(BeTrue())
+	})
+
+	It("WebhookPvcRendering should be disabled when DisableWebhookPvcRendering is true", func() {
```
awels: Can we combine all these 'It' tests into a DescribeTable? All the tests look very similar

awels (commented):
looks good minor nit pick

pkg/feature-gates/feature-gates_test.go
davmarro: @awels  good catch! 
Thanks, I'll update the tests

davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4172/commits/c235c55d6d91af2a1b31d06c8d3dc5969d4bfc72

pkg/feature-gates/feature-gates.go
```diff
@@ -80,9 +74,34 @@ func (f *CDIConfigFeatureGates) ClaimAdoptionEnabled() (bool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	return config.Spec.FeatureGates, nil
+}
+
 // WebhookPvcRenderingEnabled tells if webhook PVC rendering is enabled
+// The webhook is enabled by default, it is only disabled when
+// CDIConfigSpec.DisableWebhookPvcRendering is set to true
 func (f *CDIConfigFeatureGates) WebhookPvcRenderingEnabled() (bool, error) {
```
akalenyu: This helper probably makes more sense outside the feature gate pkg

tests/framework/framework.go
```diff
@@ -135,7 +135,7 @@ type Framework struct {
 // cannot work when called during test tree construction.
 func NewFramework(prefix string, config ...Config) *Framework {
 	cfg := Config{
-		FeatureGates: []string{featuregates.HonorWaitForFirstConsumer, featuregates.WebhookPvcRendering},
```
akalenyu, line 138: Should we also adjust the doc for the disabling path? https://github.com/kubevirt/containerized-data-importer/blob/7dc78e5cbfc006732e8458fd48f0ac9c14775e36/doc/pvc-mutating-webhook-rendering.md?plain=1#L13

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -1057,6 +1057,10 @@ type CDIConfigSpec struct {
 	PodResourceRequirements *corev1.ResourceRequirements `json:"podResourceRequirements,omitempty"`
 	// FeatureGates are a list of specific enabled feature gates
 	FeatureGates []string `json:"featureGates,omitempty"`
+	// DisableWebhookPvcRendering disables the PVC mutating webhook that completes
+	// PVC specs from StorageProfiles
+	// +optional
+	DisableWebhookPvcRendering *bool `json:"disableWebhookPvcRendering,omitempty"`
```
akalenyu: Not much conviction in this but bool is pretty limiting and thus usually harder to extend.
Normally k8s encourages something like `WebhookPvcRenderingPolicy` with `Enabled/Disabled`, defaulting to `Enabled`. Just wanted to get some discussion about this

tests/datavolume_test.go
```diff
@@ -2104,6 +2104,66 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests",
...
+			requestedSize := resource.MustParse("100Mi")
+			spec := cdiv1.StorageSpec{
+				Resources: v1.VolumeResourceRequirements{
+					Requests: v1.ResourceList{
+						v1.ResourceStorage: requestedSize,
+					},
+				},
+			}
+			dataVolume := createLabeledDataVolumeForImport(f, spec,
+				map[string]string{common.PvcApplyStorageProfileLabel: "true"})
```
akalenyu: TBH I am a bit confused myself about the use case for the rendering label on DVs. @arnongilboa Do you recall the motivation around it? Maybe we can drop this permutation of the test

akalenyu (commented):
looks great!

pkg/feature-gates/feature-gates_test.go
```diff
@@ -75,8 +75,33 @@ var _ = Describe("Feature Gates", func() {
...
+		Expect(enabled).To(Equal(expectedEnabled))
+	},
+		Entry("should be enabled by default", nil, nil, true),
+		Entry("should be disabled when DisableWebhookPvcRendering is true", boolPtr(true), nil, false),
+		Entry("should be enabled when DisableWebhookPvcRendering is false", boolPtr(false), nil, true),
+		Entry("should be enabled regardless of legacy featureGates list", nil, []string{}, true),
+	)
 })
 
+func boolPtr(b bool) *bool {
```
akalenyu: you can use the widely adopted ptr.To helper for things like that

pkg/feature-gates/feature-gates.go
davmarro: yes @akalenyu, totally agree.
What do you think if I'll move it in `pkg/controller/common/util.go`?
Keep in mind this change will lead to new files in the diff (and more noise in the PR) because this method is called from 3 different callers.

tests/framework/framework.go
davmarro: sure, I forgot it. thanks

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
davmarro: @akalenyu  ok, I didn't know about the k8s convention on this - just read about it. 
it's interesting.
Before adding the field, I checked in the current file that several `*bool` were already used for toggling features - that's why I followed the current approach.
Anyway, if we want to be more flexible I think this config option should be something like `PvcRenderingMode` with the following values :
- Webhook (enabled - default)
- Controller (disabled)

In this way we have a 1:1 map with the on/off and the possibility to extend the logic behind it.

pkg/feature-gates/feature-gates_test.go
davmarro: @akalenyu yes, I initially considered it but it would have required adding a new dependency in the BUILD.bazel file and honestly I didn't know if introducing unrelated build file changes in the PR (for testing purpose) was a good choice. 
Anyway if you think that it's ok to do it, I'm happy to switch to ptr.To.

pkg/feature-gates/feature-gates.go
akalenyu: yeah thats fine

pkg/feature-gates/feature-gates_test.go
akalenyu: yeah it's totally fine. it only goes into the test pkg of featuregates

pkg/feature-gates/feature-gates.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4172/commits/fc9766ae8197b276300fbaf79f409691949ce46c

tests/framework/framework.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4172/commits/88cd1bc4c01ab873a8773c5992e67d584db5be60

pkg/feature-gates/feature-gates_test.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4172/commits/fc9766ae8197b276300fbaf79f409691949ce46c

tests/datavolume_test.go
awels: I think the test is to verify that if someone puts the label on the DV (which makes no sense since we have a DV) it doesn't mess with the PVC.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
awels: I agree it is normally better to have policy so we can extend the approach if needed. But I can't think of a single possibility to extend this. It is either on or off. Which is pretty boolean. So I would be okay with the boolean flag.

davmarro: Yes, at the end I have decided to use a Enabled/Disabled policy just to be consistent with the k8s policy.
Thanks for the discussion!
https://github.com/kubevirt/containerized-data-importer/pull/4172/commits/04d104592987a263238c9807810f8cfce017031d

tests/datavolume_test.go
davmarro: I changed a bit the description of the test to be more clear and removed the useless label. The scope of the test is just to check that, when the webhook is toggled off, the controller is autocompleting the fields for the datavolume. 
https://github.com/kubevirt/containerized-data-importer/pull/4172/commits/c6fcb51fa79b26ee3643d3381d993c473ffdd65f

pkg/controller/datavolume/controller-base.go
```diff
@@ -1162,12 +1162,11 @@ func (r *ReconcilerBase) newPersistentVolumeClaim(dataVolume *cdiv1.DataVolume,
 	annotations[cc.AnnCreatedForDataVolume] = string(dataVolume.UID)
 
 	if dataVolume.Spec.Storage != nil && labels[common.PvcApplyStorageProfileLabel] == "true" {
-		isWebhookPvcRenderingEnabled, err := featuregates.IsWebhookPvcRenderingEnabled(r.client)
+		isWebhookPvcRenderingEnabled, err := cc.IsWebhookPvcRenderingEnabled(r.client)
 		if err != nil {
 			return nil, err
 		}
 		if isWebhookPvcRenderingEnabled {
-			labels[common.PvcApplyStorageProfileLabel] = "true"
```
akalenyu, line 1170: @arnongilboa for your thoughts 🙏

tests/datavolume_test.go
arnongilboa: @akalenyu the original idea (like @awels said), was that in this case the created PVC will get the label and rendered by the webhook and not the controller (we want to deterministically choose one of them to do the rendering). Although it's not an interesting use case, we should somehow handle and test it, unless we explicitly reject/ignore/filter the label if it appears on DV.

pkg/controller/datavolume/controller-base.go
arnongilboa: https://github.com/kubevirt/containerized-data-importer/pull/4172#discussion_r3453035349

arnongilboa:
Looks good, just a minor [comment](https://github.com/kubevirt/containerized-data-importer/pull/4172#discussion_r3453035349).

akalenyu (commented):
/approve
great stuff!

PR #4173, opened 2026-06-16
populators: fix race between ann/label propagation to target PVC

akalenyu:
In the populator path, annotations (including AnnPodPhase) and labels
were copied from PVC Prime to the target PVC in two separate API
updates. The DV controller could observe AnnPodPhase=Succeeded from
the first update and transition the DV to Succeeded before labels
(e.g. instancetype.kubevirt.io/*) landed in the second update. This
caused the DataImportCron controller to create VolumeSnapshots without
the registry image labels, and since the recovery path for snapshots
is circular (snapshot->DataSource->snapshot), the labels were
permanently lost.

Refactor updatePVCWithPVCPrimeAnnotations and
updatePVCWithPVCPrimeLabels to only mutate in-memory, letting callers
issue a single client.Update after applying all mutations. This
ensures annotations and labels are always visible atomically.

Fixes #

**Release note**:

```release-note
BugFix: instancetype labels from containerdisk doesn't get propogated to VolumeSnapshot/DataSource on multiarch clusters
```

### DIFF
--- a/pkg/controller/populators/forklift-populator.go
+++ b/pkg/controller/populators/forklift-populator.go
@@ -354,12 +354,8 @@ func (r *ForkliftPopulatorReconciler) reconcileTargetPVC(pvc, pvcPrime *corev1.P
 }
 
 func (r *ForkliftPopulatorReconciler) updatePVCPrime(pvc, pvcPrime *corev1.PersistentVolumeClaim) error {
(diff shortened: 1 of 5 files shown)


akalenyu:
/close
refactor is going to be for main, lets start with pinpointed fix

PR #4180, opened 2026-06-22
storagecapabilities: advertise RWX-Filesystem for linstor.csi.linbit.com

kvaps:

Add `{ReadWriteMany, Filesystem}` to the storage-capabilities set for `linstor.csi.linbit.com`, after the block modes:

```
{rwx, block}, {rwo, block}, {rwx, file}, {rwo, file}, {rox, block}, {rox, file}
```

LINSTOR can serve RWX in Filesystem mode through its highly-available NFS-Ganesha export (`linstor-csi-nfs-server`). Advertising it lets CDI/KubeVirt satisfy shared-filesystem workloads on LINSTOR. Concrete motivation: KubeVirt's vm-state backend-storage PVC (persistent EFI/TPM) is Filesystem-mode and must be ReadWriteMany to live-migrate; without `{rwx, file}` in the profile it is forced to ReadWriteOnce on a LINSTOR default class, pinning the VM to its node and blocking live-migration.

This also gives `linstor.csi.linbit.com` its own documented constructor instead of sharing the generic `createAllButRWXFileCapabilities()` helper with Dell Unity / Huawei.

**Ordering**: `{rwx, block}` stays first, so a fully-unspecified request still defaults to Block (DRBD dual-primary, no NFS) — the right default for normal VM disks. `{rwx, file}` is placed before `{rwo, file}`, so a request that pins `volumeMode=Filesystem` without an accessMode now defaults to RWX, i.e. a per-volume NFS-Ganesha export.

/cc @WanzenBug — as the LINBIT maintainer of linstor-csi and author of the NFS-Ganesha RWX feature, your review is the one that matters here. Two things to confirm:
1. advertising `{rwx, file}` for `linstor.csi.linbit.com` is correct now that the NFS export exists, and
2. the ordering — `{rwx, file}` before `{rwo, file}` makes a bare Filesystem request default to an NFS share; placing it after `{rwo, file}` keeps `{rwo, file}` the Filesystem default and reaches NFS only on an explicit RWX request. The current placement is the more aggressive one.

**Release note**:

```release-note
storagecapabilities: advertise ReadWriteMany + Filesystem for linstor.csi.linbit.com (served via LINSTOR's NFS-Ganesha export), enabling shared-filesystem and live-migratable vm-state volumes on LINSTOR.
```

### DIFF
--- a/pkg/storagecapabilities/storagecapabilities.go
+++ b/pkg/storagecapabilities/storagecapabilities.go
@@ -55,7 +55,7 @@ var CapabilitiesByProvisionerKey = map[string][]StorageCapabilities{
 	"rook-ceph.cephfs.csi.ceph.com":         {{rwx, file}, {rwo, file}},
 	"openshift-storage.cephfs.csi.ceph.com": {{rwx, file}, {rwo, file}},
 	// LINSTOR
-	"linstor.csi.linbit.com": createAllButRWXFileCapabilities(),
+	"linstor.csi.linbit.com": createLinstorCapabilities(),
 	// DELL Unity XT
 	"csi-unity.dellemc.com":     createAllButRWXFileCapabilities(),
 	"csi-unity.dellemc.com/nfs": createAllFSCapabilities(),
@@ -557,6 +557,28 @@ func createAllButRWXFileCapabilities() []StorageCapabilities {
 	return []StorageCapabilities{
 		{rwx, block},
 		{rwo, block},
+		{rwo, file},
+		{rox, block},
+		{rox, file},
+	}
+}
+
+// createLinstorCapabilities returns the capabilities for linstor.csi.linbit.com.
+//
+// Block is listed first so it stays the preferred volume mode: on LINSTOR/DRBD
+// shared access (RWX) is served natively at the Block level via DRBD
+// dual-primary, so a fully-unspecified request defaults to {rwx, block} (no
+// NFS). {rwx, file} is also advertised, after the block modes, because LINSTOR
+// can serve RWX in Filesystem mode through its highly-available NFS-Ganesha
+// export (linstor-csi-nfs-server). Advertising it lets shared-filesystem
+// workloads be satisfied on LINSTOR — notably KubeVirt vm-state backend storage,
+// which is Filesystem-mode and must be RWX to live-migrate. {rwo, file} is kept
+// for ordinary non-shared filesystem volumes.
+func createLinstorCapabilities() []StorageCapabilities {
+	return []StorageCapabilities{
+		{rwx, block},
+		{rwo, block},
+		{rwx, file},
 		{rwo, file},
 		{rox, block},
 		{rox, file},


kvaps:
Related upstream work: kubevirt/kubevirt#18215 (draft) is the complementary approach — keep the vm-state on a raw **Block** (RWX-Block) volume directly, via DRBD dual-primary, with no NFS. This PR (RWX-Filesystem via LINSTOR's NFS-Ganesha export) and that one are two paths to the same goal: live-migratable persistent EFI/TPM on LINSTOR. The choice affects whether shared vm-state state lands on a per-volume NFS share or on raw block. cc @WanzenBug for the LINSTOR view on both.

pkg/storagecapabilities/storagecapabilities.go
```diff
@@ -563,6 +563,28 @@ func createAllButRWXFileCapabilities() []StorageCapabilities {
...
+// createLinstorCapabilities returns the capabilities for linstor.csi.linbit.com.
+//
+// Block is listed first so it stays the preferred volume mode: on LINSTOR/DRBD
+// shared access (RWX) is served natively at the Block level via DRBD
+// dual-primary, so a fully-unspecified request defaults to {rwx, block} (no
+// NFS). {rwx, file} is also advertised, after the block modes, because LINSTOR
+// can serve RWX in Filesystem mode through its highly-available NFS-Ganesha
+// export (linstor-csi-nfs-server). Advertising it lets shared-filesystem
+// workloads be satisfied on LINSTOR — notably KubeVirt vm-state backend storage,
+// which is Filesystem-mode and must be RWX to live-migrate. {rwo, file} is kept
```
akalenyu, line 575: thanks for the PR! vm state rwo fs is live migratable since quite a while https://github.com/kubevirt/kubevirt/pull/12629

WanzenBug (commented):
Looks good to me.

If indeed File+RWO is supported for VM state, I would go for that option. It's the more "native" solution with less moving parts.

kvaps:
Thanks @WanzenBug! One clarification — the problem with File+RWO is exactly live-migration: with RWO the backend PVC pins the VM to its node (the target `virt-launcher` pod can't attach it during the migration window), so a persistent-EFI/TPM VM can't be live-migrated. File+RWO is what's used by default today, and that's the limitation.

The proposal with this fix: File+**RWX** becomes the default for VM state, since LINSTOR supports it — that's what makes those VMs live-migratable.

Btw, I'm separately working on storing the state directly on block devices (kubevirt/kubevirt#18215 + a VEP, kubevirt/enhancements#349) — but that's a longer path than simply allowing File+RWX for LINSTOR here.

akalenyu:
> Thanks @WanzenBug! One clarification — the problem with File+RWO is exactly live-migration: with RWO the backend PVC pins the VM to its node (the target `virt-launcher` pod can't attach it during the migration window), so a persistent-EFI/TPM VM can't be live-migrated. File+RWO is what's used by default today, and that's the limitation.
> 
> The proposal with this fix: File+**RWX** becomes the default for VM state, since LINSTOR supports it — that's what makes those VMs live-migratable.
> 
> Btw, I'm separately working on storing the state directly on block devices ([kubevirt/kubevirt#18215](https://github.com/kubevirt/kubevirt/pull/18215) + a VEP, [kubevirt/enhancements#349](https://github.com/kubevirt/enhancements/issues/349)) — but that's a longer path than simply allowing File+RWX for LINSTOR here.

I'm confused. The strategy in https://github.com/kubevirt/kubevirt/pull/12629 (which has been around for a long time) should achieve a live migration that is transparent and good enough for any user with RWO FS. What am I missing?

awels:
Right, state PVCs can be RWO, we simply create a new one on the target and copy the contents during the migration. This doesn't affect the migration much at all since the state PVC is so small.

kvaps:
You're right, thanks @awels @akalenyu @WanzenBug — I was carrying the pre-#12629 mental model. RWO-Filesystem backend storage already live-migrates: on migration KubeVirt creates a fresh target state PVC and copies the small blob across (kubevirt/kubevirt#12629), so advertising File+RWX for LINSTOR isn't needed for migratability. I confirmed this separately (a persistent-EFI VM on an RWO-FS backend reports `LiveMigratable=True` and the copy-on-target PVC is created on migration).

So this change isn't needed — closing it. Sorry for the noise.

For the record, I've also reframed the related kubevirt/kubevirt#18215 and its VEP accordingly: Block-mode backend storage is now presented as an optional *block-native* feature (state on the same block storage as the VM's disks, and on RWX-Block no per-migration copy), **not** a migration fix.

PR #4188, opened 2026-06-26
Remove prlimit (RLIMIT_AS) address space limiting from qemu-img execution

### LINKED ISSUE #4186: Remove prlimit (RLIMIT_AS) address space limiting from qemu-img execution

CDI currently uses `prlimit(RLIMIT_AS)` to limit the virtual address space of `qemu-img` subprocesses. The limit originates from an [OpenStack Nova recommendation](https://github.com/kubevirt/containerized-data-importer/blob/43ff23ec3a238cd49707448601dd215d1c761260/pkg/image/qemu.go#L44) (1 GiB) and was intended to prevent malicious disk images from causing `qemu-img` to allocate arbitrary amounts of memory.

However, `RLIMIT_AS` limits **virtual** address space, not physical (RSS) memory. This is a problematic distinction that has caused real issues:

1. **Incompatibility with Go runtime** — The Go runtime reserves large virtual address spaces via `mmap` (for GC arenas, span management, etc.) that never translate to physical memory usage. Setting `RLIMIT_AS` on Go programs causes sporadic crashes. This is [confirmed upstream](https://github.com/golang/go/issues/69105#issuecomment-2315846394) as expected behavior, not a bug. The Go team effectively considers `RLIMIT_AS` incompatible with Go programs. This has caused s390x unit test failures with `SIGSEGV` (#3341), sporadic amd64 unit test failures with `runtime: cannot allocate memory`, and architecture-dependent flakiness due to varying `mmap` layouts.

2. **Redundant in a containerized environment** — CDI runs in Kubernetes pods with container-level resource limits. The container runtime enforces memory limits via **cgroups** (which limit RSS/physical memory), providing a more robust and appropriate mechanism than `RLIMIT_AS`. The address space limit adds no meaningful protection on top of cgroup constraints.

3. **Limited practical scope** — The address space limit is currently only applied to `qemu-img info` calls ([qemu.go:222](https://github.com/kubevirt/containerized-data-importer/blob/43ff23ec3a238cd49707448601dd215d1c761260/pkg/image/qemu.go#L222)). All other `qemu-img` operations (convert, resize, create, rebase, commit) already pass `nil` limits. The CPU time limit + context timeout already protect against hung processes.

4. **The original motivation no longer applies** — The 1 GiB limit was borrowed from OpenStack Nova, which runs `qemu-img` in a very different environment (bare-metal hypervisors without container-level isolation). In Kubernetes, pod resource limits are the standard mechanism for constraining memory.

Remove the entire `prlimit` syscall mechanism:
1. Remove `SetAddressSpaceLimit`, `SetCPUTimeLimit`, and the `prlimit` syscall wrapper from `pkg/system/prlimit.go`
2. Remove `AddressSpaceLimit` from `ProcessLimitValues`
3. Remove `maxMemory` and `qemuInfoLimits` address space usage from `pkg/image/qemu.go`
4. Remove or simplify `prlimit_test.go` (the memory/CPU limit test cases)
5. Keep `ExecWithLimits` — it still provides value via `context.WithTimeout` for killing hung subprocesses (e.g. the 30s timeout on `qemu-img info`), plus stdout/stderr scanning and progress callbacks. `SetCPUTimeLimit` (`RLIMIT_CPU`) is unused in production code (only in tests) — the context timeout is the actual mechanism protecting against hangs.

- Bumping the limit to 2 GiB (architecture-dependent) — this was discussed in #3341 but is a band-aid; the fundamental incompatibility between `RLIMIT_AS` and Go's virtual memory model remains, and the limit value would need ongoing tuning across architectures and Go versions.
- Switching to `RLIMIT_RSS` — not reliably enforced by the Linux kernel; cgroups are the proper mechanism for RSS limiting.
- Replacing `RLIMIT_AS` with a transient cgroup for physical memory limiting — instead of `prlimit`, we could create a child cgroup per `qemu-img` invocation with a `memory.max` limit. This can be done either via [`systemd-run --scope -p MemoryMax=<limit>`](https://utcc.utoronto.ca/~cks/space/blog/linux/SystemdForMemoryLimitingII) wrapping the command, or programmatically using the [`containerd/cgroups`](https://github.com/containerd/cgroups) Go library (create a cgroup v2 manager with `memory.max` set, add the child PID, clean up on exit). This would correctly limit RSS/physical memory rather than virtual address space. However, it adds complexity (cgroup filesystem permissions, cleanup, cgroup v1/v2 compat) and may be overkill given that Kubernetes pod-level resource limits already constrain the importer container's total memory.

**Additional context**:

- #3341 — prlimit unit test fails on s390x
- #3226 — original prlimit s390x issue
- https://github.com/golang/go/issues/69105 — Go upstream: `RLIMIT_AS` is incompatible with Go programs

kenlacroix:

CDI used `prlimit(RLIMIT_AS)` to cap the address space of `qemu-img` subprocesses at 1 GiB (a value borrowed from OpenStack Nova). `RLIMIT_AS` limits *virtual* address space, not physical/RSS memory. The Go runtime reserves large virtual address ranges via `mmap` (GC arenas, span management) that never translate to physical memory, so `RLIMIT_AS` is fundamentally incompatible with Go programs ([golang/go#69105](https://github.com/golang/go/issues/69105)) and has caused `SIGSEGV` / `runtime: cannot allocate memory` flakiness across architectures (e.g. s390x, #3341).

The protection is also redundant and narrow:
- CDI runs in Kubernetes pods, where the container runtime already enforces real (RSS) memory limits via cgroups.
- The limit was only ever applied to the `qemu-img info` call; all other `qemu-img` operations already passed `nil` limits.

This PR removes the `RLIMIT_AS`/`RLIMIT_CPU` prlimit syscall mechanism and the now-unused `ProcessLimiter` abstraction.

`ExecWithLimits` and `ProcessLimitValues.CPUTimeLimit` are intentionally retained — `CPUTimeLimit` still drives the `context.WithTimeout` that kills hung subprocesses (e.g. the 30s `qemu-img info` timeout). Only the `prlimit`-based limiting (the unused `SetCPUTimeLimit`/`RLIMIT_CPU` and the `SetAddressSpaceLimit`/`RLIMIT_AS` path) is removed. `go build`, `go vet`, `gofmt`, and `go test ./pkg/system/... ./pkg/image/...` all pass locally.

Fixes #4186

**Release note**:

```release-note
Removed the RLIMIT_AS address space limit on qemu-img subprocesses, which was incompatible with the Go runtime's virtual memory model and redundant with Kubernetes pod/cgroup memory limits.
```

### DIFF
--- a/pkg/image/qemu.go
+++ b/pkg/image/qemu.go
@@ -40,9 +40,8 @@ import (
 )
 
 const (
-	networkTimeoutSecs = 3600    //max is 10000
-	maxMemory          = 1 << 30 //value from OpenStack Nova
-	maxCPUSecs         = 30      //value from OpenStack Nova
+	networkTimeoutSecs = 3600 //max is 10000
+	maxCPUSecs         = 30   //value from OpenStack Nova
 	matcherString      = "\\((\\d?\\d\\.\\d\\d)\\/100%\\)"
 )
 
@@ -75,7 +74,7 @@ var (
 	ErrLargerPVCRequired = errors.New("A larger PVC is required")
 
 	qemuExecFunction = system.ExecWithLimits
-	qemuInfoLimits   = &system.ProcessLimitValues{AddressSpaceLimit: maxMemory, CPUTimeLimit: maxCPUSecs}
+	qemuInfoLimits   = &system.ProcessLimitValues{CPUTimeLimit: maxCPUSecs}
 	qemuIterface     = NewQEMUOperations()
 	re               = regexp.MustCompile(matcherString)
 
--- a/pkg/image/qemu_test.go
+++ b/pkg/image/qemu_test.go
@@ -150,7 +150,7 @@ func init() {
 	ownerUID = "1111-1111-111"
 }
 
-var expectedLimits = &system.ProcessLimitValues{AddressSpaceLimit: 1 << 30, CPUTimeLimit: 30}
+var expectedLimits = &system.ProcessLimitValues{CPUTimeLimit: 30}
 
 var _ = Describe("Convert to Raw", func() {
 	var tmpDir, destPath string
--- a/pkg/system/BUILD.bazel
+++ b/pkg/system/BUILD.bazel
@@ -7,7 +7,6 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//vendor/github.com/pkg/errors:go_default_library",
-        "//vendor/golang.org/x/sys/unix:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
     ],
 )
--- a/pkg/system/prlimit.go
+++ b/pkg/system/prlimit.go
@@ -21,57 +21,20 @@ import (
 	"bytes"
 	"context"
 	"os/exec"
-	"syscall"
 	"time"
-	"unsafe"
 
 	"github.com/pkg/errors"
-	"golang.org/x/sys/unix"
 
 	"k8s.io/klog/v2"
 )
 
-// ProcessLimiter defines the methods limiting resources of a Process
-type ProcessLimiter interface {
-	SetAddressSpaceLimit(pid int, value uint64) error
-	SetCPUTimeLimit(pid int, value uint64) error
-}
-
 // ProcessLimitValues specifies the resource limits available to a process
 type ProcessLimitValues struct {
-	AddressSpaceLimit uint64
-	CPUTimeLimit      uint64
+	CPUTimeLimit uint64
 }
-
-type processLimiter struct{}
 
 var execCommand = exec.Command
 var execCommandContext = exec.CommandContext
-
-var limiter = NewProcessLimiter()
-
-// NewProcessLimiter returns a new ProcessLimiter
-func NewProcessLimiter() ProcessLimiter {
-	return &processLimiter{}
-}
-
-func (p *processLimiter) SetAddressSpaceLimit(pid int, value uint64) error {
-	return prlimit(pid, unix.RLIMIT_AS, &syscall.Rlimit{Cur: value, Max: value})
-}
-
-func (p *processLimiter) SetCPUTimeLimit(pid int, value uint64) error {
-	return prlimit(pid, unix.RLIMIT_CPU, &syscall.Rlimit{Cur: value, Max: value})
-}
-
-// SetAddressSpaceLimit sets a limit on total address space of a process
-func SetAddressSpaceLimit(pid int, value uint64) error {
-	return limiter.SetAddressSpaceLimit(pid, value)
-}
-
-// SetCPUTimeLimit sets a limit on the total cpu time a process may have
-func SetCPUTimeLimit(pid int, value uint64) error {
-	return limiter.SetCPUTimeLimit(pid, value)
-}
 
 // scanLinesWithCR is an alternate split function that works with carriage returns as well
 // as new lines.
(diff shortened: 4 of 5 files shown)


pkg/image/qemu.go
```diff
@@ -75,7 +74,7 @@ var (
 	ErrLargerPVCRequired = errors.New("A larger PVC is required")
 
 	qemuExecFunction = system.ExecWithLimits
```
akalenyu, line 76: ExecWithLimits is just `exec.CommandContext` at this point, maybe it's worth just whacking the entire prlimit pkg and using that directly (alongside the parsing logic etc that's also apparently part of prlimit logic)

akalenyu (commented):
Thank you!

akalenyu:
notes from sig-storage call:
we should have e2e tests covering the original use case around the prlimit pkg, worth to check those are:
- honest
- passing without prlimit

kenlacroix:
Thanks for running the tests, @akalenyu — pushed `e5ac0f680`:

**`generate-verify` / `verify-go-mod`:** both were failing on the same stale artifact. Dropping the `RLIMIT_AS` code removed the last `golang.org/x/sys/unix` use in `pkg/system`, so gazelle no longer lists it as a `go_library` dep. I regenerated `pkg/system/BUILD.bazel` to match; that's the only generated file the change affects.

**The two e2e reds look like infra, not this change:**
- `pull-...-e2e-hpp-latest` finished as `error` with no `build-log.txt` produced at all (`NoSuchKey` on GCS) — it never got to running tests.
- `pull-...-e2e-s390x` failed in `[SynchronizedBeforeSuite]` ("A BeforeSuite node failed so all tests were skipped"), i.e. cluster/env setup, not a test assertion.

Re the sig-storage point on e2e coverage of the original `prlimit` use case: the pkg wraps `qemu-img` execution, and that path is exercised today by the functional suites — e.g. `tests/import_test.go` *"Should perform fsync syscall after qemu-img convert to raw"* (qcow2→raw convert), the qcow2 import + preallocation cases, and the import/transport/datavolume suites, all of which drive `qemu-img convert`/`info` through `system.ExecWithLimits`. After this change `ExecWithLimits` still enforces the CPU-time hang-guard via `context.WithTimeout` (`ProcessLimitValues.CPUTimeLimit` is untouched); only the `RLIMIT_AS` address-space cap is removed. So the conversion behavior those tests assert is unchanged and still covered — happy to add a dedicated large-image conversion e2e if you'd prefer an explicit regression test for the no-`RLIMIT_AS` path.

akalenyu:
could you just sign off the commit? --amend --signoff

pkg/image/qemu.go
akalenyu: @kenlacroix wdyt?

davmarro: Hi @akalenyu @kenlacroix,
yes it makes sense to delete the entire `system` package. 
Anyway, all the references to the `ExecWithLimits` method are in the [qemu.go](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/image/qemu.go) file and they can be grouped into 2 different categories:
- The ones that are not using the **reportProgress** (`Info`, `Resize`, `CreateBlankImage`, `dd`): they can be totally replaced with the `exec.CommandContext(...).Output()` method
- The ones that are using the **reportProgress** (`Convert`, `Rebase`, `Commit`): they need the creation of a helper (to report the callback realtime progress) and the migration of the `scanLinesWithCR` method from the deleted `system` package.

If you agree with this approach, I can take care of it.

davmarro:
Hi @akalenyu @kenlacroix,
this [PR](https://github.com/kubevirt/containerized-data-importer/pull/4200) already covered the issue, so this PR can be closed. Thanks

PR #4189, opened 2026-06-29
Suppress ClaimMisbound Warning

HoustonBoston:

Fixes #

**Release note**:

```release-note
Supress claims misbound warning
```


pkg/controller/common/util_test.go
```diff
@@ -276,6 +279,60 @@ var _ = Describe("Rebind", func() {
...
+			Build()
+
+		recorder := record.NewFakeRecorder(10)
+		CopyEvents(&pvcPrime, &targetPVC, cl, recorder)
+		// verify no ClaimMisbound event was recorded
+		close(recorder.Events)
+		for event := range recorder.Events {
+			println("Looking at event: ", event)
+			if strings.Contains(event, "ClaimMisbound") {
+				Fail("ClaimMisbound event was recorded")
```
dsanatar: instead of calling `Fail()` here, you can use an `Expect()` which will fail when the condition isn't met

pkg/controller/common/util_test.go
```diff
@@ -276,6 +279,60 @@ var _ = Describe("Rebind", func() {
...
+			InvolvedObject: v1.ObjectReference{
+				Name: pvcPrime.Name,
+				UID:  pvcPrime.UID,
+			},
+			Type: v1.EventTypeWarning,
+		}
+		s := scheme.Scheme
+		cl := fake.NewClientBuilder().
+			WithScheme(s).
+			WithRuntimeObjects(&pvcPrime, &targetPVC, &claimMisboundEvent).
```
dsanatar: might be useful to emit another event just to see that we got the valid one and filtered out the bad one.

pkg/controller/common/util_test.go
```diff
@@ -276,6 +279,60 @@ var _ = Describe("Rebind", func() {
...
+			Reason:  "ClaimMisbound",
+			Message: "Claim is misbound",
+			InvolvedObject: v1.ObjectReference{
+				Name: pvcPrime.Name,
+				UID:  pvcPrime.UID,
+			},
+			Type: v1.EventTypeWarning,
+		}
+		s := scheme.Scheme
+		cl := fake.NewClientBuilder().
```
dsanatar: It may make more sense to have this test in `import-populator_test.go`, since they already have the recorder and clientBuilder initialization logic baked in to helper functions

dsanatar (changes requested):
looks good overall, just some minor things. Might also be good to squash some of the commits that add/remove the tests

pkg/controller/common/util.go
```diff
@@ -2373,6 +2373,12 @@ func CopyEvents(srcPVC, targetPVC client.Object, c client.Client, recorder recor
 		if _, exists := eventMap[formattedMsg]; exists {
 			continue
 		}
+
+		// simply check if event is a ClaimMisbound event, if so, skip it
+		if newEvent.Reason == "ClaimMisbound" {
```
awels: Maybe make this a constant like `EventReasonClaimMisbound = "ClaimMisbound"` this way we have it in one place if we ever need to change it.

pkg/controller/common/util_test.go
```diff
@@ -276,6 +279,60 @@ var _ = Describe("Rebind", func() {
...
+			WithIndex(&v1.Event{}, "involvedObject.uid", func(obj client.Object) []string {
+				return []string{string(obj.(*v1.Event).InvolvedObject.UID)}
+			}).
+			Build()
+
+		recorder := record.NewFakeRecorder(10)
+		CopyEvents(&pvcPrime, &targetPVC, cl, recorder)
+		// verify no ClaimMisbound event was recorded
+		close(recorder.Events)
+		for event := range recorder.Events {
```
awels: Instead of doing a for loop here, you should be able to do something like
```go
Expect(recorder.Events).ToNot(ContainElement(EventReasonClaimMisbound))
```
This reads much easier to humans.

PR #4190, opened 2026-06-29
Suppress ClaimMisbound Warning

HoustonBoston:

This PR suppresses the ClaimMisbound warning such that this warning does not enter up the Kubernetes chain. This is an expected behavior when dealing with KubeVirt's DataVolume resource, and the end user should not think there's an issue when in fact there isn't.

Fixes #

There was an earlier PR except I'm merging from my origin's dev branch rather than its main branch, which is why you may see a duplicate PR.

**Release note**:

```release-note
Suppress ClaimMisbound Warning from prime PVCs
```

### DIFF
--- a/pkg/controller/populators/import-populator_test.go
+++ b/pkg/controller/populators/import-populator_test.go
@@ -198,6 +198,69 @@ var _ = Describe("Import populator tests", func() {
 			Expect(found).To(BeFalse())
 		})
 
+		It("Should not copy ClaimMisbound event from primePVC to targetPVC", func() {
+			targetPVC := CreatePvcInStorageClass(targetPvcName, metav1.NamespaceDefault, &sc.Name, nil, nil, corev1.ClaimPending)
+			targetPVC.Spec.DataSourceRef = dataSourceRef
+			volumeImportSource := getVolumeImportSource(true, nsName)
+
+			pvcPrime := getPVCPrime(targetPVC, nil) // automatically creates a pvcPrime with the same name as the targetPVC
+
+			importSucceededEvent := corev1.Event{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "importSucceededEvent",
+					Namespace: targetPVC.Namespace,
+				},
+				InvolvedObject: corev1.ObjectReference{
+					Name: pvcPrime.Name,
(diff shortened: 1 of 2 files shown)


pkg/controller/populators/import-populator_test.go
```diff
@@ -198,6 +198,47 @@ var _ = Describe("Import populator tests", func() {
 			Expect(found).To(BeFalse())
 		})
 
+		It("Should not copy ClaimMisbound events from srcPVC to targetPVC", func() {
```
dsanatar:
```suggestion
		It("Should not copy ClaimMisbound event from primePVC to targetPVC", func() {
```

pkg/controller/populators/import-populator_test.go
```diff
@@ -198,6 +198,47 @@ var _ = Describe("Import populator tests", func() {
...
+			close(recorder.Events)
+
+			// NOTE: ContainsElement expects a slice, which recorder.Events is not
+			claimMisboundEventfound := false
+			importSucceededEventFound := false
+			for event := range recorder.Events {
+				if strings.Contains(event, fmt.Sprintf("%s [%s]", EventReasonClaimMisbound, pvcPrime.Name)) {
+					claimMisboundEventfound = true
+				}
+				if strings.Contains(event, importSucceeded) {
```
dsanatar: This only checks that the normal "import succeeded" event exists, we should see that this event got copied to the target pvc, i.e checking that it has the pvcPrime name in the event message.

pkg/controller/populators/import-populator_test.go
```diff
@@ -198,6 +198,47 @@ var _ = Describe("Import populator tests", func() {
...
+					importSucceededEventFound = true
+				}
+
+				if claimMisboundEventfound && importSucceededEventFound {
+					break
```
dsanatar: I don't think we need this break

dsanatar (changes requested):

dsanatar:
you need to add the release note to the PR description to get rid of the label.
/test all

dsanatar:
s390 lane seems super broken, however it's not marked as a "required" lane for this to merge. istio looks to have flaked
/test pull-containerized-data-importer-e2e-istio

PR #4193, opened 2026-07-01
Solve G115 Linter Errors

### LINKED ISSUE #3656: Evaluate G115 linter errors
The G115 linter will be disabled following https://github.com/kubevirt/containerized-data-importer/pull/3652;
However, it may uncover real issues in some cases.

Evaluate the G115 errors and exclude on a case by case basis instead of globally

**Additional context**:

HoustonBoston:

Rather than excluding G115 for linting globally, this fix evaluates type conversions on a case-by-case basis, changing types of variables and completely ignoring the integer conversion where applicable so that linting succeeds.

Fixes #3656. 

**Release note**:

```release-note
NONE
```


akalenyu (commented):
Thanks for tackling this!
We're definitely going to keep nolints on some specific cases, but I was wondering if we
could get rid of the nolints where the value is known to be non-negative by domain (lengths, sizes, offsets)?

Here's a reference to a large project enabling G115 that may help - https://github.com/elastic/cloud-on-k8s/pull/9302

awels (approved):

cmd/cdi-controller/controller.go
```diff
@@ -372,7 +373,7 @@ func main() {
 		klog.Fatalf("Unable to get environment variables: %v\n", errors.WithStack(err))
 	}
 
-	logf.SetLogger(zap.New(zap.Level(zapcore.Level(-1*verbosityLevel)), zap.UseDevMode(debug)))
+	logf.SetLogger(zap.New(zap.Level(zapcore.Level(-1*int8(min(verbosityLevel, math.MaxInt8)))), zap.UseDevMode(debug))) //nolint:gosec // G115: clamped to MaxInt8 above
```
akalenyu, line 376: do you still need the nolint? that's weird

pkg/importer/file.go
```diff
@@ -58,7 +58,7 @@ func GetAvailableSpace(path string) (int64, error) {
 		return int64(-1), err
 	}
 	//nolint:unconvert
-	return int64(stat.Bavail) * int64(stat.Bsize), nil
+	return int64(stat.Bavail) * int64(stat.Bsize), nil //nolint:gosec
```
akalenyu: shouldn't we tackle this properly instead of nolint? i.e. ensure that both individually and their product does not overflow?

pkg/util/util.go
```diff
@@ -83,7 +83,7 @@ func ParseEnvVar(envVarName string, decode bool) (string, error) {
 // Read reads bytes from the stream and updates the prometheus clone_progress metric according to the progress.
 func (r *CountingReader) Read(p []byte) (n int, err error) {
 	n, err = r.Reader.Read(p)
-	r.Current += uint64(n)
+	r.Current += uint64(n) //nolint:gosec
```
akalenyu: i think we can avoid the nolint here too by performing checks on n

akalenyu (commented):
thanks! some thoughts plus I think some of the nolints are not needed post your changes, did you try removing them?

cmd/cdi-controller/controller.go
HoustonBoston: Yeah, it's pretty bizarre it's still needed

pkg/importer/file.go
```diff
@@ -57,8 +58,18 @@ func GetAvailableSpace(path string) (int64, error) {
...
-	return int64(stat.Bavail) * int64(stat.Bsize), nil
+	if stat.Bavail > math.MaxInt64 {
+		return math.MaxInt64, nil
+	}
+	avail := int64(stat.Bavail)
+	bsize := int64(stat.Bsize) //nolint:unconvert // type varies by platform
+	if bsize <= 0 {
+		return 0, nil
+	}
+	if avail > math.MaxInt64/bsize {
```
akalenyu, line 69: this reads a little worse than `avail * bsize > maxint`, no?

pkg/importer/file.go
```diff
@@ -57,8 +58,18 @@ func GetAvailableSpace(path string) (int64, error) {
...
+		return math.MaxInt64, nil
```
akalenyu: I think we should just error out for all those funky cases instead of returning maxint

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -774,33 +769,39 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
 	}
 
 	if extent.Length < 1024*1024 {
+		if extent.Start < 0 || extent.Length < 0 {
```
akalenyu, line 777: does this check not eliminate the need for the nolint below?

akalenyu (commented):
i honestly dont mind keeping nolints on the vddk/imageio data source - mainly curious about nolint being needed. so just file.go comments really🙏

pkg/importer/file.go
HoustonBoston: What number should I return and what should the error be?

pkg/importer/file.go
HoustonBoston: Doesn't the multiplication cause overflow though?

pkg/importer/vddk-datasource_amd64.go
HoustonBoston: yes it does!

pkg/importer/file.go
akalenyu: right

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -730,11 +729,7 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
 
 	// Callback for libnbd.BlockStatus. Needs to modify blocks list above.
 	updateBlocksCallback := func(metacontext string, nbdOffset uint64, extents []uint32, err *int) int {
-		if nbdOffset > math.MaxInt64 {
-			klog.Errorf("Block status offset too big for conversion: 0x%x", nbdOffset)
-			return -2
-		}
```
akalenyu, line 736: I think you need this guard later on when converting back (copyrange)

pkg/importer/file.go
akalenyu: the current revision looks ok to me

pkg/importer/vddk-datasource_amd64.go
HoustonBoston: Added it back at the end of the function

akalenyu: any reason to not check it at the top? before storing?

akalenyu (commented):
/approve

thank you!

dsanatar (commented):
I don't really understand the value of removing the G115 linter exception from the global rule file if we are just going to add the exception back at each impacted line.

I did some research and apparently for older versions of the `golangci-lint` package (like the one we are using in cdi) is more strict and produces more false positives for type conversions. I think the linter engine expects these type conversions to occur within the conditional that asserts that the conversion will be safe. I would suggest trying that to see if you can then remove the nolint comments. alternatively, perhaps a package bump could also resolve some of these

cmd/cdi-controller/controller.go
```diff
@@ -372,7 +373,11 @@ func main() {
 		klog.Fatalf("Unable to get environment variables: %v\n", errors.WithStack(err))
 	}
 
-	logf.SetLogger(zap.New(zap.Level(zapcore.Level(-1*verbosityLevel)), zap.UseDevMode(debug)))
+	var level int8
+	if clampedLevel := min(verbosityLevel, math.MaxInt8); clampedLevel >= 0 && clampedLevel <= math.MaxInt8 {
```
dsanatar: do we need to take the minimum here? it makes the second part of the conditional always true. seems like we can drop one of the two.

also if the the passed in verbosityLevel is negative, this check fails and we default to 0 which is different from the previously defined `defVerbose` which is 1

cmd/cdi-operator/operator.go
```diff
@@ -83,7 +84,12 @@ func main() {
 	// implementing the logr.Logger interface. This logger will
 	// be propagated through the whole operator, generating
 	// uniform and structured logs.
-	logf.SetLogger(zap.New(zap.Level(zapcore.Level(-1*verbosityLevel)), zap.UseDevMode(debug)))
+	// Note: zapcore.Level is an int8, so we need to convert the verbosityLevel (which has no reason to be an int) to an int8.
+	var level int8
+	if clampedLevel := min(verbosityLevel, math.MaxInt8); clampedLevel >= 0 && clampedLevel <= math.MaxInt8 {
+		level = int8(clampedLevel)
```
dsanatar: same comments apply from the controller block

pkg/importer/file.go
```diff
@@ -57,8 +58,21 @@ func GetAvailableSpace(path string) (int64, error) {
 	if err != nil {
 		return int64(-1), err
 	}
-	//nolint:unconvert
-	return int64(stat.Bavail) * int64(stat.Bsize), nil
+	if stat.Bavail > math.MaxInt64 {
+		return int64(-1), fmt.Errorf("available blocks count %d exceeds int64", stat.Bavail)
+	}
+	if stat.Bavail <= math.MaxInt64 {
```
dsanatar: does it still throw a linting error if we just changed this to an else block since it's the direct inverse of the previous check?

pkg/importer/file.go
```diff
@@ -57,8 +58,21 @@ func GetAvailableSpace(path string) (int64, error) {
...
+		bsize := int64(stat.Bsize) //nolint:unconvert // type varies by platform
+		if bsize <= 0 {
+			return 0, nil
+		}
+		if avail > math.MaxInt64/bsize {
+			return int64(-1), fmt.Errorf("available space overflow: %d * %d exceeds int64", avail, bsize)
+		}
+		return avail * bsize, nil
+	}
+	return int64(-1), fmt.Errorf("available blocks count %d exceeds int64", stat.Bavail)
```
dsanatar: this line would never be hit, if it's larger than the max int64 it would be caught in the prior conditional

pkg/importer/file.go
```diff
@@ -57,8 +58,21 @@ func GetAvailableSpace(path string) (int64, error) {
...
 		return int64(-1), err
 	}
-	//nolint:unconvert
-	return int64(stat.Bavail) * int64(stat.Bsize), nil
+	if stat.Bavail > math.MaxInt64 {
+		return int64(-1), fmt.Errorf("available blocks count %d exceeds int64", stat.Bavail)
+	}
+	if stat.Bavail <= math.MaxInt64 {
+		avail := int64(stat.Bavail)
+		bsize := int64(stat.Bsize) //nolint:unconvert // type varies by platform
```
dsanatar: seems like `Bsize` is already of type int64, do we need this?

pkg/importer/imageio-datasource.go
```diff
@@ -305,26 +306,42 @@ func (is *ImageioDataSource) StreamExtents(extentsReader *extentReader, fileName
...
-				klog.Infof("Initial zero method failed, trying AppendZeroWithWrite instead. Error was: %v", err)
-				zeroRange = AppendZeroWithWrite // If the initial choice fails, fall back to regular file writing
-				err = zeroRange(outFile, extent.Start, extent.Length)
-				if err != nil {
-					return errors.Wrap(err, "failed to zero range on destination")
+			if extent.Start > math.MaxInt64 || extent.Length > math.MaxInt64 {
+				return errors.New("extent offset or length exceeds maximum int64 size")
+			}
+			if extent.Start <= math.MaxInt64 {
+				if extent.Length <= math.MaxInt64 {
```
dsanatar: can this be an else?

pkg/importer/imageio-datasource.go
```diff
@@ -305,26 +306,42 @@ func (is *ImageioDataSource) StreamExtents(extentsReader *extentReader, fileName
...
+					start, length := int64(extent.Start), int64(extent.Length)
```
dsanatar: what is this for? aren't these fields already int64s?

pkg/importer/imageio-datasource.go
```diff
@@ -305,26 +306,42 @@ func (is *ImageioDataSource) StreamExtents(extentsReader *extentReader, fileName
...
+						zeroRange = AppendZeroWithWrite
+						err = zeroRange(outFile, start, length)
+						if err != nil {
+							return errors.Wrap(err, "failed to zero range on destination")
+						}
+					}
 				}
 			}
-			is.readers.progressReader.Current += uint64(extent.Length)
+			is.readers.progressReader.Current += extent.Length
```
dsanatar: if you put this in your conditional you can likely leave the conversation

also if the end goal is to guard against this uint64 conversation, shouldn't the above checks reflect that?

pkg/importer/imageio-datasource.go
```diff
@@ -305,26 +306,42 @@ func (is *ImageioDataSource) StreamExtents(extentsReader *extentReader, fileName
...
+					responseBody, err := extentsReader.GetRange(int64(extent.Start), int64(endRange))
+					if err != nil {
+						return errors.Wrap(err, "failed to get range")
+					}
+					final := (index == (len(extentsReader.extents) - 1))
+					err = is.transferExtent(responseBody, outFile, extent, final)
+					if err != nil {
+						return errors.Wrap(err, "failed to transfer extent")
+					}
+				}
```
dsanatar: is this all even necessary? the fields used to be be int64s before your updates, is there a good reason they should now be uint64s to justify all these extra conditionals just to convert them back into int64s?

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -712,8 +712,8 @@ type NbdOperations interface {
 
 // BlockStatusData holds zero/hole status for one block of data
 type BlockStatusData struct {
-	Offset int64
-	Length int64
+	Offset uint64
+	Length uint64
 	Flags  uint32
```
dsanatar, line 717: also curious about this choice, seems related to my previous comment in imageio-datasource

dsanatar (changes requested):
just have some general questions about some changes you made to certain types in some struct fields but then are seemingly writing guards to allow for converting them back to their original type?

this should probably get another round of reviews from an approver before it can be merged since there have been a lot of new changes since the prior approval from @awels  
/hold

pkg/importer/file.go
HoustonBoston: Nope :smile:

pkg/importer/file.go
HoustonBoston: Questionable, but AI says it's different on each platform... maybe input from someone who's well-versed with this is required?

dsanatar: ah ok I see this is from a golang pkg which does seem to vary by platform, I thought this was a structure we defined.

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -750,43 +746,58 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
...
-				if lastFlags == flags && lastOffset == offset {
-					// Merge with previous block
-					blocks[last] = &BlockStatusData{
-						Offset: lastBlock.Offset,
-						Length: lastBlock.Length + length,
-						Flags:  lastFlags,
+			if offset > math.MaxInt64 {
+				klog.Errorf("Block status offset too big for conversion: 0x%x", offset)
+				return -2
+			}
```
dsanatar: is there a reason this check and conversion of offset was moved from the beginning of the function? I don't think we need this check for ever iteration of the loop since it won't change

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -750,43 +746,58 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
...
 				} else {
-					blocks = append(blocks, &BlockStatusData{Offset: offset, Length: length, Flags: flags})
+					blocks = append(blocks, &BlockStatusData{Offset: blockOffset, Length: length, Flags: flags})
 				}
-			} else {
-				blocks = append(blocks, &BlockStatusData{Offset: offset, Length: length, Flags: flags})
 			}
-			offset += length
+			if length >= 0 {
+				offset += uint64(length)
```
dsanatar: if u leave convert the offset value at the beginning, you can remove this check and conversion for `length`

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -750,43 +746,58 @@ func GetBlockStatus(handle NbdOperations, extent types.DiskChangeExtent) []*Bloc
...
+			}
 		}
 		return 0
 	}
 
 	if extent.Length < 1024*1024 {
+		if extent.Start < 0 || extent.Length < 0 {
+			return nil
+		}
```
dsanatar: do we need this guard if we aren't doing any type conversion?

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -1353,25 +1385,31 @@ func (vs *VDDKDataSource) TransferFile(fileName string, preallocation bool) (Pro
...
-		blocksize := uint64(MaxBlockStatusLength)
-		for i := start; i < vs.Size; i += blocksize {
-			if (vs.Size - i) < blocksize {
-				blocksize = vs.Size - i
-			}
+		diskSize := vs.Size
+		if diskSize > math.MaxInt64 {
+			return ProcessingPhaseError, fmt.Errorf("disk size %d exceeds int64", diskSize)
+		}
+		if diskSize <= math.MaxInt64 {
```
dsanatar: change to else?

dsanatar (commented):
looks a lot cleaner, just some extra things I noticed

pkg/importer/vddk-datasource_amd64.go
HoustonBoston: linter says else can be dropped too cause the if above contains a return statement

pkg/importer/imageio-datasource.go
```diff
@@ -308,17 +309,19 @@ func (is *ImageioDataSource) StreamExtents(extentsReader *extentReader, fileName
...
 				klog.Infof("Initial zero method failed, trying AppendZeroWithWrite instead. Error was: %v", err)
-				zeroRange = AppendZeroWithWrite // If the initial choice fails, fall back to regular file writing
+				zeroRange = AppendZeroWithWrite
 				err = zeroRange(outFile, extent.Start, extent.Length)
 				if err != nil {
 					return errors.Wrap(err, "failed to zero range on destination")
 				}
 			}
-			is.readers.progressReader.Current += uint64(extent.Length)
+			if extent.Length >= 0 {
```
dsanatar: it may be a good idea to at least log a warning or return an error if length is negative, otherwise we would be silently skipping this update.

pkg/importer/imageio-datasource.go
```diff
@@ -348,6 +351,9 @@ func (is *ImageioDataSource) transferExtent(source io.ReadCloser, dest io.Writer
 	if err != nil {
 		return errors.Wrap(err, "failed to write to file")
 	}
+	if written < 0 {
```
dsanatar: is this needed?

pkg/importer/imageio-datasource.go
```diff
@@ -584,18 +590,24 @@ func createImageioReader(ctx context.Context, ep string, accessKey string, secKe
 		total = 0
 		nonzero := int64(0)
 		for _, extent := range extents {
-			total += uint64(extent.Length)
+			if extent.Length >= 0 {
```
dsanatar: same here, we probably need to do more than just skip over the operation if we can't do the conversion

pkg/importer/vddk-datasource_amd64.go
```diff
@@ -1177,7 +1194,9 @@ func (vs *VDDKDataSource) TransferFile(fileName string, preallocation bool) (Pro
 	initialProgressTime := time.Now()
 	updateProgress := func(written int) {
 		// Only log progress at approximately 1% minimum intervals.
-		currentProgressBytes += uint64(written)
+		if written > 0 {
```
dsanatar: same here

pkg/util/util.go
```diff
@@ -83,7 +83,9 @@ func ParseEnvVar(envVarName string, decode bool) (string, error) {
 // Read reads bytes from the stream and updates the prometheus clone_progress metric according to the progress.
 func (r *CountingReader) Read(p []byte) (n int, err error) {
 	n, err = r.Reader.Read(p)
-	r.Current += uint64(n)
+	if n > 0 { // guarantee it's not negative
```
dsanatar: same here, maybe we should error if negative

PR #4194, opened 2026-07-03
fix: set default storageclass for external provider

Davo911:
When using external provider we should ensure the specified sc is annotated as the default before tests run. 
This mirrors the [behavior of the kubevirtci provider](https://github.com/kubevirt/containerized-data-importer/blob/4bd6ba382ff50e2ddf1ddf4b271f222c5a5a492d/cluster-sync/kubevirtci/provider.sh#L13-L38)

On shared external clusters, tests like `monitoring_test.go` and `datavolume_test.go` temporarily clear the
default SC annotation.

Fixes #

/cc: @akalenyu 

**Release note**:

```release-note
NONE
```


akalenyu (commented):
Thanks for tackling this!
I think messing with external cluster is a slippery slope. Folks don't expect environment changes due to the test run. Maybe we should just document/enforce the default requirement and hold the executor accountable?

For example, the job for s390 could set the default storage class

Davo911:
I considered to place this into the job directly, but felt this to be a cleaner approach here, as the executor explicitly needs to set `BLOCK_SC` and HPP is also configured here.

akalenyu:
> I considered to place this into the job directly, but felt this to be a cleaner approach here, as the executor explicitly needs to set `BLOCK_SC` and HPP is also configured here.

We can set it as part of a new job script if you prefer that (automation/s390x.sh) but IMO we can't be changing state on external kubevirtci providers

Davo911:
understood, will move it into the job, thanks!

PR #4195, opened 2026-07-03
prlimit: don't log a kill error when the command succeeds

arpitjain099:

Every successful command run through `ExecWithLimits` (qemu-img, nbdkit, and friends) logs a misleading `failed to kill the process; os: process already finished` line at ERROR level. It's noise that makes a healthy import look like something went wrong when you're reading the logs.

The cause is the deferred cleanup in `executeWithLimits`: it calls `cmd.Process.Kill()` right after `Start`, but on the normal path `cmd.Wait()` has already reaped the process by the time the defer runs, so `Kill()` returns `os.ErrProcessDone` and that got logged unconditionally. This just guards the log so it only fires when the process is genuinely still around and Kill actually fails, and drops the needless reassignment of the outer `err`.

I added a regression test that runs a successful `ExecWithLimits` and asserts no "failed to kill the process" line ends up in the klog output. It fails on the current code and passes with the fix.

The one pre-existing "killed by memory limit" spec fails the same way on `main` in my container (exit status vs signal under the memory cgroup), so it's unrelated to this change. The rest of the `pkg/system` suite plus the new test pass, and `go vet ./pkg/system/` is clean.

**Release note**:
```release-note
Stop logging a spurious "failed to kill the process" error after every successful command run in the importer/uploader
```

### DIFF
--- a/pkg/system/BUILD.bazel
+++ b/pkg/system/BUILD.bazel
@@ -23,5 +23,6 @@ go_test(
         "//vendor/github.com/onsi/ginkgo/v2:go_default_library",
         "//vendor/github.com/onsi/gomega:go_default_library",
         "//vendor/github.com/pkg/errors:go_default_library",
+        "//vendor/k8s.io/klog/v2:go_default_library",
     ],
 )
--- a/pkg/system/prlimit.go
+++ b/pkg/system/prlimit.go
@@ -20,6 +20,8 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	stderrors "errors"
+	"os"
 	"os/exec"
 	"syscall"
 	"time"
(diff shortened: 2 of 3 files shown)


akalenyu:
thank you!
/test all

arpitjain099:
Regenerated pkg/system/BUILD.bazel: the test now imports k8s.io/klog/v2, so the go_test target needed the matching dep. That was the deps-verify / generate-verify drift.

akalenyu (commented):
/approve
/test all

arpitjain099:
Closing this. #4200 removed `pkg/system` and moved qemu execution into `pkg/image`, so `prlimit.go` and the kill-error path this touched no longer exist on main. Nothing left to rebase onto.

Thanks for the approve, @akalenyu, sorry it went stale before it landed. If the same spurious kill error turns up in the new `pkg/image` code I will open a fresh one against that.

PR #4200, opened 2026-07-15
Removing pkg/system and prlimit(RLIMIT_AS) enforcement

### LINKED ISSUE #4186: Remove prlimit (RLIMIT_AS) address space limiting from qemu-img execution

davmarro:
The `pkg/system` package implements `prlimit(RLIMIT_AS)` to limit virtual address space for qemu-img operations. This is incompatible with Go 1.21+ runtime which relies on `mmap` for heap growth and goroutine stacks — causing `SIGSEGV` crashes, particularly on s390x.
Additionally, `RLIMIT_AS` is redundant in Kubernetes where cgroups already enforce memory limits via `resources.limits.memory`.
So this PR removes the entire `pkg/system` package and replaces `ExecWithLimits` with a lightweight `qemuCmd` struct in `pkg/image/exec.go` that uses `exec.CommandContext` directly, preserving streaming progress reporting for `qemu-img -p` operations.

Fixes #4186

- The `QEMUOperations` public interface is unchanged — this is an internal refactor only
- The `addPreallocation` retry logic now uses `errors.As` on `cmdExecError` instead of string-matching on stderr bytes

**Release note**:

```release-note
Removed prlimit(RLIMIT_AS) enforcement from qemu-img operations.
```


pkg/image/qemu.go
```diff
@@ -378,19 +372,23 @@ func addPreallocation(args []string, preallocationMethods [][]string, qemuFn fun
 	return err
 }
 
+func isUnsupportedPreallocation(err error) bool {
+	var execErr *cmdExecError
+	if errors.As(err, &execErr) {
+		return strings.Contains(execErr.stderr, "Unsupported preallocation mode")
+	}
```
akalenyu, line 377: is this ever not an exec err?

pkg/image/qemu.go
```diff
@@ -329,7 +325,7 @@ func execPreallocationBlock(dest string, bs, count, offset int64) error {
 		oflag += ",direct"
 	}
 	args := []string{"if=/dev/zero", "of=" + dest, fmt.Sprintf("bs=%d", bs), fmt.Sprintf("count=%d", count), fmt.Sprintf("seek=%d", offset), oflag}
-	_, err = qemuExecFunction(nil, nil, "dd", args...)
+	err = qemu.ExecRaw("dd", args...)
```
akalenyu: guess it was never truly a `qemuExec`, rather a generic command executor

pkg/image/qemu.go
```diff
@@ -219,9 +215,9 @@ func (o *qemuOperations) Info(url *url.URL) (*ImgInfo, error) {
 	if len(url.Scheme) > 0 && url.Scheme != "nbd+unix" && url.Scheme != "file" {
 		return nil, fmt.Errorf("not valid schema %s", url.Scheme)
 	}
-	output, err := qemuExecFunction(qemuInfoLimits, nil, "qemu-img", "info", "--output=json", url.String())
+	output, err := qemu.Info(qemuInfoTimeout, "info", "--output=json", url.String())
```
akalenyu: kind of weird that one has to pass `info` to `Info()`, no? maybe just accept the info args like `--output=json`

akalenyu (commented):
looks good!

pkg/image/qemu.go
davmarro: The `addPreallocation` method accept as parameter a `qemuFn func(args []string) error` that, currently, is the `Exec` method returning always an Exec error. 
Anyway it doesn't mean that in the future the method can accept are type of functions returning other errors. It's just a fallback for future usage of the method.

pkg/image/qemu.go
davmarro: yes, the "problem" is that the exec.go file is inside the qemu package so to differentiate the execution of a standard unix command from a qemu one I just added this ExecRaw method. do you think we need a different approach to execute it ?

pkg/image/qemu.go
davmarro: totally Right, it's a no-sense duplication. Good catch!

davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/f80fb6546c559c7559f2ca9c7588591942843ff4

pkg/image/qemu.go
akalenyu: no that's not a big deal

pkg/image/qemu.go
```diff
@@ -219,9 +215,9 @@ func (o *qemuOperations) Info(url *url.URL) (*ImgInfo, error) {
 	if len(url.Scheme) > 0 && url.Scheme != "nbd+unix" && url.Scheme != "file" {
 		return nil, fmt.Errorf("not valid schema %s", url.Scheme)
 	}
-	output, err := qemuExecFunction(qemuInfoLimits, nil, "qemu-img", "info", "--output=json", url.String())
+	output, err := qemu.Info(qemuInfoTimeout, "--output=json", url.String())
```
akalenyu: do we ever not do --output=json?

pkg/image/exec.go
```diff
@@ -0,0 +1,161 @@
...
+func (q *qemuCmd) Exec(args ...string) error {
+	_, err := q.run(context.Background(), qemuImg, args...)
+	return err
+}
+
+// Info runs qemu-img info with a timeout and returns stdout (for JSON parsing)
+func (q *qemuCmd) Info(timeout time.Duration, args ...string) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	return q.run(ctx, qemuImg, append([]string{"info"}, args...)...)
```
akalenyu: not sure you need `args...` for Info() - path/url always required

akalenyu:
unrelated, log probably truncated

pkg/image/exec.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := c.Start(); err != nil {
+		return fmt.Errorf("start %s: %w", name, err)
+	}
+
+	var errBuf bytes.Buffer
+	var scanErr error
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go func() {
```
akalenyu: can we do something nicer for the repeating parts of the goroutines and the magic number 2?

akalenyu: maybe we can let exec do the work for us, something like
```go
func (q *qemuCmd) defaultRunWithStream(ctx context.Context, callback func(string), name string, args ...string) error {
    c := exec.CommandContext(ctx, name, args...)

    var errBuf bytes.Buffer
    c.Stderr = &errBuf

    stdout, err := c.StdoutPipe()
    if err != nil {
        return fmt.Errorf("stdout pipe for %s: %w", name, err)
    }

    if err := c.Start(); err != nil {
        return fmt.Errorf("start %s: %w", name, err)
    }

    scanner := bufio.NewScanner(stdout)
    scanner.Split(scanLinesWithCR)
    for scanner.Scan() {
        if callback != nil {
            callback(scanner.Text())
        }
    }

    if err := c.Wait(); err != nil {
        klog.Errorf("%s failed, stderr:\n%s", name, errBuf.String())
        return &cmdExecError{name: name, stderr: errBuf.String(), err: err}
    }
    return nil
}
```

though I am not sure what the `\r` stuff was about in qemu-img, probably just stdout progress, stderr should be fine

pkg/image/qemu.go
davmarro: no, you're right. I should have passed it as parameter to the method, like "info". Thanks!

pkg/image/exec.go
davmarro: yes, I just maintained the same approach as was before with the args but it's not too much readable - better to pass explicitly the parameters. Thanks

pkg/image/exec.go
davmarro: oh I didn't realize that `exec.CommandContext` it's basically already handling the stdErr in the same way I proposed. 
At this point, this is the right approach. 
Yes the \r stuff is to process the execution output row by row (and not all together at the end of the execution).

pkg/image/qemu.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/192eb4d610570828a8af50c6c14d732b4f459933

pkg/image/exec.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/192eb4d610570828a8af50c6c14d732b4f459933

pkg/image/exec.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/192eb4d610570828a8af50c6c14d732b4f459933

pkg/image/exec.go
```diff
@@ -70,10 +70,10 @@ func (q *qemuCmd) Exec(args ...string) error {
 }
 
 // Info runs qemu-img info with a timeout and returns stdout (for JSON parsing)
-func (q *qemuCmd) Info(timeout time.Duration, args ...string) ([]byte, error) {
+func (q *qemuCmd) Info(timeout time.Duration, url *url.URL) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
-	return q.run(ctx, qemuImg, append([]string{"info"}, args...)...)
+	return q.run(ctx, qemuImg, append([]string{"info", "--output=json"}, url.String())...)
```
akalenyu: i think we don't need to build a slice here and spread it

davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/8d4cf4d8b78c42ff4dee1ac360e6eeacd79c0c28

akalenyu:
/approve
great stuff, thank you!
/cc @Acedus

pkg/image/qemu.go
```diff
@@ -74,10 +71,9 @@ type qemuOperations struct{}
 var (
 	ErrLargerPVCRequired = errors.New("A larger PVC is required")
 
-	qemuExecFunction = system.ExecWithLimits
-	qemuInfoLimits   = &system.ProcessLimitValues{AddressSpaceLimit: maxMemory, CPUTimeLimit: maxCPUSecs}
-	qemuIterface     = NewQEMUOperations()
-	re               = regexp.MustCompile(matcherString)
+	qemuIterface = NewQEMUOperations()
```
Acedus: s/Iterface/Interface

pkg/image/qemu.go
```diff
@@ -35,14 +35,11 @@ import (
...
 	metrics "kubevirt.io/containerized-data-importer/pkg/monitoring/metrics/cdi-importer"
-	"kubevirt.io/containerized-data-importer/pkg/system"
 	"kubevirt.io/containerized-data-importer/pkg/util"
 )
 
 const (
-	networkTimeoutSecs = 3600    //max is 10000
-	maxMemory          = 1 << 30 //value from OpenStack Nova
-	maxCPUSecs         = 30      //value from OpenStack Nova
+	networkTimeoutSecs = 3600 //max is 10000
```
Acedus: This const isn't used anywhere.

pkg/image/qemu.go
Acedus: I know it's a bit out-of-scope but this is also unused.

pkg/image/qemu.go
```diff
@@ -378,19 +372,23 @@ func addPreallocation(args []string, preallocationMethods [][]string, qemuFn fun
 	return err
 }
 
+func isUnsupportedPreallocation(err error) bool {
+	var execErr *cmdExecError
+	if errors.As(err, &execErr) {
+		return strings.Contains(execErr.stderr, "Unsupported preallocation mode")
+	}
+	return strings.Contains(err.Error(), "Unsupported preallocation mode")
```
Acedus, line 378: `err.Error()` already contains the error message in either case, that if appears to be redundant.

Acedus (commented):
Thanks @davmarro, while the prlimit removal itself is fine, I'm not too happy with the pkg/image design as a whole. The current code declares package-level globals and tests hack around them by swapping mutable function fields on a singleton. The `QEMUOperations` interface exists, but the testing seam is in the wrong place... There's no abstraction boundary between business logic and `os/exec` so tests can't exercise one without faking the other. This is a textbook DI violation IMO. 

What I'd like to potentially see instead is `qemuOperations` as a concrete struct that accepts its executor as an injected interface (e.g., `QEMUCmd`). Tests could then stub one layer below, exercising the real business logic while faking only the command execution boundary. The same principle could then apply one level deeper; `QEMUCMd`'s concrete impl could accept accept a `CmdRunner` interface that wraps `exec.CommandContext/StdoutPipe` so those can be stubbed independently without touching the rest of the logic (e.g., the stream parsing we do). 

Let me know WDYT, we could defer it to be a follow-up, too.

EDIT: P.S., be sure to squash and logically arrange your commits!

davmarro:
> Thanks @davmarro, while the prlimit removal itself is fine, I'm not too happy with the pkg/image design as a whole. The current code declares package-level globals and tests hack around them by swapping mutable function fields on a singleton. The `QEMUOperations` interface exists, but the testing seam is in the wrong place... There's no abstraction boundary between business logic and `os/exec` so tests can't exercise one without faking the other. This is a textbook DI violation IMO.
> 
> What I'd like to potentially see instead is `qemuOperations` as a concrete struct that accepts its executor as an injected interface (e.g., `QEMUCmd`). Tests could then stub one layer below, exercising the real business logic while faking only the command execution boundary. The same principle could then apply one level deeper; `QEMUCMd`'s concrete impl could accept accept a `CmdRunner` interface that wraps `exec.CommandContext/StdoutPipe` so those can be stubbed independently without touching the rest of the logic (e.g., the stream parsing we do).
> 
> Let me know WDYT, we could defer it to be a follow-up, too.
> 
> EDIT: P.S., be sure to squash and logically arrange your commits!

Hi @Acedus thanks for your interesting comment!
The PR scope was about removing the `pkg/system` package, so I decided to focus on removing the useless files and move the remaining useful parts elsewhere.
I kept the same approach as before for the global var with the function fields (introducing a minor improvement: using the struct instead of multiple global vars) just to stay consistent with the previous approach and to avoid polluting the PR with an out of scope redesign.

Anyway, I totally agree with what you're saying about swapping mutable function fields over a singleton - I can fix it including a `*qemuCmd` as a field of the `qemuOperations` so that each test can create its own instance. 

Regarding your suggestion about the implementation, instead, I have a bit different opinion to be honest.
We're just implementing a CLI wrapper and introducing a 2 layer interface (QEMUCmd + Cmd Runner) seems to be just an over engineering for what is essentially a simple wrapper. 

What I'd prefer to do is instead just removing the global var and inject a `*qemuCmd` directly in the `qemuOperations`, leaving the function fields approach as it is. What do you think about this approach?

Acedus:
> Regarding your suggestion about the implementation, instead, I have a bit different opinion to be honest.
We're just implementing a CLI wrapper and introducing a 2 layer interface (QEMUCmd + Cmd Runner) seems to be just an over engineering for what is essentially a simple wrapper.

> What I'd prefer to do is instead just removing the global var and inject a *qemuCmd directly in the qemuOperations, leaving the function fields approach as it is. What do you think about this approach?

I don't really think it's over-engineering it TBH, just makes `qemuCmd` more testable. But there are, of course, other ways to go about it. Anyway, let's start with the removal of those global vars.

pkg/image/qemu.go
davmarro: The if isn't redundant, it's an improvement on the previous implementation. The old code relied on `err.Error()` string matching, which is fragile: if we ever change the `Error() `format, the check would break silently.
The `errors.As` branch is the primary path:it reads the raw stderr field directly, making it independent of how the error message is formatted. The fallback at line 380 handles the edge case where the error isn't a `*cmdExecError`, since `addPreallocation` accepts a generic `qemuFn func(args []string) error` and we can't guarantee the error type.

We could simplify to a single strings.Contains(err.Error(), ...) but that would be the fragile approach I intentionally moved away from.
See also this [thread](https://github.com/kubevirt/containerized-data-importer/pull/4200#discussion_r3601511022) for more context.

pkg/image/qemu.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/6de49485411878987496306eea96f99c1c8070b4

pkg/image/qemu.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/6de49485411878987496306eea96f99c1c8070b4

pkg/image/qemu.go
davmarro: https://github.com/kubevirt/containerized-data-importer/pull/4200/commits/6de49485411878987496306eea96f99c1c8070b4

davmarro:
> > Regarding your suggestion about the implementation, instead, I have a bit different opinion to be honest.
> > We're just implementing a CLI wrapper and introducing a 2 layer interface (QEMUCmd + Cmd Runner) seems to be just an over engineering for what is essentially a simple wrapper.
> 
> > What I'd prefer to do is instead just removing the global var and inject a *qemuCmd directly in the qemuOperations, leaving the function fields approach as it is. What do you think about this approach?
> 
> I don't really think it's over-engineering it TBH, just makes `qemuCmd` more testable. But there are, of course, other ways to go about it. Anyway, let's start with the removal of those global vars.

Ok @Acedus, how do we want to track this tech debt ? a new issue ?
Global vars removed and commits squashed.

Acedus:
> Ok @Acedus, how do we want to track this tech debt ? a new issue ?
Sure.

Acedus (approved):
Thanks @davmarro!

akalenyu:
> > Ok @Acedus, how do we want to track this tech debt ? a new issue ?
> > Sure.



> Thanks @davmarro!

:pray:

So just to summarize I think we can come up with something better than the overrideable member funcs,
and that would go in a follow up PR after this merges (which should be soon)

PR #4201, opened 2026-07-15
feat: transition to using the ImageVolume API for Registery imports with pull method node

nav1s:
As of k8s 1.36 [image-volumes](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes) are now stable. 
ImageVolumes enable the mounting of container images as volume without the need for an init container.
This API allows us to streamline the existing pull method node Registry import that currently leverages a similar init container based approach for that exact use case. 

Reference:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/dataimportcron-controller.go
+++ b/pkg/controller/dataimportcron-controller.go
@@ -707,29 +707,22 @@ func (r *DataImportCronReconciler) updateContainerImageDesiredDigest(ctx context
 			Affinity:                      workloadNodePlacement.Affinity,
 			Volumes: []corev1.Volume{
 				{
-					Name: "shared-volume",
+					Name: "image-volume",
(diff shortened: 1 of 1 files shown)


Acedus (commented):
So one of the main things I wanted to solve with pull method node transition is to remove the code path it has with the http-datasource due to how it's been written: https://github.com/kubevirt/containerized-data-importer/blob/d9cdc4395c8ffcd147bb96dd1359077465889273/pkg/importer/http-datasource.go#L159-L163.

Ideally `ImageVolume` should allow us to keep the entire flow exclusively to the registry-datasource handler by skipping the `CopyRegistryImage` part in the case of pull method node instead of pod, WDYT?

Acedus:
Also @nav1s please be sure to fill out the PR template correctly including a release note if there is any.

PR #4202, opened 2026-07-15
[release-v1.64] manifests: drop explicit command overrides from test templates

Acedus:
The bad-webserver, test-proxy, and sample-populator templates had hardcoded command paths matching the old `rules_docker` `go_image` conventions. These were always redundant since `go_image` sets the image ENTRYPOINT. Remove the command override so the pod uses the image ENTRYPOINT instead.

Fixes #

See https://github.com/bazelbuild/rules_docker/blob/0e9c3b068d05f20adf7ccdea486fcb27e71593f3/lang/image.bzl#L206-L217.

**Release note**:

```release-note
NONE
```


Acedus:
/cherry-pick release-v1.63 release-v1.62 release-v1.61 release-v1.60 release-v1.59

PR #4214, opened 2026-07-26
VEP 395: Support direct OCI layer fetch with LayerDigest

Acedus:
This PR implements the CDI portion of https://github.com/kubevirt/enhancements/issues/395. It adds a `LayerDigest` field to `DataVolumeSourceRegistry` that enables fetching a single OCI blob directly by digest, bypassing manifest layer parsing and iteration. It also adds support for importing OCI artifacts with a raw+compression media type (e.g., as in the case of KubeVirt's raw+zstd) where blobs are stored as raw compressed blobs rather than tar layers.

Fixes #
VEP tracking issue: https://github.com/kubevirt/enhancements/issues/395

**Release note**:

```release-note
Added LayerDigest option for Registry DataSource to allow importing disk images from specific OCI image layers by layer digest.
```


pkg/apiserver/webhooks/util.go
```diff
@@ -158,6 +158,15 @@ func validateDataVolumeSourceRegistry(sourceRegistry *cdiv1.DataVolumeSourceRegi
 		return causes
 	}
 
+	if sourceRegistry.LayerDigest != nil && importMethod != nil && *importMethod == cdiv1.RegistryPullNode {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "LayerDigest is not supported with node pull import method",
```
0xFelix: I wonder whether the message shouldn't be the opposite? Node pull is not supported with `LayerDigest`?

pkg/importer/transport_test.go
```diff
@@ -77,8 +80,21 @@ var _ = Describe("Registry Importer", func() {
...
+		info, err := CopyRegistryImage(kubevirtOCISource, tmpDir, "disk", "", "", "", kubevirtOCISourceLayerDigest, "", false, false)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(info).To(BeNil())
+
+		file := filepath.Join(tmpDir, "disk", "disk.img")
+		Expect(file).To(BeARegularFile())
+
+		fi, err := os.Stat(file)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(fi.Size()).To(BeNumerically(">", 0))
```
0xFelix: Would these tests benefit from comparing checksums of the extracted images?

pkg/importer/transport.go
```diff
@@ -178,10 +177,39 @@ func processLayer(ctx context.Context,
...
+		return fmt.Errorf("%w: %v", errReadingLayer, err)
+	}
+	fr, err := NewFormatReaders(reader, 0, nil)
+	if err != nil {
+		klog.Errorf("%v: %v", errReadingLayer, err)
+		return fmt.Errorf("%w: %v", errReadingLayer, err)
+	}
+	defer fr.Close()
+
+	klog.Infof("Layer detected as raw disk image, streaming to disk")
```
0xFelix, line 202: Does the reader only detect that it is zstd, or does it also detect that it is a raw image?

pkg/importer/transport.go
```diff
@@ -253,10 +281,26 @@ func copyRegistryImage(url, destDir, pathPrefix, accessKey, secKey, imageArchite
...
+func copyRegistryLayer(ctx context.Context, src types.ImageSource, cache types.BlobInfoCache, layerDigest digest.Digest, destDir, pathPrefix string, stopAtFirst, preallocation bool) (*types.ImageInspectInfo, error) {
+	layer := types.BlobInfo{Digest: layerDigest}
+
+	found, err := processLayer(ctx, src, layer, destDir, pathPrefix, cache, stopAtFirst, preallocation)
+	if err != nil && !errors.Is(err, errReadingLayer) {
+		return nil, err
+	}
+
+	if !found {
+		if err := streamRawLayer(ctx, src, layer, destDir, pathPrefix, cache, preallocation); err != nil {
```
0xFelix, line 296: In case of streaming a raw layer, can we make it detect and validate the `mediaType` of the layer?

pkg/apiserver/webhooks/util.go
0xFelix, line 1: Can we add validation to this field to check if it's well formed? I.e. `sha256:a1b2c3d4....`

pkg/importer/transport.go
Acedus: It detects zstd compression and if it finds no additional headers (e.g., tar) then it treats it as a raw image.

pkg/importer/transport.go
Acedus: That would require us to parse the manifest, what would be the merit of doing that, though? Also, how would you decide what are the supported media types?

0xFelix: Supported media types could be the ones defined by KubeVirt. But I agree that parsing the manifest poses a challenge. Right now it works because the type can be auto-detected, but I was thinking checking the `mediaType` could help with handling of the blob and to ensure compatibility with it.

PR #4218, opened 2026-07-29
fix: advise csi-clone strategy for Portworx pure_block/pure_fa_file storage classes

### LINKED ISSUE #4217: pxd.portworx.com driver with backend set as  pure_block in storage class as a parameter should use clone strategy in storage profile
**What happened**:
pxd.portworx.com driver with backend parameter set to pure_block in storage class should is using snapshot strategy.

pxd.portworx.com driver with  backend parameter set to pure_block in storage class should use clone strategy. 

1. Create a storage class with storage class parameter backend set to pure_block
2. Create root PVC using DataVolumeTemplate from golden PVC using the above storage class
3. The CDI process will use snapshot and restore mechanism to create the VMs rather than clone strategy.

**Additional context**:

**Environment**:
- DV specification: N/A
- Install tools: N/A
- Others: N/A

pujitha24:
Motivation:
For a pxd.portworx.com StorageClass with parameter backend=pure_block
(or backend=pure_fa_file), the StorageProfile controller advises the
"snapshot" clone strategy instead of "csi-clone" when populating a
DataVolume from a golden PVC. This means CDI takes the less efficient
snapshot-and-restore path instead of using CSI clone, even though
Portworx supports CSI clone for these backends just like it does for
the base pxd.portworx.com and pxd.portworx.com/nfs provisioner keys.
User-visible behavior for the clone itself is otherwise unaffected
(the clone still succeeds), the only change is which cloning mechanism
is chosen.

Approach:
storageClassToProvisionerKeyMapper["pxd.portworx.com"] derives the
provisioner keys "pxd.portworx.com/pure_block" and
"pxd.portworx.com/pure_fa_file" from the StorageClass's "backend"
parameter, and CapabilitiesByProvisionerKey already has entries for
both. However CloneStrategyByProvisionerKey was missing entries for
these two derived keys, so GetAdvisedCloneStrategy() found no known
strategy and the controller fell back to its generic default
(snapshot if a VolumeSnapshotClass exists, otherwise host-assisted).
Add the two missing entries, mirroring the existing sibling entries
"pxd.portworx.com" and "pxd.portworx.com/nfs" which are both already
CloneStrategyCsiClone.

Validation:
Added a table test reproducing the bug for both backend values, and
confirmed (by temporarily reverting the map change) that it fails
before the fix with "Expected snapshot to equal csi-clone" and passes
after.

  GOWORK=off go test ./pkg/controller/ -run TestController

Ran 367 of 367 Specs in ~2.2 seconds
SUCCESS! -- 367 Passed | 0 Failed | 0 Pending | 0 Skipped

Fixes #4217

```release-note
Fixed StorageProfile controller advising the wrong clone strategy for Portworx StorageClasses with backend=pure_block or backend=pure_fa_file parameters; they now correctly advise csi-clone.
```

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/controller/storageprofile-controller_test.go
+++ b/pkg/controller/storageprofile-controller_test.go
@@ -584,6 +584,33 @@ var _ = Describe("Storage profile controller reconcile loop", func() {
 		Entry("provisioner that is known to prefer csi clone", "csi-powermax.dellemc.com", cdiv1.CloneStrategyCsiClone, false),
 	)
 
+	DescribeTable("should set cloneStrategy for pxd.portworx.com according to backend parameter", func(scParameters map[string]string, expectedCloneStrategy cdiv1.CDICloneStrategy) {
+		provisioner := "pxd.portworx.com"
+		storageClass := CreateStorageClassWithProvisioner(storageClassName, map[string]string{AnnDefaultStorageClass: "true"}, map[string]string{}, provisioner)
+		storageClass.Parameters = scParameters
+		reconciler = createStorageProfileReconciler(storageClass, createVolumeSnapshotContentCrd(), createVolumeSnapshotClassCrd(), createVolumeSnapshotCrd())
+		snapClass := createSnapshotClass(snapshotClassName, nil, provisioner)
+		err := reconciler.client.Create(context.TODO(), snapClass)
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: storageClassName}})
+		Expect(err).ToNot(HaveOccurred())
+
+		storageProfileList := &cdiv1.StorageProfileList{}
+		err = reconciler.client.List(context.TODO(), storageProfileList, &client.ListOptions{})
+
+		Expect(err).ToNot(HaveOccurred())
+		Expect(storageProfileList.Items).To(HaveLen(1))
+
+		sp := storageProfileList.Items[0]
+		Expect(*sp.Status.StorageClass).To(Equal(storageClassName))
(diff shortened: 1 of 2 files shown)


pujitha24:
The SonarCloud Quality Gate was failing because the new `"pxd.portworx.com/pure_block"` and `"pxd.portworx.com/pure_fa_file"` string literals ended up duplicated 3 times each in `storagecapabilities.go` (rule go:S1192). I extracted them into `ProvisionerPortworxPureBlock` and `ProvisionerPortworxPureFAFile` constants, following the existing `ProvisionerHPP*` convention in the same file, and used them at all three call sites. `pkg/storagecapabilities` builds cleanly (it has no test files) and the `pkg/controller` tests still pass (`GOWORK=off go test ./pkg/storagecapabilities/... ./pkg/controller/...`).

No human review comments have landed yet, so I haven't touched anything else.

akalenyu:
thanks!

akalenyu:
@pujitha24 I think there's some forbidden chars in the commit message, hence that label `do-not-merge/invalid-commit-message`:
https://github.com/kubevirt/containerized-data-importer/pull/4218#issuecomment-5121993450

Acedus (approved):

pujitha24:
Thanks for pointing that out — you're right, it was the `Fixes #4217` line in the first commit's message. Prow's invalid-commit-message check rejects closing keywords (fix/fixes/close/etc.) immediately followed by `#<issue>`, and that line matched. I've reworded it to "Addresses issue 4217 (see PR description for the closing reference)" so there's no `#` left in any commit message — the code diff itself is unchanged. Will need to force-push to update the branch.

akalenyu:
readding @Acedus' lgtm
/lgtm

PR #4220, opened 2026-07-30
Delete ConfigMaps while also Providing them with Specific Labels for Downstream Testing

HoustonBoston:
Specific ConfigMaps `['imageio-certs', 'cdi-test-proxy-certs', 'cdi-file-host-certs', 'cdi-docker-registry-host-certs']` are causing Tier2 tests to fail downstream because they do not have these three labels:
- `app.kubernetes.io/managed-by`
- `app.kubernetes.io/component`
- `app.kubernetes.io/part-of`
so the fix was to keep the ConfigMaps with those labels so at least TIer 2 tests pass while they exist in the cluster, and then Kubernetes garbage collects the ConfigMaps when the relevant pods consuming those ConfigMaps are deleted through the `OwnerReference` field.

Fixes # 
None

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils/certs.go
+++ b/tests/utils/certs.go
@@ -14,7 +14,42 @@ import (
 	"k8s.io/client-go/rest"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
+
+	"kubevirt.io/containerized-data-importer/pkg/common"
 )
+
+// helper function to add labels to config map
+func addCdiTestLabels(existing map[string]string) map[string]string {
+	if existing == nil {
+		existing = make(map[string]string)
+	}
+	existing[common.AppKubernetesManagedByLabel] = "cdi-testing"
+	existing[common.AppKubernetesComponentLabel] = "storage"
+	existing[common.AppKubernetesPartOfLabel] = "testing"
+	return existing
+}
+
+// helper function to get owning pod reference
+func getOwningPodRef(client kubernetes.Interface, namespace string) (*metav1.OwnerReference, error) {
+	podName := os.Getenv("HOSTNAME")
+	if podName == "" {
+		return nil, errors.New("HOSTNAME env var not set, cannot determine pod name")
+	}
+
+	pod, err := client.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error getting pod %s", podName)
+	}
+
+	isController := true
+	return &metav1.OwnerReference{
+		APIVersion: "v1",
+		Kind:       "Pod",
+		Name:       pod.Name,
+		UID:        pod.UID,
+		Controller: &isController,
+	}, nil
+}
 
 // CreateCertForTestService creates a TLS key/cert for a service, writes them to files
 // and creates a config map containing the cert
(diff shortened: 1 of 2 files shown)


awels (commented):
Going to be pedantic here.

Could you make a function something like this:
```go
func addCdiTestLabels(existing map[string]string) map[string]string {
   if existing == nil {
       existing = make(map[string]string
   }
   existing[common.AppKubernetesManagedByLabel] = "cdi-testing"
   existing[common.AppKubernetesComponentLabel] = "storage"
   existing[common.AppKubernetesPartOfLabel] = "testing"
   return existing
}
```
And then just call this function from both locations? I was going to ask for consts for the strings, but we have these all over the place like that.

HoustonBoston:
> Going to be pedantic here.
> 
> Could you make a function something like this:
> 
> ```go
> func addCdiTestLabels(existing map[string]string) map[string]string {
>    if existing == nil {
>        existing = make(map[string]string
>    }
>    existing[common.AppKubernetesManagedByLabel] = "cdi-testing"
>    existing[common.AppKubernetesComponentLabel] = "storage"
>    existing[common.AppKubernetesPartOfLabel] = "testing"
>    return existing
> }
> ```
> 
> And then just call this function from both locations? I was going to ask for consts for the strings, but we have these all over the place like that.

Done 🫡

awels (commented):
/approve

akalenyu (commented):
Thank you for tackling this!

I think the real issue is that we are not cleaning up these test configmaps
(created by init containers in the test pods) https://github.com/kubevirt/containerized-data-importer/blob/806cb718a3ff30c5fe5db6b026809e275b208c9f/tools/imageio-init/main.go#L38 There's other issues with leaving orphan configmaps in the install namespace, so rather than trying to make those test-only resources adhere to k8s relationship labels, I would just ensure they are cleaned up when their respective workloads are removed. We could do so via OwnerReferences here, I believe https://github.com/kubevirt/containerized-data-importer/blob/806cb718a3ff30c5fe5db6b026809e275b208c9f/tests/utils/certs.go#L45-L47

tests/utils/certs.go
```diff
@@ -14,8 +14,43 @@ import (
...
+func getOwningPodRef(client kubernetes.Interface, namespace string) (*metav1.OwnerReference, error) {
+	podName := os.Getenv("HOSTNAME")
+	if podName == "" {
+		return nil, errors.New("HOSTNAME env var not set, cannot determine pod name")
+	}
+
+	pod, err := client.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error getting pod %s", podName)
+	}
```
akalenyu, line 42: doesn't this always error out due to rbac?

akalenyu: https://github.com/kubevirt/containerized-data-importer/blob/6f791e870f40d3e45028492f27ba2dff896bf35c/manifests/templates/cdi-testing-sa.yaml.in#L16-L23

HoustonBoston: So do I need to add in a rule to allow it to get pods?

HoustonBoston: I added in that rule and the behavior seems to be that the CM is deleted by k8s when the deployment `cdi-file-host` is deleted or has been scaled down to 0 replicas. Is that the behavior you'd like to see?

akalenyu: exactly, nice work!

akalenyu (approved):

akalenyu:
nfs server dies mid run :(

PR #4221, opened 2026-07-31
Tests: Allow storage class cleanup by adding OCS external label

alromeros:

On OpenShift clusters with ODF, storage classes created by the tests are recreated by the operator during teardown because they are still referenced by the storage consumer. As a result, the test cleanup cannot delete them.

Adding the `storageclass.ocs.openshift.io/is-external` label should allow the teardown to clean them up successfully.

This fix is recommended in https://redhat.atlassian.net/browse/DFBUGS-4795?focusedCommentId=16934614. It currently is the only solution offered for the sc deletion issue.  

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/datavolume_test.go
+++ b/tests/datavolume_test.go
@@ -1854,6 +1854,9 @@ var _ = Describe("[vendor:cnv-qe@redhat.com][level:component]DataVolume tests", 
 				},
 				Provisioner: "unknown-provisioner",
 			}
+
+			utils.AddOCSExternalLabel(f.K8sClient, sc)
+
 			sc, err := f.K8sClient.StorageV1().StorageClasses().Create(context.TODO(), sc, metav1.CreateOptions{})
 			Expect(err).ToNot(HaveOccurred())
 			tempScName = sc.Name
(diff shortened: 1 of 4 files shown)


akalenyu:
I recall this being a legit bug on the operator side, and we wouldn't be exposed as long as we don't use ownerreferences to some operator resource. Anyhow, I think OCS specific labels are kind of weird in our test suite, and I'd be surprised if the maintainers of OCS operator actually accepted such a bug as WONTFIX. This is a totally legitimate sequence of operations for an admin to take (permutation of storage class with specific param)

alromeros:
> I recall this being a legit bug on the operator side, and we wouldn't be exposed as long as we don't use ownerreferences to some operator resource. Anyhow, I think OCS specific labels are kind of weird in our test suite, and I'd be surprised if the maintainers of OCS operator actually accepted such a bug as WONTFIX. This is a totally legitimate sequence of operations for an admin to take (permutation of storage class with specific param)

That's what I assumed but the related jiras I'm aware of ended up being closed as wontfix, so I'm not expecting the behavior to change. I do agree that OCS-specific labels in our tests are not ideal. I'm trying to sync with the assignee of the other jiras to see if there's a better workaround or if there's proper documentation for this label.

alromeros:
> I recall this being a legit bug on the operator side, and we wouldn't be exposed as long as we don't use ownerreferences to some operator resource. Anyhow, I think OCS specific labels are kind of weird in our test suite, and I'd be surprised if the maintainers of OCS operator actually accepted such a bug as WONTFIX. This is a totally legitimate sequence of operations for an admin to take (permutation of storage class with specific param)

I just discussed this offline and was told that the sc deletion issue isn't planned to be addressed. Instead, they plan to document the `storageclass.ocs.openshift.io/is-external` label, but there's no ETA yet as they're currently focused on other work.

PR #4225, opened 2026-08-05
Avoid PVC writes during scheduler retry to prevent SchedulerError loop

mnecas:
The import controller watches both PVCs and owned pods. When the importer pod is pending and the scheduler hits a VolumeBinding error, each reconcile writes PVC annotations, bumping resourceVersion. This fires the PVC watch causing an immediate re-reconcile (bypassing the 2s RequeueAfter), which writes again, creating a tight loop at ~10 writes/sec.

The scheduler's VolumeBinding plugin assumes a PVC resourceVersion during filtering and expects it unchanged at reserve time. The constant writes make this race unwinnable, so the pod is never scheduled.

Fix: skip PVC writes and RequeueAfter while the pod has PodScheduled=False with a transient reason (e.g. SchedulerError). Pod/PVC watches resume reconciliation once scheduling succeeds.

Fixes #MTV-6199

**Release note**:

```release-note
BugFix: Fixed excessive PVC resource writes on retry
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -2325,9 +2325,11 @@ func UpdatePVCBoundConditionFromEvents(pvc *corev1.PersistentVolumeClaim, c clie
 	anno[AnnBoundConditionReason] = "Pending"
 	anno[AnnBoundConditionMessage] = boundMessage
 
-	patch := client.MergeFrom(currentPvcCopy)
-	if err := c.Patch(context.TODO(), pvc, patch); err != nil {
-		return err
+	if !reflect.DeepEqual(currentPvcCopy, pvc) {
+		patch := client.MergeFrom(currentPvcCopy)
+		if err := c.Patch(context.TODO(), pvc, patch); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/pkg/controller/import-controller.go
+++ b/pkg/controller/import-controller.go
@@ -309,6 +309,9 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 	}
 
 	if !cc.IsPVCComplete(pvc) {
+		if pod != nil && isPodWaitingForScheduler(pod) {
+			return reconcile.Result{}, nil
+		}
 		// We are not done yet, force a re-reconcile in 2 seconds to get an update.
 		log.V(1).Info("Force Reconcile pvc import not finished", "pvc.Name", pvc.Name)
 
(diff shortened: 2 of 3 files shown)


akalenyu:
Thank you! this sounds awfully similar to https://github.com/kubevirt/containerized-data-importer/pull/4176/changes/6891189fd3a2b85f8f2d3dc3e5fde2cbbe1bcdee - we used to hit it in CI for the non-populator path, which I recall Portworx was taking.

akalenyu:
Was the DV/PVC in your case opting out of populators? (ann `cdi.kubevirt.io/storage.usePopulator` set to `"false"`)

mnecas:
@akalenyu yes! It also had `cdi.kubevirt.io/storage.usePopulator: "false"`. 
I added one more small commit to to the `UpdatePVCBoundConditionFromEvents` so we don't update when not needed.

akalenyu:
> @akalenyu yes! It also had `cdi.kubevirt.io/storage.usePopulator: "false"`. I added one more small commit to to the `UpdatePVCBoundConditionFromEvents` so we don't update when not needed.

I suspect https://github.com/kubevirt/containerized-data-importer/commit/6891189fd3a2b85f8f2d3dc3e5fde2cbbe1bcdee fixes the issue, if you want to give that a try. It was exclusively an issue with the non populator path (which is a shame to see Portworx is still doing)

solenoci:
@akalenyu Hi, do you please know if https://github.com/kubevirt/containerized-data-importer/commit/6891189fd3a2b85f8f2d3dc3e5fde2cbbe1bcdee was delivered downstream?

mnecas:
@akalenyu i don't think it's the same issue as i don't see CreatingScratchSpace in the logs

akalenyu:
> @akalenyu Hi, do you please know if [6891189](https://github.com/kubevirt/containerized-data-importer/commit/6891189fd3a2b85f8f2d3dc3e5fde2cbbe1bcdee) was delivered downstream?

Nope, just 1.66. We can backport



> @akalenyu i don't think it's the same issue as i don't see CreatingScratchSpace in the logs

As discussed offline, you don't have to see this log - we started conditionally removing the annotation based on whether it's the scratch condition. So pre fix - always removed, initiating this annoying spam, post fix - not removed unless set by scratch.

mnecas:
Closing as the issue disappeared after https://github.com/kubevirt/containerized-data-importer/commit/6891189fd3a2b85f8f2d3dc3e5fde2cbbe1bcdee

PR #4236, opened 2026-08-07
Add CDIConfig.spec.TrustedCA to hold non-proxy CA certificates

### LINKED ISSUE #4152: Refactor cluster-wide CAs
Currently, if you wanted to define a cluster-wide standard CA that gets applied for all DataVolumes, you would need to modify the `CDIConfig.spec.importProxy.trustedCAProxy` field to point to a ConfigMap in the `cdi` namespace containing your CA. 

It is semantically wrong and slightly confusing that you must edit this proxy named field regardless of whether you are intending to add a proxy or non-proxy CA to your cluster. The import controller also always considers these certs from `trustedCAProxy` as proxy certs and when creating new importer pods, will copy the certs into the pod's `/proxycerts/` directory.

For OCP clusters users modify the `proxy.spec.trustedCA` in the `openshift-config` namespace to achieve this same outcome. CDI then uses this Proxy object to populate the `trustedCAProxy` in the CDIConfig. Alternatively, there is an optional label `config.openshift.io/inject-trusted-cabundle` that can be added to a configmap definition that will trigger the CA bundle to be injected into a global operator which would then bypass the need to copy the trusted cert across the various namespaces.

1. CDIConfig should be modified to have new standard _only_ `trustedCA` field outside of the `importProxy` struct. The import controller should be modified to collect these certs and copy them to the correct `/certs/` directory, instead of only considering those from the `trustedCAProxy`
2. We should investigate if we can strip out the current cert copying mechanism that is required to get trusted certs into each new importer pod and replace it with a `certmanger`, like how OCP leverages `inject-trusted-cabundle`
3. Need to also investigate how this current vs new approach handles install time CAs.

**Additional context**:
https://github.com/kubevirt/containerized-data-importer/pull/4109 fixes a bug in this area, refer to the thread for more information.

HoustonBoston:
This adds `CDIConfig.spec.trustedCA` to hold the non-proxy CA certificates

Fixes #4152

An intern looking for feedback since I am a beginner and have not worked with large codebases before.

**Release note**:

```release-note
action required: users must set their `trustedCA` configMaps in `CDIConfig.spec.trustedCA` for non-proxy CA certificates and must use `CDIConfig.spec.ImportProxy.TrustedCAProxy` for proxy CA certificates.
```


pkg/common/common.go
```diff
@@ -99,11 +99,14 @@ const (
 	// DefaultPullPolicy imports k8s "IfNotPresent" string for the import_controller_gingko_test and the cdi-controller executable
 	DefaultPullPolicy = string(v1.PullIfNotPresent)
 	// ImportProxyConfigMapName provides the key for getting the name of the ConfigMap in the cdi namespace containing a CA certificate bundle
-	ImportProxyConfigMapName = "trusted-ca-proxy-bundle-cm"
+	// ImportProxyConfigMapName = "trusted-ca-proxy-bundle-cm" // Deprecated: Use TrustedCAConfigMapName instead
```
Acedus: Says it's deprecated but it was removed?

Acedus (commented):
Thanks @HoustonBoston. I think the issue with this PR is that it breaks a real usecase of having both a proxy and a default CA.  There are three distinct independent certificate bundles here all-in-all:
1. The cluster-wide CA which would be the one we expect to use for every import.
2. The proxy CA which would be necessary if there exists a proxy that does TLS interception in the import path.
3. The per-DV endpoint CA which would be used for a specific endpoint that may have its own self-signed cert.

The fix should therefore be additive, we can add the new `trustedCA` field alongside the existing `trustedCAProxy`. The controller would mount it at its own path, and we'd just merge certs from all of the directories (as we do now anyway).

pkg/controller/import-controller.go
```diff
@@ -351,11 +356,50 @@ func (r *ImportReconciler) copyImportProxyConfigMap(pvc *corev1.PersistentVolume
 	return nil
 }
 
+func (r *ImportReconciler) copyTrustedCAConfigMap(pvc *corev1.PersistentVolumeClaim, pod *corev1.Pod) error {
```
dsanatar: this is almost identical to `copyImportProxyConfigMap()`, could we not combine the two to reduce duplication? I think all that needs to be different is passing in the configmap names

pkg/controller/config-controller.go
```diff
@@ -635,6 +635,11 @@ func (r *CDIConfigReconciler) createProxyConfigMap(cmName, cert string) *v1.Conf
 	}
 }
 
+func (r *CDIConfigReconciler) reconcileTrustedCA(config *cdiv1.CDIConfig) error {
```
dsanatar, line 642: this isn't called anywhere?

pkg/importer/http-datasource.go
```diff
@@ -291,6 +291,19 @@ func createCertPool(certDir string) (*x509.CertPool, error) {
...
+			if file.IsDir() || file.Name()[0] == '.' {
+				continue
+			}
+			fp := path.Join(common.ImportTrustedCACertDir, file.Name())
+			if certs, err := os.ReadFile(fp); err == nil {
+				certPool.AppendCertsFromPEM(certs)
+			}
+		}
+	}
+
```
dsanatar: we can probably extract this to a common function so we can use it to populate the certPool from the two directories

pkg/controller/import-controller_test.go
```diff
@@ -1081,6 +1081,95 @@ var _ = Describe("Create Importer Pod", func() {
 		Entry("should create pod with block volume mode and scratchspace", createBlockPvc("testBlockPvc1", "default", map[string]string{cc.AnnEndpoint: testEndPoint, cc.AnnPodPhase: string(corev1.PodPending), cc.AnnImportPod: "podName", cc.AnnPriorityClassName: "p0", cc.AnnPodServiceAccount: "my-sa"}, nil), &scratchPvcName),
 	)
 
+	DescribeTable("should create pod with trustedCA cert volume", func(certConfigMapTrustedCA string, certConfigMapProxy string, expectedVolumeMounts int, expectedVolumes int) {
+		pvc := cc.CreatePvc("testPvc1", "default", map[string]string{
+			cc.AnnEndpoint:  testEndPoint,
+			cc.AnnPodPhase:  string(corev1.PodPending),
+			cc.AnnImportPod: "podName",
+		}, nil)
+		reconciler := createImportReconciler(pvc)
```
dsanatar, line 1092: this doesn't actually trigger a reconcile loop, you need to call `reconciler.Reconcile()` with the resources you want to be reconciled. if you give it the PVC then it will create the importer pod for you and you can inspect it from there

dsanatar (commented):
I think this needs a functional test to test this new process e2e. You can refer to examples of how this is done for `importProxy.trustedCAProxy` in `import_proxy_test.go`. I think we want to see when this configuration is set, when a DV is created, the associated importer pod will automatically get the global defined ca bundle.

tests/import_proxy_test.go
```diff
@@ -367,6 +376,47 @@ var _ = Describe("Import Proxy tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		Describe("Global TrustedCA tests", func() {
+			FIt("should import from https endpoint using global trustedCA without proxy", func() {
```
dsanatar: leftover focus here

pkg/controller/config-controller_test.go
```diff
@@ -953,6 +953,59 @@ var _ = Describe("GetImportProxyConfig", func() {
...
+		Expect(cdiConfig.Status.TrustedCA).To(HaveValue(Equal("my-ca-bundle")))
+	})
+
+	It("should set Status.TrustedCA to nil when Spec.TrustedCA is nil", func() {
+		reconciler, cdiConfig := createConfigReconciler()
+		cdiConfig.Spec.TrustedCA = nil
+
+		err := reconciler.reconcileTrustedCA(cdiConfig)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cdiConfig.Status.TrustedCA).To(BeNil())
```
dsanatar, line 975: I feel like this case should ensure the the status is populated as nil even when it's not explicitly defined in the spec, like what currently exists for the `importProxy` struct

pkg/controller/config-controller_test.go
```diff
@@ -953,6 +953,59 @@ var _ = Describe("GetImportProxyConfig", func() {
...
+		Expect(result).To(BeEmpty())
+	})
+
+	It("should return empty string when TrustedCA is nil", func() {
+		cdiConfig := MakeEmptyCDIConfigSpec("cdiconfig")
+		cdiConfig.Status.TrustedCA = nil
+		result, err := GetTrustedCA(cdiConfig)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(result).To(BeEmpty())
+	})
```
dsanatar: this looks to be the exact same test

pkg/controller/import-controller.go
```diff
@@ -298,9 +299,23 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 			}
 		} else {
 			// Copy import proxy ConfigMap (if exists) from cdi namespace to the import namespace
-			if err := r.copyImportProxyConfigMap(pvc, pod); err != nil {
+			cdiConfig := &cdiv1.CDIConfig{}
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, cdiConfig); err != nil {
 				return reconcile.Result{}, err
 			}
+			proxyCmName, _ := GetImportProxyConfig(cdiConfig, common.ImportProxyConfigMapName)
```
dsanatar, line 306: should read in this error return instead of ignoring it

pkg/controller/import-controller.go
```diff
@@ -298,9 +299,23 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
...
+			if proxyCmName != "" {
+				if err := r.copyTrustedCAOrProxyCAConfigMap(pvc, pod, proxyCmName, GetImportProxyConfigMapName(pvc.Name)); err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+			trustedCACmName, _ := GetTrustedCA(cdiConfig)
```
dsanatar, line 312: same here

pkg/controller/import-controller.go
```diff
@@ -317,22 +332,15 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
...
-	cdiConfig := &cdiv1.CDIConfig{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, cdiConfig); err != nil {
-		return err
-	}
-	cmName, err := GetImportProxyConfig(cdiConfig, common.ImportProxyConfigMapName)
-	if err != nil || cmName == "" {
-		return nil
-	}
+// helper to copy trusted CA or proxy CA config map to the import namespace
+func (r *ImportReconciler) copyTrustedCAOrProxyCAConfigMap(pvc *corev1.PersistentVolumeClaim, pod *corev1.Pod, cmName string, importConfigMapName string) error {
```
dsanatar: nit: function name a little long

pkg/controller/import-controller.go
```diff
@@ -317,22 +332,15 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
...
 	cdiConfigMap := &corev1.ConfigMap{}
 	if err := r.uncachedClient.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: r.cdiNamespace}, cdiConfigMap); err != nil {
```
dsanatar, line 342: you lost the check for if the `cmName` is empty, we probably should skip the Get call if we already know something went wrong

pkg/common/common.go
```diff
@@ -98,12 +98,18 @@ const (
...
 	ImportProxyConfigMapName = "trusted-ca-proxy-bundle-cm"
-	// ImportProxyConfigMapKey provides the key name of the ConfigMap in the cdi namespace containing a CA certificate bundle
+	// ImportTrustedCAConfigMapName provides the key for getting the name of the ConfigMap in the cdi namespace containing a CA certificate bundle
+	ImportTrustedCAConfigMapName = "trusted-ca-bundle-cm"
+	// ImportProxyConfigMapKey provides the key name of the ConfigMap in the cdi namespace containing a proxy CA certificate bundle
 	ImportProxyConfigMapKey = "ca.crt"
 	// ImporterProxyCertDir is where the configmap containing proxy certs will be mounted
 	ImporterProxyCertDir = "/proxycerts/"
+	// ImportTrustedCAConfigMapKey provides the key name of the ConfigMap in the cdi namespace containing a CA certificate bundle
+	ImportTrustedCAConfigMapKey = "ca.crt"
```
dsanatar: I don't see this being used?

pkg/controller/import-controller_test.go
```diff
@@ -1081,6 +1081,95 @@ var _ = Describe("Create Importer Pod", func() {
...
+
+		// Update CDIConfig with trustedCA and proxy settings
+		cdiConfig := &cdiv1.CDIConfig{}
+		err := reconciler.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, cdiConfig)
+		Expect(err).ToNot(HaveOccurred())
+		if certConfigMapTrustedCA != "" {
+			cdiConfig.Status.TrustedCA = &certConfigMapTrustedCA
+		}
+		if certConfigMapProxy != "" {
```
dsanatar: I don't think this needs to also test the trustedCAProxy

pkg/importer/http-datasource.go
```diff
@@ -327,12 +334,16 @@ func createHTTPClient(certDir string, insecureSkipVerify bool) (*http.Client, er
...
 		klog.Warningf("Unable to read proxy cert directory %v", err)
 	}
 
-	if certDir == "" && len(proxyCertDir) == 0 && !insecureSkipVerify {
+	trustedCACertDir, err := os.ReadDir(common.ImportTrustedCACertDir)
+	if err != nil && !os.IsNotExist(err) {
+		klog.Warningf("Unable to read trusted CA cert directory %v", err)
+	}
+
+	if certDir == "" && len(proxyCertDir) == 0 && len(trustedCACertDir) == 0 && !insecureSkipVerify {
```
dsanatar: nit: I probably set a bad example when I wrote this, but it might be easier/cleaner if you did an `os.Stat` on the dir instead. not a deal breaker though.

dsanatar (changes requested):

pkg/importer/http-datasource.go
```diff
@@ -327,12 +334,17 @@ func createHTTPClient(certDir string, insecureSkipVerify bool) (*http.Client, er
...
 	}
 
-	if certDir == "" && len(proxyCertDir) == 0 && !insecureSkipVerify {
+	trustedCACertDir, err := os.Stat(common.ImportTrustedCACertDir)
+	if err != nil && !os.IsNotExist(err) {
+		klog.Warningf("Unable to read trusted CA cert directory %v", err)
+	}
+
+	hasTrustedCACerts := err == nil && trustedCACertDir.IsDir()
+	if certDir == "" && len(proxyCertDir) == 0 && !hasTrustedCACerts && !insecureSkipVerify {
```
dsanatar: probably should return the error if we get a real one that isn't not exist. also if we are going to use this pattern for the one dir we should also use it for the proxy cert dir just so they are aligned

pkg/controller/import-controller.go
HoustonBoston: That causes a UT I did not write to fail, so I think ignoring it is the best option.

pkg/controller/import-controller.go
HoustonBoston: Same as above

pkg/controller/import-controller_test.go
HoustonBoston: I think it does since it has a case for when both proxy and trustedCA are configured.

HoustonBoston: Entry("when both proxy and trustedCA are configured",
			"my-ca-bundle", "proxy-ca", 3, 3), // data + proxy + trustedCA
	)

dsanatar: my point is I don't think you need the proxy case at all. the tests main objective is to test the new field, it becomes overly complex to also consider the proxy as well (esp since that's already being tested elsewhere)

HoustonBoston: Fair, I'll remove that

dsanatar (approved):

pkg/controller/config-controller.go
```diff
@@ -635,6 +639,11 @@ func (r *CDIConfigReconciler) createProxyConfigMap(cmName, cert string) *v1.Conf
...
```
davmarro, line 642: Why this reconcile doesn't check anything ? 
I'm expecting at least a sort of validation before setting the TrustedCA: the user can add any string (also referencing not existing maps) and still getting it added.

You can also avoid returning the error type if the function always return nil.

pkg/controller/import-controller.go
davmarro: Hey @HoustonBoston maybe the UT is not working well and should be fixed.
The errors should always be checked and returned if `!= nil`, we can't ignore them 😄

pkg/controller/config-controller.go
HoustonBoston: Good point. I think pre-checking for a CM's existence before setting the TrustedCA would be helpful

HoustonBoston: Now I added in some validation and unit tested them.

pkg/controller/dataimportcron-controller.go
```diff
@@ -1628,6 +1628,15 @@ func InitPollerPod(c client.Client, cron *cdiv1.DataImportCron, pod *corev1.PodT
 		volumes = append(volumes, createConfigMapVolume(CertVolName, *regSource.CertConfigMap))
 	}
 
+	if volName, _ := GetTrustedCA(cdiConfig); volName != "" {
```
davmarro: Instead of checking every time if the string got from `GetTrustedCA` is `!= nil`, you can just return an error in the main function and check for the err != nil. Always check the errors, do not ignore them.

pkg/controller/dataimportcron-controller_test.go
```diff
@@ -271,6 +271,43 @@ var _ = Describe("All DataImportCron Tests", func() {
 			Expect(err).To(HaveOccurred())
 		})
 
+		It("Should create CronJob with trustedCA volume when TrustedCA is configured", func() {
```
davmarro, line 274: Can you please test also some negative scenarios ?

pkg/controller/import-controller_test.go
```diff
@@ -1131,6 +1081,71 @@ var _ = Describe("Create Importer Pod", func() {
 		Entry("should create pod with block volume mode and scratchspace", createBlockPvc("testBlockPvc1", "default", map[string]string{cc.AnnEndpoint: testEndPoint, cc.AnnPodPhase: string(corev1.PodPending), cc.AnnImportPod: "podName", cc.AnnPriorityClassName: "p0", cc.AnnPodServiceAccount: "my-sa"}, nil), &scratchPvcName),
 	)
 
+	It("should create pod with trustedCA cert volume when trustedCA is configured", func() {
```
davmarro, line 1084: Can we cover also other scenarios other than the successful one ?

pkg/importer/http-datasource.go
```diff
@@ -271,25 +271,32 @@ func (hs *HTTPDataSource) Close() error {
...
-	certPool, err := x509.SystemCertPool()
-	if err != nil {
-		return nil, errors.Wrap(err, "Error getting system certs")
-	}
-
-	// append the user-provided trusted CA certificates bundle when making egress connections using proxy
-	if files, err := os.ReadDir(common.ImporterProxyCertDir); err == nil {
+// helper to append certs to cert pool
+func appendCertsToPool(certPool *x509.CertPool, dir string) {
+	if files, err := os.ReadDir(dir); err == nil {
```
davmarro: why are we silently ignoring the errors from os.ReadDir ?
Imagine that, when you call `createCertPool`, the `dir` doesn't exist. The Certs are not appended but you do not receive any error and you believe that everything is fine.
We can return the error and be sure that if something goes wrong, we got all the errors.

pkg/importer/http-datasource.go
```diff
@@ -326,13 +333,20 @@ func createHTTPClient(certDir string, insecureSkipVerify bool) (*http.Client, er
...
+	hasProxyCerts := err == nil && proxyCertDir.IsDir()
 
-	if certDir == "" && len(proxyCertDir) == 0 && !insecureSkipVerify {
+	trustedCACertDir, err := os.Stat(common.ImportTrustedCACertDir)
+	if err != nil && !os.IsNotExist(err) {
+		klog.Warningf("Unable to read trusted CA cert directory %v", err)
+		return nil, err
+	}
+	hasTrustedCACerts := err == nil && trustedCACertDir.IsDir()
+	if certDir == "" && !hasTrustedCACerts && !hasProxyCerts && !insecureSkipVerify {
```
davmarro: I know that this is not your code but if we could check `certDir == nil` as first in the method then we can directly early return the empty client. In this way we can skip all this "useless" logic execution.

tests/import_proxy_test.go
```diff
@@ -367,6 +376,47 @@ var _ = Describe("Import Proxy tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		Describe("Global TrustedCA tests", func() {
+			It("should import from https endpoint using global trustedCA without proxy", func() {
```
davmarro, line 390: Can we check other scenarios ?

pkg/controller/import-controller.go
```diff
@@ -1478,6 +1503,12 @@ func makeImportEnv(podEnvVar *importPodEnvVar, uid types.UID) []corev1.EnvVar {
 			Value: common.ImporterProxyCertDir,
 		})
 	}
+	if podEnvVar.certConfigMapTrustedCA != "" {
+		env = append(env, corev1.EnvVar{
+			Name:  common.ImporterTrustedCADirVar,
```
davmarro, line 1508: Correct me if I'm wrong but you set this variable but you never read it. Am i Wrong ?

pkg/controller/config-controller.go
davmarro: ok, you can take the `reconcileImportProxyCAConfigMap` as an example of how this reconcile loop can be implemented:
- if the referenced ConfigMap doesn't exist, it emits a Warning Event using r.recorder.Event()
- consider handling k8serrors.IsNotFound(err) explicitly — you can emit a warning event and either set Status.TrustedCA = nil (without blocking the entire reconciliation)
- read and validate the CM content

pkg/controller/dataimportcron-controller.go
HoustonBoston: I think the issue with not ignoring the error and returning it is that we're treating trustedCA when not configured as fatal, and that causes a test I did not write to fail unnecessarily.

HoustonBoston: Never mind, found a workaround while returning the error

pkg/importer/http-datasource.go
HoustonBoston: The case where `certDir == ""` is handled later in that function where it returns an empty client if certDir is empty and it doesn't have certs and `insecureSkipVerify` is false.

pkg/controller/import-controller.go
HoustonBoston: You're correct, but for some reason the one above that is not read either...

HoustonBoston: Changed the logic s.t. the env var is read in by the importer.

tests/import_proxy_test.go
HoustonBoston: Yes sir

pkg/importer/http-datasource.go
davmarro: Yes, it's the line I commented, right ? 
What I was saying is that the condition `certDir == ""` is in AND with other conditions and if this is false then all the other conditions are not evaluated. Given the fact that the certDir is passed as parameter in the function, we can add something like that at the start of the function :
```
if certDir == "" {
return emptyClient 
}
```
And remove the certDir check at line 349. 

In this way we can early return if the certDir is empty, without executing all those lines of codes.

HoustonBoston: I'm not fully familiar with why it needs to return early if certDir (now called `importerCertDir`) is empty in the context of this function. @dsanatar any suggestions here?

davmarro: Hey @HoustonBoston sorry but I was wrong, you can resolve the conversation!

pkg/importer/http-datasource.go
```diff
@@ -90,7 +90,7 @@ type HTTPDataSource struct {
 var createNbdkitCurl = image.NewNbdkitCurl
 
 // NewHTTPDataSource creates a new instance of the http data provider.
-func NewHTTPDataSource(endpoint, accessKey, secKey, certDir string, contentType cdiv1.DataVolumeContentType, checksum string, insecureSkipVerify bool) (*HTTPDataSource, error) {
+func NewHTTPDataSource(endpoint, accessKey, secKey, importerCertDir string, trustedCACertDir string, proxyCertDir string, contentType cdiv1.DataVolumeContentType, checksum string, insecureSkipVerify bool) (*HTTPDataSource, error) {
```
dsanatar, line 93: nit: do we have to add these new params? it seems as though it was fine to just read the env vars when we collect the certs. changing this function definition just adds all those extra changes in the unit tests (especially for them to just be empty strings)

pkg/importer/imageio-datasource.go
```diff
@@ -493,8 +493,8 @@ func (reader *extentReader) GetRange(start, end int64) (io.ReadCloser, error) {
 	return response.Body, nil
 }
 
-func createImageioReader(ctx context.Context, ep string, accessKey string, secKey string, certDir string, diskID string, currentCheckpoint string, previousCheckpoint string, insecureSkipVerify bool) (io.ReadCloser, uint64, *ovirtsdk4.ImageTransfer, ConnectionInterface, error) {
-	conn, err := newOvirtClientFunc(ep, accessKey, secKey, certDir, insecureSkipVerify)
+func createImageioReader(ctx context.Context, ep string, accessKey string, secKey string, importerCertDir string, diskID string, currentCheckpoint string, previousCheckpoint string, insecureSkipVerify bool) (io.ReadCloser, uint64, *ovirtsdk4.ImageTransfer, ConnectionInterface, error) {
```
dsanatar, line 496: nit: do we need to rename this?

pkg/importer/s3-datasource.go
```diff
@@ -166,9 +166,9 @@ func createS3Reader(ep *url.URL, accessKey, secKey string, certDir string) (io.R
 	return objectReader, nil
 }
 
-func getS3Client(endpoint, accessKey, secKey string, certDir string, urlScheme string) (S3Client, error) {
+func getS3Client(endpoint, accessKey, secKey string, importerCertDir string, urlScheme string) (S3Client, error) {
```
dsanatar, line 169: nit: same here, name change?

pkg/importer/http-datasource.go
```diff
@@ -271,68 +271,53 @@ func (hs *HTTPDataSource) Close() error {
...
-
-			fp := path.Join(certDir, file.Name())
-
-			klog.Infof("Attempting to get certs from %s", fp)
-
-			certs, err := os.ReadFile(fp)
-			if err != nil {
-				return nil, errors.Wrapf(err, "Error reading file %s", fp)
-			}
+func createCertPool(importerCertDir string, trustedCACertDir string, proxyCertDir string) (*x509.CertPool, error) {
```
dsanatar, line 295: it seems as though most of the time one or two of these params will be empty strings, seems easier to just check the env vars when we need to and collect the certs when they exist

dsanatar (commented):
I think we are in pretty good shape here, just some last minute suggestions

pkg/controller/config-controller.go
awels: Emitting an event is great. The one problem with events is that they are only there for an hour, after that they get purged and the reason is lost. This is why we have conditions on most resources. The condition can keep a permanent record of what happened. In this case a missing or invalid config map.

pkg/controller/config-controller_test.go
```diff
@@ -953,6 +953,71 @@ var _ = Describe("GetImportProxyConfig", func() {
...
+		reconciler, cdiConfig := createConfigReconciler(cm)
+		trustedCA := "my-ca-bundle"
+		cdiConfig.Spec.TrustedCA = &trustedCA
+
+		err := reconciler.reconcileTrustedCA(cdiConfig)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cdiConfig.Status.TrustedCA).To(HaveValue(Equal("my-ca-bundle")))
+	})
+
+	It("should set Status.TrustedCA to nil when Spec.TrustedCA is nil", func() {
```
awels: These 3 Its look like they are basically just different variations of the same test. You can use a DescribeTable to consolidate the test setup a bit I think. Also it looks like a missing test is if there is a `.Status.TrustedCA` set and the spec changes to something invalid or nil, the status should become nil.

pkg/controller/config-controller_test.go
```diff
@@ -953,6 +953,71 @@ var _ = Describe("GetImportProxyConfig", func() {
...
+		cdiConfig := MakeEmptyCDIConfigSpec("cdiconfig")
+		cdiConfig.Status.TrustedCA = nil
+		result, err := GetTrustedCA(cdiConfig)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(result).To(BeEmpty())
+	})
+
+	It("should return error when CDIConfig is nil", func() {
+		_, err := GetTrustedCA(nil)
+		Expect(err).To(HaveOccurred())
```
awels, line 1018: Can you verify we are getting the expected error, I think in this case it is `failed to get field, the CDIConfig is nil` a typed error is probably even nicer, but not a necessity.

tests/import_proxy_test.go
```diff
@@ -367,6 +376,235 @@ var _ = Describe("Import Proxy tests", func() {
...
+					context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{})
+			})
+
+			It("should import from https endpoint using global trustedCA without proxy", func() {
+				By("copying contents of file host cert config map to CDI NS")
+				caConfigMapName, err := utils.CopyConfigMap(f.K8sClient, f.CdiInstallNs, utils.FileHostCertConfigMap,
+					f.CdiInstallNs, "test-global-trusted-ca", "ca-bundle.crt")
+				Expect(err).ToNot(HaveOccurred())
+
+				By("setting CDIConfig Spec.TrustedCA")
```
awels: I think it would be interesting to see what happens if we set the TrustedCA before creating the config map. It should lead to a nil status, and then we create the configmap, and the status should switch to match.

tests/import_proxy_test.go
```diff
@@ -367,6 +376,235 @@ var _ = Describe("Import Proxy tests", func() {
...
+				pvc, err := utils.WaitForPVC(f.K8sClient, dataVolume.Namespace, dvName)
+				Expect(err).ToNot(HaveOccurred())
+				f.ForceBindIfWaitForFirstConsumer(pvc)
+
+				By("verifying the import fails")
+				err = utils.WaitForDataVolumePhase(f, f.Namespace.Name, cdiv1.ImportInProgress, dv.Name)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("should succeed after trustedCA is updated from nonexistent to valid ConfigMap", func() {
```
awels, line 487: To go with my previous comment, this is doing exactly what I was thinking. So this and the other test are basically the same, and we can probably eliminate one of them. This test is more interesting than the first one, so lets keep this one.

awels (commented):
Looks good some nitpicks.

pkg/controller/config-controller_test.go
HoustonBoston: I added in a string matcher from the error message.

davmarro (approved):

tests/import_proxy_test.go
```diff
@@ -367,6 +367,196 @@ var _ = Describe("Import Proxy tests", func() {
...
 		})
 
+		Describe("Global TrustedCA tests", func() {
+			AfterEach(func() {
+				_ = f.K8sClient.CoreV1().ConfigMaps(f.CdiInstallNs).Delete(
+					context.TODO(), "test-invalid-trusted-ca", metav1.DeleteOptions{})
+				_ = f.K8sClient.CoreV1().ConfigMaps(f.CdiInstallNs).Delete(
+					context.TODO(), "test-global-trusted-ca", metav1.DeleteOptions{})
+				// clear every DV in the test ns
+				_ = f.CdiClient.CdiV1beta1().DataVolumes(f.Namespace.Name).DeleteCollection(
```
Acedus: I believe the test framework already give us this for free.

pkg/controller/import-controller.go
```diff
@@ -298,9 +299,29 @@ func (r *ImportReconciler) reconcilePvc(pvc *corev1.PersistentVolumeClaim, log l
 			}
 		} else {
 			// Copy import proxy ConfigMap (if exists) from cdi namespace to the import namespace
-			if err := r.copyImportProxyConfigMap(pvc, pod); err != nil {
+			cdiConfig := &cdiv1.CDIConfig{}
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Name: common.ConfigName}, cdiConfig); err != nil {
 				return reconcile.Result{}, err
 			}
+			proxyCmName, err := GetImportProxyConfig(cdiConfig, common.ImportProxyConfigMapName)
+			if err != nil {
```
Acedus: Any reason why we no longer have `r.copyImportProxyConfigMap` and chose to inline this here? Makes the function less readable IMO.

Also, `copyImportProxyConfigMap` was returning nil on error when calling `GetImporyProxyConfig` but we don't and because `GetImportProxyConfig` returns an error on nil `config.Status.ImportProxy` `reconcilePvc` would always fail for a CDI config with not import proxy.

pkg/controller/config-controller.go
```diff
@@ -118,6 +118,10 @@ func (r *CDIConfigReconciler) Reconcile(_ context.Context, req reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	if err := r.reconcileTrustedCA(config); err != nil {
```
Acedus, line 123: We add this reconciliation logic but we don't actually watch the trusted CA CM: https://github.com/kubevirt/containerized-data-importer/blob/c2f4af455f2e5d9c09819673fa1dd5044cf3bfa3/pkg/controller/config-controller.go#L683-L705

pkg/controller/config-controller.go
```diff
@@ -635,6 +639,33 @@ func (r *CDIConfigReconciler) createProxyConfigMap(cmName, cert string) *v1.Conf
...
+	if k8serrors.IsNotFound(err) {
+		msg := fmt.Sprintf(messageResourceDoesntExist, *config.Spec.TrustedCA)
+		config.Status.TrustedCA = nil
+		r.recorder.Event(config, v1.EventTypeWarning, errResourceDoesntExist, msg)
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if _, ok := cm.Data["ca-bundle.crt"]; !ok {
```
Acedus: nit: We have a const for this as `ClusterWideProxyConfigMapKey` (might need to update the name or clone it, though).

Acedus: Also, `appendCertsToPool` doesn't actually verify a specific key, it just reads all mounted files from the dir and tries to append them as certs.

pkg/controller/config-controller.go
```diff
@@ -635,6 +639,33 @@ func (r *CDIConfigReconciler) createProxyConfigMap(cmName, cert string) *v1.Conf
...
+		config.Status.TrustedCA = nil
+		r.recorder.Event(config, v1.EventTypeWarning, errResourceDoesntExist,
```
Acedus: But the resource does exist, it's just that the specific key doesn't.

pkg/importer/http-datasource.go
```diff
@@ -271,49 +271,44 @@ func (hs *HTTPDataSource) Close() error {
...
 
-			certs, err := os.ReadFile(fp)
-			if err != nil {
-				return nil, errors.Wrapf(err, "Error reading file %s", fp)
-			}
+	trustedCACertDir, _ := util.ParseEnvVar(common.ImporterTrustedCADirVar, false)
+	proxyCertDir, _ := util.ParseEnvVar(common.ImporterProxyCertDirVar, false)
 
-			if ok := certPool.AppendCertsFromPEM(certs); !ok {
-				klog.Warningf("No certs in %s", fp)
```
Acedus, line 316: Any reason why we drop these useful logs?

pkg/common/common.go
```diff
@@ -98,12 +98,16 @@ const (
 	ImporterCertDir = "/certs"
 	// DefaultPullPolicy imports k8s "IfNotPresent" string for the import_controller_gingko_test and the cdi-controller executable
 	DefaultPullPolicy = string(v1.PullIfNotPresent)
-	// ImportProxyConfigMapName provides the key for getting the name of the ConfigMap in the cdi namespace containing a CA certificate bundle
+	// ImportProxyConfigMapName provides the key for getting the name of the ConfigMap in the cdi namespace containing a proxy CA certificate bundle
 	ImportProxyConfigMapName = "trusted-ca-proxy-bundle-cm"
-	// ImportProxyConfigMapKey provides the key name of the ConfigMap in the cdi namespace containing a CA certificate bundle
+	// ImportTrustedCAConfigMapName provides the key for getting the name of the ConfigMap in the cdi namespace containing a CA certificate bundle
+	ImportTrustedCAConfigMapName = "trusted-ca-bundle-cm"
```
Acedus: This isn't used anywhere AFAIK.

staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go
```diff
@@ -1107,6 +1109,8 @@ type CDIConfigStatus struct {
 	// ImportProxy contains importer pod proxy configuration.
 	// +optional
 	ImportProxy *ImportProxy `json:"importProxy,omitempty"`
+	// TrustedCA is the name of a ConfigMap in the cdi namespace that contains a user-provided trusted certificate authority (CA) bundle.
+	TrustedCA *string `json:"trustedCA,omitempty"`
```
Acedus, line 1114: nit: Missing `+optional`.

Acedus (commented):
Additional pass, see comments.
Also, I believe we should update `doc/cdi-config.md` to reflect this change.

tests/import_proxy_test.go
HoustonBoston: I don't think it deletes CMs in the cdi ns since it's not part of the test ns.

pkg/controller/config-controller.go
HoustonBoston: In that case I'll not check for a specific key.

pkg/importer/http-datasource.go
HoustonBoston: Probably not...

PR #4237, opened 2026-08-09
Switch import controller node-pull to image volume

### LINKED ISSUE #4253: importer: fixed 60s ready-file wait races the image pull it is waiting for (pullMethod: node)
**What happened**:

For registry imports with `pullMethod: node`, `cdi-importer` waits for the readiness file written by the `cdi-containerimage-server` sidecar. That wait is a fixed 60-second poll:

```go
// cmd/cdi-importer/importer.go
func waitForReadyFile() {
	const readyFileTimeoutSeconds = 60
	...
	os.Exit(1)
}
```

The sidecar runs **the disk image being imported as its own container image**, which is the point of `RegistryPullNode` — kubelet pulls it, then the sidecar serves its contents over localhost. So the ready file cannot appear until kubelet has finished pulling a potentially very large image.

Kubelet starts containers sequentially, so the importer container is already running when the pull of the sidecar image begins. The 60-second clock therefore has to cover the *entire* pull. Whenever that pull takes longer than 60 seconds, the importer writes a termination message and exits 1, producing a `Warning ErrImportFailed` event, even though nothing has actually failed.

Events from one affected importer pod (identifiers removed):

```
15:58:05Z  Scheduled        -> <node>
15:58:08Z  Started          container importer          <- 60s clock starts
15:58:08Z  Pulling          image "<vm-disk-image>"
16:00:01Z  Pulled           image ... in 1m54.306s      <- 114s, 54s past the deadline
16:00:01Z  Started          container server            <- ready file written
16:00:02Z  ErrImportFailed  Timeout waiting for file /shared/ready
16:00:02Z  Started          container importer          <- kubelet restart, file now present
16:00:12Z  ImportSucceeded  Import Successful
```

It is self-healing: kubelet restarts the importer, the file is present by then, and the import succeeds. The cost is a spurious failure event plus roughly 1.5-2.5 minutes of added latency per DataVolume, and the timeout is a compile-time constant so there is no way to tune it.

A registry import whose image simply takes longer than 60 seconds to pull should complete without reporting a failure. The readiness wait should tolerate a pull of arbitrary size, since image size and registry throughput are what determine it.

1. Create a `DataVolume` with a registry source using `pullMethod: node`, referencing an image large enough that pulling it to a node without a cached copy takes more than 60 seconds:

   ```yaml
   kind: DataVolume
   metadata:
     name: example-import
   spec:
     source:
       registry:
         url: docker://example.registry/vm-disk-image:tag
         pullMethod: node
     storage:
       resources:
         requests:
           storage: 100Gi
   ```

2. Ensure the target node has not already cached that image.
3. Observe the importer pod. `Started container importer` and `Pulling` occur in the same second; about 60 seconds later the importer exits with `Timeout waiting for file /shared/ready`; the `server` container starts once the pull finishes; kubelet restarts the importer and the import then succeeds.

Running several such imports concurrently makes it easier to hit, since contention for registry bandwidth pushes more pulls past 60 seconds.

**Additional context**:

Measured across 230 importer pods on two clusters over 19 hours:

| importer pods | image pull > 60s | failed on first attempt |
|---|---|---|
| no `server` container (other source types) | n/a | 0 / 69 |
| has `server`, pull <= 60s or image cached | no | 0 / 68 |
| has `server`, pull > 60s | yes | 93 / 93 |

No counterexamples in either direction. Observed pull durations for images that crossed the threshold ranged from 88.3s to 153.9s; for three of the disk images every single pull exceeded 60 seconds (22 of 22), so those imports failed on first attempt 100% of the time.

In a separate 30-day sample, 1,694 of 1,704 affected DataVolumes reached `ImportSucceeded` after the restart (median recovery about 7 seconds, p99 26 seconds), which is why this shows up as event noise and latency rather than broken imports.

Ruled out by measurement: scratch space (affected imports carry `cdi.kubevirt.io/storage.import.requiresScratch=false`), disk capacity, node pressure, image-pull failure (no `ErrImagePull` or `ImagePullBackOff`; the pulls succeed, just slower than 60 seconds), and the storage backend.

Only `pullMethod: node` is affected. Other sources leave `IMPORTER_READY_FILE` empty and `waitForReadyFile()` returns immediately, which matches the 0/69 above.

Note on history: the bound was added deliberately in #2221 so that a stuck import surfaces through the DataVolume Running condition, replacing an unbounded wait. That intent seems right; the difficulty is that 60 seconds is shorter than the legitimate duration of the thing being waited on, so the mechanism now fires on healthy imports and the signal it provides is a false one.

Possible directions, roughly in increasing order of size:

1. Keep the bounded wait and its termination message, but raise the bound. It then only has to catch a sidecar that never becomes ready, where a generous value costs nothing. Proposed in #4254.
2. Expose the value through `CDIConfig` so operators can tune it per installation.
3. Express the dependency structurally instead of by polling. Because the sidecar's readiness really means "kubelet finished pulling my image", making it a sidecar container (an init container with `restartPolicy: Always`) would let kubelet order startup and remove the need for any timeout. A genuinely stuck pull would then surface as `ImagePullBackOff`, which is arguably better than the current termination message. This is the largest change and has a minimum Kubernetes version implication, so I have not assumed it.

Happy to follow whichever direction you prefer.

**Environment**:
- CDI version (use `kubectl get deployments cdi-deployment -o yaml`): observed on v1.62.0; the constant is unchanged on `main` as of 36976df5589d4ff927946cace29f862f42b2d153
- DV specification: registry source with `pullMethod: node`, as shown above
- Cloud provider or hardware configuration: bare metal
- OS (e.g. from /etc/os-release): Azure Linux 3.0
- Kernel (e.g. `uname -a`): 6.6.130.1-1.azl3
- Install tools: N/A
- Others: found during large-scale internal durability testing; disk images in the multi-gigabyte range

nav1s:
### Background

The current node-pull flow works by:
1. Changing the datasource from registry to HTTP
2. Running a secondary container that serves the disk image and env vars over HTTP
3. Having the importer fetch both via HTTP requests
4. Using ready/done files for coordination between the containers

which is a bit convoluted.

### Changes

The suggested new flow uses Kubernetes [image volume](https://kubernetes.io/docs/concepts/storage/volumes/#image) to mount the registry image content directly into the importer container, This eliminates:
- The HTTP server in `cdi-containerimage-server`
- The source-type change from registry to HTTP
- The ready/done file coordination mechanism
- The node-pull special-casing in `HTTPDataSource`

Currently image volumes don't support environment variable propagation which we need in order to support [PVC label decoration](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/image-from-registry.md#decorate-the-pvc-of-the-imported-containerdisk-with-additional-labels), so the secondary container is still needed.

Fixes #4253

**Release note**:

```release-note
NONE
```


pkg/controller/import-controller.go
```diff
@@ -1184,12 +1178,23 @@ func makeImporterVolumeSpec(args *importerPodArgs) []corev1.Volume {
...
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
 			},
-		})
+			corev1.Volume{
+				Name: "image-volume",
+				VolumeSource: corev1.VolumeSource{
+					Image: &corev1.ImageVolumeSource{
+						Reference:  args.importImage,
+						PullPolicy: corev1.PullIfNotPresent,
```
Acedus: The image volume hardcodes `PullIfNotPresent` but the server uses the user-configured `args.pullPolicy`, we should respect the user here switch them around IMO. Kubelet pulls image volume images before starting any containers [1][2][3], and only then starts init containers [4]. Right now we may hit a case where the user explicitly stated `Always` but the image pull will actually use the cached image. 

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1890
[2] https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1877
[3] https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1892
[4] https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1917

pkg/importer/registry-datasource.go
```diff
@@ -84,24 +86,33 @@ func (rd *RegistryDataSource) Info() (ProcessingPhase, error) {
...
-	}
-	if size <= int64(0) {
-		//Path provided is invalid.
-		return ProcessingPhaseError, ErrInvalidPath
-	}
+	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod == string(cdiv1.RegistryPullNode) {
+		imageRootDir, err := util.ParseEnvVar(common.ImporterImageRootDir, false)
+		if imageRootDir == "" {
+			klog.Errorf("image root dir environment variable does not exist")
+			return ProcessingPhaseError, err
```
Acedus: Pretty sure this always returns nil error.

Acedus (commented):
Thanks @nav1s, generally looks good! Please consider adding UTs to exercise the new registry-datasource changes.

nav1s:
> Thanks @nav1s, generally looks good! Please consider adding UTs to exercise the new registry-datasource changes.

will do

pkg/controller/import-controller.go
nav1s: That's a good point, when writing this up I thought it will pull the image volume last

pkg/importer/registry-datasource.go
nav1s: Good catch, I will return an explicit error instead

pkg/importer/registry-datasource.go
```diff
@@ -84,24 +86,37 @@ func (rd *RegistryDataSource) Info() (ProcessingPhase, error) {
 
 // Transfer is called to transfer the data from the source registry to a temporary location.
 func (rd *RegistryDataSource) Transfer(path string, preallocation bool) (ProcessingPhase, error) {
-	rd.imageDir = filepath.Join(path, containerDiskImageDir)
-	if err := CleanAll(rd.imageDir); err != nil {
-		return ProcessingPhaseError, err
-	}
+	if pullMethod, _ := util.ParseEnvVar(common.ImporterPullMethod, false); pullMethod == string(cdiv1.RegistryPullNode) {
```
Acedus, line 99: Pretty sure we can keep the error here and just wrap it if there is one (even though there shouldn't).

Acedus (commented):
Looks good, small nit.
/cc @akalenyu

pkg/importer/registry-datasource.go
nav1s: Fair point

nav1s: Turns out errors.wrap returns nil if err is nil so I needed to add an if statement to avoid that

Acedus (approved):

tools/cdi-env-extractor/main.go
akalenyu, line 1: really high level pass; this is no longer a server (tools/cdi-containerimage-server/main.go)/application we really need. should we consider just deleting and using an existing minimal image that inlines this with bash? i don't think the overhead is worth it for keeping this app (being a pure go build and all that)

akalenyu (commented):
thank you for taking on this! really nice to simplify a key core flow

tools/cdi-env-extractor/main.go
nav1s: Will look into that

nav1s: Upon further investigation, won't we have the same problem from [here](https://github.com/kubevirt/containerized-data-importer/pull/3798#discussion_r2167128870)?

akalenyu: yeah my bad, the image suggestion is clearly not going to work since we need to spin off precisely from this one. I think it's still cleaner to delete the server app/introduce a new simple one

tools/cdi-containerimage-server/main.go
```diff
@@ -1,112 +1,24 @@
...
-			break
-		}
-		time.Sleep(time.Second)
-	}
-
-	os.Remove(*doneFile)
-	if err := server.Shutdown(context.TODO()); err != nil {
-		log.Printf("Shutdown failed: %v\n", err)
+	if err := os.WriteFile(*envFile, []byte(envString.String()), 0600); err != nil {
+		log.Fatalf("Failed creating \"env\" file: %v", err)
```
akalenyu: may need an `fsync` after the write, I'm not sure it's guaranteed without an unmount
(otherwise main container may get stale env file)

nav1s: Good point

tools/cdi-env-extractor/main.go
nav1s: not sure I entirely follow here since we still need a static binary (after doing some quick digging it seems like replacing it with a static `busybox` can have some issues), also worth noting that in it's current implementation we need the binary for [dataimportcron](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/dataimportcron-controller.go#L721).

akalenyu: yeah ofc this stays static it's just harder to reason about it being a "server" according to the package name. so I am mostly just suggesting a remove/reintroduce as `env-extractor`

nav1s: yeah make sense

nav1s: My suggested implementation is [here](https://github.com/kubevirt/containerized-data-importer/compare/e3264fb89b4621e0153106c3583793986408b1b5..56bb00865e79c608aa8dee0328c50a956ab605e0)

cwedgwood:
Coming from #4253, which reports the 60 second `Timeout waiting for file /shared/ready` failure. This PR removes that mechanism, so it would close that issue too if you want to reference it.

Some field data in case it is useful motivation: across 230 importer pods on bare metal, all 93 whose image pull took longer than 60 seconds failed their first attempt, and none of the 137 with a shorter or cached pull did. The imports succeeded once kubelet restarted the importer, so the current flow costs a spurious `ErrImportFailed` and roughly 1.5-2.5 minutes per DataVolume.

The ordering guarantee noted above, that kubelet pulls image volume images before starting any containers, is what removes the race.

tools/cdi-env-extractor/BUILD.bazel
```diff
@@ -4,24 +4,20 @@ load("@rules_pkg//:pkg.bzl", "pkg_tar")
...
-        "//pkg/common:go_default_library",
-        "//vendor/github.com/pkg/errors:go_default_library",
-    ],
 )
 
 go_binary(
-    name = "cdi-containerimage-server",
+    name = "cdi-env-extractor",
     embed = [":go_default_library"],
     pure = "on",
```
akalenyu, line 15: unrelated; maybe a small comment next to `pure=on` was always a good idea

pkg/controller/import-controller.go
```diff
@@ -1184,12 +1176,23 @@ func makeImporterVolumeSpec(args *importerPodArgs) []corev1.Volume {
...
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
 			},
-		})
+			corev1.Volume{
+				Name: "image-volume",
+				VolumeSource: corev1.VolumeSource{
+					Image: &corev1.ImageVolumeSource{
+						Reference:  args.importImage,
+						PullPolicy: corev1.PullPolicy(args.pullPolicy),
```
akalenyu, line 1195: consider a unit test for imagevolume pullpolicy respecting - can be valuable

pkg/controller/util.go
```diff
@@ -331,17 +331,18 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *corev1.Pod, te
...
-				return
+	for _, statuses := range [][]corev1.ContainerStatus{pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses} {
+		for _, status := range statuses {
+			if status.Started != nil && !(*status.Started) && status.State.Waiting != nil {
+				switch status.State.Waiting.Reason {
+				case "ImagePullBackOff", "ErrImagePull", "InvalidImageName":
+					anno[prefix+".message"] = fmt.Sprintf("%s: %s", common.ImagePullFailureText, status.Image)
+					anno[prefix+".reason"] = ImagePullFailedReason
+					return
+				}
```
akalenyu, line 342: this shakeup probably deserves a commit on it's own (i.e. `address image pull failure reporting with image volumes`)

pkg/importer/registry-datasource.go
```diff
@@ -84,24 +86,45 @@ func (rd *RegistryDataSource) Info() (ProcessingPhase, error) {
...
+		imageRootDir, err := util.ParseEnvVar(common.ImporterImageRootDir, false)
+		if err != nil {
+			wrappedErr := errors.Wrap(err, "failed to parse ImporterImageRootDir environment variable")
```
akalenyu: we kind of standardized over fmt.Errorf with %w for error wrapping, can we use that?

pkg/importer/registry-datasource.go
```diff
@@ -84,24 +86,45 @@ func (rd *RegistryDataSource) Info() (ProcessingPhase, error) {
 
 // Transfer is called to transfer the data from the source registry to a temporary location.
 func (rd *RegistryDataSource) Transfer(path string, preallocation bool) (ProcessingPhase, error) {
```
akalenyu, line 98: should we have a happy path test for node-pull Transfer

pkg/controller/import-controller.go
```diff
@@ -1247,20 +1250,35 @@ func makeImporterVolumeSpec(args *importerPodArgs) []corev1.Volume {
...
-			Command:         []string{"sh", "-c", "cp /usr/bin/cdi-containerimage-server /shared/server"},
-			VolumeMounts: []corev1.VolumeMount{
-				{
-					MountPath: "/shared",
-					Name:      "shared-volume",
+		initContainers = append(initContainers,
+			corev1.Container{
+				Name:            "init",
+				Image:           args.image,
+				ImagePullPolicy: corev1.PullIfNotPresent,
```
akalenyu: is it safe to always use PullIfNotPresent? at the very least this could clobber development with `pullPolicy: Always`

pkg/importer/registry-datasource.go
akalenyu, line 83: is it not an issue to return ProcessingPhaseTransferScratch here for nodepull?

akalenyu (commented):
great stuff, sorry it took some time to give this a full pass!

btw, did you have a chance to deploy this PR and test it against an openshift cluster? iirc some functests are skipped on pure k8s

pkg/controller/import-controller.go
Acedus: The container should always be influenced by the ImageVolume's policy AFAIK so I'd say it's fine.

akalenyu: this is the cp init container using some random ctrl plane image

Acedus: Oh I was looking at the wrong container, whoops. In that case, I believe you're right.

akalenyu:
unrelated https://github.com/kubernetes-csi/external-snapshotter/pull/1339

nav1s:
> Coming from #4253, which reports the 60 second `Timeout waiting for file /shared/ready` failure. This PR removes that mechanism, so it would close that issue too if you want to reference it.
> 
> Some field data in case it is useful motivation: across 230 importer pods on bare metal, all 93 whose image pull took longer than 60 seconds failed their first attempt, and none of the 137 with a shorter or cached pull did. The imports succeeded once kubelet restarted the importer, so the current flow costs a spurious `ErrImportFailed` and roughly 1.5-2.5 minutes per DataVolume.
> 
> The ordering guarantee noted above, that kubelet pulls image volume images before starting any containers, is what removes the race.

Thanks for the feedback @cwedgwood! I'm glad to hear that this PR should improve other aspects of CDI as well.

nav1s:
> sorry it took some time to give this a full pass!

No worries @akalenyu, thanks for dedicating time for the review!
> btw, did you have a chance to deploy this PR and test it against an openshift cluster? iirc some functests are skipped on pure k8s

I found `cluster-sync/external/README.md` for deploying to an external cluster, is that the flow you'd use for openshift testing, or is there something else I'm missing?

akalenyu:
> > btw, did you have a chance to deploy this PR and test it against an openshift cluster? iirc some functests are skipped on pure k8s
> 
> I found `cluster-sync/external/README.md` for deploying to an external cluster, is that the flow you'd use for openshift testing, or is there something else I'm missing?

Yeah. Feel free to ping me or @Acedus for tips

pkg/importer/registry-datasource.go
nav1s: That's what we need to return in order for [data-proccessor.go](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/data-processor.go#L174) to jump to `Transfer`.
I agree that `ProcessingPhaseTransferScratch` isn't the best fit here, but I couldn't find one that was more fitting form [here](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/data-processor.go#L38), should I create a new one?

edit: also worth mentioning that as far as I understand from [this](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/import-controller.go#L803), we won't create a scratch space for registry source when doing node pull.

tools/cdi-env-extractor/BUILD.bazel
nav1s: yeah that make sense

pkg/controller/import-controller.go
nav1s: will do

pkg/importer/registry-datasource.go
nav1s: sure

pkg/importer/registry-datasource.go
nav1s: yeah will add one

pkg/controller/util.go
nav1s: yeah will add one for that

nav1s:
> > > btw, did you have a chance to deploy this PR and test it against an openshift cluster? iirc some functests are skipped on pure k8s
> > 
> > 
> > I found `cluster-sync/external/README.md` for deploying to an external cluster, is that the flow you'd use for openshift testing, or is there something else I'm missing?
> 
> Yeah. Feel free to ping me or @Acedus for tips

I deployed and tested the code against an OpenShift cluster. After adding image volumes to the SCC, `dataimportcron` is working as expected. Let me know if there's anything else you'd like me to verify.

pkg/operator/controller/scc.go
```diff
@@ -69,6 +69,7 @@ func setSCC(scc *secv1.SecurityContextConstraints) {
 		secv1.FSTypeSecret,
 		secv1.FSTypeCSI,
 		secv1.FSTypeEphemeral,
+		secv1.FSTypeImage,
```
akalenyu, line 72: The unit test should be adjusted to pass https://github.com/kubevirt/containerized-data-importer/blob/ca62786fd609094824549273726933a8a35bbba5/pkg/operator/controller/controller_test.go#L273-L282

pkg/importer/registry-datasource_test.go
```diff
@@ -65,37 +68,123 @@ var _ = Describe("Registry data source", func() {
 		Entry("return Error on valid scratch space, but CopyImage failed", "invalid", "", "", "", "", true, true),
 	)
 
+	It("node pull Transfer should successfully return Convert on valid environment variables", func() {
+		diskDir := filepath.Join(tmpDir, containerDiskImageDir)
+		Expect(os.Mkdir(diskDir, 0755)).To(Succeed())
+		Expect(os.WriteFile(filepath.Join(diskDir, "disk.img"), []byte("image"), 0600)).To(Succeed())
+
+		GinkgoT().Setenv(common.ImporterPullMethod, "node")
```
Acedus: Setting environment variables as part of a unit test defeats the purpose of a unit test which is to be fully isolated in runtime (should not interact with the host at all). I'm aware that the registry-datasource test suite doesn't abide by that rule, but we should avoid increasing the tech debt.

pkg/controller/util.go
```diff
@@ -331,13 +331,15 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *corev1.Pod, te
...
 	anno[cc.AnnRunningCondition] = "false"
 
-	for _, status := range pod.Status.ContainerStatuses {
-		if status.Started != nil && !(*status.Started) && status.State.Waiting != nil {
-			switch status.State.Waiting.Reason {
-			case "ImagePullBackOff", "ErrImagePull", "InvalidImageName":
-				anno[prefix+".message"] = fmt.Sprintf("%s: %s", common.ImagePullFailureText, status.Image)
-				anno[prefix+".reason"] = ImagePullFailedReason
-				return
+	for _, statuses := range [][]corev1.ContainerStatus{pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses} {
```
Acedus: nit:

```go
var imagePullFailureReasons = []string{"ImagePullBackOff", "ErrImagePull", "InvalidImageName"}

for _, status := range slices.Concat(pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses) {
    if status.Started == nil || *status.Started || status.State.Waiting == nil {
        continue
    }
    if !slices.Contains(imagePullFailureReasons, status.State.Waiting.Reason) {
        continue
    }
    anno[prefix+".message"] = fmt.Sprintf("%s: %s", common.ImagePullFailureText, status.Image)
    anno[prefix+".reason"] = ImagePullFailedReason
    return
}
```

pkg/importer/registry-datasource_test.go
nav1s: I have adapted the implementation to avoid this [here](https://github.com/kubevirt/containerized-data-importer/compare/ca62786fd609094824549273726933a8a35bbba5..8bdc1099a706b3ed79bd5c44320e82d76356d68a)

pkg/controller/util.go
nav1s: That does look more readable :)) changed to this [here](https://github.com/kubevirt/containerized-data-importer/compare/ca62786fd609094824549273726933a8a35bbba5..8bdc1099a706b3ed79bd5c44320e82d76356d68a)

pkg/operator/controller/scc.go
nav1s: yeah forgot to run the unit tests :sweat_smile:
fixed [here](https://github.com/kubevirt/containerized-data-importer/compare/ca62786fd609094824549273726933a8a35bbba5..8bdc1099a706b3ed79bd5c44320e82d76356d68a)

pkg/importer/registry-datasource_test.go
Acedus: I was actually wrong about how `Setenv` works with Ginkgo, apologies.

pkg/importer/registry-datasource_test.go
```diff
@@ -65,37 +68,111 @@ var _ = Describe("Registry data source", func() {
...
+		Expect(ProcessingPhaseError).To(Equal(result))
+	})
+
+	It("GetTerminationMessage should return nil when ImporterEnvFile is empty", func() {
+		ds = NewRegistryDataSource("", "", "", "", "", true, string(cdiv1.RegistryPullNode), "", "")
+		termMesg := ds.GetTerminationMessage()
+		Expect(termMesg).To(BeNil())
+	})
+
+	It("Transfer should return error when imageRootDir is not set", func() {
```
Acedus, line 150: Isn't that the exact same test as :150?

pkg/importer/http-datasource_test.go
```diff
@@ -347,7 +346,7 @@ var _ = Describe("Http data source", func() {
 		})
 	})
 
-	It("GetTerminationMessage should return nil when pullMethod is not node", func() {
+	It("GetTerminationMessage should return nil", func() {
 		Expect(os.Setenv(common.ImporterPullMethod, string(cdiv1.RegistryPullPod))).To(Succeed())
```
Acedus: Is there any need to have `ImporterPullMethod` in this test file anymore?

pkg/importer/http-datasource_test.go
```diff
@@ -690,15 +649,6 @@ var _ = Describe("http pollprogress", func() {
 
 func createTestServer(imageDir string, env []string) *httptest.Server {
```
Acedus: env is no longer used?

pkg/controller/util.go
```diff
@@ -331,15 +332,18 @@ func setAnnotationsFromPodWithPrefix(anno map[string]string, pod *corev1.Pod, te
...
 
-	for _, status := range pod.Status.ContainerStatuses {
-		if status.Started != nil && !(*status.Started) && status.State.Waiting != nil {
-			switch status.State.Waiting.Reason {
-			case "ImagePullBackOff", "ErrImagePull", "InvalidImageName":
-				anno[prefix+".message"] = fmt.Sprintf("%s: %s", common.ImagePullFailureText, status.Image)
-				anno[prefix+".reason"] = ImagePullFailedReason
-				return
-			}
+	var imagePullFailureReasons = []string{"ImagePullBackOff", "ErrImagePull", "InvalidImageName"}
```
Acedus: nit: Can probably make this package-level as it'll never change.

pkg/importer/registry-datasource_test.go
```diff
@@ -65,37 +68,111 @@ var _ = Describe("Registry data source", func() {
...
+		ds = NewRegistryDataSource("", "", "", "", "", true, string(cdiv1.RegistryPullNode), "", "")
+		result, err := ds.Transfer(tmpDir, false)
+		Expect(err).To(HaveOccurred())
+		Expect(ProcessingPhaseError).To(Equal(result))
+	})
+
+	It("GetTerminationMessage should return nil when ImporterEnvFile is not set", func() {
```
Acedus, line 157: Same here with :157

pkg/importer/registry-datasource_test.go
```diff
@@ -65,37 +68,111 @@ var _ = Describe("Registry data source", func() {
...
+	DescribeTable("GetTerminationMessage should contain labels collected from the image", func(pullMethod string) {
+		ds = NewRegistryDataSource("", "", "", "", "", true, "", "", "")
+		envVariables := []string{
+			"INSTANCETYPE_KUBEVIRT_IO_DEFAULT_INSTANCETYPE=u1.small",
+			"INSTANCETYPE_KUBEVIRT_IO_DEFAULT_PREFERENCE=fedora",
+		}
+		if pullMethod == "node" {
+			envFile := path.Join(tmpDir, "env-file")
+			envString := strings.Join(envVariables, "\n")
+			err := os.WriteFile(envFile, []byte(envString), 0600)
```
Acedus, line 100: This still counts as interacting with the filesystem and thus violates a UT, as per my recent comment, I believe we _can_ use `GinkgoT().Setenv()`, sorry for the confusion.

Acedus (commented):
Looks good! Few minor comments @nav1s.

pkg/importer/registry-datasource_test.go
nav1s: yeah, it was relevant with the env variable design but is now redundant, fixed [here](https://github.com/kubevirt/containerized-data-importer/compare/8bdc1099a706b3ed79bd5c44320e82d76356d68a..3cd7bfb4be7596205ef765445172729f2366380f)
edit: after some additional thinking on the matter I recalled that environment variable can't be null so these tests were always redundant :))

pkg/importer/http-datasource_test.go
nav1s: not really, fixed [here](https://github.com/kubevirt/containerized-data-importer/compare/8bdc1099a706b3ed79bd5c44320e82d76356d68a..3cd7bfb4be7596205ef765445172729f2366380f)

pkg/importer/http-datasource_test.go
nav1s: indeed, I'm used to the linter catching this kind of stuff, not entirely sure why it didn't catch it there, fixed [here](https://github.com/kubevirt/containerized-data-importer/compare/8bdc1099a706b3ed79bd5c44320e82d76356d68a..3cd7bfb4be7596205ef765445172729f2366380f)

pkg/controller/util.go
nav1s: fixed [here](https://github.com/kubevirt/containerized-data-importer/compare/8bdc1099a706b3ed79bd5c44320e82d76356d68a..3cd7bfb4be7596205ef765445172729f2366380f)

pkg/importer/registry-datasource_test.go
nav1s: yeah, also redundant for the same reason as the previous comment, fixed [here](https://github.com/kubevirt/containerized-data-importer/compare/8bdc1099a706b3ed79bd5c44320e82d76356d68a..3cd7bfb4be7596205ef765445172729f2366380f)

pkg/importer/registry-datasource_test.go
nav1s: not sure I entirely follow here, could you elaborate on how we could test this without interacting with the filesystem?

Acedus: Maybe we can have an interface for loading the environment variables and we'll set the concrete implementation based on the pullMethod? Like something that loads the environment variables into a uniform format and then calls envToLabels? That way we can stub and unit test just the loading/parsing part of it.

cwedgwood:
This removes the timeout path from #4253—thanks. One compatibility question: does OpenShift resolve ImageStream names in `volumes[].image.reference`? The [admission code I checked](https://github.com/openshift/apiserver-library-go/blob/dc9a973f14ce11c6c251846997de360bd66245b6/pkg/admission/imagepolicy/imagereferencemutators/pods.go#L238-L275) only rewrites container images, so a namespace-local ImageStream could resolve for the extractor but fail for the volume. I haven’t tested this on OpenShift.

nav1s:
> This removes the timeout path from #4253—thanks. One compatibility question: does OpenShift resolve ImageStream names in `volumes[].image.reference`? The [admission code I checked](https://github.com/openshift/apiserver-library-go/blob/dc9a973f14ce11c6c251846997de360bd66245b6/pkg/admission/imagepolicy/imagereferencemutators/pods.go#L238-L275) only rewrites container images, so a namespace-local ImageStream could resolve for the extractor but fail for the volume. I haven’t tested this on OpenShift.

Thanks for pointing this out! I already have an openshift cluster deployed for testing this code so I will look into this
edit: after testing this out this seems to be a real issue so I will check how could this be fixed

PR #4245, opened 2026-08-13
Enable readOnlyRootFilesystem for all CDI containers

davmarro:
Enables `readOnlyRootFilesystem: true` for all CDI containers.
The change touches two central SecurityContext configuration points:
- `CreateContainer()` / `CreatePortsContainer()` in `pkg/operator/resources/utils/common.go` — covers the 4 static deployments (`cdi-operator`, `cdi-apiserver`, `cdi-controller`, `cdi-uploadproxy`)
- `SetRestrictedSecurityContext()` in `pkg/controller/common/util.go` — covers all transient worker pods (`importer`, `cloner`, `uploadserver`, `populators`)

**Prometheus TLS certs refactoring:**
Previously, `importer` and `cloner `generated self-signed TLS certs at runtime using `os.MkdirTemp`, requiring a writable `/tmp`. This PR refactors the approach: the controller now generates certs and stores them in a pod-owned Secret, mounted read-only at `/etc/prometheus-certs`

An emptyDir volume mounted at `/tmp` is added **only** where there is confirmed evidence of writes:
- **cdi-controller** — creates `/tmp/ready` for its readiness probe: [`cmd/cdi-controller/controller.go#L390`](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-controller/controller.go#L390)
- **Worker pods** (importer, cloner, uploadserver, dataimportcron) — no `/tmp` emptyDir needed because of the Prometheus TLS certs refactoring
- **vddk importer** — writes `/tmp/nbd.sock` and `/tmp/nbd.pid`: [`pkg/importer/vddk-datasource_amd64.go#L69-L70`](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/vddk-datasource_amd64.go#L69-L70)
- **ovirt-populator** — writes `/tmp/ovirt.pass` and `/tmp/ca.pem`: [`cmd/ovirt-populator/ovirt-populator.go#L70-L72`](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/ovirt-populator/ovirt-populator.go#L70-L72)

For cdi-controller the emptyDir is added explicitly in the deployment spec.
Static deployments without evidence of `/tmp` writes (cdi-operator, cdi-apiserver, cdi-uploadproxy) do **not** get an emptyDir — if any of them fails at runtime, CI will surface it and a targeted volume can be added.

Fixes #

I added some missing unit tests and also E2E coverage checking that really each container has the readOnlyFileSystem set to true.

**Release note**:

```release-note
Enable readOnlyRootFilesystem for all CDI containers
```

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -65,6 +65,7 @@ import (
 	featuregates "kubevirt.io/containerized-data-importer/pkg/feature-gates"
 	"kubevirt.io/containerized-data-importer/pkg/token"
 	"kubevirt.io/containerized-data-importer/pkg/util"
+	"kubevirt.io/containerized-data-importer/pkg/util/cert"
 	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/api"
 )
 
@@ -74,6 +75,9 @@ const (
 
 	// ScratchVolName provides a const to use for creating scratch pvc volumes in pod specs
 	ScratchVolName = "cdi-scratch-vol"
+
+	// PrometheusCertSecretSuffix is appended to the pod name to form the prometheus cert Secret name
+	PrometheusCertSecretSuffix = "-prometheus-certs"
 
 	// AnnAPIGroup is the APIGroup for CDI
 	AnnAPIGroup = "cdi.kubevirt.io"
@@ -1247,35 +1251,115 @@ func SetRestrictedSecurityContext(podSpec *corev1.PodSpec) {
 	if hasVolumeMounts {
 		podSpec.SecurityContext.FSGroup = ptr.To[int64](common.QemuSubGid)
 	}
-
-	addTmpVolume(podSpec)
 }
 
-const tmpVolumeName = "tmp-dir"
+// PrometheusCertSecretName returns the Secret name for a pod's prometheus certs.
+func PrometheusCertSecretName(podName string) string {
+	return podName + PrometheusCertSecretSuffix
+}
 
-// addTmpVolume adds an emptyDir volume at /tmp to support readOnlyRootFilesystem
-func addTmpVolume(podSpec *corev1.PodSpec) {
-	for _, v := range podSpec.Volumes {
-		if v.Name == tmpVolumeName {
-			return
-		}
-	}
+// AppendPrometheusCertVolume adds the prometheus cert Secret volume and mount to a pod spec.
+func AppendPrometheusCertVolume(podSpec *corev1.PodSpec, podName string) {
 	podSpec.Volumes = append(podSpec.Volumes, corev1.Volume{
-		Name: tmpVolumeName,
+		Name: common.PrometheusCertVolName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: PrometheusCertSecretName(podName),
+			},
+		},
+	})
+	mount := corev1.VolumeMount{
+		Name:      common.PrometheusCertVolName,
+		MountPath: common.PrometheusCertDir,
+		ReadOnly:  true,
+	}
+	for i := range podSpec.Containers {
+		podSpec.Containers[i].VolumeMounts = append(podSpec.Containers[i].VolumeMounts, mount)
+	}
+}
+
+// AppendTmpVolume adds an emptyDir volume mounted at /tmp to a pod spec.
+func AppendTmpVolume(podSpec *corev1.PodSpec) {
+	podSpec.Volumes = append(podSpec.Volumes, corev1.Volume{
+		Name: common.TmpVolumeName,
 		VolumeSource: corev1.VolumeSource{
 			EmptyDir: &corev1.EmptyDirVolumeSource{},
 		},
 	})
-	tmpMount := corev1.VolumeMount{
-		Name:      tmpVolumeName,
-		MountPath: "/tmp",
-	}
-	for i := range podSpec.InitContainers {
-		podSpec.InitContainers[i].VolumeMounts = append(podSpec.InitContainers[i].VolumeMounts, tmpMount)
+	mount := corev1.VolumeMount{
+		Name:      common.TmpVolumeName,
+		MountPath: common.TmpMountPath,
 	}
 	for i := range podSpec.Containers {
-		podSpec.Containers[i].VolumeMounts = append(podSpec.Containers[i].VolumeMounts, tmpMount)
+		podSpec.Containers[i].VolumeMounts = append(podSpec.Containers[i].VolumeMounts, mount)
 	}
+}
+
+// CreatePrometheusCertSecret creates a Secret containing a self-signed TLS cert/key pair
+// for the prometheus metrics endpoint of a worker pod. It must be called BEFORE creating
+// the pod so the Secret is available when the kubelet mounts the volume.
+// The returned Secret does not have an OwnerReference yet; call SetPrometheusCertSecretOwnerRef
+// after the pod is created to set it.
+func CreatePrometheusCertSecret(ctx context.Context, c client.Client, podName, namespace string, installerLabels map[string]string) error {
+	certBytes, keyBytes, err := cert.GenerateSelfSignedCertKey(podName, nil, nil)
+	if err != nil {
+		return fmt.Errorf("generating prometheus cert for pod %s/%s: %w", namespace, podName, err)
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      PrometheusCertSecretName(podName),
+			Namespace: namespace,
+			Labels: map[string]string{
+				common.CDILabelKey:        common.CDILabelValue,
+				common.PrometheusLabelKey: common.PrometheusLabelValue,
+			},
+		},
+		Data: map[string][]byte{
+			corev1.TLSCertKey:       certBytes,
+			corev1.TLSPrivateKeyKey: keyBytes,
+		},
+	}
+
+	if installerLabels != nil {
+		util.SetRecommendedLabels(secret, installerLabels, common.CDIControllerName)
+	}
+
+	if err := c.Create(ctx, secret); err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			klog.Warningf("Prometheus cert secret for pod %s/%s already exists", namespace, podName)
+			return nil
+		}
+		return fmt.Errorf("creating prometheus cert secret for pod %s/%s: %w", namespace, podName, err)
+	}
+	return nil
+}
+
+// SetPrometheusCertSecretOwnerRef updates the prometheus cert Secret to add an OwnerReference
+// to the given pod, so the Secret is garbage-collected when the pod is deleted.
+// Must be called AFTER the pod is created (since it needs the pod's UID).
+func SetPrometheusCertSecretOwnerRef(ctx context.Context, c client.Client, pod *corev1.Pod) error {
+	secret := &corev1.Secret{}
+	secretName := PrometheusCertSecretName(pod.Name)
+	if err := c.Get(ctx, types.NamespacedName{Name: secretName, Namespace: pod.Namespace}, secret); err != nil {
+		return fmt.Errorf("getting prometheus cert secret %s/%s: %w", pod.Namespace, secretName, err)
+	}
+
+	secret.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion:         "v1",
+			Kind:               "Pod",
+			Name:               pod.Name,
+			UID:                pod.GetUID(),
+			BlockOwnerDeletion: ptr.To(true),
+			Controller:         ptr.To(true),
+		},
+	}
+
+	if err := c.Update(ctx, secret); err != nil {
+		return fmt.Errorf("setting owner reference on prometheus cert secret %s/%s: %w", pod.Namespace, secretName, err)
+	}
+	return nil
 }
 
 // SetNodeNameIfPopulator sets NodeName in a pod spec when the PVC is being handled by a CDI volume populator
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -1230,6 +1230,7 @@ func SetRestrictedSecurityContext(podSpec *corev1.PodSpec) {
 			container.SecurityContext.AllowPrivilegeEscalation = ptr.To[bool](false)
 			container.SecurityContext.RunAsNonRoot = ptr.To[bool](true)
 			container.SecurityContext.RunAsUser = ptr.To[int64](common.QemuSubGid)
+			container.SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
 			if len(container.VolumeMounts) > 0 {
 				hasVolumeMounts = true
 			}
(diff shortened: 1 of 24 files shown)


pkg/operator/resources/namespaced/controller_test.go
akalenyu, line 1: So we normally test operator reconcile loop at pkg/operator/controller/controller_test.go which covers the entire path. We may be able to piggy back on existing tests too

tests/operator_test.go
akalenyu, line 1: So normally we try to avoid functests unless strictly required. In this case I'd argue there is really no benefit over unit tests, and the framework to assert on the deployments/reconcile loop exists

akalenyu (commented):
Really high level pass, makes sense! thank you
Just few testing considerations before I give this a full technical pass 🙏

pkg/operator/resources/namespaced/controller_test.go
davmarro: Thank you, I didn't realize the test file was already there. 
Moved the tests to `pkg/operator/controller/controller_test.go`.
Thanks

tests/operator_test.go
davmarro: I totally agree with you.
Changes deleted.
Thanks

akalenyu (commented):
Can you clarify how mounting certs is an issue? I assume the cloner certs for instance are a legitimate non root configmap mount?

davmarro:
> Can you clarify how mounting certs is an issue? I assume the cloner certs for instance are a legitimate non root configmap mount?

Hi @akalenyu, for the cloner and the importer the certificates creation is actually related to Prometheus - not to the services itself.
- cdi-importer — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-importer/importer.go#L102](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-importer/importer.go#L102)
- cdi-cloner — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-cloner/clone-source.go#L94](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-cloner/clone-source.go#L94)

If you see the code, both lines are included in a `startPrometheus` method that generates at runtime the certificates needed for the Prometheus server to start.

akalenyu:
> > Can you clarify how mounting certs is an issue? I assume the cloner certs for instance are a legitimate non root configmap mount?
> 
> Hi @akalenyu, for the cloner and the importer the certificates creation is actually related to Prometheus - not to the services itself.
> 
> * cdi-importer — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-importer/importer.go#L102](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-importer/importer.go#L102)
> * cdi-cloner — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-cloner/clone-source.go#L94](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-cloner/clone-source.go#L94)
> 
> If you see the code, both lines are included in a `startPrometheus` method that generates at runtime the certificates needed for the Prometheus server to start.

I see, it's a throwaway in memory cert. We should probably get rid of it at some point in favor of mounted internal cdi certs. Until then, maybe just create/mount the throwaway cert through the controller?
emptyDir/runtime generated feels off

davmarro:
> > > Can you clarify how mounting certs is an issue? I assume the cloner certs for instance are a legitimate non root configmap mount?
> > 
> > 
> > Hi @akalenyu, for the cloner and the importer the certificates creation is actually related to Prometheus - not to the services itself.
> > 
> > * cdi-importer — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-importer/importer.go#L102](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-importer/importer.go#L102)
> > * cdi-cloner — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-cloner/clone-source.go#L94](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-cloner/clone-source.go#L94)
> > 
> > If you see the code, both lines are included in a `startPrometheus` method that generates at runtime the certificates needed for the Prometheus server to start.
> 
> I see, it's a throwaway in memory cert. We should probably get rid of it at some point in favor of mounted internal cdi certs. Until then, maybe just create/mount the throwaway cert through the controller? emptyDir/runtime generated feels off

I agree on this, the current solution (emptyDir + certsDir) it's definitely a workaround to let the server just work and have the RORFS to true.
We can change the logic and generate in-memory certs without writing them to disk but I don't think this refactor can be part of this PR - it's definitely out of scope.
I can do it after the PR is merged in a proper PR with its own scope. 
Do you agree?

akalenyu:
> > > > Can you clarify how mounting certs is an issue? I assume the cloner certs for instance are a legitimate non root configmap mount?
> > > 
> > > 
> > > Hi @akalenyu, for the cloner and the importer the certificates creation is actually related to Prometheus - not to the services itself.
> > > 
> > > * cdi-importer — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-importer/importer.go#L102](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-importer/importer.go#L102)
> > > * cdi-cloner — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-cloner/clone-source.go#L94](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-cloner/clone-source.go#L94)
> > > 
> > > If you see the code, both lines are included in a `startPrometheus` method that generates at runtime the certificates needed for the Prometheus server to start.
> > 
> > 
> > I see, it's a throwaway in memory cert. We should probably get rid of it at some point in favor of mounted internal cdi certs. Until then, maybe just create/mount the throwaway cert through the controller? emptyDir/runtime generated feels off
> 
> I agree on this, the current solution (emptyDir + certsDir) it's definitely a workaround to let the server just work and have the RORFS to true. We can change the logic and generate in-memory certs without writing them to disk but I don't think this refactor can be part of this PR - it's definitely out of scope. I can do it after the PR is merged in a proper PR with its own scope. Do you agree?

Yeah I am just suggesting you shift the dummy cert creation to the controller and mount it properly instead of the emptydir, in preparation for future refactor. Then you also get rid of runtime cert creation

davmarro:
> > > > > Can you clarify how mounting certs is an issue? I assume the cloner certs for instance are a legitimate non root configmap mount?
> > > > 
> > > > 
> > > > Hi @akalenyu, for the cloner and the importer the certificates creation is actually related to Prometheus - not to the services itself.
> > > > 
> > > > * cdi-importer — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-importer/importer.go#L102](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-importer/importer.go#L102)
> > > > * cdi-cloner — os.MkdirTemp("", "certsdir") defaults to /tmp: [cmd/cdi-cloner/clone-source.go#L94](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-cloner/clone-source.go#L94)
> > > > 
> > > > If you see the code, both lines are included in a `startPrometheus` method that generates at runtime the certificates needed for the Prometheus server to start.
> > > 
> > > 
> > > I see, it's a throwaway in memory cert. We should probably get rid of it at some point in favor of mounted internal cdi certs. Until then, maybe just create/mount the throwaway cert through the controller? emptyDir/runtime generated feels off
> > 
> > 
> > I agree on this, the current solution (emptyDir + certsDir) it's definitely a workaround to let the server just work and have the RORFS to true. We can change the logic and generate in-memory certs without writing them to disk but I don't think this refactor can be part of this PR - it's definitely out of scope. I can do it after the PR is merged in a proper PR with its own scope. Do you agree?
> 
> Yeah I am just suggesting you shift the dummy cert creation to the controller and mount it properly instead of the emptydir, in preparation for future refactor. Then you also get rid of runtime cert creation

Ok, I'm going to shift the certs creation to the controller avoiding the creation of the tmpdir.
Then I'll create a new task to use the project CA to generate valid certs instead of self-signed ones.

davmarro:
@akalenyu completed the refactor with this [commit](https://github.com/kubevirt/containerized-data-importer/pull/4245/changes/8240f3374e113d47478fca37dc020a64c168e349).

pkg/controller/common/util.go
```diff
@@ -1248,6 +1253,115 @@ func SetRestrictedSecurityContext(podSpec *corev1.PodSpec) {
...
+		Name:      common.TmpVolumeName,
+		MountPath: common.TmpMountPath,
+	}
+	for i := range podSpec.Containers {
+		podSpec.Containers[i].VolumeMounts = append(podSpec.Containers[i].VolumeMounts, mount)
+	}
+}
+
+// CreatePrometheusCertSecret creates a Secret containing a self-signed TLS cert/key pair
+// for the prometheus metrics endpoint of a worker pod. It must be called BEFORE creating
```
awels: Just a nitpick on the comment, but why does the secret have to be created before the pod? You should be able to create the pod first (it remains pending) until the secret is created. At which point the pod can be properly started, the order should NOT matter.

pkg/controller/common/util.go
```diff
@@ -1248,6 +1253,115 @@ func SetRestrictedSecurityContext(podSpec *corev1.PodSpec) {
...
+		}
+		return fmt.Errorf("creating prometheus cert secret for pod %s/%s: %w", namespace, podName, err)
+	}
+	return nil
+}
+
+// SetPrometheusCertSecretOwnerRef updates the prometheus cert Secret to add an OwnerReference
+// to the given pod, so the Secret is garbage-collected when the pod is deleted.
+// Must be called AFTER the pod is created (since it needs the pod's UID).
+func SetPrometheusCertSecretOwnerRef(ctx context.Context, c client.Client, pod *corev1.Pod) error {
```
awels: This feels weird, assuming we can create the pod first, the pod gets the UID needed even if it is pending, then we could create the secret like above with the UID already instead of creating the secret first, then the pod, and then update the secret with the ownerRef. In this case the only reason we have an order is that the ownerRef needs the UID of the pod.

pkg/controller/common/util_test.go
```diff
@@ -490,3 +492,232 @@ var _ = Describe("IsWebhookPvcRenderingEnabled", func() {
...
+			Containers: []v1.Container{
+				{Name: "main"},
+			},
+			InitContainers: []v1.Container{
+				{Name: "init"},
+			},
+		}
+	})
+
+	DescribeTable("should enforce container-level SecurityContext fields",
```
awels, line 510: This is not wrong, but it feels weird. We have all these entries in the DescribeTable, but I don't understand why we don't have a single It with all the checks? We do eventually get the same checks, so logically it is not wrong, but I had to re-read this a few times before I understood what it was doing.

pkg/controller/common/util_test.go
```diff
@@ -490,3 +492,232 @@ var _ = Describe("IsWebhookPvcRenderingEnabled", func() {
...
+		pod = &v1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "test-pod",
+				Namespace: "default",
+				UID:       types.UID("test-uid"),
+			},
+		}
+	})
+
+	It("should create a secret with correct name and data (no OwnerRef)", func() {
```
awels: See comment in util.go. I think this should not happen, we should create the secret after the pod and pass the pod into the function the creates the secret.

pkg/controller/common/util.go
davmarro: An earlier version of this PR did exactly what you're suggesting: it created the Pod first and then created the Prometheus Secret using the Pod UID. You can see it [here](https://github.com/kubevirt/containerized-data-importer/commit/3e608d263298f041512c4313e7fb0a882b14f406#diff-8593ac8a9811d5f656c812921bc66e26d3743a13fa77575a27a50ca32a4297bbR527) where `EnsurePrometheusCertSecret()` was called only after the Pod had been created.
The issue was that during the [E2E runs](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/4245/pull-containerized-data-importer-e2e-nfs/2092200389968400384) I observed worker Pods stuck in ContainerCreating with FailedMount events such as secret "<pod>-prometheus-certs" not found and secret-cache timeout errors.
Now a required Secret volume prevents the containers from starting until it exists, and creating the Pod first deliberately introduces that FailedMount/retry window. For this reason I believe that it is cleaner and more deterministic to avoid that window completely because we know both the Pod name and therefore the Secret name before creating the Pod.
So I would prefer to keep the current ordering: create Secret -> create Pod -> set Secret OwnerReference

Just one thing I noticed, I can make setting the OwnerReference idempotent and ensure it again when reconciling an already existing Pod. That way a transient failure in the final update cannot leave the Secret without its owner.

pkg/controller/common/util.go
davmarro: As mentioned [here](https://github.com/kubevirt/containerized-data-importer/pull/4245#discussion_r4070621659), creating the Pod first would allow us to create the Secret with its OwnerReference immediately, but that is the ordering we previously tried and it exposes the Pod to the missing-Secret/FailedMount window.I would prefer to keep the current ordering.

I totally agree that the final OwnerReference update can be made more robust. I'll make it idempotent so that if the OwnerReference update fails transiently after the Pod has been created, a subsequent reconcile repairs it.

pkg/controller/common/util_test.go
davmarro: Yes, you're right. Better to read and maintain in this way.

pkg/controller/common/util_test.go
davmarro: See the thread [here](https://github.com/kubevirt/containerized-data-importer/pull/4245#discussion_r4070621659)

pkg/controller/common/util.go
awels: So I analyzed that run, and most of the failure can be attributed to missing /tmp not the fact that secret was missing. So I still think we should create the pod first and then the secret with the UID of the pod like you did before, and fix the missing /tmp which you likely already did.

davmarro: Ok, I'm retrying this approach [here](https://github.com/kubevirt/containerized-data-importer/pull/4245/commits/bfeb0f377c6ebb7f4182da7bf0fbead0d1bb42c8). Let's see what happens with the CI. Maybe I didn't understand correctly the errors at the time. Thanks

davmarro: You were right @awels! Thanks for spotting it, really good catch!
Here's the commit https://github.com/kubevirt/containerized-data-importer/pull/4245/commits/bfeb0f377c6ebb7f4182da7bf0fbead0d1bb42c8

davmarro:
Hey @awels, thanks again for your suggestion! I restored the previous flow: the worker pod is created first and then the secret.
I also made the flow a bit more resilient to partial failures. Now the importer and clone reconciliation loops calls the `EnsurePrometheusCertSecret`: in this way when the pod is created but the secret creation fails, on the next reconciliation the missing secret is created using the existing pod.

PR #4255, opened 2026-08-23
datavolume: emit event when PVC size is raised to minimum supported size

AdyCohen1:
  When a DataVolume/PVC is requested smaller than the storage profile's minimumSupportedPvcSize, CDI silently grows it to the minimum,    
  leaving users confused about the final size. This PR emits a Normal MinimumPVCSizeApplied event on the DataVolume explaining the change.

 **Before this PR:**
PVC is created larger than requested, with no event or explanation.

 **After this PR:**
 kubectl describe dv shows an event, e.g. "The requested storage size 1Gi is smaller than the minimum supported PVC size for storage class "my-sc"; PVC will be created with 4Gi".

**References:**
https://redhat.atlassian.net/browse/CNV-64649

  ```release-note
  A DataVolume now emits a "MinimumPVCSizeApplied" event when its requested size is raised to the storage class's minimum supported PVC   
  size.
  ```

### DIFF
--- a/pkg/controller/datavolume/import-controller_test.go
+++ b/pkg/controller/datavolume/import-controller_test.go
@@ -219,6 +219,35 @@ var _ = Describe("All DataVolume Tests", func() {
 			Expect(val).To(Equal(string(dv.UID)))
 		})
 
+		It("Should record a MinimumPVCSizeApplied event and grow the PVC when the requested size is below the storage profile minimum", func() {
+			scName := "testSC"
+			sc := CreateStorageClass(scName, nil)
+			sp := createStorageProfile(scName, []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}, corev1.PersistentVolumeBlock)
+			sp.Annotations = map[string]string{AnnMinimumSupportedPVCSize: "4Gi"}
+
+			storageSpec := &cdiv1.StorageSpec{
+				StorageClassName: &scName,
+				Resources: corev1.VolumeResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceStorage: resource.MustParse("1Gi"),
+					},
+				},
+			}
+			dv := createDataVolumeWithStorageAPI("test-dv", metav1.NamespaceDefault,
+				&cdiv1.DataVolumeSource{HTTP: &cdiv1.DataVolumeSourceHTTP{}}, storageSpec)
+
+			reconciler = createImportReconciler(dv, sc, sp)
+			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}})
+			Expect(err).ToNot(HaveOccurred())
+
+			pvc := &corev1.PersistentVolumeClaim{}
+			err = reconciler.client.Get(context.TODO(), types.NamespacedName{Name: "test-dv", Namespace: metav1.NamespaceDefault}, pvc)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pvc.Spec.Resources.Requests.Storage().Cmp(resource.MustParse("4Gi"))).To(Equal(0))
+
+			Eventually(reconciler.recorder.(*record.FakeRecorder).Events).Should(Receive(ContainSubstring(MinimumPVCSizeApplied)))
+		})
+
 		It("Should create a PVC on a valid import DV without delayed annotation then add on success", func() {
 			dv := NewImportDataVolume("test-dv")
 			AddAnnotation(dv, "foo", "bar")
--- a/pkg/controller/datavolume/util.go
+++ b/pkg/controller/datavolume/util.go
@@ -53,12 +53,26 @@ const (
 
 	// MessageErrStorageClassNotFound provides a const to indicate the PVC spec is missing accessMode and no storageClass to choose profile
 	MessageErrStorageClassNotFound = "PVC spec is missing accessMode and no storageClass to choose profile"
+
+	// MinimumPVCSizeApplied is the event reason emitted when a PVC is grown to the storage profile's minimum supported size
+	MinimumPVCSizeApplied = "MinimumPVCSizeApplied"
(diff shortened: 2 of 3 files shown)


arnongilboa:
@AdyCohen1 add a release note
/lgtm

pkg/controller/datavolume/util.go
```diff
@@ -305,7 +310,7 @@ func hasCloneSourceRef(pvc *v1.PersistentVolumeClaim) bool {
 	return dsRef != nil && dsRef.APIGroup != nil && *dsRef.APIGroup == cc.AnnAPIGroup && dsRef.Kind == cdiv1.VolumeCloneSourceRef && dsRef.Name != ""
 }
 
-func renderPvcSpecVolumeSize(client client.Client, pvcSpec *v1.PersistentVolumeClaimSpec, isClone bool, log *logr.Logger) error {
+func renderPvcSpecVolumeSize(client client.Client, recorder record.EventRecorder, dv *cdiv1.DataVolume, pvcSpec *v1.PersistentVolumeClaimSpec, isClone bool, log *logr.Logger) error {
```
akalenyu: I feel its a little off for a pure computation function to need to know about event infrastructure; what if we snapshot the size before the call and compare after?

akalenyu (commented):
Big thanks for implementing event emitting for this! noble cause indeed

pkg/controller/datavolume/util.go
AdyCohen1: snapshotting the whole call would also trigger on overhead growth. another option is to modify `renderPvcSpecVolumeSize` to just return a bool indicating whether the size reached the minimum. the caller (`pvcFromStorage`) already has the recorder + DV, so it can emit the event based on that bool. what do you think?

akalenyu: can also add a field in the render state for this purpose

arnongilboa: absolutely good catch @akalenyu !

AdyCohen1: the changes have been made

pkg/controller/datavolume/util_test.go
```diff
@@ -274,6 +309,35 @@ var _ = Describe("renderPvcSpec", func() {
 		Entry("fallback to k8s default when accessMode has no matching volumeMode", nil, &rwx, nil, &rwx, nil),
 		Entry("use the passed volumeMode and accessMode even if not in storageProfile", &filesystem, &rwx, &filesystem, &rwx, nil),
 	)
+
+	It("Should record a MinimumPVCSizeApplied event when the size is raised to the minimum", func() {
```
akalenyu: This test is approaching a full reconcile loop test model, maybe just take it out to import_controller_test to test the whole loop instead of just `renderPvcSpec`?

pkg/controller/datavolume/util.go
akalenyu, line 133: TBH, what I had in mind was more similar to 
```go
syncState.pvcSpec, syncState.renderResult, err = renderPvcSpec(r.client, log, syncState.dvMutated, syncState.pvc)
```
and then call event from regular controller paths. I understand it's going to have bigger scope, so I am not going to push too hard

@arnongilboa wdyt?

arnongilboa: we are using almost all syncState fields, so why inflate the api and not simplify it to something like
`err = renderPvcSpec(r.client, log, &syncState)`?

akalenyu: somewhat breaks the separation of concerns

arnongilboa: This helper func only caller is `(r *ReconcilerBase) syncDvPvcState()` returning `dvSyncState`, so why this change breaks the separation of concerns? in addition I guess all the rendering funcs should move out of `util.go` to `controller-base.go` or a dedicated `render.go`.

akalenyu: the render functions receives dv/pvcs which is logical for a renderind function, as soon as we start passing sync state we mix the render concept with controller machinery

arnongilboa: ok, I won't fight for that... but since the controller is the only user of this rendering func, I see no problem that it will even be a method of the controller

akalenyu: huh really i didnt realize it was the only user, i thought there was reusing from webhook

arnongilboa: The pvc mutating webhook is using RenderPvc which is a bit differet as it has no DV.
https://github.com/kubevirt/containerized-data-importer/blob/c07cdb6fbd18976c44f65f1e4c0deae03e33e718/pkg/controller/datavolume/util.go#L64

pkg/controller/datavolume/util_test.go
AdyCohen1: Good point, I made the changes

pkg/controller/datavolume/import-controller_test.go
```diff
@@ -219,6 +219,46 @@ var _ = Describe("All DataVolume Tests", func() {
...
+			found := false
+			for done := false; !done; {
+				select {
+				case event := <-reconciler.recorder.(*record.FakeRecorder).Events:
+					if strings.Contains(event, MinimumPVCSizeApplied) {
+						found = true
+					}
+				default:
+					done = true
+				}
```
akalenyu: can't we follow the event asserting pattern in this suite? https://github.com/kubevirt/containerized-data-importer/blob/335c3d5fcc3d84515ca2376972611efa567d5a92/pkg/controller/datavolume/import-controller_test.go#L1375-L1380

akalenyu (commented):
close!

pkg/controller/datavolume/import-controller_test.go
AdyCohen1: Used Gomega's Should(Receive(...)) instead of a manual drain loop:
`Eventually(reconciler.recorder(*record.FakeRecorder).Events).Should(Receive(ContainSubstring(MinimumPVCSizeApplied)))
`

PR #4261, opened 2026-08-27
controller, clone, planner: fix race when PVC capacity is unavailable

Acedus:
When cloning from a source DataVolume that is still importing, the source PVC exists but its `Status.Capacity` is not yet populated. `validateAdvancedClonePVC` treats this as a hard error, setting `AnnClonePhase=Error` and triggering exponential backoff. This misclassifies a transient pending condition as an error, and the backoff can delay recovery past timeout windows.

Check whether `sourceClaim.Status.Capacity` is populated before calling `validateAdvancedClonePVC`. If not, follow the missing-source-PVC path, set `AnnClonePhase=Pending` and requeue after 5 seconds, deferring strategy selection until the source PVC is ready.

Fixes #

Some additional context on the race: whenever a clone DV is created while its source is still importing, the source PVC exists but `Status.Capacity` isn't populated until the import completes and the PVC binds. The window between PVC creation and capacity population can span minutes for large images, so this isn't necessarily a narrow edge case. We've noticed this failure when a VM was being cloned from a DV that was only 1 second into its HTTP import.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/clone/planner.go
+++ b/pkg/controller/clone/planner.go
@@ -359,6 +359,10 @@ func (p *Planner) computeStrategyForSourcePVC(ctx context.Context, args *ChooseS
 	res.Strategy = strategy
 	if strategy == cdiv1.CloneStrategySnapshot ||
 		strategy == cdiv1.CloneStrategyCsiClone {
+		if _, ok := sourceClaim.Status.Capacity[corev1.ResourceStorage]; !ok {
+			args.Log.V(3).Info("Source PVC capacity not yet available, requeueing")
+			return nil, nil
+		}
 		if err := p.validateAdvancedClonePVC(ctx, args, res, sourceClaim); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


Acedus:
fyi @SamAlber

akalenyu (commented):
doesn't `isSourceReadyToClone` handle this by ensuring src is populated?
/cc @alromeros

Acedus:
> doesn't `isSourceReadyToClone` handle this by ensuring src is populated? /cc @alromeros

First I'll note that we noticed it in CDI 1.62, but I checked and couldn't really see any logical differences. I agree that it should be handled by that, but I also think that the added safety net is cheap and makes sense regardless.

alromeros:
> doesn't `isSourceReadyToClone` handle this by ensuring src is populated? /cc @alromeros

We've had issues in this area in the past that led us to introduce this failure to force a reconcile: https://github.com/kubevirt/containerized-data-importer/blob/c07cdb6fbd18976c44f65f1e4c0deae03e33e718/pkg/controller/clone/planner.go#L509, so I don't think `isSourceReadyToClone` provides full protection against this. However, with the clone populator watches introduced in https://github.com/kubevirt/containerized-data-importer/pull/3639 I don't think forcing the error should be necessary anymore.

akalenyu:
> > doesn't `isSourceReadyToClone` handle this by ensuring src is populated? /cc @alromeros
> 
> First I'll note that we noticed it in CDI 1.62, but I checked and couldn't really see any logical differences. I agree that it should be handled by that, but I also think that the added safety net is cheap and makes sense regardless.



> > doesn't `isSourceReadyToClone` handle this by ensuring src is populated? /cc @alromeros
> 
> We've had issues in this area in the past that led us to introduce this failure to force a reconcile:
> 
> https://github.com/kubevirt/containerized-data-importer/blob/c07cdb6fbd18976c44f65f1e4c0deae03e33e718/pkg/controller/clone/planner.go#L509
> 
> , so I don't think `isSourceReadyToClone` provides full protection against this. However, with the clone populator watches introduced in #3639 I don't think forcing the error should be necessary anymore.

yeah I don't see an obvious issue with the populator fix but it does seem like isSourceReadyToClone being limited to sizeless clones is a bug on it's own
(I can't dive deep to say exactly what breaks but I imagine the size is just one example)

PR #4262, opened 2026-08-27
controller: constrain VDDK importer pods to amd64 nodes

Acedus:
The VDDK library only ships x86_64 binaries, so VDDK importer pods can only run on amd64 nodes. Without an explicit node selector, the scheduler could place them on non-amd64 nodes in multi-arch clusters, causing runtime issues.

Set kubernetes.io/arch=amd64 on VDDK importer pods to make the requirement explicit.

Fixes #

**Release note**:

```release-note
NONE
```


akalenyu (commented):
/approve

akalenyu:
fyi @mnecas - not sure if we ever had folks hit this but it seems very real

PR #4263, opened 2026-08-31
MTV-6604 | Add nfc support

mnecas:
Use nbdkit-nfc for vsphere disk imports when the plugin is in the importer image or copied to /opt by the init image. Skip the sidecar when the init image is already the importer.

Sidecar emptyDir hides /opt in the init image; the plugin must be copied into /opt at runtime.

**Release note**:

```release-note
Add nbdkit-nfc support for vsphere disk imports.
```


awels (approved):
/approve

